PR #239, opened 2015-01-10
Add message if invoked without argument

prinzdezibel:


prinzdezibel:
This PR as it stands is not good. I didn't realize that METEOR_SETTINGS is a env variable that is not honored by the meteor app. (Rather it is used when a bundled meteor app is executed as node bundle.) 

I don't like at all the fact, that I'm supposed to set up debug & production environment in different ways.

PR #240, opened 2015-01-11
tests directories and packages

ramusus:
new directories for showing velocity tests from this PR https://github.com/reactioncommerce/reaction-core/pull/72

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,6 @@ npm-debug.log
 *.sublime-workspace
 
 reaction-dev.sh
+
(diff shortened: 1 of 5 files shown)


aaronjudd:
I changed the sym links to `ln -s ../packages/reaction-core/tests/mocha mocha`  as generally I either clone directly into packages, or `ln -s package packages/`  -ie - using the repo name, rather than the package name.  Just one less step, and I think that that's the natural order, vs renaming the directories to be package names.

PR #338, opened 2015-03-09
String shecking — products.coffee

psy21d:
\reaction-core\server\methods\products\products.coffee

replace all checks 
    check SomeThing, String
to
    check SomeThing, Match.OneOf(String, null, undefined)

This dummies just working and help to prove checking at many live problem points now


aldeed:
@psy21d @aaronjudd, why is this a general PR merging dev into master but has a title and description that suggests it's something else?

aaronjudd:
@aldeed not sure, which is one reason I haven't merged it.  @psy21d I think I should close this request, as the string checking should all be in core (and also not ready to merge to master from development).

psy21d:
Thanks!

Issue #61, opened 2014-05-13
Wire up social share features


aaronjudd:
Wire up client sharing for Twitter, Facebook, Pinterest and use the message stored in product data.  Wire default messages if no message provided.  Remove instagram, replace with G+

aaronjudd:
@lovetostrike this might be the right issue to log your thoughts on shareIt implementation

lovetostrike:
I have already added support for pinterest on my local shareit. And the way I see it, we should try to refactor social sharing buttons/functions on reaction to loop over an array that the user provides.

bstocks:
As part of this task, will there be an admin interface for authenticating into social accounts at the shop level? If so, then an additional feature would be to automatically add relevant features to the share buttons such as including the company twitter handle, etc.

aaronjudd:
@lovetostrike you can now use the new package registry to provide the templates to the social blocks, 
see https://github.com/reactioncommerce/reaction-core/blob/development/docs/packages.md#reactioncoreregisterpackage

Where in the core templates we'd add:

``` 
{{#each reactionApps provides="social"}}
     {{> dynamicTemplate template=template}}
{{/each}}
```

lovetostrike:
ok, I will update the PR to work with the new package system. That's neat.

aaronjudd:
@lovetostrike thoughts on reaction-social: 

The advantage of one package for social is that we can have the whole codebase in one repo, but easily break into separate dashboards if we want it to look like multiple packages. easier maintenance of the social package this way I think - even though I see this becoming much larger in functionality / scope. Might be nice to not have to worry about multiple packages.  Community wise, we'd want to encourage PR for additional social channels, rather than a new package if we take the single package approach.

With multiple repo / packages, we could add account-ui as a dependency - there will be more maintenance (publishing etc), but it will also be easier for developers to copy a simple social package and write their own without worrying about the scope/functionality of other social packages. ie: probably an easier implementation than what you had to do in https://github.com/lovetostrike/reaction-social/blob/master/client/templates/social.coffee#L10.  We might want to move shared functionality into core with this, or we'd need to have some shared (reaction-social) package in addition to the specific channels (as a weak dependency).

[...]

PR #352, opened 2015-03-22
social sharing for footer

lovetostrike:
Added reaction-social package and replaced manual social links with customized package settings. Reaction-social package: https://github.com/lovetostrike/reaction-social.

https://github.com/reactioncommerce/reaction/issues/61

### DIFF
--- a/client/templates/layout/footer/footer.html
+++ b/client/templates/layout/footer/footer.html
@@ -32,10 +32,9 @@
 
       <div class="col-xs-5 col-sm-5 col-md-5 col-lg-5">
       <ul class="list-unstyled">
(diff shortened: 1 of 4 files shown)


client/templates/layout/footer/footer.html
```diff
@@ -32,12 +32,11 @@
 
       <div class="col-xs-5 col-sm-5 col-md-5 col-lg-5">
       <ul class="list-unstyled">
-          <li><a href="" class="fa fa-twitter-square fa-2x pull-right" data-event-category="link" data-event-action="link-click" data-event-label="external link click" data-event-value="twitter"></a></li>
```
lovetostrike: It seems like I might need to put extra options in reactio-social package for these analytics events class.

PR #436, opened 2015-07-29
Update Dockerfile

faddat:
It works now :)!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,7 @@
 #
 ##############################################################
 
-FROM mongo:latest
+FROM ubuntu:14.04
 MAINTAINER Aaron Judd <hello@reactioncommerce.com>
 
 ENV DEBIAN_FRONTEND noninteractive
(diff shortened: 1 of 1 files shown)


faddat:
would be good to make sure to use the environment variables to set your username & password,   usage example:

docker build -t reaction .
docker run reaction -e METEOR_EMAIL="youradmin@yourdomain.com" \
-e METEOR_USER="admin" \
-e METEOR_AUTH="password" \
--name reaction \
-t

faddat:
will be uploading a second dockerfile later that has mongodb on the outside and a docker-compose.yml and a rancher-compose.yml that go along with it.

aaronjudd:
Merged. Tested local with kitematic, worked for me. nice cleanup.

aaronjudd:
well here's the bad news. this dockerfile isn't ready for our use because
- needs to expose env variables for scripting configuration of new reaction instances
- doesn't expose local mongo - (can't connect using localhost)
- doesn't build missing less themes (this goes away in meteor 1.2)

Issue #402, opened 2015-07-06
docker + orchestration + meteor reaction deployment


aaronjudd:
Production quality reaction deployment as a PaaS using Docker, Rancher or Tutum.

Meteor methods for creating and deploying reaction clusters using these orchestration services.

Testing giving each shop get it's own persistent 3 member replica set mongodb cluster service, and load balanced app servers running on docker containers.

Specs in progress with @jshimko, will add docs as we get the first prototype solution working.

faddat:
No kidding!  Surprised and glad to see this :).

Well, I suppose you felt quite the same about it that I did.

:)

-Jake

Jacob Gadikian
E-mail: faddat@gmail.com
SKYPE: faddat
Phone/SMS: +84 167 789 6421

faddat:
Tutum is also good-- I prefer that rancher is independent and of course that it is beer-free. , but tutum deserves a holler for coolness.  

Wiki will be up shortly, incidentally.  Would it be possible for me to docker up the rancher equivalent of Rancher's Build-Master dockerfile?  Any tips?

jshimko:
@faddat, not totally clear what you're asking.  Are you asking how to deploy Rancher for the use case mentioned here?

jshimko:
Our current progress on Launchdock has been officially open sourced, so we're going to close this issue and move the conversation over to the that repo instead.  (Setup docs coming very soon)  

[Check it out!](https://github.com/reactioncommerce/launchdock)

PR #503, opened 2015-10-28
Provide additional production docker images

tdecaluwe:
This PR refactors the docker setup so that images for the following purposes can be easily created:
* Fast building development image (the original image)
* Production image which includes a mongodb installation. The provided dockerfile yields images around 250mb in size.
* Production image which uses an external mongodb server (to use as a deployment image in #402 for example).

Both new production images are based on debian wheezy. The root `Dockerfile` is a symlink into the `docker` folder which contains all dockerfiles. Other images can be built as follows:

```
docker build -t reactioncommerce/reaction:mongo -f docker/reaction.mongo.docker .
docker build -t reactioncommerce/reaction:minimal -f docker/reaction.minimal.docker .
```


tdecaluwe:
This PR should be able to build after solving #504. Any feedback, thoughts or suggestions always welcome. I'm currently trying to build a tutum workflow on these images.

aaronjudd:
How are you accounting for the fact that you'll need to run in development mode to create LESS files for these other images? If you build your image in a normal "production" way - those files won't get created and you'll have a docker image that runs reaction - but without styling.

Also, we are using the current images in a tutum workflow, as part of Launchdock2 - a project we'll be sharing more about soon. @jshimko is working through the final details of this, and has also been looking at refactoring the Dockerfile to trim things down a bit.

tdecaluwe:
@aaronjudd I'm currently not. But if you or @jshimko are interested, I'll look into it, and fix up the dockerfiles.

tdecaluwe:
@aaronjudd These dockerfiles are something I'd cooked up to get a staging environment running, I'd be happy to add them, but if you've got better dockerfiles laying around, please use those!

tdecaluwe:
I read too fast, the proposed docker files do take into account the theme building. In fact, I'm using all the infrastructure in `bin/docker` already present (`build-meteor.sh`, `build-themes.sh` and `build-packages.sh`) as to maintain images which should be compatible with the image already present in the repo.

What I wanted to accomplish was this:
* A development image (I kept the one already present for this), which builds fast. This image should have a separate `aufs` layer for the meteor installation so we don't have to download this each time in development env.
* Separate production images which are very small. They build considerably slower, because they don't keep any unnecessary data, like the meteor installation, they just keep the bundle. However deploy time is considerably faster.
* Stateless app code shoudn't be in a volume like in the current image, instead one just redeploys a new image linking to the same production db.

I created a separate image with mongodb ideally suited for testing envs and another without mongodb meant for staging and production envs.

Issue #508, opened 2015-10-30
Allow bringing your own fixture data


tdecaluwe:
I've been looking into overriding the fixture data by `reaction-core`, however for any package that needs to register as a reaction package this is impossible, since it explicitly needs to depend on `reaction-core`. This is also true for loading fixtures at the app level. To override the standard fixture data we'd need a non-reaction meteor package to load our own fixture data before `reaction-core` does.

Because we'd depend explicitly on meteor load order (and this just doesn't seem right), I'm proposing an alternative: splitting out the fixture data in a separate package: `reaction-sample-data`. Simply disabling this package would allow us to prevent reaction from loading the fixtures.

tdecaluwe:
In relation to this: it seems to me the `setInterval` call in `ReactionRegistry.loadFixtures` is not really needed, since the `loadData` method seems to be executed in a synchronous way.

aaronjudd:
`setInverval` is there, because while everything works fine locally, when remotely deployed the timing of when things finish isn't as predictable, and as a result "Shops" doesn't load before other items always, and then creates a shop without the data all being tied to the  default shop. I know it's ugly.  See: https://github.com/reactioncommerce/reaction/issues/457

aaronjudd:
on your first comment: see https://github.com/reactioncommerce/reaction-shipping/blob/development/server/fixtures.js  (not impossible for packages to load their own fixtures).

tdecaluwe:
Closing this as PR #511 implements the proposed solution.

PR #511, opened 2015-11-03
Add reaction-sample-data package

tdecaluwe:
This PR adds a separate package which contains the sample data for a running reaction shop, as described in #508. 

To be functional, additional PR's in `reaction-core` and `reaction-shipping` need to be accepted:
* reactioncommerce/reaction-core#209
* reactioncommerce/reaction-shipping#4

The `reaction-sample-data` should also be transferred to `reactioncommerce`.

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -64,6 +64,10 @@ reactioncommerce:reaction-stripe
 #reactioncommerce:reaction-auth-net
 #reactioncommerce:reaction-braintree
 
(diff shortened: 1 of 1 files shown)


aaronjudd:
WIP.

 After the `reaction-sample-data` had been moved, I published `reactioncommerce:reaction-sample-data` to Atmosphere for testing purposes as well.

PR #533, opened 2015-11-25
Add forever and phantomjs to production dockerfiles

tdecaluwe:
Forever support adds about 10mb, phantomjs (which is needed for spiderable) around 40mb. ~~The docker build currently fails because of a bug in npm 3.4.1 but will complete again with npm 3.5.0.~~


aaronjudd:
FYI - Forever is what we used to use, but it stopped working the the 10.x node a while back - did you test with Debian?

tdecaluwe:
Not sure, but perhaps we don't even need this? This type of functionality is usually built in with orchestration tools.

PR #547, opened 2015-12-05
inventory/register refactoring

newsiberian:
Hello,

I was trying to find a way to make this method faster and reading of [this](http://info.meteor.com/blog/inserting-50000-documents-into-a-collection-slow-fast-and-fastest) great article gives me a solution. I did some tests with previous version of method with adding `childVariant` with 1000 items. For me average time of this operations was around 24 seconds. With [`batchInsert`](https://atmospherejs.com/mikowals/batch-insert) the same operation was taken around 0.1 seconds which is very cool)

### DIFF
--- a/packages/reaction-inventory/server/methods.js
+++ b/packages/reaction-inventory/server/methods.js
@@ -37,21 +37,24 @@ Meteor.methods({
           `inserting ${newQty - inventoryVariantCount} new inventory items for ${variant._id}`
         );
 
+        let items = [];
         while (i <= newQty) {
-          let inventoryItem = ReactionCore.Collections.Inventory.insert({
+          items.push({
             shopId: product.shopId,
             variantId: variant._id,
             productId: product._id
           });
-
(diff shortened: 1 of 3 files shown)


aaronjudd:
Good change. I thought about using `collection.rawCollection().initializeUnorderedBulkOp();`  - another way we could do bulk updates. I think that's what batch-insert is doing right? cool.

newsiberian:
[answer](https://github.com/mikowals/batch-insert/issues/2) from the developer if you are interested)

aaronjudd:
interesting.  @tdecaluwe We might want to add a method to `ReactionImport` and use that in the future -> I see inventory records and something that is likely to be imported a lot.

tdecaluwe:
I added some details to the docs in PR #555 about why I didn't add support for batch inserts. However batch inserts can still be accomplished through the use of `ReactionImport.fixture()` (also see PR, section "Batch inserts and fixture data").

Issue #553, opened 2015-12-08
broken product routes: lookup by ID if no title.


aaronjudd:
Broken behaviour, products without title should lookup by id.

<img width="1427" alt="reaction___qsldexg77zvhp4abs" src="https://cloud.githubusercontent.com/assets/439959/11646369/3ca7d10e-9d12-11e5-9d19-a30d6cd1795d.png">

newsiberian:
Hello, hmm... why not make a title mandatory?

aaronjudd:
Title is required in the schema, but we bypass it when creating a new one. I think it would be too confusing to have a default title, but nothing else when you create a new product.

aaronjudd:
It should be noted - that we create the product BEFORE the user edits it for the first time.

newsiberian:
Ok, now I get it. I will take this.

aaronjudd:
@newsiberian I added you to the "project collaborators".  That should let you assign issues to yourself, to help us track.

PR #557, opened 2015-12-08
fix for #513

### LINKED ISSUE #513: autoValue for product handle creates lowercase version of product id by default
To reproduce:  
1. Create a new product using the navbar dropdown -> `+ Add Product`
2. Fill out product details
3. Make product visible without setting a new handle.
4. Try visiting product from index

From product creation page
image

Linked product
image

newsiberian:
Hello, fix could fix #513 and maybe #553. Anyway after this changes we will have the following behavior:
screencast


aaronjudd:
:+1:

PR #568, opened 2015-12-10
Remove meteor versions from repo

tdecaluwe:
We are already removing this on `reaction reset` so we mgiht as well remove it from the repo. Additionally, this allows us, once again, to use the ever-useful `git commit -a`.


jshimko:
Oh, I don't know if we want that.  If that versions file isn't committed in the repo, we can't guarantee a build will use a particular set of versions.  As a result, we wouldn't be able to promise people that a clone of the repo at any given point in time will definitely work.  They could end up with a bunch of different dependency versions that break things.  

That file exists specifically to prevent that and most frameworks have an equivalent for the same reason.  In my opinion, overriding that with a `reaction reset` (or `rm .meteor/packages`) is an easy alternative for users that want that option.

tdecaluwe:
Yeah, I had exactly the same thoughts as you and we should definitely include one in stable releases. However on the `development` it doesn't seem to serve a real purpose. Certainly not because the one currently included, requires me to do an `--allow-incompatible-update`...

jshimko:
Hmm.  I haven't had to do an `--allow-incompatible-update` any time recently.

tdecaluwe:
Ah, I see, this was because of a forgotten `clone-packages.sh`. I'm going to close this, I wasn't really sure about this in the first place.

PR #569, opened 2015-12-11
Correct typo in schema. Says Shop but I think it means User

zenweasel:


aaronjudd:
actually the label only shows up when errors are in the console, and because almost every collection has "shopId", I made the labels read  `label: "Accounts ShopId"`so that when you see the errors you know which collection is failing (because otherwise the errors don't tell you enough detail).  Looks like the label was on the wrong property (ie userId vs shopId).

zenweasel:
Oh, ok. I will correct the correction.

aaronjudd:
@zenweasel you can just commit update to this PR..

aaronjudd:
Maybe we should close this one?  I can update quicker than trying to fix this merge.

zenweasel:
Fine by me.

PR #578, opened 2015-12-13
Added permission check to `addressBook` methods

newsiberian:


newsiberian:
WIP

packages/reaction-accounts/server/methods/accounts.js
```diff
@@ -127,20 +127,33 @@ Meteor.methods({
...
+   * @description add new addresses to an account
+   * @param {Object} doc - address
+   * @param {String} accountId - `account._id` which is need to be updated
+   * @return {Object} with keys `numberAffected` and `insertedId` if doc was
+   * inserted
    */
   "accounts/addressBookAdd": function (doc, accountId) {
     check(doc, ReactionCore.Schemas.Address);
     check(accountId, String);
+    if (accountId !== Meteor.userId() && !ReactionCore.hasAdminAccess()) {
```
mikemurray: The publication should take care of security for user accounts, as normal users are only subscribed to their own account data. The super admin gets access to everything, while shop admins can only access members of their shop.

https://github.com/reactioncommerce/reaction/blob/development/packages/reaction-accounts/server/publications/accounts.js

Doing the permission check here might be redundant, and might affect what's happening with the publication. We should make sure the publication is indeed preventing unwanted access to accounts; Then make sure these methods return false or and error that is displayed to the user. (which these ones do now; generally speaking of all client -> server method calls)

newsiberian: I've tested current version with such command sent by guest: `Meteor.call("accounts/addressBookAdd", insertDoc, adminAccountId);` and it was inserted

mikemurray: You are correct, I was mistaken about the role of publications. These server methods need explicit validation, using ongoworks:security, and / or as you have it in your updates. 

Sorry for the misinformation.

newsiberian:
There was a situation when removing address from addressBook was not affect on cart billing/shipping addresses. new cart method was added, which is used internally by `addressBookRemove` method.

If there no need to redo something - then this will be the last commit to this PR.

packages/reaction-accounts/server/methods/accounts.js
aaronjudd: Shouldn't this be:

`if (accountId !== Meteor.userId() || ReactionCore.hasAdminAccess()) `?

packages/reaction-accounts/server/methods/accounts.js
```diff
@@ -170,15 +184,21 @@ Meteor.methods({
...
+   * accounts/addressBookUpdate
+   * @description update existing address in user's profile
+   * @param {Object} doc - address
+   * @param {String} accountId - `account._id` which is need to be updated
+   * @return {Number} The number of affected documents
    */
   "accounts/addressBookUpdate": function (doc, accountId) {
     check(doc, ReactionCore.Schemas.Address);
     check(accountId, String);
+    if (accountId !== Meteor.userId() && !ReactionCore.hasAdminAccess()) {
```
aaronjudd: again, if `(accountId !== Meteor.userId() || ReactionCore.hasAdminAccess())`?

ie, If accountId = userId or hasAdminAccess

packages/reaction-accounts/server/methods/accounts.js
aaronjudd: OK, I got it, double negatives, the code works as is.

PR #585, opened 2015-12-15
address isBillingDefault/isShippingDefault behavior fix

newsiberian:
Hi, here is some changes that fix `isBillingDefault/isShippingDefault` behavior. Now any manipulations with them should have an effect in Cart, in UI and in other addresses state.

P.S. In there was a change in template, where we send `Meteor.userId()` to Methods. If we will use this template in dashboard, we will need to update this part to send `customer id`, not `Meteor.userId()`


newsiberian:
WIP

aaronjudd:
I actually knew that was a bug - but I sort of liked it as a feature which is why I was ignoring it.  :-1: 

The original behaviour was that whatever address you had selected (when you had multiple) would remain as the selected address even if you choose a new shipping option.  I think that this is probably what should happen, with another missing piece - when you enter your FIRST address, both addresses should automatically be selected.

newsiberian:
One question. What if a user intentionally turn of `isShippingDefault/isBillingDefault` when adding first address. Should we even in that case automatically select them?

aaronjudd:
I think it's more intuitive for both addresses to be selected by default. It's not until you add a second address that it's apparent (or useful) to select an alternate address.

This probably means that the insert needs to check and update the first entered address as "isDefault".

I don't mean that the first address in the array should always be "isDefault" (as there should always be one), as long as it's enforced on initial insert.

aaronjudd:
@newsiberian I wasn't sure if you were finished with this, let me know when you are finished and  I'll start reviewing.

newsiberian:
Thanks. This is ready for merging)

aaronjudd:
@newsiberian everything seems to work fine :+1:  when I tested as a new user.. the default action works great.

However, when I run tests:

`VELOCITY_TEST_PACKAGES=1 meteor test-packages --port 3006 --driver-package velocity:html-reporter`

There are failing tests.  This is the same reason the CircleCI tests are failing:

```
I20151230-07:50:13.458(-8)? jasmine-server-integration: 95 tests passed (9s)
W20151230-07:50:13.459(-8)? (STDERR) jasmine-server-integration: 5 tests failed
W20151230-07:50:13.460(-8)? (STDERR) Account Meteor method  addressBookAdd should throw error if wrong arguments were passed
W20151230-07:50:13.460(-8)? (STDERR) Expected function to throw an exception.
W20151230-07:50:13.460(-8)? (STDERR) meteor://💻app/tests/jasmine/server/integration/accounts.js:96:12: Expected function to throw an exception.
W20151230-07:50:13.460(-8)? (STDERR)
W20151230-07:50:13.460(-8)? (STDERR) Account Meteor method  addressBookUpdate should throw error if wrong arguments were passed
W20151230-07:50:13.461(-8)? (STDERR) Expected function to throw an exception.
W20151230-07:50:13.461(-8)? (STDERR) meteor://💻app/tests/jasmine/server/integration/accounts.js:292:12: Expected function to throw an exception.
W20151230-07:50:13.461(-8)? (STDERR)
W20151230-07:50:13.461(-8)? (STDERR) Account Meteor method  addressBookRemove should throw error if wrong arguments were passed
W20151230-07:50:13.461(-8)? (STDERR) Expected function to throw an exception.
W20151230-07:50:13.461(-8)? (STDERR) meteor://💻app/tests/jasmine/server/integration/accounts.js:518:12: Expected function to throw an exception.
W20151230-07:50:13.461(-8)? (STDERR)
W20151230-07:50:13.461(-8)? (STDERR) Account Meteor method  addressBookRemove should call `cart/unsetAddresses` Method
W20151230-07:50:13.463(-8)? (STDERR) Error: Match error: Expected string, got undefined
W20151230-07:50:13.464(-8)? (STDERR) meteor://💻app/packages/check/match.js:33:1: Error: Match error: Expected string, got undefined
W20151230-07:50:13.464(-8)? (STDERR)   at check (meteor://💻app/packages/check/match.js:33:1)
W20151230-07:50:13.464(-8)? (STDERR)   at [object Object].Meteor.methods.cart/unsetAddresses (meteor://💻app/server/methods/cart.js:566:5)
W20151230-07:50:13.464(-8)? (STDERR)   at maybeAuditArgumentChecks (meteor://💻app/livedata_server.js:1698:12)
W20151230-07:50:13.464(-8)? (STDERR)   at meteor://💻app/livedata_server.js:1611:18
W20151230-07:50:13.464(-8)? (STDERR)   at [object Object]._.extend.withValue (meteor://💻app/packages/meteor/dynamics_nodejs.js:56:1)
W20151230-07:50:13.464(-8)? (STDERR)   at [object Object]._.extend.apply (meteor://💻app/livedata_server.js:1610:45)
W20151230-07:50:13.465(-8)? (STDERR)   at [object Object]._.extend.call (meteor://💻app/livedata_server.js:1553:17)
W20151230-07:50:13.465(-8)? (STDERR)   at meteor://💻app/tests/jasmine/server/integration/accounts.js:559:23
W20151230-07:50:13.465(-8)? (STDERR)
W20151230-07:50:13.465(-8)? (STDERR)
W20151230-07:50:13.465(-8)? (STDERR) cart methods cart/unsetAddresses should throw error if wrong arguments were passed
W20151230-07:50:13.465(-8)? (STDERR) Expected function to throw an exception.
W20151230-07:50:13.465(-8)? (STDERR) meteor://💻app/tests/jasmine/server/integration/cart.js:92:12: Expected function to throw an exception.
W20151230-07:50:13.466(-8)? (STDERR)
```

If you can correct those tests, I'll merge and we'll push out as a part of a new release.

newsiberian:
I'm sorry, but I can't fix this tests because 4 of 5 happens because of [bug in ss](https://github.com/aldeed/meteor-simple-schema/issues/522) as I suppose.

test: "should call `cart/unsetAddresses` Method". I have a problem in such kind of tests. For example, we want to check if the method `cart/unsetAddresses` will be called or not  from inside `accounts/addressBookRemove` method.
Currently I'm doing this:
```javascript
spyOn(Meteor, "call");
expect(Meteor.call("cart/unsetAddresses")).toHaveBeenCalled();
```
but this is not working. Maybe you know how to do such tests. If no, I suppose the only way is to comment out this test.

aaronjudd:
Did you try `spyOn(Meteor.call,"cart/unsetAddresses")`, that might work..

newsiberian:
Hi, yes. This is not valid syntax, I suppose. It throws `Error: cart/unsetAddresses() method does not exist`.

newsiberian:
seems like I'm on a right way to fix this.

zenweasel:
@newsiberian When I clone and checkout this branch I am still getting 4 failing tests. Are you not seeing that? Can't approve until all tests are passing.

newsiberian:
@zenweasel, Hello, if you look at this test, you'll see, that they fail because of potential bug in SimpleSchema, so that these tests only show the bug was fixed or not. I explained this before...

zenweasel:
@newsiberian Yes you did explain before, but all four failing tests are caused by the bug in SS?
If that's true, we still don't want to have tests that are testing a bug in an external package and will cause the CI to fail every time. I guess we could mark them pending with a comment to revisit when the issue in SS is resolved?

zenweasel:
Also was confused since the last commit message was "fixed broken test;"

aaronjudd:
Clearly a bug (and an important one) that
```
check( function() { }, Schema);
```
 returns undefined.  

The question I guess is how do we handle this?  We could comment out these failing tests, knowing that some other bug is causing this (and is it `check` or simple-schema?).  

I'd suggest that we could also just change the test to expect `.toThrow(undefined)`.  which will at least allow us to notice if that behaviour changes.  We could create another issue in Reaction to monitor this bug and work separately to get a passing test here.   

I'd prefer to merge this if we can - but we're not merging any more PR's without 100% passing tests.

newsiberian:
@zenweasel, there was one broken test - I fixed it. 
This is an SS bug, as other tests shows. I think aldeed do something with it in next SS version.

~~I will comment parts that tests external packages.~~ I see what I can do here) One in `cart` and three in accounts. All tests are `should throw error if wrong arguments were passed`.

btw, thanks that take care on this PR

PR #586, opened 2015-12-15
update `addressBookRemove` method

newsiberian:
Hi, I think we don't need the whole object with address for this method. Address `_id` will be enough.


newsiberian:
This could be closed, because I replicated this changes in sibling PR.

Issue #600, opened 2015-12-17
Client currency rates: Cannot read property 'rate' of undefined


aaronjudd:
Reported by @boboci9 and #258 and related to #366 

>Hi,
I found an issue in the main.js and in the line 23 I saw a comment // fix for https://github.com/reactioncommerce/reaction/issues/248 so I figured I should report it here. 
Everything is working fine and then suddenly after some form submits I start getting this error message in the console:
```
TypeError: Cannot read property 'rate' of undefined
    at main.js:30
```
>If it's any help I debugged the Locale variable before the if and when the error occurs the Locale object is filled like the Locale.locale is holding all info like { capital, continent, currency, etc} but the Locale.currency is an empty object.

aaronjudd:
@zenweasel this sounds like it's right in the middle of the stuff you are debugging on  #366

zenweasel:
Fix by 0bf9deedbee9601cc63ddf521e1dd1462b4962d5 by @newsiberian

PR #601, opened 2015-12-17
Fix for #600

newsiberian:
Hi, quick fix for #600

### DIFF
--- a/packages/reaction-core/client/main.js
+++ b/packages/reaction-core/client/main.js
@@ -27,7 +27,7 @@ _.extend(ReactionCore, {
           typeof Locale.currency === "object" &&
           typeof Locale.locale.currency === "string") {
           const localeCurrency = Locale.locale.currency.split(",")[0];
-          if (shop.currencies[localeCurrency] &&
+          if (typeof shop.currencies[localeCurrency] === "object" &&
             typeof shop.currencies[localeCurrency].rate === "number") {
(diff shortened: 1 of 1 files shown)


packages/reaction-core/client/main.js
```diff
@@ -27,8 +27,11 @@ _.extend(ReactionCore, {
           typeof Locale.currency === "object" &&
           typeof Locale.locale.currency === "string") {
           const localeCurrency = Locale.locale.currency.split(",")[0];
-          Locale.currency.rate = shop.currencies[localeCurrency].rate;
-          localeDep.changed();
+          if (shop.currencies[localeCurrency] &&
+            typeof shop.currencies[localeCurrency].rate === "number") {
```
aaronjudd: Won't this fail if `shop.currencies[localeCurrency]` is not defined?

newsiberian:
must now work fine

newsiberian:
Hi, as far as I know in such syntax `typeof shop.currencies[localeCurrency] === "object" &&  typeof shop.currencies[localeCurrency].rate === "number"` the second part will not begins to be processed if the first part fails, but if you like `if-if` syntax more, I could update it)

aaronjudd:
OK this is good, I'm not sure why I was seeing error with the original. @newsiberian you're right - it should have worked. I just took the code out and tested and it was fine.  
Approved!

Issue #538, opened 2015-11-30
How to add a currency? I a custom one.


assiomatica:
Hi ,
I have seen the are in the Shops.json file, but how to reload only currencies?

THank YOu

aaronjudd:
In the latest development branch: to persist, you can make your own `reaction-sample-data` and modify Shops.json or import your own using `ReactionImport`.

assiomatica:
Perfect! Thank you.

Antonio Casula

mobile ch   +41 (0)79 9692274

   - ch.linkedin.com/in/antoniocasula/
   <https://mailtrack.io/trace/link/81d0859823098156b6c647708ae374b72da98f88?url=http%3A%2F%2Fch.linkedin.com%2Fin%2Fantoniocasula%2F&signature=a9183062006d1a90>

PR #602, opened 2015-12-17
Add Philipino Peso to currencies

zenweasel:
Another life-changing PR from me

### DIFF
--- a/packages/reaction-sample-data/private/data/Shops.json
+++ b/packages/reaction-sample-data/private/data/Shops.json
@@ -320,6 +320,10 @@
       "format": "%s%v",
       "symbol": "$"
     },
+    "PHP": {
+      "format": "%s%v",
+      "symbol": "PHP "
+    },
     "PLN": {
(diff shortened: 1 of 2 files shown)


aaronjudd:
@zenweasel since you just did this and it's fresh, would you mind adding an example to the docs for #538

zenweasel:
D'oh. Yes. I should have done that.

packages/reaction-sample-data/private/data/Shops.json
```diff
@@ -320,6 +320,10 @@
       "format": "%s%v",
       "symbol": "$"
     },
+    "PHP": {
+      "format": "%s%v",
+      "symbol": "PHP "
```
mikemurray: Add necessary spacing in `format`.

zenweasel: I tried that and it seemed to throw off alignment. That's why there is a space after symbol. But maybe I am missing something.

PR #610, opened 2015-12-21
fix to add-to-cart handler

newsiberian:
Hello, @mikemurray, I remember what you said about not making changes in `productDetails`, but I'm sorry, I could not stand :smile_cat:

By the way, could you explain what is this snippet for?:
```javascript
if (result && count === 0) {
  address = Session.get("address");
  if (!address) {
    return locateUser();
  }
}
```

### DIFF
--- a/packages/reaction-core/client/templates/products/productDetail/productDetail.js
+++ b/packages/reaction-core/client/templates/products/productDetail/productDetail.js
@@ -111,7 +111,7 @@ Template.productDetail.events({
     let currentProduct = selectedProduct();
 
     if (currentVariant) {
-      if (currentVariant.parentId === null) {
+      if (typeof currentVariant.parentId !== "string") {
         options = (function () {
           let _results = [];
           for (let variant of currentProduct.variants) {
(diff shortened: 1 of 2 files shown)


aaronjudd:
@newsiberian I think the changes to productDetails (the react stuff) will be for the 0.12.0 release.  As long it's not too crazy a few updates here and there for 0.11.0 can't hurt ;-)

newsiberian:
@aaronjudd, because of things need to be smooth, I think it was a bad idea to put animation login inside callback. I think I should rollback part of this commit. What do you think?

aaronjudd:
I haven't tested it yet, but looking at the code, yes, I think the animation should be outside of the callback . Also, if there were an error in the callback, then we could then update the animated text with an error message.

packages/reaction-core/client/templates/products/productDetail/productDetail.js
```diff
@@ -161,37 +161,41 @@ Template.productDetail.events({
...
-            if (!error && count === 0) {
-              address = Session.get("address");
-              if (!address) {
-                return locateUser();
+          Meteor.call("cart/addToCart", cartId, productId, currentVariant,
+            quantity, function (error) {
+              let address;
+              if (!error && count === 0) {
+                // todo understand that's going on here
+                //address = Session.get("address");
```
aaronjudd: I think that originally the behaviour was that if you selected an address during an order, that did not save it as your default address in your addressBook - or in the cart.  It was just the selected address for that checkout session.  I think this is probably not needed with the correct behaviour being that the cart specific addressbook selection is now stored in the cart, and the user's default addressbook is in the accounts.

aaronjudd:
re: snippet.  The `locateUser` function in `globals.js`,  is a crude method of inserting default address information into the addAddressBook form.  For example, for me, I get "Santa Monica" "California" defaulted (when it works).  

I'd like to see this swapped out for something cleaner (google address) in the future.
```
/**
 * locateUser
 * @return {Object} set and return session address based on browser latitude, longitude
 */
```

aaronjudd:
@newsiberian I'm closing this PR because the locateUser functionality should remain.  If you want to submit just the productDetail cleanup I'll accept that. If we want to improve the locateUser functionality let's create an issue to discuss some approaches.

newsiberian:
np, but I still suggest you to make this changes manually
here: `packages/reaction-core/client/templates/products/productDetail/productDetail.js`
and here: `packages/reaction-product-simple/client/templates/productDetail/productDetail.js`
```
-      if (currentVariant.parentId === null) {
+      if (typeof currentVariant.parentId !== "string") {
```
this changes is the main reason of this PR. Fix will not allow you to add to cart `product`without selected `option`.

aaronjudd:
@newsiberian resolved in patch.

PR #617, opened 2015-12-22
Add note about MongoDB requirements

tdecaluwe:
Reaction needs the MongoDB 2.6 bulk operations API. This PR adds a note in the docs about this requirement and will throw an `Error` on startup if the API couldn't be found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ cd reaction && git checkout master # default branch is development
 
 Additional installation options are in the [developer documentation](https://github.com/reactioncommerce/reaction/blob/development/docs/developer/installation.md).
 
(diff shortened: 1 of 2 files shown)


aaronjudd:
Approved! This looks great.

tdecaluwe:
Still one remark: This PR makes MongoDB 2.6 a hard dependency. If I move the check inside the `ReactionImport` code it will only throw when trying to use `ReactionImport`. Reaction would still be usable when not using `ReactionImport`.

aaronjudd:
@tdecaluwe I guess it might be bad form to make it hard dependency - as Meteor itself isn't enforcing any dependency there. However, there's likely to be other issues running 2.4  so I'm torn ( on which direction will lend itself to less support confusion). I suppose we should move the check inside Import.

PR #620, opened 2015-12-23
Proposing Bulgarian Translation

rkhunter:
I have only translated the strings.
Please add it to languages and all required configs.
P.S. How do I set it as a default language for new clients?

### DIFF
--- a/packages/reaction-core/private/data/i18n/bg.json
+++ b/packages/reaction-core/private/data/i18n/bg.json
@@ -1,0 +1,359 @@
+[{
+	"language": "Български",
+	"i18n": "bg",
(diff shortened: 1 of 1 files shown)


aaronjudd:
:+1:  I'll add in to the package assets,etc.

aaronjudd:
@rkhunter this is all merged into 0.10.2.  To set default, I think that should be in the shop settings via dashboard.

rkhunter:
@aaronjudd Thanks :+1:  I have just seen the currency setting is working now. I have fixed the wrong formatting of BGN and a small mistake in my translation.

Issue #624, opened 2015-12-29
Move non "mixins" from mixins.less in core-theme


mikemurray:
There are styles inside of `reactioncommerce:core-theme/default/mixins.less` that are causing that file to be dependent on variables from other less files.

We should move those styles to better locations so `mixins.less` is not dependent on anything, and can be more easily imported into your own theme.

PR #626, opened 2015-12-29
Moving around styles and mixins

mikemurray:
Split up mixins into individual files, and import them all into
mixins.less. Fixed issues with my-custom-theme-template, and added
example of importing the rtl.less mixins.


aaronjudd:
@mikemurray related to https://github.com/reactioncommerce/reaction/issues/624?

PR #613, opened 2015-12-21
RemoveFromCart refactor

newsiberian:
Hello,
I did not find a reason why `cart/removeFromCart` method is shared, so I moved it to server folder. Also method now receive only `cartItemId` instead of two args - `cartId` and `item` object!! I think this is more secure way.

Also, I don't know do you have a plans to use this method by admin to manipulate user's cart? If yes, it is needed to add few lines of code to implement this functionality. Because of this I leave tests without changes for now...

This is fixed #611.

PR #631, opened 2015-12-30
Fix test for #613

newsiberian:
- fixed broken test;

### DIFF
--- a/packages/reaction-core/tests/jasmine/server/integration/cart.js
+++ b/packages/reaction-core/tests/jasmine/server/integration/cart.js
@@ -19,93 +19,91 @@ describe("cart methods", function () {
     });
   });
 
-  //describe("cart items", function () {
-    describe("cart/addToCart", function () {
-      beforeEach(function () {
-        ReactionCore.Collections.Cart.remove({});
-      });
-
-      it("should add item to cart", function (done) {
-        spyOn(ReactionCore, "shopIdAutoValue").and.returnValue(shop._id);
-        spyOn(ReactionCore, "getShopId").and.returnValue(shop._id);
-        const product = Factory.create("product");
-        const cartId = Meteor.call("cart/createCart", userId, shop._id);
-        const productId = product._id;
-        const variantData = product.variants[0];
-        const quantity = 1;
-
-        Meteor.call("cart/addToCart", cartId, productId,
-          variantData, quantity);
-
-        let carts = ReactionCore.Collections.Cart.find({
-          _id: cartId
-        }, {
-          items: product
-        }).fetch();
-
-        expect(_.size(carts)).toEqual(1);
-        expect(_.size(carts[0].items)).toEqual(1);
-        expect(carts[0].items[0].productId).toEqual(productId);
-        return done();
-      });
-
-      it("should merge all items of same variant in cart", function (
-        done) {
-        spyOn(ReactionCore, "shopIdAutoValue").and.returnValue(shop._id);
-        spyOn(ReactionCore, "getShopId").and.returnValue(shop._id);
-        const product = Factory.create("product");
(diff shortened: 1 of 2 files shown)


packages/reaction-core/tests/jasmine/server/integration/cart.js
```diff
@@ -79,12 +79,13 @@ describe("cart methods", function () {
...
         const currentCart = Factory.create("cart");
         const cartId = currentCart._id;
 
         expect(currentCart.items[0]).toBeDefined();
         for (let cartItem of currentCart.items) {
-          Meteor.call("cart/removeFromCart", cartId, cartItem);
+          Meteor.call("cart/removeFromCart", cartItem._id);
         }
 
         let modifiedCart = ReactionCore.Collections.Cart.find({
```
aaronjudd: I'm not actually sure this test is passing. Yes, I see it work on the client, but if you uncomment and console.log `modifiedCart ` here, you'll see that the items still exist.  This could be timing/spy test related, but it could also be the method is actually failing (since this test has no other confirmation that Meteor call worked (other than it "cart/removeCart" throwing an error).

aaronjudd:
In `cart/removeFromCart`, [this](https://github.com/newsiberian/reaction/blob/removeFromCart_refactor/packages/reaction-core/server/methods/cart.js#L268) should probably be 

```
if (cart && cart.items) {
```

and [here](https://github.com/newsiberian/reaction/blob/removeFromCart_refactor/packages/reaction-core/server/methods/cart.js#L305) `cartId` is not defined.

Is Meteor.userId() always available to this method now? (does [the cart query](https://github.com/newsiberian/reaction/blob/removeFromCart_refactor/packages/reaction-core/server/methods/cart.js#L263) always work now?

newsiberian:
Fixed all you're mentioned.
> Is Meteor.userId() always available to this method now? (does the cart query always work now?

I believe, it should be always available.

aaronjudd:
Failed, but making progress ;-)
```
I20151230-20:00:44.925(0)? jasmine-server-integration: 78 tests passed (10s)
W20151230-20:00:44.926(0)? (STDERR) jasmine-server-integration: 1 tests failed
W20151230-20:00:44.927(0)? (STDERR) cart methods cart items cart/removeFromCart should remove item from cart
W20151230-20:00:44.928(0)? (STDERR) TypeError: Cannot read property '0' of undefined
W20151230-20:00:44.928(0)? (STDERR) meteor://💻app/tests/jasmine/server/integration/cart.js:94:38: TypeError: Cannot read property '0' of undefined
```

newsiberian:
I'm still working)
btw, there is a bug inside factories cart. It's very painful... Cart creates with previous user. not with current.

newsiberian:
Hello, I've turned off this test. I don't know how to make it work. Problem with it - it shows all correct within console, but in fact it doesn't do `update $pull` operation on document. At all... Maybe it is some kind of bug in `velocity`, maybe it's because of hooks over this call.

In live method works fine.

zenweasel:
So this PR does not appear to fix the test, just stops it from running? I know from the history that this test has been problematic. Are we ok with just not doing it? That seems worrisome.

aaronjudd:
@newsiberian here's some linting errors as well: https://codeclimate.com/github/reactioncommerce/reaction/pull/631 - I can merge this as soon as that's fixed.

zenweasel:
Disregard my comments about disabling the test. My mistake.

newsiberian:
@aaronjudd, should I refactor this too?:
```javascript
if (error) {
  ReactionCore.Log.warn("error adding to cart", ReactionCore.Collections
    .Cart.simpleSchema().namedContext().invalidKeys());
  return error;
}
```

Issue #621, opened 2015-12-26
[discussion] addToCart refactor


newsiberian:
Hello, I've noticed terrible lags, by doing app client tests. The reason of this you could see on the screenshot. Look at the _quantity_.
![error2](https://cloud.githubusercontent.com/assets/11130487/12005713/f3bef7ba-abd5-11e5-842e-deeb1c666f6e.png)

This led me to idea to rewrite this method.
I added an additional function `quantityProcessing`:

``` javascript
function quantityProcessing(productId, variantId, itemQty = 1) {
  let quantity = itemQty;

... calculations

  return quantity;
}
```

 which supposed to decide how many item quantity we will add to the user cart according to the following conditions:
- check the minimum threshold of the order - #225. In due to what, we will need to add an additional field to product/productVariant schemas. Or maybe we already have something for that? What do you think?

If item from stock:
- check the `inventaryQuantity` against `itemQty` and if `itemQty` > `inventaryQuantity` - return all we have in stock. Right?
- anything else here?

if item not from stock (made on request/is delivered on request - custom tailorin, custom jewelry, etc): 
- only check availability of product. Here we have an individual approach to each order.

if item is a digital product or something similar:
- only check availability of product

To make all this changes possible, I think, we need to think about expansion of product `type` property, for example, or maybe about adding some new prop. I like this approach - #152. We could add `itemType` as the stub for now. @aaronjudd, any thoughts?

aaronjudd:
@newsiberian thinking about this and will respond soon.

for reference: Related to #609

newsiberian:
Hello, @aaronjudd, if you are interested in current progress, it could be found [here](https://github.com/newsiberian/reaction/blob/cleanup_mergeCart/packages/reaction-core/server/methods/cart.js#L1-L36) and [here](https://github.com/newsiberian/reaction/blob/cleanup_mergeCart/packages/reaction-core/server/methods/cart.js#L283)

Issue #609, opened 2015-12-20
Refactor a `cart/mergeCart`


newsiberian:
Hello, I did some tests to this method and they show me that we can't rely on `sessionId` when merging carts.

Simple use case: You logged in on one device as normal user with items in cart [you have global `sessionId=1` here]. Further you, or your son put some products in cart from other device and try to sign in into this account. On second device you, probably, will have `sessionId=2`, but server will search cart  by `sessionId=1`. Carts, probably, will not be merged, because of that.

Second use case: I clean browser cache – `sessionId` changed, but server continue to search cart by `sessionId=1`.

`SessionId` is very unstable, and I think we should avoid of using it for such kind of tasks.

Currently I thinking about passing `userId` of a second device to this method by somehow and search second cart only by `userId` criteria.

Do you have any thoughts about this?

aaronjudd:
`sessionId` should be unique for each new client (moz vs firefox, mobile vs desktop etc), but not unique in the same browser instance (tabs). 

See: https://github.com/reactioncommerce/reaction/issues/393

Let me describe the cart handling, as it was meant to work:

When an **anonymous** user visits, they get a `sessionId`, this a generated server side, and stored in the browser localStorage, so that combined with their **anonymous** user account identifies them within the tab/browser/window session. The presence of a unique sessionId tells us not to create another user account for an **anonymous** user.  The same sessionId should never be used by the server for any two sessions, as it comes from the client.. it's only set on the server side in the initial *subscription to `Sessions`*.

It this same user visits on another device, browser or anonymous session they will get a new **anonymous** `userId` and `sessionId`.

When a visitor registers with a password or authentication service (FB,GH,etc),  they are **anonymous** before they register.  We then create a _new_ user and account.  If the user is `authenticated`, and has a `sessionId` that matches previous carts with the same sessionId, we then merge matching `sessionId` carts into the newly created cart.  (and we *<u>should* </u>remove the existing **anonymous** cart, user, account).

We identify an `authenticated` user  in `Roles` as a `guest` without the `anonymous` role.

When this `authenticated` user logs out of the site, the `publication` is updated and they no longer have permissions to view their `authenticated` cart.  The publication sees this, and creates a new cart for them, with a new `anonymous` cartId, but still the same `sessionId`.

[...]

Issue #592, opened 2015-12-16
CartWorkflow improvement


newsiberian:
Hello. There is a few moments, that we could improve in `cartWorkflow`.

We can have a situation where a user deletes products from `cart`, being in the last `cartWorkflow` stage and if he later put in a `cart` other products, then entering the `checkout`, it will be the last `cartWorkflow` stage. I think this is not expected. I believe, much more logical way will be to throw him back to the third `coreCheckoutShipping` step.

It may also be a situations where a user who is at any stage beyond the second `checkoutAddressBook` address may cancel the `defaultShippingAddress` or he could remove all addresses at all. Currently, when this happens, nothing has changed, as far as I understand.
Maybe if we have changes at second stage we need to react on it somehow?

#### Tasks

- [x] When you select a shipping address, we should reset the following steps and recalculate shipping rates.
- [x] With the first address entered, both shipping and address should be selected by default (and set isDefault).

UPD: ready, but need tests
- [x] if quantity changes, we reset to coreCheckoutShipping, but use the isDefault shipping address.

aaronjudd:
Your interpretation of how it should work is correct. There is an outstanding issue https://github.com/reactioncommerce/reaction/issues/250 that covers this. The shipping needs to be recalculated, and the previously completed workflow should be reset `coreCheckoutShipping`.

newsiberian:
I want to say even more, we should reset shipping when `isShippingDefault` address was changed.

I was thinking about bug/feature in #585 and now I think this is right behave. If user already filled `isShippingDefault` address in Profile or during the previous order, it is fine that he may start with `coreCheckoutShipping`. Right?

aaronjudd:
See my comments about the feature/bug - it's really two things:
1 ) When you select a shipping address, we should reset the following steps and recalculate shipping rates.
2)  With the first address entered, both shipping and address should be selected by default.
3) if quantity changes, we reset to `coreCheckoutShipping`, but use the `isDefault` shipping address.

newsiberian:
@aaronjudd, hello, I'm looking into shipping now and do not quite understand what do you mean by saying "recalculate shipping rates" in item 1. As I can see, we have a flat rates only. What need to be recalculated?

aaronjudd:
We are building additional shipping packages, such as Shippo, and support for FedEx,etc..

aaronjudd:
I've updated #250 description with more details on shipping rate recalculation timing.

aaronjudd:
[...]

Issue #536, opened 2015-11-26
Cart subscription not flush on user login/logout


newsiberian:
Hello, seems there is a bug with it.
How to reproduce:
1 way:
* before log in: leave a breakpoint somewhere on line 10 in cartDrawer.js;
* on breakpoint, type: `ReactionCore.Subscriptions.Cart.subscriptionId` in console;
* log in as admin and on breakpoint type again step 2;
* then log out and repeat step 2;
In my case every time `subscriptionId` was same;

2 way:
* before log in: leave a breakpoint somewhere on line 10 in cartDrawer.js;
* log in as admin and on breakpoint type: `Cart.findOne()`
* logout and on breakpoint type `Cart.findOne()` again.
In my case I see admin cart after logout.

aaronjudd:
I believe this is resolved with #650, #636, #536.  We should re-open if necessary. I'm not able to duplicate now though.

PR #636, opened 2016-01-01
cart methods update. part 2

newsiberian:
Hello, here is another portion of changes in `cart` methods and related places.
**status:** WIP
This PR heavily on depend on #585 and contain the following list of changes (see commits for more info):
* added destructuring to cart methods;
* reworked cart pub/sub;
* **breaking change** `cart/addToCart` now receives another set of arguments;
* fixed several typos;
* allow `shipping/updateShipmentQuotes` call within `cart/addToCart`;
* added new function `quantityProcessing` for calculating `product` quantity that was added to `cart` (WIP);
* fixed almost all broken tests for app-client-integration-testing;
* added few new tests for cart methods and cart pub/sub;

*UPD: 04.jan*
We are moving from using `ReactioCore.sessionId` in methods to real `sessionId`s which we will receive directly from client side:
* **breaking changes** in `cart/createCart`: method now receives additional `sessionId` arg;
* **breaking changes** in `cart/mergeCart`: method now receives additional `currentSessionId` arg;
* **breaking changes** in `cart/copyCartToOrder`: method now receives additional `currentSessionId` arg;
* **breaking changes** in server `cart/submitPayment`: method now receives additional `sessionId` arg;
* `ReactionCore.Schemas.Cart.sessionId` modified;

*UPD: 8.jan*
* lots of changes. See commit description.

related issues: #393, #250, #621, #225, #609, #536, #592. Maybe I missed something :smile:

This PR depend on changes from reactioncommerce/reaction-paypal#23 and from reactioncommerce/reaction-factories#3


newsiberian:
**status:** ready

zenweasel:
I've checked out the Paypal and Factory PR's and this one and the tests are not running for me. I am getting an error. I've seen this error before when faker or Factory were not available.

```.js:245
W20160111-15:23:58.254(8)? (STDERR) 						throw(ex);
W20160111-15:23:58.254(8)? (STDERR) 						      ^
W20160111-15:23:58.255(8)? (STDERR) Error: Session id is required
W20160111-15:23:58.255(8)? (STDERR)     at getErrorObject (/private/var/folders/jl/zpw4rhsn4fj9gj3vx_vd47nh0000gn/T/meteor-test-runknrc5q/.meteor/local/build/programs/server/packages/aldeed_collection2.js:450:15)
W20160111-15:23:58.255(8)? (STDERR)     at [object Object].doValidate (/private/var/folders/jl/zpw4rhsn4fj9gj3vx_vd47nh0000gn/T/meteor-test-runknrc5q/.meteor/local/build/programs/server/packages/aldeed_collection2.js:433:13)
W20160111-15:23:58.255(8)? (STDERR)     at [object Object].Mongo.Collection.(anonymous function) [as insert] (/private/var/folders/jl/zpw4rhsn4fj9gj3vx_vd47nh0000gn/T/meteor-test-runknrc5q/.meteor/local/build/programs/server/packages/aldeed_collection2.js:202:25)
W20160111-15:23:58.255(8)? (STDERR)     at Function.Factory._create (/private/var/folders/jl/zpw4rhsn4fj9gj3vx_vd47nh0000gn/T/meteor-test-runknrc5q/.meteor/local/build/programs/server/packages/dburles_factory.js:140:29)
W20160111-15:23:58.256(8)? (STDERR)     at Function.Factory.create (/private/var/folders/jl/zpw4rhsn4fj9gj3vx_vd47nh0000gn/T/meteor-test-runknrc5q/.meteor/local/build/programs/server/packages/dburles_factory.js:149:24)
W20160111-15:23:58.256(8)? (STDERR)     at```

File: packages/reaction-accounts/tests/jasmine/server/integration/publications.js [hunk 1]
```diff
@@ -1,51 +1,80 @@
...
     // reset
-    ReactionCore.Collections.Products.remove({});
-    ReactionCore.Collections.Shops.remove({});
-    ReactionCore.Collections.Orders.remove({});
-    shop = Factory.create("shop");
+    Meteor.users.remove({});
+    // ReactionCore.Collections.Products.remove({});
+    // ReactionCore.Collections.Shops.remove({});
+    // ReactionCore.Collections.Orders.remove({});
+    // shop = Factory.create("shop");
```
zenweasel: If we aren't going to do this operation, can we just remove these rather than commenting them out?

zenweasel:
Approved. I can't run the tests on this PR because I get some weird error that I think has to do with Factory/faker even with @newsiberian's changes to reaction-factory. But I think this may just be a local environment thing (though I have cloned a fresh repo, etc. and got the same results)

But I tested all three payment methods and some of the bugs that this should be fixing and it seems to be working correctly.

aaronjudd:
I need to dig into this more, I don't see the need for Session, as Session does not persist between tabs/browser,etc.- it's only a global ReactiveDict.-- so why use Session.get when sessionId is in the localStorage?

packages/reaction-core/server/methods/cart.js
```diff
@@ -706,10 +848,13 @@ Meteor.methods({
    * and adds "paymentSubmitted" to cart workflow
    * Note: this method also has a client stub, that forwards to cartCompleted
    * @param {Object} paymentMethod - paymentMethod object
+   * @param {String} sessionId - current client session id. We don't use it
+   * directly within this method, just throw down though hooks
    * @return {String} returns update result
    */
-  "cart/submitPayment": function (paymentMethod) {
+  "cart/submitPayment": function (paymentMethod, sessionId) {
```
aaronjudd: should always check methods, function arguments.

aaronjudd: a breaking change could be prevented here with a:

- adapt to using an object on these external method calls so that it's not a breaking change in the future
- use something like `sessionId = sessionIdArg || Session.get('sessionId');` as `Session` is global.
-

newsiberian:
@aaronjudd,
> I need to dig into this more, I don't see the need for Session, as Session does not persist between tabs/browser,etc.- it's only a global ReactiveDict.-- so why use Session.get when sessionId is in the localStorage?

Agree, but now it doesn't matter that to use here, because of [this](https://github.com/reactioncommerce/reaction/pull/636/files#diff-2a98d3843f1843367fe139d105786c55R258) and [this](https://github.com/reactioncommerce/reaction/pull/636/files#diff-c5bd6fa139d3e0cd292257e9739dfe34R24) - Session.get("SessionId") now always global for all tabs.

So, if it is needed it could be relpaced via `amplify`. Should I replace it for methods args?

aaronjudd:
When I three session (same browser), while logged in to a session in another browser (may not be relevant), when I log out of one browser, I'd expect all sessions in that browser to log out.
See: 
qe8x0pfmdw

aaronjudd:
Completing a cart in one browser, while logged in to two browsers - destroys other carts.


screenshot_1_11_16__9_42_pm

newsiberian:
@aaronjudd, 
> When I have three session (same browser), while logged in to a session in another browser (may not be relevant), when I log out of one browser, I'd expect all sessions in that browser to log out.

This is that I was tried to solve here partly. This is not fresh but, you could try it on `development`.

newsiberian:
> Completing a cart in one browser, while logged in to two browsers - destroys other carts.

Please, clean all browsers cache before you testing this

aaronjudd:
think we should update the the cart/submit payments - since Session is global, why even introduce this, since it's a breaking change, and not necessary..?

`Meteor.call("cart/submitPayment", paymentMethod")`

and
 ```  
"cart/submitPayment": function (paymentMethod) {
    check(paymentMethod, ReactionCore.Schemas.PaymentMethod);
    sessionId = Session.get("sessionId");
```

**Comment by aaronjudd** [comment]
@newsiberian If we can cleanup up those payment methods, the other issues we've noted are documented as separate issues, so I'd like to merge this, so we can start dealing with the smaller issues on this.

**Comment by newsiberian** [comment]
cool)

**Comment by newsiberian** [comment]
Could we discuss this via gitter?

File: packages/reaction-core/client/subscriptions.js [hunk 3]
```diff
@@ -3,36 +3,44 @@
...
+  const sessionId = amplify.store("ReactionCore.session");
+  let newSession;
+  Tracker.nonreactive(() => {
+    newSession = Random.id();
+  });
+  if (typeof sessionId !== "string") {
+    amplify.store("ReactionCore.session", newSession);
+    Session.set("sessionId", newSession);
+  }
+  if (typeof Session.get("sessionId") !== "string") {
```
aaronjudd: Perhaps Session.equals could be used here, whenever you use Session.get you are causing the computations to rerun, so it's important to avoid excess `Session.get`.

```
If value is a scalar, then these two expressions do the same thing:

(1) Session.get("key") === value
(2) Session.equals("key", value)
... but the second one is always better. It triggers fewer invalidations (template redraws), making your program more efficient.
```

packages/reaction-core/client/subscriptions.js
```diff
@@ -3,36 +3,44 @@
...
-let cart = ReactionCore.Collections.Cart.find();
-cart.observeChanges({
-  removed: function () {
-    Meteor.subscribe("Cart", Session.get("sessionId"), Meteor.userId());
-  }
+// @see http://guide.meteor.com/data-loading.html#changing-arguments
+Tracker.autorun(() => {
+  let sessionId;
+  // we really don't need to track the sessionId here
+  Tracker.nonreactive(() => {
```
aaronjudd: Why did this need to be wrapped in a nonreactive()?

packages/reaction-core/common/methods/cart.js
```diff
@@ -4,8 +4,9 @@
 //
 if (Meteor.isClient) {
   Meteor.methods({
-    "cart/submitPayment": function (paymentMethod) {
+    "cart/submitPayment": function (paymentMethod, sessionId) {
```
aaronjudd: Because this  defined in a `isClient` block , `sessionId` could just be a `Session.get('sessionId');`

packages/reaction-core/client/subscriptions.js
newsiberian: This is exactly that I want from Session. In this `autorun` we are tracking Session("sessionId") changes. We can't track `amplify.store()` but we need to push session changes to Session Collection somehow. That's why we are using Session.get

packages/reaction-core/client/subscriptions.js
newsiberian: In this part we do not need to track `SessionId`, because the one thing we need to track here is a `userId`. But we need to send `sessionId` down to server methods. If we not put Session.get("SessionId") in nonreactive, it will be cause of rerun cart subscription on every page refresh.

packages/reaction-core/common/methods/cart.js
newsiberian: agree, but we could leave it like this due to identically of all `cart/submitPayment`. Or should I update it?

aaronjudd: we can have the same method defined  on the server and client.  We can have the sessionId as param on server, but if we avoid requiring it on the client and use `Sessions` then the payment packages (which both are calling submit as a client methods) wouldn't need to pass sessionId.  I think it's a worthwhile change, even if a bit of a confusing implementation if we can avoid the breaking changes to the paymentSubmit method..

newsiberian: we need `sessionId` in `carts/cartCopyToOrder`. To get it from client the only way I found is to pass it via `cart/submitPayment` hook. We even not use `SessionId` directly inside `cart/submitPayment`. I'm not sure I understand meaning of your proposal, but if we remove SessionId from `cart/submitPayment` arguments we will not get SessionId in `cart/mergeCart`

newsiberian: additionally please, take a look at the latest comments #393

newsiberian: @aaronjudd, I think there is at least one way to not add `sessionId` to `cart/submitPayment` as arg.

We could untie `cartId` from `invoiceID`(PayPal or other). if we will do that, then we could avoid to remove cart on each order, but this could lead to more and more breaking changes. So this could be not the best scenario.

packages/reaction-core/server/methods/cart.js
```diff
@@ -344,13 +464,21 @@ Meteor.methods({
...
-    let user;
+    check(sessionId, String);
+
+    const cart = ReactionCore.Collections.Cart.findOne(cartId);
+    // security check
+    if (cart.userId !== this.userId) {
+      throw new Meteor.Error(403, "Access Denied");
+    }
+    const order = Object.assign({}, cart);
+
```
aaronjudd: we don't need to pass sessionId, just the cartId as sessionId = cart.sessionId

newsiberian: As @zenweasel says "Boom". Your right. We could just copy `sessionId` from previous cart to new one. It should be fine in case then anonymous make payment. I will try this tomorrow

aaronjudd:
@newsiberian see session updated here https://github.com/reactioncommerce/reaction/tree/no-session-cart-update-methods

PR #637, opened 2016-01-02
FIX: Use currentId from template data only if property is defined

prinzdezibel:


newsiberian:
Hi, some private data here. I you sure about this?

zenweasel:
@prinzdezibel Hey Micheal, this looks like your entire custom implementation rather than a fix to the issue described. Is that what you intended?

prinzdezibel:
Ouch!! Stupid me. Sorry!! I've messed up royally.

PR #640, opened 2016-01-03
Allow overriding less variables from reactioncommerce:core-theme more easily.

prinzdezibel:
The docs were not clear how one could easily override less variables for
custom colors. With this proposed change it is doable with much less
work for the unexperienced developer.


zenweasel:
Approved

Issue #532, opened 2015-11-25
Region dropdown doesn't pre-populate if region-detect does not work


zenweasel:
When checking out if it does not detect that I am in the Philippines it will fill the "Country" field with "United States" but the Region drop-down does not become the Select (it just seems to be a plain text field). If you select another country and then select US again it will pre-populate.

Steps to Reproduce:

1. Disable network connection so that location detection will not work. (for me it only works about half the time)
2. From fresh install place an item in the cart and proceed to checkout.
3. Observe that the Region dropdown is not populated.

zenweasel:
<img width="783" alt="no_state_dropdown" src="https://cloud.githubusercontent.com/assets/72819/12131740/2e55a022-b450-11e5-84b6-96769d355459.png">

PR #649, opened 2016-01-06
Correct Session.get and AutoForm.getValue when populating Region dropdown

zenweasel:
Closes https://github.com/reactioncommerce/reaction/issues/532 and other cases where GeoIP would detect anything other than the US that have regions (e.g. Canada)


aaronjudd:
Approved!

Issue #653, opened 2016-01-06
ReactionCore.hasPermission issue when seller shop id is considered


boboci9:
The function hasPermission both in client and server implementations result in incorrect false when the seller shop permissions are considered.

PR #654, opened 2016-01-06
Fix hasPermission issue #653

boboci9:
Hi, 

I debugged this issue and the user variable is not used correctly when seller shop permissions are considered.


aaronjudd:
Looks good.  Approved.

PR #668, opened 2016-01-08
Allow adding an entire package to the git index

tdecaluwe:
This allows me to write `git add packages/reaction-core` to add all changes in core for example.

### DIFF
--- a/packages/.gitignore
+++ b/packages/.gitignore
@@ -1,20 +1,20 @@
 *
 !.gitignore
 
-!my-custom-theme-template/*
+!my-custom-theme-template
 
-!reaction-accounts/*
-!reaction-analytics/*
-!reaction-analytics-libs/*
-!reaction-collections/*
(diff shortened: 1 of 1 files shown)


mikemurray:
@tdecaluwe this change has a pretty nasty side effect in Atom and Github Desktop. New files in those folders are ignored in Atom an cannot be searched with the fuzzyfinder. New files cannot be committed in Github Desktop as well.

This change will be reverted for the time being.

perhaps try `git add --all packages/reaction-core` instead. If that doesn't solve your issue, another PR will be welcome if it addresses those issues above.

tdecaluwe:
No problem, didn't know about Atom. However `reaction-core` has some nested `.gitignore` files (`.npm`, `bower`) so this doesn't work either. Did you try adding both `!packages/reaction-core` and `!packages/reaction-core/*`?

Issue #655, opened 2016-01-06
Issue with the Products publication


boboci9:
Hi,

I have debugged this issue and I wanted to check with you before I PR the solution. See the comments between the lines

```
// handle multiple shops
    if (shops) {
      selector = {
        shopId: {
          $in: shops // -> HERE YOU REFERENCE THE SECOND PARAM AS BEING A LIST OF IDs WHICH IS GOOD
        }
      };
      // check if this user is a shopAdmin
      for (let thisShop of shops) {
        if (Roles.userIsInRole(this.userId, ["admin", "createProduct"],
            thisShop._id)) { // HERE YOU REFERENCE THE SAME SECOND PARAM AS BEING A LIST OF PRODUCT OBJECTS
          shopAdmin = true;
        }
      }
    }
```
Solution is to change the thisShop._id to thisShop:
```
for (let thisShop of shops) {
        if (Roles.userIsInRole(this.userId, ["admin", "createProduct"],
            thisShop)) { // BECAUSE IT IS A STRING CONTAINING THE ID ALREADY
          shopAdmin = true;
        }
      }
```

I didn't add the PR right away because I remembered we had this discussion previously but I am not sure why you decided to change it back, but this way I'm not sure how you add the second param to work with both statements at the same time.

PR #680, opened 2016-01-12
fix the product publication issue #655

boboci9:
Both statements can't be correct at the same time:
```
 selector = {
        shopId: {
          $in: shops
        }
```
and 
```
 for (let thisShop of shops) {
        if (Roles.userIsInRole(this.userId, ["admin", "createProduct"],
            thisShop._id)) { 
          shopAdmin = true;
        }
      }
```
Read more in #655 issue.


aaronjudd:
:+1: Thanks @boboci9 merging.

PR #683, opened 2016-01-13
Tests fix for #636

newsiberian:


aaronjudd:
:+1: thanks for catching those.

PR #684, opened 2016-01-13
Revert "Allow adding an entire package to the git index"

mikemurray:
Reverts reactioncommerce/reaction#668

Causes new files to not be searchable in Atom, and cannot be committed in Github Desktop.


aaronjudd:
Approved. Confirmed both behavior and fix..

PR #686, opened 2016-01-13
Added key up event on input, reset form on successful submit

### LINKED ISSUE #604: When filling in refund amount, you cannot click on "Apply Refund" without clicking out of the input box
Trivial bug but while we are fiddling with the input box...

You can still issue a refund by clicking out of the box and then clicking the button but it's annoying and may not be obvious.

mikemurray:
resolves #604 

Added keyup event for the refund_amount field to properly trigger a
"change" and set the correct status on the "Apply Refund" button.

After refund is submitted, revert refund amount to 0 to disable the
"Apply Refund" button.


aaronjudd:
:+1:

PR #687, opened 2016-01-13
Switching sign in views keeps credentials filled in

### LINKED ISSUE #576: click register in login form should not lose inputs
inputs are being lost after switching between login/register in the sign in form. These should persist.

mikemurray:
resolves #576


aaronjudd:
:+1:

PR #689, opened 2016-01-13
Fixed yield region name typo

### LINKED ISSUE #588: Typo and No Use of Region in Refactored Code
Noticed  the Typo in Core "dashboard.html"
       {{> yield  "dashbaordHeaderControls"}} 

should be as           {{> yield "dashboardHeaderControls"}} .

It is not used any where in code.  Earlier it was used in Accounts and Shipping package which is now moved to Admin Controls Menu. Suggest to remove it.

mikemurray:
resolves #588


aaronjudd:
Approved. Merging.

PR #695, opened 2016-01-13
Potentially call setPaymentAddress also if cart.billing is present (but

prinzdezibel:
no billing address is specified yet).

Before this fix setPaymentAdress was not called if cart.billing
was there (even if no address was specified)

Analogous to cart.billing the same applies to cart.shipping.

NOTICE: This FIX is more targeting a "developer annoyance". There's no problem when using the shop as intended. The problem does occur when going through the checkout process until the shipment page and then PURGING the whole cart object in mongo db.  After creating the cart again and starting the checkout process again, the shipment is not clickable any more.

screen_shot_2016-01-13_at_17_54_06

### DIFF
--- a/packages/reaction-accounts/client/templates/addressBook/grid/grid.js
+++ b/packages/reaction-accounts/client/templates/addressBook/grid/grid.js
@@ -8,11 +8,9 @@ Template.addressBookGrid.helpers({
     });
 
     if (cart) {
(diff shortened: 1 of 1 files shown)


prinzdezibel:
This has an unwanted side effect. Namely, the step 3 is not disabled if one clicks on the trash button. Will be closing it therefore.

PR #698, opened 2016-01-13
Orders 570 - New Order Workflow

mikemurray:
A complete rework of the Core Order Workflow.

- Added collection hooks to prevent moving to the next workflow status if intermediate steps are incomplete. This reduces the complex login for the order push workflow and expands it into multiple methods
- Added a start button to start an order when you're ready

- Updated order workflow to allow non linear actions (jumping around between steps)
- Updated order page filters, reduced to 3 "New", "Processing", "Completed"
- Updated registry, added status for `coreOrderWorkflow` workflows. Workflow status is no longer determined by template.
- Updated schema, added status to workflow steps
- Updated look of order cards in order dashboard
- Updated the shipping "Invoice" workflow, replaced all generic text inputs with the `numericInput` component


aaronjudd:
@mikemurray there's some new failing lint issues.  I broken the already broken link in the codeclimate review section above while hopefully fixing the codeclimate implementation, but you can see the issues here https://codeclimate.com/github/reactioncommerce/reaction/compare/orders-570

Issue #656, opened 2016-01-06
Anonymous account and several browser tabs


newsiberian:
Hello,
when user interact with several tabs in one browser, we could see unexpected behavior of creation of anonymous user. Here is an illustration:
![tabs](https://cloud.githubusercontent.com/assets/11130487/12151257/0a1cf5fe-b4d0-11e5-8b7e-784afb7a9ed3.png)
Illustration is rough. I can not say exactly what sort of processes occur here.
Here is a video of this process:
![screencast-1452103572](https://cloud.githubusercontent.com/assets/11130487/12151513/540d7fa2-b4d1-11e5-996b-3c22cf104a9b.gif)

newsiberian:
related [forums.meteor.com topic](https://forums.meteor.com/t/tracker-autorun-runs-copies-of-computation-in-several-browser-tabs/15978). Any help is appreciated)

aaronjudd:
I'm not sure I understand the question, or the problem?

newsiberian:
@aaronjudd could you reproduce this bug? This issue describes a bug. On a forum - a question about, as I understand, about the way to solve it. Maybe wrong one...

zenweasel:
From @newsiberian Steps to reproduce:

1. Open more than 3 tabs with http:/localhost:3000
2. In one of these tabs log in as admin or register someone
3. Check all tabs. You should see that all tabs now logged in by user
4. Open developer tool in Recources tab (Crome) and look at LocalStorage. You should see you sessionId, userId
5. Now, in this log out from this user and look at changes in LocalStorage

newsiberian:
Hello, I did some research with situation with tabs and I have some thoughts about it.

I did few tests with #705. In first test I turned off all subscriptions, except `shop` in `main.js`, all routings, except route `/` and leave only `core`, `core-theme`, `reaction-accounts` .

I've tested with 3 browsers. Two were with 4 tabs opened on `localhost:3000` in each and one with one tab. During the test I first log in as normal user, and then log out. Expected result: – `userId` should be change only one time.

In such configuration I had got the following results:
clean means: `core`, `core-theme`, `reaction-accounts` packages

configuration | successful attempts
------------------ | ---------------------------
clean | 9/10
clean + inventory package | 6/10
clean + shipping package | 7/10
clean + two analytics packages | 7/10
clean + social | 6/10
full stack | 5/10

[...]

PR #705, opened 2016-01-16
Publications review

### LINKED ISSUE #664: User is not notified they are logged out in other tabs when logging out of one tab
Hello,

we have an issue with unexpected behavior on user logout when using several tabs [here](https://github.com/reactioncommerce/reaction/issues/656), but the most strange thing that the admin user have different behavior in the same situation.

When you attempt to logout by admin with several browser tabs, only active tab interact with server as expected. Other tabs not have such interaction. After you successfully logout, you could still see admin account inside non-active tabs by doing `Accounts.user()` or `ReactionCore.Collections.Cart.findOne()`.

newsiberian:
The main goal of this PR was to find the reason of #656 and #664.

I've added some inline-css for avatars in the hope that Mike will find the best place for these rules.

This PR fix #664 which originally was about "Admin account and several browser tabs".

#656 is not fixed here.

Also I turned off logout from other devices [here](https://github.com/reactioncommerce/reaction/blob/development/packages/reaction-accounts/client/templates/dropdown/dropdown.js#L22). Is this acceptable?

### DIFF
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -98,7 +98,6 @@ matb33:collection-hooks@0.8.1
 mdg:validation-error@0.1.0
 meteor@1.1.10
 meteor-base@1.0.1
-meteorhacks:inject-initial@1.0.3
 meteorhacks:ssr@2.2.0
 mikowals:batch-insert@1.1.13
 minifiers@1.1.7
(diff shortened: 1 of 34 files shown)


newsiberian:
Looks like CI needed `reaction-factories` update

zenweasel:
Verified fixes https://github.com/reactioncommerce/reaction/issues/664

zenweasel:
When I login as admin now I am getting a default avatar rather than the Admin avatar. Is that the expected behavior?

reaction___basic_reaction_product

aaronjudd:
There's a lot of comments left behind here that could be cleaned up. This PR really addresses two different things,  would be helpful if this was broken up. :+1:

PR #707, opened 2016-01-16
fixed #706 and uncomment "clone product" test

### LINKED ISSUE #706: Cloning product make new one with variants, which has the same ids like old one
I found this during testing UI and after this commented test in repo 
https://github.com/reactioncommerce/reaction/blob/development/packages/reaction-core/tests/jasmine/server/integration/products.js#L378

This test is fails due to error in method "products/cloneProduct".

ramusus:


zenweasel:
Approved

PR #713, opened 2016-01-18
Fix for inventory/register's batchInsert

newsiberian:
Hello. As you can see `batchInsert` doesn't know about SimpleSchema. The simplest way, I suppose, is manually add missing fields to inventory doc before `batchInsert` operation.


zenweasel:
Approved

PR #718, opened 2016-01-21
Fix for "product/tag" router

uniquarkD:
Updated the subscription for "product/tag" router.


packages/reaction-core/common/router.js
```diff
@@ -230,10 +230,8 @@ Router.map(function () {
     controller: ShopController,
     path: "product/tag/:_id",
     template: "products",
-    waitOn: function () {
-      return this.subscribe("Products", Session.get("productScrollLimit"));
-    },
     subscriptions: function () {
+      this.tagSubsription = this.subscribe("Products", Session.get("productScrollLimit"));
```
aaronjudd: `this.tagSubscription` (spelling)

uniquarkD: Oh, that's embarrassing.

aaronjudd:
@uniquarkD Can you post a short description about what this fixes? I just want to make sure we know what we're testing for before merging (will merge to a build tomorrow, if we can test).

uniquarkD:
Its infinite scrolling is reloading the page everytime we scroll down the page because we are returning its subscription on "waitOn" which triggers the global loading hook.

zenweasel:
@uniquarkD Hey, these changes are going to be irrelevant in the next release so we are closing pending v12.

PR #725, opened 2016-01-24
fixed back button for 'createPage' button from 'Flat Pages' package

ramusus:
I don't want this PR to be merged, but I want to have it to show how solve issue https://github.com/ramusus/reaction-flat-pages/issues/5 with back button while creating new Page from admin sidebar and user menu.

The problem: back button from new page form - undesirably creates another page, so it's necessary to hook event. Unfortunately I didn't find how hook this event from flat pages package. This PR solve this issue, but it's not clean way, while code of creating page is located here, not there


aaronjudd:
@ramusus I don't really think there is a better solution right now, however - the `reaction-routes` branch should be handling this quite differently in the near future. stay tuned.

PR #733, opened 2016-01-27
Only throw obsolete MongoDB error when actually using the bulk import API.

tdecaluwe:
Older versions can be used when not using the `ReactionImport` functionality.


aaronjudd:
Approved!

jshimko:
Approved!

PR #734, opened 2016-01-27
Translated left sidebar links and package names and description in dashboard

ramusus:


zenweasel:
Approved. Tested and it appears everything works correctly.

aaronjudd:
Do we need to have the label and keys in the registry?   We were automatically determining with this helper:

``` 
"registry": function () {
    // just some handle little helpers for default package i18nKey/i18nLabel
    let registry = ReactionCore.getRegistryForCurrentRoute("dashboard") || {};
    registry.nameSpace = registry.name || registry.template || "app";
    registry.i18nLabel = registry.label || registry.provides ||
      "Settings";
    registry.i18nKey = registry.nameSpace.toCamelCase() + "." + registry.i18nLabel
      .toCamelCase();
    return registry;
  }
```
This used to exist in the i18n helpers, really this should be a global function for getting package labels.

ramusus:
@aaronjudd implemented programmatic way of defining `registry.i18nLabel`

aaronjudd:
@ramusus :+1:  looks goods to me.  thanks for the update - you're not the first who wanted to put the labels in the registry, probably needs some additional documentation as well, but I think this is better, even though it's a little "magic".

ramusus:
@aaronjudd I'm only afraid of too many dependancies of auto generated i18nKey. I used only 2: stable `registry.provides` and unstable `registry.label`. I don't like using the second one because it could be changed in the future.

Auto generation of this value in your example for Action Views is much more complicated and dependent on many variables. I think we should simplify it as much as we could..

aaronjudd:
`label` could be changed, and it's easy to imagine, but I think we should view changes to the package's registry the same as any other global fixture / variable.  The original values are always preserved and available independently of the `Packages` collection - so that while we may allow UI editing of the `Packages` and `Translations` collections in the future, the `registerPackage`  data should be the stable data source.

zenweasel:
@ramusus Hey Boris, it looks like this PR introduced a new bug. When I move forward to this commit I start getting this error in the console:

```Exception in template helper: TypeError: Cannot read property 'toCamelCase' of undefined```

We need to get a fix for this if we can or I will need to revert this merge so that we can release a new version.

ramusus:
@zenweasel ok, I will check this out

ramusus:
@zenweasel, it seems like you have some new registry entry, which doesn't have `label` attribute.
I created new PR with checks if this attribute exists https://github.com/reactioncommerce/reaction/pull/771

Issue #708, opened 2016-01-18
Order in dashboard displays in local currency even though that currency was not used


zenweasel:
Some parts of the dashboard show the order in USD while other parts (the sidebar mostly) show it in the local currency. (PHP).
<img width="394" alt="reaction___djrjy7tmfdlca45uz" src="https://cloud.githubusercontent.com/assets/72819/12380948/92c2f6ea-bdba-11e5-8afc-727eda3b3334.png">

PR #737, opened 2016-01-27
Fixes for currency handeling in order dashboard

mikemurray:
resolved https://github.com/reactioncommerce/reaction/issues/708

Fixed: Numeric input now uses local currency
Fixed: Wrapping currency symbol in order list
Fixed: Discount field not being applied correctly


zenweasel:
Approved - verified fixed

Issue #719, opened 2016-01-21
We need a clarification of some aspects with product quantity


newsiberian:
Hello, there is something unclear to me in the product quantity logic and I want to figure it out. I want to show you how I see it.

We can have at least two situations with creating product. Admin can combine only top-level variants for his product:
![top_variant_2](https://cloud.githubusercontent.com/assets/11130487/12475229/4cd4b08e-c043-11e5-86d9-fab200849ab5.png)

or he could add an options:
![top_variant](https://cloud.githubusercontent.com/assets/11130487/12475235/523f4a16-c043-11e5-89b5-8071ab183a80.png)

First method looks good. We have top-level variants with 15, 10 and 20 items in each. They have their representation in `Inventory` collection. Everything looks correct to me.

If we look closely on the second picture, we will see that our T-shorts quantities now moved into `options`. We have here 20, 42 and 15. These items also has their representation in `Inventory`. BUT, we can notice that top-level variant also have 15 items. These 15 items also represented in `Inventory` collection. This is not correct from my point of view and could lead to various side effects.

So, I would like to clarify our view on the product quantities. If we add `options` to our top-level variant, we should remove `inventoryQuantity` from his doc completely or top-level `variant` could contain common quantity of all his options [with a second idea we don't need to do changes in client side at all] - some kind of denormalization -, and additionally remove all `Inventory` representations of this variant. Like on this pictures:
way 1:
![new_top_var_step_3_mod](https://cloud.githubusercontent.com/assets/11130487/12475365/525698d2-c044-11e5-81da-668a5b2859eb.png)

way 2:
![new_top_var_step_3_mod2](https://cloud.githubusercontent.com/assets/11130487/12475376/5b5672fe-c044-11e5-8656-bbad2437ada3.png)

What do you think? Or I missed something here?

newsiberian:
Additionally let's make it more clear, what the following terms mean for us. We are now talking about `productGrid` and products.

We have some notice helpers with methods:

* `isLowQuantity` – this method should return `true` if only one of many variants `quantity` < `lowInventoryWarningThreshold`? Or if every variants have a low `quantity`. I think if this notice suppose to be displayed for customer, we should use second logic, because if you want to buy an XXL size, but notice shows that one of variants (i.e. L size) has low quantity – it just will confuse.

[...]

Issue #225, opened 2014-12-04
Minimum Quantity on Product


taylorsmithgg:
Is there currently support for minimum quantities on products?

spencern:
Hey, we're going to leave the tag Pull Requests Encouraged and `needs-more-detail` on here, but we're going to close this issue because the requirements need to be defined better. If this is an issue that you would like to pursue, please create a new issue with more detail and more specific requirements and link to this old issue.

Issue #150, opened 2014-07-27
Product Inheritance Management


aaronjudd:
Manage product and variants through inheritance and overridden values.

When a product is cloned we add a cloneId to the product.
When a variant is cloned, we add a cloneId to the variant.

When you edit a product or variant attribute we're essential "overriding" this attribute if the object is a cloned object.  
-  Add functionality to reset the clones back / change to new  value set in the parent  hierarchy. 
-  Add functionality to ask change / edit only values that have not been changed in children
-  Optionally override even changed/edited child values
-  Update from anywhere in the hierarchy 
-  UI / UX to make this hierarchy visible / understandable.

The motivation to do this is for super fast mass product editing / updating.

tdecaluwe:
Is this planned to get into the beta release? How would you implement this? Would the products be nested documents in mongodb or are they supposed to be laid out flat, querying mongo for the complete subtree when needed?

aaronjudd:
@tdecaluwe I'm not sure if Product Inheritance Management will be in the beta - this is one of those that I'd really like to do asap though. Prior to this though I'd to be completely "finished" with the consumer cart->order->fulfillment flow. There's still some rough edges there, and once that's final we'll focus on the basic CMS, layout and publishing workflows. 

This feature though is critical for managing any kind of larger inventory so it's probably going to need to come as part of some product page refactor.   We have a rather vague issue https://github.com/reactioncommerce/reaction/issues/300 that just says "build product list view" and we're about to start addressing parts of that as we get into some layout and routing refactoring (see https://github.com/reactioncommerce/reaction/issues/502), as well as in the order fulfillment workflow @mikemurray is working on.  this is related to product and publishing workflows as well, so we will probably be breaking #300 issue out into several sub tasks - somewhere in that process we need to start laying the UI out for this product management.

aaronjudd:
@tdecaluwe the products, as well as the variants, all have a hierarchical set of ID's stored in them, referencing the product or variant that they were cloned from.  We'd be querying that and build a tree from there.

tdecaluwe:
@aaronjudd Doing subtree queries like that, can be become expensive in heavily nested trees. Instead of storing the parent it might be better to store the whole path (assuming moving queries will be used less)? Or even use a technique like nested interval trees, which is essentially the same, but coding the path as a continued fraction.

newsiberian:
[...]

PR #741, opened 2016-01-30
Products flattened data structure

newsiberian:
Hello, @tdecaluwe and I are introducing new way to work with products. This work is the result of the following [conversation](https://github.com/reactioncommerce/reaction/issues/150). The very beginning changelog could be found here: reactioncommerce/reaction-catalog#1.

So, the flattened multi schema – a way to keep documents with different schemas inside one collection. Combining with [ancestors data structure](https://docs.mongodb.org/manual/tutorial/model-tree-structures-with-ancestors-array/), we expect a speed boost from this approach. And, in my personal opinion, this is greatly simplified the logic of the products system.

There are a lot of changes that were done, so we need to test this PR carefully and a lot. To make flattened multi-schema works we [contributed changes](https://github.com/aldeed/meteor-collection2/pull/289) into `collection2` packages, but these changes so far hasn't approved and applied by **aldeed**. So, I hope, you help me to polish `collection2` changes to make it more appealing.

Perhaps the most daring change relates to Product publication. Currently we have all our products variants inside products docs. By getting products we have all in one. With flattened schema this is unbelievable. Instead of getting 10 product docs + 10 * 3-5-10 variants doc we fetch only 10 products docs. All data we need for `productGrid` we get via denormalization. Images we get via manipulations with `media.metadata.productId`. Also I like that we do computation only one time on write operation instead of every time on read operations. I've tested with 300 products in development branch vs this PR, and from the first glance this PR works a little bit better but still Blaze is the weak place.

Also, as a result of [this conversation](https://github.com/reactioncommerce/reaction/issues/719), the following logic was implemented for now:
> The original behavior was that if you had child variants, the top level variant would display the sum of all the child variants.

Here is a approximate changelog:

#### ancestors: Array
new field which contains ancestors of item. Currently it could be one or two `_ids`

#### parentId
**Removed**

#### cloneId
**Removed**

#### index: Number
Previously was a String. Since implementing of flattened model this property is used for keeping array index. This is needed for moving variants through list (i.e. drug'n'drop).

#### minOrderQuantity: Number
Minimum order quantity. It is used inside cart `quantityProcessing` function. This change somehow covers #225.

#### shopId: String
We need it.

#### ancestors: Array
For a product – ancestors array is always clean for now.

#### cloneId
**Removed**

#### variants
**Removed**

#### price: String
Keep in mind – this is a string. It is needed for denormalizing variants prices to display it in `productGrid`.

#### isLowQuantity: Boolean
Indicates when at least one of variants `inventoryQuantity` are lower then their `lowInventoryWarningThreshold`.

#### isSoldOut: Boolean
Indicates when all variants' `inventoryQuantity` is zero

#### isBackorder: Boolean
Indicates when the seller has allowed the sale of product which is not in stock

#### shopIdAutoValue
Method was modified. Now it ignores update operations.

## Methods

Removed a lot of lines with `this.unblock()` after I read what it does)) Please, read [this part of guide](http://guide.meteor.com/methods.html#methods-vs-rest), and [few](https://kadira.io/academy/meteor-performance-101/content/managing-waittime) [articles](https://meteorhacks.com/understanding-meteor-wait-time-and-this-unblock.html#why-thisunblock-does-not-always-work) from Kadira. I think this should be a fine tunning tool with understanding that putting it here really do not have any side effects. It is really dangerous to put it in almost every method, like I did before)

### Products

Almost all methods were rewritten. Some were removed.

##### products/cloneVariant
now accept only two arguments: `productId(String), variantId(String)`

#### products/createVariant
if we are inserting a child variant, we ran an additional function `flushQuantity()` which clean parent quantity and inventory records.

#### products/updateVariant
This method now is used only in one place – top-level variant autoform. It calls `denormalize` function to keep product denormalized fields up to date.

#### products/updateVariants
**Method removed** It is not needed anymore. It was used to manage UI sorting stuff. Now `products/updateVariantsPosition` do this. It updates only `index` field.

#### products/deleteVariant
Also fires `denormalize` function

#### products/cloneProduct
Now this method is really working as expected. It supports top-level variants and options media cloning.

#### products/updateProductField
Now work with all type of items. It also fires `denormalize` function.

#### products/updateVariantsPosition
**New Method** It updates top level variant position index (while UI sorting).

#### cart/copyCartToOrder
Now fires `orders/inventoryAdjust` method.

#### inventory/register
`batchInsert` package was removed. Method now used native `bulkOp`

#### inventory/backorder
Now use `bulkOp`

## Collections
`ReactionCore.Helpers.cartTransform` was refactored. All computations were moved to `getSummary` function.

## Hooks
* `ReactionCore.Collections.Products.before.update` renamed into `ReactionCore.Collections.Products.after.remove`
* added `ReactionCore.Collections.Products.after.insert` which previously was `modifier.$push` of `after.update`

## Functions
Template products helpers got a lot of changes. Besides, some of them we now using on server side,  so the idea is to move part of this methods to `/common` folder to global variable (`ReactionCore`) and after 1.3 make them export/imported.

#### getProductPriceRange
no need to describe

#### getVariantPriceRange
the same

#### getVariantQuantity
**New method** It calculates a sum of descendants `inventoryQuantity`.

#### getVariants
**New method** Get all parent variants.

#### getTopVariants
**New method** Get only product top level variants

#### denormalize
Function denormalized variants fields to product.

#### isSoldOut
We are stop accepting new orders if product marked as `isSoldOut`.

#### isLowQuantity
If at least one of the variants is less than the threshold, then function returns `true`.

#### isBackorder
Indicates, is product variants is still available to be ordered after summary variants quantity became zero.

#### flushQuantity
if variant `inventoryQuantity` not zero, function update it to zero. This needed in case when the option with it's own `inventoryQuantity` is created inside top-level variant. In that case top-level variant should display sum of his options `inventoryQuantity` fields.

## How to run
You need four packages to run this.
[collection2](https://github.com/newsiberian/meteor-collection2) branch `master`
[collection2-core](https://github.com/newsiberian/meteor-collection2-core) branch `flattened`
[reaction-factories](https://github.com/newsiberian/reaction-factories) branch `flattened`
[reaction-catalog](https://github.com/reactioncommerce/reaction-catalog) branch `master`

I would like to attract attention of @spencern because he use Reaction in production, and there are a lot of changes here.


aaronjudd:
I'd like to think about merging all this in v0.13 but we need a really detailed testing plan for this. We need to think about ways of merging this working is going to take some time.

newsiberian:
Totally agree, also I think, we should wait to what **aldeed** do with `collection2` changes.

newsiberian:
- product helper `globals.js` renamed to `global.js`, because we already have a file with such name
and it is ultra difficult to debug files with the same path and name;
- fixed `enable` button in dashboard package;
- in PDP from now `title` will be show to users, not `optionTitle`;
- **Temporary added `collection2` forks to packges** to make it easy access to people;
- `reaction-catalog` also **moved to packages folder**;

spencern:
I'm not sure how I missed this until today, but after a quick overview, it seems like this simplifies a lot of the complexity surrounding products, variants, and inventory management. I'll start watching this issue more closely, and echo the sentiment that great test coverage will be essential before a merge. 

Is the `collection2` merge just waiting on tests as well or are there other issues there?

aaronjudd:
@spencern I mentioned to @aldeed, hopefully he'll get time to review the PR that @newsiberian has pending this week. I think that's our primary blocker right now.

zenweasel:
I've started adding "acceptance tests" to the reaction-doc repo. If the idea of writing automated tests seems daunting we could start by writing a few simple manual tests that would just mimic how you are testing this as you develop. Then these tests can be part of the acceptance suite that we do before a release. Just a suggestion.

zenweasel:
We reviewed this PR in the weekly meeting and it's ready to be merge pending:

1. Merge to `collections2` repo (see note above)
2. Test plan
3. Resolution of conflict.

I am willing to work on a test plan with somebody if they want some help.

aaronjudd:
Merge to collections2 repo has been added now.

newsiberian:
> Test plan

added.

PR #743, opened 2016-01-30
code injection, integration with reaction-filtration

ramusus:
I don't know should we merge this PR into the core at this stage. May be there is way better to do the same. This is the way how is possible to integrate my package https://github.com/ramusus/reaction-filtration with current development branch of reaction core.
Guys, what do you think, Is there any better way to inject filtration functionality into the core?


ramusus:
Right now it adds link "Show Filters" into the bottom of the right side-bar on the product grid page only and allow reactive filtration for product grid output

packages/reaction-core/client/templates/products/productGrid/productGrid.html
```diff
@@ -1,4 +1,7 @@
 <template name="productGrid">
+
+  {{> contentFor region="dashboardControls" template="filtrationDashboardControls"}}
```
aaronjudd: This will need to be removed in v0.12 both `contentFor` and the placement of this.

aaronjudd:
I don't think you need to do this,  the v0.12 layout will allow you to inject these templates using a "layout".  We can hold merging and review after v0.12...

packages/reaction-collections/package.js
```diff
@@ -24,6 +24,7 @@ Package.onUse(function (api) {
   api.use("ongoworks:security@1.3.0");
   api.use("ongoworks:bunyan-logger@2.5.0");
   api.use("alanning:roles@1.2.13");
+  api.use("tmeasday:publish-counts@0.7.3");
```
aaronjudd: This package doesn't seem to be very well maintained,(source has deprecated Meteor code) and warns ```This package is designed primarily for correctness, not performance. That's why it's aimed at counting smaller datasets and keeping the count instantly up to date.```  Is this the best choice?

packages/reaction-core/common/router.js
```diff
@@ -230,22 +233,14 @@ Router.map(function () {
...
-        return {
-          tag: ReactionCore.Collections.Tags.findOne({
-            slug: id
-          }) || ReactionCore.Collections.Tags.findOne(id)
-        };
-      }
+      const id = this.params._id;
+      const tag = ReactionCore.Collections.Tags.findOne(id) || ReactionCore.Collections.Tags.findOne({slug: id});
+      Session.set("productFilters", {"tag": tag._id});
+      Deps.autorun(function () {
```
aaronjudd: This code is gone in v0.12, but I think `Deps` is deprecated for `Tracker`

packages/reaction-collections/package.js
ramusus: I agree completely, this notice also make me hesitate about using it. I will try to find another way of getting reactive count value.

packages/reaction-core/common/router.js
ramusus: Yes, I will update this package for 0.12 soon

ScyDev:
This doesn't work anymore with the latest code on development (yesterday).
Mainly because productGrid files have moved into another package, I guess.

ramusus:
@ScyDev I will fix it in short time

ramusus:
@ScyDev released 0.0.4 version, which is supports now version 0.12 of Reaction Core and updated this PR

newsiberian:
@ramusus, I still not happy, because of how you're using `check`. Maybe we could transfer this pub to SS validation pattern? [here](http://guide.meteor.com/data-loading.html#pagination). Then you could validate all your object for one time in one place. Good example how to use SS with complex objects could be found [here](https://github.com/reactioncommerce/reaction-stripe/blob/brent-rewrite-for-tests/server/stripeapi.js)

ramusus:
Yes, SS is good idea for this case and I agree, that validation and execution should be in separate places. May be there are package similar to ValidatedMethod,  but for Publications

ramusus:
implemented validation using SS for productFilters argument in Products publication. Now format is more readable, thanks @newsiberian. Also added translations

packages/reaction-collections/server/publications/products.js
```diff
@@ -7,12 +7,31 @@
 Meteor.publish("Products", function (productScrollLimit, productFilters) {
   check(productScrollLimit, Match.OneOf(null, undefined, Number));
```
newsiberian: We could simplify it like this:
```js
const filters = new SimpleSchema({
  "shops": {type: [String], optional: true},
  "tag": {type: String, optional: true},
  "query": {type: String, optional: true},
  "visibility": {type: Boolean, optional: true},
  "details": {type: Object, optional: true},
  "details.key": {type: String},
  "details.value": {type: String},
  "price": {type: Object, optional: true},
  "price.min": {type: Number},
  "price.max": {type: Number},
  "weight": {type: Object, optional: true},
  "weight.min": {type: Number},
  "weight.max": {type: Number}
})

Meteor.publish("Products", function (productScrollLimit, productFilters) {
  new SimpleSchema({
    productScrollLimit: { type: Number, optional: true },
    productFilters: { type: filters, optional: true }
  }).validator()
...
```
I think it should work not only in `validationMethod`. see [docs](https://github.com/aldeed/meteor-simple-schema#validating-and-throwing-validationerrors).  We don't need try/catch

ramusus: I already tried this and found, that we can not completely substitute check() with SS validator, because check package will throw error "not all arguments of publication are checked with check()".
And I like more verbose schema in my case, because it shows full structure of productFilters argument. Don't mistake ValidatedMethod with publication function. You are offering use the same strategy and imho this is right way, but there is no ValidatedPublication package for now, read [here](https://forums.meteor.com/t/package-idea-validated-publication/15516)

newsiberian: I have a feeling that we can't use `audit-arguments-check` and `SS.validator()` within one app. [here](https://forums.meteor.com/t/using-simpleschema-for-argument-check-in-publications-and-audit-arguments-check/16251)

ramusus: We can in the way I did. `audit-arguments-check ` only need that all publication arguments were checked using check(). After we can validate structure of any argument using SS validate()

zenweasel:
We reviewed this PR in the weekly meeting but we were unable to ascertain whether it was ready to merge as it seems there is an ongoing conversation (which we encourage). We leaving this open pending resolving all issues.

In addition this still has Code Climate issues that must be resolved and resolve merge conflict.

newsiberian:
Hello, this PR was merged unfortunately :smile: , but anyway, I've merged it into PR #741, and modify `products` publication. We could open an issue to continue discussion about this pub modification if you want.

aaronjudd:
@newsiberian yes, we'll need to clean it up.. we'll review it today, patch as needed for and open a new issue for the publication updates.

aaronjudd:
@ramusus I've reverted this. I think we need to look at this more. 
@newsiberian I reverted to an earlier Products publication, not sure if I missed updates for the flattened products..

PR #744, opened 2016-01-30
If we open the same Action View while it's already open - close it

ramusus:
PLUS action view explicit i18n settings

### DIFF
--- a/packages/reaction-core/client/main.js
+++ b/packages/reaction-core/client/main.js
@@ -141,8 +141,13 @@ _.extend(ReactionCore, {
    * @returns {String} Session "admin/showActionView"
    */
   showActionView: function (viewData) {
-    Session.set("admin/showActionView", true);
-    ReactionCore.setActionView(viewData);
+    if (Session.get("admin/showActionView") === true && viewData &&
+      JSON.stringify(Session.get("admin/actionView")) === JSON.stringify(viewData)) {
+        Session.set("admin/showActionView", false);
+    } else {
+      Session.set("admin/showActionView", true);
+      ReactionCore.setActionView(viewData);
+    }
   },
 
   isActionViewOpen: function () {
--- a/packages/reaction-core/client/templates/dashboard/settings/settings.js
+++ b/packages/reaction-core/client/templates/dashboard/settings/settings.js
@@ -12,7 +12,7 @@ Template.settingsHeader.helpers({
     let registry = ReactionCore.getActionView() || {};
     registry.nameSpace = registry.name || registry.template || "app";
     registry.i18nLabel = registry.label || registry.provides || "Settings";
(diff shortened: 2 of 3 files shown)


zenweasel:
@ramusus There's a couple nit-picky ESLint issues on this. Would you mind fixing those? (click the CodeClimate link above to see what they are)

packages/reaction-core/client/templates/dashboard/settings/settings.js
```diff
@@ -12,7 +12,7 @@ Template.settingsHeader.helpers({
     let registry = ReactionCore.getActionView() || {};
     registry.nameSpace = registry.name || registry.template || "app";
     registry.i18nLabel = registry.label || registry.provides || "Settings";
-    registry.i18nKey = registry.nameSpace.toCamelCase() + "." + registry.i18nLabel.toCamelCase();
+    registry.i18nKey = registry.i18nKey || (registry.nameSpace.toCamelCase() + "." + registry.i18nLabel.toCamelCase());
```
aaronjudd: requires https://github.com/reactioncommerce/reaction/pull/734 first otherwise this will error.  

See discussion there.

packages/reaction-core/client/main.js
```diff
@@ -141,8 +141,13 @@ _.extend(ReactionCore, {
...
   showActionView: function (viewData) {
-    Session.set("admin/showActionView", true);
-    ReactionCore.setActionView(viewData);
+    if (Session.get("admin/showActionView") === true && viewData &&
+      JSON.stringify(Session.get("admin/actionView")) === JSON.stringify(viewData)) {
+        Session.set("admin/showActionView", false);
+    } else {
+      Session.set("admin/showActionView", true);
+      ReactionCore.setActionView(viewData);
+    }
```
mikemurray: It should not automatically close if you open the same action view. It should stay open unless you explicitly close the panel, or navigate away from the page or view. This is because the current buttons that open Action Views do not indicate that by clicking them again will close them.

In the future when theres a need to do a toggle switch for the Action View, then we will implement something like this with all the required UI updates.

zenweasel:
@ramsus Hey Boris, so since I don't think we want the behavior that this PR introduces (closing action bar on second click) I am going to close this PR for now. If you wanted to implement a toggle as @mikemurray suggested then we could probably do this.

ramusus:
@zenweasel I would like to implement toggle, because I think this functionality is necessary. And also I think we should change url when any Action View is open. With Flow Router it's easy to implement. May be in the same PR. But I think @mikemurray mean the toggle should be implemented with React. I'm still not very familiar with React, for now I can do with Blaze only. And I think we should wait release of 0.12, where Flow Router will be introduced.

PR #748, opened 2016-01-31
Grid selection short keys - changed shift behaviour, added ctrl (cmd for mac)

ramusus:


zenweasel:
@ramusus There's a few ESLint and style issues that need to be fixed before we can merge it. Click on the Code Climate link to see what they are.

packages/reaction-core/client/templates/products/productSettings/productSettings.html
```diff
@@ -96,7 +96,7 @@ <h4 class="panel-title">Size</h4>
   {{else}}
     <div class="panel panel-default">
       <div class="panel-heading">
-        <span class="pabel-title">No Selected Products. Use the <i class="fa fa-gear"></i>, or shift + click on products to select multiple.</span>
+        <span class="pabel-title">No selected products. Use the <i class="fa fa-gear"></i> or shift|ctrl|cmd + click on products to select multiple.</span>
```
aaronjudd: I thought I fixed this a long time ago. Should be `<span class="panel-title">`

aaronjudd:
I'm cool with this - I prefer the cmd-click as long as it doesn't hijack anything else.  If we're going to use cmd though - shouldn't the shift -click be removed, or changed to a shift-contiguous behaviour?

packages/reaction-core/client/templates/products/productGrid/item/item.js
```diff
@@ -104,31 +104,28 @@ Template.productGridItems.helpers({
...
+          let indexes = [$('li.product-grid-item').index(checkbox.parents('li.product-grid-item')),
+            $('li.product-grid-item').index($('li.product-grid-item.active').get(0)),
+            $('li.product-grid-item').index($('li.product-grid-item.active').get(selected-1))];
+          for (let i=_.min(indexes); i<=_.max(indexes); i++) {
+            $('input[type=checkbox]', $('li.product-grid-item').get(i)).prop("checked", true).trigger("change");
+          }
+        } else {
+          checkbox.prop("checked", !checkbox.prop("checked")).trigger("change");
+        }
       }
```
jshimko: So these are small details, but there are a few best practice tweaks that should probably be made here.  First, there are several jQuery calls on the same elements... `$('li.product-grid-item')`, etc. which should be cached in a variable and reused. Second, all of those DOM queries can be scoped to the template by passing it into the event handler.  And finally, variables that aren't expected to change should be defined with `const` instead of `let` (This one wasn't really on you. I realize there were several `let` declarations there to begin with.  Just trying to clean up!)

So anyway, the event handler should look something like...

```js
"click [data-event-action=productClick]" (event, tmpl) {
  if (ReactionCore.hasPermission("createProduct")) {
    if (event.metaKey || event.ctrlKey || event.shiftKey) {
      event.preventDefault();
      
      const $items = tmpl.$('li.product-grid-item');
      const $activeItems = tmpl.$('li.product-grid-item.active');
      const $checkbox = tmpl.$(`input[type=checkbox][value=${this._id}]`);
      const selected = $activeItems.length;
      
      if (event.shiftKey && selected > 0) {
        const indexes = [
          $items.index($checkbox.parents('li.product-grid-item')),
          $items.index($activeItems.get(0)),
          $items.index($activeItems.get(selected-1))
        ];
        for (let i=_.min(indexes); i<=_.max(indexes); i++) {
          tmpl.$('input[type=checkbox]', $items.get(i)).prop("checked", true).trigger("change");
        }
      } else {
        $checkbox.prop("checked", !$checkbox.prop("checked")).trigger("change");
      }
    }
  }
}
```

ramusus: thanks a lot for suggestions! I added corrections. But there was some problems: tmpl var contains always only template with one product, so `tmpl.$('li.product-grid-item.active')` will not work as it should be.

ramusus:
@aaronjudd shift selection already changed to continuous behaviour - that was one of the reason of PR

packages/reaction-core/client/templates/products/productSettings/productSettings.html
ramusus: fixed

aaronjudd:
@ramusus :+1:

Issue #751, opened 2016-02-01
Tag page is empty during initial loading, products appears only after scrolling


ramusus:
I faced with this issue and it's related to global problem of filtering products, when we have them more, than scrollLimit value. We discuss it this morning with @newsiberian during analysis of https://github.com/ramusus/reaction-filtration/ package.

Steps how to reproduce:
1. create product1 with tag1 and after product2 with tag2
3. create 20 clones of product1 and after 20 clones of product2
4. add tag1 and tag2 to top navigation.
5. go to tag1 page and refresh.
5. go to tag2 page and refresh.

Tag1 page will contains 10 products, but Tag2 page will be without products after refreshing. That's because initial subscription contains only 10 products (and all of them are product1) and they are filtered by tag2 on client side - so result is empty. This is makes impossible to do any filtration and sorting on client side

zenweasel:
Verify this is still happening in development

ramusus:
I started with investigation this issue. That is my thoughts:
* move all filtration and sorting of products to the server side to subscription Products. 
* accomplish subscription Products for receiving reactive Session vars: `productScrollLimit`, `productFilters`, `productOrder`.
* update subscription to the Products in the routes `index` and `product/tag`.

aaronjudd:
Resolved with https://github.com/reactioncommerce/reaction/pull/875

PR #756, opened 2016-02-03
fix for #751

ramusus:


packages/reaction-core/common/router.js
```diff
@@ -156,7 +156,10 @@ Router.map(function () {
     name: "index",
     template: "products",
     subscriptions: function () {
-      this.subscribe("Products", Session.get("productScrollLimit"));
+      Session.set("productFilters", {});
+      Deps.autorun(function() {
```
zenweasel: Missing space

packages/reaction-core/common/router.js
```diff
@@ -230,22 +233,14 @@ Router.map(function () {
...
-        return {
-          tag: ReactionCore.Collections.Tags.findOne({
-            slug: id
-          }) || ReactionCore.Collections.Tags.findOne(id)
-        };
-      }
+      const id = this.params._id;
+      const tag = ReactionCore.Collections.Tags.findOne(id) || ReactionCore.Collections.Tags.findOne({slug: id});
+      Session.set("productFilters", {"tag": tag._id});
+      Deps.autorun(function() {
```
zenweasel: Missing space

zenweasel:
@ramsus Are you running ESLint with our `.eslintrc`? This would catch all the nit-picky format things.

packages/reaction-collections/server/publications/products.js
```diff
@@ -15,24 +15,30 @@ Meteor.publish("Products", function (productScrollLimit, shops) {
...
-      for (let thisShopId of shops) {
-        if (Roles.userIsInRole(this.userId, ["admin", "createProduct"],
-            thisShopId)) {
-          shopAdmin = true;
+    let selector = {shopId: shop._id};
+
+    if (productFilters) {
+      // handle multiple shops
+      if (productFilters.shops) {
+        check(productFilters.shops, Array);
```
newsiberian: hello,
maybe:  `if (Array.isArray(productFilters.shops)) {`?
and line 38: `if (typeof productFilters.tag === "string") {`?

did you test with both `shops` and `tag` provided. Is request valid without [AND](https://docs.mongodb.org/manual/reference/operator/query/and/)?

ramusus: I think we should check format of this argument as usual we do in the top of all publications.
Yes, I checked, AND is not necessary here

ramusus:
@zenweasel, installed eslint to webstorm, thanks for suggestion!

zenweasel:
@ramusus I found you needed to toggle "Prefer Strict Mode" in the Javascript pane to get ESLint to activate in WebStorm after installing. At least for me.

ramusus:
@zenweasel yes I did this, do you reformat code every time manually before saving this file?

zenweasel:
@ramsus Rarely need to reformat, and never need to save when using WebStorm. It saves automatically.

zenweasel:
@ramusus I am seeing an issue when testing this PR. If I follow the instructions for duplicating the error it appears that the product grid is no longer filtering by tag. For the second tag I created I get all products, even those without this tag.

aaronjudd:
Approved! this is most likely to get refactored in v0.12, but we'll see how this merges

PR #770, opened 2016-02-10
fix #766 error on signout

### LINKED ISSUE #766: Error when signing out
If I sign out after registering a user I get this error in the console and the "Ooops! You don't have permission to view this content"

```Exception in template helper: TypeError: Cannot read property '_id' of null```

Steps to reproduce:

1. Register a new user
2. Go to "Profile" page to view your orders
3. Select "Sign Out"
4. Receive error messages above.

Probably it just needs to redirect away from the profile page on signout? This might be a good "starter bug" for somebody who wants to contribute?

### PR DESCRIPTION
**Pull request by johannes-scharlach**
`typeof null === 'object'`
new comparison checks explicitly for null

This fixes #766

### DIFF
--- a/packages/reaction-accounts/client/helpers/helpers.js
+++ b/packages/reaction-accounts/client/helpers/helpers.js
@@ -49,7 +49,7 @@ Template.registerHelper("getGravatar", function (currentUser, size) {
     default: "identicon"
   };
   const user = currentUser || Accounts.user();
-  if (typeof user !== "object") return false;
+  if (user === null) return false;
   const account = ReactionCore.Collections.Accounts.findOne(user._id);
   // first we check picture exists. Picture has higher priority to display
   if (account && account.profile && account.profile.picture) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by newsiberian** [comment]
`user` also could be `undefined` here. this check is mainly against it.

**Comment by johannes-scharlach** [comment]
How would that happen?

If `user` is `undefined` here, something must have gone really wrong with `Accounts`, right? So in my opinion if that ever happens it's a bug and should be fixed properly. Otherwise we're silencing an exception, thereby basically making it impossible to ever find the actual problem.

If there is a good reason why user could ever be undefined, I would also prefer it to be checked explicitly:

```js
if (user === undefined || user === null) return false;
```

because if user is a number or a string, I would definitely want to see that exception happening, don't you?

zenweasel:
This fixes the error in the console, but doesn't fix the more important issue of redirecting them away from the profile page.

aaronjudd:
why not just `if (!user)` ?

johannes-scharlach:
@aaronjudd alright, that's more meteor style anyways.
@zenweasel I didn't know that that was the main issue, but saw what you meant once I got to that part of the code.

I hope the updated version of the PR does the job :)

zenweasel:
This was resolved in v12

PR #777, opened 2016-02-15
Fix #755 error. Orders always show created time as two minutes past the hour in Dashboard.

### LINKED ISSUE #755: Orders always show created time as two minutes past the hour in Dashboard
Regardless of when they were created the order list will show the correct hour and then :02. 

Trivial bug but putting it here so it can get tracked.

lijiming:
Fix the typo in the format of dateFormat in orders.html. Changed from "MM" for month to "mm" for minutes.


zenweasel:
Approved. Verified fixed.

PR #788, opened 2016-02-18
Fix for accounts to be compatible with the meteor-1.3

newsiberian:
Hello, we can't call method and variables by the same name


aaronjudd:
:+1:

PR #789, opened 2016-02-18
i18n labels fix

newsiberian:
Looks like logic in registry become more complex, so:
- fixed bug with creation of label for i18n;
- fixed a bug with translation payment **providers** (new term for methods;) );

before:
translation_fix

after:
translation_fix1

P.S. Translation for these and for `app.auto` will be provided later.

### DIFF
--- a/packages/reaction-core/client/helpers/apps.js
+++ b/packages/reaction-core/client/helpers/apps.js
@@ -165,8 +165,19 @@ function getReactionApps(optionHash) {
                 if (!registry.packageName) registry.packageName = app.name;
                 if (registry.enabled !== false) {
                   // TODO move this to a function and reuse with Template.dashboardHeader.helpers
-                  const registryLabel = registry.label ? registry.label.toCamelCase() : "";
-                  const i18nKey = `admin.${registry.provides}.${registryLabel}`;
+                  let registryLabel = "";
+                  let i18nKey;
+                  // first we check the default place for a label
+                  if (registry.label) {
+                    registryLabel = registry.label.toCamelCase();
+                    i18nKey = `admin.${registry.provides}.${registryLabel}`;
+                    // and if we don't find it, we are trying to look at first
+                    // registry entry
+                  } else if (app.registry && app.registry.length &&
+                    app.registry[0].label) {
+                    registryLabel = app.registry[0].label.toCamelCase();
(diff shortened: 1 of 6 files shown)


packages/reaction-core/client/helpers/apps.js
```diff
@@ -165,8 +165,19 @@ function getReactionApps(optionHash) {
...
                   // TODO move this to a function and reuse with Template.dashboardHeader.helpers
-                  const registryLabel = registry.label ? registry.label.toCamelCase() : "";
-                  const i18nKey = `admin.${registry.provides}.${registryLabel}`;
+                  let registryLabel = "";
+                  let i18nKey;
+                  // first we check the default place for a label
+                  if (registry.label) {
+                    registryLabel = registry.label.toCamelCase();
+                    i18nKey = `admin.${registry.provides}.${registryLabel}`;
+                    // and if we don't find it, we are trying to look at first
```
ramusus: Could you provide cases, when registry.label undefined? Why do want to take just first registry entry? It could lead to confusing for developer or translator what i18nKey is generated for the entry. I propose to keep them as simple as possible

newsiberian: Hi, sure:) I did it on screenshots, and you could repeat it manually.
1. Put breakpoint [here](https://github.com/reactioncommerce/reaction/blob/development/packages/reaction-dashboard/client/templates/dashboard/shop/settings/settings.js#L51)
2. go inside `ReactionCore.Apps`
3. then run [here](https://github.com/reactioncommerce/reaction/blob/development/packages/reaction-core/client/helpers/apps.js#L168)
4. check `registry` object

aaronjudd:
We have at least three spots where we are formatting the strings,  I was thinking that we should go ahead and a single method to provide the formatted string in an identical manner.

aaronjudd:
Also, a related item:  I'm thinking we should make  "registry.name" a required field on the registry schema, but need to figure the best way to autovalue from a sibling field if the name is empty.

newsiberian:
Hello,
> We have at least three spots where we are formatting the strings, I was thinking that we should go ahead and add single method to provide the formatted string in an identical manner.

done.

aaronjudd:
:+1:

PR #791, opened 2016-02-18
Few i18n fixes in dashboard

newsiberian:
- removed `Address` schema from `accounts.js`, because it is creates duplicate for `ReactionCore.Schemas.Address`;
- fixed few bugs with translation;


zenweasel:
In the future we like to keep PR's to just one issue at a time, but this just needs to get conflicts merged.

newsiberian:
resolved

PR #810, opened 2016-02-23
New shop method `shop/togglePackage`

newsiberian:
For usage instead of direct collection update.

### DIFF
--- a/packages/reaction-dashboard/client/templates/dashboard/packages/grid/package/package.js
+++ b/packages/reaction-dashboard/client/templates/dashboard/packages/grid/package/package.js
@@ -22,34 +22,38 @@ Template.gridPackage.helpers({
  * gridPackage events
  */
 Template.gridPackage.events({
-  "click .enablePkg": function (event, template) {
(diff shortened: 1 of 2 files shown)


aaronjudd:
:+1:  great, this was needed.

packages/reaction-dashboard/client/templates/dashboard/packages/grid/package/package.js
```diff
@@ -22,34 +22,38 @@ Template.gridPackage.helpers({
...
-        return Alerts.toast(
-          i18n.t("gridPackage.pkgDisabled", { app: i18n.t(self.i18nKeyLabel) }),
-          "warning"
-        );
       }
-    });
+    );
   },
-  "click .disablePkg": function (event, template) {
+  "click .disablePkg": function (event/* , template */) {
```
zenweasel: Hmmm, that's an interesting way to get around the "unused parameter" message while still leaving the signature in-tact. I think I might like that better than how I have been doing it.

zenweasel:
Approved

PR #816, opened 2016-02-24
[Meteor-1.3 compat] New dep to reaction-accounts

newsiberian:
Hello, added new dep `service-configuration` to `reaction-accounts`. W/o this service configuration not creates collection.
+ removed unused global var;


newsiberian:
I'm closing this for a while.

PR #837, opened 2016-03-01
create separate logging package

jshimko:
This removes the dependency on `ongoworks:bunyan-logger` and adds an internal logging package that accomplishes all of the same things.  This also paves the way for further expansion of logging functionality (coming soon).


aaronjudd:
reaction-collections, reaction-inventory, reaction-checkout also have `ongoworks:bunyan-logger` dependencies that need removing, and docs need updating in `logging.md`.

jshimko:
Ok.  I'll push the rest of that stuff out today.

Issue #804, opened 2016-02-22
[specification] i18n management improvement


newsiberian:
Hello, I hope Aaron forgive my arrogance. By this issue I want to asking for help in building i18n management system.

From my point of view it should:
- pick up translation files from custom packages. (i.e. from folder `pkgName/private/data/i18n/`);
- automatically update `translations` collection, each time packages added/removed;
- or at least update collection by clicking button in dashboard settings. Something like "Sync translations"
- or maybe you could offer another syncing mechanism for our purpose.

For better understanding: Why we are using such strange way for keeping translation? Because of SimpleSchema. [More info](https://github.com/aldeed/meteor-autoform/issues/483)

But the best decision I see is to write a `i18next-meteor-backend` like [this](https://github.com/i18next/i18next-node-fs-backend). With help of this package we could hold our translations inside minimongo + use all delights of i18next functionality. It could be pretty native way.

Feel free to self-assignment to this issue :wink:

ramusus:
I already thought about picking translations up from custom packages and merging everything in one big JSON. And tune i18next to use namespaces, variables inside strings and other cool stuff, But in case we will decide to use external tool for managing translations, I will completely support this decision, because this task is very complex.

And +1 for button "Sync translations" in admin. Also it would be helpful "Sync packages" (because of registry items) and "Sync shops" (because of layout features)

tdecaluwe:
@newsiberian Since the translations are currently stored as a nested tree, perhaps we could use similar approach to the new product/variant structure? :)

newsiberian:
@tdecaluwe, hello, could you explain more widely your suggestion? Why, you think we should do so?

tdecaluwe:
Now each language is a document in the db, I was just wondering if the collection wouldn't be easier to work with if we'd store each translation string as a separate document. For one thing, I know imports would be a lot easier.

aaronjudd:
let me think more about this, I'm reworking the i18n to use i18next 2.0.  2.0 brings some interesting stuff, like reading direct from files, and caching,  sprintf and interpolation... 

but let me get the current stuff working and then we can look into it some more what we should improve. 

re:  the structure - we're using the same structure as the i18n library (and with the upgrade it's now a 1-to-1 direct use of the json from the collection with no map necessary.  It's the lib that drove the schema of this.

newsiberian:
:+1:

PR #839, opened 2016-03-01
Temporary added I18n getLang function

newsiberian:
Hello, here is a temporary way to detect browser language till we find out what to do with `i18next-browser-languagedetector`. Also PR contains few translations


aaronjudd:
How did you know I couldn't figure out how to get that to work properly ;-)  .. need another set of eyes on it!

aaronjudd:
btw - re missing translations.  Set "debug: true" in the i81n init, and you'll see a list of the ones it's scanning.  We could just export all those, get them done once for en.json then use Transifiex to convert all the remaining languages.

packages/reaction-i18n/client/helpers/i18n.js
```diff
@@ -65,7 +82,7 @@ Meteor.startup(function () {
     // const defaultLanguage = lng.detect() || shopLanguage;
 
     // set default session language
-    Session.setDefault("language", defaultLanguage);
+    Session.setDefault("language", getLang());
```
aaronjudd: Should probably still have a fallback to shopLanguage.. just in case

newsiberian:
> How did you know I couldn't figure out how to get that to work properly ;-) .. need another set of eyes on it!

If you know how to use `i18next-browser-languagedetector`, I can remove `getLang` function from PR and leave only translations. Should I do this?

aaronjudd:
let's keep getLang for now.. I couldn't get i18next-browser-languagedetector to work for the life of me, complains about some missing `languageUtils`.

packages/reaction-i18n/client/helpers/i18n.js
newsiberian: are you sure? we already have `fallbackLng: shopLanguage,`

aaronjudd: yes, you're correct.

newsiberian:
In fact a spend some time to make `languageDetector` work with my react package, but I fail with it. I'm not much familiar with classes...

newsiberian:
Funny, I was not understand you first time I read you first message. According `languageDetector` I think we need a way to load translations from files, like was described in #804. I believe, we need our own `i18next-meteor-backend`, only with it help we can properly use `languageDetector` as part of i18next. Also, this means moving away from mongo `Translations` collection to minimongo-only `Translations` collection.

Issue #778, opened 2016-02-16
Guest checkout is allowed even when "Allow Guest Checkout" is unchecked


zenweasel:
Steps to reproduce:

1. Log in as admin
2. Uncheck "Allow Guest Checkout"
3. Log out
4. Add "Reaction Basic Product" to cart
5. Proceed to checkout
6. See that "Continue as Guest" is displayed and that you can checkout as a guest.

zenweasel:
Fixed in https://github.com/reactioncommerce/reaction/pull/1110

PR #850, opened 2016-03-03
Don't allow guest checkout when it is disabled

zenweasel:
Fix for issue #778 . The wrong registry setting was being checked, the logic was incorrect so it always returned true and this value was never used to control when the login was changed.

@mikemurray If you could take a quick look at what I did there and make sure that's how we want that handled? I saw some other (possibly unused?) logic about `anonymousWorkflow` but I didn't see how it was controlling the display here.

### DIFF
--- a/packages/reaction-accounts/client/templates/inline/inline.html
+++ b/packages/reaction-accounts/client/templates/inline/inline.html
@@ -1,6 +1,6 @@
 <template name="loginInline">
   <div class="accounts-dialog accounts-inline">
-    {{#unless currentUser}}
+    {{#if guestLoginOk}}
       <div class="col-md-6 pull-left checkout-guest">
         <div class="guest-checkout">
           <p class="text-justify" data-i18n="checkoutLogin.guestMessage">Continue as a guest, and you can create an account later.
@@ -17,6 +17,6 @@
       <div class="checkout-login">
         {{> loginForm startView='loginFormSignUpView'}}
       </div>
-    {{/unless}}
+    {{/if}}
   </div>
 </template>
--- a/packages/reaction-core/client/main.js
+++ b/packages/reaction-core/client/main.js
@@ -143,8 +143,7 @@ _.extend(ReactionCore, {
       ReactionCore.Log.debug("alowGuestCheckout: " + allowGuest);
       return allowGuest;
     }
-    Meteor.Error("400", "Package subscription not ready when allowGuestCheckout was checked");
-    return false;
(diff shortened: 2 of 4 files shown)


packages/reaction-core/client/main.js
```diff
@@ -133,18 +133,18 @@ _.extend(ReactionCore, {
...
+      let packageRegistry = ReactionCore.Collections.Packages.findOne({
+        name: "core",
+        shopId: this.shopId
+      });
+      allowGuest = packageRegistry.settings.public.allowGuestCheckout;
+      ReactionCore.Log.debug("alowGuestCheckout: " + allowGuest);
+      return allowGuest;
     }
-    return allowGuest;
+    Meteor.Error("400", "Package subscription not ready when allowGuestCheckout was checked");
```
aaronjudd: did you want to `throw new Meteor.Error`, or was this intended this way?

packages/reaction-accounts/client/templates/inline/inline.html
```diff
@@ -1,6 +1,6 @@
 <template name="loginInline">
   <div class="accounts-dialog accounts-inline">
-    {{#unless currentUser}}
```
aaronjudd: I believe `currentUser` is a Meteor.accounts template helper,  and we were extending with the guest logic so that the Accounts dropdowns, forms, all had the same behaviour.  are we sure the user accounts dropdown,etc are behaving the same? I'm seeing cases where the logged in user doesn't see the user accounts/profile and dashboard links, etc.

zenweasel: I didn't remove or change `currentUser`only refactored it and replaced it in this one location where it was incorrect.

zenweasel: Retested again and not seeing any issues with logged in user/dashboard, etc.

zenweasel: Ok, now I see it. Fixed

zenweasel:
Ok, this is now not showing products when the user is anonymous and anonymous login is disabled, even if I change back the template tags I changed. FFS. Also package subscription issues do not appear to be resolved always. Blocking for now.

zenweasel:
Ok, this is actually broken in the current version of `development`. If `allowGuestCheckout` ever returns `false` no products are shown. So the complexity of this has gone up a bit. Going to close this PR for now until I have a solution.

PR #851, opened 2016-03-03
Quickfix in `OnCreateUser`

newsiberian:


packages/reaction-accounts/server/methods/accounts.js
```diff
@@ -83,7 +83,7 @@ Accounts.onCreateUser(function (options, user) {
     // send a welcome email to new users,
     // but skip the first default admin user
     // (default admins already get a verification email)
-    if (!Meteor.users.find().count() === 0) {
+    if (!(Meteor.users.find().count() === 0)) {
```
jshimko: Why are the additional parens needed?  Also, a simpler equivalent to that could be:

```js
if (Meteor.users.findOne()) {
  Meteor.call("accounts/sendWelcomeEmail", shopId, user._id);
}
```

jshimko: Oh wait, nevermind.  Answered my own question there.  The simpler equivalent still stands though.

newsiberian: About using `findOne`: we need to block first and pass second, third, etc, so `findOne` can't help here.

aaronjudd:
:+1: tested, merging.

PR #852, opened 2016-03-03
Extra char causing Cart Remove Execution Flow to Break without warning

hrath2015:
Found extra char ~ while debug the remove from Cart flow in below line.
if (!~resetToIndex) return false;

### DIFF
--- a/packages/reaction-checkout/common/methods/workflow.js
+++ b/packages/reaction-checkout/common/methods/workflow.js
@@ -226,7 +226,7 @@ Meteor.methods({
     // get index of `newWorkflowStatus`
     const resetToIndex = workflow.indexOf(newWorkflowStatus);
     // exit if no such step in workflow
-    if (!~resetToIndex) return false;
+    if (!resetToIndex) return false;
     // remove all steps that further `newWorkflowStatus` and itself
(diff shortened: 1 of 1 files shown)


newsiberian:
Hello, lol, this is not just an extra char. Please, google: "Bitwise Not + indexOf"

hrath2015:
Thanks for pointing it out. Tried in Node env from command line and  it gave me undefined on console.

newsiberian:
Could you explain your use case? Maybe it will be better not to return false. Maybe throw something...

hrath2015:
As of now I am trying simple stuff to work. Remove an item from cart itself  not going through. All looks good in console print. But nothing changes in Mongo.

aaronjudd:
need to test still, but agree with @newsiberian might be better to throw an error, we have a lot of these "silent defaults" when we really should just be throwing an error and working to resolve the error.

newsiberian:
> need to test still, but agree with @newsiberian might be better to throw an error, we have a lot of these "silent defaults" when we really should just be throwing an error and working to resolve the error.

Hi, I don't think so. This PR has no deal with the bug reason. I think `workflow/revertCartWorkflow` are fine. No need to change to change anything it current method logic. I think we should close this PR.

PR #856, opened 2016-03-03
Fix for 853

newsiberian:
update `shopIdAutoValue` to deny setting `shopId` on update operations;

### DIFF
--- a/packages/reaction-schemas/common/common.js
+++ b/packages/reaction-schemas/common/common.js
@@ -17,7 +17,8 @@ _.extend(ReactionCore, {
     if (ReactionCore.getShopId()) {
       if (this.isSet && Meteor.isServer) {
         return this.value;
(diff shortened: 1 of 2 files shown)


packages/reaction-schemas/common/common.js
```diff
@@ -17,7 +17,8 @@ _.extend(ReactionCore, {
     if (ReactionCore.getShopId()) {
       if (this.isSet && Meteor.isServer) {
         return this.value;
-      } else if (Meteor.isServer || Meteor.isClient && this.isInsert) {
+      } else if (Meteor.isServer && !this.isUpdate ||
```
aaronjudd: The reason our editorconfig and eslint allow 160 chars is because we strongly believe that the 80 char norm is some legacy crap from 1980 and we're living in a new world with 3400 x 1400 screens.. we can afford longer lines..

aaronjudd:
:+1:

PR #857, opened 2016-03-03
Removed the check that prevented updating of packages on change

### LINKED ISSUE #819: Refresh Packages collection on Package.registry reload
Update `loadPackages` method to merge / update existing records when Package.registry entries are modified for packages.  Should be using ReactionImport methods for this.

mikemurray:
resolves #819 

Work in progress... (it couldn't have possibly been that easy??!!)

Removing the check around the import. This prevented the package registry from being updated when a package registry is updated.


aaronjudd:
The thing that we need to test for here is that when you have updates in reaction.json they take precedence over the package registry, and changes that have been updated in the admin take precedence over all... so on reload we maintain these edits.

aaronjudd:
Precendence: ui/db updates  ->  reaction.json -> packages registry

Ideally the changes in the db are always merged with changes the user has made in the UI/db and preserves these user changes.  For this to happen I think we need to add some revisioning to the the registry.  Why don't we add a "hashKey" to the registry, and give each registry a hash of the file contents. when the hash changes, we clone the users current data, update the db with the new file contents, then apply the user's changes back to matching keys in the registry. (object assign,etc)

So the expectation is that if a package registry or reaction.json changes, those updates are made, but that the users data always has precedence.  

@tdecaluwe any thoughts on this?

mikemurray:
This latest update now merges new properties from `reaction.json` and from package registry (on disk) into the `Packages` collection. It does not overwrite properties that are already defined. 

I'm thinking it would be better to build a tool and some settings in the dashboard that let you control how fixture data (from disk) is allowed to be imported into the database.

For example, we could have options for:
- always overwrite
    - json files will always overwrite changes in the DB
    - fields could be made read only so they couldn't be changed
- always merge (default)
    - new properties will be merged in
    - existing properties will be left alone

As well as a button to force an update from the fixture data on disk (think reset to default settings). 
While developing this would allow to test our changes quickly.

tdecaluwe:
@aaronjudd Short question: how would you know which parts are overridden by the user? Comparing the package db doc with the register contents?

aaronjudd:
@tdecaluwe yeah, thinking more about it we can only assume that the current db value is the correct value, but at least we can merge structural changes and new values with @mikemurray's updates.  I had an idea that you could hash this out (haha).. but that'd never work with structural changes..

mikemurray:
@newsiberian Object.assign seems to only do a shallow merge; just the top level keys. Since the objects are a but more complex, the sub objects weren't getting merged properly, just overwritten. This is why I used `lodash merge`

newsiberian:
Thanks, thats pretty helpful to know.

PR #862, opened 2016-03-04
fixes non-rtl bootstrap column issue

spencern:
`bootstrap.rtl.less` is no longer used in `reaction-default-theme` but the import stuck around in the custom theme template. This file causes bootstrap columns to get floated right, even if the language in use is not RTL and the `<html>` element does not have the `.rtl` class on it.

This PR fixes that issue.

### DIFF
--- a/packages/my-custom-theme-template/main.less
+++ b/packages/my-custom-theme-template/main.less
@@ -14,7 +14,8 @@
 @import "styles/variables.less";
 
 // Import Reaction Core Theme Styles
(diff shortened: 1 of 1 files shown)


aaronjudd:
:+1:
@mikemurray  need to remove the comment in https://github.com/reactioncommerce/reaction/blob/8632d5c22b7da79aee3f1aae27386827ff6d3a12/packages/reaction-default-theme/main.less#L64 as well

PR #874, opened 2016-03-08
[reaction-catalog] Implementing Validation Methods

newsiberian:
Hello, introducing my little experiment with validation methods. Originally I had an idea to move `uploadMedia` functionality to these methods, but transferring `Blob` object to the server side is so complicated thing..., so I decide to leave it and move other stuff.

### DIFF
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -98,7 +98,7 @@ localstorage@1.0.5
 logging@1.0.8
 matb33:collection-hooks@0.8.1
 mdg:validated-method@1.0.1
-mdg:validation-error@0.5.0
+mdg:validation-error@0.5.1
 meteor@1.1.10
 meteor-base@1.0.1
 meteorhacks:ssr@2.2.0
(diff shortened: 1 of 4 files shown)


zenweasel:
Shouldn't this PR include tests for these new methods?

aaronjudd:
:+1:

newsiberian:
Looks like there is `meteor` way to pass `blob` to server :smile: http://docs.meteor.com/#/full/ejson_new_binary

aaronjudd:
@newsiberian we use CFS to store blobs. (images)... in the db..  I'd take a look at that code

newsiberian:
yes, but the thing is -- this is work for us because we insert this blobs directly from client side [here](https://github.com/reactioncommerce/reaction/blob/development/packages/reaction-product-variant/client/templates/products/productDetail/images/productImageGallery.js#L46). It would be cool to move this logic to server side (for that we need to convert blobs), but I don't think this is so necessary...

PR #876, opened 2016-03-08
ProductGrid action bar translation

newsiberian:
Hello, I renamed it from `Edit Product` to `Product Settings`. I think this name more reflect the meaning of this bar


aaronjudd:
@mikemurray I think we'll need to review this to make for reactivity... but accepting as is for now.

Issue #865, opened 2016-03-04
MongoError: exception: _id index cannot be non-unique


ScyDev:
Happens during startup of reaction command with latest development changes from today.

```
/root/.meteor/packages/meteor-tool/.1.1.10.ki0ccv++os.linux.x86_64+web.browser+web.cordova/mt-os.linux.x86_64/dev_bundle/server-lib/node_modules/fibers/future.js:278
						throw(ex);
						      ^
MongoError: exception: _id index cannot be non-unique
    at Object.Future.wait (/root/.meteor/packages/meteor-tool/.1.1.10.ki0ccv++os.linux.x86_64+web.browser+web.cordova/mt-os.linux.x86_64/dev_bundle/server-lib/node_modules/fibers/future.js:398:15)
    at [object Object].MongoConnection._ensureIndex (packages/mongo/mongo_driver.js:790:1)
    at packages/aldeed_schema-index/lib/indexing.js:17:1
    at /var/www/src/.meteor/local/build/programs/server/boot.js:249:5
    - - - - -
    at Object.toError (/root/.meteor/packages/npm-mongo/.1.4.39_1.1klkbg5++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/mongodb/utils.js:114:11)
    at /root/.meteor/packages/npm-mongo/.1.4.39_1.1klkbg5++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/mongodb/db.js:1354:26
    at Server.Base._callHandler (/root/.meteor/packages/npm-mongo/.1.4.39_1.1klkbg5++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/mongodb/connection/base.js:453:41)
    at /root/.meteor/packages/npm-mongo/.1.4.39_1.1klkbg5++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/mongodb/connection/server.js:488:18
    at [object Object].MongoReply.parseBody (/root/.meteor/packages/npm-mongo/.1.4.39_1.1klkbg5++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/mongodb/responses/mongo_reply.js:68:5)
    at [object Object].<anonymous> (/root/.meteor/packages/npm-mongo/.1.4.39_1.1klkbg5++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/mongodb/connection/server.js:446:20)
    at [object Object].emit (events.js:95:17)
[...]

PR #879, opened 2016-03-08
refactor Docker build to be configurable and lean

jshimko:
This is step 1 in the effort to make the Docker image leaner and more configurable.  

The versions of Node, Mongo, and Phantom.js can now be specified via environment variables at build time.  Mongo and Phantom are now optional.  Mongo is _not_ installed by default (set `INSTALL_MONGO=true` to change that).  Phantom.js 2.1.3 _is_ installed by default, but can be skipped if you set `INSTALL_PHANTOMJS=false`. 

The ability to select versions also happens to fix the Mongo indexing issue mentioned in #865.  It looks like `aldeed:schema-index` will not work with Mongo 3.x, so installing Mongo 2.6.11 is the short term fix for that (and is now the default version set in the Dockerfile).  Not being able to use 3.x kind of sucks, so I'm trying to figure out a better way to fix that.

Aside from the usability gains, this refactor also significantly reduced the size of the final image.  Before these changes, the final build size with all layers was >2.2gb.  Now it's ~625mb.  This is mostly due to the fact that there are far less layers and a thorough cleanup happens at the end of each build layer (where relevant).

More tweaks likely on the way soon.


tdecaluwe:
Is there a reason you still include all of meteor? I'm thinking a production image shouldn't include meteor (see for example `reaction.mongo.docker` which is at 250mb), but a dev image should because of the (much) faster build times.

Also, this PR breaks building of the other docker fields present because of the removed `install-graphicsmagick.sh`.

Issue #887, opened 2016-03-09
UI popup alert for changing visibility of products on ProductGrid is always the same


ramusus:
UI popup alert for changing visibility of products on ProductGrid is always the same. When you make the product or products visible on invisible it always green and "is now visible"

ramusus:
I think it's time to implement tests for such UI behaviour. Because it's broke quite often in recent time and it takes time to get all those bugs known for developers.

aaronjudd:
@ramusus re: tests --> yes, agreed @zenweasel is taking the lead on that, but we're all waiting for some clarity in the Meteor testing for 1.3...

PR #905, opened 2016-03-11
Update product actions translation logic + fixed publushing bug

newsiberian:
Hello, here is a fix for #887. Also updated product actions translation logic. I put some of [`plural` logic](http://i18next.com/translate/pluralSimple/) in `delete` actions. Looks more complicated, but this is the only way I found to let it translate correctly, at least to Russian.


packages/reaction-product-variant/client/helpers/global.js
```diff
@@ -71,21 +72,33 @@ ReactionProduct.publishProduct = function (productOrArray) {
...
+            placement: "productGridItem",
+            id: products[0]._id,
+            autoHide: true,
+            dismissable: false
+          }
+        );
+      }
     }
-    if (!_.isArray(productOrArray)) {
+    if (products.length === 1) {
```
ramusus: I suppose, here is a bug in this case: when we do cloning on Product Grid page with only *one product selected* - we don't need redirect to the page with just created product. Redirect to another page should be only during cloning from the product page, so it is important difference between conditions *!_.isArray(productOrArray)* and *products.length === 1*

ramusus:
well done! :+1:

packages/reaction-product-variant/client/helpers/global.js
newsiberian: As far as I understand this logic was prepared for a future usage from PDP? Because currently we don't have cloning functionality within PDP. I'll revert it asap.

aaronjudd:
@newsiberian :+1: I think this is how the plurality is supposed to be implemented, looks good to me.

Issue #893, opened 2016-03-09
2 Icons for same function in Accounts Dashboard


hrath2015:
Not sure if this is on purpose. Looks like mix up with react components and old template.
![image](https://cloud.githubusercontent.com/assets/14071646/13648486/de9b905a-e65e-11e5-8bfa-7261bae89352.png)

aaronjudd:
:+1:  Yup, we need to remove the "Add Member" @mikemurray is aware of it.  A leftover from 11....

Issue #825, opened 2016-02-26
Auto insert package.layout.layout definitions into Shops.layout


aaronjudd:
on change / load of packages insert the primary layout structure definition into `Shops`.  Shops will have their own layouts, and can be altered in the Shops collections.

aaronjudd:
See related #819

PR #906, opened 2016-03-11
fix for #893

ramusus:


aaronjudd:
Although this does remove the duplicate I think that #911 fixes this in in a different way.  The structure for the layout had the wrong `dashboardControls` configuration.  As long as we have `dashboardControls": "accountsDashboardControls` in the structure, we'd need this template. I've updated the #825 to remove from the layout structure, so this is ok to merge as well.

aaronjudd:
:+1:

PR #909, opened 2016-03-11
Docker cleanup

jshimko:
- fix regressions with older Dockerfiles (`reaction.minimal.docker`, `reaction.mongo.docker`)
- fix broken bzip2 and phantom dependencies in `reaction.minimal.docker` and `reaction.mongo.docker`
- switch older Dockerfiles to `debian:jesse` base (because `wheezy` is considered an "[obsolete stable release](https://www.debian.org/releases/)")
- create a custom CI dockerfile (optimized to be leaner than devel and faster than production)
- additional cleanup and optimization of production/devel Dockerfiles 

A bunch of documentation updates to clarify using all of this stuff coming next.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -15,7 +15,7 @@ dependencies:
     - chmod +x bin/clone-packages.sh
     - ./bin/clone-packages.sh
     - if [[ -e ~/docker/image.tar ]]; then docker load -i ~/docker/image.tar; fi
-    - docker build -t reactioncommerce/prequel .
+    - docker build -f docker/reaction.ci.docker -t reactioncommerce/prequel .
     - mkdir -p ~/docker; docker save reactioncommerce/prequel > ~/docker/image.tar
 
 test:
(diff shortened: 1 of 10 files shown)


aaronjudd:
:+1:

circle.yml
```diff
@@ -29,13 +29,16 @@ deployment:
...
       - docker push reactioncommerce/prequel:$CIRCLE_BUILD_NUM
+      - docker push reactioncommerce/prequel:latest
   release:
     branch: master
     commands:
-      - docker build -t reactioncommerce/prequel .
+      - docker build -f docker/reaction.prod.docker -t reactioncommerce/reaction:$CIRCLE_BUILD_NUM .
+      - docker tag reactioncommerce/reaction:$CIRCLE_BUILD_NUM reactioncommerce/reaction:latest
       - docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
-      - docker tag reactioncommerce/prequel reactioncommerce/reaction:$CIRCLE_BUILD_NUM
```
aaronjudd: Why not reuse the build that already completed?

jshimko: Because the CI build that happens first is not using the production Dockerfile.  It uses a slightly different CI Dockerfile.  The exact same bash scripts are used in both cases, but the CI build caches all of the OS dependencies in their own layer so that subsequent builds are much faster.  As a result, the cleanup script isn't as efficient in the CI image and the container ends up being ~170MB larger.    

Also, that second build with the production Dockerfile only happens for code pushed to development or master branches.

PR #910, opened 2016-03-12
Product.positions refactoring

newsiberian:
Hello, there was a number of bugs with it. And we were using an array to store `positions` which is slow in our case, because, in that case we need to do heavy computations while sorting products.

### DIFF
--- a/packages/reaction-catalog/server/methods.js
+++ b/packages/reaction-catalog/server/methods.js
@@ -955,7 +955,7 @@ Meteor.methods({
       const updatedAt = `positions.${tag}.updatedAt`;
 
       return ReactionCore.Collections.Products.update({
-        "_id": productId
+        _id: productId
       }, {
         $set: {
           [position]: positionData.position,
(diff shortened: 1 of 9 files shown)


newsiberian:
Oh, man! I've just realize the main benefit of this approach. Now, in theory, we can completely remove sorting from client, by passing new `getTag` method's result to publication and there, we could have something like:
```js
const position = `positions.${tag}.position`;
const defaultSort = { sort: [position] }
```

ramusus:
:+1: sorting should be on the server side for sure. It's annoying to see, when new just uploaded products starts to reordering right after they appear on the site.

aaronjudd:
:+1:  approved, I tested - looks good - the important test here is that the positions and layout are specific per tag. (and that seems to be working well).

aaronjudd:
@newsiberian not sure why codeclimate didn't run for this PR, and I don't see a way to rerun, so I guess we don't get linting on this PR. :-(

newsiberian:
@aaronjudd, hello, if it is important, I can update this PR via some comment or something and codeclimate could rerun. Should I try this?

aaronjudd:
@newsiberian not that important, but I am curious..

aaronjudd:
Accepting, created an issue for the for-func-loop, need to decide how/if we really want to enforce that or not.

PR #911, opened 2016-03-13
Auto insert package.layout.layout definitions

### LINKED ISSUE #825: Auto insert package.layout.layout definitions into Shops.layout

### LINKED ISSUE #893: 2 Icons for same function in Accounts Dashboard

aaronjudd:
Loads layout from the package registry on package load
Removes necessity for including layouts in Shops.json
Provides backward compatibility for existing pre v0.12 Shops
Resolves #825
Resolves #893


jshimko:
:+1:

PR #914, opened 2016-03-14
Routing history 901

### LINKED ISSUE #795: page meta data refactor for v0.12
rewrite meta tag handling with kadira:docHead, ReactionCore.MetaData needs review/refactor as needed.

### LINKED ISSUE #901: browser back not reloading routes
browser back not reloading routes

aaronjudd:
- refactor ReactionCore.metadata to use kadira:dochead in
reaction-layout.  Replaces functionality for head updates, and relies
on the flow router triggers.enter for context.
- move metadata handling to reaction-router from layout
- remove ReactionCore.MetaData from globals
- Resolves #795
- Resolves #901


packages/reaction-product-variant/client/templates/products/productGrid/content/content.js
```diff
@@ -4,6 +4,8 @@
 
 Template.gridContent.helpers({
   displayPrice: function () {
-    return this._id && this.price.range;
+    if (this.price && this.price.range) {
```
ramusus: I suppose this.price is always object and it will always coerced to true value. Or there are cases, when it's equal to null or undefined ?

aaronjudd: this.price can be null, undefined, a Number, or an Object because it's defined with two different schemas (product, productVariant).  In the grid though, it will always be an object, but it can be empty if the price was never set on the first variant (you'd only see this on a new product)

jshimko:
:+1:

PR #916, opened 2016-03-14
Some inventoryQuantity and isSoldOut fixes

ScyDev:
- decrease inventory quantity
- prevent user putting more items in cart than are available (including the ones alread in cart)
- set product to isSoldOut=true if inventoryQuantity is < 1
- fixed check of variant quantity when denormalizing
- display only products that are not sold out

### DIFF
--- a/packages/reaction-catalog/server/methods.js
+++ b/packages/reaction-catalog/server/methods.js
@@ -185,6 +185,8 @@ function denormalize(id, field) {
   case "inventoryPolicy":
   case "inventoryQuantity":
   case "inventoryManagement":
+    ReactionCore.Log.info("denormalize() isSoldOut(variants): ",isSoldOut(variants),"doc.type:",doc.type,"id:",id,"variants:",variants);
+    
     Object.assign(update, {
       isSoldOut: isSoldOut(variants),
       isLowQuantity: isLowQuantity(variants),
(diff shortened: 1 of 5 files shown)


mikemurray:
There are some issues with this PR.

**prevent user putting more items in cart than are available (including the ones alread in cart)**
This seems to apply only to the current user in the current tab, not multiple users.

I have a product with 3 items in stock.
In one sessions (signed in) I can add only 3 of the variants.
In another session (anonymous) I can also add 3 same variants.
If I checkout in the anonymous session I get my 3 products.
If I checkout in the signed in session, I also get 3 products, even though the item should be sold out.

The current functionality is to allow for backorders, you users can checkout and the shop can fulfill the order when the items become available.

This will be configurable in the future.

---

Products need to still be visible even if they're sold out for various reasons. One being, as a shop owner, how do I add more inventory to an item thats sold out if I can't find it. Also, as a customer, even though the item may be sold out, I might have the option to backorder, or at least know that it exists.

This UI may change in the near future for this.

I can see the point to hiding a product thats no longer available, but that should be configurable behavior.

--

Other issues
You have 55 codeclimate / eslint issues that need resolving.
Your reaction log statements are too long. While we do have a 160 character line limit, they should be broken into multiple lines at that length.

As a last note; Thank you for taking the time to submit a PR, its always appreciated.

aaronjudd:
We can re-open this when it's cleaned up. Closing for now.

PR #936, opened 2016-03-21
Implementation of i18n namespaces support

newsiberian:
This is an implementation of namespaces support for `reaction-i18n`.
- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

To make it works with import, we have to add new `ns` field to `Translations` schema. Also, it is now always correctly load `en` translations within publication, don't know why, but this happens only on first load after reset and then works fine.

Package creators will need to copy directory structure to make this work: `private/data/i18n`


zenweasel:
Approved

PR #946, opened 2016-04-01
Fixed two bugs in variant-related helpers and changed behavior of progress bar

newsiberian:
Hello, why I changed progress bar behavior? Because, not all shop owners want to use inventory. This mean the products will be stored with zero `inventoryQuantity`, BUT we don't want to show transparent progress bar to customers, we want them to see green bar.

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

### DIFF
--- a/packages/reaction-product-variant/client/templates/products/productDetail/variants/variantList/variantList.js
+++ b/packages/reaction-product-variant/client/templates/products/productDetail/variants/variantList/variantList.js
@@ -6,31 +6,38 @@ Template.variantList.helpers({
     let inventoryTotal = 0;
     const variants = ReactionProduct.getTopVariants();
 
-    if (variants.length > 0) {
+    if (variants.length) {
       // calculate inventory total for all variants
-      for (let variant of variants) {
-        let qty = ReactionProduct.getVariantQuantity(variant);
+      variants.forEach(variant => {
+        const qty = ReactionProduct.getVariantQuantity(variant);
         if (typeof qty === "number") {
           inventoryTotal += qty;
         }
-      }
+      });
       // calculate percentage of total inventory of this product
-      for (let variant of variants) {
-        let qty = ReactionProduct.getVariantQuantity(variant);
+      variants.forEach(variant => {
+        const qty = ReactionProduct.getVariantQuantity(variant);
         variant.inventoryTotal = inventoryTotal;
-        variant.inventoryPercentage = parseInt(qty / inventoryTotal * 100, 10);
+        if (inventoryTotal) {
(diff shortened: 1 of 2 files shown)


packages/reaction-product-variant/client/templates/products/productDetail/variants/variantList/variantList.js
```diff
@@ -6,31 +6,38 @@ Template.variantList.helpers({
     let inventoryTotal = 0;
     const variants = ReactionProduct.getTopVariants();
 
-    if (variants.length > 0) {
+    if (variants.length) {
       // calculate inventory total for all variants
-      for (let variant of variants) {
-        let qty = ReactionProduct.getVariantQuantity(variant);
+      variants.forEach(variant => {
```
zenweasel: Generally I am strongly against changes to working code that do nothing but change the method by something is accomplished if there is no driving reason beyond the author's preferred style. (If I am missing a reason here, let me know).

I addition, I would say that we should/do prefer the easier to read iterator version over the `forEach` version.

newsiberian: I'm answered [here](https://github.com/reactioncommerce/reaction/pull/950#discussion_r58326180). In my opinion we should use Array iteration methods for iterate Arrays. Think about it like moving project to more ES6-like code style. Is that make sense?

zenweasel:
Approved

packages/reaction-product-variant/client/templates/products/productDetail/variants/variantList/variantList.js
mikemurray: The `for of` didn't need to be changed here. Also, I believe `Array.forEach` came with ES5 (JS 1.6), and `for of` is ES2015 (aka ES6). If you want to adopt a more ES6 like coding style, then you should be using `for of`.

Native array iterators are useful such as `find`, `every`, `map`, etc. In general `for of` will do everything the native `Array.forEach` method will, except give you the index. `for of` loops are also breakable.

tl;dr Prefer `for of` loops. Use the other native iterators as needed.

See other [comment here](https://github.com/reactioncommerce/reaction/pull/950/files#r58445986)

mikemurray:
It seems you need to not only disable inventory tracking on each variant, but also set the quantity to 0 in order to keep the bars green, and remove any `sold out` / `backorder` warnings.

I'd say, if inventory tracking was disabled then the bars should stay green.

basic_reaction_product

aaronjudd:
:+1:  approving, as this doesn't hurt existing functionality... but still needs some more improvements..

PR #948, opened 2016-04-01
Fix for pdp variant price range

newsiberian:
It were displayed not top variant's child variants price range but product price range.

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

### DIFF
--- a/packages/reaction-product-variant/client/templates/products/productDetail/productDetail.js
+++ b/packages/reaction-product-variant/client/templates/products/productDetail/productDetail.js
@@ -42,10 +42,12 @@ Template.productDetail.helpers({
     const current = ReactionProduct.selectedVariant();
     if (typeof current === "object") {
       const childVariants = ReactionProduct.getVariants(current._id);
+      // when top variant has no child variants we display only its price
       if (childVariants.length === 0) {
         return current.price;
       }
-      return ReactionProduct.getProductPriceRange().range;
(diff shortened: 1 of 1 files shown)


zenweasel:
Approved

mikemurray:
:+1:

aaronjudd:
I'm not sure this is the right behaviour.  Products range should be a range that is inclusive of all variants and options.  A variant price range should only displayed when you select the variant, or in the variant "inventory bar". The variant range is inclusive of all options.

newsiberian:
Hello, @aaronjudd, in PDP we always have selected variant. Try to switch between default product first and second top variants and you'll noticed that first top variant price are always be a product price.

I'm not sure it is better to show all variants prices at once...

zenweasel:
@newsiberian That's not what I am seeing here. The product price always is reflecting all variants not just the first

newsiberian:
pdp_price_bug

aaronjudd:
@newsiberian your first picture shows the problem.  the first item IS selected, and should be showing 12.99 - 19.99 because it is selected. When you first view a product you do not have any variant or options selected by default, and it show the whole range.

newsiberian:
Hello, we always have selected variant. If it were not so, then you shouldn't see any images because our products can't have images for some time. When you fist enter to PDP you always have first variant selected, at least this is the current behavior.

PR #950, opened 2016-04-02
Fix for `publishProduct`

newsiberian:
We could have a situation when admin adds product, create clean top variant, then creates child variants and work only with them. I mean he not fill top variant fields like `price`, `quantity`. This is pretty common use case. Our logic now allowed to do such things.

Also I've added a billet for a new test, but I don't see the point to write this test right now.

- [ ] Has tests.
- [ ] Has docs.
- [x] Passes all tests

### DIFF
--- a/packages/reaction-product-variant/client/templates/products/productDetail/productDetail.js
+++ b/packages/reaction-product-variant/client/templates/products/productDetail/productDetail.js
@@ -199,17 +199,20 @@ Template.productDetail.events({
       template.$(".title-edit-input").focus();
     }
     const variants = ReactionProduct.getVariants(self._id);
-    for (let variant of variants) {
-      let index = _.indexOf(variants, variant);
+    variants.forEach((variant, index) => {
       if (!variant.title) {
         errorMsg +=
           `${i18next.t("error.variantFieldIsRequired", { field: i18next.t("productVariant.title"), number: index + 1 })} `;
       }
-      if (!variant.price) {
-        errorMsg +=
-          `${i18next.t("error.variantFieldIsRequired", { field: i18next.t("productVariant.price"), number: index + 1 })} `;
+      // if top variant has children, it is not necessary to check its price
+      if (variant.ancestors.length === 1 && !ReactionProduct.checkChildVariants(variant._id) ||
+        variant.ancestors.length !== 1) {
+        if (!variant.price) {
+          errorMsg +=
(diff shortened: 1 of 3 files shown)


packages/reaction-product-variant/client/templates/products/productDetail/productDetail.js
```diff
@@ -199,17 +199,20 @@ Template.productDetail.events({
       template.$(".title-edit-input").focus();
     }
     const variants = ReactionProduct.getVariants(self._id);
-    for (let variant of variants) {
-      let index = _.indexOf(variants, variant);
+    variants.forEach((variant, index) => {
```
zenweasel: Again, why are we wholesale changing all iterators to `forEach`'s? Does this fix a bug?

newsiberian: Because, `For - of` - is for Objects, not for Arrays. For arrays es6 provides to as: `forEach`, `map`, `reduce`, `filter`, etc. Look at the [list of available Array methods](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach).
So in my opinion it is native to use Array iteration methods to iterate through it.

zenweasel: I think you are thinking of `for in` not `for of`. Please recheck the documentation. This is proper ES6 syntax.

newsiberian: Ah, yes, you're right, but still I'm thinking `For..of` is more for objects iteration.
Anyway, if you guys decide to leave `for...of`, I'll try to remember this and not change them in future.

zenweasel:
Approved

packages/reaction-product-variant/client/templates/products/productDetail/productDetail.js
mikemurray: [for of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) works for for any iterable object such as an Array, String, TypedArray, Map, Set, generators, and NodeList; but necessarily every kind of Object. 

The usage `Array.forEach` can work too, especially, if you need to access the index.

In general. Prefer `for of`.

aaronjudd: I'd like to see `forEach` reserved for use with Meteor cursor iterations and nothing else.

PR #951, opened 2016-04-03
Fix for the `productGrid` products loading

newsiberian:
Hello. `ProductsGrid` loads variants after backing from PDP via old-PDP subscription. It's just a little moment till subscription will be updated. T consider that, we have to specify our `selector`.

We have two ways to specify selector: `type: { $in: ["simple", "bundle", etc..] }` or `ancestors: []`. I decide to use second, because it is shows more respect to package creators. With this selector, they no need to add one more incompatibility to their projects.

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests


zenweasel:
Approved

aaronjudd:
:+1:

PR #965, opened 2016-04-11
Fix typos in getReactionApps function

newsiberian:
- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests


packages/reaction-core/client/helpers/apps.js
```diff
@@ -157,7 +157,7 @@ function getReactionApps(optionHash) {
           for (key in registryFilter) {
             // make sure we're dealing with valid keys
             if ({}.hasOwnProperty.call(registryFilter, key)) {
-              value = registryFilter[key];
+              let value = registryFilter[key];
```
zenweasel: This variable is redundant. You could just use registryFilter[key] in line 161

PR #969, opened 2016-04-12
Meteor 1.3.x

aaronjudd:
This introduces Meteor 1.3 and the initial working implementation imports / modules.  This is not a restructure of the existing codebase, but a functional step towards that goal, which we'll implement as we go.  You must run `npm install` at the app root before running this version.


aaronjudd:
:+1:

zenweasel:
+1

PR #980, opened 2016-04-25
proposed fix to assignRoles.js

sngn:
The assignOwnerRoles method in reaction-core/server/registry/assignRoles.js doesnt do what it says in its description.

Issues

1. The code tries to copy Arrays via simple assignment, that way globalRoles array becomes larger than intended
2. I think the process of selecting owners is flawed, it selects everyone who has a default (or package) permission set in Roles.GLOBAL_GROUP.
Might technically work, but might be a problem with more fine grained access controls (ie users who are global admins but not global owners).
3. "createProduct" is a permission defined in reaction-product-variants package and should be set automatically in every shop for (global_group) owners, without it being included in the defaultRoles array. </br>
Removing it from there has the consquence of owners not having the permission in global_group, which should theoretically be irrelevant (looking at hasPermission, the permission is checked against the current shop permissions)

Fix

- Make new arrays real copies (copy by value)
  * Split defaultRoles variable from further processing, so that it could possibly be imported from configuration
- Reduced defaultOwnerRoles array to role "owner". Also query should be less expensive.
- Removed "createProduct" permission from defaultRoles array because I think it should not be there.

- [ ] Has tests.
- [ ] Has docs.
- [x] Passes all tests. Well, I tested it, but not with automated tests 
- [x] Has been linted and follows the style guide, I think

### DIFF
--- a/packages/reaction-core/server/registry/assignRoles.js
+++ b/packages/reaction-core/server/registry/assignRoles.js
@@ -41,9 +41,9 @@ const getRouteName = (packageName, registryItem) => {
 
 ReactionRegistry.assignOwnerRoles = (shopId, pkgName, registry) => {
   const defaultRoles = ["owner", "admin", "guest"];
-  var packageRoles = defaultRoles.slice();
+  let packageRoles = defaultRoles.slice();
   packageRoles.push(pkgName);
-  var globalRoles = packageRoles.slice();
+  let globalRoles = packageRoles.slice();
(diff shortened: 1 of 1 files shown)


aaronjudd:
👍  I've tested this and it seems to work fine for me. The only note, is the `Roles.GLOBAL_GROUP` should only be assigned and used the installer's original admin account.

PR #984, opened 2016-04-28
Fixed issue with address book and added addesses

### LINKED ISSUE #827: Address Book Display in Cart Checkout- unusual Behavior
Adding the address is working fine.  I could see in Database.  When Created a new order I am asked to create address again though there is one address already available. 
 **Now the fun begins while debug**
After I put 2 console.log statement in add.js and addressbook.js oncreated() method, it starts working fine and shown the grid with addresses.  Not sure if it is due to Flow Router change or something else?? .  Will continue to debug and see if I found something will send a PR.

mikemurray:
Addresses issue where address book would always show the add address
form, even though a previous address was already added and saved.

resolves #827

### DIFF
--- a/packages/reaction-accounts/client/templates/addressBook/addressBook.js
+++ b/packages/reaction-accounts/client/templates/addressBook/addressBook.js
@@ -6,26 +6,27 @@
  */
 
 Template.addressBook.onCreated(function () {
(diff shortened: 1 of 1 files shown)


zenweasel:
Approved

PR #993, opened 2016-05-02
Slightly refactored `orders/shipmentPacked`

newsiberian:
Hello, the problem with method was -- it doesn't return result. Now it does.

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

…r place

and MAIN - now method returns result;

### DIFF
--- a/packages/reaction-core/server/methods/orders.js
+++ b/packages/reaction-core/server/methods/orders.js
@@ -56,24 +56,25 @@ Meteor.methods({
     check(order, Object);
     check(shipment, Object);
     check(packed, Boolean);
-    this.unblock();
 
     if (order) {
-      ReactionCore.Collections.Orders.update({
-        "_id": order._id,
-        "shipping._id": shipment._id
-      }, {
-        $set: {
-          "shipping.$.packed": packed
-        }
-      });
-
       // Set the status of the items as shipped
       const itemIds = shipment.items.map((item) => {
         return item._id;
       });
 
-      Meteor.call("workflow/pushItemWorkflow", "coreOrderItemWorkflow/packed", order, itemIds);
+      const result = Meteor.call("workflow/pushItemWorkflow", "coreOrderItemWorkflow/packed", order, itemIds);
+      if (result === 1) {
+        return ReactionCore.Collections.Orders.update({
+          "_id": order._id,
+          "shipping._id": shipment._id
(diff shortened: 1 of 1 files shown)


packages/reaction-core/server/methods/orders.js
```diff
@@ -56,24 +56,25 @@ Meteor.methods({
     check(order, Object);
     check(shipment, Object);
     check(packed, Boolean);
-    this.unblock();
```
aaronjudd: Do we want to block here?

packages/reaction-core/server/methods/orders.js
```diff
@@ -56,24 +56,25 @@ Meteor.methods({
...
-        }
-      });
-
       // Set the status of the items as shipped
       const itemIds = shipment.items.map((item) => {
         return item._id;
       });
 
-      Meteor.call("workflow/pushItemWorkflow", "coreOrderItemWorkflow/packed", order, itemIds);
+      const result = Meteor.call("workflow/pushItemWorkflow", "coreOrderItemWorkflow/packed", order, itemIds);
```
aaronjudd: could/should this be a callback..?

aaronjudd:
Approved.   with a caveat that this shouldn't break our current UI because we don't wait for anything.. and presumably the UI (that @newsiberian is using) could wait on this result.... might be questionable though if this was part of a bulk server operation

aaronjudd:
Approved.   with a caveat that this shouldn't break our current UI because we don't wait for anything.. and presumably the UI (that @newsiberian is using) could wait on this result.... might be questionable though if this was part of a bulk server operation

packages/reaction-core/server/methods/orders.js
newsiberian: I removed this because we have one more internally called method here. And it is still not 100% clear to me how `this.unblock()` could affect in such situation, so I believe it is safer to remove it. Anyway it is small method. Here is nothing to unblock from my point of view,

packages/reaction-core/server/methods/orders.js
newsiberian: This is exactly what we need. With callback we could don't get a result, because method call will be async with callback. We need 100% to have a method call result, that's why we should wait till it will be computed.

PR #994, opened 2016-05-02
Refactored `orders/shipmentShipped` method

newsiberian:
Hello, commit text explains all changes.

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

### DIFF
--- a/packages/reaction-core/server/methods/orders.js
+++ b/packages/reaction-core/server/methods/orders.js
@@ -160,35 +160,42 @@ Meteor.methods({
    *
    * @summary trigger shipmentShipped status and workflow update
    * @param {Object} order - order object
-   * @return {Object} return workflow result
+   * @param {Object} shipment - shipment object
+   * @return {Object} return results of several operations
    */
-  "orders/shipmentShipped": function (order) {
+  "orders/shipmentShipped": function (order, shipment) {
     check(order, Object);
     this.unblock();
 
-    if (order) {
-      let shipment = order.shipping[0];
+    let completedItemsResult;
+    let completedOrderResult;
 
-      // Attempt to sent email notification
-      Meteor.call("orders/sendNotification", order);
+    // Attempt to sent email notification
+    const notifyResult = Meteor.call("orders/sendNotification", order);
 
-      const itemIds = shipment.items.map((item) => {
-        return item._id;
-      });
+    const itemIds = shipment.items.map((item) => {
+      return item._id;
+    });
 
-      Meteor.call("workflow/pushItemWorkflow", "coreOrderItemWorkflow/shipped", order, itemIds, (error) => {
-        // Move to completed status for items
-        // TODO: In the future, this could be handled by shipping delivery status
-        if (!error) {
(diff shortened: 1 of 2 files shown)


packages/reaction-core/server/methods/orders.js
```diff
@@ -160,35 +160,42 @@ Meteor.methods({
...
 
-    if (order) {
-      let shipment = order.shipping[0];
+    let completedItemsResult;
+    let completedOrderResult;
 
-      // Attempt to sent email notification
-      Meteor.call("orders/sendNotification", order);
+    // Attempt to sent email notification
+    const notifyResult = Meteor.call("orders/sendNotification", order);
```
zenweasel: does `Email.send` even return a value? And I don't think that function returns a value on error either

packages/reaction-core/server/methods/orders.js
```diff
@@ -160,35 +160,42 @@ Meteor.methods({
...
+      completedItemsResult = Meteor.call("workflow/pushItemWorkflow", "coreOrderItemWorkflow/completed", order, itemIds);
+
+      if (completedItemsResult === 1) {
+        // Then try to mark order as completed.
+        completedOrderResult = Meteor.call("workflow/pushOrderWorkflow", "coreOrderWorkflow", "completed", order);
+      }
     }
+
+    return {
+      notify: notifyResult,
```
zenweasel: These variables should be renamed for clarity. Especially "workflow" since you aren't returning a workflow.

zenweasel:
I'm unclear how I can verify that this does what is expected it to do and what it is trying to fix

aaronjudd:
@newsiberian starting with 0.14 - our next release, we won't even review these if the description isn't sufficient for us to understand what the changes are, and there is not a linked issue.

packages/reaction-core/server/methods/orders.js
newsiberian: Partially it returns `true`. And we could also refactor `orders/sendNotification` to make it return something, because I think this is important to admin to know was notification really sent or not.

packages/reaction-core/server/methods/orders.js
newsiberian: Ok, any suggestions how which names could be used?

newsiberian:
@aaronjudd, I'm sorry for that. You're right. The mail reason for that PR and similar was -- method doesn't return any results.

packages/reaction-core/server/methods/orders.js
aaronjudd: I think `workflowResult` would be ok..

newsiberian: I'm sorry, I don't understand. Now we have an result object:
```js
{
  notify,
  workflow,
  completedItems,
  completedOrder
}
```
Should I rename some of these vars to `workflowResult` or that should I do?

zenweasel: Yes, rename the variable `workflow` to `workflowResult` because you are not returning a workflow, rename the variable `notify` to `notifyResult`, etc.

PR #1001, opened 2016-05-03
Product grid 987

### LINKED ISSUE #987: Product grid subscription tweaks

### Expected behavior
Products in the product grid should show only when the subscription is really ready.

### Actual Behavior
When switching between views, due to the subscription updating, the product grid looks incomplete; then products pop in creating a jarring experience.

Switch between 2 tags in the product grid

mikemurray:
Resolves #987 

Avoiding using find().fetch() to get products as it seems to cause slow
loading behavior.

Reduced unnecessary duplicate fetches / template
redraws which should help with performance.


mikemurray:
It seems to work when there are less than 24 products in the grid. Any more than that, and there still seems to be some kind of loading delay / jump in of products.

Maybe due to the client side sort, publication, Blaze rendering, CollectionsFS or  ¯\_(ツ)_/¯

aaronjudd:
👍

PR #1012, opened 2016-05-06
When the overridden layout is not found, fall back to the default

zenweasel:
When overriding the default layout, we need to fallback to the default whenever a custom layout/workflow combination is not found.

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests


aaronjudd:
👍

mikemurray:
👍

PR #1013, opened 2016-05-08
Shipping translation and shipping methods modification

newsiberian:
- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

Hello. I added translation for almost all phrases. Also, all shipping method now throws `Error` instead of return `false`. And I refactored `updateShippingMethods` method to allow it to accept `methodId` instead of method object. I did this because this is not beauty to send back all `object` to server then it needs only `object._id`.


packages/reaction-shipping/server/methods.js
```diff
@@ -5,37 +5,43 @@ Meteor.methods({
...
-    check(currentDoc, Object);
-    check(updateDoc, Object);
-    if (!Roles.userIsInRole(Meteor.userId(), ['admin', 'shipping'])) {
-      return false;
+  updateShippingMethods: function (providerId, methodId, updateMethod) {
+    check(providerId, String);
+    check(methodId, String);
+    check(updateMethod, Object);
+    if (!ReactionCore.hasPermission("shipping")) {
+      throw new Meteor.Error(403, "Access Denied");
```
zenweasel: Do all admins automatically have the "shipping" permissions?

packages/reaction-i18n/private/data/i18n/en.json
```diff
@@ -488,6 +488,60 @@
...
       },
+      "shipping": {
+        "addShippingProvider": "Add shipping provider",
+        "editShippingProvider": "Edit shipping provider",
+        "addShippingMethod": "Add shipping method",
+        "editShippingMethod": "Edit shipping method",
+        "deleteShippingMethod": "Delete shipping method",
+        "noSettingsForThisView": "No settings for this view",
+        "noShippingMethods": "No shipping methods are configured.",
+        "removeShippingMethod": "Are you sure you want to delete {{method}}?",
```
zenweasel: should me maybe call this "removeShippingMethodConfirm" to make it more clear?

packages/reaction-i18n/private/data/i18n/en.json
```diff
@@ -488,6 +488,60 @@
...
+        "removeShippingMethodTitle": "Remove Shipping Method",
+        "shippingMethodDeleted": "Shipping method deleted.",
```
zenweasel: These shouldn't have a period since they are almost never complete sentences.

packages/reaction-shipping/server/methods.js
```diff
@@ -44,9 +50,10 @@ Meteor.methods({
   removeShippingMethod: function (providerId, removeDoc) {
     check(providerId, String);
     check(removeDoc, Object);
-    if (!Roles.userIsInRole(Meteor.userId(), ['admin', 'shipping'])) {
-      return false;
+    if (!ReactionCore.hasPermission("shipping")) {
+      throw new Meteor.Error(403, "Access Denied");
     }
```
zenweasel: What does the user see in the UI now? Are they getting a "friendly" error?

packages/reaction-shipping/server/methods.js
newsiberian: Hello, is it so necessary to have an admin access to be allowed to change shipping data? This is your question in opposite direction :smile:

zenweasel: Actually that's not my question in reverse. What I am asking is will admins automatically have the right to edit shipping (because they should)?

packages/reaction-shipping/server/methods.js
newsiberian: Normally, customers can't get this error, because this errors happens only if someone tries to run methods without admin access. Through console ofc. We could display to him "Are you hacker or something...?" :)

zenweasel:
Approved

packages/reaction-shipping/server/methods.js
aaronjudd: owners always have permissions to all, but admin don't specifically (although they can assign themselves)

packages/reaction-shipping/server/methods.js
```diff
@@ -74,9 +81,10 @@ Meteor.methods({
   updateShippingProvider: function (updateDoc, currentDoc) {
     check(updateDoc, Object);
     check(currentDoc, String);
-    if (!Roles.userIsInRole(Meteor.userId(), ['admin', 'shipping'])) {
```
aaronjudd: These are all legitimate replacements. There may have been a reason we're using 'userInRole' here at one time, but if this is working then this is our normal way of doing these validations.  The logic in return false instead of "Access Denied" is more just security by obscurity - the less info we give for your failure the better was probably the thinking.  Again, these updates are acceptable, but we also need to make sure the client can handle these responses as well.

aaronjudd:
👍  Good updates. Let's merge.

aaronjudd:
@newsiberian don't forget those jsdoc updates ;-)

PR #1015, opened 2016-05-09
Fix issue with default payment method not auto-expanding

### LINKED ISSUE #1014: Payment method accordion not expanded by default when default payment method is set
### Expected behavior
When the default payment method is set, the accordion should show that method expanded.

### Actual Behavior
Accordion does not expand

1. Set the default payment method to Paypal
2. Add to cart and checkout till you get to the payment screen
3. Observe that Paypal accordion is not expanded

zenweasel:
Closes #1014 


aaronjudd:
Seems valid 👍  (although as discussed, wondering why the values, appDetails aren't just coming from the `reactionApps` method.  it should be able to handle the icon fallbacks..

mikemurray:
👍   Agreed, reactionApps should be doing the proper fallbacks for those values

PR #1016, opened 2016-05-09
Remove remnant from older version..

prinzdezibel:
- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

### DIFF
--- a/packages/reaction-core/server/register.js
+++ b/packages/reaction-core/server/register.js
@@ -26,9 +26,6 @@ ReactionCore.registerPackage({
     openexchangerates: {
       appId: "",
       refreshPeriod: "every 1 hour"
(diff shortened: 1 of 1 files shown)


aaronjudd:
Not accepting this, it's a valid setting and is still in use.

prinzdezibel:
Don't understand how this can be in use. It's not in ReactionCore.Schemas.CorePackageConfig schema and also I can't find references to it in code.

aaronjudd:
actually, you are correct , its `defaultPaymentMethod: shop.defaultPaymentMethod`

mikemurray:
👍 

This is ok to remove as it has defaults in code if not set.

Issue #1017, opened 2016-05-11
Populating Shops.domains via fixture is not possible


prinzdezibel:
For bugs, please fill out the three following sections completely

### Expected behavior

The usecase is to create a document of the db.Shops collection via data fixture. Additionally the shop is intended to run with a ROOT_URL different than localhost.

e.g.

 `ROOT_URL=http://tsp.dyn.projekt-und-partner.de ./reaction --port 8000`

Executing this command should update the shop document in database and run the shop with the domain given in ROOT_URL. The shop document does contain:

`    "domains" : [
        "tsp.dyn.projekt-und-partner.de"
    ],
`

### Actual Behavior

On startup, I get this error over and over:

`INFO Reaction: No shopId, waiting one second...`

This is because the default Shop document coming from loadPackages (and adhering to the Shop schema) does prepopulate the domain with `localhost`. This problem is somewhat related to https://github.com/reactioncommerce/reaction/issues/548.

### Steps to Reproduce the Behavior

Even without fixture it is reproducable. Get a fresh copy of RC and run:

 `ROOT_URL=http://tsp.dyn.projekt-und-partner.de ./reaction --port 8000`

I would expect to be able to load a shop data fixture either in onCoreInit or afterCoreInit. Neither will work.

spencern:
We experienced this issue as well and had to adjust the default value here:
https://github.com/reactioncommerce/reaction/blob/development/packages/reaction-schemas/common/schemas/shops.js#L177

prinzdezibel:
@spencern: I could resist to do that for now. But if the problem hurts too much, I'll do that as well 👍

zenweasel:
I believe that SimpleSchemas and `defaultValue` are somehow at the heart of this issue. It's only elements with a default that seem to have an issue. Correct me if I am wrong.

prinzdezibel:
Yes, It has to do with the defaultValues of the schema and how that translates to database upserts.

This problem is addressed with the filed PR.

Issue #548, opened 2015-12-06
shop/currency reset to default on every Meteor rerun


newsiberian:
Hello,
@tdecaluwe, this happens inside `ReactionImport.flush()` I suppose.

tdecaluwe:
The `ReactionImport.flush()` method just sends all the queries to mongodb. The query is still constructed in the `ReactionImport.shop()` method.

The problem is that for regular imports we want to overwrite data (concerning the same object) already in the database. To avoid this, the `ReactionImport.fixture()` method can be used which tells `ReactionImport` to use the `$setOnInsert` modifier when talking to the database.

newsiberian:
I have to re-open this issue, because shop currency still resetting after shop rerun. Even with turned off `reaction-sample-data` package.

zenweasel:
I am seeing this problem with shipping methods. Every time meteor restarts the shipping methods are reimported and assigned new _ids, which makes debugging issues with shipping methods nay impossible. Also a super bummer if you have changed any of the shipping values while developing.

Also this doesn't trigger a client reload so it allows the client and server to go out of sync.

newsiberian:
Confirmed. After I update `shipping.provider.enabled` to `false` (for testing purpose). Its revert back to `true` after reload.

prinzdezibel:
And also affects defaultPaymentMethod

prinzdezibel:
Has to do with the shop layout import in combination with the Shops schema. See screen:
![screenshot_09_05_16_22_13](https://cloud.githubusercontent.com/assets/1733229/15126782/fe0e346a-1633-11e6-8f06-b3632c34e3ee.jpg)

prinzdezibel:
A solution to the specific problem of this ticket is to omit the clean step. 

Comment out the following line in reaction-core/server/import.js :

>    //collection.simpleSchema(importObject).clean(importObject);

does fix the problem. But I don't know if this has any other unwanted side effects.

tdecaluwe:
I think omitting the clean step does have unwanted side effects since we do no longer ensure compliance with the schema. Also, we want the import to act like regular queries executed directly on the respective collections. I think we should be looking at how the cleaning mechanism interacts with the mongo `$setOnInsert` operator.

In the following line in `load.js` the `fixture()` method makes sure (or should make sure) the `$setOnInsert` operator is used:
```javascript
ReactionImport.fixture().process(..., ["name"], ReactionImport.shop);
```

[...]

PR #1018, opened 2016-05-13
FIX for #1017 and #548

prinzdezibel:
The idea is that all default values from schema are only relevant if the document is newly inserted. Therefore it is ensured that defaultValues from schemas are passed to mongo with `$setOnInsert` qualifier. 

Nevertheless we ensure full compliance with the schema. The difference between `Fixtures` and `Imports` is somewhat irrelevant with this proposal. All document updates are cumulative.

### DIFF
--- a/packages/reaction-core/server/import.js
+++ b/packages/reaction-core/server/import.js
@@ -289,7 +289,7 @@ ReactionImport.layout = function (layout, shopId) {
     _id: shopId
   };
   return this.object(ReactionCore.Collections.Shops, key, {
-    "_id:": shopId,
+    "_id": shopId,
     "layout": layout
   });
 };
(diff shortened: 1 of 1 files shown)


packages/reaction-core/server/import.js
```diff
@@ -332,18 +405,29 @@ ReactionImport.object = function (collection, key, object) {
   if (!collection.findOne(key) && !object._id) key._id = Random.id();
   // hooks for additional import manipulation.
   const importObject = ReactionCore.Hooks.Events.run(`onImport${this._name(collection)}`, object);
```
aaronjudd: nice! appreciate seeing this used, this was the intended use case as well.

zenweasel:
This is a great fix. But I wonder if it makes sense to have an option somewhere to make the write canonical, in that it would intentionally overwrite values that are already there. In cases (like testing) where you want to overwrite and have it revert to default. We don't need to hold up this PR for that, but just a thought while we are touching this code.

aaronjudd:
If you overwrite a value in the same position in the db, for example, add a value to the defaultVisitorRole manually, then add a different role to Shops.json entry, the different but new value in Shops.json will overwrite the value in the same position.  This is not entirely unexpected behaviour, but it's something we should clearly note if we accept that behaviour. I'd expect that arrays where items positions change could also be problematic.

packages/reaction-core/server/import.js
```diff
@@ -315,6 +315,79 @@ ReactionImport.tag = function (key, tag) {
...
+    })
+  };
+
+  for (let key of findRightOnlyProperties()){
+    if (typeof(rightSet[key]) === "object") {
+      // subobject or array
+      if (leftSet.hasOwnProperty(key) && (typeof(leftSet[key]) !== "object" ||
+           Array.isArray(leftSet[key])!== Array.isArray(rightSet[key]))) {
+        // This is not expected!
+        throw new Error(
```
aaronjudd: We should probably add a bunyan log here as well. just for best practices...  could also use Meteor.error

packages/reaction-core/server/import.js
```diff
@@ -289,7 +289,7 @@ ReactionImport.layout = function (layout, shopId) {
     _id: shopId
   };
   return this.object(ReactionCore.Collections.Shops, key, {
-    "_id:": shopId,
+    "_id": shopId,
     "layout": layout
```
aaronjudd: these don't actually need to be quoted... ;-)

PR #1019, opened 2016-05-16
Fix for "Can't set timers inside simulations"

### LINKED ISSUE #784: Error when checking out via Paypal Express
Not sure if I should enter this here or against the PayPal package so...

After completing Paypal Express checkout I am getting this error in the console and the confirmation page never renders. The order does seem to complete and reloading the page forces it to render.

```Exception from Tracker recompute function:
debug.js:41 Error: Can't set timers inside simulations```

Steps to reproduce:

1. Place an order using Paypal Express.
2. Once the modal closes the page clear and the above error message appears.
3. Observe that confirmation page never loads.

zenweasel:
Closes #784

Needs to be paired with https://github.com/reactioncommerce/reaction-paypal/pull/34 in order top not break Paypal processing

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

### DIFF
--- a/packages/reaction-checkout/common/methods/cart.js
+++ b/packages/reaction-checkout/common/methods/cart.js
@@ -61,11 +61,7 @@ if (Meteor.isClient) {
           Meteor.call("workflow/pushCartWorkflow", "coreCartWorkflow",
             "paymentSubmitted");
           // Client Stub Actions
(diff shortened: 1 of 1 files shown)


zenweasel:
This is the wrong approach

Issue #996, opened 2016-05-03
Items in order not being merged


zenweasel:
### Expected behavior
If an item is added to the cart more than once, it should be merged into a single line item when the order is completed.

### Actual Behavior
Identical line items are not being merged.
<img width="1177" alt="cart_completed" src="https://cloud.githubusercontent.com/assets/72819/14975560/81da83f6-1135-11e6-895a-858eeba3bbfc.png">

### Steps to Reproduce the Behavior
1. Add an item to the cart
2. Add the same item to the cart again
3. Complete the order
4. Observe that there are two entries in the database.

newsiberian:
Hello, I've spend some time by investigating this issue and it is work exactly how it designed. Look at [this](https://github.com/reactioncommerce/reaction/blob/development/packages/reaction-core/server/methods/cart.js#L561) line. This changes was added by this commit: 021dfd4e4894d58b5a587402fcdbf0c309aacc13

zenweasel:
Then it's designed incorrectly. No shopping cart should behave like that. The scenario of having 10 of the item on one line and then 3 of the same item on another will lead to confusion which will lead to chargebacks.

aaronjudd:
@zenweasel it's the **order**, not the **cart** behaving this way. the cart seems fine.  - but yes order items should be grouped, unless they are from diffferent shops

zenweasel:
My mistake, you are correct. Changing the issue title.

prinzdezibel:
Not sure, but it seems this happened with this commit:

https://github.com/reactioncommerce/reaction/commit/1eca97bf5245e83ac3807a35bc734ab6743b852e

PR #1023, opened 2016-05-18
Dashboard totals 992

mikemurray:
WIP
Resolves #922 #996 

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

---

#922
Fixes issue with currency in order fulfillment panel being localized to the user. Currency displayed in the order dashboard are now based on the shop.

Added `scale` to currencies in Shops.json and schema to better handle currencies that do not use decimal places, such as the Japanese Yen.

Converted `numericInput` component to React and update functionality to use `accounting.js` natively.

Fixes display issue with `{{formatPrice}}` template helper incorrectly displaying currencies with 0 decimal places, with two decimal places.

Fixed a bug that caused an error to display in the chrome js console while in the order dashboard.

---

#996 
Combine same products in order list.
screen shot 2016-05-18 at 3 08 10 pm

### DIFF
--- a/packages/reaction-i18n/client/helpers/helpers.js
+++ b/packages/reaction-i18n/client/helpers/helpers.js
@@ -128,7 +128,21 @@ function _formatPrice(price, originalPrice, actualPrice, currentPrice, currency,
   if (typeof currency !== "object") {
     return false;
   }
+
+  let adjustedPrice = actualPrice;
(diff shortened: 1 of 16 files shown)


aaronjudd:
👍  Looks good, no issues with this now, will merge.

Issue #1022, opened 2016-05-17
Allow users to turn off "multishop" URL's


zenweasel:
As a user who only has one shop, I want to have an option to turn off requiring my shop name to be in the URL

tdecaluwe:
Or, equivalently, add the ability to add a default shop?

aaronjudd:
or if there is just one shop in the db... don't use the shop prefix.

tdecaluwe:
@aaronjudd I agree, don't complicate things that don't have to be complicated.

Issue #1006, opened 2016-05-04
Missing Currency & Localization


Roshdy:
I'm amazed by the effort spent and the work done on this project. 
(All the below comments are from the Dashboard control,..i didn't touch the code yet)
However, i was trying to test the application by making simple customizations, I am in Kuwait now, and i can't find the KWD (Kuwait Dinar) on the currencies list. Also when changing the language to "Arabic" nothing is actually localized. the layout only goes RTL, and it's a bit messy (try to add a product with arabic language).

On another end, i tried to make the application name in 'Arabic' (only for viewing), instead it changed it in the URL (which messed up the routing). I suggest that you make it possible for a display name, and an App name (separate), to be able to fulfill such customization.

aaronjudd:
@roshdy thanks for the feedback. Leaving this open as an issue, we're aware the RTL is a little messy at the moment, it's on short list to give it all another review.  The url issue is documented elsewhere, and we'll fix as a critical bug in 0.13.  Leaving this open to track this stuff.

Roshdy:
Thanks @aaronjudd. Looking forward to this version 😃

mikemurray:
RTL issues for the product detail and some navigation drop-downs have been resolved in PR #1095.

aaronjudd:
I believe the url routing is now resolved as well.  updated in 0.14.

mikemurray:
The Arabic language not loading is due to it being added to the `translations` collection with a null `shopId`. When we try to select the language in the dropdown, it fails to return because the publication is fetching translations by `shopId` and `i18n`(name).

mikemurray:
Related to https://github.com/reactioncommerce/reaction/issues/960 missing shopId

PR #1024, opened 2016-05-19
add slug prefix and absolute routes

### LINKED ISSUE #955: Changing shop name breaks urls

### Expected behavior
Changing shop name in admin section should change urls in a manner that does not break urls.

### Actual Behavior
When changing the site name in the admin panel to a name containing spaces, urls do not work.

The workaround I've been using is to change the name of the shop in the admin UI to a url friendly one("my_shop_name") and then manually enter the correct display name ("My Shop Name") in the brand.html file.

aaronjudd:
- issue #1022
-allow “absolute” url definition, registry routes without “/“ will be
treated as absolute
- resolves #955 with added getSlug
- related to #1006

### DIFF
--- a/packages/reaction-router/common/init.js
+++ b/packages/reaction-router/common/init.js
@@ -88,7 +88,7 @@ const getRegistryRouteName = (packageName, registryItem) => {
  */
 ReactionRouter.initPackageRoutes = () => {
   const pkgs = ReactionCore.Collections.Packages.find().fetch();
-  const prefix = ReactionCore.getShopName().toLowerCase(); // todo add shopId
+  const prefix = getSlug(ReactionCore.getShopName().toLowerCase()); // todo add shopId
 
   // initialize index
   // define default routing groups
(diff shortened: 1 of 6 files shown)


packages/reaction-router/common/init.js
```diff
@@ -88,7 +88,7 @@ const getRegistryRouteName = (packageName, registryItem) => {
  */
 ReactionRouter.initPackageRoutes = () => {
   const pkgs = ReactionCore.Collections.Packages.find().fetch();
-  const prefix = ReactionCore.getShopName().toLowerCase(); // todo add shopId
+  const prefix = getSlug(ReactionCore.getShopName().toLowerCase()); // todo add shopId
```
zenweasel: Should we create a global `getShopSlug`? Asking for a friend.

aaronjudd: well at this point, getSlug is a global function, and currently, this is the only place that the shop name would be useful slugified - I think.

zenweasel: Except you know, if somebody had to create a `return_url` for something. Just speculating.

zenweasel:
Approved.

packages/reaction-router/common/init.js
aaronjudd: maybe instead we need our own absoluteUrl method..

PR #1025, opened 2016-05-19
Renamed i18n locale code of Simplified Chinese from "cn" to "zh"

andretw:
- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

Hi all, I noticed the i18n file (cn.json) of Simplified Chinese and the locale code do not follow the naming convention by http://i18next.com/translate/. (RFC 4646)

> 1. en-US language + region
> 2. en language only

i.e. the file's name and the value of the key "i18n" should be "zh" or "zh-CN".

### DIFF
--- a/packages/reaction-i18n/package.js
+++ b/packages/reaction-i18n/package.js
@@ -64,7 +64,6 @@ Package.onUse(function (api) {
   // i18n translations
   api.addAssets("private/data/i18n/ar.json", "server");
   api.addAssets("private/data/i18n/bg.json", "server");
-  api.addAssets("private/data/i18n/cn.json", "server");
   api.addAssets("private/data/i18n/cs.json", "server");
   api.addAssets("private/data/i18n/de.json", "server");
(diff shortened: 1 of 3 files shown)


aaronjudd:
@andretw thanks, I'm not exactly sure where the original locale code came from ;-) appreciate the fix. 👍

zenweasel:
Approved

reidlai:
why not rename zh to zh-CN?  I found all characters are in Simplified Chinese while HK and Taiwan are using Traditional Chinese.  If you can rename zh to zh-CN, I can help to create zh-TW

aaronjudd:
@reidlai I think that's a reasonable proposal, in general other languages/regions also have dialects, and we've just been supporting the "root" version -> but only because we never got around this implementation.  We should create a new issue to track, rather than this old PR and we can move forward from there.

reidlai:
@aaronjudd will do.  Thanks.

PR #1030, opened 2016-05-23
New `shop/createTag` method

newsiberian:
- [x] Has tests.
- [x] Has docs.
- [x] Passes all tests

Hello, I needed this method for blog package I'm working on and I suppose it could be used in future by reaction itself. I think this is a good way to pick out this functionality.

Also, I added new file for `shop` hooks. This is how I manage logs in my packages. I believe this is a good pattern to manage logs separately from method. It looks cleaner. It would be cool to implement such pattern for all reaction methods in future. What do you think?

I'm not using `newTag.shopId = ReactionCore.getShopId();` because `shopId` is `autovalue`.
Also added two tests for new method. All tests passed well for me.

### DIFF
--- a/packages/reaction-core/server/methods/shop.js
+++ b/packages/reaction-core/server/methods/shop.js
@@ -363,6 +363,35 @@ Meteor.methods({
   },
 
   /**
+   * shop/createTag
+   * @summary creates new tag
+   * @param {String} tagName - new tag name
+   * @param {Boolean} isTopLevel - if true -- new tag will be created on top of
+   * tags tree
+   * @return {String} with created tag _id
+   */
+  "shop/createTag": function (tagName, isTopLevel) {
+    check(tagName, String);
+    check(isTopLevel, Boolean);
+
+    // must have 'core' permissions
+    if (!ReactionCore.hasPermission("core")) {
+      throw new Meteor.Error(403, "Access Denied");
+    }
+
+    // we assume what tag is not exist.
+    const tag = {
+      name: tagName,
+      slug: getSlug(tagName),
+      isTopLevel: isTopLevel,
+      updatedAt: new Date(),
+      createdAt: new Date()
+    };
+
+    return ReactionCore.Collections.Tags.insert(tag);
+  },
+
+  /**
    * shop/updateHeaderTags
    * @summary method to insert or update tag with hierarchy
    * @param {String} tagName will insert, tagName + tagId will update existing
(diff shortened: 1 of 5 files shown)


packages/reaction-core/server/methods/shop.js
```diff
@@ -363,6 +363,35 @@ Meteor.methods({
...
+    check(tagName, String);
+    check(isTopLevel, Boolean);
+
+    // must have 'core' permissions
+    if (!ReactionCore.hasPermission("core")) {
+      throw new Meteor.Error(403, "Access Denied");
+    }
+
+    // we assume what tag is not exist.
+    const tag = {
```
zenweasel: I assume you meant "we assume this tag does not exist" and why do we assume that? Wouldn't a simple check for an existing tag be prudent?

packages/reaction-core/server/methods/shop.js
```diff
@@ -424,12 +453,8 @@ Meteor.methods({
       });
     }
     // create newTags
-    newTag.isTopLevel = !currentTagId;
-    newTag.shopId = ReactionCore.getShopId();
```
zenweasel: Can you help me understand why we don't want to set the shopId of a new tag?

aaronjudd:
Doesn't this provide this functionality? 
```
   * shop/updateHeaderTags
   * @summary method to insert or update tag with hierarchy
   * @param {String} tagName will insert, tagName + tagId will update existing
   * @param {String} tagId - tagId to update
   * @param {String} currentTagId - currentTagId will update related/hierarchy
   * @return {Boolean} return true/false after insert
   */
  "shop/updateHeaderTags": function (tagName, tagId, currentTagId) 
```
https://github.com/reactioncommerce/reaction/blob/175d5015612395fdf18be51c8f5a20776413cf63/packages/reaction-core/server/methods/shop.js#L366

aaronjudd:
👎  I think we should reject this, perhaps we could/should reconsider renaming the existing method though, as it's not very clear.

packages/reaction-core/server/methods/shop.js
newsiberian: Hello, because this happens automatically [here](https://github.com/reactioncommerce/reaction/blob/development/packages/reaction-schemas/common/schemas/tags.js#L32)

packages/reaction-core/server/methods/shop.js
newsiberian: I forgot to remove this comment from previous redaction. But you are right about simple check for tag existence and this check happens before this method calls. In other words, when we call it, we already know that there is no such tag.

newsiberian:
> Doesn't this provide this functionality?

This method doesn't override `shop/updateHeaderTags`. It extracts the part of functionality from it to separate method.

newsiberian:
Hello, lets think about this PR one more time. Here is a situation: I'm making a blog (server functionality) package and I need an ability to add tags to posts. I can work with my Posts collection by adding existing tags to posts, but if tag does not exist, I need to create new tag, right? What API reaction provides to me?, because I don't want to write method for adding document to reaction internal collection (Tags).
So, of course, If we are looking from current functionality point of view, then this new method is duplicate, but from another point of view, this is necessary part of reaction API for package creators.
Any thoughts?

aaronjudd:
@newsiberian any chance you already refactored for the new development/modules branch? if so, can you close this PR and make a new one and I'll merge it?  Otherwise, I can copy into the development and close this.

newsiberian:
@aaronjudd, hello, I'll check this PR tomorrow, ok.

PR #1039, opened 2016-05-24
suppress reset cli output

aaronjudd:
suppress “No such file or directory” when removing bower and
node_modules directories when using `reaction reset`


zenweasel:
Approved.

PR #1058, opened 2016-06-01
Use current versions of AuthNet and Braintree packages

### LINKED ISSUE #970: In fresh clone, Braintree is enabled by default
### Expected behavior
By default only the "Example Payment Provider" should be enabled

### Actual Behavior
Example and Braintree are enabled, but of course Braintree doesn't work because it's not configured (which it should not display without an API key present anyway)

Run a fresh clone

I would just fix this but it wasn't immediately obvious where this value is getting stored, so this is just a reminder.

zenweasel:
Closes #970 (after re-publishing these packages)

Zenweasel, closing trivial bugs that I myself opened since 2015


zenweasel:
Will reopen on modules branch (when ready)

PR #1060, opened 2016-06-03
Module analytics

aaronjudd:
- updated permissions
- todo: enable Mixpanel (no script added)
- GA and segment updated
- basic page tracking restored
- event tracking updated

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -56,7 +56,7 @@
      */
     "comma-dangle": [2, "never"], // http://eslint.org/docs/rules/comma-dangle
     "no-cond-assign": [2, "always"], // http://eslint.org/docs/rules/no-cond-assign
-    "no-console": 1, // http://eslint.org/docs/rules/no-console
+    "no-console": [2, { allow: ["warn", "error"] }], // http://eslint.org/docs/rules/no-console
     "no-debugger": 1, // http://eslint.org/docs/rules/no-debugger
     "no-alert": 1, // http://eslint.org/docs/rules/no-alert
     "no-constant-condition": 1, // http://eslint.org/docs/rules/no-constant-condition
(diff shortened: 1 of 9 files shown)


jshimko:
👍

jshimko:
This is ready to be merged, right?

aaronjudd:
👍  good to merge, resolves the main issue (permissions writing to AnalyticsEvents)

We should merge, two items pending that we can review again later

1.  Just added GA script to head, should review add with script.src and logic to not load if Ga isn't enabled.
2. Add or remove the mixpanel methods.  (Segment covers this, as well).

jshimko:
GA and Mixpanel are both in the free tier on Segment.  Do you think we should possibly consider recommending that path to users so it's a single tracking API in the Reaction code and Segment can transform all of the calls on their end?  

At the very least, maybe it's worth running Mixpanel through Segment.  It seems reasonable to assume that if you're using a more sophisticated tool like Mixpanel, you'd likely be open to using Segment (since it'd be free and can also tie in GA in the same place).

aaronjudd:
Segment cost money. Google is free (and not being passed through Segment). Seems like a good reason to have the both of them as defaults. I Agree with you on Mixpanel - because on the other hand, it costs.. if you are going to use MixPanel that it makes sense for you to also use Segment.

jshimko:
Segment's free "developer" plan will let you use GA and Mixpanel for free (among others).  That was the only reason I mentioned it.  It's probably not worth it for GA though.  I could see a lot of people being frustrated with having to deal with another product just to use GA - even if it's free.

spencern:
I've got a branch I've been working on that adds Enhanced Ecommerce to GA tracking and makes GA and mixpanel work independently of segment while still sending data to all. Currently in our private repo, but plan to open source it/pr this repo it out once we transition to modules.

PR #1062, opened 2016-06-04
Make email notification multilingual and refactoring of

prinzdezibel:
package reaction-email-template

- email notifications are i18n enabled
- Loose coupling through usage of ReactionCore.Hooks.Events
- Email notifications can be disabled in Dashboard (even if I don't know
  why somebody should do this, but hey..)
- Compile templates during init process. That way it is possible to
  properly override templates (in alignment to aldeed's
  Template.replace)
- Prevent redundant HTML and STYLES by having only one template for
  all languages.
- Also prevent having templates for different languanges in Mongo-DB,
  which IMHO also brings a lot of redudancy to the game. Having all
  templates as files is also in alignment with normal HTML-Templates.
- Copying templates to custom package and override/extending base templates is
  possible
- Shop is usable even if package reaction-email-templates is removed
- Email generation is in one central place and not cluttered among
  serveral packages

===> Because of the above enhancements I propose a package name change:

   ```reaction-email-templates```
        should be renamed into
   ```reation-email-notifications```


prinzdezibel:
Hi. This is a kind of a WIP. I've translated the "order.new" email for demonstration purposes. If I find more time, I can l10n the other templates.

prinzdezibel:
forgot to mention: I've connected the email templates to actual workflow step and/or UI elements because I think there might be a plan to trigger notifications automatically when certain workflow steps are entered (or maybe via flow router triggers when a certain route is entered, etc, etc.)

But this is optionally of course and the referencing of templates in register.js is not technically necessary.

packages/reaction-core/server/main.js
```diff
@@ -23,6 +23,17 @@ _.extend(ReactionCore, {
...
     // hook after init finished
     ReactionCore.Hooks.Events.run("afterCoreInit", this);
+
+    // Make sure this event is registered last.
+    ReactionCore.Hooks.Events.add("orders/sendNotification", (result) => {
+      if (!result) {
+        //  If event was successfully captured the result will be falsy.
+        //  No-op
+      } else {
+         ReactionCore.Log.warn("Package email-notifications seems not being " +
```
zenweasel: This should probably be a `fatal` and the correct English would "seems to not be installed. Try to fix it with....". Is the package not being installed the main reason this doesn't work? (being that it will be installed by default)

prinzdezibel: you're correct. The package is installed by default. This is just intended as safety net, which is not necessary at all. Feel free to remove it if you don't feel it adds value.

prinzdezibel:
Another thing: At the moment the currently selected language is stored in Accounts document under key profile to have it available on server side.

Another option would be to pass the desired language in the event via parameter. Then we won't need the schema change to Profile. Not sure about what suits us best, though...

packages/reaction-email-templates/README.md
```diff
@@ -10,19 +10,28 @@ Provides functionality to load local default email templates, or optionally load
...
-    html: SSR.render("<template>", {
-      homepage: Meteor.absoluteUrl(),
-      shop: shop,
-      order: order,
-      shipment: shipment
-    })
+    subject: `You have been invited to join ${shop.name}`,
+    html: html
   });
+} catch (_error) {
```
aaronjudd: this `_error` should be just `error`.  I think this was a remnant from when we converted from CoffeeScript.

packages/reaction-email-templates/server/hooks.js
```diff
@@ -0,0 +1,297 @@
...
+    // also add english as second fallback..
+    fallbackLng = [fallbackLng, "en"];
+  }
+
+  if (!Template[template]) {
+      ReactionCore.Log.fatal(`Template '${ template }' does not exist. ` +
+        `Did you register it in the approriate workflow step (register.js)?`);
+  }
+  Template[template].helpers({
+    '_': function(key){
```
aaronjudd: rather that we didn't have helpers/methods named `_`.   Easily confused with underscore library, and also isn't descriptive. Hard to search for as well. This is one of the things I did not like about the `tap-i18n` package. This isn't doing the same thing the current i18n helper does?

packages/reaction-email-templates/templates.js
```diff
@@ -1,40 +0,0 @@
...
-    lang = shopLocale.locale.languages;
-  }
-
-  // using layout where in the future a more comprehensive rule based
-  // filter of the email templates can be implemented.
-  const tpl = ReactionCore.Collections.Packages.findOne({
-    "layout.template": template
-  });
-
-  if (tpl) {
```
aaronjudd: This functionality was here to allow for language specific templates, that could be edited in a UI (that is yet to be built).  By removing this we'd be moving to only being able to edit the translations.  This may be ok just for translations, but wouldn't allow for the easy customization of the layout of the templates.

packages/reaction-email-templates/server/hooks.js
prinzdezibel: _  for translations is quite common in python land. but you right: could be
confusing. please feel free to change anything you don't like or tell me
what to do

prinzdezibel: i haven't pulled in tap-i18n, just used their naming convention (which I'm
used to from python..)

packages/reaction-orders/server/register.js
```diff
@@ -59,34 +59,41 @@ ReactionCore.registerPackage({
     label: "Order Processing",
     status: "created",
     workflow: "coreOrderWorkflow",
-    audience: ["dashboard/orders"]
+    audience: ["dashboard/orders"],
+    emailTemplates: ["orders/new"]
```
aaronjudd: Previously the logic just used `let tpl = `orders/${order.workflow.status}`;`, and didn't need a `emailTemplates`... do you think it's better to have to define a template for each, or to just assume that every workflow state that supports a template, does has the `emailTemplate` assignment?

aaronjudd:
Overall, good PR. The use of the event hooks is nice, and clearly this adds required functionality.  I had hoped to have this handled by using translated emails stored in the Templates collection- previously you just needed to add an entry to the Template collection to override the email that the default package supplied,(and language specific) I'd like to keep that functionality, but include the Translation functionality as well - particularly because we didn't translate the mail fields before.

packages/reaction-orders/server/register.js
prinzdezibel: Sent from mobile device. Please excuse brevity and typos.
On Jun 6, 2016 11:50 PM, "Aaron Judd" <notifications@github.com> wrote:
>
> In packages/reaction-orders/server/register.js:
>
> > @@ -59,34 +59,41 @@ ReactionCore.registerPackage({
> >      label: "Order Processing",
> >      status: "created",
> >      workflow: "coreOrderWorkflow",
> > -    audience: ["dashboard/orders"]
> > +    audience: ["dashboard/orders"],
> > +    emailTemplates: ["orders/new"]
>
> Previously the logic just used let tpl =orders/${order.workflow.status};,
and didn't need a emailTemplates... do you think it's better to have to
define a template for each, or to just assume that every workflow state
that supports a template, does has the emailTemplate assignment?

I'm totally open to that. I'm not even sur if we want to have the templates
coupled to workflow steps/UI containers. it could make sense if we plan to
automatically trigger notification when a certain state is entered.

packages/reaction-email-templates/templates.js
prinzdezibel: Ok, was not sure if the code is really working and/or in use. Looked more like a not-finished-thing to me. 

Let me try to re-integrate it.

prinzdezibel:
Reasining: My main goal with this PR is to be able to override the templates in a file-based fashion. I need to override the email templates for easywine and also be able to extend the helpers, because I'm going to display more information about the order in the orders/new template. I think this could be a quite common use case.

packages/reaction-email-templates/templates.js
```diff
@@ -1,40 +0,0 @@
...
-  const shopLocale = Meteor.call("shop/getLocale");
-
-  if (shopLocale && shopLocale.locale && shopLocale.locale.languages) {
-    lang = shopLocale.locale.languages;
-  }
-
-  // using layout where in the future a more comprehensive rule based
-  // filter of the email templates can be implemented.
-  const tpl = ReactionCore.Collections.Packages.findOne({
-    "layout.template": template
```
prinzdezibel: I don't get the intention of that query. Also it seems that "layout.template" query does not represent a valid document schema.

prinzdezibel:
@aaronjudd: If you have time please check again and let me know if you want me to fix/tidy-up/improve something else.

aaronjudd:
@prinzdezibel we need to refactor this for the development/modules branch, and I wanted to see if it was also still a WIP.  Otherwise, I would be good to merge this.  If you don't have time to update for development let me know, we can also handle that.

prinzdezibel:
@aaronjudd: I'm very busy this and the next week. Would it meet your timeline if I'm going to update it after that?

prinzdezibel:
Please see #1136 for a first shot with new module structure.

PR #1064, opened 2016-06-07
Product detail page updates and more

mikemurray:
- Fixed overlapping issues with responsive styles and the login forms.
- First pass at updating CSS and templates for the product detail page.
- This will help with modernizing the look of the controls on the page
during edit mode, as well as fix issues with RTL support.
- Added product images to child variants.
- Updated variant form for action view.
- Changed layout of variant form so its no longer a table allowing more
controls and info to be displayed.
- Added image upload for product variants in action view.
- Added select rings around variant and child variants.
- Removed stock amount from variants, leaving "sold out" and
"backorder" badges.
- Removed quantity progress bars from variants.
- Child variant form now shows all attached media items.
- Variant list child variant now sorts and displays the highest
priority item in the option select grid.


client/modules/product-variant/templates/products/productDetail/productDetail.js
```diff
@@ -85,6 +94,7 @@ Template.productDetail.helpers({
                     handle: result
                   });
                 }
+                return null;
```
zenweasel: Just a question. Why do we return `null` here instead of `undefined` like we do elsewhere?

client/modules/product-variant/templates/products/productDetail/productDetail.js
```diff
@@ -153,16 +163,20 @@ Template.productDetail.helpers({
...
+
+    return null;
   },
   tags: function () {
     let product = ReactionProduct.selectedProduct();
     if (product) {
       if (product.hashtags) {
         return _.map(product.hashtags, function (id) {
-          return Tags.findOne(id);
+          return Reaction.Collections.Tags.findOne(id);
```
zenweasel: Wasn't `Tags` imported already from Collections so it would just be `Tags.findOne`?

client/modules/ui/components/index.js
```diff
@@ -1,8 +1,3 @@
 // export ButtonGroup from "./buttonGroup/buttonGroup";
-import Icon from "./icon/icon";
-import Seperator from "./separator/separator";
-
-export {
-  Icon,
-  Seperator
-};
+export { default as Icon } from "./icon/icon";
+export { default as Seperator } from "./separator/separator";
```
zenweasel: Did not know you could do that

zenweasel:
Approved

client/modules/product-variant/templates/products/productDetail/productDetail.js
aaronjudd: Correct, updated.

PR #1071, opened 2016-06-08
Transliteration fixes

jshimko:
- adds client side lib for [transliteration](https://www.npmjs.com/package/transliteration) (because Node lib doesn't work in the browser)
- removes client side imports of Node lib in favor of global
- fixes router prefix slug
- updates transliteration to 1.0.7

In order to make the source of the transliteration lib conditional for files that use it on client _and_ server, this also updates the app to `METEOR@1.3.3-rc.4` (which allows conditional ES6 imports)

Like...

```js
if (Meteor.isServer) {
  import { slugify } from "transliteration";
}
// else use the client side global
```


aaronjudd:
Simply importing slugify doesn't handle empty values.
old code:
```
if (Meteor.isClient) {
  getSlug = function (slugString) {
    if (slugString) return TR.slugify(slugString);
  };
} else {
  getSlug = function (slugString) {
    if (slugString) return Transliteration.slugify(slugString);
  };
````

So if you try to set an empty handle, you'll get: Exception in delivering result of invoking `'products/setHandle': TypeError: Cannot read property 'substring' of undefined`

aaronjudd:
Also, I think it's worth wrapping these "Reaction.getSlug".. .

In the future, we may want to revert back to using "speakingURL", or we might want a combo of the two...
speakingUrl doesn't handle Arabic, Chinese, or Thai for example, but is very lightweight in comparison.  We might want to switch transliterators based on language.  Best to leave a wrapper around this.

aaronjudd:
👍  looks good

PR #1080, opened 2016-06-15
replace underscore with lodash

aaronjudd:
- _.findWhere -> _.find
- _.where -> _.filter
- _.contains -> _.includes

Todo: `lodash` exports individual function so we can import just what
we need.


aaronjudd:
Merging iodash.

PR #1091, opened 2016-06-23
Fix race condition that was happening at checkout

spencern:
- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

Fixes issue with `Cart.update` being called asynchronously which would trigger a race condition between the cart update and the functionality in the `cart/submitPayment` method. If the database was not updated fast enough, `copyCartToOrder` would never get called, leaving the customer with a blank white checkout complete screen, leaving the order in it's cart stage, and processing the payment anyway.

### DIFF
--- a/packages/reaction-core/server/methods/cart.js
+++ b/packages/reaction-core/server/methods/cart.js
@@ -972,14 +972,11 @@ Meteor.methods({
       };
     }
 
-    return ReactionCore.Collections.Cart.update(selector, update,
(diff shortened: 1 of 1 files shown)


jshimko:
This has been fixed in the modules restructure branch (along with several other async calls in cart methods in that same file).  Hopefully that solves the issue.

aaronjudd:
Closing, as the patch has been made to development.

Issue #422, opened 2015-07-21
Improved error handling for payment packages


aaronjudd:
Ensure that all the exceptions are handled and displayed to user in friendly manner.

zenweasel:
I think this ticket is mostly covered by the work @kieckhafer is doing, but we can review once all the PR's are in.

kieckhafer:
Yes, I believe what I'm doing will cover this ticket.

PR #1105, opened 2016-06-29
Add stripe luhn check

zenweasel:
This is code migrated from PR#20 on the Stripe repo
https://github.com/reactioncommerce/reaction-stripe/pull/20/

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests


zenweasel:
Making this WIP since when this fails it just returns "Oops! Match failed [400]" rather than a meaningful error to the user.

aaronjudd:
👍  let's accept and add the better messaging in #422

PR #1135, opened 2016-07-06
Add example payment method

zenweasel:
Reconfigure the "example payment method" to the new plug-in layout. Also change all references from "generic" to "example".

- [x] Has tests.
- [ ] Has docs.
- [x] Passes all tests


mikemurray:
👍

PR #1137, opened 2016-07-07
don't slugify productId when used as handle, add random id fallback

### LINKED ISSUE #1127: unable to view newly created product
### Expected behavior

When you add new product, you should be able to view it as admin.

### Actual Behavior

Unable to view product without title,etc. Just an empty page, after navigation away from initially created product.

Add new product, navigate away, click in main grid. Empty page.

jshimko:
Fixes #1127 

Handle was being set as a lowercase product ID when title wasn't set on creation.  Now remains mixed case so a query by ID in the handle field returns a product.

Updates the regex check that tries to discern between a handle and an ID (based on `SimpleSchema.Regex.Id`).


zenweasel:
Approved

mikemurray:
👍

PR #1144, opened 2016-07-10
Add custom plugins and missing imports

zenweasel:
This is a rollup of all changes I made to the main directory while rewriting the Customization Guide. Mostly just missing imports.


imports/plugins/included/product-variant/client/templates/products/products.js
```diff
@@ -3,6 +3,7 @@ import { ReactionProduct } from "/lib/api";
 import { Products, Tags } from "/lib/collections";
 import { Session } from "meteor/session";
 import { Template } from "meteor/templating";
+import { ITEMS_INCREMENT } from "/client/config/defaults";
```
aaronjudd: How will this be overwritten?

aaronjudd:
👍  accepting, but with the caveat that this doesn't allow a customization of ITEM_INCREMENTS.  @mikemurray is creating a new issue to resolve that regression.

PR #1145, opened 2016-07-11
Reaction.getShopId() race condition

### LINKED ISSUE #960: shopID is not available for `createDefaultAdmin` when running RC in separate containers
This bug is currently blocking running the most recent version of RC in LaunchDock so it's probably a P1 bug. I am putting this here just so we have a place to track it.

### Expected behavior
Since import methods are run before `createDefaultAdmin`, the function `getShopId` should return a shopId.

### Actual Behavior
In some configurations of Docker (any configuration where you have separate app and db containers on separate machine, and on my local machine as well) either the shopId is not available for several seconds, or in the case of separate containers, not available at all.

If you remove the current workaround, I am able to replicate this locally by running `docker-compose` against my local version of the app. The problem in LD however would seem to be either different or more serious since `shopId` never becomes available.

jshimko:
Fixes #960 and #1006

`Reaction.getShopId()` was returning `null` for more than half of the times it was being called on initial startup.  This was mostly just an issue with a fresh database because the shop hadn't been inserted yet.  This was the main issue behind the `createDefaulAdminUser` issue in #960 and the translation loading issue in #1006.  (both should be fixed now)

Part of the solution was to reorder several imports that were trying to use `Reaction.getShopId()` before the initial shop was created and the other part of it was because the query behind that method was not very efficient.  It was querying for the entire shop document (which is a fairly large object), but it really only needed the `_id`.  That said, I changed the query to only return the `_id`.  It didn't totally solve the problem, but it should be significantly more performant considering how often `Reaction.getShopId()` gets called.

The issue with the translations loading was because it was using Node's native `fs` utility to read the `i18n` directory and get all available translations files.  By default, Meteor runs all `fs` commands before compiling the app and executing anything.  The problem with _that_ is it was trying to then insert those translations at a point in time where the shop document hadn't been created yet.  That would cause a few of the first translations to have a `null` shopId and _that_ would cause them to not be returned in the `Translations` publication (which relies on that `shopId` to fetch the translations).  

My solution for that was to use `fs` to scan for all of the translation files and then insert them into an `Assets` collection.  Then I read the `Assets` collection at startup and load the translations from there.  This guarantees that Meteor will run it when things are _actually_ ready.  I figure it'll also be useful to store file names, paths, and content of server side asset files in there in the future (since native Node, NPM, fs, etc. is the future of Meteor).  For now, only `i18n` stuff is being imported there.

As far as I can see, all shopId related issues appear to be fixed now and `Reaction.getShopId()` never returns `null` at any point in first startup or any subsequent startup.

### DIFF
--- a/server/startup/i18n.js
+++ b/server/startup/i18n.js
@@ -53,7 +53,7 @@ export default function () {
   /**
    * Hook to setup core i18n imports during Reaction init
    */
-  Hooks.Events.add("afterCoreInit", () => {
+  Hooks.Events.add("onCoreInit", () => {
     loadCoreTranslations();
   });
 }
--- a/server/startup/i18n.js
+++ b/server/startup/i18n.js
@@ -1,4 +1,5 @@
 import fs from "fs";
+import { Assets } from "/lib/collections";
 import { Hooks, Logger, Reaction } from "/server/api";
 
 // taken from here: http://stackoverflow.com/a/32749571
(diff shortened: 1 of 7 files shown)


zenweasel:
Approved. Nice Work!

server/startup/i18n.js
```diff
@@ -21,9 +22,29 @@ export function loadCoreTranslations() {
...
+              content: json
+            }
+          }, {
+            upsert: true
+          });
         }
       }
+
+      Assets.find({ type: "i18n" }).forEach((t) => {
+        Logger.info(`Importing ${t.name} translations...`);
```
aaronjudd: Can we update `translations...`  to just be `translation`. Minor detail, but inline with the rest of the startup messaging.

aaronjudd:
In the UI, I don't see any of the translations actually working at the moment. Doesn't even look like the english translations are working -> this might be related to the jquery-ui issue mentioned/fixed elsewhere.  I tested with a `reaction reset && reaction`.

aaronjudd:
Also, the Translations collection seems to be empty.

aaronjudd:
👍  looks good.

Issue #779, opened 2016-02-16
i18n translation key missing in translation files


saad91:
I have noticed that "Drop file to upload" button in product edit page is missing the translation key in "en.json" file. Im not sure if this is  okay or not, but i was expecting to find it in that file. Check the attached images for more details on this issue.
<img width="471" alt="screen shot 2016-02-16 at 10 10 24 pm" src="https://cloud.githubusercontent.com/assets/17137214/13087755/8b3bc724-d4fa-11e5-8f9a-5808c1ade4b7.png">
<img width="1279" alt="screen shot 2016-02-16 at 10 12 09 pm" src="https://cloud.githubusercontent.com/assets/17137214/13087756/8b4f64be-d4fa-11e5-919a-04783a7b730a.png">

aaronjudd:
It's ok (for english), as as long as there is fallback text in the html, but we should add to the en.json (and then replicate as needed to other languages).

aaronjudd:
@kieckhafer just need to verify that       "dropFiles": "Drop files to upload", key is attached here and then close (or fix).

kieckhafer:
The issue here is in our i18n files (en.js and ru.js are the only two where it's available), we are using dropFiles (plural) as the as the key, where in the data-i18n attribute, as well as in one other location, we are using dropFile (singular).

![screen shot 2016-07-12 at 7 03 08 pm](https://cloud.githubusercontent.com/assets/4482263/16789541/601131ba-4863-11e6-886c-e07a988c9280.png)

I've fixed this and created a pull request.

PR #1150, opened 2016-07-13
Fixes mismatched key name on i18n data-attribute (Closing #779)

kieckhafer:
The issue (#779) here is in our i18n files (en.js and ru.js are the only two where it's available), we are using dropFiles (plural) as the as the key, where in the data-i18n attribute, as well as in one other location, we are using dropFile (singular). I've updated this to use dropFiles in all locations.


aaronjudd:
👍  Approved.

mikemurray:
:thumbsup:

PR #1154, opened 2016-07-14
Fix Stripe dashboard panel saving key to database

spencern:
- [X] Description explains the issue / use-case resolved
- [X] Only contains code directly related to the issue
- [X] Has been linted and follows the style guide

Prior to this fix, the Stripe package would not save the stripe API key to the database from the dashboard. This PR fixes that issue.

thanks to @paulgrever for the discovery on this.


aaronjudd:
👍

mikemurray:
👍

PR #1161, opened 2016-07-18
Add "JobControl" plugin

### LINKED ISSUE #924: Failed to get currency exchange rates error
### Expected behavior
When I have a valid Open Exchange Rates API key Reaction should fetch exchange rates between my base currency and all other currencies

### Actual Behavior
Receive "Failed to get currency exchange rates." in console. Prices never show in Locale currency.

1. Start a Reaction session where you locale currency is not USD. (if in the US, a VPN is the easiest) 
2. Receive error message at console
3. Observe prices are shown in USD even though that's not your local currency.

zenweasel:
Closes #924 

This is a "simple as possible but no simpler" implementation of a job-control plug-in that just adds the one job that we currently run by default, `shop/fetchCurrencyRates`.

This obviously could be expanded to have a UI component, but trying to keep this as simple as possible to not introduce any new bugs for 14.

The code for actually adding a job is pretty much taken verbatim from @jshimko's code posted in Slack.

Here is the section for the docs:

# Job Control

Reaction Commerce has a Job Control system based on the excellent [job-collection](https://atmospherejs.com/vsivsi/job-collection) package. This allows you to run tasks in the background, separate from any user actions. You can easily extend or modify this to add your own jobs to do any kind of work where a task is appropriate.

There is a `Job` collection which is based on the `JobCollection` object provided in the package. You can then use then to build and queue your jobs.

There are essentially two parts you need to make your job active and that's a "Job Definition" and a "Job Schedule". You can look at the code we have built (in `imports/plugins/included/jobcontrol`) for fetching exchange rates as an example.

The `Jobs.processJobs` command creates the job definition (in this case just calling a Meteor method but it could be any code that can run on the server).

You can then use `new Job` to add the job to the queue. If you add the `cancelRepeats` option to the `save` portion, you can avoid adding duplicate jobs when you restart your server. You add them to the Event hooks for `onJobServerStart` or if you need to wait until the server is fully started you can use `afterServerInit`.


aaronjudd:
👍  approved.

aaronjudd:
👍  for the job control, but on further testing in relation to the OpenExchangeRates configuration.

 ```
17:07:06.390Z  INFO Reaction: OpenExchangeRates API not configured. Not adding fetchRates job
17:07:06.390Z  INFO Reaction: OpenExchangeRates API not configured. Not adding flushRates job
```
However, I actually do have the OpenExchangeRates key in my setup information, and it's displaying in the dashboard/core settings as well. 

Also, saving the OpenExchangeRates in the config does work and executed `shop/fetchCurrencyRates` .. but I noticed that there are now two entries in the `Jobs` collection for `shop/fetchCurrencyRates`.

File: imports/plugins/included/jobcontrol/server/jobs/cleanup.js [hunk 1]
```diff
@@ -0,0 +1,47 @@
...
+export default function () {
+  Jobs.processJobs("cleanup", {pollInterval: false, workTimeout: 60 * 1000}, function (job, cb) {
+    Logger.info("=======> Running cleanup Job");
+    let current = new Date();
+    current.setMinutes(current.getMinutes() - 60);
+    const ids = getJobIds();
+    if (ids.length > 0) {
+      Jobs.removeJobs(ids);
+    }
+    const success = `Removed ${ds.length} stale jobs`;
```
jshimko: Small variable typo.  This...
```js
const success = `Removed ${ds.length} stale jobs`;
```
should be...
```js
const success = `Removed ${ids.length} stale jobs`;
```

imports/plugins/included/jobcontrol/server/jobs/cleanup.js
```diff
@@ -0,0 +1,47 @@
+import { Jobs } from "/lib/collections";
+import { Hooks, Logger } from "/server/api";
+
+if (Hooks) {
```
jshimko: Now that we have module imports, we don't really need to check for the `Hooks` global anymore.  If that was null, the import at the top of this file would throw an error before you even get to that check.

imports/plugins/included/jobcontrol/server/jobs/cleanup.js
```diff
@@ -0,0 +1,47 @@
...
+  Hooks.Events.add("afterCoreInit", () => {
```
jshimko: Should me maybe create an `onJobsServerStarted` hook for scheduled jobs that kick off at startup?  I don't think the order actually matters much in this case, but it feels like it would make sense since there's a callback for `Jobs.startJobServer()`.

server/api/core/core.js
```diff
@@ -14,6 +14,7 @@ export default {
     // start job server
     Jobs.startJobServer(() => {
       Logger.info("JobServer started");
+      Hooks.Events.run("onJobServerStart", this);
```
jshimko: There's no object to alter for this particular hook, so no arg is needed there.  The `this` isn't really useful for anything there.  Additionally, since it's in an arrow function, the `this` is actually from the outer scope (which is just an anonymous function).  So it's certainly not breaking anything, but it's not ever going to be doing any useful either.

mikemurray:
👍

PR #1162, opened 2016-07-19
Plugin loader

jshimko:
This PR adds a plugin loader that automatically scans any plugin folders present in the `/imports/plugins/{core,custom,included}` folders and imports the following from each plugin (if they exist):

```
my-plugin/client/index.js
my-plugin/server/index.js
my-plugin/register.js
```

The loader generates two files with imports for all found files at `/client/plugins.js` and `/server/plugins.js`

Since the plugin loader requires the app to run in development mode at least once for a `meteor build` command to compile those imports, the Docker builds were failing.  So I created a bash script that runs in the Docker build process (right before the Meteor build process) and it does the exact same thing as the JS version that Meteor runs in development.

All tests pass and CircleCI is building/running the app successfully.


zenweasel:
Approved. Tested by dropping the `beesknees` plugin into the custom folder and it worked like a champ.

mikemurray:
So, assuming someone doesn't use `docker build` and doesn't run their meteor app in development mode first; how will the plugin index be generated then?

Needs documentation on how this works, and how to manually add imports, or that they need to run some script. @zenweasel has a PR https://github.com/reactioncommerce/reaction-docs/pull/49 in reaction-docs that could be updated with this information.

Also perhaps adding the plugin script to the `npm postinstall`; considering the node modules *have* to be installed.

aaronjudd:
I think `reaction reset` should also remove the existing `plugins.js` contents.

mikemurray:
@aaronjudd well if you want to get all fancy, then `reaction build` and `reaction start` can also build the `plugin.js` contents.

aaronjudd:
@mikemurray yup. I was thinking of just for testing, if you aren't removing those on reset (optionally with node_modules) then we really aren't testing a clean install, since none of that exists.

For instance , right now if I `rm -rf node_modules` and `client/plugins.js` and `server/plugins.js`

I get
```

Unable to resolve some modules:

  "jquery-ui/autocomplete" in /Users/aaronjudd/ongoworks/reaction/imports/plugins/included/product-variant/client/templates/products/productDetail/tags.js
(web.browser)
  "jquery-ui/sortable" in /Users/aaronjudd/ongoworks/reaction/imports/plugins/included/product-variant/client/templates/products/productGrid/item.js (web.browser)

If you notice problems related to these missing modules, consider running:

  meteor npm install --save jquery-ui

```

aaronjudd:
the jquery-ui issue, we just need to pin to the current version, which I will do in the development branch.  We do however need this branch to be updated from `development`, before merging.

mikemurray:
👍

PR #1166, opened 2016-07-19
remove tests and packages folders

aaronjudd:
these are no longer used by reaction core


zenweasel:
Should we add the `.gitkeep` for `public/custom` and `private/custom` in this PR as well?

zenweasel:
Approved

Issue #607, opened 2015-12-19
Cart not display all items on a low res


newsiberian:
Hello,
![screencast1](https://cloud.githubusercontent.com/assets/11130487/11914231/34720ee4-a69c-11e5-93ca-f545936a67bc.gif)
browsers: Chromium, Opera

aaronjudd:
@mikemurray @kieckhafer The swiping action is now broken in 0.14 -> there is no boundary bounce.  This needs review for 0.15rc

PR #1175, opened 2016-07-20
Fix boundary bounce on cart drawer

kieckhafer:
Fixw #607 

Boundary bounce was not working on initial page load, due to the Swiper
instance being in a hidden div, which means the width of Swiper was not
being set initially. This change moves Swiper initiation into our
openCartDrawer function, so the width is being set as it’s called
inside a visible div.

- [ ] Has tests.
- [ ] Has docs.
- [x] Passes all tests


aaronjudd:
👍  good to merge into 0.14

PR #1187, opened 2016-07-22
make sure Windows generates Posix import paths for plugins

### LINKED ISSUE #1186: Development plugin loader generates wrong paths on Windows
### Expected behavior

The plugin loader should generate import files at `client/plugins.js` and `server/plugins.js` with the JS imports for all plugins in the `/imports/plugins` directory.

### Actual Behavior

File generation is working, but the import paths added to the files are [being generated](https://github.com/reactioncommerce/reaction/blob/master/server/startup/plugins.js#L82-L84) as Windows paths with `\` when running in development on a Windows machine (instead of posix paths with `/` as needed in JS imports and everywhere else).  This is the [expected behavior of Node's path module](https://nodejs.org/api/path.html#path_windows_vs_posix) and there is [a method to force posix style paths](https://nodejs.org/api/path.html#path_path_posix), BUT it wasn't implemented until Node 0.11.15 and Meteor currently uses 0.10.46.  

Run Reaction on Windows.

Fix for this is on the way.

jshimko:
Fixes #1186 

The [development plugin loader](https://github.com/reactioncommerce/reaction/blob/master/server/startup/plugins.js) generates import files at `client/plugins.js` and `server/plugins.js` with the JS imports for all plugins in the `/imports/plugins` directory.  This was working, but the import paths added to the files were using Windows paths with `\`'s when running in development on a Windows machine.  

The paths are now normalized to use `/`'s instead via a simple regex.

### DIFF
--- a/server/startup/plugins.js
+++ b/server/startup/plugins.js
@@ -80,7 +80,8 @@ function getImportPaths(baseDirPath) {
 
   // create the import path
   const getImportPath = (pluginFile) => {
-    return "/" + path.relative(appRoot, pluginFile);
+    const importPath = "/" + path.relative(appRoot, pluginFile);
+    return importPath.replace(/\\/g, "/");
   };
 
   // get all plugin directories at provided base path


mikemurray:
👍

aaronjudd:
👍

PR #1188, opened 2016-07-22
Fix for broken loading of non-core translation files

newsiberian:
Hello, we should pass additional `ns` field to correctly load files with different translation namespaces. Before this PR only language name was passed to Assets doc. This is not enough in situations like so, when I have several files: `en.json`, `en-plugin1.json`, `en-plugin2.json`.

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

- added additional "ns" field for Assets schema  which
 is translation namespace;
- `loadCoreTranslations` now correctly loads non-core namespace
translation files;

### DIFF
--- a/lib/collections/schemas/assets.js
+++ b/lib/collections/schemas/assets.js
@@ -8,6 +8,13 @@ export const Assets = new SimpleSchema({
     type: String,
     optional: true
   },
+  /**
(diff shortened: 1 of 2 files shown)


aaronjudd:
👍  looks good.

mikemurray:
👍

PR #1190, opened 2016-07-22
Updated order of imports.

### LINKED ISSUE #1189: Update order of plugin imports
Update the order of plugin imports so that custom is loaded last. 

This is helpful with CSS overrides, and other cascading changes. For example, any CSS in plugins in the `included` directory will override any CSS in the `custom` directory; which is not ideal.

**imports currently currently**
- core
- custom
- included

**imports updated to**
- core
- included
- custom

mikemurray:
Fixes #1189 

Update order of variable declarations to be consistent with order of
imports.

### DIFF
--- a/.reaction/docker/scripts/plugin-loader.sh
+++ b/.reaction/docker/scripts/plugin-loader.sh
@@ -17,7 +17,7 @@ tee $CLIENT_PLUGINS_FILE $SERVER_PLUGINS_FILE > /dev/null <<EOF
 EOF
 
 # define plugin subdirectories
-PLUGIN_SUB_DIRS=("core" "custom" "included")
+PLUGIN_SUB_DIRS=("core" "included" "custom")
 
 # generate plugin import files on client and server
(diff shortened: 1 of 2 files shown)


jshimko:
👍  Approved.

jshimko:
FYI, this is going to create a merge conflict with my PR for the Windows path fix.  We both changed the same file.

jshimko:
Oh by the way, you need to update the order [in the bash script](https://github.com/reactioncommerce/reaction/blob/master/.reaction/docker/scripts/plugin-loader.sh#L20) too.  Just change that array:
```sh
# from this
PLUGIN_SUB_DIRS=("core" "custom" "included")

# to this
PLUGIN_SUB_DIRS=("core" "included" "custom")
```

mikemurray:
got it, thanks

jshimko:
Ok, now 👍 for real.

aaronjudd:
👍

PR #1196, opened 2016-07-22
release 0.14.1

aaronjudd:
- patches for windows
- updated load order
- product detail updates
- braintree fixes
- example and authnet settings form fixes

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,4 +1,7 @@
 <!-- Thanks so much for your PR, your contribution is appreciated! -->
+**Are you submitting to `development`?**
+
+We currently only accept PR's to `development` not `master`
 
 - [ ] Description explains the issue / use-case resolved
 - [ ] Only contains code directly related to the issue
--- a/.reaction/docker/scripts/plugin-loader.sh
+++ b/.reaction/docker/scripts/plugin-loader.sh
@@ -17,7 +17,7 @@ tee $CLIENT_PLUGINS_FILE $SERVER_PLUGINS_FILE > /dev/null <<EOF
 EOF
 
 # define plugin subdirectories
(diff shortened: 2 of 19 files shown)


zenweasel:
Approved

3Maestros:
I just pulled the master code which I believe is v14, I see that project hierarchies have changed ex : folder "packages" no longer exists, but  I dont see any mention of this major change any where? Is there ay documentation, I liked the old structure better is there any reason why this has changed?

jshimko:
@3Maestros The original Meteor package system is going to be deprecated in the (somewhat) near future in favor of NPM modules.  This was the first step toward complying with Node/NPM packages and the greater javascript ecosystem.  Meteor packages are certainly good for a variety of things that NPM packages can't do with Meteor yet (e.g. isomorphism), but they're highly proprietary to Meteor and can only be distributed on Atmosphere (which will also be going away).  So, while this part of the refactor was technically optional to do right now, it won't be optional for long.

There's certainly nothing stopping you from using Meteor packages of your own in the packages directory or from Atmosphere, but you won't be able to `import` any modules from the root Meteor app itself (the major limitation of Meteor packages that made us go for the change now instead of later).

So my personal recommendation is to embrace the Node/NPM best practices for building modules and save yourself from having to tear a Meteor package apart 6 months from now (or however long it takes Meteor to move completely to NPM).

You can either use traditional NPM packages or build a plugin module in the `/imports/plugins/custom` directory.  The conversion process of Meteor packages to the new modules is outlined here:
https://docs.reactioncommerce.com/reaction-docs/master/converting

aaronjudd:
Release notes are also located here: https://github.com/reactioncommerce/reaction/releases/tag/v0.14.0

PR #1198, opened 2016-07-25
Register inventory records on startup

### LINKED ISSUE #961: Inventory is not available until admin visits the PDP
### Expected behavior
Imported inventory should be available before the first user checks out

### Actual Behavior
Imported inventory is not inserted into the Inventory collection until an admin visits the PDP page (other actions might trigger this too, that's just what I stumbled upon). If you don't do this, then all orders are placed into backorder because there is no available inventory.

* do `./reaction reset && reaction`
* place an order
* Observe the following messages:
```
02:05:06.354Z  INFO Reaction: totalRequiredQty 1
02:05:06.354Z  INFO Reaction: availableInventoryQty 0
02:05:06.354Z  INFO Reaction: no inventory found, create 1 backorder
02:05:06.358Z  INFO Reaction: created 1 backorder records for product BCTMZ6HTxFSppJESk, variant CJoRBm9vRrorc9mxZ
```

zenweasel:
Closes #961 
Previously inventory registration wasn't happening until a Product collection hook fired which would not happen until you visited the PDP and edited a product in the dashboard. So I added a hook to register all products on startup. Also, since the Meteor method checked permissions, I moved the body of the method out into a regular function and made the Meteor method just a wrapper that checks permission.


aaronjudd:
👍  Looks good to me. There's some linting issues, but something funky is going on there...

zenweasel:
In this case the linter is wrong. This was existing code that was just moved, but according to the linter you should not indent for a `switch/case` statement which is clearly wrong so I am ignoring.

jshimko:
👍

PR #1206, opened 2016-07-26
Fix errant import

zenweasel:
If you use the current Logger imported here you get a `can't call method "warn" of undefined`. Just changed it to the standard import. Didn't bother creating a ticket.


aaronjudd:
👍

Issue #1148, opened 2016-07-12
Refund Error Handling not displayed in admin


aaronjudd:
Error handling for refunds, seems to be broken.  Additionally we need to limit refunds, they cannot be greater than the order value (and any discounts previously applied).

### Expected behavior
Admin user should see any errors that come back from refund handling in any of the payment providers.

Related to #1147 

### Actual Behavior

No errors are displayed. 

Server console displays a fatal log:

```
18:46:56.591Z FATAL Reaction:
  Attempt for refund transaction failed { _id: 'QCrf92AJGgodzMqty',
    sessionId: '3C4a9M6YkvAdKog4K',
    userId: 'nfK8ktR5FpDj4LRhu',
    shopId: 'J8Bhq3uTtdgwZx3rz',
    workflow:
     { status: 'coreOrderWorkflow/processing',
       workflow: [ 'coreOrderWorkflow/created', 'coreOrderWorkflow/processing' ] },
    billing:
     [ { address: [Object],
         _id: 'jCQYTEpdnErpgpAuM',
         paymentMethod: [Object],
         invoice: [Object] } ],
    shipping

```

### Steps to Reproduce the Behavior

Attempt to refund an amount in the order workflow during admin.

zenweasel:
Actually, this applies to all Payment Provider-related error messages in the admin. Also, methods for which we don't/can't support refunds should have that option removed/greyed out.

aaronjudd:
Currently it looks like `refunds` just aren't working... in any payment package. refunds don't seem to work anywhere at the moment. Also the currency needs formatting in the refund alerts.

aaronjudd:
Popup handling resolved in #1219, only service that's currently not having errors is Stripe, and we need to test that that isn't just a silent warning.

aaronjudd:
Authorize.net errors 
```
I20160728-15:16:13.785(-7)? Exception while invoking method 'orders/refunds/create' Error: Did not check() all arguments during internal call to 'authnet/refund/create'
I20160728-15:16:13.785(-7)?     at [object Object]._.extend.throwUnlessAllArgumentsHaveBeenChecked (packages/check/match.js:462:1)
I20160728-15:16:13.785(-7)?     at Object.exports.Match._failIfArgumentsAreNotAllChecked (packages/check/match.js:110:1)
```

Capt-Slow:
NOTE: Verify when submitting Braintree refund, alert box appear and when closed does not present the following error in the console:

[...]

PR #1210, opened 2016-07-26
Updates to the way discount amounts are processed

### LINKED ISSUE #1149: Should not be able to discount an amount greater than order.
### Expected behavior
UI and methods should not allow a discounts amount that is greater than the order, or a previously created approval.

### Actual Behavior
Admin is permitting values larger than the order to be entered into the discounts input and applied.
No error is shown.

Create an order, sign in as admin, apply an larger than order discounts value and apply.

kieckhafer:
Fixes #1149

Updated discounts code inside of the shippingInvoice panel to only allow discounts up to the amount of the total cart value.

A few other small fixes:
Typos
included lodash in file page that was missing it

- [ ] Has tests.
- [ ] Has docs.
- [x] Passes all tests


server/methods/core/orders.js
```diff
@@ -139,6 +139,18 @@ Meteor.methods({
...
 
+    // Server-side check to make sure discount is not greater than orderTotal.
+    let orderTotal =
+      order.billing[0].invoice.subtotal
+      + order.billing[0].invoice.shipping
+      + order.billing[0].invoice.taxes;
+
+    if (discount > orderTotal) {
+      const err = "Discount is greater than the cart total";
+      Logger.error(err);
```
zenweasel: I'm not sure we need to log this error here at least not at this level since it's just a user error that we handle. There is nothing an admin can/needs-to do

server/methods/core/orders.js
```diff
@@ -139,6 +139,18 @@ Meteor.methods({
...
+      throw new Meteor.Error("orders/approvePayment.discount-amount", err);
```
zenweasel: Do we have a standard we are using there? Can we have one?

imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
```diff
@@ -58,14 +59,33 @@ Template.coreOrderShippingInvoice.events({
...
+    if (discount > orderTotal) {
+      Alerts.inline("Discount cannot be greater than original total price", "error", {
+        placement: "coreOrderShippingInvoice",
+        i18nKey: "order.invalidDiscount",
+        autoHide: 10000
+      });
+    } else {
+      Meteor.call("orders/approvePayment", order, discount, (error) => {
+        if (error) {
+          // Log & show error if client-side fails
```
zenweasel: I think you might mean "server-side" in that comment? And do we actually want to display that error in the console?

server/methods/core/orders.js
kieckhafer: @jshimko is this a standard we are using other places, or just something that was thought for this instance?

We definitely should come up with our own error reporting standard, as @jshimko and I were discussing, the current system with http status related codes doesn't make a lot of sense for some of the things we need it for.

I'd be happy to discuss further and take the lead on adding it into the docs somewhere if we don't already have something similar.

aaronjudd: I thought we had settled this discussion previously (error object, error string) for Logger?  I thought this was what @jshimko had proposed.  (otherwise we just used a string). The http status code were traditionally for `throw new Meteor.error` not `Logger`.

aaronjudd: If it there is a standard, it's NOT a standard until it's documented here: https://docs.reactioncommerce.com/reaction-docs/master/styleguide and here: https://docs.reactioncommerce.com/reaction-docs/master/logging

zenweasel: I remember having a discussion and agreeing that the HTTP code were irrelevant but I don't remember it being a settled issue.

aaronjudd: just nitpicking, but I'd rather see `error` than `err` or `e`. in this case, it doesn't bother me, but it does bother me on try/catch statements.

imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
kieckhafer: I did mean client-side. The thought process on this is `if (discount >orderTotal) {` is in place for client-side actions, i.e. clicking the "Apply Discount" button. If someone is able to bypass this by not using the UI and directly calling `Meteor.call("orders/approvePayment", order, discount, (error) => {` we still want something to explain what's happening. This should never happen if anyone is using platform in a traditional way however, so maybe we do want to remove the log?

imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
```diff
@@ -7,6 +7,7 @@ import { ReactiveVar } from "meteor/reactive-var";
 import { Reaction, i18next, Logger } from "/client/api";
 import { NumericInput } from "/imports/plugins/core/ui/client/components";
 import { Media, Orders, Shops } from "/lib/collections";
+import _ from "lodash";
```
aaronjudd: we did agree, and it's not documented yet, so I will add now, but the general order of import statements should be:

import npm
import meteor
import local

imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
zenweasel: Are you not calling a server-side method and that is in fact what would fail there?

No we don't want to pass on any secure information that may leak out of an unredacted error message. We could however check for "expected" (if any) errors and pass on friendly messages.

imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
zenweasel: I would suggest one step further in

import npm
import meteor
import third-party
import local

server/methods/core/orders.js
aaronjudd: Just to clarify, just in case, the currently documented standard for Logger is `"string", error` as well as for Meteor.error - the old http status thing is dead.

imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
kieckhafer: Correct, the server side method is not / should not be ever called if the discount > totalPrice, the validation is / should be all done client-side. 

The `// Log & show error if client-side fails` is there, and a second check is done once we get to the server-side, just in case somehow, someway `Meteor.call("orders/approvePayment"...` gets called without using the UI.

zenweasel: Ok, but my point being the comment is confusing. You call a server-side method, check for a server-side error and then say "Log & show error if client-side fails" inside the check for a server-side error.

imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
aaronjudd: when isn't a npm package 3rd party?

zenweasel: Sorry, should have said third-party Meteor packages or something similar. e.g. AutoForm, SimpleSchema

imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
kieckhafer: Ah, ok, I see how it's confusing. Is the best solution taking that comment out, as it's pretty self explanatory that an error is being thrown?

zenweasel: I agree that the code is pretty self-explanatory and the comment is unnecessary.

server/methods/core/orders.js
```diff
@@ -139,6 +139,18 @@ Meteor.methods({
...
     }
 
+    // Server-side check to make sure discount is not greater than orderTotal.
+    let orderTotal =
+      order.billing[0].invoice.subtotal
+      + order.billing[0].invoice.shipping
+      + order.billing[0].invoice.taxes;
+
+    if (discount > orderTotal) {
+      const err = "Discount is greater than the cart total";
```
zenweasel: Probably should be "order total". There is no cart involved here.

aaronjudd:
Pinned jquery to 1.10.5, resolves console errors and broken easing. 

Created update issue for jquery #1215

aaronjudd:
👍  approved.  the discount handling seems to be correct now.  Refunds are a different issue, see: #1148

mikemurray:
👍

PR #1219, opened 2016-07-28
Pass isConfirm when calling callback

zenweasel:
The problem was that refunds were not working for any payment method. This was because the confirmation alert was always returning `undefined`. @mikemurray informed me that a new version of Alerts had been rolled out and he had to patch this version to use callbacks.

I discovered that when calling the callback, `isConfirm` was not being passed in, which was why it was never `true`.

To Verify that this PR works:
Place an order.
Approve and Capture.
Issue a Refund and verify that it works.

### DIFF
--- a/imports/plugins/core/layout/client/templates/layout/alerts/reactionAlerts.js
+++ b/imports/plugins/core/layout/client/templates/layout/alerts/reactionAlerts.js
@@ -68,7 +68,7 @@ Object.assign(Alerts, {
         ...titleOrOptions
       }).then((isConfirm) => {
         if (isConfirm === true) {
-          messageOrCallback();
+          messageOrCallback(isConfirm);
         }
       });
     }
@@ -83,7 +83,7 @@ Object.assign(Alerts, {
       ...options
     }).then((isConfirm) => {
       if (isConfirm === true) {
-        callback();
+        callback(isConfirm);
       }
     });
   },


kieckhafer:
👍

aaronjudd:
Tested with PaypalExpress, refunds are not being processed.

aaronjudd:
Also this PR has the reverted layout in it.

kieckhafer:
-1' d my +1. Misunderstood what payment platforms should and should not be working.

aaronjudd:
It's my opinion that if refunds aren't working for a particular payment method, then we need to make that obvious in some way. Otherwise, my expectation is that refunds work for all payment methods.

zenweasel:
@aaronjudd That's fair but I am not sure how that relates to this PR

aaronjudd:
My comments don't generally need to relate to the specific PR.  However, that comment is to head off the comment that was sure to follow that some payment method doesn't actually support refunds yet.

zenweasel:
@aaronjudd How does this PR have the reverted layout in it?

aaronjudd:
Interesting, I actually rechecked that several times, but now it doesnt.  (Honestly). I'll re-review, I know this sounds hokey - but GH was being bad.

aaronjudd:
👍  for the fixed popup sending working with refunds.  However the only service that I didn't have errors with was Stripe.

PR #1220, opened 2016-07-28
To fix issue #1197.

### LINKED ISSUE #1197: Can't add email after purchase as guest
### Expected behavior
Be able to add an email address after purchase as **guest**

### Actual Behavior
At the "Thank you page" user can't add email, in the console says:
``
Error invoking Method 'orders/addOrderEmail': Access Denied [403]
``

1. Add to cart Basic Reaction Product
2. Continue as guest
3. Checkout with paypal
4. At the "Thank you page", fill the email field
5. Click button "Send"

thank-you-page

hrath2015:
Yes

- [ X] Description explains the issue / use-case resolved
- [X ] Only contains code directly related to the issue
- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests
- [ X] Has been linted and follows the style guide

Resolves #1197

### DIFF
--- a/server/methods/core/orders.js
+++ b/server/methods/core/orders.js
@@ -523,10 +523,15 @@ Meteor.methods({
   "orders/addOrderEmail": function (cartId, email) {
     check(cartId, String);
     check(email, String);
+    /**
+    *Instead of checking the Orders permission, we should check if user is
+    *connected.This is only needed for guest where email is
+    *provided for tracking order progress.
+    */
 
-    if (!Reaction.hasPermission("orders")) {
-      throw new Meteor.Error(403, "Access Denied");
-    }
+    if (!Meteor.userId() {
+       throw new Meteor.Error(403, "Access Denied. You are not connected.");
+     }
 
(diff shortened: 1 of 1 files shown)


aaronjudd:
` if (!Meteor.userId() {` does solve the issue.  I want to look a little more at this though, because every user has a Meteor.userId (unless this were some random non ui client trying to update an order).

hrath2015:
`orders` permission check was introduced after RC  V.12. Till V12 there was no check.   Orders permission are specific to Admins who works on Order workflow. So in this case it can not be used.   Instead of removing it completely,  check is done to see if client is connected.   For  full-proof solution we need to introduce another guest role "orderemail" or ask for  email earlier in the checkout flow.

aaronjudd:
@hrath2015 that's exactly what i was thinking, adding a "orderaccess" or something like that upon successful completion. `orders` isn't correct, you are right I think only admin has that permission. I'll accept this PR but will make a note to revisit and add some more permissions.  I believe the security roles will only let you modify your own record anyways, I just wanted to double check that.

Issue #1134, opened 2016-07-05
downloadable font: rejected by sanitizer /fonts/fontawesome-webfont.woff2?v=4.6.3


jujes:
### General question   

**Development branch build of reactioncommerce:reaction**

Someone know about this font issue?

![screen shot 2016-07-05 at 11 46 46 pm](https://cloud.githubusercontent.com/assets/2881017/16601066/c7d8f5f2-430a-11e6-94fd-e4b9c020d439.png)

```
unreachable code after return statement f474c6bbefe48578647e20819f049973ea16351c.js:299:23956
unreachable code after return statement f474c6bbefe48578647e20819f049973ea16351c.js:299:24274
unreachable code after return statement f474c6bbefe48578647e20819f049973ea16351c.js:299:23956
unreachable code after return statement f474c6bbefe48578647e20819f049973ea16351c.js:299:24274
mutating the [[Prototype]] of an object will cause your code to run very slowly; instead create the object with the correct initial [[Prototype]] value using Object.create f474c6bbefe48578647e20819f049973ea16351c.js:26:20071
downloadable font: rejected by sanitizer (font-family: "FontAwesome" style:normal weight:normal stretch:normal src index:1) source: https://rootsoap.com/fonts/fontawesome-webfont.woff2?v=4.6.3 /:6985:12
downloadable font: rejected by sanitizer (font-family: "FontAwesome" style:normal weight:normal stretch:normal src index:2) source: https://rootsoap.com/fonts/fontawesome-webfont.woff?v=4.6.3 /:6985:12
downloadable font: rejected by sanitizer (font-family: "FontAwesome" style:normal weight:normal stretch:normal src index:3) source: https://rootsoap.com/fonts/fontawesome-webfont.ttf?v=4.6.3 /:6985:12
```

Thanks in advance,

aaronjudd:
Prequel is the `development` branch.   Need to look into that error though, that's the first I've seen of it. @jshimko ideas?

jshimko:
I can't reproduce this, but [a quick Google search](https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=downloadable%20font%20rejected%20by%20sanitizer) turns up a few common causes for that error with Firefox.

[Memory issues or a corrupt font file](https://support.mozilla.org/en-US/questions/913498)...
[...]

PR #1223, opened 2016-07-29
Add Reaction CDN sources for font and favicon assets

jshimko:
- move FontAwesome and Fontello fonts to new Reaction Assets CDN
- create default favicons for all platforms and move to CDN
- set up local fallback for fonts an favicons so everything works offline 
- update Browser Policy configs
- add custom headers for CDN assets
- update to Meteor 1.4.1
- update dependencies


jshimko:
Forgot to mention, this also happens to resolve #1134 (which was the original point of doing this in the first place).

aaronjudd:
👍  looks good.

kieckhafer:
👍

PR #1225, opened 2016-07-29
Do not allow refunds greater than original purchase amount

kieckhafer:
Partial fix of #1148

We were allowing refunds greater than the original purchase price to be sent to the payment providers. The providers themselves would send back and error, but it should not have gotten that far. This PR fixes that by validating on the client-side before it's sent to the provider.

There will be a second part to the solution of this issue, and that will consist of fixing the error messaging and callbacks from each separate payment provider.

- [ ] Has tests.
- [ ] Has docs.
- [x] Passes all tests


aaronjudd:
I get ```Exception in delivering result of invoking 'orders/refunds/create': TypeError: Cannot read property 'split' of undefined
    at setParameters (http://localhost:3000/packages/modules.js?hash=27d96b14b1570bf781e9cfcc23643d61bf5c54d0:1162:36)
    at modalDependant (http://localhost:3000/packages/modules.js?hash=27d96b14b1570bf781e9cfcc23643d61bf5c54d0:1381:5)``` after paypalExpress refund..  I didn't expect express to work, but I think this is a new error.

aaronjudd:
👍  `Refund(s) total cannot be greater than captured total` display when amount too large.

kieckhafer:
I haven't seen that PayPal error yet, but I know there are a lot of different ones I"ll be working through with the different providers, I'll add that to the list for ones to look out for.

mikemurray:
can't refund an exact amount

screen shot 2016-07-29 at 2 16 15 pm

kieckhafer:
@mikemurray the issue should be fixed, needed to round the adjustedTotal - it was displaying 5.98 but in reality it was 5.97999999999, so it wasn't allowing a 5.98 refund.

mikemurray:
Also getting this error when I apply a discount.

It applied the discount, and also displayed it as a refund.

screen shot 2016-07-29 at 3 30 27 pm

kieckhafer:
Something fishy is going on with refunds. This is going to be a deeper dive. I am able to replicate what @mikemurray posted above.

It seems if you give a discount, this also doubles down and turns into a refund. I'm thinking this is just a UI issue, but I need to do a little more research into it.

I gave a $75 discount, which was applied, but then it was also applied a second time as a refund, as soon as I hit "Capture Payment", with no further user input:
screen shot 2016-07-29 at 5 37 27 pm

Because of this, it looks like the store owes the customer $65.07, instead of the customer paying the store $9.93.

kieckhafer:
I am closing this PR for the time-being. While technically I think the original issue being addressed is fixed, it was coded and tested, and the PR opened, with data that is not being correctly populated. We are going to do a deeper dive into getting that data correctly populated, at which time we can re-open this PR.

PR #1234, opened 2016-08-02
Add convenience function to globally change layouts.

zenweasel:
If you want to use a custom layout template you need to override it in DEFAULT_LAYOUT but also in every layout record in Shops. This just adds a convenience feature so that plugins can easily override all layout records. It is not called in current code.


jshimko:
👍

kieckhafer:
👍

PR #1240, opened 2016-08-03
Adjust inventory on shipment

### LINKED ISSUE #964: Implement simple inventory methods in reaction-inventory
Currently the `reaction-inventory` module doesn't subtract inventory when an item is shipped. It should probably do that by default if inventory tracking is turned on.

zenweasel:
Closes #964 
Implements two new status changes:

1. When an order is placed, moves inventory from "reserved" to "sold".
2. When an order is shipped moves inventory from "sold" to "shipped".

How to test:

1. Add items to cart
2. Check that an appropriate number of items are marked `workflow.status` = "reserved".
3. Finish your order
4. Verify those items are now marked `workflow.status` = "sold".
5. Log into the dashboard.
6. Ship the same order
7. Verify that the same items are now marked `workflow.status` = "shipped".

I think that `inventory/setStatus` probably needs to be restructured to handle different types of logic for different transition states (and to be able to extend it without modifying the existing code).


imports/plugins/included/inventory/server/startup/hooks.js
```diff
@@ -89,3 +89,52 @@ Products.after.insert((userId, doc) => {
   }
   Meteor.call("inventory/register", doc);
 });
+
+function markInventoryShipped(doc) {
```
aaronjudd: these functions could be exported and then tested independent of the hooks

aaronjudd:
the naming of inventory, inventory2 is leftover from the module refactor and should be merged/cleaned up, as well as not exactly sure why hooks are in the `startup` folder and not in a `hooks` folder.. don't have to cleanup in this issue.. but something I've been meaning to look at.

aaronjudd:
👍 

[Approved with PullApprove](https://pullapprove.com/reactioncommerce/reaction/pull-request/1240/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=reactioncommerce/reaction)

mikemurray:
👍

PR #1247, opened 2016-08-04
fix for #1072

### LINKED ISSUE #1072: Should not be able to create multiple tags with same slug
### Expected behavior

Case should behave consistently tags / hashtags. Case should either be respected (`Camping` is a different tag than `camping` is different than `CAMPING`) or should be normalized (`Camping` is the same tag as `camping`).
### Actual Behavior

If I create an uppercase tag called `Camping` and then add a lowercase tag `camping` both tags are created and both have a slug `camping` but visiting `/tag/camping` only filters products connected to the first `camping` tag by ID.

Add a few products to project. Add a capitalized tag to Product A, add lowercase tag to Product B. Visit `/tag` route.

users/developers

nialexsan:
Resolve #1072


zenweasel:
Approved. Seems to fix this issue. Needs to be merged with Development

aaronjudd:
👍

jshimko:
👍

PR #1248, opened 2016-08-04
data-id for correct tag dragging in navbar

nialexsan:

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

without data-id outputs error and hangs js

### DIFF
--- a/imports/plugins/core/ui/client/components/tags/tagItem.html
+++ b/imports/plugins/core/ui/client/components/tags/tagItem.html
@@ -34,7 +34,7 @@
 </template>
 
 <template name="tagEditable">
-  <div class="rui item edit draggable">
+  <div class="rui item edit draggable" data-id={{tag._id}}>
     <div class="rui tag edit draggable {{className}}" data-id={{tag._id}}>
       {{> button icon="bars" className="js-drag-handle" status="default"}}
       {{> textfield name="tag" value=tag.name onChange=handleTagNameUpdate}}


aaronjudd:
I don't see what this is fixing, or maybe this was already fixed in development?  I'm able to drag and re-arrange top menu, sub menus, and tags on the product page.  While this change doesn't break the top menus, the product page tag dragging breaks with this change.

nialexsan:
Try to drag a tag from one tag group to another like this (from shop1 to shop2):
image

image

aaronjudd:
@mikemurray is that the intended behaviour? I was able to drag the overall group of tags, but not the individual sub-tags from one group to another.  If this PR is enabling that behaviour, cool?  

@nialexsan  However, this PR still seems to break the tag component in the context of the product detail page.. can you test development version versus this update and verify?

nialexsan:
@aaronjudd sorry, I can't find where the PR breaks existing component

aaronjudd:
@nialexsan on the product detail page, dragging and dropping of the tags there. This is a shared component. 

@mikemurray can you review?

mikemurray:
@aaronjudd Yes, you should be able to move sub tags in the groups to other groups. But groups themselves, at the moment, can only be rearranged in the dropdown menu.

Also you can move tags from the product detail page to the navigation, and vise versa; however, it seems the tag is removed and never added to the navigation bar. In `development` it just gives me an error.

@nialexsan your PR fixes the js error in the navigation bar, but now causes tags dragged from the product detail page to the navigation bar to be removed.

If a tag is dragged from the product detail page to a group in the dropdown menu, it stays.

Any tag dragged from the navigation bar / dropdown to the product detail page, it is removed.

Your PR might not be the root cause of these issues though; it does fix the issue of not being able to re-order between groups.

zenweasel:
No response so closing for now for this release. Please reopen whenever you have some time to address the issues.

Issue #1098, opened 2016-06-28
Cannot use triggersEnter and triggersExit in registry


spencern:
### Expected behavior
To be able to add functions to the `triggersEnter` and/or `triggersExit` array and have them run when visiting that route.

E.g. in the registry:

Ideal scenario - pass array of functions similar to how flowRouter currently works
```
// reaction-product-variant
  { route: '/tag/:slug?',
    name: 'tag',
    template: 'products',
    workflow: 'coreProductWorkflow',
    triggersEnter: [ function () { console.log('trigger fired'); } ] },
```

suboptimal scenario - passing a string representing a function (how to pass args?)
```
// reaction-product-variant
  { route: '/tag/:slug?',
    name: 'tag',
    template: 'products',
    workflow: 'coreProductWorkflow',
    triggersEnter: [ 'consoleLog' ] },
```

### Actual Behavior
Adding a function to the triggers enter array causes an empty array to be added to the document
e.g.
```
// reaction-product-variant
  { route: '/tag/:slug?',
    name: 'tag',
    template: 'products',
    workflow: 'coreProductWorkflow',
    triggersEnter: [] },
```
because the Schema requires an array of strings (`triggersEnter: [String]`);

Adding a String to the array, causes the following client error to be thrown and the page does not render.
```
Exception from Tracker recompute function:
debug.js:41 TypeError: Cannot read property 'length' of undefined
    at parse (index.js:85)
    at Function.compile (index.js:104)
    at Router.path (router.js:56)
    at Router.go (router.js:170)
    at router.js:68
    at withValue (dynamics_browser.js:23)
    at Router.reload (router.js:67)
    at Router.ReactionRouter.initPackageRoutes (init.js:178)
    at router.js:6
    at Tracker.Computation._compute (tracker.js:323)
```
### Steps to Reproduce the Behavior
Add function/string to `triggersEnter` property of registry in reaction package.

prinzdezibel:
@spencern I have had the same problem. My workaround for now is to re-define the route manually again like this:

```
Meteor.startup(function(){
  ReactionRouter.group({name:"shop"}).route("/", {
    name: "index",
    action: function () {
      ReactionLayout({
        adminControlsFooter : "adminControlsFooter",
        dashboardControls : "dashboardControls",
        layout : "coreLayoutEasywine",
        layoutFooter : "footerEasywine",
        layoutHeader : "layoutHeaderEasywine",
        notFound : "productNotFound",
[...]

PR #1253, opened 2016-08-05
Add route hooks API

jshimko:
This PR add the `Router.Hooks` namespace which provides a router-agnostic API for registering functions to be called either on specific routes or on _every_ route.  Like FlowRouter, there is an option for `onEnter` and `onExit` and the function hooks are passed the same `context` object to optionally do something with.

## API
```js
import { Router } from '/client/api';

// Register a hook on a specific route
// (can be called as many times as needed to add more than one callback)
Router.Hooks.onEnter(routeName, callback);
Router.Hooks.onExit(routeName, callback);

// Register a hook on every route
// (can be called as many times as needed to add more than one callback)
Router.Hooks.onEnter(callback);
Router.Hooks.onExit(callback);

// Get all registered hooks for a specific route (returns array)
Router.Hooks.get("onEnter", routeName);
Router.Hooks.get("onExit", routeName);

// Get all registered hooks that run on every route (returns array)
Router.Hooks.get("onEnter", "GLOBAL");
Router.Hooks.get("onExit", "GLOBAL");

// Run all registered hooks for a specific route
// (context object is optional)
Router.Hooks.run("onEnter", routeName, context);
Router.Hooks.run("onExit", routeName, context);

// Run all registered global hooks
// (context object is optional)
Router.Hooks.run("onEnter", "GLOBAL", context);
Router.Hooks.run("onExit", "GLOBAL", context);
```

That's the whole public API, but most users will only need to register new hooks (onEnter, onExit) because running all of the hooks is already taken care of internally.

#### Import
```js
import { Router } from '/client/api';
```
#### Route-specific Hooks
```js
// create a function to do something on the product detail page
function logSomeStuff() {
  console.log("We're arriving at the product page!");
}
// add that to the product detail page onEnter hook
Router.Hooks.onEnter("product", logSomeStuff);
```
#### Global Route Hooks (every route)
```js
// create a pageview tracking function
function trackPages() {
  analytics.page();
}
// track page views on every route
Router.Hooks.onEnter(trackPages);

// or put more simply...
Router.Hooks.onEnter(() => analytics.page());
```
#### Using route context in your hook
The same context object from FlowRouter is available to every callback
```js
function logSomeContext(context) {
  console.log("The current route details...");
  console.log("Params: ", context.params);
  console.log("Query Params: ", context.queryParams);
  console.log("Path: ", context.path);
  console.log("The route object: ", context.route);
}
// log out route details on every route
Router.Hooks.onEnter(logSomeContext);
```

One of the main motivations here was to provide an API for both internal and external contributors to add hooks to any or all routes without having to manually register routes with the regular FlowRouter API. But perhaps most importantly, this aims to start standardizing our own router-agnostic route hook API, so we can keep it consistent in the event that we switch routers in the future (because Javascript).


jshimko:
Oh yeah, the original point of this effort was to address/fix #1098.

aaronjudd:
@jshimko I'd like to see a PR doc added/updated before we pull this. mostly a copy and paste of your PR comment.

@zenweasel @Capt-Slow I think this is a good example of where I'd like to start seeing some client tests added.  I think we're all a little fuzzy on how to approach client tests within the new meteor testing paradigm..   I won't push for them on this PR, but some guidelines / examples for client PRs would be great.

aaronjudd:
👍  very nice.

zenweasel:
👍  Added an example to the `reaction-example-plugin`, works great.

PR #1259, opened 2016-08-06
Create a new workflow for the product list route and template

spencern:
- [X] Description explains the issue / use-case resolved
- [X] Only contains code directly related to the issue
- [X] Has been linted and follows the style guide

We're having issues overriding the `/tag/:slug?` route template - `products` from a custom theme. It seems that the issue is that both the `product` route and the `tag` route use the same workflow (`coreProductWorkflow`) and different templates. When we override this workflow from a different module and layout - `getoutfittedLayout` in our case, we are not able to override each template separately.

What this PR accomplishes is separating the `tag` product list browsing workflow from the `product` product detail page workflow. This permits us to override each workflow and thus each template separately from a custom module.


spencern:
Would be interested in some feedback on this issue. Is this a reasonable direction to enable overriding of core templates?

 I'm now seeing that this is also needed in `coreCartWorkflow` to override the `cartCompleted` workflow. 

Any thoughts on if there's a better way to permit overriding of these templates or if this may cause future trouble because of the long term design of the `workflow` idea?

zenweasel:
A few thoughts:

1. You guys are taking the idea of "not modifying core" to include never modifying any of the data that's written by core packages. That is a little farther than we have been suggesting to take it and that's what necessitating the need for a new `cartWorkflow`.  If you modify that record in your plugin, that's also repeatable. Curious why the need to have such a separation? If there was a declarative method of doing the same thing would that make you more comfortable?

2. I don't think our current implementation of workflows as "an ordered list of templates" really applies to anything other than checkout (separate from the idea of "workflow" for a cart). I would like to see us remove that element from parts of the site that are not workflows such as everything about browsing products or "about us" type pages. Maybe things like specifying templates, etc. could be move to route declarations if one could override routes? (I feel like this is more what you guys are trying to accomplish?)

spencern:
I think my mindset is being driving in part by past experiences with platforms (Shopify, Wordpress, Spree, etc). It seems like one should be able to install and uninstall 'themes' in a production enviornment.

Right now if I add my theme as a plugin, I'm not sure what the correct approach to changing the data that the core theme installs. I haven't been able to do that from within the registry, without changing these workflows which is where this PR came from. I was under the assumption that RC was trying to use the registry to setup and modify 'core' style data such as routes, templates, layouts etc.

I can think of 4 ways to accomplish updating _all_ of the customer facing templates in RC right now. If there's another way that I'm missing, I'd love to know.

1. By changing the `layout` using `Session.set("DEFAULT_LAYOUT", "myLayout")` and then adding registry entries that attach my new templates to existing workflows with my new layout. This feels _right_, but currently doesn't work for all templates (hence this PR).  I'm able to change some templates using the registry, but I haven't been able to make changes to all templates using this process. Definitely possible that I'm doing something wrong. 

2. By writing some type of server auto-run code that waits until the shop layouts, routes, etc are populated in the db and then overwrites/updates all layouts and routes in the Shops collection in the database to use the installed theme's templates instead of the core templates. This approach feels brittle and prone to error to me, but it's possible I'm missing something here.

3. Directly edit the `register.js` files in core modules such as `product-variant` and `checkout`. This would work, but isn't something that can be done from a custom plugin

4. The last option that I can think of is what we have been doing up until 0.14 and that is to edit all of the template files directly while keeping the registries unmodified. This involves changing the most core files and is perhaps the easiest path to start down and the hardest to maintain.


To answer your 2nd point, it seems like routes and workflows and templates are all tied up together in ways that I don't fully understand. All we are trying to do is apply our custom 'templates' (in the Blaze sense) so that we can change the look and feel of the customer facing parts of RC. In some cases our look and feel updates involves changing the schemas or adding server methods as well, but currently that part seems pretty straightforward and it's the UI that's challenging to make changes to.

zenweasel:
So to separate out the two problems here:

1. For the product-related sections, if you could override template/layout, etc. in a route definition would that give you what you are looking for?

2. For the cart workflow, if you could say something to the effect of "cartWorkflow = []" and specify a custom workflow that would get used instead of the default would that give you what you are looking for? So basically the way you would modify the cart workflow would be declaring a new one and using that instead? Maybe there could be an "inherit or replace" option so that you wouldn't need to re-specify the entire workflow just to change one template.

These are just ideas I have for trying to iterate forward.

spencern:
I agree that these are better identified as two separate issues.

1. Yes, I think that would solve our problems.

2. This approach sounds like a good start for the cart workflow. I like the idea of being able to swap out certain templates with a replace option.

zenweasel:
@spencern Agreed (this is an issue not a PR). Do we want to move any of this to your "Challenges" ticket?

zenweasel:
Closing this for now as we should consolidate discussion on the issue @spencern created.

PR #1269, opened 2016-08-08
Adds namespace field to layout schema

spencern:

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests

Our current need for this is to override checkout templates. 

Currently if I add templates to the registry for checkout steps, they are added to the existing checkout steps and as far as I can tell there isn't a good way to replace or override these checkout step templates.

This PR permits overriding of checkout step templates in the following way: 

```
{{ each reactionTemplate workflow="coreCartWorkflow" container="checkout-steps-main" namespace="getoutfitted" }}
  // Loads only checkout step templates with `namespace: getoutfitted`
  {{> checkoutSteps }}
{{/each}}
```

and in a custom module defining checkout step layouts that look like:
```
{
      template: "goCheckoutLogin",
      label: "Login",
      workflow: "coreCartWorkflow",
      container: "checkout-steps-main",
      namespace: "getoutfitted",
      audience: ["guest", "anonymous"],
      priority: 1,
      position: "1"
    },
```

I think there will be other uses that become obvious as I continue to override templates and build out our custom theme.

### DIFF
--- a/lib/collections/schemas/layouts.js
+++ b/lib/collections/schemas/layouts.js
@@ -110,6 +110,10 @@ export const Layout = new SimpleSchema({
     type: String,
     optional: true
   },
+  namespace: {
+    type: String,
+    optional: true
+  },
   container: {
     type: String,
     optional: true


zenweasel:
I am tempted to say just merge this since it doesn't break anything and will unblock you, but my reluctance is whether this muddies anything in terms of how we override templates/workflows. Specifically, does RC have any plans to use this `namespace` field anywhere.

spencern:
I had merged this with our current branch internally, but after some research and #1278 we're not currently using this method as there are other blockers. Might still be worth discussing this path as I think it still opens up a few new paths, but not as critical as I'd hoped.

aaronjudd:
Closing will review and respond in #1278

PR #1273, opened 2016-08-09
Fix for decrement cart function 

### LINKED ISSUE #1003: Calling "cart/removeFromCart" with a quantity does not decrement the quantity
### Expected behavior
Calling "cart/removeFromCart" with a quantity should decrement the quantity for that cartItem

### Actual Behavior
Quantity is not decremented

I don't believe this functionality is currently used in the stock shop since we don't currently allow you to change quantities.

1. Add an item to the cart
2. Execute `ReactionCore.Collections.Cart.findOne()` in the browser console to get the current cart
3. Observe the quantity is 1
4. Store the cart Item ID
5. Execute Meteor.call("cart/removeFromCart", <youritemID>, 1);
6. Observe there is no error
7. Execute `ReactionCore.Collections.Cart.findOne()` again (or however you want to look at the db)
8. Observe that the quantity is still 1

Additionally there probably needs to be a check that if the decrement brings the quantity to `<= 0` that the item should just be removed. I also don't see a test for this but I will try to add one if it's missing.

Looking at the method code I can't see why this isn't working. The Mongo call returns a 1 and the values seem to be correct.

zenweasel:
Closes #1003 
Fixes the existing code to decrement cart and add a check if the amount goes below 0.


server/methods/core/cart.js
```diff
@@ -473,26 +473,31 @@ Meteor.methods({
...
-          "error removing from cart");
-        return error;
-      }
-      if (result) {
-        Logger.info(`cart: removed variant ${
-          cartItem._id} quantity of ${quantity}`);
-        return result;
-      }
-    });
+    if (quantity >= cartItem.quantity) {
```
jshimko: We could also remove this conditional on 476 and add the check to line 450 above like:
```js
if (!quantity || quantity >= cartItem.quantity) {
```
...instead of re-calling the method with no quantity arg.  It accomplishes the same thing, but with a slightly easier to read logic flow.

jshimko:
Commented on some minor implementation details, but otherwise 👍

kieckhafer:
👍

kieckhafer:
@zenweasel a couple linting issues are popping up. I fixed one, tried to fix another (line 482) but was seeing a new error show up, either `Inconsistently quoted property` OR `Unexpected token`, depending on what I tried.

jshimko:
@kieckhafer You have to quote object keys that have dashes or periods, so `Inconsistently quoted property` can be ignored.

kieckhafer:
@jshimko  OK. What about the final remaining issue, an expected return in the function?
Does the following need to be outside an if (result) statement?
```
if (result) {
    Logger.info(`cart: deleted cart item variant id ${
        cartItem.variants._id}`);
     return result;
  }
```

kieckhafer:
👍  Looks good

jshimko:
👍

PR #1275, opened 2016-08-09
Fix PayPal PayFlow discounts and refunds

kieckhafer:
This fix confirms PayPal PayFlow refunds are working, fixes discounts, which were not working at all, and adds the ability to give a 100% discount, which effectively voids the authorization.

- [x] Has tests.
- [ ] Has docs.
- [x] Passes all tests


aaronjudd:
👍  as long as you are testing with a unique, new, luhn10 card, such as from http://saijogeorge.com/dummy-credit-card-generator/  this is passing for me.  Errors happen when using previously used cards.

PR #1276, opened 2016-08-10
PDP publish workflow

### LINKED ISSUE #1178: Product detail page editing workflow improvements
Simple publishing workflow for products and variants.

Goals:
- [ ] Allow the admin users to edit a product without having changes showing up immediately for all users.
- [ ] Save any changes to a product as a draft revision in the `Revisions` Collection.
- [ ] Use a "Publish Changes" button to publish all draft changes to a product and all variants with a draft.

mikemurray:
Resolves #1178 

Basic draft / revision workflow for the Product detail page and products in the product grid.

- [ ] Has tests.
- [ ] Has docs.
- [ ] Passes all tests


imports/plugins/core/revisions/server/publications.js
```diff
@@ -0,0 +1,24 @@
+import { Products, Revisions } from "/lib/collections";
```
aaronjudd: Is there a specific reason that Products is imported here?

lib/collections/schemas/revisions.js
```diff
@@ -0,0 +1,61 @@
...
+
+  workflow: {
+    type: Workflow,
+    optional: false
+  },
+
+  documentId: {
+    type: String,
+    label: "Reference Document Id"
+  },
```
aaronjudd: Perhaps "index: true" would be appropriate here?

imports/plugins/core/revisions/server/publications.js
mikemurray: Probably none as I went with doing the revision publication from the products publication.

This file might need a rewrite or removal.

aaronjudd:
Should I be seeing empty change lines when adding details? This occurs after the first time you publish changes with details, prior to publishing you see [Object,Object] (in the change column).
basic_reaction_product

aaronjudd:
Some ui thoughts.. for now or in the future, having warnings display with the publish changes could be helpful.

As a user I felt it could be a little unclear what is published or not, blockers/warnings would be good, but also I'm not sure the color coding on the table tells me clearly enough what the current status of the live product is.

Btw - the pencil icon confused me there.. I was hoping I could edit that change. 

ctlndpqiram9ilmkm

mikemurray:
I think each field / section needs some indication that its been edited, not just that diff in the sidebar, which is a pretty rough summary at the moment. Perhaps thats a colored badge or something else.

aaronjudd:
I was scrolled down in a smaller window and watching the publish view while testing, so I was oblivious to the warnings at the top, so I guess it's always been an issue but now we have somewhere to put them.

imports/plugins/included/product-variant/client/templates/products/productSettings/productSettings.js
```diff
@@ -17,19 +19,34 @@ Template.productSettings.onCreated(function () {
...
       });
-
+      this.state.set("productIds", productIds);
       const products = Products.find({
         _id: {
           $in: productIds
         }
-      }).fetch();
+      }).map((product) => {
+        console.log(product);
```
aaronjudd: Errant console log is spitting the product object out on the browser.

aaronjudd:
Seems to be incompatible with current product structure, if I switch to this branch with existing product data, I get "No Products Found".  This would be a blocker, I think.

imports/plugins/core/revisions/server/startup.js
```diff
@@ -0,0 +1,175 @@
+import { Products, Revisions } from "/lib/collections";
+import { Logger } from "/server/api";
+import { diff } from "deep-diff";
+
+
+Products.before.insert((userId, product) => {
```
aaronjudd: I think hooks should move to `server/hooks/hooks.js` for more consistent implementation of hooks.  Is there a specific reason it's in startup.js?

zenweasel:
This has been moved to v16 so let's reopen when we are ready to review then.

PR #1279, opened 2016-08-10
Fix Authorize.net refunds and discounts

kieckhafer:
- Fix discounts so that they are correctly sent to authorize.net

- Allow 100% discounts by adding the “voidTransaction” function and
voiding transactions

- Update error messaging for Refunds, as we do not (yet) allow them
from Authorize.net, directing the user to go to their auth.net portal to perform refunds - Auth.net has told us there will be a new SDK ready within a few days, which we will support refunds with, so this was the best option for a temporary “No Refunds” flow

- [ ] Has tests.
- [ ] Has docs.
- [x] Passes all tests


aaronjudd:
lots of empty info....
2__reaction_reset____reaction__node_

kieckhafer:
Is this in the right place? This PR will not have Braintree fixed in it, only Auth.net

aaronjudd:
oh yes, sorry, was dragging quickly before I got distracted by something...

Capt-Slow:
Verified the fix.

mikemurray:
👍

aaronjudd:
Browser console error on click of refund popup: ```reactionAlerts.js:86 Uncaught (in promise) TypeError: callback is not a function(…)(anonymous function) @ reactionAlerts.js:86(anonymous function) @ dynamics_browser.js:51```

kieckhafer:
This is fixed in the other open PR (Braintree).


On August 16, 2016 at 1:50:32 PM, Aaron Judd (notifications@github.com)
wrote:

Browser console error on click of refund popup: reactionAlerts.js:86
Uncaught (in promise) TypeError: callback is not a function(…)(anonymous
function) @ reactionAlerts.js:86(anonymous function) @
dynamics_browser.js:51

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/reactioncommerce/reaction/pull/1279#issuecomment-240234150>,
or mute the thread

imports/plugins/included/authnet/server/methods/authnet.js
```diff
@@ -90,7 +92,33 @@ Meteor.methods({
...
+            response: captureResult
+          };
+        } else {
+          result = {
+            saved: false,
+            error: captureResult
+          };
+        }
+      } catch (error) {
+        Logger.fatal(error);
```
aaronjudd: thought more than review:  should this be a `warn` since it's not actually a `fatal` error because we're capturing it?

imports/plugins/included/authnet/server/methods/authnet.js
```diff
@@ -153,6 +188,18 @@ function priorAuthCaptureTransaction(transId, amount, service) {
...
+    refTransId: transId
+  };
+  // This call returns a Promise to the cb so we need to use Promise.await
+  let transactionRequest = service.sendTransactionRequest.call(service, body, function (trans) {
+    return trans;
+  });
+  return Promise.await(transactionRequest);
+}
+
 ValidCardNumber = Match.Where(function (x) {
```
aaronjudd: just a thought for the future, these card validations could perhaps move into the schemas and we validate against schemas rather than this object locally.

imports/plugins/included/authnet/server/methods/authnet.js
```diff
@@ -1,13 +1,15 @@
...
 // meteor modules
+import accounting from "accounting-js";
 import { Meteor } from "meteor/meteor";
 import { check, Match } from "meteor/check";
 import { Promise } from "meteor/promise";
 
 import AuthNetAPI from "authorize-net";
 import { Reaction, Logger } from "/server/api";
 import { Packages } from "/lib/collections";
+import { PaymentMethod } from "/lib/collections/schemas";
```
aaronjudd: curious why we aren't using the plugin specific schema?

import { AuthNetPayment } from "../../lib/collections/schemas";

aaronjudd:
@kieckhafer ok then we can't have approved this PR until the braintree PR is merged, because you can't complete an order currently in this PR. right?

kieckhafer:
You should still be able to complete an order. If you aren't able to, that's a different unrelated issue (that I have not come across).

aaronjudd:
Is this part functional? 
> Allow 100% discounts by adding the “voidTransaction” function and
voiding transactions

kieckhafer:
Yes, 100% discounts are valid, but we don't support refunds at all for Auth.net (until we switch to their new platform which is coming out any day now).

aaronjudd:
👍

PR #1281, opened 2016-08-11
Update circle node

aaronjudd:
maintenance update.
- updates CI node version
- removes bunyan-loggly ( I couldn't see anywhere we're using this right now, and logging seemed to work without)
- removed meteor-eslint as it was enforcing some errors that we didn't agree with and was causing some of the CI testing to fail.


jshimko:
Yeah, the latest version of the Logger code was [copied from Launchdock](https://github.com/reactioncommerce/launchdock/blob/development/server/api/logger.js#L26-L41) and I removed the Loggly lines because we weren't using it.  I guess I forgot to remove it from the package.json.  We can add back later (preferably after they get around to updating that vulnerability).

Approved. 👍

PR #1286, opened 2016-08-12
resolve i18next updates

aaronjudd:
updated package.json:

- updates jquery-i18next
- updates i18next-browser-languagedetector

Resolve Bithound warnings.


jshimko:
👍

PR #1289, opened 2016-08-13
Taxes

### LINKED ISSUE #972: Tax Implementation
Provide a basic flat rate tax implementation package (reaction-taxes) and a provider package (reaction-taxcloud)

Basic steps are likely something like:
- [x] Implement Taxes collection, update schemas as needed. 
- [x] Add hook after shipping, prior to cart review return and save tax rates to the cart
- [x]  Ensure taxes added to cart totals / charges
- [x] Display Tax in review and order dashboards.
- [x] Create as reaction-taxes package
- [x] Create generic method (similiar to payments, shippings) in core for calling specific tax packages
- [x] Add one specific tax vendor (suggest we start with TaxCloud) (additional package?)
- [x]  UI and Settings configurations for Tax API for initial provider
- [ ]  implement tax refund / return order handling
- [ ] Auth / Trans payment hooks for tax auto-filing requirements

# Tax 
We should accommodate these initial taxation types:

 *  Manual / Flat Tax
    * match address fields (ie: state, city, country)
 *  US Nexus Sales Tax
    * implement as taxcloud 
       * requires taxcloud and usps id

In a future release we could add providers:
* VAT / GST (Value Added Tax)
    * implement as avatax 
* Marketplace taxes
   * implement TaxJar marketplace api

## Options

* Include Shipping in tax total (can vary by origination locale)
* Tax by Destination
* Tax by Origination
* Product options
  *  `Tax Included` at product
  *  `Tax Exempt ` at product
  *  `Taxable Goods` product tax category array

## Providers

These seem to be the most common API providers in the US market right now:

* [TaxCloud](https://dev.taxcloud.com/)  (free rates, automated filings)
* [TaxJar] (http://www.taxjar.com/)
* [Avalara](https://www.avalara.com/) (US and EU Vat))
* [Taxify](http://taxify.co/) (automated filings)

Some random notes:

- If you are VAT registered, then the prices displayed on the website should be VAT inclusive.
- Discounts / coupons should be applied before tax calculations.  Perhaps an option?

aaronjudd:
Resolves #972 and provide sales tax rate lookups where an address has been added to the cart.

Custom Tax Rate implementation for region, postal, country rate lookup from Taxes collection.

Initial implementation of Avalara provides rate lookups.

Initial implementation of TaxCloud provides free rate lookups and tax filings for United States.

Independant plugins:

Plugin         | Path
-------------- | ---------------------------------------
reaction-taxes | imports/plugins/core/taxes
taxes-avalara  | imports/plugins/included/taxes-avalara
taxes-taxcloud | imports/plugins/included/taxes-taxcloud
taxes-taxjar   | imports/plugins/included/taxes-taxjar

Note: `taxes-taxjar` is incomplete and disabled.

Currently these methods work if `product.taxable` = true. TaxCode schema and some work has been done for allowing products to have specific taxcodes. This will be a new task.

These plugins all use MethodHooks after "taxes/calculate" to update `product.tax` with a fractional tax rate. They also store the raw tax lookup results in the `product.taxes` array. `product.taxes` is not published to the client, only `tax`. The `product.tax` is used in the cart.transform logic, and remains largely unchanged.

Schemas are provided in the plugins and imported from plugin rather than a core schema.

There is not any logic implemented yet for taxation options, such as including shipping. However the UI exists, currently commented out. We'll need to add logic for handling the rules for each method.

Adds an `ui-grid` component to the core plugins, which provides MeteorGriddle.

Adds `registry/update` method that updates plugin registry settings.

There is not any capture or refund logic, this should be added next.

Dashboard documentation is provided in https://github.com/reactioncommerce/reaction-docs/pull/77


jshimko:
👍 

I feel like I can't possibly have tested every moving part, but all of the functionality I've tested seems to be working for me.

kieckhafer:
Not seeing the taxes displayed in a couple places:

1) We didn't list it here prior, so it makes sense why it's not here, but the totals here don't add up correctly without the taxes displayed:
screen shot 2016-08-16 at 1 53 30 pm

2) We have a display field for taxes, but they aren't registering:
screen shot 2016-08-16 at 1 54 57 pm

aaronjudd:
@kieckhafer looking into the order pages

aaronjudd:
@kieckhafer resolved tax display.
orders

kieckhafer:
Shopping cart drawer:

1) Before I enter my address (first time customer), taxes always display $0.00 in the cart drawer, even on taxed products. I'm not sure if this is a problem, since we can't calculate them, but maybe we put a note saying "Can't be calculated until address is known".

2) After I put my address in, ALL products are being taxed, even those that are not marked as taxable.
taxes mov

aaronjudd:
@kieckhafer line items should now work for all services. Still working on the UI issue in product detail (changes not being saved).

kieckhafer:
👍  taxes rule!

PR #1265, opened 2016-08-08
Fix Braintree discounts and refunds

kieckhafer:
This fix confirms braintree refunds are working, fixes discounts, which were not working at all, and adds the ability to give a 100% discount, which effectively voids the authorization.

- [x] Description explains the issue / use-case resolved
- [x] Only contains code directly related to the issue
- [x] Has tests.
- [ ] Has docs.
- [x] Passes all tests
- [x] Has been linted and follows the style guide

PR #1298, opened 2016-08-15
update cardNumber schemas

kieckhafer:
cardNumber schemas were not allowing valid credit cards, such as Amex
(15 digits), some Visa (13 digits), and foreign Maestro (12 - 19
digits). This updates to allow these lengths to be input. Braintree is left out of this PR, as it is updated inside of #1265.


zenweasel:
👍 I feel like we should document this somewhere, but I don't know exactly where. Should we start a section under "Plugins" for each of the Payment Provider plugins?

kieckhafer:
Yes... We have some docs already in our GH repo, but they aren't live in the docs, I was wondering about that. I will go through them this evening and see what's old info, what's new, and try to update the docs.

Capt-Slow:
Verified and merged.

PR #1299, opened 2016-08-16
Remove bash script from postinstall to fix Windows installs

jshimko:
Solve one more bash script issue for Windows.

`npm install` no longer runs a bash script to load plugins (because `reaction-cli` takes care of it).  The fallback local icon fonts have been moved to the public folder to avoid needing the `cp` command to run on all platforms.

App behavior should be identical as long as you're using `reaction-cli`.


zenweasel:
Approved

aaronjudd:
👍  added update to package version 0.15.0

PR #1302, opened 2016-08-16
Fix Stripe refunds and Double Discounts

### LINKED ISSUE #1233: "Double discount" being applied when Stripe is the payment provider
### Expected behavior

When an order is authorized with Stripe, and a discount is applied to the order, we should apply that discount one time on payment capture.

### Actual Behavior

When an order is authorized using Stripe, if a discount is given before the order is captured, it is being applied a second time, as a "Refund", once the order is captured. From (Stipe's Documentation)[https://stripe.com/docs/api#capture_charge], it seems that this is due to stripe using the authorization as the basis for the charge, and instead of changing the amount that was authorized for the capture, they "refund" the difference - at least on paper that's how they present the data to us.

1) Use Stripe to purchase an item
2) Apply a discount to the order before Capturing the payment
3) Capture the payment
4) See that the discount is applied a second time as a "refund", and the "adjusted total" is actually the Total - (discount x 2), instead of Total - discount.

kieckhafer:
Resolves #1233

The way Stripe works, 'discounts' aren't a thing - to give a discount, you charge the full authorized amount, and immediately refund the said discount. Because of the way we calculate our Adjusted Total, we were seeing discounts applied twice, once as a discount, once as a refund.

In addition, 100% 'discounts' on unauthorized charges weren't allowed, so you couldn't give a 100% discount.

This PR fixes both these issues.

1) When Stripe is the payment provider, discounts are re-added into the "Adjusted Total", so they don't apply twice.
2) To give a 100% discount, we now capture the full authorized amount, and then immediately refund it.

- [x] Has tests.
- [ ] Has docs.
- [x] Passes all tests


zenweasel:
Is this intentionally against `master` instead of `development`?

kieckhafer:
No it's not. I will close this, and opened up #1304 into Development instead.

PR #1307, opened 2016-08-17
Fix wrong method hooks context

priezz:
This PR contains a small change to fix passing the wrong context into `before` and `after` method hooks. With this fix hooks get the same context as the original method, so the goodies like `this.userId` become available for them as well.

As the changes proposed are really small, there are no additional tests/docs.

### DIFF
--- a/server/api/method-hooks.js
+++ b/server/api/method-hooks.js
@@ -83,15 +83,13 @@ MethodHooks._initializeHook = function (mapping, methodName, hookFunction) {
   MethodHooks._originalMethodHandlers[methodName] = method;
 
   MethodHooks._wrappers[methodName] = function () {
-    // Save `this` context
-    const self = this;
     // Get arguments you can mutate
     let args = _.toArray(arguments);
     let beforeResult;
     // Call the before hooks
     let beforeHooks = MethodHooks._beforeHooks[methodName];
-    _.each(beforeHooks, function (beforeHook, hooksProcessed) {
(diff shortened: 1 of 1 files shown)


jshimko:
Thanks for the PR!

I think we might want to consider changing those callbacks to arrow functions so the `self` isn't necessary.  That's what we tend to do everywhere else.  It'd be nice to be consistent.  

Thoughts @aaronjudd @mikemurray?

jshimko:
To clarify...

```js

  MethodHooks._wrappers[methodName] = function () {
    // Get arguments you can mutate
    let args = _.toArray(arguments);
    let beforeResult;
    // Call the before hooks
    let beforeHooks = MethodHooks._beforeHooks[methodName];

    _.each(beforeHooks, (beforeHook, hooksProcessed) => {
      beforeResult = beforeHook.call(this, {
        result: undefined,
        error: undefined,
        arguments: args,
        hooksProcessed: hooksProcessed
      });

...
```

priezz:
Sure, arrow functions will do the thing as well

aaronjudd:
👍  accepting, @jshimko  we can always update/cleanup (now if you want), or when we review hooks overall.

priezz:
Just changed to the arrow function use... Are there any other steps required from my side regarding this PR or I just wait for the merge?

aaronjudd:
@priezz thanks, looks good.

jshimko:
👍

PR #1323, opened 2016-08-23
Release 0.15.1

kieckhafer:
* Meteor 1.4.1

* native implementation bcrypt (meteor npm install --save bcrypt)

* update eslint-2 note

* add .bithoundrc & more bitHound rules

* update to eslint-3

* linting updates (updated rules, enforced constant preferences, update consistent-return rule, use strict: 0 per babel-eslint, warn if not valid-jsdoc, warn on max-len, allowImportExportEverywhere for Meteor’s reify conditional imports, warn on prefer-const)

* updated READAME.md

* disable & remove CodeClimate

* add jquery ignore unused

* add app-test to test scoring (http://blog.bithound.io/scoring-test-files-in-bithound/)

* guard for in loop

* updated var assignments (const / let)

* update imports

* updated for testing (methods updates to pass testing)

* updated fast-render

* updated sweetAlert (fix error “message unknown parameter” from sweetAlert)

* update core label, description

* updated README.md (add CC badge back until we can pass on warnings from BitHound)

* updated packages (move newly added shell-server to “meteor” packages)


jshimko:
👍

PR #1329, opened 2016-08-26
Added babelrc and stage-2 preset

mikemurray:
Added .babelrc file as well as the stage-2 preset.

This will allow the use of some features still in draft, but that are most likely going to be added into the spec.


aaronjudd:
👍

jshimko:
👍

PR #1332, opened 2016-08-27
Remove internal plugin loader in favor of reaction-cli loader

jshimko:
Since the Reaction plugin loader logic is now in `reaction-cli`, the built-in plugin loader is running redundantly on every startup.  But more importantly, it writes to the filesystem in the middle of Meteor building the app and that file change causes Meteor to restart the build in the middle and sometimes obscures the log output that shows the default admin user credentials (which can't be shown again without a database reset).  Ideally, the whole plugin loader logic should be built into a Meteor build tool plugin, but that process was really quite complicated, so that's how we ended up here.  :)  

The only issue this potentially presents is now plugin loading will be dependent on starting the app with `reaction-cli` (which everyone should be doing anyway).  Using the regular `meteor` command wouldn't generate new plugin import files.  So this is not _the_ fix, but it's a fix that solves more than it breaks.  And being able to see the default credentials on first usage is probably more important than anything else here.

So the longer term goal would likely be to move the plugin loader logic into a Meteor build tool plugin so `reaction-cli` isn't required and so hot changes to plugins (like deleting a plugin folder while the app is running) will force a refresh of the plugins imports files.

But in the short term, this fixes the fact that the app almost always builds twice on startup and occasionally hides the default login info from new users.  This should also reduce the build time by at least a few seconds since it's no longer reading/writing all the file system stuff.


aaronjudd:
👍

Issue #1221, opened 2016-07-28
Replace or maintain CollectionFS


aaronjudd:
Goal is to evaluate alternate file uploading packages, or write our own, or take over maintenance of CollectionFS.

Reason is that support for CollectionFS has stopped and the project has been deprecated. 

Current with Meteor 1.4 the inclusion of CFS is producing an error:

```
{ [Error: Cannot find module '../build/Release/bson'] code: 'MODULE_NOT_FOUND' }
 (STDERR) js-bson: Failed to load c++ bson extension, using pure JS version
```

This is identified in the Meteor issue https://github.com/meteor/meteor/issues/7477

The package needs to support:
- MongoDB blob storage
- Alternate, configurable CDN storage. AWS S3 is the minimum requirement.
- Transforms for variety of configurable sized upon upload
- Video
- All image formats
- Metadata storage with images
- multiple file uploads
- import from file system (probably through Reaction.import)

Tasks involve replacement, then updating the image gallery uploader and template image helpers.

zenweasel:
In terms of possible replacements I am putting this here because I keep forgetting it. Suggested by @newsiberian: https://github.com/VeliovGroup/Meteor-Files

aaronjudd:
Here's another Meteor package:  [UploadFS](https://github.com/jalik/jalik-ufs).  I'm only adding because I stumbled across it in a Meteor forum thread. Doesn't seem as mature as Meteor-Files

spencern:
If we build our own solution based on npm packages there doesn't seem to be anything that does both of these things well out of the box.

We could consider using something such as https://www.npmjs.com/package/gridfs-stream (or  https://www.npmjs.com/package/gridfs-locking-stream if we need to support locking) to get started with GridFS support and then using an npm S3 lib (maybe https://github.com/andrewrk/node-s3-client) to support AWS S3.

Would be more work than dropping in another meteor lib, but might be more sustainable in the long run.

spencern:
It looks like the core of https://github.com/CollectionFS has been merged back into a single repo and the individual meteor packages that we are currently using are "deprecated" - though it doesn't appear that the single core package has had any additional work done.

aldeed:
Note to self: Also remove `raix:ui-dropped-event` (the custom "dropped" event) as part of this

aldeed:
Will remove the `Media` collection hooks as part of this, too

PR #1336, opened 2016-08-31
update cfs:gridfs npm dependencies to fix bson error

jshimko:
Since Meteor 1.4, we've been getting the following error on every startup...

```sh
{ [Error: Cannot find module '../build/Release/bson'] code: 'MODULE_NOT_FOUND' }
(STDERR) js-bson: Failed to load c++ bson extension, using pure JS version
```

It doesn't break anything, but it's a performance issue.  It turns out that it's being caused by the outdated npm dependencies in the deprecated `cfs:gridfs` package, so I copied that package into the local packages folder and updated its dependencies.  Problem solved.

See further discussion at https://github.com/meteor/meteor/issues/7477


jshimko:
More specifically, I changed this...

```js
Npm.depends({
  "mongodb": "1.4.35",
  "gridfs-stream": "0.5.3"
});
```

to this...

```js
Npm.depends({
  "mongodb": "2.2.9",
  "gridfs-stream": "1.1.1"
});
```

Otherwise, the package is identical.

aaronjudd:
This is part of the work that is defined in the issue #1221 that is scheduled for 0.17.0.

jshimko:
Yeah, I know.  I just though a temporary/simple solution was reasonable since people are going to continue to ask about the error.  

I'm fine with whatever though.

aaronjudd:
That comment was to ensure that this PR was tracked to #1221.

This patch works (tested), and is something that I considered doing for the last release.  I was probably just being stubborn, but re-polluting the packages directory was the primary deterrent, and because I thought the CFS replacement issue is of sufficient criticality that we should keep the errors screaming at us every day as a non-breaking warning. 

The work is done now though, and now that we've connected the dots,etc  I'm 👍

jshimko:
Yeah, using the packages directory felt a little gross to me too, but so did the ugly error.  But then I convinced myself the temporary fix was probably worth the trouble/ugliness since it shouldn't be there for too long.

zenweasel:
No one asked me, but I too agree that it's worth fixing in the short-term so people don't perceive Reaction as "broken". I think GridFS is a big enough pain that we won't forget it.

zenweasel:
Approved

PR #1338, opened 2016-08-31
persistent profile language

### LINKED ISSUE #890: i18n Selected language doesn't save. Always reset to english after F5
If i'm not mistaken saving selected language was already implemented, but now it doesn't works.
Tested in Safari

aaronjudd:
Refactors the way default languages are stored, moving them to the user profile. Now when user selects a language, the changes are reactive to the `user` collection (rather than Session).  We were only using Session to track before, now we're tracking in `Meteor.user.profile.lang`.  While, it's not generally best practice to use the user.profile, but this seems to be the most transportable location in the user-> accounts flow.  Publications were updated in UserProfile ShopMembers to allow this additional property to be accessed.  Refactors publication so that multiple languages can be subscribed too (but actually subscriptions need to be updated for this).

Ideally I think this will move into the loginHandlers and use the client i18next.language browser detection, but I was having unpredictable results and need to put some more thought into that workflow.  `i18nextBrowserLanguageDetector` now has options and is wired up, but not really fully implemented.  In working on this, it feels as though this should move to `imports/plugins/core/i18n` as a complete plugin in the future as well.  

The behavior now is:

- resolves #890
- new anonymous user is created using shop default language. 
- user selects alternate language, i18n transform occurs
- profile.lang is preserved on reload, new tabs
- user registers new account, retains profile.lang
- user logs out of account, 
	- *new anonymous user created, reverting to shop default*

**Todo/discuss**

- *new anonymous user created on logout, retain user selected language*
- *alternate storage rather than user.profile*


jshimko:
👍

PR #1339, opened 2016-09-01
Release 0.15.2

kieckhafer:
* update cfs:gridfs npm dependencies to fix bson error
* Preserve custom profile fields in Accounts.onCreateUser
* Remove internal plugin loader in favor of reaction-cli loader
* update to Meteor 1.4.1.1
* Added babelrc and stage-2 preset


jshimko:
👍

Issue #1347, opened 2016-09-01
Speed acceptance tests up


Capt-Slow:
Tests right now have a 22min runtime. Fix this.

Capt-Slow:
So after removing a lot of the pauses, it came as no surprise to me that the waitFor functions are proving unreliable. So I am having to troubleshoot a new series of issues/failures.

zenweasel:
Closed via https://github.com/reactioncommerce/reaction/pull/1355

PR #1355, opened 2016-09-06
Josh test speed increase #1347

Capt-Slow:
This is still just running synchronously. As tests grows we will need
to look into running tests in parallel, however this might require some
infrastructure in place, i.e. a hosted QA environment. An environment
that can handle the load of 5-10 tests hitting it at a time.


zenweasel:
Approved

Capt-Slow:
screen shot 2016-09-07 at 10 39 23 am

kieckhafer:
👍  Seeing errors, but legit errors that we need to fix, not the fault of the tests.

PR #1357, opened 2016-09-07
Shipping 1320

### LINKED ISSUE #1320: Edit of Shipping settings is not saved, changes are lost on server restart

### Expected behavior
Changes to Shipping are always persisted
### Actual Behavior
Changes to Shipping are not persisted
I edited  default "Free Shipping" to "Paid Shipping" as a test , on app resart the changes are lost it goes back to "Free Shipping".

aaronjudd:
- resolves #1320 
- fixes a number of UI issues
- refactors shipping methods naming conventions (breaking change)
- updated Reaction.Import exports
- adds Fixture export
- updated Shipping Schema
- updated on Core Init Load Data to use Fixture
- updated SweetAlert2
- disabled un-implemented range inputs


Capt-Slow:
screen shot 2016-09-07 at 1 19 39 pm
screen shot 2016-09-07 at 1 20 48 pm

I noticed when adding price to Cost, Handling & Rate I see what appears to be correct-ish (off by .01) shipping cost in 'Select Shipping option' during checkout '$0.35'. However,  when I look at the 'Review' summary I see shipping cost being '$0.25'.

Capt-Slow:
If a user adds a product to the cart using a shipping option that has been changed or removed while items are still in the cart. That cart session will still use the old shipping method, that may not exist anymore.

aaronjudd:
@Capt-Slow I think what you are describing is expected behaviour. We don't change prices (currently), or shipping, or taxes unless the cart itself has been modified. The

Capt-Slow:
I receive an error when Capturing Payment for AuthNet transaction only when there is a Handling price in the shipping method.

I20160908-10:37:42.956(-7)? 17:37:42.956Z FATAL Reaction: The amount requested for settlement cannot be greater than the original amount authorized.
I20160908-10:37:42.957(-7)?   Error: The amount requested for settlement cannot be greater than the original amount authorized.

imports/plugins/core/taxes/server/api/import.js
```diff
@@ -1,5 +1,5 @@
 import { Reaction } from "/server/api";
-import Import from "/server/api/core/import";
+import { Import } from "/server/api/core/import";
 import * as Collections from "../../lib/collections";
```
zenweasel: I wonder if we shouldn't refactor this to not be something that's also a keyword?

imports/plugins/included/shipping/server/methods.js
```diff
@@ -1,19 +1,26 @@
 import { Shipping } from "/lib/collections";
 import { Reaction } from "/server/api";
 
-Meteor.methods({
-  /*
+export const methods = {
+  /**
```
zenweasel: Is this the standard on how we are doing this now? If so I like it except I have been doing one thing differently in that I have kept the Role/Permission checks in the Meteor method but not in the exported method so that they can be called in code (e.g. hooks) without having a logged in user. I don't feel like we need to check permissions if they are being called from server-side code. Not sure what your thoughts are.

imports/plugins/core/taxes/server/api/import.js
aaronjudd: Agree, that's bugged me for a while.  I created https://github.com/reactioncommerce/reaction/issues/1364 to track.

imports/plugins/included/shipping/server/methods.js
aaronjudd: I guess that it make sense that permissions are included in the wrapper (I've been thinking for a while that we could just extend the Meteor.method with our own Reaction.method and just include the permissions etc at that level.  You have a link to your implementation?

aaronjudd:
@Capt-Slow I believe the issue you discovered with Authorize.net + handling getting errors is resolved now with the updated cart transforms.

kieckhafer:
👍  Verified that shipping details stay saved on reboot

kieckhafer:
👍  #1320 is confirmed fixed, tests pass, looks good

PR #1369, opened 2016-09-12
load default user from env

### LINKED ISSUE #1368: load default admin user with ENV variables
### Expected behavior
From https://docs.reactioncommerce.com/reaction-docs/master/configuration#environment-variables

`_Note: Environment variables will override variables set in settings.json_`

The expected behaviour is that if you set ENV variables it will always use these instead of values in `settings.json`.  This is how it does work if you are using `meteor` instead of `reaction-cli`.

### Actual Behavior

In `run.js` we're adding `--settings  dev.settings` if no _settings.json_ exists, which is fine, but changes older behaviour where we simply never loaded dev.settings -> it was just an example file. 

Because we now are loading, dev.settings, the default user/email/pass are being set to "" and the **env** variables are never used.

aaronjudd:
resolves #1368

### DIFF
--- a/server/api/core/core.js
+++ b/server/api/core/core.js
@@ -223,9 +223,11 @@ export default {
    */
   createDefaultAdminUser() {
     Logger.info("Starting createDefaultAdminUser");
-    let options = {};
     const domain = getRegistryDomain();
+    const env = process.env;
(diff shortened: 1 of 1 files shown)


zenweasel:
Approved. Tested, works as expected

PR #1384, opened 2016-09-14
remove card cursor:pointer

### LINKED ISSUE #1309: Some Dashboard Cards Not working
### Expected behavior
Should navigate to selected card (i.e Inventory)

### Actual Behavior
Some Cards don't navigate to any route

http://g.recordit.co/eQY6F9j9jj.gif

docker reaction:latest fresh install

aaronjudd:
Resolves #1309

### DIFF
--- a/imports/plugins/included/default-theme/client/styles/cards.less
+++ b/imports/plugins/included/default-theme/client/styles/cards.less
@@ -83,7 +83,7 @@
 .rui.card .content {
   .flex(1 1 auto);
   padding: 10px;
-  cursor: pointer;
+  cursor: default;
 }
 
 .rui.card .content .title {


zenweasel:
"resolves" might be too strong a word here ;)

imports/plugins/included/default-theme/client/styles/cards.less
```diff
@@ -83,7 +83,7 @@
 .rui.card .content {
   .flex(1 1 auto);
   padding: 10px;
-  cursor: pointer;
+  cursor: default;
```
mikemurray: Why even had the cursor at all?

aaronjudd: Are you suggesting that you don't think the card should be clickable at all? Just the little > section?

mikemurray: That the `cursor: default` is not necessary if its not overriding anything.

- **For future consideration** -
And looking forward to Bootstrap 4 cards:  http://v4-alpha.getbootstrap.com/components/card/

I feel like maybe the clickable content might be confusing since there are other action buttons in the card as well. Maybe we always force the use of the arrow buttons.

aaronjudd: `That the cursor: default is not necessary if its not overriding anything.` but it is.. because the cards are clickable by default.. the logic to add/remove the anchor is missing from the card component.

mikemurray: Ok then, this is good for now.

mikemurray:
👍

PR #1386, opened 2016-09-15
Check if billing address exists

priezz:
Prevents exception if there is no billing address in the `cart/submitPayment` after-hook.

### DIFF
--- a/server/methods/core/hooks/cart.js
+++ b/server/methods/core/hooks/cart.js
@@ -22,7 +22,11 @@ MethodHooks.after("cart/submitPayment", function (options) {
 
     // create order
     if (cart) {
-      if (cart.items && cart.billing[0].paymentMethod) {
+      if (!cart.billing) {
+        Logger.info("MethodHooks after cart/submitPayment. No billing address after payment! userId:", Meteor.userId(), "options:", options);
+      }
+
+      if (cart.items && cart.billing && cart.billing[0].paymentMethod) {
         const orderId = Meteor.call("cart/copyCartToOrder", cart._id);
         // Return orderId as result from this after hook call.
         // This is done by extending the existing result.


jshimko:
Thanks @priezz!

I definitely agree that we need more useful debugging data in the logs, but I'm not sure that particular formatting of the logger arguments is ideal.  I think we probably need to standardize that kind of stuff so that when we make use of the bunyan log output in the future (for things like Loggly, Logstash, etc.), the JSON is outputs always has the same format.  Right now, there's not really much consistency with that.

Probably even better for a case like this would be implementing a [child logger](https://github.com/trentm/node-bunyan#logchild) (which was on the "some day" roadmap already) so that every log in that method/hook/whatever gets consistent output.  For example, see [how I did it in Launchdock](https://github.com/reactioncommerce/launchdock/blob/development/server/methods/stacks.js#L16-L20).  Every `logger` call in that method has the same args added to it so all that is needed is the standard message string.  

So I'm not saying it needs to look exactly like that.  Just saying I think it'd be useful for us to start discussing and enforcing some logging standards.  Maybe we could even add some wrapper that does arg checks on the Logger methods.  We obviously don't need to have this completely sorted out to accomplish your PR, but I'd definitely like to at least discuss log formatting with @aaronjudd first.

And thanks for the contribution!

priezz:
Hi, Jeremy! Thank you for your feedback. Actually, this PR is more about the `billing` field checking in the line 29, but not about logging itself. So, I kept the format exactly the same as used a few lines above. However I (almost, see below) like your approach  used in the Launchdock and if everyone is happy with it I could definitely log this way. I am going to create another dozen of PRs (small, like this one), so I could follow the proposed logging way there.

What I don't like to do - is copy/pasting, even if it is just 5 lines. What about the wrapper around `Logger.child()`? To be called somehow like that:

```
const logger = Logger.methodChild(this, 'rancher/createStack', arguments);
```

So, for event handlers it would be `eventChild`, etc. What do you think about this shortening?

aaronjudd:
👍   My only comment about this is that this assumes that you have a paymentMethod before you can place an order.  I can think of a variety of cases where that might not be true, but I guess in those cases we should probably require some kind of dummy payment method.

zenweasel:
@aaronjudd Doesn't the existing code already check for a payment method? He is just adding a check whether the billing address exists before checking for a property on it so it doesn't throw an `undefined` I assume.

priezz:
>  just adding a check whether the billing address exists before checking for a property on it so it doesn't throw an `undefined`

exactly

aaronjudd:
👍

PR #1389, opened 2016-09-15
Close cart and scroll to top on payment

priezz:
Close cart and scroll to top on payment

### DIFF
--- a/imports/plugins/core/checkout/client/methods/cart.js
+++ b/imports/plugins/core/checkout/client/methods/cart.js
@@ -65,6 +65,12 @@ Meteor.methods({
           Reaction.Router.go("cart/completed", {}, {
             _id: cartId
           });
+          // Close cart drawer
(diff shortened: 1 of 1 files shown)


zenweasel:
@priezz It would be helpful in the future if you could provide for your PR's:

1. Why you are making this change (An issue is strongly preferred).
2. How we can verify that the code is doing what it's supposed to do by providing us steps to test.

We want to take your changes but we also need to make sure that it doesn't have any unwanted side-effects. That quicker we can do that the quicker we can get your PR's through. Thanks.

zenweasel:
Testing this now it seems like this behavior is masked by the fact that it continues directly to the completed screen which happens before this animation completes. I suppose there's no harm in this but still not sure what problem this is trying to solve.

aaronjudd:
50/50 torn on accepting, but as I can't see that this is fixing an issue. please feel free to re-open if we can identify a problem here...

PR #1390, opened 2016-09-15
Exclude anonymous users from accounts publication

priezz:
Exclude anonymous users from accounts publication when logged in as admin

### DIFF
--- a/server/publications/collections/accounts.js
+++ b/server/publications/collections/accounts.js
@@ -1,5 +1,6 @@
 import * as Collections from "/lib/collections";
 import { Reaction } from "/server/api";
+import _ from "lodash";
 
 /**
  * accounts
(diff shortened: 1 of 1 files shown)


aaronjudd:
👍

server/publications/collections/accounts.js
```diff
@@ -1,5 +1,6 @@
 import * as Collections from "/lib/collections";
 import { Reaction } from "/server/api";
+import _ from "lodash";
```
zenweasel: Import order, npm packages first

priezz: Done

zenweasel:
Approved (once updated)

zenweasel:
Testing this I am seeing this error in the server console:
`Exception from sub Accounts id omBLyCHcGmm5LFP5S TypeError: _.pluck is not a function`
and this error in the browser console
`Error invoking SubsManager.subscribe("Accounts", "RcktvhiypBSDce4tp"):  Internal server error`

aaronjudd:
@zenweasel resolved with a commit direct to development ;-)

_.pluck is deprecated in lodash 4.x and instead uses _.map.   The load order of the npm import probably brought this to light,  as I think _ (underscore) is still being exported by Meteor...?..

priezz:
Right, I think it was underscore which worked for me, not lodash... This commit was put without the development branch was broken at the time of commit. @aaronjudd, thanks for your quick fix!

PR #1392, opened 2016-09-15
Prevent no image being found

priezz:
Prevent no image being found if no image has priority 0. `"metadata.priority": 0` filter will fail to find an image if none has priority 0, so sort is used instead.


zenweasel:
Approved - tested and seems to be working fine

aaronjudd:
👍

aaronjudd:
Although, this is likely to get replaced in 0.17 with new "react" product detail page...

PR #1393, opened 2016-09-15
Make close button work on tag nav bar in mobile mode

priezz:
Make close button work on tag nav bar in mobile mode by auto-clicking close button to make navbar left disappear.


aaronjudd:
👍

PR #1397, opened 2016-09-16
metafields -> details for client visibility

kieckhafer:
Changed “Metafields” to “details” in the search settings panel.

Metafields was confusing to people we were showing the search to


aaronjudd:
👍

PR #1409, opened 2016-09-19
browserstack example

Capt-Slow:
This turned into more then just an example.

New Features:
 - Enabling Browserstack support
 - Test Suite is now configurable, allowing the user to pick and choose through a config what set of test suites to enable/run
 - Test restructure and general organization
 - Added npm run scripts to kick off tests
 - Added optional Allure reporting. Which creates pretty testing metrics
 - Changed logged in tests to register new user and checkout, for each of the payment processors

NOTE: Docs for this can be found on https://github.com/reactioncommerce/reaction-docs/pull/100


kieckhafer:
One "issue" i'm coming across. In my settings.json, my username is "Erik", but these tests require the username to be "admin" or else they fail. There should be an option to add a username into the settings, just like you add your email / pass.

screen shot 2016-09-19 at 3 11 01 pm

Capt-Slow:
@kieckhafer The hardcoded assertion should be fixed now.

zenweasel:
The `elem_map` variables should all be changed to camel case. (including `elem_map`)

tests/acceptance-tests/lib/basic-user-actions.js
```diff
@@ -3,11 +3,13 @@ const yaml = require("js-yaml");
 const fs   = require("fs");
 
 const eleMap = yaml.safeLoad(fs.readFileSync("./tests/acceptance-tests/elements/element-map.yml", "utf8"));
+const usrData = yaml.safeLoad(fs.readFileSync("./tests/acceptance-tests/config/user-data.yml", "utf8"));
+const browserConfig = yaml.safeLoad(fs.readFileSync("./tests/acceptance-tests/config/settings.yml", "utf8"));
+
```
zenweasel: Can we just convert these either to JSON or just JS Objects to eliminate the need to load YAML?

Capt-Slow: I suppose we could. I will look into getting a fix in for that. I'll change the `elem_map` variables too while I'm at it.

Issue #1402, opened 2016-09-19
Visibility glitch with Search results


zenweasel:
### Expected behavior
Results should scroll under search bar

### Actual Behavior
Intermittently search results scroll on top of search bar and obscure it

### Steps to Reproduce the Behavior
1. Use current `development` branch
2. Navigate to "Basic Reaction Product" detail page
3. search for "basic"
4. Scroll results above search bar then scroll back down. Observe this glitch:

<img width="987" alt="example-product" src="https://cloud.githubusercontent.com/assets/72819/18620736/fe9c72a0-7e4a-11e6-8c8b-1ab6790dc43a.png">

5. In addition, if you are already on a product page, the underlying page seems to mix with results in a way that's confusing. This might be because I am a relatively small monitor. (15" MBP)

3Maestros:
Even I am seeing this exact behavior.

kieckhafer:
I cannot replicate on my machine, but I have seen this error in action on another machine, and think we have a fix in the branch (search-ui-updates)[https://github.com/reactioncommerce/reaction/tree/search-ui-updates].

kieckhafer:
This is fixed in PR #1410. Closing this issue.

zenweasel:
Closed via #1410

3Maestros:
Awesome confirmed this is fixed, thank you!

PR #1410, opened 2016-09-19
Search UI updates

kieckhafer:
Various UI updates to the ui-search package

- New button component for search icon
- Ability to use ESC to close modal even when input is not focused
- Fix for #1402 - Visibility glitch with search results


zenweasel:
Verified this fixes #1402

mikemurray:
👍  Looks like its fixed

PR #1417, opened 2016-09-22
Renamed field_group to fieldGroup to be consistent with other files

### LINKED ISSUE #1416: Fix inconsistent file name convention in UI package
### Expected behavior

No underscores in file names.
### Actual Behavior

Underscore in `field_group.js`
### Why this is Important

Because its inconsistent. And when importing files from the UI package directly, it will make users have to guess the format of the name.

mikemurray:
Resolves #1416 

Renamed field_group to fieldGroup to be consistent with other files.

We should avoid adding `yet-another_fileNaming.convetion` if we can help it.


aaronjudd:
👍

PR #1432, opened 2016-09-25
Prevent error in `cartPayerName` helper if buyer name contains non-latin characters.

priezz:
Prevent error in `cartPayerName` helper if buyer name contains non-latin characters. The latter are not accepted by payment gateways, resulting the error in the payment form. See the discussion in #1388.


aaronjudd:
👍

PR #1449, opened 2016-09-28
Fixes #1337

### LINKED ISSUE #1337: StripeApi.methods.createCharge returning null
`StripeApi.methods.createCharge` in `stripe/server/methods/stripe.js` is always returning null.

This should be returning the charge result.

Error handling is also broken, because `chargeResult` is null.  See related #1117 

This is causing failing tests in #1315 as well.

joseconstela:
Fixes incorrect promise causing it to return null


aaronjudd:
👍

PR #1461, opened 2016-10-03
Set submitting state to handle error case correctly

### LINKED ISSUE #1459: Example Payment/Stripe Uncaught ReferenceError
In checkout, the example payment causes an `Uncaught ReferenceError` when submitted with invalid fields.

### Expected behavior

In checkout, the example payment should validate and handle the error event, letting the user update invalid fields.

### Actual Behavior
`example.js:95 Uncaught ReferenceError: submitting is not defined`

cart_checkout

After this the inputs are disabled as well and the user cannot update invalid inputs.

Checkout and submit an example payment without required fields.

zenweasel:
Closes #1459


jshimko:
👍

aaronjudd:
👍

Issue #1420, opened 2016-09-23
Reaction docker images should run without MONGO_URL


aaronjudd:
You should be able to run Reaction from a Docker image (like when you use Kitematic) without supplying a mongo url.

zenweasel:
@jshimko Was this fixed in #1468?

jshimko:
yessir.

PR #1468, opened 2016-10-04
Docker build improvements

jshimko:
- clean up unused or unnecessary deps in Docker build
- install all npm deps in first Meteor build step (remove --production)
- create new base Docker image for better build caching
- install MongoDB inside app container and use it if no external MONGO_URL is defined
- move deps versions to Dockerfile for easier update maintenance (instead of having to update multiple places)
- fix custom Meteor package file script
- add build script for full build of base image and app image
- fix lodash/camelcase import that was breaking the production build
- npm version updates
- update CircleCI scripts to use new Docker build

This creates a new base base image with all dependencies installed (other than Meteor), so builds are much faster after the first time.  It also includes Mongo, so you can run the container without specifying an external database.  To test the new build locally...

**Build**
```sh
.reaction/docker/build.sh
```

**Run**
```sh
docker run -d -p 80:80 reactioncommerce/reaction:latest
```

Since `reactioncommerce/base:latest` is published, the normal build process will be identical to the current one.  The above build script is just for building both from scratch in development.

```sh
# build 
docker build -t reactioncommerce/reaction:latest .

# run
docker run -d -p 80:80 reactioncommerce/reaction:latest
```


jshimko:
Build/run instructions added to initial PR comment.

aaronjudd:
👍  tested build and run w/o db.

aaronjudd:
PR resolves #1420

zenweasel:
👍 Tested, works.

Issue #1455, opened 2016-09-30
[PDP] Revisions UX updates


aaronjudd:
Implement remaining UX update for https://github.com/reactioncommerce/reaction/pull/1356
- [x] move revisions dashboard card into Catalog card  (and i18n update) **PR** #1544
- [x] updated "Loading" spinner (should be universal) **PR** #1471
- [x] UX move delete button to same row (shouldn't need row to be active to delete)
- [x] UX to collapse accordions, focus edit target and open **PR** #1569
- [x]  Reimplement a confirmation UX when updating inputs (formerly turned green and fade on success) **PR** #1558
- [x] Ensure all inputs have placeholders **PR** #1472, #1542

zenweasel:
- [ ] Some fields like details and manufacturer and not submitted when pressing enter while others like Tag are. Probably any field that makes sense to be submitted by enter should be.

aaronjudd:
#1471 did not close #1455 (loading indicator for PDP).  I am still seeing the text loader.

![example-product_and_reactioncommerce_reaction_and_inbox__14_759_messages__3_unread_](https://cloud.githubusercontent.com/assets/439959/20197926/c36085ca-a756-11e6-9f59-28f6869adeda.png)

mikemurray:
Different loading issue, as this one is with reaction layout.

PR #1471, opened 2016-10-05
Loading 1455

mikemurray:
Part of #1455

Uploaded loading component to use already available spinner component.

Renamed loading component from `loading.jsx` to `loading.js`. The JSX extension is not necessary.

Added a basic circular progress component, with indeterminate spinning animation.

Updated loader for product detail container.

Removed domkit component. If the need arises to do CSS in JS please consider using React Radium instead.(https://github.com/FormidableLabs/radium).


aaronjudd:
👍

PR #1472, opened 2016-10-05
Added placeholders product fields

mikemurray:
Part of #1455 

Added placeholders to inline product fields.

Added placeholders to some inputs in the product detail admin.

**Note:** Did not add placeholders for social message fields. Need to discuss how we do placeholders when we have labels, or if they're even necessary.


aaronjudd:
👍  verified resolved.  Probably unrelated, inputs don't seem to have placeholders on the variant details form (but they do on children).

Issue #1458, opened 2016-09-30
[PDP]  publish workflow and changes UX


aaronjudd:
This is a task from https://github.com/reactioncommerce/reaction/pull/1356

The current publish workflow and "changes" is confusing and it's not always clear how it works. 

This is an an issue to propose some design changes to the UX and implement.

PR #1499, opened 2016-10-17
Publish 1490

### LINKED ISSUE #1490: [PDP] Publish Controls & Product Details
- [x] Move publish controls to a more visible location on the Product Detail Page so users that make changes will have a better indication that they need to publish their changes after editing.
- [x] Remove publish controls from the admin sidebar for the moment.
- [x] Make the "Product Settings" sidebar a "Product Details" to make it apparent that its the details your changing for the product, not settings for the view.
- [x] Move delete and visibility buttons to a better location in sidebar
- [x] Hide "show changes" button and UI for as it's a determent in it's current form.

mikemurray:
Resolves #1490 #1458

Adds a toolbar to the new product detail page that includes controls for publishing and view switching.


aaronjudd:
I thought I saw the weird behavior of public/private once, but then I didn't see it again.  Really the workflow is pretty nice and make sense to me now.  I did experience one weird issue though, I deleted a product (which should probably be "Archived" a product).  Then I published changes to the "deleted" product, which causes some strange things to happen (lost the variants because they weren't also published).

mikemurray:
I agree, that delete product should probably be "Archive" instead, since that's what you're actually doing.

New variants probably wouldn't be saved since there is no published version to update, archive or delete.

mikemurray:
The same thing happens to me public / private bug. Sometimes it shows up, most times its fine. Not sure what causes it.

aaronjudd:
👍  approved and merging created issue #1503 for private/public debugging, and #1502 for "Archive" labeling.

PR #1501, opened 2016-10-18
New Email Template for Order Confirmation

kieckhafer:
First of the new batch of email templates.


aaronjudd:
👍  Best email we've ever had. ;-)

server/methods/core/orders.js
```diff
@@ -1,8 +1,11 @@
+import path from "path";
+import { getSlug } from "/lib/api";
```
jshimko: Import order standards...

1. npm
2. meteor
3. local

https://docs.reactioncommerce.com/reaction-docs/master/styleguide

server/methods/core/orders.js
```diff
@@ -328,8 +331,84 @@ Meteor.methods({
...
     this.unblock();
 
+    // Get Shop information
     const shop = Shops.findOne(order.shopId);
-    const shipment = order.shipping[0];
+    const shopContact = shop.addressBook[0];
+
+    // Get shop logo, if available
+    let emailLogo;
+    if (_.isArray(shop.brandAssets)) {
```
jshimko: Admittedly super nitpicky... but in general, we can/should probably use the native [`Array.isArray()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray) for `Array` type checks.  Surely we're only talking about microseconds of performance gain, but it's probably still a good best practice to adopt.

Aside from that, it looks like you weren't importing `lodash` anyway (which means the global `underscore` that Meteor forces on us was being used and is the only reason this didn't throw an error).

jshimko: `_.find()` is used below as well.

PR #1504, opened 2016-10-18
Rename settings file so it's more clear

zenweasel:
I get this question all the time of "I changed my settings in dev.settings.json but it's not working". Hopefully this makes it more clear?


aaronjudd:
This would need updates made to the cli and to the docs as well, I think.

PR #1513, opened 2016-10-20
Change "Delete" to "Archive" in PDP

### LINKED ISSUE #1502: [PDP]  Use "Archive" label instead of "Deleted"
Replace "Deleted" with "Archive" in the product "delete" functionality.  (as the product isn't really deleted, Archive is more correct.)

zenweasel:
Resolves #1502


imports/plugins/included/product-variant/client/templates/products/productGrid/controls.html
```diff
@@ -13,8 +13,8 @@
 
       {{#if product.isDeleted}}
         <div>
-        <span class="badge badge-danger" data-i18n="app.deleted">
-          <span>Deleted</span>
+        <span class="badge badge-danger" data-i18n="app.archived">
+          <span>Archived</span>
```
aaronjudd: Does this work? shouldn't `data-i18n="app.archived"` be on the inner span?

private/data/i18n/en.json
```diff
@@ -355,6 +359,8 @@
         "restoreProduct": "Restore Product from Trash",
         "deleteThisProduct": "Delete this product?",
```
aaronjudd: We should consider removing this phrase, as it's not used anywhere now.

imports/plugins/included/product-variant/client/templates/products/productGrid/controls.html
mikemurray: This works. The inner span is because Blaze and jQuery fight each other for who has the right to update the text contents of an element. The `<span>` around the `Archived` prevents this from happening. Don't know why this is... but this solution seems to fix that issue.

mikemurray:
Still tagged as deleted. They are archived as well.

screen shot 2016-10-20 at 2 13 38 pm

zenweasel:
I didn't change these on purpose because I thought we were still deleting variants. If I accidentally add a variant I can't immediately delete it?

mikemurray:
tl;dr No. 

Since variants are products, its all covered by the new revision control system, including all product hooks.

Meaning, it would just be published it as "deleted". There is a collection hook that blocks all deletion attempts for all product types unless you bypass the hook with `_direct`. That hooks assigns `isDeleted: true` instead.

mikemurray:
The variant might no be visible on the product after you delete/archive it. but it's still around in database.

This is a case where we need some sort of list view for all products in the dashboard so we can see what's active / archived / what-ever-other-status

zenweasel:
Ok. Out of scope for this PR obviously but maybe we could so a check that if something has no orders and has never been added to cart it could actually be deleted. I think we had talked about this at some point.

zenweasel:
Should be reflecting Archived for variants as well

example-product

mikemurray:
👍 

looks good to me!

aaronjudd:
👍  If you archive / delete a variant, where on the product grid was previously showing a range, and you leave only one item, the grid is still showing price range based on variants (that are now archived). This might be related to #1519, but I'll create a new issue.  Otherwise from just a review of "does deleted now say archived", yes, this is accepted.

Issue #1515, opened 2016-10-20
Ability to modify transactional email templates via the dashboard


kieckhafer:
An expansion of #423 and #1429, this ticket concerns adding the ability to edit transactional emails from inside the dashboard, by admin users.

This ticket should include work that:
- [x] Creates a new schema for email templates
- [x] Takes the file-system HTML email templates and loads them into the correct Collection
- [x] Creates a raw HTML editor (text-area) inside the dashboard that allows templates to be changed and saved into the collection. Fields should include:
  - [x] Email Subject
  - [x] Email Body
  - [x] Email Language (en is default)
  - [x] Title
- [x]  Checks to see if a template is inside the database, and if so uses that Template. If it's not inside the database, the file-system email will be the fallback.

PR #1522, opened 2016-10-24
Add templates to database

mikemurray:
Related To: #1515 

Adds emails that are on disc in `private/email/templates/` into the `Templates` collection in the database.

Fetches email template from database when it exists using `Reaction.Email.getTemplate()`.


kieckhafer:
👍 

Looks great... updated the email in the DB, changes were seen. Deleted the Template from the DB, and the file-system version was used.

PR #1523, opened 2016-10-24
Admin Invite & Password reset emails

kieckhafer:
Email templates for shop invitations from an admin, and password resetting


server/api/core/accounts/password.js
```diff
@@ -47,16 +49,65 @@ export function sendResetPasswordEmail(userId, optionalEmail) {
...
+    passwordResetUrl: Accounts.urls.resetPassword(token),
+    user: user
+  };
 
   return Reaction.Email.send({
     to: email,
     from: Reaction.getShopEmail(),
-    subject: `${shopName} - Reset your password`,
-    html: SSR.render("resetPassword", { shopName, user, url })
+    subject: `${dataForEmail.shopName}: Here's your password reset link`,
```
aaronjudd: We need to fetch/store these in the Templates collection, along with i18n language.

PR #1526, opened 2016-10-24
only register routes for enabled packages

callmephilip:
# Expected behavior

Once a package is disabled, any routes associated with it should **not** be registered any more. 

# Actual behavior

Routes associated with a package get registered regardless of whether the package is enabled or not


aaronjudd:
👍

PR #1527, opened 2016-10-24
do not use revisions if the package is disabled

callmephilip:
# Expected behavior

When I disable Revisions package, I expect revisions to not be used any more

# Actual behavior

Revisions are still applied


aaronjudd:
👍

aaronjudd:
@mikemurray can you review, is this actually the only place we'd need to check if the package is enabled?

mikemurray:
👍

PR #1528, opened 2016-10-25
update const to conform to new file name

kieckhafer:
Will take care of `WARN Reaction: Template not found:
accounts/reset_password. Falling back to coreDefault.html` error


aaronjudd:
👍  (approved and merged)

PR #1532, opened 2016-10-26
When building an accountsearch collection, actually build it.

### LINKED ISSUE #1531: buildAccountSearch does not build an Account Search
### Expected behavior

On launch the AccountSearch collection should be created
### Actual Behavior

AccountSearch is not created

From a fresh clone launch the app. Every time you will launch the app you will get the message "AccountSearch collection (re)built successfully" but the AccountSearch collection is empty. Because it checks for an empty AccountSearch on each launch. Every time you relaunch the app you will get that message.

Not sure when this bug got introduced since this used to work.

zenweasel:
Resolves #1531


aaronjudd:
👍

Issue #517, opened 2015-11-12
PDP Package.layout implementation


aaronjudd:
The goal of this feature is to allow repositioning (layout) of functional blocks on the PDP page, as well as the addition, removal and permission based viewing of blocks as necessary.  

We'll need to refactor the Product Detail Page to use `ReactionTemplates`, using `Packages.layout` entries to define different product types and associate layout and templates with the PDP based on the product's `Product.type` entry.

This will make the PDP less rigid, and makes implementation much simpler for  #152, #343, #330, #191, #368, #367, #443, #500, #516 , #225.

ProductDetail.html will become a layout loader page, essentially just:

```
{{#each reactionTemplate provides=product.type}}
    {{> Template.dynamic template=template}}
{{/each}}
```

and we add a new `layout` entry in `register.js` for the `simple` PDP layout:

```
 {
    template: "simpleProduct",
    label: "Simple Product",
    provides: "simple",
    audience: ["guest", "anonymous"]
  }
```

assuming `type` = `simple`, and where we've created a new product layout template:

```
<template name="simpleProduct">
  ...
  <div class="col-sm-7 pdp-right-column">
    {{#each reactionTemplate provides=product.type container="pdp-right-column"}}
        <div class="row">
            {{> Template.dynamic template=template}}
        </div>
    {{/each}}
    </div>
...
<template>
```

the individual blocks get moved into their own templates:

```
<template name="productDescription">
   <div class="product-description">
     {{> fieldComponent field="description" value=description type="textarea"}}
   </div>
</template>
```

The `register.js` will get a new `layout` entry for each PDP template:

```
 {
    template: "productDescription",
    label: "Description",
    container: "pdp-right-column",
    audience: ["guest", "anonymous"],
    position: "2"
  }
```

aaronjudd:
@mikemurray have you considered this during your updates.. let's discuss...

PR #1536, opened 2016-10-28
PDP Package Layout - ReactionLayout

mikemurray:
Related Ticket: #517 

This PR represents the first step to rendering any page or view in the app, through package registry entries, for React components; starting with the Product Detail Page.

**Goals this PR tries to achieve:**
- Add Ability to create a any layout with a valid JSON object from the database.
- Render the Product Detail Page using ReactionLayout with permission checks
- Ability to properly register custom react components to use with the new React based ReactionLayout renderer
- Ability to create registry entries for custom Product Detail Page templates

---

**Testing**
- Verify that the Product Detail Page functions the same as it used to.


kieckhafer:
👍  PDP functions as it did

PR #1541, opened 2016-10-31
Docker security updates

jshimko:
Meteor 1.4.2 attempts to enforce not using root to build the app and tanks the Docker build in some cases (not totally sure why it doesn't always happen).  This refactor fixes the build error, corrects permissions in the production bundle, and uses [gosu](https://github.com/tianon/gosu) to run the app as a non-root user (which is highly recommended for security anyway)

Commit details...

- refactor Docker build to run app as a non-root user
- cleanup the cleanup scripts
- remove app source after build is done
- don’t install mongos if mongo install happens in container
- update container port in docker-compose example file
- add a delay option to the Docker entrypoint
- remove unused script from CircleCI
- fix custom package file script

To test...

```sh
reaction init -b docker-security-updates docker-test

cd docker-test

.reaction/docker/build.sh

docker-compose up -d

# app available on http://localhost
```

Note that non-root users can't run processes on port 80, so the container now exposes port 3000 instead.  The `docker-compose.yml` that gets used above takes care of mapping port 3000 to port 80 and you can do the same thing with `docker run -p 80:3000 ...`

### DIFF
--- a/.reaction/docker/scripts/build-meteor.sh
+++ b/.reaction/docker/scripts/build-meteor.sh
@@ -1,30 +1,35 @@
 #!/bin/bash
 
 #
-# builds a production meteor bundle directory
+# builds a production meteor bundle
 #
 set -e
-
-printf "\n[-] Building Meteor application...\n\n"
-
-: ${APP_BUNDLE_DIR:="/var/www"}
-: ${BUILD_SCRIPTS_DIR:="/opt/reaction"}
 
 cd $APP_SOURCE_DIR
 
 # Customize packages
-bash $BUILD_SCRIPTS_DIR/build-packages.sh
+$BUILD_SCRIPTS_DIR/build-packages.sh
 
 # Generate plugin import files
-bash $BUILD_SCRIPTS_DIR/plugin-loader.sh
+printf "\n[-] Running Reaction plugin loader...\n\n"
+$BUILD_SCRIPTS_DIR/plugin-loader.sh
 
 # Install app deps
+printf "\n[-] Running npm install...\n\n"
 meteor npm install
 
-# build the source
+# build the production bundle
+printf "\n[-] Building Meteor application...\n\n"
 mkdir -p $APP_BUNDLE_DIR
-meteor build --directory $APP_BUNDLE_DIR
-cd $APP_BUNDLE_DIR/bundle/programs/server/ && meteor npm install --production
+meteor build --unsafe-perm --directory $APP_BUNDLE_DIR
+
+# run npm install in bundle
+printf "\n[-] Running npm install in server bundle...\n\n"
+cd $APP_BUNDLE_DIR/bundle/programs/server/
+meteor npm install --production
 
 # put the entrypoint script in WORKDIR
 mv $BUILD_SCRIPTS_DIR/entrypoint.sh $APP_BUNDLE_DIR/bundle/entrypoint.sh
+
+# change ownership of the app to the node user
+chown -R node:node $APP_BUNDLE_DIR
--- a/.reaction/docker/base.dockerfile
+++ b/.reaction/docker/base.dockerfile
@@ -1,7 +1,7 @@
 FROM debian:jessie
 MAINTAINER Reaction Commerce <admin@reactioncommerce.com>
 
-ENV NODE_VERSION "4.6.0"
+ENV NODE_VERSION 4.6.1
 
 # Install MongoDB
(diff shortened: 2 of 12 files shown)


jshimko:
FYI, this PR was pretty much entirely motivated by this issue after we updated to 1.4.2...

https://github.com/meteor/meteor/issues/7959

zenweasel:
This is failing for me on my machine so maybe someone else want to try it approve if it works for them? @mikemurray? @kieckhafer?

aaronjudd:
@jshimko the build succeeds for me, and all is good.  However, should we just wait for Meteor 1.4.2.1 with this patch: https://github.com/meteor/meteor/pull/7975/commits/e4acc36f63f98243237b5e6b3e46b083822b95fd 
Or should we move forward with this anyways..?

jshimko:
I'm all for merging it now.  As long as it stays synced with the `.meteor/release` file, it should work fine (and fixes some build issues that came with 1.4.2)

aaronjudd:
👍

jshimko:
👍

callmephilip:
This is failing both on my local machine and our CI with the following message:

```
You are attempting to run Meteor as the "root" user. If you are developing,
this is almost certainly *not* what you want to do and will likely result in
incorrect file permissions. However, if you are running this in a build process
(CI, etc.) or you are absolutely sure you know what you are doing, add the
`--unsafe-perm` flag to this command to proceed.
```

Docker build is run using the following command:

```
docker build -t bakery/shop .
```

.reaction/docker/scripts/build-meteor.sh
```diff
@@ -1,30 +1,35 @@
...
+printf "\n[-] Running npm install...\n\n"
 meteor npm install
 
-# build the source
+# build the production bundle
+printf "\n[-] Building Meteor application...\n\n"
 mkdir -p $APP_BUNDLE_DIR
-meteor build --directory $APP_BUNDLE_DIR
-cd $APP_BUNDLE_DIR/bundle/programs/server/ && meteor npm install --production
+meteor build --unsafe-perm --directory $APP_BUNDLE_DIR
```
callmephilip: this looks suspicious. since the assumption is that this is run in non root mode, why would you need ```--unsafe-perm```? unless i am missing something

jshimko: The app is built as root because there were a lot of random issues in the build process (depending on 3rd party packages a user may have).  As soon as the build is done, the permissions of the entire directory are changed to non-root and the app is started as that same user.  

Building as non-root would sometimes fail because of permissions issues caused by some npm packages and building as root would fail because Meteor 1.4.2 won't allow it without that --unsafe-perm flag.  I assure you, this was the least awful solution until Meteor releases 1.4.2.1 with a better answer for --unsafe-perm (already fixed, just waiting for release).  

In the end, this is accomplishing the exact same thing while avoiding fussy permissions on some module builds.

jshimko: @callmephilip Have you tried running `.reaction/docker/build.sh`?  The image is two parts now, so I recommend you switch to using that script so you definitely have the right stuff in the base image.

jshimko:
@callmephilip Have you tried running `.reaction/docker/build.sh`? The image is two parts now, so I recommend you switch to using that script so you definitely have the right stuff in the base image.

callmephilip:
@jshimko yeah, i just switched to ```.reaction/docker/build.sh``` and it seems to be working - waiting for the CI server to finish build. i've also noticed that container ports changed. does this mean  ```docker run``` will need to be updated with a new port mapping ```-p 80:3000```?

jshimko:
Yes.  As a security measure, non-root users can't run processes on ports lower than 1024 on Linux.

https://www.w3.org/Daemon/User/Installation/PrivilegedPorts.html

.reaction/docker/scripts/build-meteor.sh
callmephilip: Ok. Just to clarify - the build does happen as **root** while the container is run in a non privileged mode?

jshimko: Yes, the entire build process is done by root.  Then the resulting built app is `chown`'d to the non-root `node` user ([here](https://github.com/reactioncommerce/reaction/blob/master/.reaction/docker/scripts/build-meteor.sh#L34-L35)).  When the container gets started, the `node` user is runs the process.  All of the security benefits, none of the permissions hassles.

Issue #1456, opened 2016-09-30
[PDP] Revision images


aaronjudd:
Task from https://github.com/reactioncommerce/reaction/pull/1356

Currently in the PDP revisions do not control images. If you delete (or update) an image but haven't published the image, it's deleted in the published product as well (and it should not).

We can use the CFS metadata to store revision / publish state for the images.

zenweasel:
Closed by PR #1555

PR #1555, opened 2016-11-07
Revision control for images

zenweasel:
Resolves: #1456 
As it says, adds revision control to images. As with the other revision the main cruz of what is happening in `hooks` where the changes are deferred and moved into a revision record.

I made some schema changes to accommodate changes coming from other than the Products collection.

The biggest challenge was that the `Media` collection does not act like other collections because it is not a regular collection. Specifically it does not seem to allow attributes to be assigned (e.g. adding the `__revisions` property like is done for Products`) so assigning this had to happen on th client-side.

How to test:
Add/remove/edit images. You should not see these changes published until you hit "Publish Changes". When there is a pending change, you should see the "edit" icon on an image. (the only thing you can really "edit" about an image is the sort order)


mikemurray:
When you remove an image the edit icon is blank. Otherwise it seems to work as intended.

screen shot 2016-11-07 at 9 22 00 am

kieckhafer:
Seeing the same thing as mike with the blank edit icon, but good aside from that. 👍

kieckhafer:
Merging this into release-0.18.0, with errors above, which need to be fixed.

PR #1556, opened 2016-11-07
Discounts

aaronjudd:


kieckhafer:
Partial implementation: discount, ui, and payment refactoring

👍

PR #1559, opened 2016-11-08
Minor docker build updates

jshimko:
- update gosu to 1.10
- remove redundant `chown` in the entrypoint that was causing slow start time
- move Mongo process into the bg and use stdout for logs instead of using `--fork` and a logfile
- update Node to 4.6.1 on CircleCI


aaronjudd:
👍 

Related to #1541

zenweasel:
Approved

PR #1561, opened 2016-11-10
Allow for disabling/overriding of search modal from a plugin

### LINKED ISSUE #1399: disabled search is displaying
I think the search icon should check if the registry entry is enabled / disabled (like ReactionApps and ReactionTemplates helpers).  Maybe should be using ReactionApps?

Should not appear, and should use the appropriately enabled search if it's another package.

zenweasel:
Resolves #1399 

Provide a registry setting providing "ui-search" and a template name and that will be used. Don't provide it or don't enable one? No search button.


imports/plugins/core/ui-navbar/client/components/navbar/navbar.html
```diff
@@ -6,8 +6,9 @@
     <div class="menu">
       {{> tagNav tagNavProps}}
     </div>
-    
-    <div class="search">{{> React IconButtonComponent}}</div>
+    {{#if isSearchEnabled}}
```
aaronjudd: Since we are still using Blaze here, why not use 
```
  {{#each reactionApps provides='ui-search' enabled=true}}
  {{#each}}
```

zenweasel: Because I already needed to query on the db to get the template and I thought it made the code there easier to read.

aaronjudd: Reaction Apps returns the template, as well as i18n values etc. 
```
Reaction.Apps({provides: "ui-search"}).template;
`

zenweasel: Changed to use `Reaction.Apps`

aaronjudd:
👍 conditionally approved.  The issue here is that all the components could use Reaction.Apps, not just the icon, and a shortcut registry entry for the icon.  The implementation as it stand now, is just a boolean visibility, versus the ability to have multiple potential templates and display only active ones.  We'll need to review this more, but @mikemurray will hit this in update the nav bar to reaction components and the template registry.

PR #1563, opened 2016-11-11
[WIP] Fix Template Loading Issues

kieckhafer:
Templates were not loading on the initial startup of Reaction due to the shopId not yet being created.

This fix adds a _sleep function to make sure shopId is already created before the Templates are inserted.


aaronjudd:
The onCoreInit hook event runs after shops are loaded in `server/startup/load-data.js`.  If you take a look at `server/startup/i18n.js`, we're using the `Assets` collections as fixture for data that is loaded before Shops are created, and which are the processed as imports after the hook event has been triggered.  Using this pattern would allow you avoid the _sleep (which I don't think will be reliable), and also could be considered as an alternate place to store the default templates -> as this is already more or less fixture data in the Assets collection.

kieckhafer:
#1583 replaces this PR.

PR #1564, opened 2016-11-11
Fix remove icon not displaying / revision change

zenweasel:
Fix a couple minor bugs in previous PR

 - Don't show the remove button if item has a pending remove
- only publish changes for product if the product exists


aaronjudd:
👍

kieckhafer:
👍

PR #1570, opened 2016-11-11
Loading component added for reactionLayout

mikemurray:
see https://github.com/reactioncommerce/reaction/issues/1455#issuecomment-259834092


kieckhafer:
Looks good, seeing the loading circle now instead of text.

aaronjudd:
👍

PR #1572, opened 2016-11-11
PDP 500

### LINKED ISSUE #500: Products without Variants
Is there currently a way to post products that don't have Variants? Looks like it's the default (and required) method for posting, however, I can see many shops _not_ needing this functionality.

It seems a bit odd for a user to have to see the "Options/Variations" section if there's only 1 variant.

Perhaps variant fields could be not required, and only show on the front-end if an admin has filled them out?

mikemurray:
Resolves #500

After reviewing issue #500, I have determined that a product cannot be without at least one variant. It is simply hard coded into the cart that a product, using the `ProductVariant` schema, must be supplied to the cart items. That immediately invalidates any product using the `Product` schama, which is actually used as a display wrapper around `ProductVaraint`. 

tl;dr ProductVariants are the product you buy, Product is just for informational detail page.

This PR is mostly UX to help make a product more presentable when there is only one top level variant.

- Hide available options divider, if there are no available options.
- Add an add (+) button to the PDP page when all variants have been removed.


zenweasel:
I'm seeing the Add Option button even when not signed in as an admin.
dumb_product

zenweasel:
It appears we are still not stopping people from publishing a product that doesn't contain at least one variant? I feel like this is the final piece of this puzzle. If we don't solve it here we should open a separate ticket, but it feels like this is where it should be handled.

Other than that it seems like this works as expected. The issue above occurs on `master` so it's not a regression.

mikemurray:
@zenweasel Maybe instead of not allowing a product to be published without variants, we simply disable/hide the add-to-cart button, and throw up some messaging like "This product has been discontinued" or "This product is currently not for sale", or even "Call 1-800-call-me" to discuss pricing. This way, you could keep the product around for historical reasons, or use the page for content instead of purchasing. 

Not sure if that makes sense at the moment, but maybe perhaps in the future.

aaronjudd:
re: `publishing a product that doesn't contain at least one variant so it can't be purchased`, I'd suggest that we leave as is for now, open a new issue if we want to rethink/gather feedback.  I can think of cases where you might want a deprecated/legacy product to be visible for seo/search but not purchaseable..  Might be other cases, but not sure that's necessarily bad to publish without options.  Maybe some better messaging might help.

aaronjudd:
Clicking on the variants visibility icon, and also after the variant is deleted.

``` 
 at packages/meteor.js:785:14
Exception in queued task: Error: Could not find element with id YZkeBjyi9nrotcSDc to change
    at SessionCollectionView.changed (packages/ddp-server/livedata_server.js:188:13)
    at Session.changed (packages/ddp-server/livedata_server.js:400:10)
    at Subscription.changed (packages/ddp-server/livedata_server.js:1254:19)
    at Object.changed (server/publications/collections/products.js:294:18)
    at [object Object].observeChangesCallbacks.changed (packages/minimongo.js:3755:28)
    at self.applyChange.changed (packages/minimongo.js:3659:44)
    at packages/mongo/observe_multiplex.js:183:30
```

aaronjudd:
5n4wgmnpna3tk27fa

aaronjudd:
👍

kieckhafer:
👍

PR #1573, opened 2016-11-12
move import settings

### LINKED ISSUE #1050: "Import" button in dashboard does not work
### Expected behavior

Import button should allow you to upload a file and import the data within An error should be shown if you try to upload an invalid file type.
### Actual Behavior

You can choose a file but when you click on the "Import" button nothing happens
1. Go Dashboard -> Core -> Import
2. Choose a file
3. Press the "Import" button
4. Observe that nothing seems to happen

aaronjudd:
- moved “Import” from settings/settings.html to import/import.html
- is currently not used and won’t be visible, but can be basis for
import settings.

resolves #1050

### DIFF
--- a/imports/plugins/core/dashboard/client/templates/import/import.html
+++ b/imports/plugins/core/dashboard/client/templates/import/import.html
@@ -1,3 +1,26 @@
+<template name="importSettings">
+    <div class="panel panel-default">
+      <div class="panel-heading">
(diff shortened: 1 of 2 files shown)


zenweasel:
Tested. Verified fixed.

kieckhafer:
👍

Issue #1398, opened 2016-09-17
elevate current default logging levels


aaronjudd:
I suggest that we replace the current use of logging _info_ levels with _debug_ , and update _debug_ to use _trace_. 

```
Logger.info -> Logger.debug
Logger.debug -> Logger.trace
```

Our current use of info is just too noisy, and is mostly suited for development information.  The cases where we actually use debug are really worst case issues, so I think it make sense to just bump up the levels used, and use info for just the absolute "info" type messages (ie: a system process that doesn't occur very often -> not debugging or event logging).

This would likely mean a giant search and replace, as well as a update to the debug log level configuration. (both implying a potential conflict heavy branch, ideally timed right before a release I think). This is also related to #808 and could be an ideal time to make some of those updates.

impactmass:
`.debug -> .trace` - affects ~60 files
`.info -> .debug` - affects 9 files.

With the current default log level set as "INFO" (except if overridden by env var), a bunch of logs which are moving to debug and trace won't show up. So.. we need to look at those that we really want to show.

Question: This issue is titled "elevate current default logging levels", but going from "info" to "debug" and "debug" to "trace" is actually demoting the log level. Is that the plan? cc @aaronjudd

aaronjudd:
I feel as though we've largely done this, in the way we've been determining the levels since this issue was created.   Now a wholesale search and replace might not be advisable, as it could be a lot of instance we're the level is/will be appropriate ( maybe a manual walkthrough, but this might require a good sense of what should be logged for info ).  Perhaps adding the debug log level as a prettify console, and trace as the raw json debug would be appropriate as a first step.

impactmass:
@aaronjudd I've added the first step you mentioned of using raw json when trace is enabled [here](https://github.com/reactioncommerce/reaction/pull/1997).
Any other modifications to be made under this task before I open the PR for review?

One side effect here: When `trace` log format is set as "json", debug logs (since they're are a level above trace) also appear as json, except if defined otherwise. I think it's fine, or would you want this defined?

hrath2015:
Time in logs is shown in UTC.  Is there a way to override it to use Timezone of choice(or at least local time) 
While you are at it, can this be addressed. Sorry if it is not a right place to ask or suggest.

jshimko:
[...]

PR #1574, opened 2016-11-12
add Loggly integration

jshimko:
- add [Loggly](https://www.loggly.com/) integration for Bunyan logger
- refactor logger config to clarify level settings (I feel like `isDebug` implies a boolean, but it can also be used to set the actual log level - which is probably far more often what people use it for rather than setting `true`)

#### New Environment Variables
`REACTION_LOG_LEVEL`
`LOGGLY_TOKEN`
`LOGGLY_SUBDOMAIN`
`LOGGLY_LOG_LEVEL`

Right now, there's no UI for adding Loggly settings.  It's environment variables only because I just wanted to get something working and merged without having to make decisions about where things should go or how they should look in the UI.

To test this PR...

```sh
reaction init -b loggly-integration loggly

cd loggly

LOGGLY_TOKEN="" LOGGLY_SUBDOMAIN="" reaction
```

If both variables are set, logs will be sent to Loggly at DEBUG level by default.  That can be overridden by setting the `LOGGLY_LOG_LEVEL` env var.

Adding a UI for this later should be fairly trivial.  I just wanted to get something available so advanced users can pipe their logs somewhere other than stdout for debugging.


aaronjudd:
If we're going to deprecate `isDebug` maybe we should consider #1398 at this point.  Just the level formatting, not the updating of all uses (that we can do quickly in 0.18).   We also need to update our docs, examples to reflect the new uses.  I think `isDebug` supporting boolean was in the case where it's actually `Meteor.isDebug`, but I'm not sure, and separating this is for the best I think.

zenweasel:
Approved

kieckhafer:
This seems like it's a good idea and the code looks good, but we are making a lot of changes here, I think we definitely need to update the docs to go along with this before it's pulled in to dev.

aaronjudd:
Be great if we could merge this in #1584

jshimko:
I think it'd be great if we could merge this as-is and do the other logger updates as a separate PR.  I feel like that part is going to be quite a bit of effort and probably could use a bit more group discussion.  Mostly, I think it'd be great to define some standards around what qualifies something as falling into a specific log level.  

Also, there are currently a lot of debug logs that don't output very useful information and it's unclear where they even came from without doing a search through the code.  I'd love to discuss coming up with some standards for the amount of provided detail when it comes to debug logs (kind of related to the child logger idea where context data gets embedded in the logger within the scope it's currently running in - for example, the name of the Meteor method or function that the logger instance was triggered by).  

So... maybe we merge this now and then open an issue to draw up some acceptance criteria around further logging improvements?  

@kieckhafer agreed, I'll take care of updating the docs right now.  There's not really that much changing here though.  It's really just `isDebug` becomes `REACTION_LOG_LEVEL`.  Worst case, anybody using `isDebug` would end up having their log level fall back to the default as if they hadn't set anything.  Definitely not a breaking change and I'd be surprised if many people were even using it anyway.

aaronjudd:
@jshimko sure thing, just attach to docs PR to this issue and we're good to go.

jshimko:
Doc updates
https://github.com/reactioncommerce/reaction-docs/pull/140

aaronjudd:
👍  @jshimko thanks. looks good.

PR #1575, opened 2016-11-12
Meteor 1.4.2.1

jshimko:
- update Meteor to 1.4.2.1
- install `babel-runtime` as [recommended](https://forums.meteor.com/t/meteor-1-4-2-1-is-an-important-patch-for-1-4-2-users/31190)
- update Docker build script to use new `METEOR_ALLOW_SUPERUSER` variable instead of `--unsafe-perm` flag
- Send the `meteor build` warning about building as root to `/dev/null` (since some people have assumed this is an error and the build can take a long time in larger apps).  The warning is irrelevant in this case because all permissions are updated after the build and the app is then run by a non-root user.


aaronjudd:
👍  bumped version, let's merge.

zenweasel:
Approved

PR #1580, opened 2016-11-14
Update Order Progress Workflow

### LINKED ISSUE #1224: Order progress status is not correctly being updated on the Orders admin page
### Expected behavior

The "Start" button should change to our "Arrow Button" once the first step in the order fulfillment process has been completed. It shouldn't change immediately when you click the order card itself, as you might want to view an order without starting it, but should change once the "Approve" button is pressed.
### Actual Behavior

You can "Start" an order by clicking anywhere inside the order detail card, you don't necessarily need to click the green "Start" button to do so. When you start an order after clicking the card OUTSIDE the start button itself, the Start button does not change. As you can see in this screen shot, I have already gone through the whole flow and approved the payment, and the start button is still showing. Even after a refresh, it shows.

screen shot 2016-07-29 at 10 32 36 am

1) Order an item
2) Log in as an admin and click the "Orders" button
3) Click on there main part of an order card (not the green "Start" Button
4) Use the admin panel to complete the order
5) See that the start button is still present in the main order card screen

kieckhafer:
Fixes #1224 

When an order card is clicked in the main part of the card, and not the active “Start Order”, the workflow is never updated.

This change adds a Meteor call so if the workflow was never updated the first time, the workflow is now updated once the payment is captured.

### DIFF
--- a/imports/plugins/core/catalog/client/index.js
+++ b/imports/plugins/core/catalog/client/index.js
@@ -1,2 +1,2 @@
 import "./templates/settings.html";
-import "./templates/settings.js"
+import "./templates/settings.js";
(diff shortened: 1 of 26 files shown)


aaronjudd:
👍

PR #1585, opened 2016-11-15
more email templates

kieckhafer:
Having issues with the last PR, so making a new one.

PR #1588, opened 2016-11-18
Send order notification when user adds email

### LINKED ISSUE #431: Email confirmation on guest order update entry
Confirm email account and register a user account for a guest who adds their email to the order confirmation page.

kieckhafer:
Fixes #431


kieckhafer:
I am aware there is an error reported in the log regarding `Alerts` not being defined. This is fixed in #1585, I just branched this off of `release-0.18.0`, where that fix is not yet pulled in.

zenweasel:
👍  Tested: Verified works

zenweasel:
Actually, the subject of the email I got said "Your order has shipped"

kieckhafer:
The subject line is / should be fixed in #1585. Since it's just going into the release branch, and not into development, I think we should merge if everything else is OK, and once #1585 is merged I will double check to make sure the subject is correct.

Issue #1578, opened 2016-11-14
[Marketplace] Ban account + Close account functionality


lcampanis:
Adding functionality for Ban and Close account.

**Ban account**
In a marketplace setup - Allow admins and other moderator roles to ban a user account from the site.

- The account will be blocked without notice
- Email is blocked from sign in and register
- Username is unavailable
- Products related to that user account are unpublished (not deleted for safety, but also for the ability to reinstate an account easily in the future)
- Disable Shop in Marketplace

Future enhancements:
Enhancements for email templates and other hooks can be added in the future.

**Close account**
For single shop and marketplace: Allow all users (except the admin) to Close their account at will.
- Warn users about the consequences of closing their account
- Delete user account entirely
- Delete all associated products
- Delete Shop in Marketplace
- Keep orders (Implications of removing invoices may vary between countries)
Question: Keeping orders is safe, but can an order exist in Reaction if the user, shop and products are deleted entirely? And is it necessary?

**UI**
- Ban account button can sit under Accounts > Manage Account in a new panel above Permissions.
- Close account can sit in the Account > Profile page at the bottom of the page
- Templates for both will be available so that buttons can be added anywhere 

We'll be taking this one on this week. 

Has this been discussed or thought of before? 
Any thoughts?

lcampanis:
While we are trying to do this as a standalone plugin, in some cases it is needed to alter core. 
In this case dashboardAccounts so that the Ban User button is part of the manage user settings screen, but most importantly so that banned users are not in the list of Guests under accounts.

So far the only way I have found is to override reaction templates entirely in the new package, but this doesn't cut it as the package will always have to update its views every time reaction updates them.
This will require some Template hooks, for something say the registry provides: "settings" renders in settings when the same route is used, with a topToBottom priority?
Here's an example in action.
![image](https://cloud.githubusercontent.com/assets/670954/20383475/a4fe5d48-acb8-11e6-824d-0a1e3886f874.png)

lcampanis:
This will need some core changes to get it done correctly. I am putting it on hold for now as we need to sort out some marketplace issues before.

spencern:
@lcampanis This may be more feasible from a plugin now with the new accounts components and group based permission features. Let me know if there's anything I can do to help you deliver this feature

aaronjudd:
Closing - insufficient details, see new user management utilities.

PR #1591, opened 2016-11-23
[WIP] - Marketplace: Admin role permissions for sellers, Allow guests to become sellers, Create a shop and Post products

lcampanis:
**This branch is WIP.**
Please code review when you can.
- Allow Sellers to add products via `Add Product`to their shop only, when they have the `createProduct` role
- Amended `shopIdAutoValue` to check loggedIn user has a shop, or fallback on parent shop
- Amended `getCurrentShop` as a start towards shop switching UI
- Amended `Product` publication to allow sub-sellers (admin) to add products to their shop
- Amended `gridControls` so that sub-sellers with `createProduct` permission can edit their products from the grid (create a new grid view for this to work)
- Allow guests to become sellers when marketplace module is enabled, `defaultSellerRoles` in Account Schema
- Added Become a seller panel in user profile page only when Marketplace plugin is enabled and `AllowGuestSeller` option is enabled
- Amended `shop/createShop` to allow guests to create a shop only once, when marketplace is enabled
- New meteor method `shop/getSellerShopId` to safely get seller shop Id, or default to parent shopId
- New `lib/api/core` for `Reaction` 
- New core functions `Reaction.addRolesToVisitors` and `Reaction.isPackageEnabled`
- Fixed Reaction.getPackagesSettings
- Marketplace dashboard settings under `Catalog` settings with option to `allowGuestSellers` (might be better under `Settings`?)
- Fix loggedIn User dropdown perms - PR https://github.com/reactioncommerce/reaction/pull/1639

### In  Progress
1. **Fix Permission issues**
    - Fix for posting a new product as a seller **(RC to investigate)** 
https://github.com/reactioncommerce/reaction/pull/1591/files#diff-db3ed98289e6a832d284ffae19a90194R132
    - Fix product image upload permissions **(RC to investigate)**
2. Allow Sellers to edit their shop settings
3. New `Products` Grid View and Publications showing products from multiple sellers, with permissions for logged in seller to edit directly from the grid

### To Do
1. Product cart flow to allow products from multiple sellers **TBD**
1. Payment flow to allow payments from multiple sellers **TBD**
1. Allow Sellers to edit their shop URL by updating Shop.domains (domain, subdomain, path based URLs)
1. Allow Tag management per shop 
1. Expand on dashboard settings, with `enableBanAccounts` see https://github.com/reactioncommerce/reaction/issues/1578, etc)


aaronjudd:
@lcampanis I had a few issues while testing with my user account/shop logging out on my unexpectedly, but I suspect that it's an exaggeration of an existing issue.  Curious if you've seen that?  I'm testing your PR now with release-0.18 merged in as well. Small conflict in `core.js`, but otherwise seems to be fine with all the updates coming soon.  

Not sure I understand why you added the ShopTaxes schema, could you elaborate?

lcampanis:
@aaronjudd Very happy with your feedback. Thanks for that!
Re logout - yes I noticed, we'll fix that now that we're on the same page.
ShopTaxes had to be added in the Shop schema as the new shop is cloned/taken from the 1 shop in the DB (and the original has Taxes in it) which aren't in the default Shop Schema for the check to pass. It was hitting on line 38 here https://github.com/reactioncommerce/reaction/pull/1591/files#diff-a85b4a331f34703f51a72e9728869fe2R38
Passing a clean shop object as an argument in the createShop function would solve that, assuming Taxes aren't in it.

imports/plugins/included/marketplace/server/startup/init.js
```diff
@@ -2,14 +2,10 @@ import { Shops } from "/lib/collections";
...
-  // Add the artlimes/become-seller permission to all default roles since it's available to all
-
-}
 
 /**
  * Hook to make additional configuration changes
  */
 Hooks.Events.add("afterCoreInit", function() {
-
+  //Reaction.addRolesToVisitors("marketplace/");
```
lcampanis: Packages can now add permissions like this

lib/api/helpers.js
```diff
@@ -4,12 +4,22 @@ import { FlowRouter } from "meteor/kadira:flow-router-ssr";
 import { Shops } from "/lib/collections";
 
 /* eslint no-unused-vars: 0 */
-//
-//  TODO review this slugify import in lib/api/helpers
-//
+
+// Export Reaction using commonJS style for common libraries to use Reaction.*
+// https://docs.meteor.com/packages/modules.html#CommonJS
+let Reaction;
```
lcampanis: We need a common Reaction object for common libs to use safely. 

This is only a suggestion and we'd like your feedback on this to see if it suits everyones needs.
Meteor suggests to use commonJS require() style as the only way to include modules conditionally (see page for more)

lib/collections/schemas/shops.js
```diff
@@ -122,24 +122,6 @@ export const BrandAsset = new SimpleSchema({
 });
 
 /**
- * Shop Taxes
```
lcampanis: No need for this anymore. We trust the owner's cloned shop from the DB and only check(shop) when a shop is passed as an argument. Should be safe for creating shops on the fly - but needs testing (as does creating shop after reset - WIP)

server/api/core/core.js
```diff
@@ -211,6 +211,38 @@ export default {
...
+    const settings = this.getPackageSettings(name);
+
+    return !!settings.enabled;
+  },
+
+  /**
+   * Add default roles for new visitors
+   * @param {String|Array} roles A string or array of roles and routes
+   */
+  addRolesToVisitors(roles) {
```
lcampanis: It's flexible to add an array of roles or simply a string

server/methods/core/shop.js
```diff
@@ -65,11 +60,15 @@ Meteor.methods({
...
+    // We trust the owner's shop clone, check only if shopData is passed as an argument
+    if(shopData) {
+      check(shop, Schemas.Shop);
+    }
+
     try {
       Collections.Shops.insert(shop);
     } catch (error) {
-      return Logger.error("Failed to shop/createShop", sanitizedError);
+      return Logger.error("Failed to shop/createShop", error);
```
lcampanis: Not sure where sanitizedError is coming from, but error should be it?

aaronjudd: Agree, looks like a typo, plus our preference is to always use "error".

imports/plugins/core/dashboard/client/templates/shop/settings/settings.html
```diff
@@ -26,11 +26,11 @@
         </div>
       </div>
       <div id="general" class="panel-collapse collapse in" role="tabpanel" aria-labelledby="General">
-        <div class="panel-body">
+        <div class="panel-body">-
```
aaronjudd: Are these `-` dashes, remnants? Debug?

imports/plugins/included/marketplace/lib/api/marketplace.js
```diff
@@ -0,0 +1,50 @@
...
+
+export function hasMarketplaceGuestAccess() {
+  const currentUser = Meteor.user();
+
+  // parent shop owners have full access
+  if(Roles.getGroupsForUser(currentUser, ['owner']).length) {
+    return true;
+  }
+
+  const packageSettings = Reaction.getPackageSettings('reaction-marketplace');
```
aaronjudd: `getPackageSettings` is a much cleaner implementation . We should create an issue to clean up the implementation in the rest of the core packages, and make sure that docs are updated with these new methods.

A comment on styles,  `eslint` should be warning on these,  but http://eslint.org/docs/rules/no-extra-semi and http://eslint.org/docs/rules/quotes are common offenders in this code.  `eslint -fix` is the easiest way to clean all those up.   Generally, we've been "double quote" advocates.

imports/plugins/included/marketplace/server/startup/init.js
```diff
@@ -0,0 +1,11 @@
+import { Shops } from "/lib/collections";
+import { Hooks, Logger } from "/server/api";
+import { Reaction } from "/server/api";
+
+
+/**
+ * Hook to make additional configuration changes
+ */
+Hooks.Events.add("afterCoreInit", function() {
```
aaronjudd: Is this a dangling block now?  Did `marketplace/` work with the trailing `/`?

server/api/core/core.js
```diff
@@ -162,6 +174,11 @@ export default {
         _id: 1
       }
     }).fetch()[0];
+
+    if (!shop) {
+      Logger.debug(domain, "Add a domain entry to shops for ");
```
aaronjudd: This probably warrants `Logger.warn`.

imports/plugins/included/marketplace/register.js
```diff
@@ -0,0 +1,41 @@
...
+    name: "Marketplace",
+    public: {
+      allowGuestSellers: true
+    }
+  },
+  registry: [
+    // page
+
+    // Dashboard card
+    {
```
aaronjudd: I wonder if we should just place the settings under the `Catalog` card, and not give Marketplace it's own dashboard card (we're trying to consolidate these as much as possible).  All the registry entries would still be here.  Easy to move later one way or another though.

lib/api/helpers.js
```diff
@@ -42,6 +52,33 @@ export function getShopName() {
...
+  const user = Meteor.users.findOne({ _id: userId });
+  const domain = url.parse(Meteor.absoluteUrl()).hostname;
+
+  let query = {
+    domains: { $in: [domain] }
+  };
+  let shopId;
+
+  if(user) {
+    shopId = Roles.getGroupsForUser(userId, 'admin')[0];
```
aaronjudd: More linting here with http://eslint.org/docs/rules/quotes

lib/api/helpers.js
```diff
@@ -4,12 +4,22 @@ import { FlowRouter } from "meteor/kadira:flow-router-ssr";
 import { Shops } from "/lib/collections";
 
 /* eslint no-unused-vars: 0 */
-//
-//  TODO review this slugify import in lib/api/helpers
-//
+
+// Export Reaction using commonJS style for common libraries to use Reaction
+// https://docs.meteor.com/packages/modules.html#CommonJS
+let Reaction;
```
aaronjudd: I think this makes sense, @jshimko care to comment on this?  I think this make some thing easier, is this something we should also do for Schemas, Collections?

imports/plugins/included/marketplace/lib/api/marketplace.js
```diff
@@ -0,0 +1,50 @@
+import { Meteor } from "meteor/meteor";
+import { Packages } from "/lib/collections";
+import { Reaction } from "/lib/api";
+
+const name = "reaction-marketplace";
```
aaronjudd: Unused var declaration here.

server/methods/core/shop.js
jshimko: The error object also needs to be the first arg to the logger method to be parsed correctly.

jshimko: `Logger.error(error, "Some message string")`

server/api/core/core.js
jshimko: Agreed.  Definitely feels like a `warn` scenario.  And I think it'd be a good idea to clarify the output a little.  I usually make the first arg an object so it's a little clearer what you're seeing when that outputs to the console (or wherever).  So...

```js
Logger.warn({ domain }, "Shop not found");
```

Otherwise, when you use a string for that first arg, it renders that as if it's the main log message.

Or maybe better yet, just use a single arg as a string...

```js
Logger.warn(`Add a domain entry to shops for ${domain}`);
```

But I also think it could use a more helpful message for debugging.  Perhaps...

```js
Logger.warn(`getShopId: Shop not found with domain ${domain}`);
```

That tells you where and why it happened.

lib/api/helpers.js
jshimko: Yeah, I guess that makes sense.  We'd probably want to update all of the imports all over the app too.  And yes, it should probably extend with Collections/Schemas there instead of doing it redundantly on both the [client](https://github.com/reactioncommerce/reaction/blob/development/client/modules/core/index.js#L10-L11) and [server](https://github.com/reactioncommerce/reaction/blob/development/server/api/core/index.js#L14-L15).

However, I definitely don't think that's the right file for that stuff.  It should probably live in `/lib/api/core.js`.

jshimko: Actually, it just occurred to me that pulling the Collections/Schemas into there and removing from the client/server locations would be a breaking change for anyone not importing `Reaction` from `/lib/api`.  I think we need to leave that part alone for now.

jshimko:
Lots of linting errors in this PR.  Check out the style guide for setting up your editor to detect that stuff.

https://docs.reactioncommerce.com/reaction-docs/development/styleguide

imports/plugins/included/marketplace/client/templates/becomeSellerButton/becomeSellerButton.js
```diff
@@ -0,0 +1,36 @@
...
+    return classes.join(' ');
+  }
+});
+
+
+Template.becomeSellerButton.events({
+  "click [data-event-action='button-click-become-seller']": function() {
+    Meteor.call("shop/createShop", Meteor.userId(), function(err, response) {
+      if(err) {
+        return Alerts.toast("Unable to create shop with specified user", "error");
```
zenweasel: Needs i18n?

lib/api/helpers.js
```diff
@@ -42,6 +52,33 @@ export function getShopName() {
...
+    domains: { $in: [domain] }
+  };
+  let shopId;
+
+  if(user) {
+    shopId = Roles.getGroupsForUser(userId, 'admin')[0];
+    query._id = shopId;
+  }
+
+  const shop = Shops.find(query, {
```
zenweasel: Could you just use `findOne` here?

server/methods/core/shop.js
```diff
@@ -19,36 +20,55 @@ Meteor.methods({
...
     const userId = shopAdminUserId || Meteor.userId();
-    const adminRoles = Roles.getRolesForUser(currentUser, Reaction.getShopId());
+    let adminRoles = Roles.getRolesForUser(currentUser, Reaction.getShopId());
     // ensure unique id and shop name
     shop._id = Random.id();
     shop.name = shop.name + count;
 
-    check(shop, Schemas.Shop);
+    // admin or marketplace needs to be on and guests alowed to create shops
+    if(currentUser && Marketplace.hasMarketplaceGuestAccess()) {
```
zenweasel: linting, spaces between `if` and condition

server/publications/collections/product.js
```diff
@@ -14,7 +14,14 @@ Meteor.publish("Product", function (productId) {
     return this.ready();
   }
   let _id;
-  const shop = Reaction.getCurrentShop();
+  let shopId;
+
+  if(this.userId) {
```
zenweasel: linting: spacing here

zenweasel:
Testing this now. As a registered user (not an admin) I have the "Add a product" option before electing to become a seller. If I select that option I get an error in the console.

```
Exception in delivering result of invoking 'products/createProduct': Error
    at http://localhost:3000/app/app.js?hash=f9910344a20c449f69997e9306302dbdb5bacdc9:47141:19
    at MethodInvoker._callback (http://localhost:3000/packages/meteor.js?hash=e3f53db3be730057fed1a5f709ecd5fc7cae1229:1105:22)
    at MethodInvoker._maybeInvokeCallback (http://localhost:3000/packages/ddp-client.js?hash=bc32a166cd269e06a394f9418e0024d805bab379:3557:12)
    at MethodInvoker.receiveResult (http://localhost:3000/packages/ddp-client.js?hash=bc32a166cd269e06a394f9418e0024d805bab379:3577:10)
    at Connection._livedata_result (http://localhost:3000/packages/ddp-client.js?hash=bc32a166cd269e06a394f9418e0024d805bab379:4736:9)
    at onMessage (http://localhost:3000/packages/ddp-client.js?hash=bc32a166cd269e06a394f9418e0024d805bab379:3385:12)
    at http://localhost:3000/packages/ddp-client.js?hash=bc32a166cd269e06a394f9418e0024d805bab379:2736:11
    at Array.forEach (native)
    at Function._.each._.forEach (http://localhost:3000/packages/underscore.js?hash=cde485f60699ff9aced3305f70189e39c665183c:149:11)
    at SockJS.self.socket.onmessage (http://localhost:3000/packages/ddp-client.js?hash=bc32a166cd269e06a394f9418e0024d805bab379:2735:11)
```

zenweasel:
If I elect to become a seller and try to add a product, I can't edit variants (no pencil icon)

my_personal_product

zenweasel:
I also get a 403 when trying to upload an image

zenweasel:
If a turn off "Allow guest sellers" I still have the "Add product" options available from the dropdown

zenweasel:
@lcampanis Since I can't add a product I can't really do much more testing on this. Let me know when this is updated and I can give it another go.

aaronjudd:
>Note - after merging release 0.18 the dropdown menu shows more items than expected, we'll investigate.

 this is likely due to a significant change in how the Packages collection is published, and how Reaction.Apps filters/displays packages.  my comment ` // todo could add audience permissions to registry` was about adding an additional "audience" permission check (feel free to do this).  In the updated [Packages publication line 20](https://github.com/reactioncommerce/reaction/blob/release-0.18.0/server/publications/collections/packages.js#L20), should be checking doc.name (package name) permissions exist,  seems like that needs to extended to look at the registry permissions object as well...

lcampanis:
Hi all,
Requested changes are all done and code is clean ready to review. :)
We'll create separate PRs for the rest.
Updated description at the top to see where we are, only missing a couple of things to call it ready.

lcampanis:
@aaronjudd Not sure why this PR is showing more files and commits, as we've now merged and checked across all ends since release 18 was closed and there shouldn't be any additional commits, and only **28 files** changed.

Creating an identical PR with the same base/branch combo will actually give you the correct files. Weirdest thing! But to save us time give it a go rather than trying to fix this one ;)

aaronjudd:
@lcampanis that is weird.. I'm sure the base got screwed up somehow with some magic.. the changes happened automatically when I merged release-0.18.  In any case, a new PR would be fine, just reference this one and we'll close this.

PR #1600, opened 2016-12-02
Templates fixes

### LINKED ISSUE #1594: Receive validation error when trying to change email template in Dashboard
### Expected behavior
No errors when valid form submitted

### Actual Behavior
Receive "Name is required" error when submitting form even when name field is filled and not been changed

1. Edit a template
1. Don't change anything (or do, no difference)
1. Hit update
1. Observe error

kieckhafer:
- Fixes #1594
- Fixes "Add template" issue from #1585 by removing the settings button from the dashboard

### DIFF
--- a/imports/plugins/core/templates/client/templates/email.html
+++ b/imports/plugins/core/templates/client/templates/email.html
@@ -7,7 +7,7 @@
             <th>{{i18n "mail.templates.templateName" "Template Name"}}</th>
             <th>{{i18n "mail.templates.type" "Type"}}</th>
             <th>{{i18n "mail.templates.language" "Language"}}</th>
-            <th>{{i18n "mail.templates.enabled" "Enabled"}}</th>
+            <!-- <th>{{i18n "mail.templates.enabled" "Enabled"}}</th> -->
             <!-- <th><i class="fa fa-plus pull-right add-shipping-method" title="{{i18n 'mail.templates.addTemplate' 'Add email template'}}" data-event-action="addEmailTemplateMethod"></i></th> -->
             <th>Actions</th>
           </tr>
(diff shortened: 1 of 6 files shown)


zenweasel:
If I try to update the subject of an email, that change is lost when I hit "Update"

email_update

kieckhafer:
Either way it needs to be fixed, but...

Is the change "saved" in the overall scheme of things, even though the state isn't updated in the side panel, and shows when the page is refreshed?

Or is it not saved at all, even on refresh?

Guessing it's the state not being updated initially.

zenweasel:
Yes, the changes are there after a refresh.

zenweasel:
So was the fix to the error when adding templates to just remove the ability to add templates?

zenweasel:
Weren't we going to standardize these on "Confirm" or something similar?

templates

zenweasel:
Not sure where this fits in with the scope of this PR but I wonder what happens to the "reset password" functionality if the reset password template is disabled or deleted.

kieckhafer:
@zenweasel 

1) Reactivity is now working in sidear
2) Yes, we did discuss standardizing, this is now done.
3) Yes, fixing the error is akin to removing the ability to add a new template, for the time being.

imports/plugins/core/templates/client/templates/email.html
```diff
@@ -7,9 +7,7 @@
             <th>{{i18n "mail.templates.templateName" "Template Name"}}</th>
             <th>{{i18n "mail.templates.type" "Type"}}</th>
             <th>{{i18n "mail.templates.language" "Language"}}</th>
-            <th>{{i18n "mail.templates.enabled" "Enabled"}}</th>
-            <!-- <th><i class="fa fa-plus pull-right add-shipping-method" title="{{i18n 'mail.templates.addTemplate' 'Add email template'}}" data-event-action="addEmailTemplateMethod"></i></th> -->
-            <th>Actions</th>
+            <th class="textAlignRight">Actions</th>
```
zenweasel: this doesn't need an i18n?

imports/plugins/core/templates/client/templates/email.js
```diff
@@ -99,7 +56,26 @@ Template.emailTemplatesDashboard.events({
 /*
  * Template emailTemplatesSettings Helpers
  */
+Template.emailTemplateSettings.onCreated(function () {
+  this.state = new ReactiveDict();
+
+  this.autorun(() => {
+    const currentData = Template.currentData();
+    const template = Templates.findOne(currentData._id);
```
zenweasel: While the use of the variable name "template" is totally logical, I wonder if it doesn't lead to some confusing-to-read code? Just a thought.

PR #1601, opened 2016-12-03
Fix of nodemailer secure connection configuration.

actuosus:
We cannot figure out nodemailer secure connection based on port that is converted to Number and compared with String.

Also JSDoc of return value is changed to structured Object.


zenweasel:
Approved - Seems to be working

jshimko:
Good catch.  Thanks!

👍

Issue #1595, opened 2016-11-28
Error in server console when placing order


zenweasel:
I am migrating this from a comment on #1585 

Seeing this error in the server console after placing an order

```
03:08:52.788Z ERROR Reaction: Error in orders/sendNotification for order W2BGyczBMJoqQ9egH
  Error: Access Denied [403]
      at [object Object].ordersRefundsList (server/methods/core/orders.js:857:13)
      at packages/check.js:130:16
      at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
      at Object.exports.Match._failIfArgumentsAreNotAllChecked (packages/check.js:129:41)
      at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1709:18)
      at packages/ddp-server/livedata_server.js:1625:18
      at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
      at Server.apply (packages/ddp-server/livedata_server.js:1624:45)
      at Server.call (packages/ddp-server/livedata_server.js:1567:17)
      at packages/kadira:flow-router-ssr/server/plugins/ssr_data.js:67:23
      at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
      at Object.Meteor.(anonymous function) [as call] (packages/kadira:flow-router-ssr/server/plugins/ssr_data.js:66:44)
      at [object Object].ordersSendNotification (server/methods/core/orders.js:390:33)
      at packages/check.js:130:16
      at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
      at Object.exports.Match._failIfArgumentsAreNotAllChecked (packages/check.js:129:41)
      at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1709:18)
      at packages/ddp-server/livedata_server.js:1625:18
      at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
      at Server.apply (packages/ddp-server/livedata_server.js:1624:45)
      at Server.call (packages/ddp-server/livedata_server.js:1567:17)
      at packages/kadira:flow-router-ssr/server/plugins/ssr_data.js:67:23
```

kieckhafer:
@zenweasel I believe this was fixed in a different PR than #1585, and I cannot replicate. Can you please check again if you are still seeing this in release-0.18.0

zenweasel:
Can't replicate. Closing

zenweasel:
[...]

PR #1607, opened 2016-12-07
Fix permissions on Order when "guest" user

### LINKED ISSUE #1602: Receive error when adding email to anonymous order
### Expected behavior
No error in console

### Actual Behavior
Receive error:
`Error invoking Method 'orders/sendNotification': Access Denied [403]`

1. Place order as guest
1. Put email in "Add your email" field
1. Observe error message

kieckhafer:
Fixes #1602 and #1595

When a guest / non-admin user created an order, our permissions were not allowing refunds to be accessed, which in turn was not allowing emails to be sent, as refunds are listed inside order emails.

This update grants extra permissions so if an order is owned by the current user, they are allowed to access the needed information


zenweasel:
Approved. Tested and verified

PR #1612, opened 2016-12-09
Move Template editing to Sidebar

kieckhafer:
- Convert form to Griddle
- Move template editing to be fully in sidebar


zenweasel:
Are we intentionally showing these in the form and then telling people they can't edit them? Should we hide the ones they can't edit?

dashboard

zenweasel:
When I filter by "orders" I still see templates that aren't for orders?
dashboard

kieckhafer:
We purposely left the message about React Templates not working in the sidebar. It creates a message that this is all the Templates on the site, even if you can't edit some of them from here, at least you have an easy to access list of every available template.

PR #1616, opened 2016-12-12
Use product handle for `pathFor`

### LINKED ISSUE #1586: Cart product link doesn't work
When clicking of the items you previously added to the cart it does not open the respective product page.

- Let's say you have 2 products - Product A and Product B. 
- If you add 5 of Product A to the cart and then open Product B
- Then click of the Product A item in the cart
- It does not work

It seems that the cart link structure is wrong and missing the product ID after "product/..." e.g. http://localhost:3000/reaction/product/%2F/BQTAZfTfCJeHjf3H9

zenweasel:
Resolves #1586


aaronjudd:
@mikemurray I think you are correct, because this is supposed to select the product with the options, not just the product.

PR #1617, opened 2016-12-12
Don't blank out titles when cloning variants

### LINKED ISSUE #1610: Duplicating a variant produces empty variant rather than duplicate
### Expected behavior
When editing a product, clicking on the "Duplicate" button within a variant should duplicate that variant for editing

### Actual Behavior
Creates an empty variant

1. Create a new product
1. Create a variant and fill out fields
1. Click on "Duplicate" icon
1. Observe that new variant is created, but it is empty

zenweasel:
Resolves #1610 

This is not so much a bug fix as a change in intentional behavior but I feel like it makes sense.

If I created a shoe with 15 different sizes and now I wanted to clone that to another color, producing a clone with all the option names blanked out pretty much negates most of the benefit I get from cloning a product rather than creating it from scratch. It also makes it hard to determine which is which.

Maybe there was a reason for this that I don't see in which case this PR and bug can just be closed.

### DIFF
--- a/server/methods/catalog.js
+++ b/server/methods/catalog.js
@@ -349,7 +349,7 @@ Meteor.methods({
         type = "parent";
         Object.assign(clone, variant, {
           _id: variantNewId,
-          title: ""
+          title: `${variant.title} - copy`
         });
       } else {
         const parentIndex = variant.ancestors.indexOf(variantId);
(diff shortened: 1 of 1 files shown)


mikemurray:
👍

PR #1620, opened 2016-12-13
Checkout discount code input

aaronjudd:
DiscountList and DiscountForm React components added to checkout as a payment method. 
Discounts stored as billing.paymentMethods in the Cart/Orders.  Some handling is updated for handling of multiple payment methods ~, but WIP~.  A number of todo's also remain for conditional and limit handling.


aaronjudd:
Discounts should be functional.  To test, Enable "Codes" in the "Payment" settings. Add a code and an amount.  Checkout, apply code. Code can be added as well in order fulfillment. 

Known problems are:  
- no conditional/limit validation.
- multiple instances of billing[0] (single payment method handling) remain
-  deleting discount in order fulfillment can cause a payment provider error because the order total is now more than the originally authorized amount.  
- if the paymentMethod[0] is not the credit this could lead to issues. These need to be updated to handle each `billing.paymentMethod.method = credit` instance independently.

imports/plugins/included/discount-codes/server/methods/methods.js
```diff
@@ -32,6 +32,140 @@ export const methods = {
...
+    check(codeId, String);
+    check(collection, String);
+    const Collection = Reaction.Collections[collection];
+    //
+    // delete code from cart
+    //
+    const cart = Collection.findOne(id);
+    let hasInvoice = false;
+    let currentDiscount = 0;
+    for (billing of cart.billing) {
```
zenweasel: const here?

imports/plugins/included/discount-codes/server/methods/methods.js
```diff
@@ -32,6 +32,140 @@ export const methods = {
...
+      if (billing.paymentMethod && billing.paymentMethod.processor === "discount-code" && billing._id !== codeId) {
+        currentDiscount += parseFloat(billing.paymentMethod.amount);
+      }
+      if (billing.paymentMethod && billing.invoice) {
+        hasInvoice = true;
+      }
+    }
+    // only if this is an order
+    if (hasInvoice) {
+      selector = {
```
zenweasel: `const` here?

imports/plugins/included/discount-codes/server/methods/methods.js
```diff
@@ -32,6 +32,140 @@ export const methods = {
...
+            }
+            if (billing.paymentMethod && billing.invoice) {
+              hasInvoice = true;
+            }
+          }
+
+          Collection.update(id, { $set: { discount: currentDiscount } });
+          // if this is an order, we'll update the invoice as well.
+          if (hasInvoice) {
+            selector = {
```
zenweasel: const?

spencern:
Coming back to this in an hour.

imports/plugins/core/discounts/client/components/form.js
```diff
@@ -0,0 +1,127 @@
...
+        <Translation defaultValue="Discount submitted." i18nKey="discounts.submitted"/>
+      </a>
+    );
+  }
+
+  // handle keydown and change events
+  handleChange(event) {
+    const { attempts } = this.state;
+    // clear input if user hits escape key
+    if (event.keyCode === 27) {
```
spencern: Might be worth handling the `enter` key as well so that the payment form doesn't submit if the user hits enter after inputting the code.

imports/plugins/core/discounts/client/components/form.js
```diff
@@ -0,0 +1,127 @@
...
+  handleClick(event) {
+    event.preventDefault();
+    this.setState({ validatedInput: true });
+  }
+
+  // loader button
+  loader() {
+    const { attempts, discount } = this.state;
+    let loader;
+    if (discount && discount.length >= 10 && attempts >= 12) {
```
spencern: Is the idea that after a certain number of attempts the input will be throttled or will the form no longer attach a correct code?

imports/plugins/included/discount-codes/server/methods/methods.js
```diff
@@ -32,6 +32,140 @@ export const methods = {
...
+      // apply to cart / order
+      Meteor.call("payments/apply", id, paymentMethod, collection, (error, result) => {
+        if (result) {
+          const Collection = Reaction.Collections[collection];
+          const cart = Collection.findOne(id);
+          let hasInvoice = false;
+          let currentDiscount = 0;
+          for (billing of cart.billing) {
+            if (billing.paymentMethod && billing.paymentMethod.processor === "discount-code") {
+              currentDiscount += parseFloat(billing.paymentMethod.amount);
```
spencern: This is the line that's throwing an error if `discount.discount` is not parsable as a number.

imports/plugins/core/discounts/client/components/list.js
```diff
@@ -0,0 +1,103 @@
...
+  listItems: PropTypes.array
+};
+
+function composer(props, onData) {
+  const currentCart = Reaction.Collections[props.collection].findOne({
+    _id: props.id
+  });
+
+  const listItems = [];
+  for (billing of currentCart.billing) {
```
aaronjudd: Yes, I believe so.

aaronjudd:
@spencern thanks for the review.  Adding percentages and rates as an option (in a few)... just wanted to get the whole thing working with an amount, will calculate different kinds of rates now that the "flat amount" is working.  Will add validation for the code input and the amounts as well.

imports/plugins/core/discounts/client/components/form.js
aaronjudd: yes, but I was going to leave the details of this to another UI task.  Right now this doesn't throttle but, does show a different icon as it becomes more unlikely that a match will be made.   I wasn't sure exactly what rules we'd want to enforce.. a max length, or number of bad attempts past some character count...? Open to ideas..

imports/plugins/core/discounts/client/components/form.js
aaronjudd: good idea.. (event.preventDefault) should do this as well.. (maybe?)

spencern:
@aaronjudd that's what I figured re: percentages and rates, but wanted to note the potential confusion point. Validation will fix the issue I think. 👍

aaronjudd:
@spencern so I decided to go with the Discount input can be either a string or a number.  The admin validation needs some updating to validate this input per discount type.  
Right now, there is discount, credit, sale, shipping calculations.  With discount, credit, sale, we use the `discount` input, as a number for the formula calculation (defined in calculations.method). If you select "shipping", and this will match a shipping "Name". so ideally if you selected that type, the form input should be a dropdown with valid shipping methods.

aaronjudd:
I believe `checkout-codes` is ready for a final test, and if everyone agrees, merge to release-0.18, and thus 0.18 merge to development.  I will create another task/pr for redemption limits and conditionals as these are incomplete features, but can be tasked independently.

zenweasel:
Testing this now I enabled the "Codes" payment method and created my first code, but I don't see the prompt for code in checkout.

aaronjudd:
@zenweasel not reactive, a refresh does the trick.

spencern:
I'll try to look at this this evening. @aaronjudd  - is percentage discount is bumped for another PR with conditionals and limits?

aaronjudd:
@spencern I'm referring to percentage discounts as the method "discount", so that's already included.

zenweasel:
So I am able to put a String into the discount formula field but it throws an error when I go to checkout. `Error invoking Method 'discounts/codes/apply': Discount must be a number [400]`

zenweasel:
Not an error but I am unclear what the behavior of the "sale" discount type is supposed to be.

aaronjudd:
it's not much use right this second, but this is a discount based on a new price of each item.  (a sale price). However, add tags or product/variant filtering to that, and you have sale price per item, or per tags, or qty conditions.  (so,  product X is 5.99 instead of 9.99).

aaronjudd:
re: String as discount -> only for method type = "Shipping".  Yes, some UI validation is needed.

zenweasel:
For the "shipping" type, what would a valid value be?

aaronjudd:
dashboard

aaronjudd:
@spencern regarding the "default payment method".  I'm not even sure if this is useful anymore.  I started to remove it, but for some reason that is escaping me now, I decided to leave it. hmm

zenweasel:
Super minor item but if you use a code and then delete it, everything works great 👍  except that it then prompts you to enter another discount code in the orders dashboard panel which doesn't work.

imports/plugins/included/discount-codes/server/methods/methods.js
```diff
@@ -12,6 +13,94 @@ Discounts.attachSchema(DiscountSchema, { selector: { discountMethod: "code" } })
...
   /**
+   * discounts/codes/discount
+   * calculates percentage off discount rates
+   * we intentionally passed ids, instead
+   * of the cart,discount Object
+   * for a smaller request providing an
+   * additional level of validation.
+   * @param  {String} cartId cartId
+   * @param  {String} discountId discountId
+   * @return {String} returns discount total
```
zenweasel: JSDoc: return type is not a string

aaronjudd:
Can you break down the step to replicate that some more?  I deleted a code that in the admin, that I had already placed an order with, and then completed processing the order. That worked fine for me as well, or are you saying that you had not yet placed the order, it was still cart, with a discount applied and then you deleted? I tried that as well, and was able to place the order after I had applied the discount(but before touching anything else on the order).   I did catch one error with default where I had added a "rate" versus a discount.  "Rates" are not done at all, although I've been trying to keep that update and functional as I go with discounts.

zenweasel:
Yeah, so you have to go a _long_ way to make this be a problem. Create a discount, use it on a order then delete the discount. That all works great. The discount still displays correctly. But the "Apply a discount" prompt appears for that order in the dashboard, and if you **then** go and apply another discount, that doesn't work (it doesn't reflect the discount).

So you sort of have to work really hard to run into this problem, and even then it doesn't really cause an error or anything. You just can't apply another discount so ` ¯\_(ツ)_/¯`. I just ran into it so I thought I would note it.

aaronjudd:
that might be a ahem, feature...

Made me think though that it might be a good idea to have "archive" or "disabled" rather than deleting codes.  Even though the details are preserved on the order...

zenweasel:
One last thing is that I noticed that you can't see the name of the discount anywhere including the dashboard. It would be great to see which discount has been used on an order, especially when the amounts are not immediately clear.

zenweasel:
Approved - 5/5 Would release

spencern:
Thanks for the clarification re percentage discount vs shipping vs flat amount. I think that makes sense and UI validation will make it solid.

Re: default payment method ¯\_(ツ)_/¯ - as long as these work with other payment methods as well, that's good by me. Feels weird to have 'Codes' be a payment method though.

Basic docs would be a nice-to-have.

otherwise :shipit:

aaronjudd:
@zenweasel updated orders list summary to display multiple payment methods.. orders

PR #1629, opened 2016-12-13
Fix Safari websocket connection browser policy

jshimko:
Fixes a browser policy config that prevented ~~Internet Explorer~~ Safari from being able to make a websocket connection to the app.

It's 2016 Apple.  Get it together, man.   ;)


aaronjudd:
👍

Issue #1249, opened 2016-08-05
Build integration package for prerender.io


zenweasel:
As a developer I want to use the prerender.io service with Reaction Commerce to render Search Engine friendly HTML result for SEO.

PR #1632, opened 2016-12-14
Add Prerender.io Integration - Feature #1249

spencern:
Feature #1249 

Initializes prerender.io if PRERENDER_TOKEN and PRERENDER_HOST are set as ENV variables.

Not really sure how I could add tests for this, but let me know if there are any changes this PR needs.

Could also potentially extract this to it's own module/package but it's so simple and necessary with spiderable not being reliable that it seems like it might belong in core.


zenweasel:
@spencern Since this isn't a bug-fix, can you submit this to `release-0.18.0` instead?

spencern:
I can do that.

spencern:
Closed in favor of #1643

PR #1633, opened 2016-12-15
Fix price updates

### LINKED ISSUE #1519: Price doesn't show up in product landing page and database.
### Expected behavior

price object would show up in `Products` collection after insert product variant price.
price preview should displayed in products landing page.
### Actual Behavior

price object doesn't exist in database.
image
price doesn't have preview after published.
image
image

clone 0.17.0 repo =>  `reaction` => add a product => add option and variant => publish => price is not included.

### LINKED ISSUE #1521: Product grid pricing not respecting publish status of variants
Related to #1513, #1519 

If you archive / delete a variant, where the product grid was previously showing a range, and you archive variants, and leave only one item, the grid is still showing price range based on multiple published variants (that are now archived).

### LINKED ISSUE #1597: getProductPriceRange(productId) does not return range
### Expected behavior
Function `getProductPriceRange` - should return range of the Product
### Actual Behavior
`Range is not returned -> range == undefined even if we have correct ProductId`
In function 

```js
Template.productSettingsGridItem.helpers({
   displayPrice: function () {
      if (this._id) {
         return ReactionProduct.getProductPriceRange(this._id).range;
      }
      return null;
   },
});
```

we recieve error in `ReactionProduct.getProductPriceRange(this._id).range`  

`getProductPriceRange(productId)` is defined in `lib/api/products.js`, `getProductPriceRange` recieves correct id but doesnt returns range, that means that there is some error in core `lib/api/product.js` file

kieckhafer:
- Fix price not updating when variants are added
- Fix price not accounting for revisions of visible / notVisible revisions

Fixes #1521 
Fixes #1519
Fixes #1597


zenweasel:
Encountered this error while creating products. @kieckhafer is looking into it

```
Exception while invoking method 'products/updateProductField' Error: Match error: Failed Match.OneOf, Match.Maybe or Match.Optional validation
    at exports.check (packages/check.js:57:15)
    at [object Object].productsUpdateProductField (server/methods/catalog.js:778:5)
```

kieckhafer:
This is fixed with my latest commit.

The issue seemed to stem with the `price` object not being created when a new product is created, because the revision flow stubbed that step until the product / revision was actually `published`.

Therefore a new, unpublished product did not have a `price` object attached to it, and `products/updateProductField` was not finding the `price` object and returning this error.

This fix uses the schema to add the `price` object by default.

zenweasel:
Minor issue but price range not being calculated correctly in admin view as you add variants.

another_dumb

zenweasel:
When revisions are turned off, I am getting this when I create a new product.
another_dumb_product_and_reaction_and_inbox__4__-_brent_thebuddhalodge_com_-_the_buddha_lodge_mail_

kieckhafer:
Regarding the range not being updated, I think that's because you have that option selected, so it only shows that options price as the larger displayed price. The range would show on the variants box, which isn't fully in your image so I can't confirm if it's there. If you just select the variant, "white", the range should show up.

kieckhafer:
@zenweasel my most recent push should fix the issue where price ranges weren't updated when revision control was disabled.

zenweasel:
Approved - Tested verified fixed

PR #1635, opened 2016-12-16
Fix typo in i18n file so that error message showed key instead of text

### LINKED ISSUE #1470: Adding order email rejects bad email but it's not reflected in the UI
### Expected behavior

If you put in a invalid email address you should get a message telling you to try again
### Actual Behavior

An error in the console is thrown but nothing in the UI changes
1. Place an anonymous order
2. In the "Hello! Add an email and receive order updates." input box put an invalid address.
3. Observe that the UI does not change but an error is thrown in both the server and client console.

zenweasel:
Resolves #1470  Which was probably fixed by code by Erik and Jeremy but this fixes an issue with the error message.


mikemurray:
👍  code looks good

PR #1643, opened 2016-12-19
Prerender.io Integration - Feature #1249

spencern:
Feature #1249

* Integrates `prerender-node` npm package
* Initializes prerender.io if PRERENDER_TOKEN and PRERENDER_HOST are set as ENV variables.
* Adds an escaped fragment meta tag so web crawlers know to look for prerendered versions of pages.
* Uses `window.prerenderReady` in the products grid as an example of how to tell prerender not to capture a page until a subscription has finished.
* Sets `prerender-status-code` to `404` when rendering the `notFound` template
* Sets `prerender-status-code` to `403` when rendering the `unauthorized` template
* Docs added in [reaction-docs#146](https://github.com/reactioncommerce/reaction-docs/pull/146)

### DIFF
--- a/server/startup/prerender.js
+++ b/server/startup/prerender.js
@@ -9,7 +9,5 @@ export default function () {
     prerender.set("protocol", "https");
     WebApp.rawConnectHandlers.use(prerender);
     Logger.info("Prerender Initialization finished.");
-  } else {
-    Logger.info("Prerender Not Configured. Skipping");
   }
 }
--- a/server/startup/prerender.js
+++ b/server/startup/prerender.js
@@ -1,0 +1,15 @@
+import prerender from "prerender-node";
+import { WebApp } from "meteor/webapp";
+import { Logger } from "/server/api";
+
+export default function () {
+  if (process.env.PRERENDER_TOKEN && process.env.PRERENDER_HOST) {
+    prerender.set("prerenderToken", process.env.PRERENDER_TOKEN);
+    prerender.set("host", process.env.PRERENDER_HOST);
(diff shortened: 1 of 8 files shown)


server/startup/prerender.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default function () {
+  if (process.env.PRERENDER_TOKEN && process.env.PRERENDER_HOST) {
+    prerender.set("prerenderToken", process.env.PRERENDER_TOKEN);
+    prerender.set("host", process.env.PRERENDER_HOST);
+    prerender.set("protocol", "https");
+    WebApp.rawConnectHandlers.use(prerender);
+    Logger.info("Prerender Initialization finished.");
+  } else {
+    Logger.info("Prerender Not Configured. Skipping");
```
zenweasel: Tiniest of items, but generally I am against `info` logging when we **don't** do things

zenweasel:
So this seems to require `PRERENDER_TOKEN` and `PRERENDER_HOST` but I only have a token when I sign up and I am only seeing `PRERENDER_TOKEN` in the docs. What would be my `PRERENDER_HOST`?

spencern:
`PRERENDER_HOST` is your current domain - essentially - for us it's `getoutfitted.com` 

The npm package will attempt to determine the current host if one isn't provided, but most implementations I've seen included some enforcement of setting the host.

If this seems suboptimal or if there are use cases that include multiple domains with the same app, we could change this to be optional.

zenweasel:
Approved

PR #1655, opened 2016-12-20
select and open newly created variants

### LINKED ISSUE #1614: Adding a new Variant options expands the first options, not the one just created
### Expected behavior
When you click on the "+" button, you expect the option that you just created to be expanded.

### Actual Behavior
The first option is expanded, which must be closed before you can expand the option you just created.

1. Create a product
1. Add an option
1. Click on the "+" to create a second option
1. Observe that the option you just created is not expanded, but the first option is

This is not really a "bug" as much as just an annoying UI behavior

kieckhafer:
When a new variant option is created, we keep focus on whatever variant was previously opened.

This update closes all other variants, opens the newly created and focuses on the title field for editing.

Fixes #1614


zenweasel:
Approved - Tested, works great.

PR #1658, opened 2016-12-21
PR #1655 re-merge

### LINKED ISSUE #1614: Adding a new Variant options expands the first options, not the one just created

kieckhafer:
These changes were merged into the last release-0.19.0 branch, which I killed because there were a crazy amount of conflicts when i tried to pull `development` in. This was due to me branching from `release-0.18.0` instead of `development` initially.

This was the only PR merged into the new `release-0.19.0` branch, so once it’s pulled in we are back where we were an hour ago, with no conflicts.

Fixes #1614


mikemurray:
👍  Work for me

PR #1660, opened 2016-12-21
add check to make sure subscription is ready before displaying Shipping options

### LINKED ISSUE #1650: During checkout, shipping options display error briefly while loading
Smooth out the transitions in the checkout.

kieckhafer:
Fixes #1650 

In checkout, our “No shipping packages are configured” message was showing when our shipping subscription had not yet loaded.

This change adds a check that shows the loading circle when the subscription isn’t yet loaded, instead of that error.

### DIFF
--- a/imports/plugins/core/shipping/client/templates/checkout/shipping.js
+++ b/imports/plugins/core/shipping/client/templates/checkout/shipping.js
@@ -65,6 +65,17 @@ Template.coreCheckoutShipping.helpers({
       return "active";
     }
     return null;
+  },
+
+  isReady() {
+    const instance = Template.instance();
+    const isReady = instance.subscriptionsReady();
+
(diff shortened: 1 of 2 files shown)


imports/plugins/core/shipping/client/templates/checkout/shipping.js
```diff
@@ -65,6 +65,17 @@ Template.coreCheckoutShipping.helpers({
       return "active";
     }
     return null;
+  },
+
+  isReady() {
+    const instance = Template.instance();
+    const isReady = instance.subscriptionsReady();
+
+    if (isReady) {
```
zenweasel: could we just return `isReady`

aaronjudd:
👍  approved for merge

PR #1664, opened 2016-12-22
Variants without media to inherit media from their parents

### LINKED ISSUE #1322: Product variant must have a option to inherit picture from parent
### Expected behavior

Product variant must have a option to inherit picture from parent 

users/developers

mikemurray:
Resolves #1322

Child variants with missing media will now attempt to find media from their parents.

This applies only to the image gallery when the variant is selected.

This should lower / prevent cases where a placeholder image is shown instead of an available product image.


kieckhafer:
👍  Second level variants display top level variant image if no image is directly added to the second level variant.

PR #1667, opened 2016-12-24
Discount limit handling

### LINKED ISSUE #1662: Implement discount limit handling
Resolve limit handling requirements of #149 and #1512 

- Store discount applied counters on Discounts, after discount is used in copyCartToOrder.
- Store pending counter stats on Discounts when discount applied to order.
- Credit counters if removed from an order.
- Display invalid code notification if attempt to use in DiscountForm component

aaronjudd:
Resolve #1662 
- Store discount applied counters on Discounts, after discount is used in copyCartToOrder.
- Display invalid code notification if attempt to use in DiscountForm component
- Store counter stats in Discounts on copyCartToOrder hook


zenweasel:
If create a promo code with an account limit of 1 and a total limit of 3. I can use it once fine and then if I try again from the same account I get the "code expired" message as expected. But when I try to use the code from an anonymous account via incognito I am also getting the "code expired" message.

aaronjudd:
@zenweasel good catch, not sure what I was thinking there.. but resolved now.

kieckhafer:
Discount limiting is working as it should, so I can give this a 👍 

However, I am a bit confused about the differences between the "`Rates` table, under the Catalog dashboard card, and the `Codes` table under the Payments card. Can you go into detail about how they are different?

zenweasel:
"Codes" are for discounts associated with a coupon/discount code ("Take 10% off with the discount code PARDO"). "Rates" are just for discounts applied across the board ("20% off all knife sharpeners through Thursday"), which is why Codes is a payment type and Rates are in Catalog.

I don't believe Rates is fully developed yet.

mikemurray:
Just a few little issues I found.

**If I don't enter a number amount into the Discount field. I did `10%` at first.**

screen shot 2017-01-03 at 1 25 45 pm

The order still goes through with the invalid code "applied" ☹️  - but no actual money was discounted 😄 .

**Some help text / placeholder examples:**

dashboard

Otherwise it appears to work well.

aaronjudd:
@mikemurray agreed the input form needs to be have some more validation logic.  before "rates" really work we need a more elegant form for the conditional logic (order max, min, expiry, tags, products,etc) so I was thinking we can build some components for handling creating rules.  I really just used the blaze/autoform for now as the quickest way to create the forms, but there really needs to be some UI for "if string, default to shipping, if number use discount percentages" .

imports/plugins/included/discount-codes/server/methods/methods.js
```diff
@@ -211,11 +211,13 @@ export const methods = {
       // existing usage count
       if (discount.transactions) {
         const users = Array.from(discount.transactions, (t) => t.userId);
+        const transactionCount = new Map([...new Set(users)].map(
```
zenweasel: Oooh, fancy

zenweasel:
I'm getting something sort of weird where the first time I enter an expired promo code nothing happens (neither applied nor the expired message). If I backspace and re-enter or do anything to trigger it again it works.
cart_checkout

zenweasel:
Approved

aaronjudd:
👍

PR #1672, opened 2016-12-27
Route theme body class

lcampanis:
Add body classes for current route and theme to help with CSS scoping, so that themes can target specific pages on a route and/or theme basis.

To test this easily, just see the classes change in the body

- [ ] Has tests.
- [ ] Has docs.
- [x] Passes all tests

EDIT - 
Something I forgot to mention is that there is no DEFAULT_THEME or in other words a theme for "index" and so the theme will always be `default` for that route, until a theme can be declared in INDEX_OPTIONS or a DEFAULT_THEME is introduced.


mikemurray:
Looks good 👍

PR #1686, opened 2017-01-05
Fix console error in checkout form

kieckhafer:
We were receiving a console error in the checkout form when choosing a shipping options.

it stemmed from a combination of autoforms `firstOption` setting, which created a `null` first option, combined with our Template Helper that puts a placeholder message in the Month and Year Select boxes.

This update removes the `firstOption=false` from the payment forms, which fixes the error, and then uses the i18n translation directly in the html file, instead of in the helper.


kieckhafer:
I tried stripping i18n out, and just using a string of text, and was still seeing the error, so I think the issue was deeper than just i18n being moved.

PR #1691, opened 2017-01-06
payment credit methods

aaronjudd:
Fixes payment methods to always use method = “credit”.  To get response payment method use the transaction object.


zenweasel:
Approved

PR #1706, opened 2017-01-11
save selected gridItems to profile

kieckhafer:
Saving the selected items to `Meteor.user().profile.preferences["reaction-product-variant"].gridItems` allows us to easily recall the items a user previously had selected, and allows the actionView panel to have better context when returning to the grid view.


zenweasel:
Approved

PR #1727, opened 2017-01-18
IOS / safari flex-wrap not working #1695

### LINKED ISSUE #1695: IOS / safari flex-wrap not working
The flex-wrap IOS safari issue is pretty well known and unfortunately it seems to be having an impact of the reactioncommerce platform. 

BTW: on a Mac chrome, firefox and safari seem ok and on mobile chrome works as well. The only time I find this error is on an iPhone.

Two example and fixes I would like to share include:

Example 1 - 
Viewing the product details like in the screenshot below, the page is trying to render 2 columns which isn't wrapping. The easiest way I found to fix this is to force the flex-basis to 100% either with .flex(1 1 100%) or -webkit-flex-basis: 100%.

Solution:

.rui.items.flex .item.half {
  flex: 1 1 50%;
  @media @mobile {
    -webkit-flex-basis: 100%;
  }
}
screen shot 2017-01-07 at 11 27 59 am

Example 2 - 
Product grid...
Solution:

  // Small product: default
  &.product-small {
    @media screen and (max-width: @screen-xs-max) {
      -webkit-flex-basis: 100%;
    }
  }

screen shot 2017-01-07 at 11 38 41 am

chluke123:
Resolves #1695 (3 instance of flex-wrap issue fixed on IOS / Safari)


jshimko:
@mikemurray This is probably not even needed after your style loader work today, right?  Autoprefixer will take care of it.  We probably don't want to have to add unnecessary prefixes to the code anyway, yes?

mikemurray:
@jshimko Yea, I think they would need to be replaced by `flex-basis` instead of `-webkit-flex-basis`.

PR #1728, opened 2017-01-18
1709 Allow users to change their default currency across the app using a dropdown  

### LINKED ISSUE #1654: Sidebar behavior fixes
What happens when you transition to a new page? Does the sidebar stay open, close, or show relevant information.

lcampanis:
Credit goes to @stevaras from @artlimes.

Fixes #1654

New currency change dropdown allows users to change their default currency to any of the `enabled` currencies in the shop. 

The new currency dropdown follows the exact same code structure as the language dropdown. 

image

## Notes

Client: Changing the dropdown affects all Prices in:
  - Product Grid
  - Product Detail
  - Checkout Flow
  - Order confirmation page
  - Dashboard Orders page
  - Orders in Profile page

Server: All currencies are calculated based on the default shop's currency. 
The default currency is kept in:
  - All Order emails (code stays as is)
  - Dashboard Orders Sidebar ActionView (code stays as is)

Notes:
- `numericInputProps` was replaced with `formatPrice` for all `Orders` templates - Except `CoreOrderShippingInvoice`
- If a Locale has multiple currencies, ie. USD, USS, LocalStorage stores only the first one.

## How it works
 ```cucumber
Feature: Change currency using the currency change dropdown

Scenario: First visit by an Anonymous user
  Given an Anonymous user has no currency in `LocalStorage.currency`
  And the currency for that Locale is enabled in Reaction
  Then get `Locale.currency` for that Locale
  And set `LocalStorage.currency` to `Locale.currency`
  But if the currency for that Locale is disabled
  Then get `Shop.currency`
  And set LocalStorage with `Shop.currency`

Scenario: Anonymous user Registers as a guest 
  Given an anonymous user is visiting from the UK
  And has a currency set to `GBP` in LocalStorage 
  And registers
  Then set `user.profile.currency` based on `LocalStorage.currency`
  And display all prices in `GBP`

Scenario: Any signed-in user changes currency from dropdown
  Given any signed-in user changes the currency from the dropdown to `USD`
  Then set `LocalStorage.currency` to `USD`
  And set `user.profile.currency` to `USD`
  And display all prices in `USD`

Scenario: Any user places an order in their default currency
  Given any user places an order in `GBP`
  Then the payment should be made in `GBP`

Scenario: Seller is editing a product
  Given the seller's shop has `USD` as a default currency
  And the user has selected `GBP` as his default currency from the dropdown
  When the user is viewing the product edit page as Administrator
  Then the product display price should be in `USD`
  And all variant display prices should be in `USD`
  And when the user clicks on the pencil to edit the product variant
  Then all variant prices should be in `USD` (note here the prices are correct - but no symbol present when value is present, or when adding a new variant)
  But when the user is switching the product view to Customer
  Then all prices should be displayed in `GBP`
```
**Edit -** We find this solution flexible in the sense that sellers are now able to add products in the shop's currency, but see the product quickly in other prices and compare.

## Scenarios that need to be looked at by RC
 ```cucumber
Scenario: Any user receives an email for an order placed in their favourite currency
  Given any user placed an order in `GBP`
  And Received an invoice by email
  Then all costs, taxes and discounts should be shown in `GBP` (currently in Shop's default currency)

Scenario: From the dashboard, a Seller is viewing an order that was placed in a different currency from the shop's
  Given a shop has `USD` as a default currency
  And the order was placed in `GBP`
  Then the order should be visible in `USD` in the dashboard
  And have `GBP` display as the order's transaction currency (currently missing)
```

And a last one...
  - Invoice - PDF version of the order has NO currency symbol and the values default to the shop's currency.


lcampanis:
Need to squash this sorry

client/modules/i18n/currency.js
```diff
@@ -34,20 +63,19 @@ export function formatPriceString(formatPrice) {
     try {
       // we know the locale, but we don"t know exchange rate. In that case we
       // should return to default shop currency
-      if (typeof locale.currency.rate !== "number") {
+      if (!Match.test(userCurrency.rate, Number)) {
```
aaronjudd: Is there an advantage to using Match instead of typeof here?
Also, I think this requires adding `import { Match } from "meteor/check"` (to be on the safe side, as it's getting exported by Meteor -> but that's one of the meteor bad behaviors that we expect to see fixed)

client/modules/i18n/templates/currency/currency.js
```diff
@@ -0,0 +1,81 @@
+import { Meteor } from "meteor/meteor";
+import { Reaction } from "/client/api";
+import { Shops } from "/lib/collections";
+
```
aaronjudd: import { Match } from "meteor/check";

client/modules/i18n/currency.js
lcampanis: We thought consistency, but in this case it won't matter much. We'll put typeof back ;)

jshimko: `Match` also doesn't give you any control over how you log/handle the error.  It just throws a big stack trace.  That's not necessarily bad or wrong, but I personally prefer more customized error handling when it's not inside a server side Meteor method.

aaronjudd:
Minor UI issue on order completion, maybe not introduced here, need issue created:
screenshot 2017-01-18 08 28 34

Need issue to track the remaining items for us here, particulary a UI for enabled/disable currencies, ala Translations.

lcampanis:
@aaronjudd yes there were a few minor UI issues that were there from before, we thought they could be part of the UIX fixes and didn't attempt to fix them. We just highlighted the important ones here.

lcampanis:
Here they are:
image 

image

image

lcampanis:
Also fixing the Add Product page - It should be the Shop's default currency when adding a product in PDP. 
Fixing and pushing soon.

lcampanis:
@aaronjudd @jshimko Just pushed the fix for PDP. Description updated above with an additional more complex scenario covered. Minor UI fix needed when editing variants.
One commit to review here: ab8f944

We needed an easy way to format prices in the user's selected currency or in the default shop currency. Instead of introducing yet more react props across the app, we thought making the functions `formatPriceString` and `findCurrency` a bit more flexible.
That should be it.. Any changes let us know..

jshimko:
Looks/works good for me.

👍

PR #1742, opened 2017-01-23
[WIP] upgrades to PDP image gallery

kieckhafer:
Fixes 1615 (will add # when ready, don't want to move the card in the waffle)

- Remove onMouseLeave from gallery, so you can stay on the last hovered image, instead of the image switching back to featured when your mouse leaves

- Add 1st / featured image to the Thumbnails, instead of just showing it as the big image and disappearing when hovering over a second image


kieckhafer:
Opened #1794 to replace this PR

PR #1745, opened 2017-01-24
Silence of the Logs

zenweasel:
Out of frustration with too much logging data.  Related to #1398 

Change most (but not all) `Logger.info` to `Logger.debug`. If you disagree with a particular change just comment.

### DIFF
--- a/imports/plugins/included/payments-stripe/server/methods/stripe.js
+++ b/imports/plugins/included/payments-stripe/server/methods/stripe.js
@@ -124,7 +124,7 @@ Meteor.methods({
           response: chargeResult
         };
       } else {
-        Logger.info("Stripe Call succeeded but charge failed");
+        Logger.debug("Stripe Call succeeded but charge failed");
         result = {
           saved: false,
           error: chargeResult.error.message
@@ -183,7 +183,7 @@ Meteor.methods({
     let result;
     try {
       const refundResult = StripeApi.methods.createRefund.call({ refundDetails });
-      Logger.info(refundResult);
+      Logger.debug(refundResult);
       if (refundResult && refundResult.object === "refund") {
         result = {
           saved: true,
--- a/server/methods/email.js
+++ b/server/methods/email.js
@@ -74,7 +74,7 @@ Meteor.methods({
 
     delete settings.password;
 
-    Logger.info(settings, "Email settings updated");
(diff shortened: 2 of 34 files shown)


imports/plugins/included/payments-stripe/server/methods/stripe.js
```diff
@@ -183,7 +183,7 @@ Meteor.methods({
     let result;
     try {
       const refundResult = StripeApi.methods.createRefund.call({ refundDetails });
-      Logger.info(refundResult);
+      Logger.debug(refundResult);
```
jshimko: I know this was already like this before changing it to debug, but while we're here, do we want to consider adding more detail to logs like this?  Otherwise, that'll just output some object in the console and you won't be able to tell where it came from or why it was output.  Possibly a good candidate for the child logger API?  (See [how I did that in Launchdock](https://github.com/reactioncommerce/launchdock/blob/development/server/methods/stacks.js#L15-L19))

server/methods/email.js
```diff
@@ -74,7 +74,7 @@ Meteor.methods({
 
     delete settings.password;
 
-    Logger.info(settings, "Email settings updated");
+    Logger.debug(settings, "Email settings updated");
```
jshimko: This happens quite infrequently.  It may be worth having stuff like this at info level.

imports/plugins/included/payments-stripe/server/methods/stripe.js
```diff
@@ -124,7 +124,7 @@ Meteor.methods({
           response: chargeResult
         };
       } else {
-        Logger.info("Stripe Call succeeded but charge failed");
+        Logger.debug("Stripe Call succeeded but charge failed");
```
zenweasel: No, this is just a normal charge failure and the error is passed back up the response chain.

imports/plugins/included/payments-stripe/server/methods/stripe.js
zenweasel: Yeah, that's fair and it is frustrating getting a barf of data with no context.

zenweasel: We should probably be looking at these more closely in the future in code reviews.

PR #1810, opened 2017-02-08
moved accounts registry into new core package

### LINKED ISSUE #1808: Some dashboard panel translations aren't loading
Check to see why not all translations are loading in the dashboard menu. Could be missing, could be they aren't translated yet.

kieckhafer:
Fixes #1808 

Created a new core account package, moved registry info from server/startup into new package

Update a few dashboard icons


aaronjudd:
@mikemurray @kieckhafer sent you new LingoHub invites (and pushed translations to this branch). Updated the import files as well.  Now when testing in another language, you should be seeing a valid view of missing translations (as this branch now has all translations current).

PR #1822, opened 2017-02-13
Marketplace - Allow Seller to View and Edit their products from the homepage grid and product detail page

lcampanis:
Product Grid updated so that sellers can display and edit their products from the homepage grid.
Also fixed permissions so that new ProductDetailPage works for sellers as well.

You will notice `Reaction.hasPermission` has the "sellers" check commented. 
This is TBD, but in essence the code always returns true for sellers, even when the product does not belong to the seller, which is an issue. With the new fix in place, now `hasPermission` needs to be called explicitly with the `userId` and `group` as arguments so that a seller's group is also in the checks.

Regarding cross-importing between included and core - To avoid having to import the `SellerShops` client-only collection from Marketplace directly, we have put it along with the core collections, and initialised only on Meteor.isClient.

To test, simply do:
1. Register as guest
2. Become a seller
3. Change your shop settings
4. Add a product
5. Edit product from the homepage grid
6. Edit product from PDP

We have tested the cart/payment/order flow for the owner (logged in buying his own product) and guest (on owner's product) and it works as expected, with Shippo and Stripe. 

A few general things to iron out especially on how the "Edit mode" toggle works. But as soon as the new UI is in place, these should be quick to fix. 

Any changes let us know.

### DIFF
--- a/client/modules/core/main.js
+++ b/client/modules/core/main.js
@@ -87,7 +87,14 @@ export default {
    * @return {Boolean} Boolean - true if has permission
    */
   hasPermission(checkPermissions, checkUserId, checkGroup) {
-    let group = this.getSellerShopId();
+    let group;
+    // default group to the shop or global if shop isn't defined for some reason.
+    if (checkGroup !== undefined && typeof checkGroup === "string") {
+      group = checkGroup;
+    } else {
+      group = this.getSellerShopId() || Roles.GLOBAL_GROUP;
+    }
+
     let permissions = ["owner"];
     let id = "";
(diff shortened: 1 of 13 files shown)


client/modules/core/main.js
```diff
@@ -116,7 +123,7 @@ export default {
       if (Roles.userIsInRole(userId, permissions, group)) {
         return true;
       }
-      // global roles check
+      /*// global roles check
```
zenweasel: ESlint complains there's not a space after the comment mark

PR #1833, opened 2017-02-15
Resolve access-denied error on non-admin checkout

zenweasel:
Method being called by `copyCartToOrder` (which is called by `cart/submitPayment` on the client-side) requires "order" permissions which non-admins don't have. Moved it to a non-Meteor method (so that it can't be called client-side) and removed the security check.

### DIFF
--- a/server/methods/core/cart.js
+++ b/server/methods/core/cart.js
@@ -3,7 +3,6 @@ import { Meteor } from "meteor/meteor";
 import { check } from "meteor/check";
 import * as Collections from "/lib/collections";
 import { Logger, Reaction } from "/server/api";
-import { ordersInventoryAdjust } from "/server/methods/core/orders";
 
 /**
  * quantityProcessing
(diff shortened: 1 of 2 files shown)


kieckhafer:
Looks like the inventory updates were broken because of revision control... there is a revision control hook that stops updates to the products collection, and this was triggered by our attempted update here. This is why no error was shown, because it was never running the update. When we skip revision control (which we should on inventory), we are seeing the same error as before, because an anon user doesn't have permission to update the products collection.

We think the solution is to probably move the inventory update into an admin section, perhaps when the order payment is actually captured, instead of right when the order is confirmed.

zenweasel:
Not arguing, but unclear how just removing the security check rebreaks inventory

mikemurray:
Non admin users are denied access to update the products collection in `/server/security/collections.js`. Maybe that causing some of the issues.

PR #1839, opened 2017-02-15
Fix accounts import

mikemurray:
- Also prevent Accounts (meteor) from colliding with Accounts (reaction
collection)


server/methods/accounts/accounts.js
```diff
@@ -1,3 +1,4 @@
+import { Accounts as MeteorAccounts } from "meteor/accounts-base"
```
zenweasel: Should go after the npm imports

zenweasel:
Verified fixed

PR #1847, opened 2017-02-16
Updated Shippo and rate handling in order fulfillment

### LINKED ISSUE #1778: add retry for failed shipping methods
In case a shipping methods fails to return tracking information (capture error),  create a UI element that can be used to re-trigger the shipping method request.

### LINKED ISSUE #1785: Tracking number is not automatically populated while using Shippo
In release-0.19.0

Shippo tracking should be automatically generating and tracking number and applying it to the related order. This is no longer working.

aaronjudd:
Resolves #1778
Resolves #1785

Adds a "Generate Rates" button for whenever a flat rate hasn't been used.  Adds method "shipping/status/refresh" that is a stub to be observed to trigger updates (in the case of multiple packages that need to refresh rates).   Adds updated handling for flat rate editing, to be aware of other shipping views.  Adds print label into the shipment section.  Update UI elements in shipping details.

To test, walk through both a flat rate, and a shippo calculated rate checkout, then fulfillment.


aaronjudd:
WIP removed, @kieckhafer created issue for a more elegant shipping loader #1872

PR #1876, opened 2017-02-22
[WIP] Add line item return UI

### LINKED ISSUE #1849: Add UI for line item returns in Order dashboard
We need to have a way to do returns/refunds by line item rather than just at the total order level. Currently there is no way to do that in the Orders dashboard.

Related to #1793

zenweasel:
Resolves #1849


zenweasel:
Deferring this till we have more design direction

PR #1883, opened 2017-02-22
Migrations for package registry

aaronjudd:
- deletes, then re-adds registry entries for version 0.19
- does not update settings

To test: 
checkout master, reaction reset, check settings. git checkout development, do not reset.  `reaction update`, then `reaction` . Migrations should run, and revised registry entries should be used. This should also work for all custom packages.
reaction


aaronjudd:
The downside of this, is that every new install is effectively loading packages, then deleting the registry again as the migration runs.  Be better to add some version checking, to only run migrations when actually needed.

Issue #1787, opened 2017-02-06
Fix sold-out issue on search grid


kieckhafer:
Search grid displays "Sold Out" on all products, no matter their inventory.

Related to #1776

aaronjudd:
![reaction](https://cloud.githubusercontent.com/assets/439959/23243016/09d4772e-f931-11e6-83ba-984576e0e05e.png)

Should released earliest possible.

PR #1889, opened 2017-02-23
update search results to use their own notice / badges template

kieckhafer:
Create a new `searchGridNotice` template to be used in our search results modal to display badges for low inventory and sold out products.

Currently, the search grid shares a template, `gridNotice`, with our main grid view. When navigating to a PDP page, we lose the publication data for all products, and only receive the current PDP product data, which was causing display issues in search.

This update creates a new template to be used only on the search results page, which uses new data output through the search publication.

This update will temporarily cause no badges to display in our search results, which will be remedied in `release-0.20.0` by denormalizing inventory data and attaching it to the top level product.


aaronjudd:
👍  this is a good patch for #1787, let's go!

PR #1870, opened 2017-02-21
Avalara Tax Compliance

zenweasel:
Resolves: #1793 
Resolves: #1860 
Resolves: #1699 

One pending ticket required for compliance is #1857 but it can be PR'd in separately.

#### With Avalara **disabled**

1. Ensure that everything still works
1. Ensure that there is a "Tax Code" field and fields for dimension in the PDP. The Tax Code field should be a simple text input
1. Ensure that no extra fields show up in Accounts/Manage

#### With Avalara enabled (the API info should be in the WIki and contact me for Company code)

1. You should see a dropdown in the PDP for tax code
1. If you navigate to Accounts then Manage, you should be able to enter exemption number or select from a customer entity type.
1. In the Avalara configuration panel you should be able to "Enable Logging" and a panel should appear with all transactions logged and the ability to select an individual one and select it.
1. You should be able to go through the entire checkout and order processing flow (including refunds) without problems
1. These transactions should be reflected in the Avalara "transactions" view
1. When a exemption number or customer entity type is supplied, their transactions should be flagged as tax exempt.

PR #1865, opened 2017-02-18
Allow packages to override Meta Title from the registry

lcampanis:
Introduced a new `register.$.meta` to allow packages to override the default page title, currently being the package name alone.

To test it easily just add the following to any `register.js` route and see the title change.
```js
meta: {
  title: "My new title"
}
```

Example:
```js
registry: [{
    label: "Notifications",
    name: "notifications",
    route: "/notifications",
    meta: {
      title: "Notification Title Updated"
    },
    workflow: "coreWorkflow",
    permissions: [{
      label: "Notifications",
      permission: "notifications"
    }],
    template: "notificationRoute"
  }]
```
Then visit http://localhost:3000/reaction/notifications

PR #1893, opened 2017-02-23
Release 0.20.0 (Working Branch)

kieckhafer:
Release 0.20.0 will focus on bug fixes and polish.

https://waffle.io/reactioncommerce/reaction?milestone=v0.20.0

# Major Features
- Added UI to allow switching between templates on the Product Detail Page ( #1909 )
- Avalara Tax integration ( #1870 )

## Other Features
- New `meta` key in registry ( #1865 ) 
-- Allow package registry to override default meta title ( #1865 ) 
-- Allow packages to determine the size of their `actionView` panel
- Better organization of localization options ( #1947 )

## Other Changes / Fixes
- Convert various dashboard panels to React ( #1848 , #1854 , #1947 )
- Add switch component to list item in dashboard ( #1936 )
- Various Lint fixes ( #1937 )
- Add feedback / loading spinner when waiting on refund data ( #1926 )
- Update product handle when title changes ( #1898 )
- Fix Sold Out / Low Inventory warning always showing on products in search view ( #1904 )
- Add context to actionView / sidebar throughout checkout process ( #1932 , #1956 )
- Remove inventory subscription that is not in use (  #1951 )
- Get package data by shopId, in addition to package name ( #1949 )
- Localization of currency in order emails ( #1957  )

## Breaking Changes
- #1914 - We have removed `react-bootstrap` from our `npm` packages. If you need `react-bootstrap`, you can re-add it to your dependencies.

## Contributors

@reactioncommerce/core
@reactioncommerce/engineering 
@lcampanis 
@hrath2015


zenweasel:
When I create a new product and click to edit the variant I am getting a 404

aaronjudd:
When editing variants, I am not seeing what @zenweasel sees (I think), but I am seeing a console error 

Browser console:
```
Exception in delivering result of invoking 'avalara/getTaxCodes': Error
```

Server console:
```
00:19:34.676Z FATAL Reaction: The Avalara package cannot function unless username is configured
00:19:34.680Z FATAL Reaction: The Avalara package cannot function unless password is configured
00:19:34.687Z FATAL Reaction: The Avalara package cannot function unless apiLoginId is configured
00:19:34.693Z FATAL Reaction: The Avalara package cannot function unless companyCode is configured
00:19:34.702Z FATAL Reaction: The Avalara package cannot function unless shippingTaxCode is configured
```

We should not be spitting out Fatal error just because a package is not configured. We should always have a graceful fallback.

zenweasel:
@aaronjudd So is the issue just that we should change the logging level?

aaronjudd:
On PDP Tax UI, `Origin Country` should use the countries collection / provide a dropdown for the product origination. I very much doubt people will want to type in Country every time, and also doubt they'd type the right country every time.

zenweasel:
I have `reaction reset` a couple of times and I see the 404 every time when I try to edit the variant on a newly created product.

### Steps to reproduce

1. `reaction reset -n`
1. Select "Add Product" from the dropdown
1. Enter data for Title, Subtitle, Description, and Vendor.
1. Click on the pencil icon next to variant where it says "Label"
1. Observe the "Ooops! This is not the product you were looking for" message.

aaronjudd:
@zenweasel re: product variant, ok now it's reproduceable.  the url that it's going to is `http://localhost:3000/reaction/product/<new product title>/id`. I think that in this case <new product title> has not yet been published.

zenweasel:
Not a blocker or a big deal, but I think currencies should be sorted in the currency enable/disable panel. Just because they appear _mostly_ sorted so it's confusing.

zenweasel:
Visual corruption when clicking on the "Base Currency" dropdown in the localization panel
dumb-product

zenweasel:
Same for Unit of Measure
dumb-product

aaronjudd:
re: avalara errors.  The error level doesn't feel correct at FATAL,  [our logging docs](https://docs.reactioncommerce.com/reaction-docs/master/logging) imply that these should likely be WARN, because in the case of adding to the PDP data, it's not likely to cause any specific PDP error. More generally the fact that that Avalara is enabled, but not configured might warrant an ERROR, and a client warning somewhere, just not sure that the PDP is the place for it.

zenweasel:
@aaronjudd re: avalara Understood. I had hoped that we could utilize the `Logs` collection to somehow surface global errors to admin. Making logging some sort of `configuration-error` shows a global warning in the dashboard. `Next`

I do think the PDP is one place where they should be notified that Avalara is not configured correctly, because they will not be able to assign the correct Tax Code. I will work on making that messaging more clear and context-appropriate.

zenweasel:
I am still seeing this error when placing more items in the cart than are in stock
`skipped bulk operations backorder updates.`

zenweasel:
There are still 6 EsLint warnings on this branch, but they are React props errors , so the someone else who is more familiar with the code should fix them. (silly Bithound)

zenweasel:
The "Email logs" panel collapses every time there is a new update. Probably not a **huge** deal, but if you had a busy system and were wanting to deal with emails it would be a drag.

zenweasel:
Also, the Email Log filter doesn't seem to work, or at least I can't figure out how it's supposed to work

zenweasel:
Service says "Not Set" when it should say "Custom". Status shows as red even though I am sending emails fine.

basic_reaction_product

imports/plugins/included/product-detail-simple/lib/layout/twoColumn.js
```diff
@@ -30,11 +30,23 @@ export default function twoColumn() {
       ]
     },
 
-    // Variant block
+    // Product Details Block
+    // Contains
+    // - Title
+    // - Sibtitle
```
zenweasel: Subtitle?

zenweasel:
If I go into the SMS settings panel and save settings. I get the "Saved Successfully" message, but the button never goes back to "Save" so I can't save any more changes.

basic_reaction_product

server/api/core/core.js
```diff
@@ -203,10 +206,22 @@ export default {
     return settings.settings || {};
   },
 
+  getShopCurrency() {
```
zenweasel: Did you intend to add this function twice? If so, see my comment on other instance.

server/methods/core/cart.js
```diff
@@ -551,6 +562,36 @@ Meteor.methods({
...
+    const currentUser = Meteor.user();
+    let userCurrency = Reaction.getShopCurrency();
+    let exchangeRate = "1.00";
+
+    if (currentUser && currentUser.profile && currentUser.profile.currency) {
+      userCurrency = Meteor.user().profile.currency;
+    }
+
+    if (userCurrency !== Reaction.getShopCurrency()) {
+      const getExchangeRate = Meteor.call("shop/getCurrencyRates", userCurrency);
```
zenweasel: I think normally we would only want to use names that start with `get` or `set` for function, not a variable.

server/methods/core/cart.js
```diff
@@ -702,6 +743,65 @@ Meteor.methods({
...
+   */
+  "cart/setUserCurrency": function (cartId, userCurrency) {
+    check(cartId, String);
+    check(userCurrency, String);
+    const cart = Collections.Cart.findOne({
+      _id: cartId,
+      userId: this.userId
+    });
+    if (!cart) {
+      Logger.error(`Cart not found for user: ${ this.userId }`);
```
zenweasel: I'm seeing this error as `Cart for not found user: null` in the logs a few times. Seems like this should not normally happen?

server/methods/core/cart.js
```diff
@@ -702,6 +743,65 @@ Meteor.methods({
...
+      throw new Meteor.Error(404, "Cart not found",
```
zenweasel: I think we are trying to get away from using HTTP codes here in errors?

server/methods/core/cart.js
aaronjudd: Agreed, this should not happen.

zenweasel: This looks like it is also breaking some tests
```
I20170311-17:31:01.406(8)?   1) Account Meteor method  addressBookAdd should disabled isShipping/BillingDefault properties inside sibling address if we enable their while adding:
I20170311-17:31:01.406(8)?      Error: Cart not found [404]
I20170311-17:31:01.406(8)?       at [object Object].Meteor.methods.cart/setUserCurrency (server/methods/core/cart.js:761:13)
I20170311-17:31:01.408(8)?       at packages/check.js:130:16
```

zenweasel:
Changed "Origin Country" from text to select in PR: https://github.com/reactioncommerce/reaction/pull/1967

mikemurray:
@zenweasel fixed the issue with the variant 404 with new product handle: https://github.com/reactioncommerce/reaction/pull/1893#issuecomment-285813920

imports/plugins/core/email/client/components/emailLogs.js
```diff
@@ -0,0 +1,132 @@
...
+
+    // helper adds a class to every grid row
+    const customRowMetaData = {
+      bodyCssClassName: () =>  {
+        return "email-grid-row";
+      }
+    };
+
+
+    const ColumnData = React.createClass({
```
mikemurray: and also, don't create a class in a class. That needs to be moved to its own file.

imports/plugins/included/product-detail-simple/lib/layout/twoColumn.js
mikemurray: fixed

mikemurray:
Sorta fixed. https://github.com/reactioncommerce/reaction/pull/1893#issuecomment-285825610

While the dropdown will no longer overlap one another, the arrow for the panel will still be above the dropdown in some cases.

I removed the z-index that @kieckhafer and I added to fix this, as it causes more issues than it resolves.

client/modules/core/main.js
```diff
@@ -243,6 +243,14 @@ export default {
     return settings.settings || {};
   },
 
+  getShopCurrency() {
+    const shop = Shops.findOne({
+      _id: this.shopId
+    }) || {};
+
+    return shop.currency;
```
kieckhafer: How do we handle this usually then? I just reference the function right above it which does the same thing:

```
getShopSettings() {
    const settings = Packages.findOne({
      name: "core",
      shopId: this.shopId
    }) || {};
    return settings.settings || {};
  },
```

server/api/core/core.js
kieckhafer: Yeah, one for server, one for client.

zenweasel: Got it. Makes sense.

client/modules/accounts/templates/profile/profile.js
```diff
@@ -1,6 +1,7 @@
 import * as Collections from "/lib/collections";
```
kieckhafer: Fixed in PR #1973

client/modules/core/main.js
kieckhafer: Fixed in PR #1973

client/modules/i18n/templates/currency/currency.js
```diff
@@ -74,5 +74,10 @@ Template.currencySelect.events({
     //
     Meteor.users.update(Meteor.userId(), { $set: { "profile.currency": this.currency } });
     localStorage.setItem("currency", this.currency);
+
+    const cart = Cart.findOne({ userId: Meteor.user()._id });
```
kieckhafer: Fixed in PR #1973

server/api/core/core.js
kieckhafer: Fixed in PR #1973

server/methods/core/cart.js
kieckhafer: Fixed in PR #1973

server/methods/core/cart.js
kieckhafer: Fixed in PR #1973

server/methods/core/cart.js
kieckhafer: Fixed in PR #1973

PR #1898, opened 2017-02-24
Update product handle when title changes

### LINKED ISSUE #1648: Product URL is not being set automatically by product title
### Expected behavior
When a new product is created, and a title is added, the product URL slug should update to be based off of the product title.

### Actual Behavior
The URL slug does not update automatically

Add a new product
See URL slug is not updated.

screen shot 2016-12-19 at 5 06 09 pm

### LINKED ISSUE #1887: Product Title is required error on new products.
Warning on PDP. 
ugpjenviupkdjk8uv

Perhaps related to #1605.  Doesn't however seem to cause the product to be un-editable.

To replicate, add new product, click to edit, view console. 

```
Error invoking Method 'products/updateProductField': Product Title is required [400]
```

mikemurray:
Resolves #1648 
Resolves #1887

Ensure the handle is always set to the slugified product title, custom
input from user, the product id.

**If handle is identical to the product id (new product)**
Handle will become the slugified title if one exists

**If handle is a slug that is not identical to the product id (modified by user)**
Handle will stay the same.

**If handle is blank (deleted by user)**
Handle becomes a slug of product title if one exists, else becomes the product id


zenweasel:
Maybe I am misunderstanding what is happening here, but if I go to the "Basic Reaction Product" and remove the permalink it changes into the slugified title (cool). But if I go to the grid and try to navigate to via that new handle I get a 404.  So basically I can never get back to the that product again either via the new handle or the old.

zenweasel:
If I make this change (remove the handle so it changes in place) and then hit "publish" the page changes to a 404 because the url in the url bar is still the old one.

zenweasel:
Something is not happening reactively here. If I change the permalink, then publish, then go to the grid and click on the product it tries to go to the old URL. If I open a separate tab it's updated.

imports/plugins/core/revisions/server/hooks.js
```diff
@@ -447,6 +450,64 @@ Products.before.update(function (userId, product, fieldNames, modifier, options)
...
+            const slugDocId = getSlug(documentId);
+            const revisionTitle = productRevision.documentData.title;
+            const revisionHandle = productRevision.documentData.handle;
+
+            // Checks
+            const hasNewHandle = _.isEmpty(newHandle) === false;
+            const hasExistingTitle = _.isEmpty(revisionTitle) === false;
+            const hasNewTitle = _.isEmpty(newTitle) === false;
+            const hasHandle = _.isEmpty(revisionHandle) === false;
+            const handleMatchesId = revisionHandle === documentId || revisionHandle === slugDocId || newValue === documentId || newValue === slugDocId;
```
zenweasel: line length?

imports/plugins/core/revisions/server/hooks.js
```diff
@@ -447,6 +450,64 @@ Products.before.update(function (userId, product, fieldNames, modifier, options)
...
+              // Set the handle to be the slug of the product.title
+              // when documentId (product._id) matches the handle, then handle is enpty, and a title exists
+              revisionModifier.$set["documentData.handle"] = getSlug(newTitle || revisionTitle);
+            } else if (hasHandle === false && hasExistingTitle === false) {
+              // If the handle & title is empty, the handle becomes the product id
+              revisionModifier.$set["documentData.handle"] = documentId;
+            } else if (hasNewHandle === false && property === "handle") {
+              // If the handle is empty, the handle becomes the sligified product title, or document id if title does not exist.
+              // const newTitle = modifier.$set["title"];
+              revisionModifier.$set["documentData.handle"] = hasExistingTitle ? getSlug(newTitle || revisionTitle) : documentId;
```
zenweasel: line length?

mikemurray:
@zenweasel Updated with refresh on handle change. Uses a `location.href` because `Router.go` / `Router.refresh` doesn't seem to want to refresh the current view.

As for the grid not being reactive, well it is sometimes, sometimes not (probably another ticket). With the `location.href` mentioned above it should be fixed as you end up reloading everything.

zenweasel:
Sometimes when I try to set the URL, it's resetting to the product title slug.

product_url

zenweasel:
For what it's worth, changing the URL and hitting publish now gives you a 404 on the current development

mikemurray:
I can't seem to pinpoint whats causing it since I've only seen it twice in 50 tries.

One of those times was on the `Basic Example Product`. For that one, it may be related to some weirdness I've seen where the product fixture data get partially re-imported every so often. Then when you try to change the handle, it sometimes reverts back to `example-product`.

With that said, I've made some changes to the checks in hopes it makes it more consistent.

PR #1902, opened 2017-02-27
Fix i18next-jquery error that causes app to not work

### LINKED ISSUE #1908: Cannot set property 't' of undefined. Labels missing
I am brand new to Reaction. Really eager to try it out. I got everything installed and running. However, when i navigate around I get the below console error.

image

### Expected behavior
No javascript errors

### Actual Behavior
Uncaught TypeError: Cannot set property 't' of undefined ----- modules.js?hash=c4f639d…:53710

Fresh install

users/developers

kieckhafer:
Update the way we use`i18next-jquery` to fix an issue caused by the recent update to `i18next` version `7.1.0`.

Issue opened here: https://github.com/i18next/jquery-i18next/issues/29

Fixes #1908


zenweasel:
:+1 tested again. Also works.

PR #1931, opened 2017-03-02
Add dimensions/tax code to PDP

### LINKED ISSUE #1699: PDP UX for Shipping, Taxes
For implementation of #1618,  we need UX for designed for the following information on product details page:

Dimensions [Width (x) Length, Depth]
Customs Category (select from list)
Tax Code Category  (select from list)

@rymorgan this is likely some additional fields, in the same area vein as the existing inventory, pricing.

kieha:
Resolves #1699

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -96,3 +96,5 @@ johanbrook:publication-collector
 
 
 # Custom Packages
+aldeed:autoform-select2
+natestrauser:select2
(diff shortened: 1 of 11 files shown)


mikemurray:
Theres a lot of space between form fields, looks like they're all wrapped in `panel-body`. You should only need one, with all the fields inside of it.

untitled_product

PR #1947, opened 2017-03-07
Updated i18n settings panel

### LINKED ISSUE #1773: Add enable/disable currency settings under localization
Since the drop-down, just a few currencies are disabled by default, but we actually a support a ton of currencies, so shop operators should be able to configure which ones they want to support.

### LINKED ISSUE #1797: Localization Sidebar Fix
Put all languages is in the card component. 
screen shot 2017-02-07 at 11 23 23 am

mikemurray:
Resolves #1797 
Resolves #1773 

- added new package for i18n settings
- converted to react
- added currency list
- added ability to save open state to SettingsCard to user preferences
- added ability to pluck fields in form component to enable concise
forms


kieckhafer:
Should we have the all off / all on toggle?

kieckhafer:
Base Currency dropdown should filter our all disabled currencies: 

reaction

imports/plugins/core/i18n/client/components/localizationSettings.js
```diff
@@ -0,0 +1,182 @@
...
+                type: "select",
+                options: this.props.currencyOptions
+              },
+              baseUOM: {
+                type: "select",
+                options: this.props.uomOptions
+              },
+              language: {
+                type: "select",
+                options: this.props.currencyOptions
```
hrath2015: `this.props.currencyOptions`  should be language Options.

hrath2015: Also when you try to select value(Base Language) from available options, expandable arrows from below 2 cards (Languages and Currency) are visible. See the  attached pic. 
capture

kieckhafer:
@hrath2015 you're right, fixing now

Issue #1382, opened 2016-09-14
Only subscribe to Inventory if Inventory package is installed


zenweasel:
Inventory should be an optional package but the app throws an error if the package is removed

`Error invoking SubsManager.subscribe("Inventory"):  Subscription 'Inventory' not found`

hbakhtiyor:
i deployed newest master branch to heroku and got the error
how to resolve it?

hbakhtiyor:
found the same issue https://github.com/reactioncommerce/reaction/issues/1363

zenweasel:
If you are not using the inventory plugin you can just comment out that line as a temporary workaround.

andela-oadebayo:
@zenweasel  am having this same issue, where will i comment out the inventory plugin line, can't seem to find it. Thank you

zenweasel:
It's right here: https://github.com/reactioncommerce/reaction/blob/da4a379586c347a39adb3984d842490a210c7392/client/modules/core/subscriptions.js#L41-L41

vigyano:
commenting the subscribe inventory does not work, just a blank page is showing up if master, development, and other branches

zenweasel:
@aaronjudd This can be closed can't it?

PR #1951, opened 2017-03-08
Reduce Inventory records subscribed to by admin

### LINKED ISSUE #1370: Reduce number of inventory records subscribed to by admin
By default if you are logged in as an admin you are subscribed to every individual inventory product, which for a store with a lot of inventory can mean being subscribed to tens of thousands of records which on a live store are constantly changing. This has caused performance problems for some sites.

Probably this can be changed to subscribe to an aggregate of each products quantity in each workflow status. (available, reserved, etc)

kieha:
Resolves #1370 

I went through the app and realized that the Inventory subscription wasn't being used (client-side), and it was affecting app performance. This resolves that issue.

### How to test

Add quantity of products. This should not affect the app's performance.

### DIFF
--- a/client/modules/core/subscriptions.js
+++ b/client/modules/core/subscriptions.js
@@ -36,10 +36,6 @@ Subscriptions.Tags = Subscriptions.Manager.subscribe("Tags");
 
 Subscriptions.Media = Subscriptions.Manager.subscribe("Media");
 
-// admin only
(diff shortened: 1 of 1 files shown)


zenweasel:
This also closes: https://github.com/reactioncommerce/reaction/issues/1382 I believe

PR #1967, opened 2017-03-12
Add origin country dropdown

zenweasel:
Change Origin Country from text field to Select


aaronjudd:
Is this supposed to add a DD, vs select on the PDP? I don't see it.

asd_and_reaction___waffle_io

PR #1979, opened 2017-03-14
Remove code which uses no longer defined event

### LINKED ISSUE #1845: Endless spinner when going to the product grid in Firefox
### Expected behavior
The product grid should always render

### Actual Behavior
If you are an admin user and go to the Dashboard and then try to return to the product grid you an error in the console and the product grid never renders

1. Login in as admin (in FF)
1. Go to the dashboard
1. Return to the product grid by clicking on "Reaction" in the upper left.
1. Observe endless spinner
1. Observe error in console.

kieckhafer:
Fixes #1845 

Remove an old reference to an event which is never used anymore, and breaks firefox

### DIFF
--- a/imports/plugins/included/product-variant/client/templates/products/productGrid/productGrid.js
+++ b/imports/plugins/included/product-variant/client/templates/products/productGrid/productGrid.js
@@ -11,7 +11,7 @@ import Sortable from "sortablejs";
  */
 
 Template.productGrid.onCreated(function () {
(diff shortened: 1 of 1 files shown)


aaronjudd:
BH says: https://www.bithound.io/github/reactioncommerce/reaction/blob/aa9399bdd9c3e069c6684e7ace7df328a31e1a11/imports/plugins/included/product-variant/client/templates/products/productGrid/productGrid.js#filter-lint

PR #1989, opened 2017-03-15
add Route to onClick for search

### LINKED ISSUE #1959: Clicking on search results doesn't take you to product detail
### Expected behavior
When you click on a product in the search results it should take you to that product detail pager

### Actual Behavior
Seems to do nothing? Notice that the URL never changes from `tag/shop`

search_detail

1. `reaction reset -n`
1. Browser to home page in incognito window
1. Click on Search icon
1. Search for "basic"
1. Click on product
1. Observe you are not taken to detail page

Just to make this more interesting, a couple of times when I first stumbled on this it did work. But now it never seems to work for me.

kieckhafer:
Fixes #1959 

Adds Route to onClick event of search item

### DIFF
--- a/imports/plugins/included/ui-search/client/templates/productSearch/productItem.html
+++ b/imports/plugins/included/ui-search/client/templates/productSearch/productItem.html
@@ -11,7 +11,7 @@
         data-event-category="grid"
         data-event-action="productClick"
         data-event-label="grid product click"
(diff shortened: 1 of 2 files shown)


aaronjudd:
OK so, yes, this seems to have resolved getting the PDP when clicking the search result.

Maybe as a new issue, these views seem like they could use some love... 
accounts
orders

PR #1996, opened 2017-03-15
Tax code field styling in PDP 

### LINKED ISSUE #1988: Tax Code field not styled correctly when populated
### Expected behavior
All fields should have consistent styling

### Actual Behavior
When populated, Tax Code field is not using the proper styling.

basic_reaction_product

1. Enabled a tax provider that provides tax codes (e.g. Avalara or TaxCloud)
1. Go to PDP
1. Observe Tax Code field

Seen in PR #1974

kieha:
Fixes #1988 

Copied over CSS attributes for the other fields into the tax codes field for consistency purposes.

### How to test
1. Enable tax provider that supplies tax codes (Avalara or TaxCloud)
2. Head over to PDP
3. Observe that the fields are consistently styled

### DIFF
--- a/imports/plugins/included/default-theme/client/styles/products/variantForm.less
+++ b/imports/plugins/included/default-theme/client/styles/products/variantForm.less
@@ -29,3 +29,17 @@
 .select2-container {
   width: 370px;
 }
+.select2-container--default .select2-selection--single {
+  border-radius: 2px;
+  border: 1px solid #cccccc;
+  box-shadow:inset 0 1px 1px rgba(0, 0, 0, 0.075);
+}
+.select2-container .select2-selection--single {
+  height: 34px;
+}
+#select2-taxCode-container {
+font-size: 14px;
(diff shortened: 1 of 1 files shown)


zenweasel:
Still not exactly right. 

basic_reaction_product

basic_reaction_product

I'm being super nit-picky and if @mikemurray or @rymorgan want to say it's ok then it's okay with me.

kieha:
@zenweasel should I match the field to the other dropdowns in the PDP, or to the fields on the page that it's on? I was going for the latter.

zenweasel:
@kieha Is there a case where that is not the same thing? There shouldn't be.

kieha:
By other fields I mean something like this:

screen shot 2017-03-16 at 3 15 31 am

So the tax code field looks like the tax description field, and they are both in the same section, which is the product variant form page. But not like the origin country field you pasted above.

PR #1997, opened 2017-03-15
Use JSON format for trace log level

### LINKED ISSUE #1398: elevate current default logging levels

impactmass:
Resolves #1398

To test:
- Nothing to see on the UI, since this is about logging
- Start the app using `reaction`, confirm that the usual logs still work as before
- Restart the app with `REACTION_LOG_LEVEL` set as `TRACE`, check that trace logs appear as raw json. (you'll get loads of logs, as you'll now be seeing all debug and trace logs).

NB: Setting log format for `trace` as json, makes Bunyan use same for `debug` (so debug logs also appear as json).


impactmass:
Ok. Just got back to this ticket. Taking a look at the issue.

impactmass:
Initial thought of having "short" for debug and "json" for trace doesn't work in same stream. 
Can't have both for same stream.... that's why there was double logs.

Fixed that now.

PR #1998, opened 2017-03-15
Fix form styles

### LINKED ISSUE #1982: AutoForm styling (payments)
Seen in #1974 

Some of the AutoForm inputs in the dashboard are missing styles. See payments.

basic_reaction_product

I think this was previously resolved with a global style patch that was applied elsewhere.. (but not by reattaching class="form-control").

mikemurray:
Resolves #1982 

Reintroduced form styles for auto form fields with a selector to ignore
any `rui` from fields.


zenweasel:
Are things like these supposed to be fixed as well?

reaction

aaronjudd:
@zenweasel they "should" be -> since this is a regression and those were handled as well before..

aaronjudd:
@zenweasel should be ok now..

Issue #1969, opened 2017-03-13
Can no longer override workflow/layout structures from a plugin


zenweasel:
### Expected behavior
Able to override an existing workflow/layout combo by adding a `layout` record in the registry.

### Actual Behavior
Layout record is written to the Shops collection, but the original record is always selected because it's first in the collection.

### Steps to Reproduce the Behavior
1. Clone the example-plug into `import/plugins/custom`
1. `reaction reset`
1. Observe that the new layout/workflow structure is not used.

I believe at one point there was a `reverse` on Shops.layout` so that custom layouts got loaded first but that appears to be gone and I have a recollection of trying that again and it broke other things.

We probably need a way to "merge" layout records where custom ones just take the place of the stock ones.

gepolv:
There might be another reason causing this issue.

Inside this function:
```
export function ReactionLayout(options = {}) {
   const layout = options.layout || Session.get("DEFAULT_LAYOUT") || "coreLayout";
   const workflow = options.workflow || Session.get("DEFAULT_WORKFLOW") || "coreWorkflow";
```
I found "options.workflow" already had a value which was not "coreWorkflow". Then later on, it caused this line to pick up a wrong layout:
`shop.layout.find((x) => selectLayout(x, layout, workflow)) `

impactmass:
I was able to reproduce this by the steps listed. 
Next steps is looking at possible solutions starting with the hints added on the issue, while I also need to go through the docs to get more understanding on layouts and workflows. More update later..

gepolv:
I have a pull request to fix this issue, please take a look:
https://github.com/reactioncommerce/reaction/pull/2003

aaronjudd:
@zenweasel now that priority field has been added to layouts, is this still an issue?

zenweasel:
This was fixed via #2023. Now there is a problem the example plugin product grid, but that's a separate issue.

PR #2003, opened 2017-03-16
[WIP] change default "INDEX_OPTIONS"

gepolv:
change "INDEX_OPTIONS" from "coreProductGridWorkflow" to "coreWorkflow".
Reason:
when the root path "/" is rendered in "ReactionLayout",  it will look for combination of layout and workflow to retrieve template from collection "Shop". If "INDEX_OPTIONS" is "coreProductGridWorkflow", "ReactionLayout" will look for "coreLayout" and "coreProductGridWorkflow" in "Shop". Instead, we should retrieve "coreLayout" and "coreWorkflow".

Please provide the following information to help us approve your PR. Failure to provide this
information may result in your PR being closed without comment.

- [x] Clearly explain what your PR is trying to accomplish. A link to an issue is best
Issue 1969 : https://github.com/reactioncommerce/reaction/issues/1969
With this change, the issue mentioned in Issue 1969 can be resolved.


impactmass:
Hi @gepolv,

I tested this and it still didn't solve the problem of the default template being used over the set custom one.

What is really happening here is that when there is a custom template defined (which means there are two matching Layouts - one being the default and the other the custom), the custom layout should be the one used in rendering. That selection happens [here](https://github.com/reactioncommerce/reaction/blob/master/client/modules/router/main.js#L124). So the solution should be to add logic to determine and use the custom layout.

mikemurray:
Perhaps logic, or maybe even some kinda UI to select the one you want, kinda like the template selector on the PDP page.

impactmass:
Thanks @mikemurray. I haven't considered using a UI like the PDP. But my guess it that, that would need to be considered properly, to determine if it's needed or not. 

Considering that the issue (#1969) was labeled as "regression", there was a time it was using the custom template when available (over the default).  Has there been any previous mention of having a template selector for this page?

mikemurray:
We've talked about the ability for users to select templates for anything. Consider users that use the PaaS only; they'll need a way to select different templates for pages as well.

In this case, the way the index page works is different from all others. At the very least, it needs to be in the registry so it can be overriden from the database.

gepolv:
@impactmass 
Suppose we have installed the https://github.com/reactioncommerce/reaction-example-plugin .

To better explain the problem, I have two screenshots of Collection "Shop":
reactionshops1112

The first figure shows Item 11 and Item 12. Our current behavior is selecting Item 11 because layout is set as "coreLayoutBeesknees" and workflow is set as "coreProductGridWorkflow". As you can see, the "structure" under Item 11 is not correct.

reactionshop1213

This figure shows Item 12 and Item 13. As you can see, both have the same workflow and layout. however, their "structure"s are different. "Structure" under Item 12 is from the example plugin whose template is "productsLanding" defined https://github.com/reactioncommerce/reaction-example-plugin/blob/master/register.js#27.
"Structure" under Item 13 is the default one.
So Item 12 should be the correct result and should be selected.

Conclusion: the logic here is correct because Item 12 (custom plugin) comes before Item 13 (default one). What we need to do is to change "INDEX_OPTIONS" from "coreProductGridWorkflow" to "coreWorkflow", Item 12 will be selected and the issue mentioned in https://github.com/reactioncommerce/reaction/issues/1969  will be resolved.

zenweasel:
@gepolv I agree with your analysis of the problembut I don't agree with your solution. It would only patch this one case where it's grabbing the wrong layout record and not have any affect on other custom layouts. I think the real solution is to load records in such a way that "custom" layouts always override default ones when the records are loaded in the router [here](https://github.com/reactioncommerce/reaction/blob/8dd49b43c45a6ab6f7d3712b7344bdf64df25c25/client/modules/router/main.js#L124-L124)

We also should not be relying on load order as that is not something that we can assume will always be the same.

gepolv:
@zenweasel 
Suppose we are able to overwrite the default layout+workflow, which means Item 12 will overwrite Item 13 in my Figure 2, in this case, what will be loaded by https://github.com/reactioncommerce/reaction/blob/8dd49b43c45a6ab6f7d3712b7344bdf64df25c25/client/modules/router/main.js#L124-L124 ?
It is still Item 11 in Figure 1, which is still wrong.

You basically believe the default workflow should be "coreProductGridWorkflow". What is the reason behind this? In my opinion, "coreProductGridWorkflow" cannot be the default workflow. Instead, "coreWorkflow" should be the default workflow.

Correct me if I am wrong.
Thanks.

zenweasel:
Tested this again and it still doesn't work. Closing as issue was fixed via #2023 

@gepolv I am sorry we didn't see eye-to-eye on this, but the issue is fixed now, and we welcome your input in the future.

Issue #1411, opened 2016-09-20
anonymous user cleanup job


aaronjudd:
anonymous users are not cleaned up currently. 

the cleanup job should incorporate some abandoned cart functionality and ping abilities.

zenweasel:
From a conversation/debugging session with @spencern. We can probably purge empty carts/users pretty aggressively (7 days, 1 day?) , and then longer for populated carts/account. Probably configurable in shop settings.

aaronjudd:
We could be using the cart aging.. every session will get a cart, so if you said..
cart.created > 30 days, where there are no items, delete where cart.userId from roles, users, carts, sessions if user role = anonymous.

spencern:
I think this should work. We deleted old carts + users, Accounts, Sessions and it cleaned up our db pretty well. Anon carts are pretty small though so we also purged accounts where cart.updatedAt was > 30 days even if there were items in it.

zenweasel:
Let's double-check that Jobs are being removed from the Job collection when they are completed

andela-aatanda:
I have been able to grasp a full understanding of this issue and I've implemented it locally and currently testing it to make sure it's in an okay state.

andela-aatanda:
I'll have to populate the database with dummy data to adequately tests current implementation. All seems fine at the moment

andela-aatanda:
@zenweasel jobs are currently been removed as soon as they are completed. I've been able to check that.

andela-aatanda:
This ticket has been completed with a pull request [here](https://github.com/reactioncommerce/reaction/pull/2013)

PR #2013, opened 2017-03-18
Cart Cleanup Job

andela-aatanda:
Resolves [1411](https://github.com/reactioncommerce/reaction/issues/1411)

**Test**
- Add `"cart": { "cleanupDurationDays": "older than 3 days" }` to "settings" in `reaction.json`

***Anonymous User***
- Pull this branch to your local machine
- Launch the application and put an item in the cart without login and checkout attempt.

***Registered User***
- Pull this branch to your local machine
- Launch application and Login with your credentials
- Select a product and place in your cart without checking out

**Note:**
You'll have to reset the schedule in `imports/plugins/included/jobcontrol/jobs/cart.js` as current schedule is daily


imports/plugins/core/dashboard/client/templates/shop/settings/settings.html
```diff
@@ -102,6 +102,7 @@
             {{> afQuickField name='settings.openexchangerates.refreshPeriod' placeholder="every 1 hour"}}
             {{> afQuickField name='settings.google.clientId'}}
             {{> afQuickField name='settings.google.apiKey'}}
+            {{> afFormGroup name='settings.cart.cleanup' options=options placeholder="Every 3 days"}}
```
zenweasel: This should just be a text field. Also the name should be more descriptive like `cleanupDurationDays`

imports/plugins/included/jobcontrol/server/jobs/cart.js
```diff
@@ -0,0 +1,91 @@
+import later from "later";
+import moment from "moment";
+import { Accounts, Cart, Jobs } from "/lib/collections";
+import { Hooks, Logger, Reaction } from "/server/api";
+
+const olderThan = moment().subtract(3, "days")._d;
```
zenweasel: You aren't using the settings variable you created.

imports/plugins/included/jobcontrol/server/jobs/cart.js
```diff
@@ -0,0 +1,91 @@
...
+        backoff: "exponential" // delay by twice as long for each subsequent retry
+      })
+      .repeat({
+        schedule: later.parse.text("every day")
+      })
+      .save({
+        cancelRepeats: true
+      });
+  } else {
+    Logger.info("No stale carts at the moment");
```
zenweasel: This shouldn't be an `info` level log. We don't need to log things we **don't** do.

imports/plugins/included/jobcontrol/server/jobs/cart.js
```diff
@@ -0,0 +1,91 @@
...
+            Logger.debug(success);
+            job.done(success, { repeatId: true });
+          }
+        } else {
+          const settings = Reaction.getShopSettings();
+          const schedule = Number(settings.cart.cleanup); // configurable in shop settings
+          if (!schedule) {
+            Logger.warn("Cleanup Schedule not configured");
+          } else {
+            if (cart.updatedAt <= moment().subtract(schedule, "mins")._d) {
```
zenweasel: I don't understand what you are doing here? Removing all items from the cart?

imports/plugins/included/jobcontrol/server/jobs/cart.js
```diff
@@ -0,0 +1,91 @@
...
+      updatedAt: {
+        $lte: olderThan
+      }
+    }).fetch();
+    cartDetails.map(cart => {
+      if (cart.items) {
+        const user = Accounts.findOne({ userId: cart.userId });
+        if (!user.emails.length) {
+          const result = Cart.update(
+            { userId: user._id },
```
zenweasel: By "cleaning the cart" I think we meant removing the cart record. We should also be removing stale Account and Session records as noted in the ticket.

imports/plugins/included/jobcontrol/server/jobs/cart.js
```diff
@@ -0,0 +1,91 @@
...
+          }
+        }
+      } else {
+        Logger.debug("Cart is currently empty");
+      }
+    });
+    callback();
+  });
+  Jobs.find({
+    type: "cart/removeFromCart",
```
zenweasel: What is this supposed to be doing?

imports/plugins/included/jobcontrol/server/jobs/cart.js
andela-aatanda: Yeah, I'm removing all items from the cart here. The `mins` on line 64 should have been `days`.

zenweasel: My point being, removing items from the cart is not what we want. We want to remove all unused/unneeded cart records.

imports/plugins/included/jobcontrol/server/jobs/cart.js
andela-aatanda: Noted.

imports/plugins/core/dashboard/client/templates/shop/settings/settings.html
andela-aatanda: Noted.

imports/plugins/included/jobcontrol/server/jobs/cart.js
```diff
@@ -0,0 +1,97 @@
...
+              const success = "Stale anonymous user cart and account successfully cleaned";
+              Logger.debug(success);
+              job.done(success, { repeatId: true });
+            }
+          } else {
+            const user = Accounts.findOne({ _id: cart.userId });
+            if (cart.updatedAt <= moment().subtract(Number(schedule[0]), "days")._d) {
+              const userCleanup = Cart.update(
+                { userId: user._id },
+                { $unset: { items: [] } }
```
zenweasel: I'm still unclear why we are ever removing items from the cart.

zenweasel:
When I start this after `reaction reset` I get the message ` WARN Reaction: No cart cleanup schedule` and nothing is added.

imports/plugins/core/dashboard/client/templates/shop/settings/settings.html
```diff
@@ -102,6 +102,7 @@
             {{> afQuickField name='settings.openexchangerates.refreshPeriod' placeholder="every 1 hour"}}
             {{> afQuickField name='settings.google.clientId'}}
             {{> afQuickField name='settings.google.apiKey'}}
+            {{> afQuickField name='settings.cart.cleanupDurationDays' placeholder="Every 3 days"}}
```
zenweasel: Unless I am misunderstanding this is a duration, but it's stated as a schedule i.e. "every 3 days". This job runs once a day regardless correct? So this should be "Older than X days" or something similar.

imports/plugins/included/jobcontrol/server/jobs/cart.js
```diff
@@ -0,0 +1,97 @@
...
+        if (cart.items) {
+          if (anonUser) {
+            const removeCart = Cart.remove({ userId: anonUser._id });
+            const removeAccount = Accounts.remove(
+              {
+                _id: cart.userId,
+                emails: []
+              }
+            );
+            const destroySession = ServerSessions.remove({ _id: cart.sessionId });
```
zenweasel: This logic is a little hard to follow and this is hard to test (not your fault, it's just the nature of these jobs).

Can we break this out into a function that could be called at the command line e.g. `purgeAnonymousCarts(3)` that would purge Anonymous carts older than 3 days (or whatever argument is passed there)? Even better if we could specify time unit, e.g. "1 day" or "2 hours". Then I would feel better about being able to test that it does what it says it's doing as I could just call it from the meteor shell.

imports/plugins/included/jobcontrol/server/jobs/cart.js
```diff
@@ -0,0 +1,97 @@
...
+        } else {
+          Logger.debug("No item in cart");
+        }
+      });
+    } else {
+      Logger.warn("No cart cleanup schedule");
+    }
+    callback();
+  });
+  Jobs.find({
```
zenweasel: Can you help me understand how this is supposed to work?

andela-aatanda:
That's correct @zenweasel You have to setup cleanup schedule on the dashboard shop `options` before the job can be added.

zenweasel:
@andela-aatanda Yes, I did that. It doesn't work

imports/plugins/included/jobcontrol/server/jobs/cart.js
andela-aatanda: If the entire cart of a user is removed, there's a brief disappearance of the cart icon on the landing page until there's a reload and I don't think that is a wanted behavior which was what informed my decision to remove items instead.

imports/plugins/core/dashboard/client/templates/shop/settings/settings.html
andela-aatanda: Yeah. It's misleading. It has been corrected

imports/plugins/included/jobcontrol/server/jobs/cart.js
zenweasel: But if this job is working correctly, when would you ever be eliminating the cart for a currently active user?

zenweasel: Just to be clear, the entire point of this job is that lots of cart records increases seek time. So just removing items from the cart would not going with what this code should be trying to accomplish.

andela-aatanda: The assumption here is that the user has items that are older than a certain number of days and does not check if the user is currently logged in. Hence, the user may or may not be logged in when this job runs.

zenweasel: So the use case you are talking about is of a user who goes to the site, places something in the cart, and then stares at the cart icon for three days to see it briefly disappear?

andela-aatanda: lol. That's correct

imports/plugins/included/jobcontrol/server/jobs/cart.js
andela-aatanda: This is supposed to be an observer that triggers a job after it has been completed. I see it's that way for currently existing jobs

imports/plugins/included/jobcontrol/server/jobs/cart.js
andela-aatanda: This has been implemented on line 35.

imports/plugins/included/jobcontrol/server/jobs/cart.js
```diff
@@ -0,0 +1,91 @@
...
+    Logger.warn("No cart cleanup schedule");
+  }
+});
+
+/**
+ * {Function} that fetches stale carts
+ * @param {Object} olderThan older than date
+ * @return {Object} stale carts
+ */
+const staleCarts = (olderThan) => {
```
zenweasel: this function should be called `getStaleCarts`. Function names should always contain verbs

imports/plugins/included/jobcontrol/server/jobs/cart.js
```diff
@@ -0,0 +1,91 @@
...
+    workTimeout: 180 * 1000
+  }, (job, callback) => {
+    Logger.debug("Processing cart/removeFromCart");
+    const settings = Reaction.getShopSettings();
+    if (settings.cart) {
+      const schedule = (settings.cart.cleanupDurationDays).match(/\d/);// configurable in shop settings
+      const olderThan = moment().subtract(Number(schedule[0]), "days")._d;
+      const carts = staleCarts(olderThan);
+      carts.forEach(cart => {
+        if (cart.items) {
```
zenweasel: So if I am reading this right, if the cart is empty we aren't deleting it? That definitely not what we want to do.

PR #2023, opened 2017-03-21
Set custom layout to override based on priority field

### LINKED ISSUE #1969: Can no longer override workflow/layout structures from a plugin

impactmass:
Resolves #1969 

**How to Test:**
- Create a custom plugin with a layout that you want to take precedence over another existing default layout. An easy way is to drop [reaction-example-plugin](https://github.com/reactioncommerce/reaction-example-plugin) into your project.
- Default `priority` is now 999.... so set a lower value for `priority` for your custom layout in the `register.js`
- Run `reaction reset -n && reaction`
- Load the app in the browser and confirm that your defined custom template is used over the default

**Changes introduced in this PR include (do take a look at the diff)**:
- Sorting of the layout array based on the priority field (before picking a match in selectLayout (client/modules/router/main.js) 
- Updated other registry entries using the old value of 1 to 999. This only applies to imports/plugins/core/checkout/register.js
- Added migration for templates already in db, to the new default 999 (4_update_templates_priority.js)
- Updated the `audience` field in Layout schema. It is being used as array of strings in the app... but schema was set as array of objects.


mikemurray:
I think this could use a migration to set the priority of all of the templates in our packages (core / included) to 999.

mikemurray:
Also might be useful to update the registry entries of the templates to 999

zenweasel:
Can we also make sure and update the example plugin and the tutorial with any changes here?

impactmass:
Current status on this:

* I've updated other registry entries using the old value of 1 to 999. This only applies to `imports/plugins/core/checkout/register.js`. 
* I'm currently writing a migration to update old records already in the db to this new default. 
* After that, I'll make needed updates to the example-plugin

impactmass:
@mikemurray you can take a look at this again. I've updated based on your comments.

imports/plugins/core/versions/server/migrations/4_update_templates_priority.js
```diff
@@ -0,0 +1,40 @@
+import { Migrations } from "/imports/plugins/core/versions";
+import { Packages } from "/lib/collections";
+
+const query = { layout: { $type: 3 } }; // gets docs with layout objects
```
mikemurray: It seems this query will affect all packages with layout, not just our own packages. might need to include a list of our packages like so:

```
name: { $in: ["reaction-core", "reaction-<other_name>", ...] }
```

You probably don't need all package names, only ones with actual layout.

impactmass: Ok..cool.

So... instead of making a list of packages, which might not capture all that is required based on the time the migration is run, how about this:

```const query = {
  name: { $regex: /reaction-/ }, // picks reaction pkgs
  layout: { $type: 3 } // docs with layouts set
};```

**Comment by zenweasel** [comment]
Not directly related but this makes me wonder if we shouldn't add a "vendor" field or something similar to plugins so we could differentiate between internal and external packages. Maybe a type = "core/included" too. Just so we wouldn't need to rely on a string, because I can see someone naming their plugin "reaction-myplugin". Because it's their plugin for reaction.

**Comment by impactmass** [comment]
The second part of your comment is a major concern about doing this:

```
{  name: { $regex: /reaction-/ }, // picks reaction pkgs
  layout: { $type: 3 } // docs with layouts set
}
```
It might be preferable to list our the packages (that way it's sure those that will be affected).

imports/plugins/core/versions/server/migrations/4_update_templates_priority.js
impactmass: OK.. I made a list of the affected packges: `"reaction-accounts", "reaction-checkout", "reaction-dashboard", "reaction-email", "reaction-orders", "reaction-ui", "reaction-inventory", "reaction-product-variant".`

So.. the update will be for only these ones.

PR #2030, opened 2017-03-22
[WIP] 2014 Change Delete Product Tooltip To Archive Product

### LINKED ISSUE #2014: Archive Product function still labeled "Delete Product"
When in edit mode, I tried to delete a product by clicking on the delete (red trash can) button but instead it asks if I want to archive the product and it will archive it. Even when I hover over the trash can button it has a tooltip that says: "Delete Product".

andela-akolawole:
Resolves
#2014 

**Manual Test**
- Login as admin
- Click on a product
- Hover for 3 seconds on the archive icon on the selected panel header by the right part of your screen
- The tooltip should show `Archive product.`
- Also, click on the product to view the details.
- Click the pencil icon on a variant to view the details.
- On the panel header you would see an archive icon, with a tooltip saying `Archive Variant`
- You can also click a variant option to view its details.
- Below the edit form, you would see an archive icon with a tooltip saying `Archive Option`


zenweasel:
Let's change the icon there to be `fa-archive`

private/data/i18n/en.json
```diff
@@ -255,7 +255,7 @@
         "duplicateProduct": "Duplicate Product",
         "pinProduct": "Pin Product",
         "toggleSize": "Toggle Size",
-        "deleteProduct": "Delete Product",
+        "archiveProduct": "Archive Product",
         "restoreProduct": "Restore Product from Trash",
         "deleteSelectedProducts": "Delete selected products?",
```
andela-akolawole: No, we do not.

zenweasel:
Still seeing the trashcan here and here
basic_reaction_product

basic_reaction_product

PR #2036, opened 2017-03-24
Orders Pagination

### LINKED ISSUE #1921: Pagination for orders
Goal should be that orders should perform faster, if pagination (and optional, filters) for orders implemented. Pagination and limit order for order publishing. 

The orders subscribe to accounts as well, not sure if this is done client, if so, move/re-arch to server, normalize the order in the publication.

kieha:
Fixes #1921 

#### How To Test

1. Make a couple of orders (15 or so).
2. Go to the Orders tab in the dashboard.
3. Observe that the number of orders being displayed is less than what the count at the tabs is.
4. Click `Show More` and observe the rest of the orders being loaded.
5. Use Meteor Dev Tools to check subscription of `Orders` now happens per tab, per number of orders displayed.


imports/plugins/core/orders/client/templates/orders.js
```diff
@@ -89,22 +79,35 @@ const OrderHelper =  {
...
-
-    this.state.set("orders", orders);
+    const limit = this.newLimit.get() || this.processingLimit.get() || this.completedLimit.get();
+    const subscription = this.subscribe("Orders.paginated", (limit + 1));
+
+    if (subscription.ready()) {
+      const filter = Reaction.getUserPreferences(PACKAGE_NAME, ORDER_LIST_FILTERS_PREFERENCE_NAME, DEFAULT_FILTER_NAME);
+      const query = OrderHelper.makeQuery(filter);
+
+      if (filter === "new") {
```
andela-akolawole: I think doing multiple `if else` statement is absolute due to the fact that the `switch` statement in the `OrderHelper.makeQuery` handles the same thing.

You could leverage on the switch statement to return both query and the limit.

imports/plugins/core/orders/client/templates/orders.js
```diff
@@ -89,22 +79,35 @@ const OrderHelper =  {
...
-  // fetch available orders
   this.autorun(() => {
-    this.subscribe("Orders");
-    const filter = Reaction.getUserPreferences(PACKAGE_NAME, ORDER_LIST_FILTERS_PREFERENCE_NAME, DEFAULT_FILTER_NAME);
-    const query = OrderHelper.makeQuery(filter);
-    const orders = Orders.find(query).fetch();
-
-    this.state.set("orders", orders);
+    const limit = this.newLimit.get() || this.processingLimit.get() || this.completedLimit.get();
+    const subscription = this.subscribe("Orders.paginated", (limit + 1));
```
andela-akolawole: First I think the `this` context for your subscription is referring to `Template` not `Meteor`
I think this subscription would limit the number of orders before the query is used with the `Orders.find(query)` in line `101`.

The best way to go about this would be to pass the query to the subscription.

`Meteor.subscribe("Orders.paginated",query,  (limit + 1))`
If this is used, the orders query would change to `Orders.find({})`(This would return the orders result based on the subscription).

imports/plugins/core/orders/client/containers/ordersActionContainer.js
```diff
@@ -84,7 +20,6 @@ function composer(props, onData) {
   Meteor.subscribe("Orders");
```
andela-akolawole: Since a query is not made to the orders collection, I don't think we need this.

imports/plugins/core/orders/client/templates/orders.js
```diff
@@ -89,22 +79,35 @@ const OrderHelper =  {
 
 Template.orders.onCreated(function () {
   this.state = new ReactiveDict();
+  this.newLimit = new ReactiveVar(2);
```
zenweasel: Could we combine these into one dictionary so you could use the filter as the key? e.g. orderLimit["new]]. That would help get around having to keep doing there different if/switch branches

`const orders = Orders.find(query, { limit: orderLimit[filter] }).fetch();`

imports/plugins/core/orders/client/templates/orders.js
```diff
@@ -160,6 +178,33 @@ Template.orders.helpers({
...
+
+    let limit = instance.processingLimit.get();
+    limit += 5;
+    instance.processingLimit.set(limit);
+  },
+  "click .completed": function (event, instance) {
+    event.preventDefault();
+
+    let limit = instance.completedLimit.get();
+    limit += 5;
```
zenweasel: Should/could we make this configurable somehow?

imports/plugins/core/orders/client/containers/ordersActionContainer.js
kieha: This is needed for the `Counts.publish` to work.

imports/plugins/core/orders/client/templates/orders.js
kieha: Was thinking the same. Also, the limit variables.

server/publications/collections/orders.js
```diff
@@ -14,8 +87,41 @@ Meteor.publish("Orders", function () {
...
+    shopId: shopId,
+    userId: this.userId
+  });
+});
+
+/**
+ * paginated orders
+ */
+
+Meteor.publish("Orders.paginated", function (limit) {
```
zenweasel: We don't use the dot notation in the product. Use camel-case instead.

imports/plugins/core/orders/client/templates/orders.js
kieha: @andela-akolawole you still have to pass a query to `Orders.find()` otherwise, it will just return every order in the DB, regardless of the query passed to the subscription.

joykare:
When there isn't any new/processed/completed orders it doesn't display _0_. Its empty.
screen shot 2017-03-31 at 1 13 06 pm

imports/plugins/core/orders/client/templates/orders.js
```diff
@@ -89,22 +79,29 @@ const OrderHelper =  {
 
 Template.orders.onCreated(function () {
   this.state = new ReactiveDict();
+  this.orderLimits = new ReactiveDict();
   this.state.setDefault({
     orders: []
   });
+  this.orderLimits.setDefault({
+    new: 2,
```
zenweasel: I wonder if we shouldn't increase this a little bit. 2 seems like very few before showing a "Show More".

kieha:
Fixed @joykare's comment above.

imports/plugins/core/orders/server/methods/orderCount.js
```diff
@@ -0,0 +1,27 @@
...
+import { Reaction } from "/server/api";
+import { Orders } from "/lib/collections";
+
+const workflowStatusOrderCounts = {};
+
+/**
+ * orders/count
+ *
+ * @summary Count number of orders per workflow status
+ * @param {String} filterName - string to filter by
```
zenweasel: It doesn't seem like it takes this parameter?

zenweasel:
It does not seem like the order counts are reactive any longer?

imports/plugins/core/orders/client/templates/orders.js
kieha: Yeah, I'd set it to 2 for my own testing purposes, but in production people would have a lot more orders. What would you suggest the number be?

imports/plugins/core/orders/client/templates/orders.js
```diff
@@ -160,6 +162,19 @@ Template.orders.helpers({
...
 
+/**
+ * orders events
+ */
+Template.orders.events({
+  "click .show-more-orders": function (event, instance) {
+    event.preventDefault();
+    const filter = Reaction.getUserPreferences(PACKAGE_NAME, ORDER_LIST_FILTERS_PREFERENCE_NAME, DEFAULT_FILTER_NAME);
+    let limit = instance.orderLimits.get(filter);
+    limit += 5;
```
kieha: @zenweasel This is a small number to increment by as well. Any suggestions?

zenweasel: I think 10 and 10 would be hard to argue against?

joykare:
Show more button shows up even when there aren't any more orders to show.

kieha:
@zenweasel yeah, the order counts aren't reactive any more. Mongo aggregations are not reactive. I could make them reactive by piping the result of the aggregation to a new `OrderCount` collection and publishing/subscribing to that.

imports/plugins/core/orders/client/templates/orders.js
```diff
@@ -141,6 +138,11 @@ Template.orders.helpers({
     return Template.instance().state.get("orders") || false;
   },
 
+  hasMoreOrders() {
+    const filter = Reaction.getUserPreferences(PACKAGE_NAME, ORDER_LIST_FILTERS_PREFERENCE_NAME, DEFAULT_FILTER_NAME);
+    return Template.instance().state.get("orders").length >= Template.instance().orderLimits.get(filter);
```
joykare: The length of the orders in state is always 2 or less coz you set limit to 2 orders.  So it returns true everytime you have two or more orders. So if I actually have just 2orders in total it'll still show the "Show more" button. Which isn't ideal.

zenweasel:
You should be able to create a publication for the aggregate using a technique like this: http://meteorcapture.com/publishing-anything/

imports/plugins/core/orders/client/templates/orders.js
kieha: Fixed

aaronjudd:
Curious why/if the [publish-counts](https://github.com/percolatestudio/publish-counts) approach didn't work for the order counts? There are some legit reasons to not use those, but wasn't sure of the logic here?

kieha:
@aaronjudd I've revisited implementing the counts using `publish-counts`, and what I initially thought wouldn't work, actually works.

The other issue with it is performance.

`This package is designed primarily for correctness, not performance. That's why it's aimed at counting smaller datasets and keeping the count instantly up to date.`

The above is from the package README. So I thought that if a person had a very large dataset of orders, the orders would still be slow.

zenweasel:
Wait, so now aren't using aggregates for doing the counts? Are we still not oversubscribed to `Orders`?

kieha:
@zenweasel no we aren't. Since we have a new `PagintedOrders` publication, we're subscribed to that instead.

kieha:
Update:
Subscribing to the `PaginatedOrders` publication for `publish-counts` does not work as expected. 
The initial idea was subscribing to as many documents as are visible per tab, i.e. 10 initially, then 20 when `Show More` is clicked, e.t.c. 
What is happening now is that, since the `PaginatedOrders` subscription in `imports/plugins/core/orders/client/containers/ordersActionContainer.js` does not receive a limit, it's subscribing to every document per tab, i.e. if there are 13 new orders, it subscribes to all 13 of them, not the 10 that are shown initially.

zenweasel:
So if this isn't working, should we put this back in WIP?

kieha:
Update:
Passing the limit as props from the template to the React component that renders the tabs fixes the issue of subscribing to as many documents as are visible per tab.

zenweasel:
When I visit a tab that has fewer than 10 orders I get a spinner that never disappears (and should never appear because we aren't waiting for anything).

reaction

### Steps to reproduce

1. Add 12 orders
1. Click on 2 of them so they are in "processing"
1. Observe spinner in Processing tab

zenweasel:
Retested this and seems to be working correctly.

zenweasel:
Just waiting for approval of changes from @aaronjudd

kieha:
@aaronjudd this is ready to go, just waiting for your approval of changes.

aaronjudd:
Yes, I am actually testing it right now.

Issue #1445, opened 2016-09-27
Add to cart alert behaviour


aaronjudd:
### Expected behavior

When I rapidly add multiple items to cart, I expect there to be some recurring updates in the "Add to Cart" slideout. 
### Actual Behavior

Formerly this was implemented using jQuery easing, which provided some queueing, and would display additional updates when additional items were added to the cart.  Now using just CSS, we aren't getting additional updates
### Steps to Reproduce the Behavior

Add item to cart, see slider.  Add to cart again (and again, and again) and slide out will not appear

andela-aatanda:
I'll reproduce this behavior and work on solving it over the next couple of days.

andela-aatanda:
@aaronjudd reducing the slide out delay `time` from `4s` to `0.4s` somehow fixes this. Is there any reason why it was kept at `4s`?

zenweasel:
The slideout delay is 4s so that the toast remains on the screen long enough for you to read it. You could probably reduce it some, but making it .4s is way too fast.

aaronjudd:
@andela-aatanda agree with Brent, the goal here is a smooth visual effect that is noticeable and gives time for a user to see the action, this issue is not about the behaviour of adding one item, but the effect that occurs while you add more than one item and it's already slide out once.

andela-aatanda:
Gotcha.

andela-aatanda:
@zenweasel I was thinking if the `Add To Cart` button is deactivated pending the time the `cart-alert` goes back in this behavior will be prevented. Will deactivating the `Add To Cart` button be a behavior we'll want to have?

andela-aatanda:
PR has been raised fixing this ticket and is currently been reviewed.

rymorgan:
@zenweasel and I were talking this through and here's what we came up with. 
When you click quickly to add to cart multiple times, it should update the original alert and restart the animation.  It should count up and if it was starting to slide out it would start the time over again. So it’s always 4 seconds from the last time you clicked the add to cart button. 

When it updates the alert, display a spinner. (you can use the same spinner we use on buttons like the complete payment button)

See rough mockup. 

![cart - toast](https://cloud.githubusercontent.com/assets/20409254/24482795/cc27ee9a-14a8-11e7-8db6-b07304923895.png)

andela-aatanda:
@rymorgan This put things in better perspective. Thanks

PR #2039, opened 2017-03-27
[WIP] 1445 Add to Cart Alert Behavior

andela-aatanda:
Resolves [1445](https://github.com/reactioncommerce/reaction/issues/1445)

***Test***
- Clone this branch
- Start the application
- Select product and click add to cart rapidly and notice the button becomes deactivated until cart alert is hidden


imports/plugins/included/product-detail-simple/client/containers/productDetailContainer.js
```diff
@@ -112,14 +115,15 @@ class ProductDetailContainer extends Component {
...
             [direction]: 0
           }, 600)
-          .delay(4000)
+          .delay(2000)
           .animate({
             [oppositeDirection]: "auto",
             [direction]: -alertWidth
           }, {
             duration: 600,
             complete() {
```
zenweasel: Is there a way to re-enable add to cart at the start rather than the end of this animation?

zenweasel: Also, could we provide some sort of visual indicator that the "Add to cart" button is disabled? Otherwise I think people might be confused as to why it's not responding

andela-aatanda: Enabling the button at the start of the animation will defeat this solution as it will re-create the problem since users will be able to click `Add To Cart` rapidly.

andela-aatanda: We can have that. I'm on it

andela-aatanda: I have fixed this on `line 76`

zenweasel:
I'm still not seeing the add to cart button showing disabled while I am waiting for the toast to close.

andela-aatanda:
I think I understand what you meant by a visual notification that the button is deactivated. I thought the change in `cursor` would pass that message. I'll fix that soon

imports/plugins/included/product-detail-simple/client/containers/productDetailContainer.js
```diff
@@ -72,8 +72,15 @@ class ProductDetailContainer extends Component {
         });
       } else {
         productId = currentProduct._id;
-        this.setState({ disableAdding: true }); // disable add-to-cart button
         this.setState({ disabledStyle: { cursor: "not-allowed" } }); // sets cursor property style on button
+        this.setState({ disableAdding: true }); // disable add-to-cart button
+
+        // message to user when add-to-cart button gets deactivated
+        Alerts.inline("Add to cart button is currently inactive", "info", {
```
zenweasel: No, this is not what I mean. You need to show the button disabled, e.g. show it greyed-out.

andela-aatanda: Gotcha!

zenweasel:
So while this technically solves the issue I don't think this is going to work for us because it makes the system seem slow and unresponsive. I spoke with @rymorgan and I think what we want to do is leave the add to cart button alone and stack the messages along the side similar to how we do with error message toasts. I am closing this PR as this will probably involve going back to the drawing board a bit. Will make additional notes on the original ticket.

PR #2060, opened 2017-03-29
Make sure createDefaultAdminUser works when at least 1 user is present in the system

callmephilip:
### Current behavior

Admin user check in the **createDefaultAdminUser** returns true even in the case where there's no admin user in the system but 1 or more "regular" users. As per **Roles.getUsersInRole** [documentation](https://alanning.github.io/meteor-roles/classes/Roles.html#method_getUsersInRole), if an array is passed as the first arg, `users returned will have at least one of the roles specified but need not have all roles`.

### Expected behavior

An admin must be created even if there are already regular users in the system but no actual admin


callmephilip:
The problem i am now seeing is when there's another non admin user in the system,  and createDefaultUserAdmin is run, the following exception is thrown:

```
Error: Meteor.userId can only be invoked in method calls. Use this.userId in publish functions.
    at AccountsServer.userId (packages/accounts-base/accounts_server.js:82:13)
    at AccountsServer.user (packages/accounts-base/accounts_common.js:58:23)
    at Object.Meteor.user (packages/accounts-base/accounts_common.js:261:19)
    at AccountsServer._onCreateUserHook (server/startup/accounts.js:98:36)
    at AccountsServer.Ap.insertUserDoc (packages/accounts-base/accounts_server.js:1308:21)
    at createUser (packages/accounts-password/password_server.js:1009:25)
    at AccountsServer.Accounts.createUser (packages/accounts-password/password_server.js:1078:10)
    at Object.module.export.exports.default.createDefaultAdminUser (server/api/core/core.js:297:28)
    at Object.module.export.exports.default.init (server/api/core/core.js:34:10)
    at module.export.exports.default (server/startup/init.js:8:12)
    at module.export.exports.default (server/startup/index.js:19:3)
    at server/main.js:6:3
    at Function.time (/Users/philip/projects/it-furniture/.meteor/local/build/programs/server/profile.js:309:28)
    at /Users/philip/projects/it-furniture/.meteor/local/build/programs/server/boot.js:312:13
    at /Users/philip/projects/it-furniture/.meteor/local/build/programs/server/boot.js:353:5
    at Function.run (/Users/philip/projects/it-furniture/.meteor/local/build/programs/server/profile.js:510:12)
```

aaronjudd:
Ok, so yes, this doesn't break, and you are correct that the array returns any match, but....  two things..  `createDefaultAdminUser()` is really creating the "owner", not the admin, so the single role to check would be `owner` instead of `admin`. The intended use of of `createDefaultAdminUser` is really just a one time use, (which is why this hasn't been an issue/caught before), it wasn't intended to be re-used to add new store owners, as this method also adds a global role, so.. if you added additional users with this method, they'd have access to everything / all shops, not just be an admin of a single shop.

callmephilip:
The situation i found myself in is as follows:
- i had no way to log in as the admin (did not have admin credentials set using env vars and failed to grab auto-generated admin info)
- removed auto-generated admin user model from the DB, set env vars accordingly for the new admin user
- i was expecting RC to reinitialize the admin but stumbled upon it saying the admin already exists which is not strictly true but is due to how the check works (see my original comment)

Now, i understand your arguments about how it's supposed to be used once and only once but what would be a reasonable way to recover from the situation described above? Also, I still maintain that the admin check should be modified since it's misleading right now. I would also be curious to know where the exception comes from since it is very confusing

zenweasel:
I think we could just make this change look for `owner` (there should be only one owner, and it should always exist) and I think we would be okay with it. I think @aaronjudd was just explaining why it worked the way it did.

zenweasel:
@callmephilip Did you want to make that change to the PR or should I just close it?

Issue #2026, opened 2017-03-22
Price shows as $0.00 in product grid when adding a product with just one option


zenweasel:
### Expected behavior
If you add one option with a price of $12, the price in the product grid should be $12

### Actual Behavior
Price shows as $0.00

### Steps to Reproduce the Behavior
1. Change branch to `master`
1. `reaction reset`
1. Login as admin
1. Create a new product, fill in description etc
1. Add one option
1. Publish
1. Observe price displays as $0.00 in the product grid

![dumb_product](https://cloud.githubusercontent.com/assets/72819/24189356/9636a904-0f1f-11e7-94f8-c1f9c623d349.png)

![reaction](https://cloud.githubusercontent.com/assets/72819/24189367/a16b0d38-0f1f-11e7-8489-8002365bd2f0.png)

hrath2015:
This is happening because  Product Revision is not available on server.  This API is supposed to work on client where expected Revision is embeded in Published data. 
https://github.com/reactioncommerce/reaction/blob/development/server/methods/catalog.js#L185
https://github.com/reactioncommerce/reaction/blob/development/server/methods/catalog.js#L187
https://github.com/reactioncommerce/reaction/blob/development/server/methods/catalog.js#L207

What we are doing in here:
https://github.com/reactioncommerce/reaction/blob/development/imports/plugins/core/revisions/server/hooks.js#L21
 needs to be done in `denormalize()` method.

If nobody is working on it please let me know, I have already made some changes to fix it. Will send it for review.

Issue #2059, opened 2017-03-29
When a product is made visible its variants remains invisible in database


hrath2015:
### Expected behavior
when a product is made visible all the variants should also be made visible
### Actual Behavior
Only top level product is updated in DB.
### Steps to Reproduce the Behavior
Create a new product
Add variants.
Click the eye icon in toolbar to make is visible.
Check in DB(revisions) only Top Level product is set to `isVisible=true`

michael-alade:
A PR has already been opened for this issue #2222

zenweasel:
Closed via #2222

PR #2064, opened 2017-03-30
Fix issue #2026 changed price not displayed for prod on grid page

hrath2015:
- [ ] Clearly explain what your PR is trying to accomplish. A link to an issue is best
To fix issue #2026 
This is happening because Product Revision is not available on server. This API is supposed to work on client where expected Revision is embeded in Published data.
https://github.com/reactioncommerce/reaction/blob/development/server/methods/catalog.js#L185
https://github.com/reactioncommerce/reaction/blob/development/server/methods/catalog.js#L187
https://github.com/reactioncommerce/reaction/blob/development/server/methods/catalog.js#L207

What we are doing in here:
https://github.com/reactioncommerce/reaction/blob/development/imports/plugins/core/revisions/server/hooks.js#L21
needs to be done in denormalize() method.
   1.  Can  change the price of  any variant loaded by default. Go to index page, it should show the updated one.
   2.  If wanna try new product then make sure to create, make it visible, publish and then update  `isVisible=true` for variant in db. (because of issue #2059)

### DIFF
--- a/imports/plugins/core/revisions/server/hooks.js
+++ b/imports/plugins/core/revisions/server/hooks.js
@@ -154,6 +154,14 @@ export const ProductRevision = {
       }
     });
     return variants;
+  },
+  getVariantQuantity(variant) {
+    const options = this.getVariants(variant._id);
+    if (options && options.length) {
+      return options.reduce((sum, option) =>
+      sum + option.inventoryQuantity || 0, 0);
+    }
+    return variant.inventoryQuantity || 0;
   }
 };
 
--- a/imports/plugins/core/revisions/server/hooks.js
+++ b/imports/plugins/core/revisions/server/hooks.js
@@ -18,7 +18,7 @@ function convertMetadata(modifierObject) {
   return metadata;
 }
 
-const ProductRevision = {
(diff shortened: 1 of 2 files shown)


zenweasel:
While this works and fixes the problem structurally it feels sort of weird to me, importing from `hooks`. If we need a server-side version of this library (which lib should work shouldn't it?) maybe we should be refactoring and calling from hooks.

hrath2015:
Though it is in lib, the result will not be as expected in this case.  If you run query `Products.find({_id:"someid"}) ` for a product  with revision in browser console and meteor shell on server you will get different output.  I think that was the reason the Revision API was implemented in Hooks thinking it will not be needed anywhere else.  But not working in this case.  I tested with `denormalize()`  function commented and  found  it working. But I am not sure if this will be good idea.   So to keep regression at minimum I imported API which is already there.   And you are right,  It should be refactored to have server API which can be used in existing Hooks and Catalog methods and may be in future usecases.  
 Till then if ok can merge and have a  issue opened to refactor.

hrath2015:
👍

Issue #2031, opened 2017-03-22
Move/add variant icon buttons.


rymorgan:
1. Move Public/Private icon button from body of the page to sidebar. 
2. When the variant is set to private, fade it out
3. Change placement of icon buttons for variant options in sidebar
4. Add ability to clone a variant option

![pdp_-_action_view-2](https://user-images.githubusercontent.com/20409254/30404551-fefeaee0-989b-11e7-94fb-63a52704d442.png)
![pdp_-_action_view-1](https://user-images.githubusercontent.com/20409254/30404550-fefe8014-989b-11e7-930e-9f4d0cf67bd4.png)

rymorgan:
Related to https://github.com/reactioncommerce/reaction/issues/1647

aaronjudd:
Demo'ed this morning... so assigned to @kieckhafer

kieckhafer:
^^ Yes this is done in PR #2893

kieckhafer:
Closed by #2893

PR #2065, opened 2017-03-30
Update read only Quantity in top variant #1919

### LINKED ISSUE #1919: Update read-only Quantity in top-level variant when children are present
When a Variant has options / second-level variants, inventory is moved away from the top-level, and onto the child.

However, in the Edit Variant screen, we still show the most recent quantity of top-level variant products, never updating it to use the combination of second-level variants, which is the true inventory count.

We should update the top level variant field to show the correct inventory numbers.

inventory

Related to #1867

joykare:
Resolves #1919 

**Description**
When a variant has second level variants the edit variant screen still shows the most recent quantity of top-level variant products, not updating it to use the combination of second-level variants, which is the true inventory count.

**Test**
1.  Create product.
2.  Create child variants and add quantities.
3. View edit variant section. On the top variant quantity expect to see total quantity of the child variants.


hrath2015:
I would say we  need to re consider this.  When Midvariant has Options (Child) It should be treated similar to Top Level Product.  Quantity  not applicable(u can check in inventory) and Price is the Range based on what children's prices are. 

What  we should do:
 1.  Remove Qty field from UI if Options are added.
 2.  Price Should be derived UI field only like we have on Left Side of PDP.    No need to change anything in DB schema. 
I would say it is working fine except some UI changes in line with Product structure what RC has. 
We  should take it up as UI refactor/revisit as part of issue #2031.

joykare:
@hrath2015 seeing as the UI change is `in progress` should we just close this issue and create a new issue with the new specifications you mentioned about the changes in the UI. Coz doing this with the current UI would be futile.

joykare:
@zenweasel Wanna check this out?

mikemurray:
@hrath2015 I think this is just a matter of display. We just want to show the total quantity, not actually change it on the higher level variant.

Also, I would advise against removing UI elements because it adds mystery and inconsistency to the interface, and when things disappear / reappear, users may not understand why.

Disabled with maybe a tooltip explaining why would be a good option.

zenweasel:
When I test this I am seeing this error and I can't create a second option.

```
Error invoking Method 'products/updateProductField': [Quantity must be a number]
```

hrath2015:
@mikemurray   Disabled field with tooltip is also  good suggestion to keep UI consistent. We can avoid DB changes/updates as quantity has no meaning at parent level(it should always be at option level).  I think we should do this as part of issue #2031.

zenweasel:
@joykare What's the status of this? Should this be changed back to `WIP` or do you need some help?

joykare:
No @zenweasel I think you should take another look at it.

I couldn't get it to change the field immediately but it is disabled for the time till it updates. Also it'll be a lot different once @kieha is done with the PDP change to react.

kieha:
@joykare @zenweasel this won't be any different once the conversion to React is done because the `Quantity` field will still be there, it'll just be moved into a different card from where it was before.

zenweasel:
@joykare What's the status of this? Njeri is saying that the PDP React changes will not affect the field. Can we please finish this up and get it closed? This PR is almost two weeks old.

kieha:
The most recent commit sort of fixes this issue. However, the behavior is a bit weird. Reason being, the `inventoryQuantity` field is updated together with every other field from auto form (the updates are sent as an object to `products/updateVariant`), and we are trying to manipulate it externally with another meteor method (updating the one field with `products/updateProductField`).

This might be a bit better executed once I'm done with the PDP change to React because once there, every field is going to be updated separately.

zenweasel:
This seems to work as expected now

imports/plugins/included/product-variant/client/templates/products/productDetail/variants/variantForm/variantForm.html
```diff
@@ -88,7 +88,7 @@
                 {{/if}}
               </div>
 
-              <div class="col-xs-3 col-sm-3 form-group {{#unless hasChildVariants}}{{#if afFieldIsInvalid name='inventoryQuantity'}} has-error{{/if}}{{/unless}}">
+              <div class="col-xs-3 col-sm-3 form-group {{#unless updateQuantityIfChildVariants}}{{#if afFieldIsInvalid name='inventoryQuantity'}} has-error{{/if}}{{/unless}}">
```
zenweasel: This seems a little bit tricky but I'll allow it.

PR #2073, opened 2017-04-03
Don't subscribe to all media

### LINKED ISSUE #1815: Add media handling and storage options
- Add settings to collectionFS to stream/store to S3. (and cdn)
- Update Media publication to return URL's instead of blobs, so the Media (all) subscription is more performant.
- Review default aspect ratios, and allow these to be changed in admin UI.
- perhaps review lambda service for generating responsive images

mikemurray:
Fixes #1815 

- Publish brand assets globally in media publication
- Publish product media in the `Products` and `Product` publications
- Limit media to only published products

After much back and forth with another branch, this is the solution I came up with. Simple yea?

Let’s call this a first step towards media performance. Everything else like CDN / S3 will be next.

### DIFF
--- a/server/publications/collections/product.js
+++ b/server/publications/collections/product.js
@@ -1,4 +1,4 @@
-import { Products, Revisions } from "/lib/collections";
+import { Media, Products, Revisions } from "/lib/collections";
 import { Logger, Reaction } from "/server/api";
 import { RevisionApi } from "/imports/plugins/core/revisions/lib/api/revisions";
 
@@ -156,7 +156,17 @@ Meteor.publish("Product", function (productId) {
         handle2.stop();
       });
 
-      return this.ready();
+      const mediaCursor = Media.find({
+        "metadata.productId": _id
+      }, {
+        sort: {
+          "metadata.priority": 1
+        }
+      });
+
+      return [
+        mediaCursor
+      ];
     }
 
     // Revision control is disabled
@@ -164,5 +174,18 @@ Meteor.publish("Product", function (productId) {
   }
 
   // Everyone else gets the standard, visibile products and variants
-  return Products.find(selector);
+  const productCursor = Products.find(selector);
+
+  const mediaCursor = Media.find({
+    "metadata.productId": _id
+  }, {
+    sort: {
+      "metadata.priority": 1
+    }
+  });
+
+  return [
+    productCursor,
+    mediaCursor
+  ];
 });
--- a/server/publications/collections/product.js
+++ b/server/publications/collections/product.js
@@ -22,16 +22,14 @@ export function findProductMedia(publicationInstance, productIds) {
     selector["metadata.shopId"] = shopId;
   }
 
+  // No one needs to see archived images on products
+  selector["metadata.workflow"] = {
+    $nin: ["archived"]
+  };
+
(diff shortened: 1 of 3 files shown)


zenweasel:
When I go to add a product (after adding images to the default product, maybe not related) I get this error:

```
Exception in queued task: Error: Could not find element with id NCHtKw9CaT5WssF4s to change
    at [object Object]._.extend.changed (packages/ddp-server/livedata_server.js:188:13)
    at [object Object]._.extend.changed (packages/ddp-server/livedata_server.js:403:10)
    at [object Object]._.extend.changed (packages/ddp-server/livedata_server.js:1262:19)
    at Object.Revisions.find.observe.changed (server/publications/collections/products.js:346:20)
    at [object Object].observeChangesCallbacks.changed (packages/minimongo/observe.js:162:28)
    at self.applyChange.changed (packages/minimongo/observe.js:66:44)
    at packages/mongo/observe_multiplex.js:182:30
    at Array.forEach (native)
    at Function._.each._.forEach (packages/underscore.js:139:11)
    at Object.task (packages/mongo/observe_multiplex.js:176:9)
    at [object Object]._.extend._run (packages/meteor.js:807:18)
    at packages/meteor.js:785:14
```

zenweasel:
Other than getting that error though, it seems to work well.

server/publications/collections/product.js
```diff
@@ -1,7 +1,47 @@
...
+  if (shopId) {
+    selector["metadata.shopId"] = shopId;
+  }
+
+  // Product editors can see both published and unpublished images
+  if (!Reaction.hasPermission(["createProduct"], publicationInstance.userId)) {
+    selector["metadata.workflow"] = {
+      $in: [null, "published"]
+    };
+  } else {
```
zenweasel: Shouldn't this be outside the `if/else` construct if it applies to everyone?

mikemurray: I dare not question the zenweseal's masterpiece https://github.com/reactioncommerce/reaction/blame/1815-media-take2/server/publications/collections/media.js#L30

But on a serious note, I suppose it makes sense, but not sure if you built it that way for a reason.

mikemurray:
`Exception in queued task: Error: Could not find element with id NCHtKw9CaT5WssF4s to change`

@zenweasel I can't seem to reproduce that error with adding / removing media. This error is probably related to the ongoing issues with revision control and observables. 

Made a ticket for it https://github.com/reactioncommerce/reaction/issues/2079 as I believe the issue still existed in rare cases before these changes to the Media collection.

My steps

1. fully kill meteor
1. full reset, reset, and reset!
1. start, mindful that no tabs have active meteor sessions.
1. Had a logged in session, and incognito session
1. add images, remove images, add products / variants with images, remove, the works.

Both seems to work and the error doesn't show. I think using collection observables to piece together two collections may be too fragile; however, changing that may require a significant rework as per https://github.com/reactioncommerce/reaction/issues/2079.

server/publications/collections/product.js
zenweasel: Oh, so now we're using "but it's your code I copy/pasted" as an excuse?

jk my code is probably wrong.

zenweasel:
Starting fresh I can't replicate this error either.

PR #2086, opened 2017-04-05
PDP Cleanup - Basic Details Card, Tax Card & Inventory Card

### LINKED ISSUE #2032: PDP Cleanup- Basic Details Card, Tax Card & Inventory Card Updates
See annotated .png for details. 

inventory
inventory-annotated

kieha:
Fixes #2032 

### How to test
- Clone this branch
- Open PDP variant sidebar on an existing/new product
- Open browser console to catch console errors (if any)
- Navigate through the fields/buttons/switches to ensure everything works as it did before with the Blaze template


mikemurray:
Here's some pre-release docs I owe you for some of the components you might need.

Setting Card
http://localhost:3040/reaction-docs/component-docs/settings-card

Switch
http://localhost:3040/reaction-docs/component-docs/switch

zenweasel:
I'm not able to edit the tax code

zenweasel:
If I edit a product and then create a new product, I can't get the side panel to show the new variant when I click on the edit icon unless I click on a top-level field and click back.

dumb_product

zenweasel:
If I set a "Warn At" I get the limited supply badge even though I'm not anywhere near the "Warn At" quantity.

zenweasel:
When I create a product with variants, the price range is shown correctly as an admin but not for a standard user.

another_dumb_product_and_another_dumb_product_and_slack_-_reaction

kieha:
@zenweasel the price range shows correctly for a normal user. When you click on a variant option (child variant) is when it displays the price of that particular variant option (child variant). When you click on the variant itself (parent variant) is when you see the full price range.

When I click on the parent variant:

screen shot 2017-04-27 at 1 49 48 am

When I click on one of the child variants:

screen shot 2017-04-27 at 1 49 58 am

zenweasel:
@kieha Did you test this using a new product, not the default product as I mentioned in my comment? I am aware of how it works and specifically tested to make sure it wasn't about which variant was selected.

kieha:
@zenweasel I've tested it just now with a new product. Works as it should.

screen shot 2017-04-27 at 2 11 41 am

screen shot 2017-04-27 at 2 11 50 am

zenweasel:
Okay, let's change the default for Taxable and Inventory Management back to `true` and then we should be good to go.

private/data/i18n/en.json
```diff
@@ -273,20 +273,20 @@
...
         "weight": "Weight",
         "length": "Length",
         "width": "Width",
         "height": "Height",
         "taxable": "Taxable",
-        "taxCode": "Tax Codes",
+        "taxCode": "Tax Code",
         "taxDescription": "Tax Description",
-        "inventoryManagement": "Inventory tracking",
+        "inventoryManagement": "Inventory Tracking",
```
aaronjudd: @rymorgan Case use in strings is a design choice that I think we should be enforcing some rules around.  I consider `Inventory tracking` to be a correct form versus `Inventory Tracking`, just as I prefer to not see a `:` used on labels.  I'm not asking for a change here, but it's something I think could be covered ASAP in a style guide. (which ever way the case may go ;-) )

aaronjudd:
I tried to change the option title, which didn't save and got the server console error 
```Exception in queued task: Error: Could not find element with id u3hQWdjKwnTpywxBk to change...
    at Object.Revisions.find.observe.changed (server/publications/collections/product.js:181```

aaronjudd:
Not sure if this is a new issue, but we need to handle long strings better here in the UI.

basic_reaction_product

aaronjudd:
Interesting.. I'm able to create a product with a negative qty, and a price of 0.00 and add it to the cart... well at least the "add to cart slider" comes out..  but nothing actually gets added.  
basic_reaction_product

kieha:
@aaronjudd I tried replicating `Creating a product with a negative qty, and a price of 0.00 and add it to the cart`. The first time I did, I got the same result as yours, adding to card slider popped out, with this error in browser console:

screen shot 2017-04-27 at 8 45 50 pm

This seems to only happen when you have a product with the top-level variant and child variants because the second time I tried it, with a new product, same conditions, but only with the top-level variant, I got this:

screen shot 2017-04-27 at 9 21 02 pm

zenweasel:
@kieha The issue where the side-panel is not opening when I click on the variant has not been fixed.

zenweasel:
Whenever I set inventory quantity or "warn at" it toggles the backorder switch

backorder_quantity

zenweasel:
The issue of the "Backorder" or "Limited Quantity" badge showing when the "Warn At" quantity is set is still occurring. I have verified that this doesn't happen on the `development` branch.

kieha:
@zenweasel I've tried reproducing the `Warn At` but I can't seem to. I'm only getting the badges under the correct conditions, e.g. when I set the `Warn At` to 9 and the `Quantity` to 7, I get the `Limited Supply` badge, which is correct.

Do you mind giving me steps to reproduce it?

zenweasel:
@kieha Just set `warn at` to any quantity (e.g. quantity to 1000, warn at to 500) and you get the "limited supply" badge, which should not show until quantity < 500.

kieha:
@zenweasel setting `Quantity` to 1000, `Warn At` to 500, I don't get the badge:

screen shot 2017-05-02 at 10 21 34 pm


Setting `Quantity` to 100, `Warn At` to 500, I get the badge:

screen shot 2017-05-02 at 10 19 32 pm

zenweasel:
@kieha Okay, I think I see the issue. I had quantity on an option set to 100 after I had set the quantity at 1000 at the top-level. So disregard that.

PR #2092, opened 2017-04-07
Add `requirements` subsection and `my-reaction-app` param for quick test

Haeresis:
## What is the problem this PR should resolved

**Trying to quickly test the solution to see if it fit it my needs I follows this steps**

**Step**

```bash
git clone https://github.com/reactioncommerce/reaction.git
cd reaction
npm install -g reaction-cli
reaction init
```

produce

```bash
Directory 'reaction' already exists.
Use 'reaction init somename' to install in a different directory.
```

**Screen**

Example

**Config**

Windows : 10
Node.js : 3.10.9
Meteor : 1.4.3.2
NPM : 3.10.9

**And I think that**

There is no way for Windows OS to knows the difference between a `reaction` directory or a `reaction` file. So use the command `reaction init` will always result on error saying `reaction directory already exist`. So if you're smarter you will find the CLI documentation and add a param after `init` command BUT in my case, Meteor was not mentioned as a dependency and my message was `you have not Meteor`. After installing Meteor, when I have use another time `reaction init` command I get `reaction directory already exist` and I have thinking « Ok, first time all was installed, and it's just because Meteor was not installed ». After that I have been a lot of dependencies problems, etc. I have past 2 hours just for `test` application and I use Node.js ecosystem really often so I imagine a beginner want just try « out-of-the-box ».

## Why this modifications

So just run your project on Windows system without understand `git clone` is not useful, and you will always have `reaction directory already exist` message when you will use `reaction init`.

In order to avoid this problem for futur beginners I propose:

- first: propose to install node.js and meteor before starting.
- second: place installation command in dedicated subsection to understand this command not require any `git clone` usage and is self-suffisant.
- three: encourage usage of this command with a param. It's why in the README.md now I say `reaction init my-reaction-app`

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ npm install -g reaction-cli
 
 For more information on installation check out the [installation](https://docs.reactioncommerce.com/reaction-docs/development/installation) doc.
 
-### reaction-cli usage
+### reaction-cli init usage
 
 Into your workspace:
 
@@ -47,6 +47,8 @@ reaction init my-reaction-app
 cd my-reaction-app
 reaction
 ```
+
+Go on the url provided and log you with your login/password information also provided.
 
 For more information on configuration check out the [configuration](https://docs.reactioncommerce.com/reaction-docs/development/configuration) doc.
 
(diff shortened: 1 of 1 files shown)


aaronjudd:
Thanks! This (doc) was already updated in #2085, and also the CLI checking was updated in https://github.com/reactioncommerce/reaction-cli/commit/1920f0de2887cb0b37d4d4b40c4cf3999b196a18

PR #2097, opened 2017-04-10
Remove legacy PDP

### LINKED ISSUE #2080: Remove legacy PDP
This is the version that used to be displayed when revision control is disabled. Related to #2079

impactmass:
Resolves #2080.

How to Test:
* Run the app and visit the PDP.
* Check that the PDP is working as it should (using React components)
* Check with an old revision (since this is removing the isRevisionControlEnabled check).


impactmass:
Tested with v18... Created a new product below:

screenshot 2017-04-11 13 17 16


Switched to current development, then loaded up PDP below:

screenshot 2017-04-11 13 10 45

cc @zenweasel

PR #2106, opened 2017-04-11
trying again with renamed repo for submodule eePLugins

katiebvogel:


katiebvogel:
Mistake while experimenting with submodules.

PR #2107, opened 2017-04-11
Fix "Set Custom Value" on Avalara Usage Type

### LINKED ISSUE #2101: [Avalara] "Set Custom Value" does not work for "Customer Usage Type"
### Expected behavior
In the Account screen you should be able to select "Set Custom Type" and type a custom type rather than selecting one from the dropdown.

### Actual Behavior
Selecting "Set Custom Type" doesn't do anything

1. Login as an admin
1. Select "Manage" for a user
1. Choose "SET CUSTOM VALUE"
1. Observe nothing happens

impactmass:
Resolves #2101

How to test:
- Login as an admin
- With Avalara enabled, Select "Manage" for a user
- Choose "SET CUSTOM VALUE" under Tax Settings
- A "Custom Value" input should appear. 
- Type in a code and save.
- Ensure saving works as it should


impactmass:
Would test this again. And open up for review shortly.

impactmass:
customtype

zenweasel:
When I test this I see that you can set it and it's coming across but the item is still showing as Taxable. I have an email into Avalara to see what might be happening there.

avatax_development_admin_console

imports/plugins/included/taxes-avalara/client/accounts/exemption.js
```diff
@@ -6,11 +6,16 @@ import { Packages, Accounts } from "/lib/collections";
 import { Accounts as AccountsSchema } from "/lib/collections/schemas/accounts";
 import { TaxEntityCodes } from "/client/collections";
 
+let entityCodeList = [];
+let currentAccount;
+
 Template.taxSettingsPanel.helpers({
   account() {
     const sub = Meteor.subscribe("UserAccount", this.member.userId);
```
zenweasel: Just noticing this now, but should thus subscription be in the helper? Seems like it should be in an `onCreated` event?

impactmass: Yes. I believe you mentioned this before. 

The reason the subscription was done here is because when the subscription is done outside of the `account()` helper, Autoform doesn't pick the new member data change that happens when the "Manage" button is clicked for another member.

Hence the form data remains at the very first member info that was clicked when the taxsetting form was created.

But I'll confirm this again and update.

zenweasel: I thought that was why we created that separate Publication?

impactmass: We created a separate publication so we can have a publication that takes in a userId and just returns the account with that id. Which is different from the way the existing [`Accounts` publication](https://github.com/reactioncommerce/reaction/blob/master/server/publications/collections/accounts.js#L9-L45) works.

impactmass: **Update:**
I was wrong. The subscription can be moved into `onCreated` and it worked as expected (like you said).

```
Template.taxSettingsPanel.onCreated(function () {
  this.subscribe("UserAccount", Meteor.userId());
});

Template.taxSettingsPanel.helpers({
  account() {
    if (Template.instance().subscriptionsReady()) {
      console.log('sub ready');
      const account = Accounts.findOne({ _id: this.member.userId });
      console.log({ account });
      currentAccount = account;
      return account;
    }

    return null;
  },
........
```

I'll modify it.

PR #2114, opened 2017-04-12
Fix Avalara document id

### LINKED ISSUE #2103: [Avalara] Order ID sent to Avalara and Order ID in the Dashboard are not the same
### Expected behavior
The Order ID in the dashboard and the one being sent to Avalara are not the same

### Actual Behavior
They should be the same so you can cross-reference the two

1. With Avalara enabled place an order and then view the order in the Avalara admin console.
1. Observe that the order # displayed in the dashboard (if that is the order ID) does not match with anything in the Avalara admin.

zenweasel:
Resolves #2103 

Turns out just dropping the `code` fields from the SalesOrder allows you to submit the document a second time with an id, so now we pass in the Order ID to Avalara so that you can cross reference them from the Dashboard.


imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
```diff
@@ -369,7 +369,6 @@ function cartToSalesOrder(cart) {
   const salesOrder = {
     companyCode: companyCode,
     type: "SalesOrder",
-    code: cart._id,
```
impactmass: Just to confirm: So when we call `estimateCart` which requires `salesOrder` object... no form of Id/code is needed on it?

impactmass:
I tested like so:

* On current development: I created an Order with Avalara enabled. I checked the order code and compared it to `request.data.code` and `result.code` found in the most recent log in the Avalara logs in the dashboard. They do not match.

* I switched to the branch brent-fix-avalara-documentid, and did the same process and confirmed that the codes match.
Works as described @zenweasel.

Just need a response to my last question and good to go.

imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
zenweasel: That's correct. You **can** pass an ID if you want, but it's not necessary since we don't do anything with the document.

impactmass: OK

PR #2119, opened 2017-04-13
Don't break out orders with quantities

### LINKED ISSUE #2102: [Avalara] Don't break out order items
Currently the order items (**not** cart items) are "exploded" out, adding a new line item for every quantity (e.g. 30 X Item = 30 line items with a quantity of 1).

We want to change the behavior so that order items are kept as an aggregate until they are operated upon, e.g. split-shipped/returned etc. Then they should be broken out.

This change should be made in a backwards-compatible way so that people with existing orders don't experience any issues.

zenweasel:
Closes #2102 and resolves blocker with Avalara

Changes the behavior `copyCartToOrder` to not break out lines with quantities into individual lines

### To Test

This change should have no visibility to the user except for seeing new orders not broken out. The most important thing to test is that nothing is broken. With the exception for adding in some conditionals for Inventory hooks I didn't see any compatibility issues with moving forward from old versions.

### DIFF
--- a/imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
+++ b/imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
@@ -515,7 +515,7 @@ Template.coreOrderShippingInvoice.helpers({
         });
       } else {
         cart = {
-          cartItemId: item._id,
+          _id: item._id,
           productId: item.productId,
           shippingRate: shipment.shipmentMethod.rate,
           items: [item]
--- a/imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
+++ b/imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
@@ -309,16 +309,13 @@ Template.coreOrderShippingInvoice.helpers({
   refundAmount() {
     return Template.instance().refundAmount;
   },
-  /**
-   * Discount
-   * @return {Number} current discount amount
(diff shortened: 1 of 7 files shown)


imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
```diff
@@ -495,43 +490,13 @@ Template.coreOrderShippingInvoice.helpers({
...
-          lineNumber: item.cartItemId
+          lineNumber: item._id
         });
         return _.extend(item, { taxDetail });
       });
     } else {
       items = returnItems;
     }
-
-    /**
```
aaronjudd: We need more comments, not less.

imports/plugins/included/inventory/server/hooks/hooks.js
```diff
@@ -112,9 +112,10 @@ function markInventoryShipped(doc) {
 function markInventorySold(doc) {
   const orderItems = doc.items;
   const cartItems = [];
+  // If a cartItemId exists it's a legacy order and we use that
   for (const orderItem of orderItems) {
     const cartItem = {
-      _id: orderItem.cartItemId,
+      _id: orderItem.cartItemId || orderItem._id,
```
aaronjudd: Is this change / behaviour documented elsewhere? Should be added to release docs?

PR #2130, opened 2017-04-17
Circle release branch builds

jshimko:
This removes the Docker builds for the development branch and replaces it with builds that run on any `release-x.x.x` branches.  If all tests pass and the build succeeds, a release candidate build is pushed to Docker Hub with an RC tag.

For example, any commits to the `release-1.1.1` branch will build and push: `reactioncommerce/reaction:v1.1.1-rc`

We could possibly version the release candidates too (`rc1`, `rc2`, etc), but that felt a little excessive at this point and I opted to just have one RC tag per release branch.


aaronjudd:
Holding this until after the next release (1.2?) -> I'd like the find the time to review the CI 2.0 implementation.

zenweasel:
So I think we are still using `development` as the "latest" branch, so do we still want to do this?

jshimko:
The `:latest` tag is always the latest build of the tagged master branch (and is also equivalent to the most recent version tag).  

reactioncommerce_reaction_-_docker_hub

The thinking there was that you can opt to use the `:latest` tag and you can count on it always being the latest official release.  

Before this PR, the development branch gets built and pushed to a different Docker Hub repo altogether.  `reactioncommerce/prequel:latest` And every new build of it just gets tagged with the Circle build number.  Same idea with the `:latest` tag... it's equivalent to the most recent Circle build number. 

reactioncommerce_prequel_-_docker_hub

If/when this PR is merged, the master branch stays the same.  However, the development branch deployment goes away and the `release-x.x.x` branches trigger the build of release candidate images instead.  This allows all testing of new releases to happen on the `release-x.x.x` branches and we can merge them straight to master when they're ready (which was the original reason for this PR in the first place).  

But to answer your question... if we're still using the development branch like we always have been, then no, this PR isn't necessary.  The whole point of it was just to have some kind of development build happening before we merge stuff to master for release.

zenweasel:
Yes, I believe the decision was made to continue using `development` as we have been.

jshimko:
Ok, hadn't heard that.

aaronjudd:
@jshimko this was still something I wanted to do, but I want it done with Circle 2.0 (as mentioned in comments)...

Issue #2113, opened 2017-04-12
Currency and Language dropdown are not disabled on first start


l0qii:
### Expected behavior
When I create a Shops.json with only 1 currency and 1 language enabled and start reaction for the first time, the currency and language menus in the header should be disabled

### Actual Behavior
The currency and language menus are still active, each with just the single option showing that I had enabled in the Shops.json

### Steps to Reproduce the Behavior
Copy my Shops.json to private/data
reaction reset
reaction
goto localhost:3000, you will see the menus each with a single option

[Shops.zip](https://github.com/reactioncommerce/reaction/files/915235/Shops.zip)

zenweasel:
I verified that this is true. However as a workaround if you click on "All Off" on both Currency and Language both dropdowns disappear for everyone.

kieckhafer:
Yeah, good idea @zenweasel, I didn't see that ticket initially. I'll pick it up.

zenweasel:
Closed via #2148

PR #2148, opened 2017-04-21
language dropdown updates

### LINKED ISSUE #2083: English disappears from language selector
### Expected behavior
When English is enabled it should always appear in dropdown

### Actual Behavior
English disappears from dropdown

1. Login as admin
1. Select "All off" under Languages (English remains enabled)
1. Enable a few other languages
1. View dropdown and observe that English is missing
reaction

kieckhafer:
Fixes #2083 and #2113 

This update does two things.

1) It expands on the “All on / All off” action in the Languages admin, and makes it so the store default language is always set to `true` in the database, since you cannot turn it off.

2) It updates the language dropdown to only show when more than one language is selected


zenweasel:
@kieckhafer This works and fixes the problem, but I wonder if we could do the same for currency and fix #2113 as well?

kieckhafer:
Yeah, good idea @zenweasel, I didn't see that ticket initially. I'll pick it up.

kieckhafer:
@zenweasel I've updated the PR to fix #2113 as well.

PR #2160, opened 2017-04-24
Spencer marketplace shop settings fixes

spencern:
Fixes failing tests and linter errors in marketplace-shop-settings.


spencern:
Deps will get upgraded when we pull in updated RC. Some need to be pinned anyway.

PR #1942, opened 2017-03-06
[WIP][Marketplace] 799 marketplace stripe connect

tdmoneybanks:
This PR aims to provide the reaction application with the ability for the owner to enable stripe connect and set a transaction fee on orders placed through shops on the marketplace. This pr also allows sellers to connect their stripe account to the platform in order to collect payments.

PR #2167, opened 2017-04-26
[Marketplace] Spencer 799 marketplace stripe connect

spencern:
This PR is a replacement for #1942 Marketplace Stripe Connect. This replacement includes almost all of the code in the original, but has been linted and cleaned up for the merge.

A few notes.

1. There is some commented out code in `imports/plugins/included/payments-stripe/server/methods/stripe.js` this code was not ready to be merged and was experimental in nature. Because this PR is replacing #1942, i've commented out the code rather than removing it for now so that the progress towards the Stripe Connect integration is intact. It will be removed or refactored in a future update. I am not indicating that I believe the commented out code is the correct direction, that is a decision that should be made by the implementer when we finish this plugin.

2. There are a number of `import`s that were removed during this cleanup. I _have not_ left those intact or included them in comments. I don't believe that the imports indicated any specific direction and could not tell if they were related to the Stripe Connect integration.

3. I've removed code in `server/methods/core/cart.js` that started to implement splitting items by shopId in an order. I believe that should be a separate PR and issue and the progress was not far enough along to warrant commenting out the code or leaving it in. There is also some higher level discussion that should happen before we start working on cart/order splitting.


imports/plugins/included/payments-stripe-connect/client/settings/settings.js
```diff
@@ -0,0 +1,43 @@
...
+    onError: function () {
+      return Alerts.toast(`${i18next.t("admin.settings.saveFailed")} ${error}`, "error");
+    }
+  }
+});
+
+Template.stripeConnectRedirect.onCreated(function () {
+  // TODO: Verify that this works and define steps to reproduce.
+  // grab stripe connects oauth values and redirect the user
+  const authCode = FlowRouter.getQueryParam("code");
```
aaronjudd: I think this should be Reaction.Router -> directly using flow router could be an issue during upgrades.

spencern: 👍 Not sure how that slipped through

PR #2170, opened 2017-04-26
React unit testing example

### LINKED ISSUE #2066: Propose/Implement React client testing framework
Since we are doing a full push to convert to React, this seems like a prime time to get some testing in place before we end up with 100+ components without tests. The main thrust of this ticket is not to have every component be tested, but to have something in place where developers can quickly and easily write tests for new components they create (and then we will go back and add tests for existing components).

Steps to complete this ticket

1. Evaluate current options for React testing that will work from the command line and in CI
1. Propose a framework/tooling
1. Create a few example tests for existing components
1. Have these run in CI
1. Write a tutorial on how to create tests for components that also includes best practices

joykare:
Resolves #2066


package.json
```diff
@@ -99,12 +99,17 @@
   },
   "devDependencies": {
     "babel-eslint": "^7.2.1",
+    "babel-jest": "^19.0.0",
     "babel-plugin-lodash": "^3.2.11",
-    "babel-preset-stage-2": "^6.22.0",
```
Pheonix7284: This is being used above - it probably still needs to be included or babel will be upset.

jshimko: Uh, yeah... that's not gonna work.

zenweasel: Made this change. No difference.

zenweasel:
So just trying something I started to try and work through the issues it was throwing and they actually look like they are mostly legitimate errors. Possibly there is a change/rule that doesn't allow implicit globals? The only thing that's breaking is declaring `ReactionCore` as a global. I am working through the issues on another branch (in case this turns out to be a dead end)

Whatever this rule is, it's probably a good one to have in place.

zenweasel:
So it appears that adding the `es2015` by default also does `use strict` everywhere (which I didn't realize we *weren't* doing already. As I mentioned most of these are valid errors that should be fixed. The last one is failing that tests and that is this line here in `/server/methods/core/cart.js` where it is trying to assign `exchangeRate` to `currencyCode` which is a string. 

```
    order.billing[0].currency.exchangeRate = exchangeRate;
```

I honestly don't know how this ever worked.

client/api/index.js
```diff
@@ -1,11 +1,6 @@
 import { Reaction } from "/client/modules/core";
 import { Router } from "/client/modules/router";
 
-// Legacy globals
-// TODO: add deprecation warnings
```
zenweasel: Do we want to keep these global?

aaronjudd:
`use strict` use to have more complex implications for Meteor. As of 1.3,  looks like it's safe.  Quick google will show there is some history there.. but looks like as of now, to quote abernix 
>it's up to the developer to add use strict if that's something they would like to use. As mentioned above not doing so could cause inconsistency for developers working on the same project. Additionally, Meteor 1.3 supports ES2015 modules, so if you want to use strict, you'll be able to work around Meteor's global variable style which was previously could have been problematic when combined with use strict.

zenweasel:
So this works except now the products don't load on startup and the failing tests around `currency.exchangeRate`.

zenweasel:
Ok, this appears to be working now? It probably still needs to be tested as undefined variables on the client side don't throw until they actually get evaluated.

zenweasel:
I am running `jest` from the command line and getting this error:

```
  ● OrderSummary snapshot test

    TypeError: Cannot read property 'fullName' of undefined

      at OrderSummary.render (imports/plugins/core/orders/client/components/orderSummary.js:19:106)
      at node_modules/react-dom/lib/ReactCompositeComponent.js:796:21
      at measureLifeCyclePerf (node_modules/react-dom/lib/ReactCompositeComponent.js:75:12)
      at ShallowComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (node_modules/react-dom/lib/ReactCompositeComponent.js:795:25)
      at ShallowComponentWrapper.performInitialMount (node_modules/react-dom/lib/ReactCompositeComponent.js:362:30)
      at ShallowComponentWrapper.mountComponent (node_modules/react-dom/lib/ReactCompositeComponent.js:258:21)
      at Object.mountComponent (node_modules/react-dom/lib/ReactReconciler.js:46:35)
      at ReactShallowRenderer._render (node_modules/react-dom/lib/ReactShallowRenderer.js:126:23)
      at _batchedRender (node_modules/react-dom/lib/ReactShallowRenderer.js:79:12)
      at Object.batchedUpdates (node_modules/react-dom/lib/ReactDefaultBatchingStrategy.js:60:14)
      at Object.batchedUpdates (node_modules/react-dom/lib/ReactUpdates.js:97:27)
      at ReactShallowRenderer.render (node_modules/react-dom/lib/ReactShallowRenderer.js:106:18)
      at ReactShallowRenderer.render (node_modules/enzyme/build/react-compat.js:180:39)
      at node_modules/enzyme/build/ShallowWrapper.js:128:26
      at ReactDefaultBatchingStrategyTransaction.perform (node_modules/react-dom/lib/Transaction.js:140:20)
      at Object.batchedUpdates (node_modules/react-dom/lib/ReactDefaultBatchingStrategy.js:62:26)
      at batchedUpdates (node_modules/react-dom/lib/ReactUpdates.js:97:27)
      at performBatchedUpdates (node_modules/enzyme/build/ShallowWrapper.js:106:42)
      at node_modules/enzyme/build/ShallowWrapper.js:127:9
      at withSetStateAllowed (node_modules/enzyme/build/Utils.js:284:3)
      at new ShallowWrapper (node_modules/enzyme/build/ShallowWrapper.js:126:38)
      at shallow (node_modules/enzyme/build/shallow.js:19:10)
      at Object.<anonymous> (imports/plugins/core/orders/client/__tests__/components/orderSummary.spec.js:35:39)

  ✕ OrderSummary snapshot test (6ms)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        0.877s, estimated 2s
```

mikemurray:
In the `package.json` we should add under `scripts`.

```
"scripts": {
    "test": "jest"
}
```

so we can do `npm test`, and use the local version of jest, vs having to install a global version.

And, add how to run tests to the docs.

circle.yml
```diff
@@ -16,6 +16,9 @@ dependencies:
 test:
   override:
     - reaction test
+    - npm install -g jest
+    - npm update
+    - jest
```
mikemurray: This is unnecessary because you can now do `npm test` and it will run the local copy of `jest` installed with reaction. (If the above code installs a different version of jest than you expect, then unforeseen issues may occur)

zenweasel: Good comment but we probably don't want to call this just `test` as we already have 3 other test runners going on here. So maybe `test-react`? Or just call the script command "jest"?

zenweasel:
@joykare What's the status on the docs for this?

Issue #2115, opened 2017-04-12
Inconsistent Product Image of Cart Items


andela-aatanda:
### Expected behavior
When Cart Icon is clicked and drawer opens, products should retain their uploaded images when any of the items in cart are clicked

### Actual Behavior
When one of two products with an uploaded image in Cart is clicked, the other product defaults to the Basic Reaction Product image and vice versa.

### Steps to Reproduce the Behavior
1. Add an image to an option in the Basic Reaction Product
1. Create a new product and upload an image for it.
1. Publish your changes and toggle edit mode
1. Add both the Basic Reaction Product that has an image and the New Product to cart
1. Open your cart and click any of the product and observe that the product image of the other item defaults to no image and vice versa

PS: The following error appears in console: 
```
TypeError: Cannot read property 'refs' of undefined
    at ProductField.<anonymous> (app.js:20270)
    at CallbackQueue.notifyAll (modules.js:89661)
    at ReactUpdatesFlushTransaction.close (modules.js:89384)
    at ReactUpdatesFlushTransaction.closeAll (modules.js:90820)
    at ReactUpdatesFlushTransaction.perform (modules.js:90767)
    at ReactUpdatesFlushTransaction.perform (modules.js:89414)
    at flushBatchedUpdates (modules.js:89497)
    at ReactUpdatesFlushTransaction.close (modules.js:89372)
    at ReactUpdatesFlushTransaction.closeAll (modules.js:90820)
    at ReactUpdatesFlushTransaction.perform (modules.js:90767)`
```

zenweasel:
I was able to reproduce using the instructions provided.

![product_imges](https://cloud.githubusercontent.com/assets/72819/24987519/f79a493e-2032-11e7-8ddc-48c543e9123a.gif)

andela-aatanda:
This behavior is caused by the `href` in the anchor as I noticed that data is been re-fetched for `cartItems` when trying to fetch data for `PDP`

andela-aatanda:
As a last resort on this, `cartDrawer` will be closed when the `PDP` is visited.

andela-aatanda:
Querying the `Media` Collection seems to be generating inconsistent result as redirecting to `PDP` resends query to `Media` Collection

andela-aatanda:
I was able to work around the `PDP` removing images of other product by using a new Pub/Sub for `cartItemImage`

PR #2171, opened 2017-04-27
[WIP] 2115 Inconsistent Product Image of Cart Item

andela-aatanda:
Resolves [2115](https://github.com/reactioncommerce/reaction/issues/2115)

**Testing**
- Clone branch
- Run on you local machine
- Add an image to an option in the Basic Reaction Product
- Create a new product and upload an image for it.
- Publish your changes and toggle edit mode
- Add both the Basic Reaction Product that has an image and the New Product to cart
- Observe that both products maintain their images when either product is clicked and no errors in console

### DIFF
--- a/imports/plugins/core/checkout/client/container/cartDrawerContainer.js
+++ b/imports/plugins/core/checkout/client/container/cartDrawerContainer.js
@@ -83,9 +83,9 @@ function composer(props, onData) {
   const userId = Meteor.userId();
   const shopId = Reaction.getShopId();
   let productItems = Cart.findOne({ userId, shopId }).items;
-  Meteor.subscribe("CartImages");
 
   productItems = productItems.map((item) => {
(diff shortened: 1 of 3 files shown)


zenweasel:
Looking at this code it doesn't look like it will work if you have multiple items in your cart. Are you sure you have thoroughly tested this?

andela-aatanda:
@zenweasel I have tested this a couple of times with up to 3 items in `Cart` and it behaved as expected.

andela-aatanda:
screen shot 2017-04-30 at 11 06 22 am

zenweasel:
So the behavior should be that when an option does not have an image, it should use the image from the variant or the top-level product (whichever comes first). People are not going to assign images to 20+ individual shoe sizes for example, but we should still show the image of the shoe. Or if an image was assigned for a color, we should show that.

another_dumb_product

PR #2172, opened 2017-04-27
Marketplace core fixes

lcampanis:
Core updates for Marketplace.

The following PR fixes a number of issues around Media, SellerShops and Shops permissions and publications for admin and owner in respect to those collections.

SellerShops is now using observers and passes only the shops needed to the client, with full read/write access to the current shop of the seller.

A new paymentMethods field has added to Shops for knowing which payment methods are enabled for each seller during checkout. In our case we set this upon opening a new shop. It is useful with SellerShops for knowing which method each seller supports, and alleviate some of the pressure in forming complex data structures. 

The code is self-explanatory, but please let us know your thoughts and if you have any questions.


client/modules/accounts/templates/profile/profile.js
```diff
@@ -71,5 +71,9 @@ Template.accountProfile.helpers({
 
   isMarketplaceGuest: function () {
     return (Reaction.hasMarketplaceAccess("guest") && !Reaction.hasAdminAccess());
+  },
+
+  isMarketplaceSeller: function () {
+    return (Reaction.hasMarketplaceAccess() && !Reaction.hasOwnerAccess());
```
spencern: For my knowledge more than anything else, what's the reasoning behind having custom marketplace permissions checkers rather than using the `hasPermission` or `userIsInRole`?

lcampanis: There is always a case where Marketplace might be disabled as a module, or is enabled but doesn't allow for shops to be created if that one option is unchecked. This function covers these cases and makes sure the module is enabled, seller shops can be created, and the user has permission. Otherwise if any of these fail, access is false. We're not very fond of it, but we always had to check these two options along with permissions. What is needed is a global way to know whether the app is operating in marketplace mode or not. and then just have that checked in permissions for us every time. So hasPermission could do the job for us maybe?

server/publications/collections/product.js
```diff
@@ -37,12 +37,12 @@ export function findProductMedia(publicationInstance, productIds) {
   if (sellerShopId) {
     selector.$or = [
       {
-        "metadata.workflow.$in": [null, "published"]
+        "metadata.workflow": { $in: [null, "published"] }
       }, {
         "metadata.shopId": sellerShopId
       }
-    ]; // Product editors can see both published and unpublished images
-  } else if (!Reaction.hasPermission(["createProduct"], publicationInstance.userId, shopId)) {
```
spencern: Why remove this permission check?

server/publications/collections/product.js
```diff
@@ -66,48 +66,36 @@ Meteor.publish("Product", function (productId) {
...
   let productShopId;
   const shop = Reaction.getCurrentShop();
   // verify that parent shop is ready
   if (typeof shop !== "object") {
     return this.ready();
   }
 
-  // selector should come first as default, alterations take place later depending on role
-  const selector = {
-    isVisible: true,
```
spencern: Not sure I understand why we're removing the requirement that a product be visible by default. Seems safer to do it that way and remove that requirement based role than vice-versa.

server/security/collections.js
```diff
@@ -41,7 +41,7 @@ export default function () {
   Security.defineMethod("ifHasSellerRole", {
     fetch: [],
     deny: function (type, arg, userId) {
-      const isDenied = Roles.userIsInRole(["admin", "owner", "createProduct"], Reaction.getSellerShopId(userId));
+      const isDenied = !Roles.userIsInRole(userId, "createProduct", Reaction.getSellerShopId(userId));
```
spencern: So a seller role is anyone who can `createProduct` for a given shop? Would be nice to have a comment here explaining this a little bit.

server/publications/collections/product.js
prosf: This is changed with the last push

server/publications/collections/media.js
```diff
@@ -32,78 +32,82 @@ Meteor.publish("Media", function (mediaFilters) {
...
+          revisionHandle.stop();
+        });
+      }
     }
   }
 
-  return Media.find({
-    "metadata.type": "brandAsset"
+  return Media.find(selector, {
+    sort: {
```
prosf: This Media publication has more filter options than the Reaction's one as it can take a
shop-Ids array or product-Ids array, and return the corresponding images. So to limit the results to
metadata.type "brandAsset" seems a little bit counter-intuitive.

prosf:
Hi Spencer. 
Fixes have been submitted

spencern:
Hi @prosf 
We've still got 3 linter issues: https://www.bithound.io/github/reactioncommerce/reaction/7c1e8b995efd4885b45d0239a35d03bbf31d0fbe/files#lint

There is also still one failing test.

I'd like to get these issues fixed before merging. I don't mind helping with this either if you'd like.

imports/plugins/included/marketplace/server/methods.js
```diff
@@ -13,8 +13,8 @@ Meteor.methods({
    * @returns {Object|null} - The user hash if found, null otherwise
```
spencern: @prosf - This doesn't look like it will ever return null if there are any accounts present

server/publications/collections/media.js
prosf: In the last pushed version ,  "metadata.type": "brandAsset" gets set in the default Media-selector, so has the same behavior as before,(except when Media subscription is called with a filter parameter like  product-Ids))

imports/plugins/included/marketplace/server/methods.js
prosf: These methods ("marketplace/getSeller", "marketplace/getSellerShop") are of no use anymore. (They are replaced by getSeller ,getSellerShop ) of lib/api/core.js

spencern:
@prosf Changes look good. Linter is clean now. Still have that one failing test, once we fix that I think we can merge.

spencern:
@prosf I know you guys have been busy with your launch. Any ETA on when you might be able to come back to this? (that failing test is our blocker here right now)

server/methods/catalog.js
```diff
@@ -832,7 +832,8 @@ Meteor.methods({
     let update;
     // handle booleans with correct typing
     if (value === "false" || value === "true") {
-      update = EJSON.parse(`{${field}:${value}}`);
+      const booleanValue = (value === "true" || value === true);
```
spencern: the second half of this evaluation seems unnecessary. 
Because we're filtering for either `"false"` or `"true"` in the `if` statement above, `value` can never equal `true`.

server/publications/collections/product.js
```diff
@@ -66,48 +58,37 @@ Meteor.publish("Product", function (productId) {
...
-    // we need productShopId in order to check if user can edit this product or view its revisions
+
     const product = Products.findOne(_id);
     if (product) {
       productShopId = product.shopId;
     } else {
       return this.ready();
     }
   } else {
-    const newSelector = Object.assign({}, selector, {
```
spencern: @prosf @lcampanis - this change here breaks the test that we're blocked on. It causes `newSelector` to find invisible products for non-authenticated users.

Is there a marketplace related reason for this or can I fix it by undoing this change?

spencern:
Closing in favor of #2394

PR #2178, opened 2017-04-28
Merge Development into Marketplace

spencern:
Manually resolve conflicts
* imports/plugins/core/checkout/client/templates/cartDrawer/cartItems/cartItems.js
* imports/plugins/core/ui-navbar/client/components/navbar/navbar.js
* imports/plugins/included/payments-stripe/client/checkout/stripe.js


imports/plugins/core/checkout/client/components/cartSubTotal.js
```diff
@@ -0,0 +1,67 @@
+import React, { Component, PropTypes } from "react";
+import { Currency } from "/imports/plugins/core/ui/client/components/";
+
+class CartSubTotal extends Component {
+  validateDiscount() {
+    if (Number(this.props.cartDiscount) > 0) {
+      return (
+        <tr>
+          <td data-i18n="cartSubTotals.discount">Discount</td>
```
aaronjudd: @mikemurray is this pattern for data-i18n correct in this use case?

Issue #2182, opened 2017-05-01
ADA Compliant Links on Social Component


minimart:
### Proposed Enhancement
Social links do not currently comply with ADA or [ WCAG 2.0](https://www.w3.org/TR/WCAG20/) requirements. These links (Facebook, Pinterest, Google Plus, and Twitter) should be ADA/WCAG 2.0 compliant by having text readable by screen readers.

### Benefit to Reaction and Users 
Compliance with these standards is legally mandated in the United States, EU and several other governing bodies. 

### Proposed changes
In imports/plugins/included/social/client/components adding an aria-label to each component, ex:

`<a className="btn btn-flat facebook-share" href="#" onClick={this.handleClick} aria-label="Share to Facebook" target="_blank">
        <Helmet
          meta={getOpenGraphMeta(this.props)}
        />
        <i className={iconClassNames} />
        {this.renderText()}
      </a>`

I will begin to work on this issue, and create a PR for these simple changes to the social plugin when completed.

aaronjudd:
👍 Excellent points.  Related to #405, which we're trying to give a higher priority.

kieckhafer:
Closed in PR #2196

PR #2196, opened 2017-05-01
made social share buttons on PDP ADA compliant

minimart:
Solves Issue #2182 

Aria (and other screen-readers) will now state that each of the social buttons, if clicked, will "Share on {SocialNetworkName}".

#### Testing

Using [aXe](https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd?hl=en-US) or any other ADA compliance checker, review the warnings present on the PDP. You could also utilize a screen reader to hear the text being read to the user. 

_Note: There are a number of other complaints from this tool, some of which I would like to keep making Issues and PRs for to make sure that Reaction Commerce is, on its base, ADA compliant._


imports/plugins/included/social/client/components/facebook.js
```diff
@@ -88,7 +88,9 @@ class FacebookSocialButton extends Component {
     });
 
     return (
-      <a className="btn btn-flat facebook-share" href="#" onClick={this.handleClick} target="_blank">
+      <a className="btn btn-flat facebook-share" aria-label="Share to Facebook" href="#" onClick={this.handleClick}
```
aaronjudd: Perhaps the aria labels also be i18n translated?

minimart: Good call! I'll make a separate namespace for all of the aria translations as well. 👍

Pheonix7284: Generally if you are using a href="#" the a tag should probably be a button tag as you aren't 'linking' or 'anchoring' to another part of the page!

aaronjudd:
@minimart both suggestions are good ones.. I can accept this PR right now though as it certainly causes no harm.

minimart:
Sounds good! This is working for now-- and I can keep working on it and either keep this PR open and update it or open another one-- whichever works better for you!

Issue #1638, opened 2016-12-17
When logged out, cart completed shows "Order Processing"


aaronjudd:
### Expected behavior

When logged out, cart completed page should show a login page.
ie:
`http://localhost:3000/reaction/cart/completed?_id=cE4EBEeYFBXFPqjhG`
should display login/access denied.  

### Actual Behavior
Displays "Order Processing". 

### Steps to Reproduce the Behavior

Open a valid cart/completed url in an incognito window. 
![cart_completed_and_cart_completed_and_your_order_is_confirmed_-_6dd9ge4mwvqwqb42p_-_testing_reactioncommerce_com_-_reaction_commerce_mail](https://cloud.githubusercontent.com/assets/439959/21283976/c05c6258-c3c2-11e6-9a7b-abbbaf438b72.png)

andela-aatanda:
There's a PR open for this [here](https://github.com/reactioncommerce/reaction/pull/2201)

zenweasel:
Closed via #2201

zenweasel:
When you checkout now as an anonymous user with Paypal Express you get a login screen rather than the completed screen. Confirmed that this doesn't happen in `master`. Most likely it is related to the fix for this issue.

impactmass:
@andela-aatanda asked me if the new issue here might have been caused by my permissions PR.... but I've confirmed it's not.

I created a branch with permissions issue fixed (but without React router)... and the checkout worked and I see cart completion page. This _might_ be related/caused by React router.

hrath2015:
@zenweasel  are u able to  do it on latest dev.  Because I am not able to see any payment options listed for guest user due to issue# 2231

PR #2201, opened 2017-05-02
1638 Completed Order Logged Out State

andela-aatanda:
Resolves [1638](https://github.com/reactioncommerce/reaction/issues/1638)

**Testing**
- Add Items to your cart and checkout
- On Order Completed page, copy `URL` and paste in incognito browser mode
- Observe that a `notFound` page is presented to user

**Screenshot**
screen shot 2017-05-03 at 12 06 31 pm


imports/plugins/core/checkout/client/templates/checkout/completed/completed.html
```diff
@@ -35,13 +35,7 @@
...
   {{else}}
     <div class="container">
-      <div class="row text-center">
-        <h1>Order Processing</h1>
-        <p>Please do not click back button.</p>
-        <br />
-        <br />
-        <i class="fa fa-spinner fa-pulse fa-5x fa-fw"></i>
-      </div>
+      {{> unauthorized}}
```
zenweasel: So this change throws "unauthorized" whenever the order is not completed. What if the user is valid but the order is not completed (which can happen with something like Paypal which can take a few seconds)? Rather than eliminating the "Order Processing" page, you need to add another check for whether that user is the user for this order.

imports/plugins/core/checkout/client/templates/checkout/completed/completed.js
```diff
@@ -20,6 +20,22 @@ Template.cartCompleted.helpers({
     }
     return false;
   },
+  ifUserOrder: function () {
```
zenweasel: Let's not start a function name with `if`. If it returns a boolean you want it to start with `is`.  `+{{#if ifUserOrder}}` is some awkward syntax.

imports/plugins/core/checkout/client/templates/checkout/completed/completed.js
```diff
@@ -20,6 +20,22 @@ Template.cartCompleted.helpers({
...
+    const id = Reaction.Router.getQueryParam("_id");
+    if (id) {
+      const ccoSub = Meteor.subscribe("CompletedCartOrder", Meteor.userId(), id);
+      if (ccoSub.ready()) {
+        const order = Orders.findOne({
+          userId: Meteor.userId(),
+          cartId: id
+        });
+        if (order) {
```
zenweasel: Can we  just return !!order?

zenweasel:
So when I checkout now, I see a the "Page Not Found" page for a loooong time and the finally the order page displays. Something is broken with the logic here. I confirmed that this doesn't happen on `development`. This also makes me think that you didn't really spend much time testing this as I discovered this problem **literally** following your testing instructions.

I am closing this PR now and you can reopen once you have fully tested it.

Issue #1882, opened 2017-02-22
Order number is not shown on order completed screen


zenweasel:
### Expected behavior
Order ID should be shown on completed screen

### Actual Behavior
It's not

### Steps to Reproduce the Behavior
1. Place an order
1. Observe that order ID is in the URL, but no unique identifier is visible.

zenweasel:
Closed via #2202

PR #2202, opened 2017-05-02
Add OrderId To Checkout Completed Screen

michael-alade:
**Resolves**
#1882 

**Test**
1. Make an order.
1. Pay for the order
1. Order Id should be display on the right corner with time the order was placed


zenweasel:
@rymorgan Do you have an opinion on how/where to display this order ID in the completed screen?

cart_completed

Issue #2207, opened 2017-05-03
Order Summary not showing the correct data


kieckhafer:
When using Reaction as an Admin, the Orders panel is not correctly displaying order information in the Summary section.

1) If you have not yet attached an Address to your Admin account, then the summary section does not display at all

![pasted_image_at_2017_05_03_04_10_pm](https://cloud.githubusercontent.com/assets/4482263/25684184/e507063a-3014-11e7-89cd-a738ace59c3a.png)

2) If you have an address attached to your Admin account, then the Summary panel displays the address of the Admin, not the address attached to the order.

![pasted_image_at_2017_05_03_04_12_pm](https://cloud.githubusercontent.com/assets/4482263/25684232/0c359ca8-3015-11e7-9506-699522002ad6.png)

spencern:
Was able to confirm the issue.
My Repro steps:

1. clone `development` branch
2. setup shipping and payment, do not setup admin account address
3. Checkout with non-admin account (incognito, anonymous checkout works fine)
4. Attempt to fulfill order. Note that Summary is empty.
5. Add address to admin account - make sure it's different than the address you checked out with.
6. Re open order from admin account. Ship To address and all other address information will be related to administrator address, not customer address in the summary panel.

zenweasel:
@kieckhafer @spencern This is ready to be closed right? Do I need to confirm fixed?

kieckhafer:
Yes this can be closed.

PR #2208, opened 2017-05-03
update the way address data is provided

kieckhafer:
Fixes issue #2207

### DIFF
--- a/imports/plugins/core/orders/client/containers/orderSummaryContainer.js
+++ b/imports/plugins/core/orders/client/containers/orderSummaryContainer.js
@@ -125,7 +125,7 @@ const composer = (props, onData) => {
       "shipping._id": props.fulfillment._id
     });
 
-    const profile = order.shipping[0].address;
+    const profileAddress = order.shipping[0].address;
 
     if (order.workflow) {
       if (order.workflow.status === "coreOrderCreated") {
@@ -136,7 +136,7 @@ const composer = (props, onData) => {
 
     onData(null, {
       order: order,
-      profile: profile
+      profileAddress: profileAddress
     });
   }
 };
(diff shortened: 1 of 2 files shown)


imports/plugins/core/orders/client/containers/orderSummaryContainer.js
```diff
@@ -116,25 +116,19 @@ class OrderSummaryContainer extends Component {
...
-    }
 
+  if (orderSub.ready()) {
+    // Find current order
     const order = Orders.findOne({
       "_id": props.orderId,
       "shipping._id": props.fulfillment._id
     });
 
+    let profile = {};
```
zenweasel: Why do we initialize this and then immediately reassign it?

kieckhafer: Yeah.... good question, definitely don't need to do that... 

was rushing through, just copy/pasting as much as needed, in the old version we could re-assign it a couple times and it was necessary. Updated to make more sense.

imports/plugins/core/orders/client/containers/orderSummaryContainer.js
```diff
@@ -116,25 +116,17 @@ class OrderSummaryContainer extends Component {
...
-    }
 
+  if (orderSub.ready()) {
+    // Find current order
     const order = Orders.findOne({
       "_id": props.orderId,
       "shipping._id": props.fulfillment._id
     });
 
+    const profile = order.shipping[0].address;
```
zenweasel: Since this is an address not a profile I wonder if we shouldn't call this variable `profileAddress`? I realize that's how it was in the original code but maybe we can make this easier to read.

spencern: Agreed that this would make more sense if we named it `profileAddress` or even `profileShippingAddress`

spencern:
1 new failing dependency we should fix in another branch. Twilio has a new major version

Issue #2125, opened 2017-04-14
Email address not verified for new users after clicking through email


spencern:
### Issue
When creating a new user, an email is sent out with a link to "Get Started". This link does not mark the user's email as "verified"

![image](https://cloud.githubusercontent.com/assets/1203639/25052955/c44f7cc2-2111-11e7-8aa3-464123d51db8.png)

![image](https://cloud.githubusercontent.com/assets/1203639/25053010/0697feec-2112-11e7-9edb-498d8a71b428.png)

I'd expect that after clicking the "Get Started" link in the email, the email address would be verified.

### Reproduction Steps
1. Clone from dev
2. Add email provider via dashboard
3. Create new user
4. Follow link in email
5. Observe database showing email is still `verified: false`

michael-alade:
A PR #2216 has been opened for this issue.

michael-alade:
@spencern After attaching a verification link to the "Get Started" button for the welcome email, everything seemed to be fine with having a new user verifying his account via email. But we discovered that if an admin tries inviting a new user the user gets two emails, the add welcome email and the add password email. At this point, it's a little bit confusing what should be done about verification because this might have its effect on the marketplace feature. What's your opinion?

spencern:
My opinion is that we should try to combine the welcome email and the create password email into a single email when invited. This email should accomplish a few things:
1. Notify the user that they've been invited to join shop {shopname} by {user that invited them}
2. Provide the user with an opportunity to finish filling out their account details, username, password, address, etc
3. Verify the users email address if they choose to use the email they were invited with, OR send an new verification email if the user chooses to use a different email address than they were invited with.
4. If the user was added into any permission groups/audiences or given any shop specific permissions, make sure those are granted.

5. *Future feature - Trigger drip tutorial campaign based on the permissions the user has been granted. E.g. shop owner campaign might get a set of emails like this over the next month ["Setting up your shop", "adding your first product", "automate shipping with Shippo", "fulfilling orders", "understanding reaction analytics", "how to effectively market your shop", etc]

spencern:
@michael-alade When at all possible we should try to accomplish trigger based actions ("verify email address", etc) with existing emails ("welcome to Reaction / your shop").

michael-alade:
@spencern I think the new functionality for the invite email should be created as a new issue.

zenweasel:
Closed via #2216

PR #2216, opened 2017-05-05
Add Verification Url to Welcome Email

michael-alade:
**Resolves**
#2125 

**Testing**
1. Clone from dev
1. Add email provider on dashboard
1. Sign as a new user
1. Follow the get started button on the welcome email
1. Observe that the user's document in the database is now showing `true` for the verified field


zenweasel:
So I seem to be getting two welcome emails now.
inbox__1__-_brent_reactioncommerce_com_-_reaction_commerce_mail

aaronjudd:
@zenweasel I'm not sure that's new. I've been seeing that as well in my recent testing.

zenweasel:
After following the link and setting my password the verified field still shows as `false`.

robomongo_0_9_0-rc9

zenweasel:
I verified that you get the two emails on `development` as well.

Well, this is a bit of a mess. So when I invite a user I get two emails, one that allows me to set my password and one that supposedly verifies my email, except it doesn't actually verify my email. Since the first one allows me to set my password and I can't login until I set my password it seems like that's the one that we want and this second serves no purpose.

client/modules/accounts/templates/verify/verifyAccount.js
```diff
@@ -0,0 +1,19 @@
+import { Meteor } from "meteor/meteor";
+import { Template } from "meteor/templating";
+import { Reaction } from "/client/api";
+import { Accounts } from "/lib/collections";
```
zenweasel: Unused import

zenweasel:
The issue is with this is that the email in the link is incorrect. The link sent to `brent@reactioncommerce.com` looks like this:
`http://localhost:3000//reaction/account/profile/verify?email=brent@hoover.net`

zenweasel:
The `verified` field in the `user` collection now shows verified, but the field in the `Accounts` collection still shows `false`. Not sure why we have both of these fields or which one is important. @spencern's original ticket does not specify either.

spencern:
@zenweasel I don't think I was aware of the field in Accounts. Probably should set both to `true`

aaronjudd:
Not 100% sure about this, but I think `verified` probably only exists in `Accounts` because when the user is first created, the user record is clone to the Accounts collection.. ? (thus before verification..)

zenweasel:
When I click on the email I am getting a 404 page. I think this was redirecting me to the home page before (could be React Router related?)

Steps to Replicate:

1. `reaction reset`
1. Click on "Sign in" from the main menu
1. Enter email and password
1. Open email and click on "Get Started" link
1. Observe "Oops! This page was not found"

server/methods/accounts/accounts.js
```diff
@@ -1,5 +1,6 @@
 import _ from "lodash";
 import moment from "moment";
+import { Meteor } from "meteor/meteor";
```
zenweasel: Import order

server/methods/accounts/accounts.js
```diff
@@ -145,6 +146,35 @@ function validateAddress(address) {
  * Reaction Account Methods
  */
 Meteor.methods({
```
zenweasel: So the style that we want to use for Meteor methods is to `export` them as individual method, and then just declare them as Meteor methods at the bottom of the file. This allows them to more easily be tested or used by other server-side code.

zenweasel:
It looks like this flag is still not being set correctly.

robomongo_0_9_0-rc9

aaronjudd:
@zenweasel @michael-alade what is the status of this PR?

michael-alade:
@zenweasel The changes have been made, and it's ready for another review.

server/methods/accounts/accounts.js
```diff
@@ -8,10 +9,46 @@ import { Logger, Reaction } from "/server/api";
...
+  }
+  return false;
+}
+
+/**
  * @summary Returns the name of the geocoder method to use
  * @returns {string} Name of the Geocoder method to use
  */
-function getValidator() {
+export function getValidator() {
```
zenweasel: these functions were purposefully not exported as they are not part of the API

server/methods/accounts/accounts.js
```diff
@@ -141,65 +178,147 @@ function validateAddress(address) {
...
+      Accounts.update({
+        "userId": userId,
+        "profile.addressBook.isBillingDefault": true
+      }, {
+        $set: {
+          "profile.addressBook.$.isBillingDefault": false
         }
-      }
-      // then change the address that has been affected
+      });
```
zenweasel: While I support moving these into a more standard format, it is best practice to limit your PR to just changes that affect the issue you are trying to fix (it's in our docs and in the PR template). These changes make it difficult to see what is affecting this issue and what is "refactoring". It also makes it difficult for me to know what to test now.

zenweasel:
This flag is still not being set correctly. This is now the fourth time I have noted this.

robomongo_0_9_0-rc9

aaronjudd:
Is this [WIP]?

michael-alade:
@aaronjudd I'm just making a few changes before I have one of the team members test to make sure it works for them.

server/methods/accounts/accounts.js
```diff
@@ -1,3 +1,4 @@
+import { Meteor } from "meteor/meteor";
```
michael-alade: I've made changes to the import order on this file, but it doesn't seem to be reflecting on this comment.
https://github.com/reactioncommerce/reaction/blob/michael-verify-email-address-2125/server/methods/accounts/accounts.js

zenweasel:
Testing this with registering a user this works, but inviting a user still does not set the `Account` to `true`. The original replication steps were vague about how exactly to create a user but it should be for either method.

michael-alade:
I am currently working to make sure the same verification functionality works when a user is invited to the shop from the dashboard.

michael-alade:
@zenweasel The email verification for an invited user works fine now.

Issue #1923, opened 2017-02-28
Payment provider name inconsistently translated


zenweasel:
### Expected behavior
All payment provider names should either be translated or not translated (probably not translated?)

### Actual Behavior
Some are and some are not. 
![reaction](https://cloud.githubusercontent.com/assets/72819/24530874/742d9592-15e7-11e7-8e0f-2fb84a5d826d.png)

### Steps to Reproduce the Behavior
1. Log in as an admin
1. Go to Payment in the dashboard
1. Set language to be something besides English
1. Observe that some payment providers are translated, some are not.

aaronjudd:
PayPal etc probably are brands.. not translatable word? not sure though..

zenweasel:
Right, but Stripe should probably not be translated, as well as Braintree.

zenweasel:
"Example Payment Method" could be translated, since it's not brand

aaronjudd:
True. Is there a translation in i18n?

zenweasel:
```
[{
  "language": "Arabic",
  "i18n": "ar",
  "ns": "reaction-stripe",
  "translation": {
    "reaction-payments": {
      "admin": {
        "shortcut": {
          "stripeLabel": "شريط"
        },
        "dashboard": {
          "stripeLabel": "شريط",
          "stripeDescription": "المدفوعات شريط"
        },
        "paymentSettings": {
          "stripeLabel": "شريط",
          "stripeSettingsLabel": "شريط",
          "stripeSettingsDescription": "لم يكن لديك معرف العميل الشريط API؟",
          "stripeSettingsGetItHere": "أحضره هنا"
        }
      }
    }
  }
}]
```

zenweasel:
Trying to change these in LingoHub (I had already modified a bunch of these in code but some I can't really fix). But when I go into LingoHub the word shows as "locked" with the status "ordered". Have a support request in with LIngoHub

![editor___lingohub](https://cloud.githubusercontent.com/assets/72819/24784874/18229ae4-1b88-11e7-8943-8b94208fe98b.png)

aaronjudd:
I actually think you need to leave a message for their support, I think that we paid for some translations that were never fulfilled...

zenweasel:
I have been working in Lingo hub to set up "Terms" which you can specify global translations but I keep getting distracted or need to ask their support a question so I get delayed a day to wait for their response. So progress is happening, just super slowly.

zenweasel:
[...]

PR #2218, opened 2017-05-05
LingoHub based on development

aaronjudd:
Pull request created by LingoHub based on branch development


imports/plugins/included/payments-stripe/server/i18n/es.json
```diff
@@ -6,11 +6,11 @@
...
+          "stripeLabel": "Stripe"
         },
         "dashboard": {
-          "stripeLabel": "Raya",
-          "stripeDescription": "pagos de la raya"
+          "stripeLabel": "Stripe",
+          "stripeDescription": "pagos de la Stripe"
         },
         "paymentSettings": {
           "stripeLabel": "Raya",
```
zenweasel: I don't understand why it fixed the translations 2 lines above but not the translation here

zenweasel:
Okay, pretty sure this solves #1923 now. Audited every instance of payment methods I could find. Also added Terms for social networks.

imports/plugins/included/payments-stripe/server/i18n/es.json
zenweasel: This diff shows this as not being reset back to Stripe but that's not what's in the file. 🤔

Issue #1424, opened 2016-09-23
Translations.json - the shopId is not populate at shop creation time


EpifaniaCollection:
### Expected Behavior
shopId filed in the Translations.json should be filled it when a brand new shop created when a brand new env is created.

### Actual Behavior
shopId is null at startup of a clean env => language change is not saved.

### Steps to Reproduce the Behavior

`docker-compose -f docker-compose-reactioncommerce-dev.yml up`
and check the created mondgo

jshimko:
This was "fixed" multiple times, but I believe we finally got it in #2224.  The changes are in the development branch now.  If you're still experiencing this issue, make sure to try the development branch from any time after 13 days ago.

I'm closing this for now, but feel free to reopen if the problem returns.

PR #2224, opened 2017-05-08
Fix missing shop

jshimko:
This update ensures the default fixture data loading happens before `Reaction.init()` even runs.  It was previously running in the `onCoreInit` hook and that created a potential for race conditions because there's no simple way to guarantee the order that those hooks run in.  By default, they run in the order in which they're added to the queue and that is largely dictated by the Meteor load order and a bunch of imports across the app.

The original reason this was a problem (#1424) was because translations were _also_ loaded in the `onCoreInit` hook and this occasionally allowed it to run before the initial shopId was available and that prevents translations from loading properly on the client because of the translations pub/sub relies on using the current shopId as a filter.

Now all the fixture data loads before anything else and there's a `while` loop that checks for the default shopId to exist before allowing the rest of `Reaction.init()` to continue.

Unrelated, this also fixes a debug log that wasn't at the right scope and was returning an inconsistent result.  Essentially, it was saying `reaction.json` was not loaded and then immediately afterward it said it _was_ loaded (which it definitely wasn't).

While I was there, I also took the opportunity to fix a few `Logger.error()` instances that had the args in the wrong order.  For the Bunyan error object parsing to work, the error object needs to be the first argument, so it should always be: 

```js
Logger.error(errorObject, "Some human readable string");
```

### DIFF
--- a/server/api/core/core.js
+++ b/server/api/core/core.js
@@ -388,11 +388,11 @@ export default {
       // or attempt to load reaction.json fixture data
       try {
         registryFixtureData = Assets.getText("settings/reaction.json");
+        Logger.info("Loaded \"/private/settings/reaction.json\" for registry fixture import");
       } catch (error) {
         Logger.warn("Skipped loading settings from reaction.json.");
         Logger.debug(error, "loadSettings reaction.json not loaded.");
       }
-      Logger.info("Loaded \"/private/settings/reaction.json\" for registry fixture import");
     }
 
     if (!!registryFixtureData) {
--- a/server/api/core/core.js
+++ b/server/api/core/core.js
@@ -15,8 +15,15 @@ import { getMailUrl } from "./email/config";
 export default {
 
   init() {
+    // make sure the default shop has been created before going further
+    while (!this.getShopId()) {
+      Logger.warn("No shopId, waiting one second...");
+      Meteor._sleepForMs(1000);
+    }
+
     // run onCoreInit hooks
-    Hooks.Events.run("onCoreInit", this);
+    Hooks.Events.run("onCoreInit");
+
     // start job server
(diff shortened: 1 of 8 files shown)


server/api/core/core.js
```diff
@@ -385,11 +388,11 @@ export default {
       // or attempt to load reaction.json fixture data
       try {
         registryFixtureData = Assets.getText("settings/reaction.json");
+        Logger.info("Loaded \"/private/settings/reaction.json\" for registry fixture import");
       } catch (error) {
         Logger.warn("Skipped loading settings from reaction.json.");
         Logger.debug(error, "loadSettings reaction.json not loaded.");
```
aaronjudd: Hey, what about me? Why am I not deserving of a fix as well?

zenweasel:
While I think this is the fix we are looking for, I can't replicate this error when running a local Docker container so I can't verify that it fixes the problem.

server/startup/index.js
```diff
@@ -12,7 +11,6 @@ export default function () {
   Accounts();
```
zenweasel: Should we move the rest of these (or maybe just `initTemplates`) to `init.js`. I feel like even though I have done it as well, having functions run in an `index.js` is sort of an anti-pattern.

jshimko: That definitely occurred to me, but I opted not to because that meant I should probably move all of those files from `/server/startup` to `/server/api/core`.  There's certainly nothing wrong/difficult about doing that, but I was trying to keep things as simple as possible for the sake of a clear PR.  I'm totally not against it though.  I think it makes a lot of sense to kick everything off right inside the init function.  That makes it way easier to see what's going on and in what order.

Issue #2044, opened 2017-03-27
Convert to react: Alerts, Language and other dropdowns


rymorgan:
Convert to react: Alerts, Language and other dropdowns related to #2040

zenweasel:
@rymorgan Can you clarify what you mean by Alerts here? Do you mean the notifications?
![reaction](https://cloud.githubusercontent.com/assets/72819/25654704/b3574c3e-3024-11e7-8a23-1e21ca4e5ef9.png)

joykare:
Notes:

- Branch `joykare-navbar-dropdowns-2044`
- @rymorgan hoping that alerts means the notifications. If so its already in React.
- Starting with the currency dropdown /client/modules/i18n/templates/currency

joykare:
Notes:
- Using DropDownMenu Component on  `/imports/plugins/core/ui/client/components/menu`
- Working on the language dropdown `/client/modules/i18n/templates/header/i18n.html`

joykare:
@mikemurray I'm using <DropDownMenu /> component for the drop downs and I'm trying to figure out a way to close the `menu` when one of the `menuitems` is selected. Any ideas on how to make that happen?

mikemurray:
**This will require some additions to both DropDownMenu and the Popover component**

1. The Popover manages state for the open / close status. That will need to be exposed to props so it could/should be managed by the parent component. __Probably best to remove all the state management from that component and replace with callbacks, and accept the open/close state as props.__
1. The dropdown itself menu should have an open/close state var. 
1. In the `handleMenuItemChange` callback in the dropdown would set the open/close state and pass that to the popover component via props.

joykare:
Thanks that worked @mikemurray

PR #2230, opened 2017-05-09
Navbar dropdowns conversion to React

joykare:
Resolves:
#2044

**Description**
- The language and currency dropdown on the navbar should work exactly as they were in the Blaze implementation

**Testing**
- Click on the currency || language button on navbar
- Change language || currency
- Observe that there aren't any errors and that language || currency actually change


zenweasel:
Comparing this with the code on `master` I see a couple of discrepancies, nothing major.

1. There is no "down arrow" next to the language icon which helps to make it clearer that this is a dropdown.
1. When you select a language/currency the dropdown doesn't close.
1. The search bar seems to have disappeared when I am not logged in?
1. The spacing between the currency and language selectors seems to be different.
1. When I enable all currencies the dropdown opens upward  
reaction

kieha:
I've just tested this out on `development` and the search bar does not appear when logged out @zenweasel.

joykare:
@zenweasel Ready for another check

zenweasel:
Just noting that this actually **fixes** a bug in `master` where both the currency and language dropdowns would extend past the browser border and there was no way to get to them.

reaction_and_inbox__36__-_brent_thebuddhalodge_com_-_the_buddha_lodge_mail_

client/modules/i18n/templates/currency/containers/currencyContainer.js
```diff
@@ -0,0 +1,96 @@
...
+        currentCurrency = localStorageCurrency + " " + shop.currencies[localStorageCurrency].symbol;
+      } else if (locale && locale.currency && locale.currency.enabled) {
+        currentCurrency = locale.locale.currency + " " + locale.currency.symbol;
+      } else {
+        currentCurrency = shop.currency + " " + shop.currencies[shop.currency].symbol;
+      }
+    }
+
+    if (Match.test(shop, Object) && shop.currencies) {
+      for (const currencyName in shop.currencies) {
```
joykare: Why `of` I want to iterate over the object `shop.currencies` and get all the keys. I think `for-in` works well for this.

zenweasel: Because `in` iterates over not just the property of this object but everything up the prototype chain, so typically you would want to pair it with a `hasOwnProperty` to avoid leakage from a parent object.

I would probably use `of`` and `Object.keys` but if you are sure this won't ever cause any problems I am fine with it.

PR #2145, opened 2017-04-21
Fix permissions for items shown in admin dashboard

impactmass:
First part of the resolution of #1651.
Also resolve #1641. 

**What this PR solves:**
- When owner gives dashboard access to an account, the account has access to everything manageable in the dashboard. This PR fixes that. So, only the packages that the account has access to shows up in the dashboard.

**To test:**
- Run the app on this branch.
- Create an account other than admin account (you can use incognito or another browser... so you have both the admin and the new account opened)
- Grant dashboard access to that account (i.e toggle on "Dashboard" in Accounts -> Manage for that user)
- Go through items showing in the dashboard for that account
- Confirm that not all available settings show up in the dashboard for that account.
- Also confirm that publish buttons/controls don't show up at the top bar (except with access to "createProduct")... [see screenshot](https://cloud.githubusercontent.com/assets/439959/25412459/b41c7a3e-29d7-11e7-8c18-73059ad90387.png)
- Toggle on and off dashboard items for this account from your own admin account, then go to the dashboard for the test account to see if access is being given accordingly.

PR #2241, opened 2017-05-11
Revert "Fix permissions for items shown in admin dashboard (#2145)"

zenweasel:


spencern:
Was able to check out

image

image

aaronjudd:
Was able to checkout out, and permissions to view payment methods as guest. However @mikemurray what I saw, was the bug I reported in router, that I wasn't able to test at the last second after I merge the permissions pr.  Mainly...  that everyone should be testing these PR's with  paypal express..  as I now am able to replicate that I get a login screen after order/completed.

reaction-paypal_paypaldone_and_reaction

PR #2248, opened 2017-05-11
Add the Vietnamese Dong as a supported currency

fuo:
Hi kieckhafer This is a PR based on the examples for adding new currency:
https://docs.reactioncommerce.com/reaction-docs/master/reaction-localization


zenweasel:
@fuo I think you want to also add an entry around line 472, there needs to be one entry for each shop. I would just do it for you but I don't think I have write permissions

fuo:
Thanks, @zenweasel I will update it later today.

Issue #1313, opened 2016-08-19
Auto-completed tags in nav should be filtered so you cannot see already used tags


kieckhafer:
### Expected behavior

Tags that have already been used in the navigation should not show up in the auto-complete dropdown.
### Actual Behavior

When adding new tags into the navigation, the auto-complete feature shows tags that have already been used. When you click on an already used tag, it just kind of disappears and nothing happens.
### Steps to Reproduce the Behavior

Add new tags to a product or the nav. Try to add the same tag again, and see it show up in the auto-complete dropdown.

![screen shot 2016-08-19 at 10 08 54 am](https://cloud.githubusercontent.com/assets/4482263/17817972/11e7040a-65f6-11e6-8743-59810c9febad.png)

zenweasel:
@kieckhafer Are you still working on this?

kieckhafer:
@zenweasel yeah, this still needs to be done. I'll get going on it again.

zenweasel:
@kieckhafer If you are still working on other things can we take this back from you? We want to start working in the Navbar React rewrite.

kieckhafer:
Yes please do... I just removed myself from the assignment.

impactmass:
will be fixed in #2043

impactmass:
resolved in #2306

Issue #1312, opened 2016-08-18
Errors when changing tag hierarchy in main navigation


mikemurray:
### Expected behavior

When adding an existing tag that is a tag group, a nested child of a top level navigation item to the main navigation, there should be no errors.
### Actual Behavior

Errors appear and become persistent.

```
Exception in template helper: TypeError: Cannot read property 'relatedTagIds' of undefined
Exception in template helper: TypeError: Cannot read property 'isEditing' of undefined
```
### Steps to Reproduce the Behavior
1. Create a new top-level navigation item
2. Expand drop down
3. Add a `Tag Group` with a new name
4. Add an item to that tag group
5. In the main navigation, add a new top-level, use the same name you used for the `Tag Group`
6. Be amazed as your js console blows up with errors.

zenweasel:
Verified that this error still happens in version 1.1.1.

impactmass:
fix going in #2251

impactmass:
resolved in #2306

PR #2251, opened 2017-05-11
Convert to react: Tag navigation

### LINKED ISSUE #2043: Convert to react: Tag navigation
Convert to react: Tag navigation related to #2040

impactmass:
Update: 22/05
Resolves #2043 TagNav React Conversion

**Things to Test:**

1.  TopLevel Tags List
    * Adding a new Tag 
        * Type into the tag box, save by clicking "enter". Also save by clicking the "+" icon
    * Drag tags around to re-order them
    * Delete tags by clicking the "x" icon
    * Add more tags

2.  Second Level Tags (TagTree)
    * In Edit mode, click the drop caret beside a TopLevel tag to expand dropdown
    * Add header tags
    * Add body tags under those header tags
    * Confirm you can edit the tags and can delete (with the "x" icon)

3.  Drag & Drop
    * Drag to re-order tags under a tagTree header
    * Dragging the tagTree header should lift all subtags under it (in-prog)

4.  Related feature to check
    * Product tags: TagNav share components with Product tags... check to confirm product tag functionality is still intact.

5.  Confirm that these bugs are fixed
    * #1312
    * #1313

Pending issue: On mobile, clicking the hamburger icon doesn't show the tagNav. Being worked on


imports/plugins/core/ui-tagnav/client/components/tagTree.js
```diff
@@ -0,0 +1,163 @@
+import React, { Component, PropTypes } from "react";
```
impactmass: @kieha I need little help with this component.... specifically making it share a number of functions from the container file above. I've done the usual passing them in as props, but much of them are not doing what is expected. 

Running this branch gives a clearer picture of what I'm saying.... so we can sync on this. My guess is that I'm making a React beginner mistake there.

kieha:
You're rendering the `TagTree` component here https://github.com/reactioncommerce/reaction/blob/c0c5081169a41dc9b681a20d949f7ad7b7155321/imports/plugins/core/ui/client/components/tags/tags.jsx#L117, but you're not passing in the props you want to use down to it.

Also, you're using functions that you don't have, e.g. https://github.com/reactioncommerce/reaction/blob/seun-convert-tagnav-react-2043/imports/plugins/core/ui-tagnav/client/components/tagTree.js#L60, yet you haven't defined the function `handleTagMouseOut` in that component.

That's part of the reason that stuff isn't working.

impactmass:
Thanks @kieha, fixing those moved me a step further. I intended passing those from the parent,done that now.

impactmass:
Hi @joykare.... do you mind taking a first pass test on this PR? I know I still have the WIP tag on it... I'm resolving some style issues currently, but I don't want to hold back getting test feedback till later.

imports/plugins/core/ui-navbar/client/components/navbar/navbar.js
```diff
@@ -90,27 +92,36 @@ Template.CoreNavigationBar.helpers({
     };
   },
 
-  tagNavProps() {
+  tagNav() {
```
joykare: This function should be moved into the Tag Nav container in a compose function. Which would pass data into the container as props. 
Example code: https://github.com/reactioncommerce/reaction/blob/brent-milestone-54/client/modules/i18n/templates/currency/containers/currencyContainer.js#L44

imports/plugins/core/ui-tagnav/client/components/tagNav.js
```diff
@@ -0,0 +1,71 @@
...
+
+const TagNav = (props) => {
+  return (
+    <div className={`rui tagnav ${props.navbarOrientation} ${props.navbarPosition} ${props.navbarAnchor} ${props.navbarVisibility}`}>
+      <div className="navbar-header">
+        <p>Header</p>
+      </div>
+      <DragDropProvider>
+        <TagList
+          {...props}
```
joykare: With { ...props } you don't have to list out all the props. https://zhenyong.github.io/react/docs/jsx-spread.html

imports/plugins/core/ui-tagnav/client/components/tagNav.js
```diff
@@ -0,0 +1,71 @@
...
+          onTagSave={props.handleTagSave}
+          onTagUpdate={props.handleTagUpdate}
+          onTagSelect={props.onTagSelect}
+        />
+      </DragDropProvider>
+    </div>
+  );
+};
+
+TagNav.propTypes = {
```
joykare: With the spread operator you don't require type-checking with PropTypes.

imports/plugins/core/ui-tagnav/client/components/tagTree.js
```diff
@@ -0,0 +1,173 @@
...
+          <span className="title">{this.state.parentTag.name}</span>
+          <a href="#">View All <i className="fa fa-angle-right" /></a>
+        </div>
+        <div className="content">
+          {this.renderTree(this.tags)}
+          {this.props.editable &&
+            <div className="rui grouptag create">
+              <div className="header">
+                <TagItem
+                  blank={true}
```
joykare: You dont have to pass this down as props. Just set whatever you want to be `true` as `true`

imports/plugins/core/ui-tagnav/client/containers/tagNavContainer.js
```diff
@@ -0,0 +1,367 @@
...
+const NavbarAnchor = {
+  Top: "top",
+  Right: "right",
+  Bottom: "bottom",
+  Left: "left",
+  None: "inline"
+};
+
+const TagNavHelpers = {
+  onTagCreate(tagName, parentTag) {
```
joykare: Just call the `TagHelpers` directly when needed. This seems redundant.

imports/plugins/core/ui-tagnav/client/containers/tagNavContainer.js
```diff
@@ -0,0 +1,367 @@
...
+        name: ""
+      }
+    };
+  }
+
+  componentDidMount() {
+    $(window).on("resize", this.onWindowResize).trigger("resize");
+  }
+
+  componentWillReceiveProps(nextProps) {
```
joykare: I dont think this is necessary. Instead of passing props into state then using that state, just use them as props.

imports/plugins/core/ui-tagnav/client/containers/tagNavContainer.js
```diff
@@ -0,0 +1,367 @@
...
+  handleClearSuggestions = () => {
+    this.setState({ suggestions: [] });
+  }
+
+  get canEdit() {
+    return this.props.hasEditRights && Reaction.isPreview() === false;
+  }
+
+  attachBodyListener = () => {
+    document.body.addEventListener("mouseover", this.closeDropdown);
```
joykare: jQuery in React. ouch..
https://facebook.github.io/react/docs/events.html#mouse-events

imports/plugins/core/ui/client/components/tags/tags.jsx
```diff
@@ -55,53 +59,117 @@ class Tags extends Component {
...
-            onTagMouseOver={this.handleTagMouseOver}
-            onTagRemove={this.handleTagRemove}
-            onTagSave={this.handleTagSave}
-            onTagUpdate={this.handleTagUpdate}
-            suggestions={this.props.suggestions}
-            tag={tag}
-          />
+          <div className={classAttr} key={index}>
+            <TagItem
+              {...this.props.tagProps}
```
joykare: { ... this.props } already takes care of that. Refer above

imports/plugins/core/ui-tagnav/client/components/tagTree.js
impactmass: I'm not sure I get you here. The blank `prop` can be true or false. The component renders differently  based on this.

imports/plugins/core/ui-navbar/client/components/navbar/navbar.js
impactmass: I saw these two patterns in the code. I thought any can be used. I'll update to this method.

imports/plugins/core/ui-tagnav/client/containers/tagNavContainer.js
impactmass: But those values get modified at some point, which is why they were added to state. Or I'm missing your point?

imports/plugins/core/ui-tagnav/client/containers/tagNavContainer.js
impactmass: Ok. I'll see to resolving this. This was an existing template helper function moved into here.

imports/plugins/core/ui/client/components/tags/tags.jsx
impactmass: Not quite.. because TagItem needs the things in `tagProps` to be top-level.

imports/plugins/core/ui-tagnav/client/components/tagNav.js
impactmass: ok. got it

impactmass:
@zenweasel you can also review this. (I've worked on most of the comments from @joykare. I need to checkin with her about the two remaining)

zenweasel:
The styling and icons don't match (`development` on right)
shop_and_reaction

imports/plugins/core/ui-tagnav/client/containers/tagNavContainer.js
```diff
@@ -0,0 +1,389 @@
...
+        return true;
+      }
+    }
+    return false;
+  }
+};
+
+let editable = false;
+let selectedTag = null;
+
```
joykare: These two can be instantiated in `this.state = {}`

imports/plugins/core/ui-tagnav/client/containers/tagNavContainer.js
```diff
@@ -0,0 +1,389 @@
...
+        [NavbarStates.Orientation]: NavbarOrientation.Horizontal,
+        [NavbarStates.Position]: NavbarPosition.Static,
+        [NavbarStates.Anchor]: NavbarAnchor.None,
+        [NavbarStates.Visible]: false
+      });
+    }
+  }
+
+  handleNewTagSave = (tag, parentTag) => {
+    editable = true; // keep editing state after re-render
```
joykare: You shouldn't edit `state` directly.
https://facebook.github.io/react/docs/state-and-lifecycle.html#do-not-modify-state-directly

`this.setState()` causes a re-render every time it is encountered.

impactmass: That's not what's happening. See comment above

imports/plugins/core/ui-tagnav/client/containers/tagNavContainer.js
impactmass: I used these two vars to keep track of values that's needed when the tagnav is re-rendered. E.g when you are editing a taglist and you add a new tag, the list is re-rendered, but a that point, we want to keep it at the editable state. THere needed to be a way to capture that the list was being edited

impactmass: Wrong. The behavior I described above seen to have gone with some other refactoring... so I'll update

PR #2255, opened 2017-05-11
Fix #2243 paypal express

### LINKED ISSUE #2243: [development] order/completed displays login to authenticated user
Check out by registering a user and using PayPal Express as the payment method. `order/completed` displays a login page, instead of displaying the expected order.
reaction-paypal_paypaldone_and_reaction

spencern:
Fixes #2243 by adding paypal roles that are required for accessing the redirect templates within the Paypal Express payment package.

Video:
http://recordit.co/hGX73J8t2J

### DIFF
--- a/imports/plugins/included/payments-paypal/server/startup/startup.js
+++ b/imports/plugins/included/payments-paypal/server/startup/startup.js
@@ -1,0 +1,9 @@
+import { Reaction, Hooks } from "/server/api";
+
+Hooks.Events.add("afterCoreInit", () => {
+  Reaction.addRolesToDefaultRoleSet({
+    allShops: true,
+    roleSets: ["defaultRoles", "defaultVisitorRole"],
+    roles: ["reaction-paypal/paypalDone", "reaction-paypal/paypalCancel"]
+  });
+});
--- a/imports/plugins/included/payments-paypal/server/index.js
+++ b/imports/plugins/included/payments-paypal/server/index.js
@@ -2,3 +2,4 @@ import "./i18n";
 import "./methods/express";
 import "./methods/payflow";
 import "./security/paypal";
+import "./startup/startup";


imports/plugins/included/payments-paypal/server/startup/startup.js
```diff
@@ -0,0 +1,9 @@
+import { Reaction, Hooks } from "/server/api";
+
+Hooks.Events.add("afterCoreInit", () => {
```
zenweasel: You just wanted to try out your fancy new method

imports/plugins/included/payments-paypal/server/startup/startup.js
```diff
@@ -0,0 +1,9 @@
...
+  Reaction.addRolesToDefaultRoleSet({
```
zenweasel: Shouldn't this permission be defined on the routes that are defined in the `register.js`?

imports/plugins/included/payments-paypal/server/startup/startup.js
spencern: ha, well, it seemed like the perfect opportunity. If we can set the `audience` for it, that is probably a better solution

spencern:
@zenweasel 
I attempted to set the route permissions with the audience keyword in the register.js for the `reaction-paypal` plugin. Doing it that way resulted in the same issues where the redirect stops because it's asking a user to login.
```
 {
      route: "/paypal/done",
      template: "paypalDone",
      workflow: "coreWorkflow",
      audience: ["guest", "anonymous"]
    }, {
      route: "/paypal/cancel",
      template: "paypalCancel",
      workflow: "coreWorkflow",
      audience: ["guest", "anonymous"]
    }
```

zenweasel:
Tested. Verified fixed.

Issue #2225, opened 2017-05-09
Receive server error when Label information isn't entered


zenweasel:
### Expected behavior
Some validation should be performed if you don't enter fields that are required to create the record

### Actual Behavior
Receive a server error and "There are no changes to publish".

### Steps to Reproduce the Behavior
1. Create product
1. Bypass "Variant Details" and create options.
1. Publish
1. Observe server error

```
Exception while invoking method 'revisions/publish' Error: Label is required
    at getErrorObject (packages/aldeed_collection2-core.js:480:15)
    at [object Object].doValidate (packages/aldeed_collection2-core.js:462:13)
    at [object Object].Mongo.Collection.(anonymous function) (packages/aldeed_collection2-core.js:214:25)
    at [object Object].Mongo.Collection.(anonymous function) [as update] (packages/dispatch_run-as-user.js:325:19)
    at [object Object].Meteor.methods.revisions/publish (imports/plugins/core/revisions/server/methods.js:142:38)
    at packages/check.js:130:16
    at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
    at Object.exports.Match._failIfArgumentsAreNotAllChecked (packages/check.js:129:41)
    at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1734:18)
    at packages/ddp-server/livedata_server.js:719:19
    at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
    at packages/ddp-server/livedata_server.js:717:40
    at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
    at packages/ddp-server/livedata_server.js:715:46
    at [object Object]._.extend.protocol_handlers.method (packages/ddp-server/livedata_server.js:689:23)
    at packages/ddp-server/livedata_server.js:559:43
```

### Versions
```
Node: 7.10.0
NPM: 4.2.0
Meteor Node: 4.8.2
Meteor NPM: 4.5.0
Reaction CLI: 0.8.1
Reaction: 1.1.1
Reaction branch: development
Docker: 17.03.1-ce
```

spencern:
I'm seeing this error as well. It happens when you don't fill in the title on the product, the label on a variant, or the label on a option. 

Shows the "No changes to publish" alert to the user, even though there are icons indicating changes need to be published.

Same error as @zenweasel on the server

[...]

Issue #2274, opened 2017-05-16
Fix validation on PDP


zenweasel:
All fields that are required or have a set type need to provide some user feedback.

zenweasel:
This should be part of the style guide, but for now we need a way to consistent display errors that require action.

zenweasel:
Right now everything is just being thrown up in a toast. This is actually a regression for some of these fields like price and dimension where we would see the field highlighted, etc.
![dumb_product](https://cloud.githubusercontent.com/assets/72819/26132329/231b1480-3ad2-11e7-81cb-1ee174354192.png)

zenweasel:
I should be able to enter $0.00 as a valid price on a product. Sometimes things are free. Right now I get a "Price is required" toast.

zenweasel:
We don't need to wait for design here, I think we all know the basics of how validation should work. The field should be focused, the error should show next to the field, etc.

rymorgan:
@zenweasel totally agree. If you guys take a stab at it and send it my way when it's done. But yes it should just as you describe.

zenweasel:
I just talked to @kieckhafer and he is willing to take this on. Hopefully we can expand the Autoform replacement work that @mikemurray has been working on

rymorgan:
@kieckhafer let me know when you work on this and we can talk through how the validation should be styled. I'm hoping this is something that does some automatic validation magic with our forms so we're consistent everywhere.

aaronjudd:
Removed "critical impact" with the merge of #2531

mikemurray:
This PR (https://github.com/reactioncommerce/reaction/pull/2893) should be able to close this issue once it's merged

mikemurray:
Also assigned @kieckhafer since he's fixing this in his PR #2893

kieckhafer:
Closed by #2893

PR #2276, opened 2017-05-16
Add validation for required fields on PDP

michael-alade:
*Resolves*
#2225 

*Testing*
- Login as admin
- Create a product
- Add a variant option.
- Fill all fields apart from product title, variant details label and variant options label.
- Click publish
- A toast message should show what fields are required.
- Now fill in all field required, then click publish
- The changes should be successfully published.


zenweasel:
Well, this is not great. You still get "Label is required" whether you didn't enter an option of a variant title and the way "Untitled Option" shows up there is also not great. At it looks like if there are multiple errors it only shows one.

But we we probably need to always bubble up these errors from SimpleSchema so I guess this is an improvement. But we really need to do #2276

rymorgan:
Jumping in on this too. I think there needs to be some indication of what fields are required. See example. 

validation

zenweasel:
@rymorgan Yeah, so we are waiting for design on #2274 and we can revisit this. Should I assign that to you? Not really sure what the best way to do that is.

rymorgan:
Yea, add it to the design project, which I think you have or I did. I'll pull it into view so it doesn't get lost.

PR #2327, opened 2017-05-22
Better handling for unique routes for React Router

mikemurray:
The old logic for handling route uniqueness was not sufficient in preventing duplicate routes in our new React Router world.

This PR attempts to correct this by handling route uniqueness by route path, and only after all routes have been had an opportunity to be processed.

**To Test:**

1. Add a duplicate route to a custom plugin register.js:

```

Reaction.registerPackage({
  label: "Test",
  name: "test-plugin",
  icon: "fa fa-cubes",
  autoEnable: true,
  registry: [{
    route: "/product/:handle/:variantId?",
    name: "product",
    template: "test",
    workflow: "coreProductWorkflow"
  }]
});

```

2. Put this in any `.html` file in a client folder outside the imports directory" For example "/client/templates"

```
<template name="test">
  HELLO
</template>
```

3. Observe that your custom template overwrites the default PDP route.


imports/plugins/core/router/lib/router.js
```diff
@@ -500,24 +489,33 @@ Router.initPackageRoutes = (options) => {
...
-          Router.routes.push(route);
+          routeDefinitions.push(route);
         }
       }
     } // end package loop
 
+    // Uniq-ify routes
+    // Take all route definitions in the order that were received, and reverse it.
+    // Routes defined later, like in the case of custom routes will be then have a
+    // higher precidence. Any duplicates after the first instance will be removed.
```
zenweasel: precedence

imports/plugins/core/router/lib/router.js
```diff
@@ -500,24 +489,33 @@ Router.initPackageRoutes = (options) => {
...
-
-          Router.routes.push(route);
+          routeDefinitions.push(route);
         }
       }
     } // end package loop
 
+    // Uniq-ify routes
+    // Take all route definitions in the order that were received, and reverse it.
+    // Routes defined later, like in the case of custom routes will be then have a
```
zenweasel: "will then"

Issue #2177, opened 2017-04-28
Docker build: FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory


owenhoskins:
I am curious if anyone else has been running into this while running building a custom Reaction docker image with `docker build`. 

I am unable to reproduce this issue with a fresh reaction install. However, my custom app which includes an additional dozen or so npm packages and lots of additional code in `/imports/plugins/custom/` no longer succeeds when running `docker build`. It does however successfully generate a tar ball when running `meteor build`. 

I've been following this related [issue](https://github.com/meteor/meteor/issues/8157) and the suggestion to remove `standard-minifiers-js` or use `abernix:standard-minifier-js@2.0.4` does not have an effect. In fact, the error message varies greatly in the stack trace as to where it hit memory issues. 

Here is a sample of the error:

```
<--- Last few GCs --->

  347149 ms: Mark-sweep 1387.2 (1457.1) -> 1387.8 (1457.1) MB, 1402.8 / 0 ms [allocation failure] [GC in old space requested].
  348463 ms: Mark-sweep 1387.8 (1457.1) -> 1391.0 (1457.1) MB, 1515.8 / 0 ms [allocation failure] [GC in old space requested].
  350069 ms: Mark-sweep 1391.0 (1457.1) -> 1382.3 (1457.1) MB, 1605.7 / 0 ms [last resort gc].
  351539 ms: Mark-sweep 1382.3 (1457.1) -> 1386.8 (1457.1) MB, 1470.5 / 0 ms [last resort gc].


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x1867241b4629 <JS Object>
    1: def_function [0x1867241041b9 <undefined>:~3517] [pc=0x2972eb9453c5] (this=0x18eea5efc0a1 <an AST_Function with map 0x120175a62fb1>,symbol=0x8848be30a79 <an AST_SymbolDefun with map 0x3c70d8c56851>)
    2: visit [0x1867241041b9 <undefined>:~3317] [pc=0x2972eb938507] (this=0x2e939bded2e9 <a TreeWalker with map 0x3c70d8c0b061>,node=0x8848be30a79 <an AST_SymbolDefun with map 0x3c70d8c56851>,...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory
[...]

PR #2334, opened 2017-05-23
CircleCI 2.0 and docker base update

jshimko:
### CircleCI

This converts the CI build/test/deploy to the new CircleCI 2.0 config format.

Config reference:
https://circleci.com/docs/2.0/configuration-reference/

Latest builds of this branch:
https://circleci.com/gh/reactioncommerce/reaction/tree/circleci-2.0-and-docker

I also added a Docker test build for commits to any branch with "docker" in the name (great for testing updates to the Docker build before merging a PR into development)

Aside from the more flexible config, the single biggest gain here was the build time.  This was largely achieved by caching as many of the slowest steps as possible.  As a result, the average build times were dropped significantly.

Running all Reaction tests, but no Docker build (most branches):
- Before:  9-10mins
- After: 4-5mins

Running all Reaction tests, build/test/deploy a fresh Docker image:
- Before:  23-25mins
- After: 16-18mins

Sweet, right?!

### Docker

I also updated the Docker base image to v1.4.0.  This adds the ability to set Node flags that the Meteor build tool will use at install/build time. 

Example usage:

```sh
docker build --build-arg TOOL_NODE_FLAGS="--max-old-space-size=2048" -t reactioncommerce/reaction:latest .
```

That was added to solve a problem some users were experiencing with memory consumption during the build.  See #2177.


.circleci/config.yml
```diff
@@ -0,0 +1,61 @@
...
+          key: node_modules
+          paths:
+            - /home/reaction/node_modules
+
+      # run tests
+      - restore_cache:
+          name: Restoring Meteor dev_bundle cache
+          key: dev_bundle
+
+      - run: reaction test
```
zenweasel: Just a question, but does this new format only allow one line for the executable, as opposed to the old format where you could any number of executables?

aaronjudd: This? https://circleci.com/docs/2.0/defining-multiple-jobs/

aaronjudd:
Shouldn't we have seen an image pushed for  https://circleci.com/gh/reactioncommerce/reaction/5723 ?

deploy.sh returned `Not running a deployment branch.`

PR #2371, opened 2017-05-31
[WIP] Meteor 1.5

aaronjudd:
- include Buffer polyfill
- replaces abernix:standard-minifier-js with standard-minifier-js
- properly pins job-collections to 1.4.0


aaronjudd:
@mikemurray `Warning: ReactClass: You're attempting to include a mixin that is either null or not an object.` this has the mixin errors on the Griddle. Can you take a look at those?

hrath2015:
Flat Rate Shipping Options are not visible... 
image

Checked in DB and added 1  option.  Could be due to react mixing error on console.

hrath2015:
it is because of  in latest react-meteor-data@0.2.11 **`ReactMeteorData`** is not exported anymore. Hence it is breaking in 2 places.  `LogGriddle` and `MeteorGriddle`.  there is issue https://github.com/meteor/react-packages/issues/223 and PR https://github.com/meteor/react-packages/pull/224 submitted.  Till then it can be pinned to 0.2.9

zenweasel:
@aaronjudd Do we want to add the eslint change here where it will catch variables that aren't declared?

aaronjudd:
WIP with no assigned or approved reviewers generally means... not ready for development.

zenweasel:
Do we want to revert this merge? Seems like it was premature?

mikemurray:
:shipit:

spencern:
image

🤔

jshimko:
For what it's worth, I tested this branch a bunch and hadn’t had any issues.  And I brought it up in Slack and the meeting today and nobody objected to the idea.  Until that last devel merge, it passed CircleCI tests 7 times in a row with no failures (and the only failure was a false positive that just passed after clicking rebuild)
https://circleci.com/gh/reactioncommerce/reaction/tree/Meteor-1.5
So I'm happy to revert it if that's the preference.  And in the future, I'll wait for other reviewers.

aaronjudd:
I think it's good. I also didn't have any issue with this branch but had wanted to see if we could also get the updated packages merged in and tested. however, now that it's merged.. I guess -let's let it go, unless anyone else objects.  My main goal was to get the dev dependencies that @spencern mentioned resolved, but we can merge that in later.

Issue #1983, opened 2017-03-14
Invite user doesn't allow inviting users with the same name as an existing user


zenweasel:
### Expected behavior
You should be able to invite anybody to your shop, even if there is someone with the same name already a member.

### Actual Behavior
If you already have a "Brent Hoover" and try to invite another person named "Brent Hoover" it fails, even when they have different email addresses.

### Steps to Reproduce the Behavior
1. Create a user with the name "Test User"
1. Go to Account panel and try to invite another user named "Test User"
1. Observe that you will get the "UserName already exists" error even when the emails are unique

Seen in #1974

zenweasel:
Closed via #2386

PR #2386, opened 2017-06-02
Admin should be able to invite a user with the same name

michael-alade:
**Resolves**
#1983 

**Test**
- Login as admin
- Go to the accounts on action menu
- Invite a user by adding the person's name and email address
- If successful, invite another user with the same name but different email address
- You should be able to see the user below where other users are displayed.


jshimko:
Since we can't override the fact that Meteor enforces unique usernames, would it possibly be more useful to just create a new field on the user document called `name`? (or whatever).  Having a random number in the username renders it pretty much useless to the user, no?  Nobody is going to remember their username and having names with random numbers in the UI doesn't sound great either. 

Meteor will let you login with that username, so it might be nicer to let users set usernames themselves when they register and update the login forms to allow them to use either email or username to log in.  I personally prefer to login with a username whenever possible because it can be shorter than my email and faster to type.  But I also like having the option to log in with my email because it's easy to forget what username you used.

Any UX thoughts on this @rymorgan?

michael-alade:
@jshimko Is it a viable alternative to allow the invited user reset the username to his preferred username when the enroll email link is sent? There is a password form that pops up for the user to add his/her password

jshimko:
Yep, agreed.  @michael-alade As Ryan said, I'm thinking you should just change that to a `name` field on the user object and then update the few places that name gets used in the UI (not many).  Then we can open another issue to discuss allowing usernames to be used the way Meteor intended them to be (as a method of login and an optional unique user identifier).

zenweasel:
We discussed this again and I think we all agreed that @jshimko's direction made the most sense, so I think that's the solution we want to pursue.

zenweasel:
@michael-alade What's the status on this?

michael-alade:
This is ready to be reviewed.

zenweasel:
If I try to invite a user with an email that I have already invited I get the following server error:
```
Exception while invoking method 'accounts/inviteShopMember' ReferenceError: subject is not defined
    at [object Object]._inviteShopMember (server/methods/accounts/accounts.js:554:7)
    at packages/check.js:129:16
    at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
    at Object.exports.Match._failIfArgumentsAreNotAllChecked (packages/check.js:128:41)
    at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1734:18)
    at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
    at packages/ddp-server/livedata_server.js:717:40
    at [object Object]._.extend.withValue (packages/meteor.js:1122:17)
```

zenweasel:
When I click on invite link I get this broken page (in addition to what you can see the Update and Cancel buttons didn't display until I moused over them)

reaction

zenweasel:
Like this:

reaction

zenweasel:
Both of the above issues happen in `development` so they aren't related to this ticket. Tested and the functionality work as described.

Issue #2297, opened 2017-05-17
Setting route permissions via the registry doesn't work


zenweasel:
### Expected behavior
You should be able to specify permissions/audience in a route in `register.js` and have it take affect.

### Actual Behavior
It doesn't work, you need to programmatically add these.

### Steps to Reproduce the Behavior
1. Create a plugin with a route (you could use the `reaction-example-plugin` as a starting point).
1. Add permissions to the route according to the documentation
1. Observe you still get an error

michael-alade:
This was caused due to the change from Flow Router to React Router.

aaronjudd:
Closed in #2388

PR #2388, opened 2017-06-02
Use route permissions on the registry

michael-alade:
**Resolves**
#2297 

**Testing**
- Create a plugin according to the documentation
- Add permissions in the plugins `register.js` according to the documentation
- Try to access the defined route for your plugin. If the user is meant to have the permission to view the plugin then it should display. 

**Read**
Look into this documentation on creating routes for plugins
https://docs.reactioncommerce.com/reaction-docs/master/plugin-routes-6


mikemurray:
Can checkout and everything looks good.

One issue other testers may see, but is not related to this PR is: https://github.com/reactioncommerce/reaction/issues/2399

impactmass:
@michael-alade I dropped `reaction-example-plugin` on to this branch. I modified the `about` registry entry to have permission like so:
```
  registry: [
    {
      route: "/about",
      name: "about",
      template: "aboutUs",
      workflow: "coreWorkflow",
      permissions: [{
        label: "About-perm",
        permission: "about-perm"
      }]
    }
  ],
```
Then, `reaction reset`. Then loaded up the page, and added the permission to the anonymous user data.
With the user having that permission, I expect the /about to load, but got redirected to /not-found.

I'm not sure if this is an issue with the example plugin or the pr.. so I'll probably do another test shortly.

PR #2394, opened 2017-06-02
Marketplace Core Fixes

spencern:
Updated PR in place of #2172 
- Contains #2172 
- Fixes Broken Tests
- Merges updated marketplace branch

### DIFF
--- a/server/publications/collections/product.js
+++ b/server/publications/collections/product.js
@@ -19,31 +19,23 @@ export function findProductMedia(publicationInstance, productIds) {
     selector["metadata.productId"] = productIds;
   }
 
-  // TODO: This block is commented out in Marketplace, why??
-  // if (shopId) {
-  //   selector["metadata.shopId"] = shopId;
-  // }
+  // The default is to see only published images of products
+  selector["metadata.workflow"] = { $in: [null, "published"] };
 
-  // No one needs to see archived images on products
-  selector["metadata.workflow"] = {
-    $nin: ["archived"]
-  };
-
-  // get seller-shop id if user is a seller;
-  const sellerShopId = Reaction.getSellerShopId(publicationInstance.userId, true);
-
-  // sellers can see unpublished images only of their shop
-  // Users with create-product access in the Parent shop can see all of them
-  if (sellerShopId) {
-    selector.$or = [
-      {
-        "metadata.workflow.$in": [null, "published"]
+  const isUserOwnerOrModerator = Reaction.hasPermission(["owner", "moderator"], publicationInstance.userId);
+  if (isUserOwnerOrModerator) {
+    selector["metadata.workflow"] = { $nin: ["archived"] };
+  } else {
+    // get seller-shop id if user is a seller;
+    const sellerShopId = Reaction.getSellerShopId(publicationInstance.userId, true);
+    // sellers can see unpublished images only of their shop
+    if (sellerShopId) {
+      selector.$or = [{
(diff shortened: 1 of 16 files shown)


spencern:
@lcampanis and/or @prosf 

Could either of you review this PR (specifically the last commit I've made) and ensure that it doesn't break anything you've changed intentionally.

server/publications/collections/product.js
```diff
@@ -53,42 +54,56 @@ export function findProductMedia(publicationInstance, productIds) {
...
     isDeleted: { $in: [null, false] },
     isVisible: true
   };
-  let productShopId;
-  const shop = Reaction.getCurrentShop();
-  // verify that parent shop is ready
-  if (typeof shop !== "object") {
-    return this.ready();
+
+  // Permits admins to view both visible and invisible products.
```
spencern: @prosf @lcampanis 
I added this section back into the code. I'm still unclear as to why it was removed, but tests fail without it. Specifically the tests that ensure that an owner/admin/createProduct user can view hidden products.

server/publications/collections/product.js
```diff
@@ -53,42 +54,56 @@ export function findProductMedia(publicationInstance, productIds) {
...
+
     if (product) {
+      // Take productShopId in order to check if user can edit this product or view its revisions
       productShopId = product.shopId;
     } else {
       return this.ready();
     }
   } else {
-    const newSelector = {
+    const newSelector = Object.assign({}, selector, {
```
spencern: This is the other change that fixes a broken test. Without this section, the test ensuring that invisible products are not available to non-admins is broken.

I can see how this seems unnecessary with the section above that permits admins to view invisible products, so I'm curious what the bigger picture is here if this was removed on purpose.

aaronjudd:
I was unable to view any of the product settings (for instance, the template selector is empty). Other parts of the PDP appear to be broken as well (social messaging).  @spencern did you see these?

spencern:
I am seeing a number of console errors, but I can see the template selector and product settings.

image

If I click into the product with the action view open, then it's empty, but after clicking the "edit" icon on a variant, it populates.

A lot of console errors though, and many things don't work.

spencern:
That's as the marketplace owner.

As a seller shop, I don't have any of the navigation
image

spencern:
Product Social Media editing does appear to be broken as well.

spencern:
Additionally,

Just ran these checks on the `marketplace` branch as well and it appears that they are issues there as well, so not confined to this PR. 

My guess is something in marketplace is not playing well with the React conversion.

server/publications/collections/product.js
lcampanis: Hi @spencern, there's already a role checking on line 132
```js 
if (Reaction.hasPermission("createProduct", this.userId, productShopId)) {
```
There's also no need for [`owner` or `admin`], as `createProduct` perm is strong enough to be together with any of those users' roles...
Leaving visible undeclared will just look to return `any` visibility anyway.

So to keep the same pattern you see in other publications, where the default is always last, we have the following in place which might be tiny bit neater.. 
Move it instead to line 242
```js
  // Everyone else gets the standard, visible products and variants
  selector = Object.assign(selector, {
    isVisible: true,
    $or: [
      { type: "simple" },
      { type: "variant" }
    ],
    isDeleted: { $in: [null, false] }
  });
```
Hope I'm not forgetting something, but I think that's it.

PR #2395, opened 2017-06-03
[WIP] Updated dependencies

aaronjudd:
- preserved nodemailer, griddle, react libs
- update all other package


zenweasel:
So do we need to correct all the `React.proptype` references?

PR #2405, opened 2017-06-07
[WIP] React Permission

mikemurray:
Adds components to help with defining permissions for React components.

Examples.

```
import { Permission, withPermissions } from "@reactioncommerce/reaction-permissions";

<Permission roles={["admin"]}>
   <span>{"Admin Only"}</span>
</Permission>

const permissibleComponent = withPermissions(["admin"])(MyComponent);
```


mikemurray:
@spencern @impactmass This is a simple approach to adding permissions to components. Nothing in-app depends on this yet so feel free to modify.

spencern:
@mikemurray thanks, this looks like a good direction.

aaronjudd:
@spencern is that an approved review?

imports/plugins/core/permissions/lib/permission.js
```diff
@@ -0,0 +1,106 @@
...
+          return true;
+        }
+      }
+    }
+  }
+  // no specific permissions found returning false
+  return false;
+}
+
+export function getShopId() {
```
spencern: This getShopId() function won't work with the updated marketplace style of getting an active shopId

imports/plugins/core/permissions/lib/permission.js
```diff
@@ -0,0 +1,106 @@
...
+  // return if user has permissions in the group
+  if (Roles.userIsInRole(userId, permissions, group)) {
+    return true;
+  }
+
+  // global roles check
+  const sellerShopPermissions = Roles.getGroupsForUser(userId, "admin");
+
+  // we're looking for seller permissions.
+  if (sellerShopPermissions) {
```
spencern: Would be nice to have a more explicit comment here explaining what this section does. Seems like it has some/a lot of redundant code.

Per the [getGroupsForUser](https://github.com/alanning/meteor-roles/blob/master/roles/roles_common.js#L488-L530) method - this should return an array of groups (shopIds in our case).

With that being the case, I'm unclear as to why we're using this method of looping and checking. It seems like something as simple as the following would be more readable and accomplish the same thing.
```
return sellerShopPermissions.some((shopId) => Roles.userIsInRole(userId, permissions, shopId));
```

spencern:
@aaronjudd It's not. I do think the permissions component is the right direction, but after looking through the code, there's a few things that should be cleaned up.

imports/plugins/core/permissions/lib/permission.js
mikemurray: then should that be fixed in the marketplace branch?

mikemurray:
@spencern since nothing is using this component, can any updates you need be done in the `marketplace` branch or otherwise?

imports/plugins/core/permissions/lib/permission.js
spencern: We could wait until it's merged to fix it. Might be easier to just fix it now.

spencern:
@mikemurray I think we can probably do updates in the marketplace branch, but I'd prefer if it's marketplace ready when we merge so we don't need a separate task for it.

I can probably update that getShopID method for you

spencern:
@mikemurray I updated the getShopId method here. But have one question. I noticed you were exporting it from this file - was it being used anywhere else? I couldn't find any place using it, but want to make sure.

spencern:
We should probably revisit this or close it if it's not needed any more

PR #2406, opened 2017-06-07
[WIP] React SSR

mikemurray:
- Added packages needed for SSR support
- Added super quick demo `App` component to test SSR


janus-reith:
What is the current road here? Are you still working on this?

zenweasel:
Quote from @aaronjudd re SSR:
> I think we should take a slightly more complicated path towards SSR. we've been discussing a direction where we get the reaction-api-base graphql services implemented in client react components as an optional data source, and start creating a lightweight client layer, where all the client consumer UI (checkout, profile, cart, pdp, grids) can be implemented independent of the full meteor package . This would allow admin and services to be deployed separate from the current meteor build, in theory that more cleanly separated client UI should become a lot simpler to implement SSR on (as we really don't need SSR on admin functionality).

janus-reith:
Well what makes me wonder: Wouldn't this render the current Meteor client implementation obsolete when its finished? At least I don't really understand how both would play together in such a scenario.

janus-reith:
Or would the client part still be delivered out by the Meteor Server, and just use graphql to feed data instead of the Meteor ddp connected Mongo collections?

sneerin:
So as far as I understand SSR is still not ready? prerender.io _escaped_fragment_ marked as deprecated by google, so this is e-commerce solution without no clear view on SEO and how to implement it?

andru:
@sneerin fwiw, [prerender.io now user-agent sniffs for Google](https://prerender.io/documentation/google-support) instead of using _escaped_fragment_

PR #2418, opened 2017-06-09
Add closeOnClick option to DropDownMenu component

spencern:
Adds `closeOnClick` option to DropDownMenu component.
Adds getter and setter to popover `isOpen` to determine isOpen from either props or state.
Adds `isControlled` to check if isOpen prop is being passed in.


zenweasel:
Looks like there's a React lint issue to resolve

spencern:
Nice catch @zenweasel / bithound. Just pushed up a fix.

PR #2425, opened 2017-06-12
Bunyan and Loggly updates

jshimko:
- Bunyan 2.0.0
- [`node-loggly`](https://github.com/winstonjs/node-loggly) has been abandoned and [Loggly finally created an official fork](https://github.com/winstonjs/node-loggly/issues/68#issuecomment-307174847).  I removed [`bunyan-loggy`](https://github.com/MauriceButler/bunyan-loggly) (which depended on `node-loggly`) and I rewrote a new connector to pipe logs to Loggly using the new official lib.
- added a new Loggly environment variable (`LOGGLY_BUFFER_LENGTH`) which sets the API batching amount.  The default value is 1, so every logger call gets sent to Loggly immediately.  So to make it wait and only send logs on every 10th logger call, you can now do this...

```sh
export LOGGLY_BUFFER_LENGTH=10

reaction
```

I have tested everything with Loggly credentials and confirmed all expected data makes it there in the correct format.  No breaking changes should be expected with the existing `Logger` API in Reaction.  

One more constantly outdated dependency chain gone!


zenweasel:
@jshimko How would I go about testing this? Also looks like you have a conflict to resolve?

server/api/logger/main.js
```diff
@@ -1,26 +1,23 @@
...
+const outputMode = "short";
 
 level = level.toUpperCase();
 
-if (!includes(levels, level)) {
+if (!_.includes(levels, level)) {
   level = "INFO";
 }
 
-if (level === "TRACE") {
```
aaronjudd: Are we not going to have a "trace" dump then, or was this just an unlikely use case?

jshimko: Well, my thinking here was that the default stdout output stream (the console) is intended entirely for human consumption, so readability should be the main focus of the stdout configuration.  Also, there isn't really anything that an unformatted output gives you that's unique or useful.  The only things the formatter strips out is the hostname and pid (process ID).  Any other objects (errors, data, etc.) passed to the logger calls for debugging are still included in the formatted log output.  The main advantage being that they're much easier to read when formatted like that.  So in short, all of the same data gets shown, it's just easier to parse visually when it's run through the formatter.

However, in case someone _really_ wants to have that raw JSON output, I added a new env var (`REACTION_LOG_FORMAT`) which you can use to set the formatter.  So...

```
REACTION_LOG_FORMAT=json reaction
```
outputs this...

1____code_meteor_reaction_reaction__node__and_main_js_ ___code_meteor_reaction_reaction

And technically you could set that to anything [`bunyan-format`](https://www.npmjs.com/package/bunyan-format) supports.  Which is...

`short|long|simple|json|bunyan` (default: short)

Sort of relevant... there's not a single instance of `Logger.trace()` in the whole app, so this shouldn't have any effect on anyone's current experience.

jshimko:
@zenweasel The Loggly connector setup is documented in our logger docs (which I totally forgot).  It's just a few environment variables.

```sh
# required
LOGGLY_SUBDOMAIN=""
LOGGLY_TOKEN=""

# optional
LOGGLY_LOG_LEVEL=""
```
(scroll to bottom)
https://docs.reactioncommerce.com/reaction-docs/master/logging

Issue #2212, opened 2017-05-04
[Marketplace] Packages are not added for new shop until server restart


spencern:
### Expected behavior
When a new shop is created, the associated packages that shop requires should be created and inserted into the Packages collection of the database.

### Actual Behavior
When a new shop is created, Packages are inserted into the database until the server is restarted. Packages are inserted during the Reaction startup sequence.

### Steps to Reproduce the Behavior
1. Checkout the `marketplace` branch
2. Open up a `meteor mongo` shell
    - Run `db.Shops.find({}, {_id: 1})` and select one of the shopIds
    - Run `db.Packages.find({shopId: "copyPasteShopId"})
  - Observe that packages are present for existing shops
3. Register a new user.
4. Navigate to the new user's profile and click the "Become A Seller" button
5. Go to your `meteor mongo` shell
    - Run `db.Shops.find({}, {_id: 1})` and select the newly created shop
    - Run `db.Packages.find({shopId: "copyPasteShopId"})
    - Observe that packages do not exist for newly created shop.
6. Restart your server, wait until it is completely restarted.
7. Go to your `meteor mongo` shell
    - Run `db.Packages.find({shopId: "sameShopAsStep5"})
    - Observe that packages now exist for this shop.

### Versions
```
$ reaction -v

Node: 7.8.0
NPM: 4.5.0
Docker: 17.03.1-ce
Reaction CLI: 0.8.0
Reaction: 1.1.1
Reaction Branch: marketplace
```

@impactmass confirmed this bug on his end as well and created an example video https://www.dropbox.com/s/1tw9eibedtld1hn/mkpl.mov?dl=0

There is more discussion and details on this issue in #2145

spencern:
Resolved by #2429

PR #2429, opened 2017-06-13
[Marketplace] Feature - Dashboards for multiple shops

spencern:
Change the context of a users dashboard shortcuts and user dropdown shortcuts based on active shop.

For a user with the following roles for the secondary shop: 
```
"FXybdgAYSKYndkRef" : [ 
  "notifications", 
  "dashboard", 
  "seller", 
  "guest", 
  "manage-users", 
  "orders", 
  "account/profile", 
  "accounts", 
  "account/verify", 
  "dashboard/accounts", 
  "product", 
  "createProduct", 
  "tag", 
  "index", 
  "cart/checkout", 
  "cart/completed", 
  "shop"
]
```
Their dashboard and shortcuts dropdown looks like the following:
image


spencern:
@impactmass 
Assigning you to this PR since you're working on permissions also. I think it'd be good if you are I are reviewing all permissions related PRs

impactmass:
These are the items I see in the dropdown with the permissions set as listed in above description (shown on the side). Taken while on the secondary shop too.

screenshot 2017-06-15 10 41 22

I added the permissions directly on the user in the DB as going through the Accounts dashboard wasn't adding the permissions under the secondary shop. Is this the right way to test the PR?

spencern:
@impactmass While we don't have identical permissions available, I think we probably went about setting it up slightly differently. I don't think that's an issue as what this PR is trying to accomplish is changing the dashboard contextually based on the active shop and enabled permissions. It looks like that is working for you as well.

If you can do a code review for me as well, I think we can probably go ahead and merge this one in.

We can figure out why the same set of permissions are granting different options for the dashboard later, but my guess would be something that changed in the base `marketplace` branch that you merged in.

impactmass:
ok, I'll do a code review, plus test it while stepping back that last commit. Will update in a shortly

server/publications/collections/packages.js
```diff
@@ -69,10 +69,13 @@ function transform(doc, userId) {
 //
 //  Packages Publication
 //
-Meteor.publish("Packages", function (shopCursor) {
-  check(shopCursor, Match.Optional(Object));
+Meteor.publish("Packages", function (shopId) {
```
impactmass: Any chance of this affecting any current place "Packages" is being called as previously written (i.e with shopCursor)... now that we're changing it to shopId string?

spencern: Good question and nice find.

I should have mentioned that I checked on that, I looked at every instance where we subscribe to the Packages publication and not a single one used the shopCursor. I'm guessing it's an undeleted artifact of the way things used to be done, but not sure.

@aaronjudd any thoughts here?

impactmass: From my checks as well, I don't see any references that can break (so far). All the "subscribes" seem to already be set to use shopId. So, I think this is good... any objections?  cc @aaronjudd

spencern:
This PR got out of hand, and is a lot bigger than I'd like, but the issues kept cascading.

This solves a number of issues now: 
* Changes the context of the dashboard based on active shop
* When creating a shop, automatically add packages, setup permissions, and create layouts so a server restart is not necessary (Resolves #2212)
* Removes the dependency on getSellerShopId through out the app, with the exception of leaving it in the `marketplace` plugin specific templates.
* Modifies the getShopId method on the server to return the users active shop
* Fixes an issue where routes were not loading on a direct route because the packages were not found for the shop.

@impactmass - If you could give this another review asap, we can merge it and get cranking on the rest of marketplace

client/modules/core/helpers/permissions.js
```diff
@@ -59,6 +59,10 @@ Template.registerHelper("hasDashboardAccess", function () {
   return Reaction.hasDashboardAccess();
 });
 
+Template.registerHelper("hasDashboardAccessForAnyShop", function () {
+  return Reaction.hasPermissionsForAnyShop();
```
impactmass: I searched for this function `hasPermissionsForAnyShop` and couldn't find it. Is it already on the marketplace branch? Or there's some typo here? Is it meant to be
```
Template.registerHelper("hasDashboardAccessForAnyShop", function () {
  return Reaction.hasDashboardAccessForAnyShop();
});
```

spencern: Good catch. I think I added this before the navbar was fully converted to react and have since removed the rest of it.

spencern: This is fixed now.

spencern:
Pretty sure this is ready to merge. Waiting on a final 'ok' from @impactmass, which I'll get whenever he gets back. We're planning to walk through the PR at that point, anyone is welcome to join. Will be some time next week. (Monday hopefully). In the mean time if anyone else wants to review, we can go ahead and merge.

PR #2435, opened 2017-06-14
[Marketplace] Feature - getShopId server side

spencern:
Please review and merge #2429 if it has not been merged yet first. Creating separate PRs and branches here even though they build off of each other to keep concerns separated.

This PR accomplishes the following
1. When an administrative user selects a shop, saves the selected shop as a user preference in the database as a user preference using `Reaction.setUserPreference`
2. Adds a server method `Reaction.getUserPreference` that permits us to retrieve user preferences on the server
3. Changes `Reaction.getShopId` on the server to accept a userId param and return the users activeShopId if it's set
4. Changes all Collection permit/deny security code to use the active shop
5. On the client, select the active shop on app load if it's available.

This PR should be reviewed in depth for security vulnerabilities

### DIFF
--- a/client/modules/accounts/templates/dropdown/dropdown.js
+++ b/client/modules/accounts/templates/dropdown/dropdown.js
@@ -76,7 +76,7 @@ Template.loginDropdown.events({
 Template.accountsDropdownApps.helpers({
   reactionAppsOptions() {
     // get shortcuts with audience permissions based on user roles
-    const roles = Roles.getRolesForUser(Meteor.userId(), Reaction.getSellerShopId());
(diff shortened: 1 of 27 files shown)


spencern:
@lcampanis You may want to review this as well as this is the point where we're starting to move away from some of the `getSellerShopId` functionality.

I still like the idea of using the concept of "my shop" or sellerShop for some functionality, but at the core I think we need more flexibility from the `getShopId` method than it was providing. This PR starts giving us the flexibility in the core that we need for marketplace implementations.

spencern:
Closing in favor of #2429 which has these changes and more pulled in

Issue #2184, opened 2017-05-01
[Marketplace] Permissions should be settable by placing a user into a predefined group


spencern:
Permission groups should exist.

Users should be able to be placed in multiple groups. Groups should be able to cascade in some way. Perhaps Permission Groups should be able to inherit permissions from other groups as well.

By placing a user into a permission group, said user should be granted all permissions that are defined in that group for the specific shop context that was granted

By removing a user from a permission group, all permissions defined in that group - that are not available in some other group the user is in - should be revoked from that user.

If a user's permissions are customized by an admin, the user should be removed from any groups associated with a specific set of permissions and re-granted individual permissions that match the customized set of permissions they now have. The user should no longer be associated with a `Permission Group` in the Accounts dashboard but should instead be labeled as having a `Custom Permissions` set in the Accounts dashboard. Multiple users with Custom Permissions may be grouped together, but should not be assumed to have similar permission sets.

#### Possible Default Permission Groups
-   Marketplace Owner
-   Marketplace admin
-   Shop Owner
-   Shop Admin
-   Affiliate Shop Owner
-   Affiliate Shop Admin
-   Customer
-   Guest

spencern:
#372 has some old discussion around groups.

impactmass:
WIP PR: Opening shortly

Todos:
- [ ] Creating a group permission
- [ ] Updating a group permission and updating users belonging to that group
- [ ] Adding a user to a group

impactmass:
Update:
_ Group methods completed
_ Not all the group methods I've written can be tested from the UI yet...  So, still keeping the WIP label on the PR.

_ But feedback can be given on this file: [`server/methods/core/groups.js`](https://github.com/reactioncommerce/reaction/pull/2448/files#diff-f8c0fc2581fcce57e4037e6f5e9df21a). As it contains the main work.
 ... I added comments to explain some parts.

_ Next steps:
Implement design from Ryan (as it relates to groups), and connect them to the methods. At that point, the whole groups PR can be tested.

impactmass:
*Status:
_ I'm working on adding more tests for the group methods
_ and also adding email notification
_ while @rhenshaw56 is working on UI for accounts dashboard (part of which the groups methods will be used with)

impactmass:
I was out for 2 days so there's been no update here. Here's the current status: 
_ I've written tests for group methods (can be seen in `methods/core/groups.app-test.js`
[...]

Issue #2194, opened 2017-05-01
[Marketplace] Marketplace and Shop owners should be able to create custom permission groups


spencern:
Marketplace and shop owners should be able to create additional permission groups with custom names and a custom set of permissions. 

Owners should be able to define granular permissions for these groups.

Once created these permission groups should be available anywhere other default permission groups are available.

rymorgan:
**Links for designs below** 

Design annotations in Zeplin :)

Reaction core team link: 
Web: https://zpl.io/1YphHq
Zeplin App: zpl://project?pid=59487558eca97e6140dd4b5b

Public link: 
https://scene.zeplin.io/project/59487558eca97e6140dd4b5b

rymorgan:
Here are the latest screens for Permissions. After looking through it with @kieckhafer made some small edits. 

1. A different way to change the default group.
2. Changed one icon button from a cog to a pencil b/c it makes more sense.
3. Added a way to delete groups.

Screens: 
Intial User Screen- https://zpl.io/Z162xqi
Edit Permissions Screen - https://zpl.io/nVX8n
Edit Groups - https://zpl.io/Zh8Rdn
Add Group - https://zpl.io/ZTEMLM
Edit Ownder -  https://zpl.io/15nVaK

impactmass:
Thanks @rymorgan. I'll take a look at the designs. If there be questions, I'll ask you/Erik. 

Just a note: I'm using this other ticket #2184 to post updates of the state of the implementation.... there's a number of tickets related to this mockup, but I'm just posting in one ticket ( #2184)

impactmass:
A functional (base) UI of the groups view (creating and editing groups, switching user from one group to another, and adding users to groups) was merged in with this PR #2543.  
Now, admins can create permission groups from the UI (using the already merged in methods #2448) .

Issue #2183, opened 2017-05-01
[EPIC][Marketplace] Marketplace Permissions


spencern:
# Permission model for Marketplace.

## Tasks

-   [x] ~Permissions should be settable by placing a user into a predefined group~ (#2184)

-   [ ] Registry should permit group based permission allowance (#2185)

-   [x] ~Inviting someone to create a new Shop or Affiliate should automatically grant that user appropriate permissions when registered (#2186)~ (closed by #2538)

-  [x] ~Marketplace, Shop, and Affiliate owners should be able to add administrators (#2187)~ by #2588

-   [x] ~"Owner" permission should be transferrable to another member of a Marketplace / Shop / Affiliate (#2188)~ (closed by #2609)

-   [x] ~Permission to grant and remove permissions from non-owner users should be grantable (#2189)~

-   [x] Marketplace and shop owners, and account administrators should be able to view users with permissions (#2190) (Functionally complete, but needs UI work)

-   [x] ~~If a user is a member of multiple shops, they should be able to switch permission and shop dashboard contexts (#2191)~~ Solved by #2429

-   [ ] Marketplace, Shop, and Affiliate owners should be able to define the default permission set for new shops, affiliates, and admins they invite (#2192) (need to be able to edit default customer & guest groups)

-   [ ] Users with permission to edit permissions of other users should be able to 'bulk edit' the permissions of users (#2193)

-  [x] ~Owners should be able to create custom permission groups (#2194)~ closed by #2543 and #2448 

-   [x] ~New permissions added to an existing Reaction Commerce app should update existing users permission sets appropriately~ (#2195)

-   ~~Custom Permission Groups (#2509)~~ Closed

## Docs

-   Docs explaining what permissions in RC do and the basics of how they work

-   Define default permission levels for each audience

-   Define what each permission grants and what possible side-effects are (for each permission in core, also for each permission in included plugins)

-   Docs explaining best practices for checking permissions

-   Docs explaining best practices for granting permissions (both default and dynamically)

#### Possible Permission Groups
-   Marketplace Owner
-   Marketplace admin
-   Shop Owner
-   Shop Admin
-   Affiliate Shop Owner
-   Affiliate Shop Admin
-   Customer
-   Guest

Some marketplace permissions discussion in issue #357

spencern:
@impactmass @aaronjudd @zenweasel 

Can you guys take a look through these issues and let me know where you think there's enough detail and where there needs to be more detail?

[...]

PR #2448, opened 2017-06-15
[Marketplace] Feature - Permissions should be settable by placing a user into a predefined group

impactmass:
Permission groups. Implements #2184 
This PR has the Meteor methods to manage permission groups. It specifically adds `group/createGroup`, `group/updateGroup`, `group/addUser`, `group/removeUser`.
The UI for this is being handled separately (as that involved changing a lot of the current Accounts section of admin dashboard).

**To test:**
_ Take a look at the tests, to have a sense of how the methods work
_ The methods can be tested from the browser console of a running Reaction (via `Meteor.call` and passing the needed values). See the jsdocs added for the descriptions of the arguments.
* Start Reaction
* Create a shop (i.e register as a user (not the owner). Go to "Profile", click "Become a Seller"
* Create a group, by calling `group/createGroup` from console, passing in the required options.
* Confirm in db that a new doc is created in Groups collection
* Add a user to the group by calling `group/addUser` as described
* Confirm that the user's permissions gets updated (i.e Meteor.user.roles) and that the user's Account doc now has the groupId.
* The other methods follow similar pattern as shown in the tests

Other notes:
_ Groups exits as a schema
_ Group Ids are referenced on Accounts to track which users belong to a group


lib/collections/schemas/shops.js
```diff
@@ -107,6 +107,15 @@ export const ShopTheme = new SimpleSchema({
   }
 });
 
+const Groups = new SimpleSchema({
```
spencern: This should be singular perhaps - Group: has many permissions.

lib/collections/schemas/shops.js
```diff
@@ -244,6 +253,10 @@ export const Shop = new SimpleSchema({
       "cart/completed"
     ]
   },
+  "groups": {
+    type: [Groups],
```
spencern: this would then be an array of [Group] objects.

server/methods/core/groups.js
```diff
@@ -0,0 +1,207 @@
...
+
+    // must have needed permissions.. .TODO: Review who can create a group permission for a shop
+    if (!Roles.userIsInRole(Meteor.userId(), "admin", shopId)) {
+      throw new Meteor.Error(403, "Access Denied");
+    }
+
+    const shop = Shops.update({ _id: shopId }, { $addToSet: { groups: groupData } });
+
+    if (!shop) {
+      throw new Meteor.Error(400, "Bad request data");
```
spencern: I _think_ (but might be wrong) that we've been using `(400, "Bad Request") for these types of errors.

server/methods/core/groups.js
```diff
@@ -0,0 +1,207 @@
...
+import { Roles } from "meteor/alanning:roles";
+import { Logger } from "/server/api";
+import { Accounts, Shops } from "/lib/collections";
+
+/**
+ * Reaction Group Permission Methods
+ * TODO: Confirm if better to move this to shop.js since groups are on shops
+ */
+Meteor.methods({
+  /**
```
spencern: Really nice work creating good jsdoc here

server/methods/core/groups.js
```diff
@@ -0,0 +1,207 @@
...
+   * group/createGroup
+   * @summary creates a new permission group for a shop
+   * creates permission group for a given shop with passed in roles
+   * @param {Object} groupData - info about group to create
```
spencern: Is this the group "description"?

server/methods/core/groups.js
```diff
@@ -0,0 +1,207 @@
...
+    check(group.name, String);
+    check(group.permissions, [String]);
+
+    check(newGroupData, Object);
+    check(newGroupData.name, String);
+    check(newGroupData.permissions, [String]);
+    check(shopId, String);
+
+    // must have needed permissions.. .TODO: Review who can create a group permission for a shop
+    if (!Roles.userIsInRole(Meteor.userId(), "admin", shopId)) {
```
spencern: I think "admin" is fine for now. We should be using `Reaction.hasPermission` rather than `Roles.userIsInRole` where possible

server/methods/core/groups.js
```diff
@@ -0,0 +1,207 @@
...
+      updateUsersGroupName([user], shopId);
+      return { status: 200 };
+    } catch (error) {
+      Logger.error(error);
+      throw new Meteor.Error(500, "Could not add user");
+    }
+  }
+});
+
+function updateUsersGroupName(affectedUsers, shopId) {
```
spencern: This function seems really expensive. Maybe premature to optimize, but might be worth a conversation about if there's a better way to assign / check group membership

server/methods/core/groups.js
```diff
@@ -0,0 +1,207 @@
...
+    const matchQuery = { $elemMatch: { shopId: shopId, names: group.name } };
+    const affectedUsers = Accounts.find({ groups: matchQuery }).fetch();
+    let error;
+
+    if (permissionsChanged) {
+      error = updateUsersPermissions(affectedUsers, group, newGroupData, shopId);
+    }
+
+    // update group names on user account
+    error = updateUsersGroupName(affectedUsers, shopId);
```
spencern: Why does this function run here? What are we ensuring?

server/methods/core/groups.js
```diff
@@ -0,0 +1,207 @@
...
+    check(userId, String);
+    check(groupData, Object);
+    check(groupData.name, String);
+    check(groupData.permissions, [String]);
+    check(shopId, String);
+
+    const user = Accounts.findOne({ _id: userId });
+
+    try {
+      Meteor.call("accounts/addUserPermissions", user.userId, groupData.permissions, shopId);
```
spencern: One thing I've been thinking about is that it might be easier to do group assignment if any user can only have one group per shop for now. You could then add/remove the difference in permissions between users old group and users new group and just add the group name. It would negate the need to do the hard determining which group(s) a user is in by their permissions.

server/methods/core/groups.js
impactmass: OK. Thanks. I'll check to confirm and update

server/methods/core/groups.js
impactmass: Cool. I now get this after our call. Going this direction will help making it easy. And it'll also answer some questions around how expensive "updateUsersGroupName" function is.

server/methods/core/groups.js
impactmass: The comment [here](https://github.com/reactioncommerce/reaction/pull/2448/files#r123016688) will likely eliminate the need to do this

server/methods/core/groups.js
impactmass: I need to add description. I missed that out .. So I'm thinking like this now:
```
{
  "name":  "Customer Service",
  "description": "Handles all ... blah blah"
  "slug": (to be formed from the name. Will use this to reference on user Accounts)
}
```

spencern:
@impactmass FYI left some general permissions comments based on our recent discussion on the permissions epic #2183

server/methods/core/groups.js
spencern: Looks good. Might be worth having some meta information on each group as well, what do you think?

E.g. createdAt, updatedAt, createdBy: username, etc

impactmass: Yes. That makes sense too.

spencern:
@rymorgan and I were also just discussion that we should probably have `EventHooks` built into these methods as well so that we can send notifications and emails when a user's group is changed.

lib/collections/schemas/accounts.js
```diff
@@ -38,6 +38,17 @@ export const Profile = new SimpleSchema({
   }
 });
 
+const Group = new SimpleSchema({
+  shopId: {
+    type: String,
+    optional: true
+  },
+  groupId: {
+    type: [String],
```
spencern: I'm curious as to why both this (line 47) and line 117 have type as an array

lib/collections/schemas/shops.js
```diff
@@ -107,6 +108,45 @@ export const ShopTheme = new SimpleSchema({
...
+    type: String
+  },
+  description: {
+    type: String,
+    optional: true
+  },
+  slug: {
+    type: String
+  },
+  _id: {
```
spencern: let's keep `_id` at the top for consistency.

lib/collections/schemas/shops.js
```diff
@@ -119,7 +119,7 @@ const Group = new SimpleSchema({
   slug: {
     type: String
   },
-  _id: {
+  groupId: {
```
spencern: what's the thought behind this change? Using `groupId` instead of `_id` implicates to me that the groupId is pointing to another document or object elsewhere

server/methods/core/groups.js
```diff
@@ -202,6 +214,6 @@ function setUserPermissions(users, oldGroup, newGroupData, shopId) {
   return affectedUsers.forEach(user => Roles.setUserRoles(user._id, newGroupData.permissions, shopId));
 }
 
-function getGroupPermissions(groups, name) {
-  return _.get(_.find(groups, { name }), "permissions", []);
+function getGroupPermissions(groups, groupId) {
+  return _.get(_.find(groups, { groupId }), "permissions", []);
```
spencern: I think you can use Object prototype find here instead of lodash
Something like this perhaps
```
const group = groups.find(group => group.groupId === groupId);
return group.permissions || [];
```

server/methods/core/groups.app-test.js
```diff
@@ -0,0 +1,135 @@
...
+    Meteor.call("group/createGroup", sampleGroup, shop._id);
+    const group = Groups.findOne({ shopId: shop._id });
+    Meteor.call("group/addUser", user._id, group._id);
+    let updatedUser = Meteor.users.findOne({ _id: user._id });
+    expect(updatedUser.roles[shop._id]).to.include.members(sampleGroup.permissions);
+
+    const newGroupData = Object.assign({}, sampleGroup, { permissions: ["new-permissions"] });
+    Meteor.call("group/updateGroup", group._id, newGroupData, shop._id);
+    updatedUser = Meteor.users.findOne({ _id: user._id });
+    expect(updatedUser.roles[shop._id]).to.include.members(newGroupData.permissions);
```
spencern: Might be worth adding a test to make sure the permissions that the old group had, but the new group does _not_ have are no longer included.

server/methods/core/groups.js
```diff
@@ -0,0 +1,161 @@
...
+    });
+
+    try {
+      Groups.insert(newGroupData);
+    } catch (error) {
+      Logger.error(error);
+      throw new Meteor.Error(400, "Bad request");
+    }
+
+    return { status: 200 };
```
spencern: We should probably return the `_id` of the created group here as well?

server/methods/core/groups.js
```diff
@@ -0,0 +1,161 @@
...
+import { Roles } from "meteor/alanning:roles";
+import { Reaction, Logger } from "/server/api";
+import { Accounts, Groups } from "/lib/collections";
+import { getSlug } from "/lib/api";
+
+/**
+ * Reaction Permission Group Methods
+ */
+Meteor.methods({
+  /**
```
spencern: Great job with jsdoc here.

server/methods/core/groups.app-test.js
```diff
@@ -0,0 +1,135 @@
...
+      return methods[method].apply(this, arguments);
+    });
+  }
+
+  it("should create a group for a particular existing shop", function () {
+    sandbox.stub(Reaction, "hasPermission", () => true);
+    spyOnMethod("createGroup", shop._id);
+
+    Meteor.call("group/createGroup", sampleGroup, shop._id);
+    const group = Groups.findOne({ shopId: shop._id });
```
impactmass: Finding groups belonging to a shop can be done with the slug property. More can be seen in `addDefaultRoles.app-test.js`. This is just doing a findOne because just one group was added in the beforeEach.

lib/collections/schemas/accounts.js
```diff
@@ -107,6 +107,11 @@ export const Accounts = new SimpleSchema({
     type: Profile,
     optional: true
   },
+  groups: {
+    type: [String],
+    optional: true,
+    defaultValue: []
```
impactmass: No. It's just means at the schema level, they don't belong to a group yet. Although, on account creation, the array then receives either "guest" or "customer".

impactmass:
Brought the other work of default groups into here (since they're closely linked, and better reviewed from together after discussing with Spencer). So, server implementation for #2184 along with #2194 is here.

client/modules/core/helpers/layout.js
```diff
@@ -19,6 +19,7 @@ Template.registerHelper("reactionTemplate", function (options) {
   const shopId = options.hash.shopId || Reaction.getShopId();
   // get shop info, defaults to current
   const Shop = Collections.Shops.findOne(shopId);
+  const defaultRoles = Collections.Groups.findOne({ slug: "customer", shopId }).permissions;
```
spencern: Do we need to ensure that defaultRoles are loaded at this point? (e.g. `subscription.ready()`)?

spencern: This helper is only used in the checkout right now it seems, and even if we load the checkout directly default roles are ready, so I think we're good to go here.

PR #2450, opened 2017-06-16
Fix ESLint4 errors

zenweasel:
(info from @jshimko) 
ESLint 4.0 doesn't like invalid parameters but ESLint 3 ignored them. So if you fix the invalid parameter here (e.g `ecmaFeatures` at top level rather than inside `parserOptions`) you realize you have a lot more errors. This PR fixes all these lint errors.

Also this remove the deprecated `import { Proptypes } from "react"` and replaces it with importing from the `prop-types` package


package.json
```diff
@@ -77,6 +77,7 @@
     "react-avatar": "^2.3.0",
     "react-color": "^2.11.4",
     "react-copy-to-clipboard": "https://registry.npmjs.org/react-copy-to-clipboard/-/react-copy-to-clipboard-5.0.0.tgz",
+    "react-create-class": "^1.0.0",
```
aaronjudd: ```npm WARN deprecated react-create-class@1.0.0: This package is empty. You probably meant create-react-class.```

zenweasel:
@aaronjudd Corrected `react-create-class` package

zenweasel:
Ugh, so isn't ESLint supposed to catch undefined variables? If not, what are you even doing ESLint?

zenweasel:
Yeah, so no `no-undef` and excluding `browser` gives us:
```
✖ 2183 problems (2183 errors, 0 warnings)
```

aaronjudd:
Well, which is it? Browser or undefined?

zenweasel:
Pardon?

aaronjudd:
@zenweasel it'd be good to bring this up to date, I suspect there are some fixes in here that might actually help with our build issues.

zenweasel:
@aaronjudd I still had about 150 left (down from about 1300). I can try to finish those today. There are a few I need to leave because they look like actual bugs we need to fix. It's just slow, tedious work. 90% are just missing imports but there are few actual problems.

If we want to tweak the ESLint at all, this is probably a good time to do it, so maybe we want to put our heads together and see what changes we might want to make.

.eslintrc
```diff
@@ -102,6 +102,7 @@
     "react/jsx-boolean-value": 0,
     "react/jsx-closing-bracket-location": 1,
     "react/jsx-curly-spacing": 1,
+    "react/jsx-indent": [1, 2],
```
zenweasel: This change seems to be causing a conflict with the existing indentation rule that is prevent ESLint from autofixing files.

zenweasel: Yeah, they are definitely conflicting. Fix errors from one causes error from the other.

zenweasel:
Okay, I have finished everything that was obvious and that I could figure out relatively easier. The rest are either packages I can't find where they come from (maybe globals, Gravatar, Package) and a few that seem to be undefined variables there what they are is not obvious

zenweasel:
The packages that seem to be globals are:

```
amplify
Package (I think this is some sort of legacy Meteor thing)
Gravatar
sAlert
ReactionUI
```
I also wasn't sure there to import `slugify` from as I believe there was some issue when importing that on the client.

zenweasel:
Doing `import { amplify } from "meteor/amplify";` definitely does not work

.eslintrc
```diff
@@ -160,7 +172,8 @@
     "no-self-compare": 2, // http://eslint.org/docs/rules/no-self-compare
     "no-sequences": 2, // http://eslint.org/docs/rules/no-sequences
     "no-throw-literal": 2, // http://eslint.org/docs/rules/no-throw-literal
-    "no-with": 2, // http://eslint.org/docs/rules/no-with
+    "no-with": 2, // http://eslint.org/docs/rules/no-with,
+    "no-undef": 2,
```
zenweasel: Here's the cause of all the problems/solutions

imports/plugins/core/layout/client/templates/layout/header/header.js
```diff
@@ -16,38 +16,12 @@ Template.layoutHeader.events({
 });
 
 Template.layoutHeader.helpers({
-  TagNav() {
```
zenweasel: I removed this helper because it was using an undefined `ReactionUI` component and did not seem to be used (should all be React now anyway right?)

zenweasel:
@jshimko If you have a chance to look at this I would appreciate it

PR #2453, opened 2017-06-16
refactor default admin user creation

jshimko:
This updates several aspects of the default admin creation as a result of fairly regular feedback from users that have been getting stuck with the initial setup of their Reaction shops in production environments (Docker image deployments).  There are two regularly recurring issues that this aims to address.  

First, if default admin credentials aren't provided by the user through environment variables or Meteor settings, random credentials are generated for them.  If mail sending isn't already configured, those new credentials get output to the console one time and can't ever be retrieved or regenerated again.  The major problem with that is if you no longer have access to those first console logs, or are new to Docker, or just generally didn't realize you needed to look out for that, then you end up in a situation where you can't log in without manually deleting your default admin from the database to get a new user created.  That's less of a problem in development (`reaction reset` solves that), but that's not pleasant in a production deployment where logs are a little more hidden and resetting the database is a lot more manual.

The second part to that is the requirement for email validation in production deployments. If you don't have email sending configured with a third party provider (which is extremely common for new users starting their shop the first time) then you don't receive the validation email that gets sent to the default admin.  As in the case above, it outputs a validation link to the console.  But if you happen to not have access to those logs, your only options to log in to your new shop are to delete the user and start over or edit your user in the database to set their email to "validated".  And that's assuming you know that user's credentials and now how to edit the database.

So both of these scenarios combined have ended up causing the same support questions to come up over and over from people that are locked out of their new shop.  In an effort to make that initial user experience a little more flexible, the following changes have been made in this PR.

- if no credentials are provided via environment or Meteor settings, the default credentials are: 
  - email: `admin@localhost`
  - password: `r3@cti0n`

- the default admin user's email is set to "validated" so they can log in without requiring them to have email sending configured or access to server logs (this already happens in development, but now happens in production as well)

However, if you preferred the previous behavior (random credentials and required email validation), you can re-enable the process to be 100% identical to the previous state by setting the following environment variable:

```sh
REACTION_SECURE_DEFAULT_ADMIN=true
``` 

That will cause random credentials to be generated (when none are provided) and will enforce email validation (link will either be emailed or printed to the console).  

Totally unrelated, I also discovered that the shop domain was being updated with whatever the value of `Meteor.settings.ROOT_URL` was.  When this isn't set (which is the default on a fresh copy of Reaction), it would insert an empty string into the domains array of the current shop.  I updated this to check the current shop for the current domain and only add it if it doesn't already exist.  I don't think this was actually breaking anything, but it appears that nearly all shops were getting an empty string inserted into the domains array if they didn't use a settings.json on first startup.  Oops!  :)

Possible discussion for future changes... I think the ideal for the default admin creation would be that we have an onboarding welcome screen like Wordpress or Ghost that shows up any time no admin users exist that lets someone create that default admin right in the browser.  There's certainly value in having it be something we can automate with environment variables, but I think a lot of people would appreciate being able to just start the app and do the rest in the browser.  That's obviously a bit more work than I did here, but I think it's probably worth discussing so we can make that initial setup a little more user-friendly for non-developer folks.

If we're ok with merging this, I'll update the docs to reflect the changes.


server/api/core/core.js
```diff
@@ -1,3 +1,4 @@
+import _ from "lodash";
 import url from "url";
 import packageJson from "/package.json";
 import { merge, uniqWith } from "lodash";
```
aaronjudd: Since you are importing _ from lodash... isn't this a redundant import?

server/api/core/core.js
```diff
@@ -268,29 +261,49 @@ export default {
...
-    if (env.REACTION_EMAIL && env.REACTION_USER && env.REACTION_AUTH) {
+    // If $REACTION_SECURE_DEFAULT_ADMIN is set to "true" on first run,
+    // a random email/password will be generated instead of using the
+    // default email and password (email: admin@localhost pw: r3@cti0n)
+    // and the new admin user will need to verify their email to log in.
+    // If a random email and password are generated, the console will be
+    // the only place to retrieve them.
+    // If the admin email/password is provided via environment or Meteor settings,
+    // the $REACTION_SECURE_DEFAULT_ADMIN will only enforce the email validation part.
+    const isSecureSetup = process.env.REACTION_SECURE_DEFAULT_ADMIN === "true";
```
aaronjudd: This is defaulting to the secure setup?

server/api/core/core.js
aaronjudd: Since you are importing _ from lodash... isn't this a redundant import?

server/api/core/core.js
```diff
@@ -301,15 +314,27 @@ export default {
...
-        domains: Meteor.settings.ROOT_URL
+        }
       }
     });
 
+    // get the current shop
+    const shop = Shops.findOne(shopId);
+
+    // add the current domain to the shop if it doesn't already exist
+    if (!_.includes(shop.domains, this.getDomain())) {
```
aaronjudd: Our goal should be to use native methods as much as possible, someday maybe no underscore or lodash.. `shop.domains.includes(this.getDomain());`?

server/api/core/core.js
```diff
@@ -348,22 +369,20 @@ export default {
     //
     // Set Default Roles
     //
-
+    const defaultAdminRoles = ["owner", "admin", "guest", "account/profile"];
     // we don't use accounts/addUserPermissions here because we may not yet have permissions
-    Roles.setUserRoles(accountId, _.uniq(defaultAdminRoles), shopId);
```
aaronjudd: Was the reason that _.uniq was here because at some point we were allowing `defaultAdminRoles` to be overridden, or perhaps set in `Shops`.  Not sure that ever happened, so this change isn't changing the current behavior, but feels as though something got lost. @impactmass you'll want to be aware of this for implementation of permission groups.

server/api/core/core.js
jshimko: Yep, good catch.  Didn't see the other lodash import.

server/api/core/core.js
jshimko: No, it's setting `isSecureSetup` to the boolean value of this equality check:
```js
process.env.REACTION_SECURE_DEFAULT_ADMIN === "true"
```

server/api/core/core.js
jshimko: Yeah, I have no idea why that `_.uniq` was there.  Presumably from some other time when setting custom roles was possible.  At this point, those role values are hard coded, so there was no need to run that lodash method on them.  (unrelated, that `_.uniq` was using the `underscore` global from Meteor because `lodash` wasn't being imported previously)

jshimko:
All requested fixes/comments submitted.

zenweasel:
This is what a good PR and code review looks like

PR #2472, opened 2017-06-20
Fix on tag menu-sidebar glitch

### LINKED ISSUE #2356: New Tags Menu SideBar Glitch
It looks like this is related to some previous issues that HAVE been fixed.  One more detail that doesn't seem to be what we'd expect: 

When in small screen view, the drop down side bar appears and allows us to view all tags (parent and child-level).  We should be able to select the parent-level tag by clicking on "View All" 

When selecting "View All", we see "#"   added to the url, but no re-route to the list page. 

screen shot 2017-05-26 at 2 01 08 pm

rhenshaw56:
Resolves #2356 
**Testing**
- Start the app and open in either mobile view or full screen mode
- Hover on any tag on the tags menu and notice a dropdown pops out with child tags
- Click on `View all` and observe that it routes to the required page

### DIFF
--- a/imports/plugins/core/ui-tagnav/client/components/tagGroup.js
+++ b/imports/plugins/core/ui-tagnav/client/components/tagGroup.js
@@ -123,7 +123,7 @@ class TagGroup extends Component {
       <div className="rui tagtree">
         <div className="header">
           <span className="title">{this.state.parentTag.name}</span>
-          <a href="#">View All <i className="fa fa-angle-right" /></a>
+          <a href={this.state.parentTag.slug}>View All <i className="fa fa-angle-right" /></a>
         </div>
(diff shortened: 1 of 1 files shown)


zenweasel:
When I test this it is taking me to `/reaction/this` rather than `/reaction/tag/this` (where `this` is the name of my tag

rhenshaw56:
@zenweasel i have refactored it to include the `tag` in the `url`

zenweasel:
Ok, are you testing these before you submit them? Seems like that should have been caught by just following your own testing instructions.

rhenshaw56:
@zenweasel yeah, i tested it, but i failed to catch the part where if the user is on the index page, it does not route to the required page.

Issue #2438, opened 2017-06-15
"Ooops you don't have permission" screen shown briefly when logging in


zenweasel:
### Expected behavior
A login screen should not be shown in the main screen while the login is processing

### Actual Behavior
"Ooops you don't have permission" screen is briefly shown while login screen is processing
![login](https://user-images.githubusercontent.com/72819/27163286-eeacaac6-51b8-11e7-9981-e8dc65c8703f.gif)

### Steps to Reproduce the Behavior
1. Open Reaction
1. Login
1. Observe login screen is shown briefly

### Versions
```
Node: 7.10.0
NPM: 4.2.0
Meteor Node: 4.8.3
Meteor NPM: 4.6.1
Reaction CLI: 0.9.3
Reaction: 1.2.0
Reaction branch: ek-ordersPlaceholder
Docker: 17.03.1-ce
```

rhenshaw56:
I believe there is a fix for this on #2459

zenweasel:
@rhenshaw56 That PR was merged. Can you verify this was fixed and close this ticket if it is?

Issue #2459, opened 2017-06-19
RangeError: Maximum call stack size exceeded in Development Branch


stevaras:
**Expected behavior:**
When I register I shouldn't see a flickering in the screen and get an exception error.

**Actual Behavior:**
When I registered I actually saw a flickering in the screen and got an exception error.

![reaction-issue](https://user-images.githubusercontent.com/13280180/27282705-c3c4133c-54f9-11e7-8475-017cdb9e04a0.png)

**Steps to Reproduce the Behavior:**
Create a new user through the anonymous dropdown menu process. 

**Versions:**
Node: 7.1.0
NPM: 5.0.3
Meteor Node: 4.8.3
Meteor NPM: 4.6.1
Reaction CLI: 0.9.3
Reaction: 1.3.0
Reaction branch: development

aaronjudd:
@reactioncommerce/engineering this is a priority blocker that needs fixing for 1.3.0 release.

aaronjudd:
I also see max call stack errors when clicking on the "profile" link within the user dropdown.. ```modules.js?hash=714f292…:87208 Uncaught RangeError: Maximum call stack size exceeded
    at http://localhost:3000/packages/modules.js?hash=714f292fe02c3b8afedec4d1ebf4a03c187b0eb7:87208:23
    at Object.confirmTransitionTo (http://localhost:3000/packages/modules.js?hash=714f292fe02c3b8afedec4d1ebf4a03c187b0eb7:86946:7)
    at Object.replace (http://localhost:3000/packages/modules.js?hash=714f292fe02c3b8afedec4d1ebf4a03c187b0eb7:87199:23)
    at Function.Router.replace (http://localhost:3000/app/app.js?hash=ce3cabe137ab2549139e11d64d88145e4171d397:78621:14)
    at BrowserRouter._this.handleLocationChange (http://localhost:3000/app/app.js?hash=ce3cabe137ab2549139e11d64d88145e4171d397:79611:17)
    at listener (http://localhost:3000/packages/modules.js?hash=714f292fe02c3b8afedec4d1ebf4a03c187b0eb7:86956:24)
    at http://localhost:3000/packages/modules.js?hash=714f292fe02c3b8afedec4d1ebf4a03c187b0eb7:86975:23
    at Array.forEach (native)
    at Object.notifyListeners (http://localhost:3000/packages/modules.js?hash=714f292fe02c3b8afedec4d1ebf4a03c187b0eb7:86974:15)
[...]

PR #2476, opened 2017-06-20
2459 router range error fix

prosf:
https://github.com/reactioncommerce/reaction/issues/2459

> 
I looked into it a little bit, and the problem seems to be present when the router is re-initialized (As it happens when a user subscribes/logins):

Accounts.onLogin(() => { if (Meteor.loggingIn() === false && Router._routes.length > 0) { initBrowserRouter(); } });

@ /imports/plugins/core/router/client/startup.js

So if a Router.go(route) is triggered at the time which a reinitialization of the Router happens (which empties _routes before rebuilding them) ,
the route as well as the "index" fallback can't be found at _routes so the Router has nowhere to go(!).

I think that the Router and its methods should be more aware of its 'readiness' in order to avoid similar circumstances.
I tried a small fix which seems to work for the current case, which I 'll push as PR just for reference for a probably more robust solution from the official assignees

### DIFF
--- a/imports/plugins/core/router/client/startup.js
+++ b/imports/plugins/core/router/client/startup.js
@@ -7,8 +7,17 @@ import { Router } from "../lib";
 Meteor.startup(function () {
   Tracker.autorun(function () {
     // initialize client routing
+
     if (Reaction.Subscriptions.Packages.ready() && Reaction.Subscriptions.Shops.ready()) {
-      initBrowserRouter();
+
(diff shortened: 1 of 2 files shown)


kieckhafer:
This seems to fix the issues we were seeing with the Range / call stack, I haven't been able to replicate those issues on this branch.

I think this is a good solution, if not just temporary, until we can dig into routing.

rhenshaw56:
This also resolves #2438

PR #2477, opened 2017-06-20
[WIP] Fix on dashboard analytics settings

### LINKED ISSUE #2301: Dashboard analytics settings not working.
The settings for analytics are not working in the dashboard. They can be configured in reaction.json, but the UI is broken. 

basic_reaction_product

## Versions:
Node: 7.5.0
NPM: 4.1.2
Meteor Node: 4.8.2
Meteor NPM: 4.5.0
Reaction CLI: 0.8.2
Reaction: 1.2.0
Reaction branch: 1863-toolbar-take2
Docker: 17.03.1-ce

rhenshaw56:
[WIP] Resolves #2301 
**Testing**
- Clone this branch
- Follow the directions on the [Reaction docs](https://docs.reactioncommerce.com/reaction-docs/master/analytics) to setup `Analytics` (preferably both `Google` and `Segment.io`)
- Login as admin and try to configure analytics from the dashboard
- Click on the gear icon and observe that the pane for the analytics settings shows


rhenshaw56:
@aaronjudd @zenweasel the flow of the UX is a bit off on the analytics page. For instance if a user clicks on the gear icon under the `Segment` section and all three are configured to work, then the settings reveal all three. I would like your input on how to proceed with it.

aaronjudd:
This actually needs a rework @rymorgan.  For now, what I think we should do is remove line https://github.com/reactioncommerce/reaction/blob/development/imports/plugins/included/analytics/register.js#L25 from the registerPackages statement, which will remove the weird use of Analytics twice in the dashboard.  We can then combine the templates <reactionAnalytics> and <reactionAnalyticsSettings> on the same page.  As long as every form works, you could just leave them visible, this should be quick.  What really needs to happen..   to get converted over to React, as these are pretty simple settings forms that should be using our standard collapsible settings pattern.

rymorgan:
@aaronjudd So is the requirement that we want the ability to enable integrated analytics tools (segment, GA and Mixpanel into an analytics dashboard along side our Rection analytics?) Are we also going to enable them to pull analytics from those sources into our UI? I'd love to hammer out the requirements for analytics at some point to get the whole picture.

rymorgan:
This fix needs some tweaks. Removing the design review-label until it's ready again. 

analytics

zenweasel:
I believe we are still waiting for some direction per your note @rymorgan? That's why I placed it in design review. If not I guess we can try to implement Aaron's last comment from 6/21? As per Rowland's last comment, he is waiting for direction on how to proceed.

mfdoss:
STOP SENDING ME THESE EMAILS. YOUR UNSUBSCRIBE LINK IS BROKEN.

zenweasel:
@mfdoss We don't have anything to do with which emails you receive, those are all in your settings and is between you and Github. We can neither subscribe you nor unsubscribe you from emails.

aaronjudd:
@mfdoss you probably need to unwatch the project-> (button at top of project ). As @zenweasel says has nothing to do with us.. we are not GitHub.  Assuming that you will see this comment, so I will be deleting both your comment and ours as neither are relative to the issue at hand.

rymorgan:
@zenweasel You didn't see my above comments? (from two days ago?)

zenweasel:
@rymorgan I did, but I don't see how that answers Rowland's question. What am I missing?

zenweasel:
@aaronjudd so do we want the three settings for Segment, GA, and Mixpanel like in the one above (that we will be removing)? If that's so then I think we can fix this up pretty quickly. That current settings panel just has Segment.

rymorgan:
@aaronjudd Is that what you had in mind? Can we just remove the duplicate analytics item, collapse into one and have it work like this? 

https://zpl.io/1F9Wy

public link: 
https://scene.zeplin.io/project/58af8900737c137280322bd6/screen/5966eced903ca17a5cb7a171

zenweasel:
@rhenshaw56 So I think we want to eliminate the first Analytics panel as Aaron mentioned above and then implement the design that Ryan has provided, replacing the existing settings panel. I think we want to rewrite it in React as well. You may be able to use the `<Form>` component that Mike created to use instead of autoform. You see it in use on a couple of the existing settings panels.

imports/plugins/included/analytics/client/components/settings.js
```diff
@@ -0,0 +1,86 @@
...
+import PropTypes from "prop-types";
+import _ from "lodash";
+import {
+  CardGroup,
+  SettingsCard,
+  Form
+} from "/imports/plugins/core/ui/client/components";
+import { ReactionAnalyticsPackageConfig } from "/lib/collections/schemas/analytics";
+
+const analyticsProviders = [
```
zenweasel: This should probably be declared in the `settings` portion of `registry.js` rather than in a presentation component. Then you can just grab it when you grab the rest of the settings.

imports/plugins/included/analytics/client/containers/analyticsSettingsContainer.js
```diff
@@ -0,0 +1,41 @@
+import { Meteor } from "meteor/meteor";
```
zenweasel: Import order: https://docs.reactioncommerce.com/reaction-docs/master/styleguide

zenweasel:
This should probably read `Segment.io` rather than `Segmentio`.

reaction

zenweasel:
I don't seem to be able to flip the switches on GA and Mixpanel.

zenweasel:
There doesn't appear to be any validation happening. If I try to save an empty API key it just lets me.

zenweasel:
@rhenshaw56 What's the status of this?

rhenshaw56:
@zenweasel i've fixed the changes you requested for actually, just needed to update the branch with changes from development and a circle ci rebuild

zenweasel:
When I enable Segment I am getting this toast with the translation key:

reaction

zenweasel:
When I click on the toggle or enter an API key for GA or Mixpanel I get this error message and can't enable either one?
reaction

rhenshaw56:
@zenweasel the toast and error messages are originating from this [file](https://github.com/reactioncommerce/reaction/blob/b277f14f66fed84a2f0b322bc11db7266ac9a79d/imports/plugins/included/analytics/client/startup.js#L152). I believe it was originally setup so that if there's no api key and then a user decides to enable analytics by toggling, the alert comes up.

zenweasel:
Yes, but I can't put in an API either without the message coming up

aaronjudd:
Regarding design: yes, that's what I had in mind (and this looks great).  I think though, that we need to make sure the flow actually works, I seem to remember, that the three analytics options were child options of Segment, while there maybe was also analytics that didn't require Segment (GA?).  In any case, not sure these options work without Segment configured first.   

In testing this, there wasn't any confirmation that the settings were saved, and I checked the https://segment.com/reaction-commerce/sources/reactioncommerce-core/debugger and didn't see any activity.  Actually fixing analytics might have been out of the scope of this.. but... should be looked at.

aaronjudd:
I closed this Analytics PR, only because it's not moving forward this second. There do seem to be events being reported, but we're missing the majority of the event that were being tracked previously -> through a slow deprecation of Blaze elements, and we've failed to add the necessary tracking and funnels to the replacement React components.  There is a larger analytics refactor on the roadmap but if anyone ones to take at bringing the current analytics back to life, this is a good place to start.

PR #2489, opened 2017-06-22
Fix hard refresh on clicking tag link on product detail page

### LINKED ISSUE #2482: clicking tag link on PDP causes hard refresh
### Expected behavior

Clicking link should immediately render page.

### Actual Behavior

The browser does a hard refresh and downloads/re-renders the entire app (which takes several seconds instead of milliseconds).  This is most likely related to #2246 and probably caused by the recent Router updates.

Visit the product detail page, click a tag link under the product image.

### Versions

Node: 8.1.2
NPM: 5.0.3
Meteor Node: 4.8.3
Meteor NPM: 4.6.1
Reaction CLI: 0.10.0
Reaction: 1.3.0
Reaction branch: Release-1.3.0
Docker: 17.03.1-ce

basic_reaction_product

awadesmoline:
This resolves #2482 

### Steps to test
1. Visit the product detail page, click a tag link under the product image.


jshimko:
While you're there, here's another link that's doing the same thing...

reaction_swag_shop_and_slack_-_reaction

jshimko:
@mikemurray @kieckhafer  I'm also wondering... why aren't we using [React Router's `<Link/>` helper](https://reacttraining.com/react-router/web/api/Link)?  Isn't this what it's specifically intended to be used for?  Having to make click handlers with `Router.go()` for links all over the app seems like pretty tedious and repetitive work.  

Maybe it's because we have some custom logic happening in our route changes, but I think we need to make our own `<Link/>` component if that's the case.  This is just far too common to require click handler boilerplate everywhere.  Replacing `<a/>` tags with `<Link/>` tags is also a bit more intuitive than having to attach event handlers to every link that lives inside markup.

aaronjudd:
I think part of the reason was to maintain backward compatibility. But no reason not to migrate to a better technique if in core if we maintain that for (a while?).

jshimko:
Yeah, I agree that we probably shouldn't import directly from the React Router package, but we can certainly wrap it in the Reaction Router package to maintain backward compat.  

Wait on merging this and I'll take a quick look.

PR #2491, opened 2017-06-22
Remove redundant risul:moment-timezone

aaronjudd:
- Removes the Meteor package risul:moment-timezone
- Implements missing imports for moment


zenweasel:
@jshimko Is there a reason we didn't merge this?

PR #2501, opened 2017-06-26
Marketplace merge development

spencern:
Merges latest development into Reaction


zenweasel:
I suspect problems with the shop switcher, etc. are already known so I am ignoring.

spencern:
What problems did you have with the shop switcher?

zenweasel:
When I select "blank site" from the switcher I got a bunch of errors

spencern:
oh, hmm. I actually haven't done much with the default second site, so I can check on that, but I don't think it's been introduced in this merge, so approval seems okay still.

spencern:
CircleCI is finished and passed, not sure why it's saying queued still. Merging.

PR #2515, opened 2017-06-30
Default to using shop country code or GeoIP result for new addresses

spencern:
For the address add template, default the country dropdown to use the active shop's country if it exists instead of defaulting to unselected.

To test: 
1. Make sure your shop has an address set.
2. Go through the checkout process as a new / logged out user (incognito works).
3. Observe that the country for your new address is set by default to the shops country.


jshimko:
I took this a step further and added a call to our GeoIP server that determines the user's current country.  Your changes are still in effect, but now it also reactively updates if/when the request returns a country code.  

So the checks go in this priority order and the last value found wins: 

1. default shop country
2. Session (if set)
3. user's actual location based on their IP

FYI, to check out the available results from GeoIP, you can:

```sh
curl https://geo.getreaction.io/json/
```

When no arguments are supplied, it assumes it should check the IP of the requester (automatically provided in the request headers).

Issue #1993, opened 2017-03-15
Desktop-Safari: RC does not work properly in incognito/private mode


prinzdezibel:
Stock RC , running from localhost

Open incognito/private window in desktop Safari browser. Navigate to localhost:3000.

### Expected behavior
Reaction Commerce start page to be shown.

### Actual Behavior
Empty screen appears.

### Elusive solution
Hit refresh and everything seems fine. Except that it's not.

### Steps to reproduce
Put some items into cart and start Paypal-Express checkout. Ensure auth+capture is enabled in settings. After returning from PayPal's site (In my case the checkout process was NOT done in-context, but redirected to domain paypal.com. Be aware that this may differ depending on many things) an error is shown:

```
There was a problem with your payment.
ACK Failure: This transaction cannot be processed. The amount to be charged is zero.:10525
```

prinzdezibel:
Apparently the implementation of private mode differs between Safari and other browsers. This could be the reason, why they behave differently after redirect from paypal.

https://gist.github.com/JohnAllsopp/1470989

spencern:
I think this may have something to do with Safari not implementing LocalStorage in private mode.

We experience similar issues (though we don't use paypal) with GetOutfitted.

First load - safari private mode
https://www.dropbox.com/s/b19tibr025zo6i6/Screenshot%202017-03-15%2014.38.53.png?dl=0

Refresh and everything is fine.

This StackOverflow answer was helpful for me in understanding the safari private browser issue

http://stackoverflow.com/a/27081419/1179341

spencern:
I was able to shim localStorage and prevent it from throwing errors, but the rendering issue persists, so it does not appear that it's related to the localStorage issue.

prinzdezibel:
To have PayPal express checkout to work in Safari private mode, RC needs a way to re-login through preserving the current userLogin. 

I see several possibilities to achieve this:

1) Through setting a token/sessionId as cookie before navigating away to PayPal and read it again after PayPal has done its job and calls the redirect (`window.location.href = <shopUrl>/paypal/done`)

2) Setting a loginToken as CUSTOM field when calling `getExpressCheckoutToken` and pass it through the PayPal process. Retrieve this CUSTOM field after PayPal has done its job and calls the redirect (`window.location.href = <shopUrl>/paypal/done`) through PayPal's getExpressCheckoutDetails API and login via `loginWithToken`

[...]

PR #2518, opened 2017-07-02
Fix for issue #1993

prinzdezibel:
Resolves: #1993 
Safari incognito mode does not support local storage.

 - This prevents re-login through local storage
 - Can happen whenever app is navigating away from home domain, e.g. Paypal-Express checkout flow
 
The fix is to have the application set a login token as cookie whenever the page is going to be unloaded (window.onbeforeunload). After the external process the app will be initialized again: A good place to read the cookie again and log in.


prinzdezibel:
@aaronjudd  Sorry for being a bit late with this PR, even though I offered it quite a while ago. I was very busy the last couple of months.

client/modules/core/startup.js
```diff
@@ -2,7 +2,12 @@ import store from "amplify-store";
 import { Meteor } from "meteor/meteor";
 import { Tracker } from "meteor/tracker";
 import { Accounts } from "meteor/accounts-base";
+
 import Reaction from "./main";
+import Logger from "/client/modules/logger";
```
zenweasel: I think we want to import `Reaction` and `Logger` from `.client/api` like we do elsewhere for consistency.

client/modules/core/startup.js
```diff
@@ -14,17 +19,66 @@ Meteor.startup(function () {
   // initialize anonymous guest users
   return Tracker.autorun(function () {
     const userId = Meteor.userId();
+
+    if (userId && !isLocalStorageAvailable() && !readCookie(cookieName)) {
+      Logger.info("No local storage available. About to set up fall-back login " +
```
zenweasel: This should probably not be an `info` level log.

client/modules/core/startup.js
prinzdezibel: sure. I'll fix.

client/modules/core/startup.js
prinzdezibel: what log level do you find appropriate?

zenweasel: debug?

zenweasel: we're all about the `debug` these days. `info` is so 2016

prinzdezibel: :)

client/modules/core/startup.js
prinzdezibel: done.

client/modules/core/startup.js
prinzdezibel: done.

PR #2531, opened 2017-07-06
[WIP] Adds inline validation to form variant form

mikemurray:
Adds inline validation to variant form, attempts to make it as reusable as possible.

### DIFF
--- a/imports/plugins/core/ui/client/components/textfield/textfield.js
+++ b/imports/plugins/core/ui/client/components/textfield/textfield.js
@@ -14,6 +14,18 @@ class TextField extends Component {
     return this.props.value || "";
   }
 
+  get isValid() {
+    const { isValid, validation } = this.props;
+
+    if (typeof isValid === "boolean") {
+      return isValid;
+    } else if (typeof validation === "object") {
+      return false;
+    }
+
+    return undefined;
+  }
+
   /**
    * onValueChange
    * @summary set the state when the value of the input is changed
@@ -134,6 +146,14 @@ class TextField extends Component {
   }
 
   renderHelpText() {
+    if (this.props.validation) {
+      return (
+        <span className="help-block">
+          <Translation defaultValue={this.props.validation.message} i18nKey={this.props.validation.message} />
+        </span>
+      );
+    }
+
     if (this.props.helpText) {
       return (
         <span className="help-block">
@@ -155,6 +175,8 @@ class TextField extends Component {
       "rui": true,
       "textfield": true,
       "form-group": true,
+      "has-error": this.isValid === false,
+      "has-success": this.isValid === true,
 
       // Alignment
       "center": this.props.align === "center",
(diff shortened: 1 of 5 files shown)


imports/plugins/core/collections/lib/validation.js
```diff
@@ -0,0 +1,60 @@
...
+    const messages = {};
+
+    // clean object, removing fields that aren't in the schema, and convert types
+    // based on schema
+    const cleanedObject = this.schema.clean(objectToValidate, this.cleanOptions);
+
+    // Validate the cleaned object
+    const isValid = this.validationContext.validate(cleanedObject);
+
+    // Avoding the reactive-stuff built into simple-schema, grapy invalid
```
zenweasel: typo in this and the next line?

zenweasel:
So if I leave off the "label", I am still not getting an inline validation? Just getting the Toast that was there before.

mikemurray:
@zenweasel are you leaving the label off of the `top level variant` or a `child variant`?

For me, both do inline validation. It's only when I try to publish does the toast come back into play.

zenweasel:
@mikemurray Yeah, it's when I do publish that I am getting the toast. Are we not doing a validation before doing a publish?

mikemurray:
@zenweasel no validation before publishing. It's only on blur event of the fields so the user knows that `label` is required. But, it would stop submission if the publish button is active and clicked.

Publish is far removed from the context of the fields. And publish can do multiple documents; all docs to be published will need to be validated and their errors presented in some fashion, (toast, inline, or otherwise).

mikemurray:
This PR at the moment may not **100%** resolve all validation issues, only fix functionality that was lost with the react conversion away from autoform.

zenweasel:
That's a serious problem imho. Label is especially problematic because it's really easy to skip and then the Toast gives you no idea where that field even is. While the inline validation is good I don't feel like it deal with the original issue or the usability problems.

zenweasel:
@mikemurray discussed this and decided that while this doesn't solve all the validation problems it solves and lot of them and is worth merging this PR (when complete) and leaving the original issue open.

### Current open issues with this PR:

1. Selecting "Origin Country" does not change the field value when any other field is invalid
1. Cannot turn on the "Allow Backorder" switch when any other field is invalid
1. Cart Label is required but no validation is shown unless you fill the field and then remove the value
1. Should Option label be required since we default it to "Untitled Option" and that's what is shown in the Product Grid?
1. Should the publish button be disabled when there are validation errors? I wonder if that wouldn't help solve the above problem?

mikemurray:
Latest changes:

- Forms are validated initially
- Valid fields will save, invalid fields will not. This should fix the issues with the switches and dropdown not working
- Fields flash red when invalid on blur, flash green when valid on blur


Some another issue I've found:

When creating a new product with the (+) button in the toolbar, with the Action Panel open as if you were editing a variant, you get an error when trying to create a child variant. You have to close the side panel and reopen for it to work. 

This is may be related reactivity as the new selected top level variant is not passed into the Action Panel. (I feel like it's been like this for a while )

zenweasel:
@mikemurray This is working well and ready to merge except for the linting error.

aaronjudd:
@mikemurray it'd be good to get this merged into 1.4. Eta?

PR #2534, opened 2017-07-07
[WIP] Fix inconsistent product grid tag views

### LINKED ISSUE #2213: Product Grid tag views are sometimes inconsistent
### Expected behavior

When viewing a product grid filtered by a `tag`, there will sometimes be products visible in that tag that don't belong.

### Actual Behavior

Sometimes products show up on tag pages that should not be there

With a fresh install
1. Add a new product ( add title, update variants, etc)
2. Publish
3. Edit product and add a new tag
4. Add tag to navbar
5. observe that on shop / <your tag> ( there are products that don't belong )

This doesn't always happen, and with some fiddling, you can fix it by juggling tags on products that are showing up incorrectly.

### Versions

Node: 7.2.0
NPM: 4.0.5
Docker: 17.03.1-ce
Reaction CLI: 0.8.0
Reaction: 1.1.1
Reaction Branch: master

awadesmoline:
This resolves #2213 
### Steps to test
1. Sign in as admin
1. Click on the (+) sign on the toolbar to create a product (edit but don't publish yet)
1. Go to the product grid, and see that the unpublished product is there
1. Click on `Shop` tag and observe that the unpublished product is not there.

### DIFF
--- a/imports/plugins/included/product-variant/containers/productsContainer.js
+++ b/imports/plugins/included/product-variant/containers/productsContainer.js
@@ -127,7 +127,8 @@ function composer(props, onData) {
   const currentTag = ReactionProduct.getTag();
   const productCursor = Products.find({
     ancestors: [],
-    type: { $in: ["simple"] }
+    type: { $in: ["simple"] },
+    isVisible: true
   }, {
     sort: {
       [`positions.${currentTag}.position`]: 1,
--- a/imports/plugins/included/product-variant/containers/productsContainer.js
+++ b/imports/plugins/included/product-variant/containers/productsContainer.js
@@ -124,18 +124,38 @@ function composer(props, onData) {
     window.prerenderReady = true;
   }
 
+  let productCursor;
+
   const currentTag = ReactionProduct.getTag();
-  const productCursor = Products.find({
-    ancestors: [],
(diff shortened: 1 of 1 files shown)


zenweasel:
So I am not seeing the unpublished product under the Shop tag, I am now also not seeing it under an actual tag I created. e.g. I added a new tag to the nav and added a product to it and now I no longer see my product there. However I do see it on `development`

awadesmoline:
@zenweasel I figured this is a bug in itself. When you add a tag to an unpublished product, it doesn't update the product in the database until you publish it. The `products/updateProductTags` method gets called which fails at [this point](https://github.com/reactioncommerce/reaction/blob/development/server/methods/catalog.js#L913) to update the database. I'm not sure how/why this doesn't update the product. When I manually update the products `hashtags` with the tag created on the nav, the product displays under the tag as it is supposed to.

zenweasel:
@awadesmoline I am going to have to disagree since this is how it works on `development` currently.

awadesmoline:
@zenweasel I looked into it further and still couldn't figure out why products don't display correctly. 

I'm pausing on this to continue work on another ticket.

zenweasel:
Pretty sure this isn't the right approach to this problem

PR #2545, opened 2017-07-10
Fixed Failed Transliteration Import showing blank screen on IE11

### LINKED ISSUE #2535: IE11 Blank White Page in Development Branch
**Expected behavior:**
See the Reaction Homepage and everything.

**Actual Behavior:**
Get a blank white page and nothing happens. 
Many debugger errors in console but not sure which ones are the important ones.
( Can't post them now because I can't copy them from a Windows Virtual Machine )

**Steps to Reproduce the Behavior:**
Load Reaction Development Branch in IE11

**Versions**
Node: 7.1.0
NPM: 5.0.4
Meteor Node: 4.8.3
Meteor NPM: 4.6.1
Reaction CLI: 0.10.0
Reaction: 1.3.0
Reaction branch: development

rhenshaw56:
Resolves #2535 
**Testing**
- Pull this branch
- Start the app and run on Internet explorer 11

### DIFF
--- a/lib/api/helpers.js
+++ b/lib/api/helpers.js
@@ -1,5 +1,5 @@
 import url from "url";
-import { slugify } from "transliteration";
+import "transliteration/lib/browser/transliteration.js";
 import { Meteor } from "meteor/meteor";
 import { Router } from "/imports/plugins/core/router/lib";
 import { Shops } from "/lib/collections";
--- a/lib/api/helpers.js
+++ b/lib/api/helpers.js
@@ -1,5 +1,5 @@
 import url from "url";
-import "transliteration/lib/browser/transliteration.js";
+import { slugify } from "transliteration";
 import { Meteor } from "meteor/meteor";
 import { Router } from "/imports/plugins/core/router/lib";
 import { Shops } from "/lib/collections";
--- a/server/api/core/utils.js
+++ b/server/api/core/utils.js
@@ -1,5 +1,5 @@
 // import { slugify } from "transliteration";
-import "transliteration/lib/browser/transliteration.js";
+import { slugify } from "transliteration/lib/browser/transliteration.js";
 
 
 /**
--- a/server/api/core/utils.js
+++ b/server/api/core/utils.js
@@ -1,6 +1,5 @@
 // import { slugify } from "transliteration";
-import { slugify } from "transliteration/lib/browser/transliteration.js";
-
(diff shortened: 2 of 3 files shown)


spencern:
@rhenshaw56 I figured it would be something like this, however you've got some failing tests here that need to be resolved before we can merge.

server/api/core/utils.js
```diff
@@ -1,4 +1,6 @@
-import { slugify } from "transliteration";
+// import { slugify } from "transliteration";
```
zenweasel: if we are going to remove this, let's remove it, not comment it out.

lib/api/helpers.js
```diff
@@ -1,5 +1,5 @@
 import url from "url";
-import { slugify } from "transliteration";
+import "transliteration/lib/browser/transliteration.js";
```
zenweasel: You've introduced a linting error there.

zenweasel: That's also what's causing the tests to fail

zenweasel: I think we might need to do something like this (this is what was there in a previous incarnation)
```
if (Meteor.isServer) {
  import { slugify } from "transliteration";
}

```

zenweasel: None of that is working. It looks like this has fixed the problem on the clients-side but broken it on the server-side.

I'm getting this error on the server-side whenever I start and it's also what is failing the tests
```
  TypeError: _slugify is not a function
      at _getSlug (lib/api/helpers.js:86:27)
      at Object.autoValue (lib/collections/schemas/products.js:467:9)
      at Object.runAV (packages/aldeed_simple-schema.js:1282:26)
      at packages/aldeed_simple-schema.js:1383:13
      at Array.forEach (native)
      at Function._.each._.forEach (packages/underscore.js:139:11)
      at packages/aldeed_simple-schema.js:1382:7
```

adding in any import for the server causes the client-side to fail again. I am convinced that https://github.com/andyhu/transliteration/issues/38 is related. We might want to pull in https://github.com/andyhu/transliteration/pull/44 and see if that fixes it.

zenweasel:
I've tested this using `development` and this version of `transliteration` and that appears to solve the problem.

zenweasel:
@dzcpy Thanks so much for the update. We'll update and give it a try. Would love to be back on the "mainstream" version.

Issue #2547, opened 2017-07-11
UI bug on Safari: navbar links not vertically centered


machikoyasuda:
### Expected behavior

The Shop `tag link` text should be vertically centered relative to the navigation bar.

### Actual Behavior

In Safari, the link is not vertically centered. The issue stems from the parent `div`'s height: the `navbar-item` height should be 100% of the navbar, but that attribute is not being rendered by Safari.

### Steps to Reproduce the Behavior

1. open http://localhost:3000 in Safari and Chrome on desktop
2. compare the navbar on both browsers:

![screen shot 2017-07-11 at 1 55 16 pm](https://user-images.githubusercontent.com/3673236/28090454-2c68ee3c-6641-11e7-92ed-ae4c6c2d0016.png)

### Versions

```
Node: 7.0.0
NPM: 3.10.8
Meteor Node: 4.8.3
Meteor NPM: 4.6.1
Reaction CLI: 0.10.2
Reaction: 1.3.0
Reaction branch: master
```

Browser versions:
Chrome 59 on OS X 10 http://whatsmybrowser.org/b/OKAUJ6M
Safari 10 on OS X 10 http://whatsmybrowser.org/b/KFCF49L

zenweasel:
Closed via #2551

PR #2550, opened 2017-07-11
fixes issue #2547: navbar links are now vertically centered on Safari

machikoyasuda:
- Fixes https://github.com/reactioncommerce/reaction/issues/2547
- Sets `.navbar-item` height to the height of the navbar with the `@navbar-height` variable instead of the percentage. Percentage height breaks in Safari.

## Steps to Reproduce & Test:
- open http://localhost:3000 in Safari and Chrome on desktop
- compare the navbar on both browsers:

screen shot 2017-07-11 at 2 44 12 pm

screen shot 2017-07-11 at 2 48 51 pm

### DIFF
--- a/imports/plugins/included/default-theme/client/styles/tagNav.less
+++ b/imports/plugins/included/default-theme/client/styles/tagNav.less
@@ -294,7 +294,7 @@
 }
 
 .rui.tagnav .navbar-item {
-  height: 100%;
+  height: @navbar-height;
 }
 
 .rui.tagnav.vertical .navbar-item {


zenweasel:
@machikoyasuda Thanks for taking the initiative to fix this. I had noticed this before but didn't get around to making an issue, much less fixing it.

Couple of notes around process. You will normally just want to make your PR's directly from the Reaction repo rather than a fork. This makes is easier for us to update, etc. Secondly you should normally choose a reviewer (whomever you think would be the most relevant) so that person can test/review and merge. If the PR is large or complicated you can choose more than one or if you just want to get feedback from more than one person.

PR #2555, opened 2017-07-13
Fix You may not observe cursor with fields _id: 0

spencern:
Fixes console error
```
Error: You may not observe a cursor with {fields: {_id: 0}}
```
Not really sure when it got introduced, perhaps in Meteor `1.5.1rc5`?

@erik and I had both seen this issue intermittently

Not really sure why we're trying to suppress the `_id`s for Translations anyway, but if there's a good reason to do that, please let me know.

### DIFF
--- a/client/modules/i18n/startup.js
+++ b/client/modules/i18n/startup.js
@@ -51,11 +51,7 @@ Meteor.startup(() => {
       //
       return Meteor.subscribe("Translations", language, () => {
         // fetch reaction translations
-        const translations = Translations.find({}, {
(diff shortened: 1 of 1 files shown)


zenweasel:
I'm seeing this error after switching to 1.5.1

spencern:
@zenweasel does this PR fix it?

aaronjudd:
I'd guess the logic here was just to reduce the overhead -> there's a lot of translation info, no need for all of it, when we just needed the changed id.

PR #2563, opened 2017-07-17
Don't add a prefix to Paypal return URL when there is a single shop

zenweasel:
The Paypal plugin was adding the shop slug to the return url even when we only had one shop, where we shouldn't be using the shop prefix. Changed the behavior of `getShopPrefix` slightly but tested the other places it is used and seems to be fine.

### Steps to test
1. Place an order via Paypal Express with default data
1. Use a `Shops.json` with only one store
1. Place an order via Paypal Express and confirm that it completes.


zenweasel:
@spencern If you get a change could you take a quick look at this?

spencern:
Checking it out now, sorry didn't see I was on it before!

PR #2565, opened 2017-07-17
Longer timeout for circleci to run build.sh

lcampanis:
Currently CircleCI will timeout after 10mins of waiting for `build.sh` to run.
```bash
[-] Building Meteor application...

Even with METEOR_ALLOW_SUPERUSER or --allow-superuser, permissions in your app
directory will be incorrect if you ever attempt to perform any Meteor tasks as
a normal user. If you need to fix your permissions, run the following command
from the root of your project:

  sudo chown -Rh <username> .meteor/local

artlimes:spiderable: updating npm dependencies -- chrome-remote-interface...
cfs:tempstore: updating npm dependencies -- combined-stream...
cfs:gridfs: updating npm dependencies -- mongodb, gridfs-stream...
verso:intercom: updating npm dependencies -- intercom-client...
Too long with no output (exceeded 10m0s)
```

The following fix allows for 30mins before CI times out.


jshimko:
Have you confirmed that this works with a full build?  According to their docs, you're supposed to specify a unit (seconds, minutes, hours, etc.) for the `no_output_timeout` config key.

https://circleci.com/docs/2.0/configuration-reference/#run

> `no_output_timeout:`  Elapsed time the command can run without output. The string is a decimal with unit suffix, such as “20m”, “1.25h”, “5s” (default: 10 minutes)

FYI, you can trigger a full Docker build on Circle by pushing a branch with `docker` anywhere in the name.

Regardless of whether this works as-is, I'm thinking it would be worth updating with a unit so there's no confusion about what the unit is.  Obviously, most people will assume seconds because of the context, but for example, one easy assumption to make would be milliseconds since it's the standard in JS.  I'm thinking we just change it to `30m` so there's less of a chance of confusion.

jshimko:
I updated it.  Thanks for taking the time to submit a PR!

lcampanis:
Noted thanks!

PR #2558, opened 2017-07-14
Release 1.4.0

aaronjudd:
Release 1.4.0

##  Changes / Fixes

- __Meteor 1.5.1__
- __Adds inline validation to variant &amp; child variant form (#2531)__
- __Longer timeout for circleci to run build.sh (#2565)__
- __Translations updated based on development (#2566)__
- __Fix You may not observe cursor with fields id: 0 (#2555)__
- __Fix LoadMoreProducts button not working as it should (#2554)__
- __Provide sort parameter to Products subscription and remove duplicate sorting of products (#2552)__
- __Fix issue 2535. Transliteration in Safari/IE (#2553)__
- __Fix 2547: vertically-center navbar link using @navbar-height height variable instead of percentage height (#2551)__
- __Update eslint to v4 (#2542)__
- __Fix order action button reverting status to processing (#2541)__
- __Update Docker base to v2.0.2 (#2537)__
- __Fix Exception in delivering result of invoking &#39;email/verifySettings&#39; (#2512)__
- __Fix unable to open fulfill order action view (#2530)__
- __Fix for issue #1993 (#2518)__
- __Fix for no scrollbar in Dashboard (#2511)__
- __Convert product grid to React: Milestone 57 (#2402)__
- __Fix Errors importing exported product data  (#2513)__
- __Fix send sms method to read settings of shop and buyer userId (#2508)__
- __Fix issue dragging images in PDP when settings drawer is open (#2506)__
- __Add Tunisian Dinar currency (#2517)__
- __Default to using shop country code or GeoIP result for new addresses (#2515)__
- __Fix missing reference (#2507)__
- __Show proper error messages when Stripe payments fail (#2499)__
- __Fix linting problems introduced in new SortableTable (#2505)__
- __New SortableTable component to replace MeteorGriddle (#2416)__
- __Fix ESLint4 errors correct eslint config errors (#2450)__
- __Remove redundant risul:moment-timezone (#2491)__
- __Fix hard refresh on clicking tag link on product detail page (#2489)__
- __Fixed product images not displaying on order completed page (#2490)__
- __Updated timeout time. (#2492)__
- __Fix error when inviting already invited user (#2433)__
- __Rename emailTableColumn -> shippoTableColumn (#2564)__
- __LingoHub based on development (#2566)__
- __Longer timeout for circleci to run build.sh (#2565)__
- __Adds inline validation to variant & child variant form (#2531)__

## Thanks To Our Community Contributors
**Changes and Fixes**
@prinzdezibel 
[...]

PR #2567, opened 2017-07-17
Release 1.4.0 (#2558)

spencern:
Merge 1.4.0 into development


zenweasel:
I think I speak for everyone when I say, WHY? (why is this diff so large?)

spencern:
@zenweasel I had that same thought. Should this be done by just merging locally and pushing?

zenweasel:
I _think_ you end up with the same result? I think it's just this diff display that's messed up. I think that's what we normally do though?

spencern:
Closed in favor of doing it locally.

PR #2524, opened 2017-07-05
Split Orders and Shipping

spencern:
This PR is experimental and currently WIP

The goal of this PR is to setup the schema and any methods necessary for having multiple shipments and multiple payments for a single order..

PR #2577, opened 2017-07-20
Marketplace settings, publications, subscriptions, and routing

spencern:
*   New registry `provides` option for `"shopSettings"` which will insert a template into the shop settings panel
*   providesShopSettings panels now collapsable and will display icon from registry

*   New settings marketplace settings, publications, and schemas for marketplace settings
    *   marketplaceNakedRoute (not implemented yet)
    *   merchantTheme (false)
    *   merchantCart (false)
    *   merchantLocale (false)
    *   merchantShippingRates (false)
    *   allowMerchantSignup (false)
    *   merchantFulfillment (true)
    *   enabledShopTypes (merchant, affiliate)
    *   enabledPackagesByShopTypes (packages defined for merchant and affiliate)
    *  

*   Cache marketplace settings on the client
    

*   Add PrimaryShop and MerchantShops pubs/subs. 
    *   Remove Shops pub/sub & SellerShops pub/sub

*   Replace getShopId with getPrimaryShopId or conditionally determine which shopId to use in
    *   email config
    *   ReactionLayouts
    *   language, currency, locale
    *   brand
    *   cart
    *   revision control
    *   templates

*   Removes default 2nd shop for marketplace			
*   Disable packages when new shop is created based on shop type
*   New methods for interacting with the marketplace and primary shop

*   add guards for getShopPrefix and getLogo on client			

*   Newly created shops can't be shopType "primary"

*   Flag merchantLanguage and merchantLocale for removal			

*   Publish tags for both the active shop and the primary shop			

*   Deprecate and replace getCurrentShop and getCurrentShopCursor			
*   Add getPrimaryShopId to /lib/api			

*   Fixes for tests based on marketplace updates
    *   stub getPrimaryShopId


spencern:
This is the other marketplace feature branch that is pretty much ready and will add a lot of value. 
I was considering rolling the new stripe-connect stuff into this, but I think if we merge this and #2524 into marketplace, then it will be better to just rebuild the stripe connect stuff at the base.

PR #2584, opened 2017-07-22
Reaction Shopify connector

spencern:
This PR does two major things:
1. Provides a new dashboard settings panel for "Connectors"

Adding a template to the connectors panel is as simple as `provides: connectorSettings`
For example a registry entry that looks like this will add the provided template (`shopifyConnectSettings` in this case) to the Connectors panel in the dashboard.
```
  registry: [{
    label: "Shopify Connect Settings",
    name: "settings/connectors/shopify",
    icon: "fa fa-exchange",
    route: "/dashboard/connectors/shopify",
    provides: "connectorSettings",
    container: "dashboard",
    template: "shopifyConnectSettings"
  }]
```

A user has to have one of these roles to access: `owner`, `admin`, or `connectors`

2. Provides a new Shopify connector which can import products from a specific shop.


spencern:
Will merge updated marketplace with this once it's been merged with 1.5 and then it should be ready to go.

PR #2588, opened 2017-07-24
Marketplace, Shop, and Affiliate shop owners should be able to add administrators via dashboard

### LINKED ISSUE #2187: Marketplace, Shop, and Affiliate shop owners should be able to add administrators
**Update June 29**
This issue is focused on inviting new admins to existing shops.

5 thoughts from a chat @impactmass and I had

1. This invitation will be sent via the “accounts” dashboard (or the new dashboard that’s created to manage shop users).
2. An email is generated and sent to the user.
3. I think the flow might be:
  - Invitation is created
  - New user is generated with existing permissions and with the email specified in the invitation
  - User clicks from email to join -> is logged in via token -> lands on password set/reset
  - User sets new password -> routed to user profile

4. The invitation form will need a dropdown (I think @rymorgan is working on this/has finished this) to select the correct group to invite the new user to
5. ‘Appropriate group permissions’ means permissions for the group the user was invited to.
6. If the user already exists (as a customer or other shop admin), they should be given permissions for the site and sent an email, but an additional user should not be created.

**Original Issue**
Marketplace, Shop, and Affiliate shop owners should be able to invite people to become administrators. 

Newly invited administrators should be granted appropriate permissions based on the marketplace, shop, or affiliate user that invited them. E.g. If a shop owner invites a new admin, that admin should be granted admin permissions for that shop only.

If adding an user or email that is already connected to a Reaction account on this app, we should grant permissions to the existing user.

If we are adding a user or email that does not exist yet, we should grant the new account permissions upon activation.

impactmass:
Resolves #2187 (and also includes updates for #2184)

In detail, this PR contains:
* UI implementation for #2187:
_ Adds Groups dropdown to `AdminInviteForm`.  
_ Thus enabling inviting an admin directly as part of a permission group
_ The invite form picks up the defaultInviteGroup set [here](https://github.com/reactioncommerce/reaction/blob/seun-update-user-group-2184/imports/plugins/core/accounts/client/components/manageGroups.js#L28-L41). Basically any group available other than the "owner" group
_ The invited admin gets the permissions defined in the group invited to

* Updates for #2184:
_ The "remove" button removes a user from a group and adds them to customer group, instead being "groupless"
_ When a new group is created, default customer permissions + "dashboard" is used the base permissions.


impactmass:
Adding back a WIP on this PR since it will need to be updated after dev goes into marketplace. As soon as dev<>marketplace merge is complete, I'll update.

impactmass:
I've updated this with marketplace branch. it can be reviewed now.

server/methods/accounts/accounts.app-test.js
```diff
@@ -495,29 +495,32 @@ describe("Account Meteor method ", function () {
...
     it("should let a Owner invite a user to the shop", function (done) {
       this.timeout(20000);
       this.retries(3);
       sandbox.stub(Reaction, "hasPermission", () => true);
-      // TODO checking this is failing, even though we can see it happening in the log.
-      // spyOn(Email, "send");
-      expect(function () {
-        return Meteor.call("accounts/inviteShopMember",
+      // TODO: Need to udpate this test to properly check the account created
+      // there's currently an error with Media branding asset when trying to do that
```
spencern: Not a blocker, but let's chat about why you think this is happening. Maybe next week sometime we can pair on tests together for a bit and see if we can knock this out.

impactmass: Thanks. I'll note that so we check on it after

impactmass:
Re: the dropdown not showing until a group is created: I also thought about it a bit. An option is showing the dropdown box that shows a message about not having groups.... but it wasn't looking good UI wise too. 

Since the PR for adding the Shop Manger default group was not far off (currently in WIP ... possibly getting merged today) I went with that.

spencern:
Yeah, I think that was the right call.

Issue #2593, opened 2017-07-25
i18n and currency elements not all updating on change


aaronjudd:
Change currency or i18n in a page where there are multiple localized elements, and you will see many of the components are not update on language / currency changes.  
Behavior was introduced in #2536.

<img width="1193" alt="reaction" src="https://user-images.githubusercontent.com/439959/28579879-c2fc3364-7112-11e7-9251-e267477da1ed.png">
<img width="920" alt="basic_reaction_product" src="https://user-images.githubusercontent.com/439959/28579966-f789ecc0-7112-11e7-982c-62f195bc5708.png">

aaronjudd:
Verified not in `master` -> discussed with @mikemurray @kieckhafer as well in components meeting.

aaronjudd:
@mikemurray @jshimko assigning this to you guys, whomever gets there first.

zenweasel:
Closed via #2742

PR #2599, opened 2017-07-27
[WIP] Merge development marketplace two

spencern:
Merge development / 1.5 into marketplace take 2.

Tests pass and the app loads, but we've got some errors in certain areas..

The new accounts / groups admin section seems to be the main area throwing errors, but the analytics section seems to have been reset to the old panel as well, so there may be other problems.


impactmass:
From what I've observed re: the last commit about the warning, the lint rule at play (https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-multi-comp.md) enforces one component per file (even for func style stateless component. So it's treating the `Cell` field in that particular object as a new comp declaration. 

I've found that when the component is changed from func style to Class the warning stops. I can do that, or another option is to update/extend the rule with  `{ "ignoreStateless": <boolean> }`.

aaronjudd:
I think this should be merged, I see two main issues, there's some UI issues in the account login and the accounts @impactmass does not load.

spencern:
@jshimko and @impactmass are working through a few of the merge related issues and then will merge it after they are done cleaning it up

jshimko:
The issues mentioned by @aaronjudd should be fixed now.  And there's a decent possibility that I fixed most of #2593 because I removed all of the `PureComponent` instances that were blocking re-renders of components when the language changed.

spencern:
Tested this again after the merges and groups works now as it did before. With @aaronjudd's approval above I'm gonna go ahead and merge this so we can get the rest of our outstanding marketplace branches updated and merged

PR #2602, opened 2017-07-27
Add "Shop Manager" as a default group

impactmass:
NB: To be merged after #2601  

Add "Shop Manager" as a default group to the currently existing Owner, Customer and Guest groups. 
This makes two admin groups: Owner and Shop Manager

To Test:
* Start the app, and check that there's a default group for "Shop Manager"

Other UIX updates in this PR:
* Sort groups to display higher permission groups at the top. With this "Owner" group stay at the top
* Use first part of email, if account has no name
* Show no dropdown popover if there's no groups in the list


spencern:
Works. Merging.

@impactmass We should probably make the "enter" button submit the form during the design updates.

PR #2609, opened 2017-07-31
[Marketplace] Owner permission should be transferable #2188

### LINKED ISSUE #2188: [Marketplace] Owner permission should be transferable
Marketplace, Shop, and Affiliate shop owners should be able to transfer ownership to another administrator of their marketplace or shop.

#### Rules:
-   Should not be able to remove permissions from the owner except by transferring the owner title. -   Each shop should only be able to have one owner.
-   If there are no other administrators in a given shop, owner can not be transferred.
-   Process for transferring ownership to a non admin user should be: Add user to admin group, then transfer ownership, then new owner can remove previous owner from admin group.

impactmass:
Resolves #2188 (transfer of ownership of a Shop or Marketplace to another administrator)

**How to test:**
* Transfering Marketplace Owner
_ In the default (primary) shop, log in as default admin and invite a user as shop admin (e.g Shop Manager).
_ On the newly invited user, click the group dropdown, and select "Owner"
_ Confirm that there's a pop up warning about the intended change 
_ On proceeding, the current owner should be replaced with the new owner. 
_ NB: The old owner (who is currently logged in) losses the right to change the owner back (since she has transferred ownership)
_ Log in as the new user that the ownership was transferred to (in new browser) 
_ Go to the admin panel for that user, and confirm that you can transfer the ownership back

* Test for another shop
_ Create a new shop. (e.g using the "Invite Owner Form" from the Marketplace tab in Shop settings)
_ Log in as the owner of the shop, and run through a similar process like above


spencern:
👌 http://recordit.co/1bRTX8YIIi

PR #2610, opened 2017-08-01
Update cart schema to include all product data

kieckhafer:
Update the cart schema, and `addToCart` function to include all data we store on the product.


zenweasel:
Looks like this is breaking a lot of tests

kieckhafer:
If you're wondering what `getAutoValues` is, this is the reference we used to figure out that it's a good fix: https://github.com/aldeed/meteor-collection2/issues/245

Issue #1278, opened 2016-08-10
Challenges overriding cart workflow.


spencern:
### Expected behavior

I'd like to be able to use non-default templates for the checkout workflow without having to rewrite or remove existing core templates such as `checkoutLogin` or `checkoutAddressBook`. We've written our own namespaced templates (e.g. `goCheckoutLogin`) that we'd like to be able to use instead of the core templates.

I'd also like to change the order of the cart workflow by adding additional workflow items. For example we have a workflow step `0` that we've added that permits customers to review the items in their cart before continuing with the checkout process. We also have a workout step `4.5` that requires users to agree to our Terms of Service before continuing on to entering their payment information. Currently we've had to hack these additional steps on to existing steps, but ideally they should be their own workflow steps.

I've been hacking at this for a day and a half now and have determined now that there is no way forward that doesn't involve changing core templates significantly. There are some simple changes we can make that will solve some of issues that make overriding these templates difficult (see: #1269) , but there are other bigger problems that I didn't see until getting deeper that seem to require significant changes to solve.

We've been able to override the existing templates to an extent with the method described in #1269, but in order to change the order, insert new steps, remove existing steps or otherwise modify an existing workflow, it seems there will need to be significant changes to core.
#### Significant Challenges
1. Core cart methods have `coreCartWorkflow` steps built in. [mergeCart](https://github.com/reactioncommerce/reaction/blob/development/server/methods/core/cart.js#L212-L215),
   [copyCartToOrder](https://github.com/reactioncommerce/reaction/blob/development/server/methods/core/cart.js#L618-L624),
   and [setShipmentMethod](https://github.com/reactioncommerce/reaction/blob/development/server/methods/core/cart.js#L700-L701) all call `workflow/pushCartWorkflow` and have existing - core - workflow steps built in.
2. Workflow steps are directly tied to their template. In fact, as far as I can tell, the workflow template _is_ the workflow step. There doesn't seem to be any way to decouple a core template from a workflow step. If there was a way to overwrite named workflow steps with alternative templates, perhaps we could build around the limitation in step one.
#### Solutions?

I'm not really sure what the best way to solve these problems, but I doubt our organization is the only one that will be interested in changing these workflows.

It seems like the first step might be to decouple workflows from the templates that they serve, but I'd imagine that's a fairly complicated process.

[...]

PR #2613, opened 2017-08-01
[Marketplace] Set order(s) to a specific status

### LINKED ISSUE #2341: Order managers should be able to set an order or multiple orders to a specific status
Users with permission to view and process orders should be able to update the status of multiple orders in bulk.

If any orders are selected and bulk set to a state that is not the next status in the order flow, there should be a confirmation such as:

```
You've requested that 17 orders be set to the "Shipped" status, but 4 of these orders are not in the "Packed" state and would skip all steps leading up to the "Shipped" state. Are you sure you want to do this?
[Yes, Set All Selected Orders] [No, Cancel] [Set All Packed Orders]
```

While setting the state of the selected orders, the orders should have an event added to their history showing that they have been set into a specific state by a bulk operation.

See: #2325 for how to select orders for bulk processing.

kieha:
Resolves #2341 

This allows users with permission to view and process orders to be able to set/update the status of multiple orders in bulk.

#### How to test

-  Have at least 10 (preferably new) orders made in your shop
- Visit the orders page on the dashboard
- Select one order/many orders (hover over the avatar: it rolls over to a check box)

##### Testing `Bulk Actions`

- On the bar that appears at the top, click on the `Bulk Actions` button
- Select the list items in order of how they appear, and you should see the labels in the `Shipping` column change according to what you select
- Select another order/set of orders and randomly choose any order of list item (e.g. select the `Labeled` option on a new order). You should see a modal that asks if you're sure you want to skip the shipping flow

##### Testing `Capture`

- On the bar that appears at the top, click on the `Capture` button
- Observe the waiting state on the button, then the modal that shows up on success of the action
- Open the sidebar for the item selected, scroll to the `Invoice` section and observe that the payment has been captured

P.S. The `Print Invoice and Pick Sheet` step that appears on the list under `Bulk Actions` is not functional yet in this PR.


aaronjudd:
Some UI issues in opening the order details.
reaction

server/methods/core/orders.js
```diff
@@ -112,6 +165,57 @@ export const methods = {
...
+  "orders/shipmentLabeled": function (order, shipment, labeled) {
+    check(order, Object);
+    check(shipment, Object);
+    check(labeled, Boolean);
+
+    if (!Reaction.hasPermission("orders")) {
+      throw new Meteor.Error(403, "Access Denied");
+    }
+
+    if (order) {
```
zenweasel: Same comments as above

server/methods/core/orders.js
```diff
@@ -62,6 +62,56 @@ export function ordersInventoryAdjust(orderId) {
  */
 export const methods = {
   /**
+   * orders/shipmentPicked
```
zenweasel: Not specific to this PR but this file is already >1000 LOC. I wonder if we should start breaking these down?

zenweasel:
Is "Capture" supposed to be working here, because it's not.

zenweasel:
Once I have set a couple of orders to picked, I can no longer select "Picked" again because the status never clears. (in this screenshot I picked two orders successfully then selected two other orders and see that I cannot select Picked any longer)

cart_completed

zenweasel:
Just a nitpick (and may be a design issue) but the status says "Generate Picking Report" but there is no reporting being generated, just changing the status.

zenweasel:
I changed an order from New to Picked (and it properly showed me "1 Order picked") but then the Bulk Actions window shows me the status from the last operation I did (move orders to Shipped).

cart_completed

zenweasel:
When I select all orders and move them to "Picked" where some orders are moving backwards in status (from Shipped to Picked) I am not getting the warning telling me I am moving them into a backwards status. Shouldn't it be warning me?

zenweasel:
I'm attempting to move an order from Shipped to Picked but it's telling me the order is already Picked even though you can see in the screenshot that it's status is "Shipped".

cart_completed

zenweasel:
When I select an order the pagination controls disappear. Am I not allowed to select orders across pages?

zenweasel:
When I select an order that is already in a certain state and try to set again I get the message "Order is already in Picked status" (or the appropriate) except for Shipped. It will let me set the status to Shipped over and over again.

zenweasel:
That's a few notes but generally this works well. Just need to make the UI a little more consistent.

imports/plugins/core/orders/client/containers/ordersListContainer.js
```diff
@@ -140,30 +137,263 @@ class OrdersListContainer extends Component {
...
+   * @param {String} status - the shipping status to be set
+   * @return {null} no return value
+   */
+  shippingStatusUpdateCall = (selectedOrders, status) => {
+    let orderText;
+
+    if (selectedOrders.length > 1) {
+      orderText = "orders";
+    } else {
+      orderText = "order";
```
kieha: I'm passing this variable to `i18next.t` below to be translated depending on context.

imports/plugins/core/orders/client/containers/ordersListContainer.js
```diff
@@ -1,16 +1,14 @@
...
+import { Loading } from "@reactioncommerce/reaction-ui";
+import { Orders, Media } from "/lib/collections";
+import { i18next, Reaction } from "/client/api";
 import OrdersList from "../components/orderList.js";
-import {
-  PACKAGE_NAME,
-  ORDER_LIST_FILTERS_PREFERENCE_NAME,
-  ORDER_LIST_SELECTED_ORDER_PREFERENCE_NAME
-} from "../../lib/constants";
+import { PACKAGE_NAME, ORDER_LIST_FILTERS_PREFERENCE_NAME, ORDER_LIST_SELECTED_ORDER_PREFERENCE_NAME, shippingStates } from "../../lib/constants";
```
zenweasel: Line too long. Maximum length is 120 chars (so it fits in GH code review screens)

imports/plugins/core/orders/client/containers/ordersListContainer.js
```diff
@@ -28,26 +26,40 @@ class OrdersListContainer extends Component {
...
     });
   }
 
   handleSelect = (event, isInputChecked, name) => {
     this.setState({
-      multipleSelect: false
+      multipleSelect: false,
+      renderFlowList: false
+    });
+    shippingStrings.forEach((string) => {
```
zenweasel: Not sure settings the variable name to be the name of a type is probably a good idea.

imports/plugins/core/orders/client/containers/ordersListContainer.js
```diff
@@ -140,30 +162,347 @@ class OrdersListContainer extends Component {
...
+      this.shippingStatusUpdateCall(selectedOrders, status);
+      // display alert if order(s) are already in this state
+    } else if (!falsePreviousStatuses && !falseCurrentState && trueCurrentState) {
+      Alerts.alert({
+        text: i18next.t("order.orderAlreadyInState", { orderText: this.displayOrderText(selectedOrders), status: status })
+      });
+    }
+  }
+
+  displayRegressionAlert = (selectedOrders, ordersToRegress, status, whichFalseState, falsePreviousStatuses, falseCurrentState, trueCurrentState) => {
```
zenweasel: There's a bunch of places where the line is too long. Your editor should be restricting you to 120 chars.

imports/plugins/core/orders/client/containers/ordersListContainer.js
```diff
@@ -140,30 +162,347 @@ class OrdersListContainer extends Component {
...
+   */
+  setShippingStatus = (status, selectedOrdersIds) => {
+    this.setState({
+      renderFlowList: true
+    });
+    const selectedOrders = this.state.orders.filter((order) => {
+      return selectedOrdersIds.includes(order._id);
+    });
+
+    if (status === "picked") {
```
zenweasel: I wonder if there was a way to make this dynamic so we didn't need a conditional for each type which feels brittle. In the interest of getting this done I would be okay with the implementation as it is though.

imports/plugins/core/ui/client/components/list/list.js
```diff
@@ -3,12 +3,12 @@ import PropTypes from "prop-types";
 import classnames from "classnames";
 import { registerComponent } from "@reactioncommerce/reaction-components";
 
-const List = ({ children, isAdmin }) => {
+const List = ({ children, isAdmin, className }) => {
```
zenweasel: Do we need to add this to the docs?

zenweasel:
Based on the design spec it looks to me like "Capture" should work here.

zenweasel:
Something about having to close that status panel every time is super annoying but I don't know what else you could do about it.

Otherwise I think this works pretty well. I'd like to look at making Approve and Capture work though because I think that's a pretty common use case.

imports/plugins/core/orders/client/containers/ordersListContainer.js
kieha: Yeah, my bad. my editor's line length max is 160. I'll change that.

spencern:
On a UI/UX note, I think the way that inbox handles the visibility of the checkbox makes a lot of sense.

Before anything has been clicked:
image

After at least one is selected, we're aware of the users intent (to select at least one order), and in inbox they toggle the images to checkboxes without enforcing the hover to make it easier to select multiple.
image

@rymorgan curious what you think about this. I think it would make selecting multiple easier and wouldn't be a drag on the UI when not selecting orders.

imports/plugins/core/orders/client/components/orderBulkActionsBar.js
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    setListItemsToDefault() {
+      if (this.state.renderFlowList === false) {
+        this.setState({
+          shipping: {
+            picked: false,
+            packed: false,
+            labeled: false,
+            shipped: false
```
spencern: How does this work if there is a shipping workflow that has additional or fewer steps?

imports/plugins/core/orders/client/templates/workflow/shippingTracking.js
```diff
@@ -77,7 +77,7 @@ Template.coreOrderShippingTracking.events({
   "click [data-event-action=shipmentPacked]": () => {
     const template = Template.instance();
 
-    Meteor.call("orders/shipmentPacked", template.order, template.order.shipping[0], true);
+    Meteor.call("orders/shipmentPacked", template.order, template.order.shipping[0]);
```
spencern: We can't assume that there is only one shipment for any given order. 

`template.order.shipping[0]` is too brittle for multiple shipments per order and in a marketplace setting we have to support multiple shipments per order.

In general, hard-coding the 0th index for an array is a code smell, especially as we continue to build multi-shop support

imports/plugins/core/orders/client/templates/workflow/shippingTracking.js
```diff
@@ -193,6 +193,7 @@ Template.coreOrderShippingTracking.helpers({
     const order = Template.instance().order;
     const shipment = order.shipping[0];
 
-    return shipment.packed && shipment.tracking;
+    return _.includes(shipment.workflow.workflow, "coreOrderWorkflow/packed") && shipment.tracking ||
+    _.includes(shipment.workflow.workflow, "coreOrderWorkflow/packed");
```
spencern: this feels brittle as well. What if there is a custom workflow that doesn't include `coreOrderWorkflow/packed` (e.g. uses a different workflow, or pre-packed items, etc.)

imports/plugins/core/orders/lib/constants.js
```diff
@@ -37,3 +37,15 @@ export const orderFilters = [{
...
+/**
+ * Order shipping states
+ * @type {String}
+ */
+export const shippingStates = {
+  picked: "Picked",
+  packed: "Packed",
+  labeled: "Labeled",
+  shipped: "Shipped",
+  delivered: "Delivered"
```
spencern: How would this be overridden or extended?

server/methods/core/cart.js
```diff
@@ -671,6 +671,11 @@ Meteor.methods({
     order.shipping[0].items.shipped = false;
     order.shipping[0].items.delivered = false;
 
+    // begin a new shipping workflow for the order
+    order.shipping[0].workflow = {
```
spencern: We can't assume that this shipment is the only shipment.

imports/plugins/core/ui/client/components/list/list.js
kieha: Done. PR to docs opened here https://github.com/reactioncommerce/reaction-docs/pull/256

server/methods/core/cart.js
```diff
@@ -671,6 +671,11 @@ Meteor.methods({
...
+      status: "new",
+      workflow: ["coreOrderWorkflow/notStarted"]
+    };
     order.billing[0].currency.exchangeRate = exchangeRate;
```
spencern: This issue seems like it is older than this particular PR because this code is older, but needs to get solved ASAP.

Can't assume that this is the only payment in this order, or that the 0th payment corresponds with this shipment.

I'll repeat myself a few more times: hardcoded 0th indexes are an indicator that code is not ready for marketplace.

Hard coded indexes, 0th especially, in general are a code smell. It's a sign that that the data structure doesn't fit the application, or that you don't have a better way to identify elements in the array other than just arbitrarily taking the first one.

server/methods/core/orders.js
```diff
@@ -62,53 +62,115 @@ export function ordersInventoryAdjust(orderId) {
...
+   *
+   * @summary update picking status
+   * @param {Object} order - order object
+   * @param {Object} shipment - shipment object
+   * @return {Object} return workflow result
+   */
+  "orders/shipmentPicked": function (order, shipment) {
```
spencern: This feels brittle to me as well.

I'm always open to new ways to do things, but it seems like we should have "orders/progressStatus" and "orders/regressStatus" methods instead of one method per status.

We could then have a set of workflows that identifies the next or previous status from the existing status and proceeds. There could be different states for items than for orders, and each should be overridable. 

This seems like it would make it easier to extend or replace functionality because if you replace the workflow set, then you replace the way that orders progress.

server/methods/core/orders.js
```diff
@@ -62,53 +62,115 @@ export function ordersInventoryAdjust(orderId) {
...
    * orders/shipmentPacked
    *
    * @summary update packing status
    * @param {Object} order - order object
    * @param {Object} shipment - shipment object
-   * @param {Boolean} packed - packed status
    * @return {Object} return workflow result
    */
-  "orders/shipmentPacked": function (order, shipment, packed) {
+  "orders/shipmentPacked": function (order, shipment) {
```
spencern: See comments on unifying status progression methods in the shipmentPicked comment.

server/methods/core/orders.js
```diff
@@ -62,53 +62,115 @@ export function ordersInventoryAdjust(orderId) {
...
-      });
+  /**
+   * orders/shipmentLabeled
+   *
+   * @summary update labeling status
+   * @param {Object} order - order object
+   * @param {Object} shipment - shipment object
+   * @return {Object} return workflow result
+   */
+  "orders/shipmentLabeled": function (order, shipment) {
```
spencern: See comments on unifying status progression methods in the shipmentPicked comment.

rymorgan:
@spencern Yes, totally agree, once you check one box, all the boxes should appear. That makes sense. Will update the design specs.

kieha:
@spencern I agree that the implementation of most of this (the 0th index hard coding) isn't marketplace ideal, and needs to be refined. I also agree with rethinking the order fulfillment workflow to be more flexible and extensible, and I'll look into it, with a bit of direction.

spencern:
@kieha I don't mind guiding or helping with the direction for any of this, let me know if you would like to pair on it sometime.

spencern:
@kieha - here's a ticket that is related
https://github.com/reactioncommerce/reaction/issues/1278

rymorgan:
@spencern @kieha I've updated the design to reflect the functionality that once you check on check all the rows show as checks. 

https://zpl.io/Nbrejea

spencern:
Merging, additional work will be performed in #2724

PR #2624, opened 2017-08-02
Ignore C9 project files.

mikeumus:
Git Ignore files from the Cloud 9 IDE, by Amazon. 

https://c9.io


.gitignore
```diff
@@ -2,6 +2,7 @@
 .fileStorage/
 .vscode
 .idea
+.c9/
```
zenweasel: Actually, is there a reason for that trailing slash? The other directory-wide ignores don't have that.

mikeumus: Tested and works with just `.c9`. Fix in c1d450b

Issue #2620, opened 2017-08-02
[Marketplace] Receive error when trying to archive product


zenweasel:
### Expected behavior
Should be able to archive a product w/o an error

### Actual Behavior
Receive an error on the server-side:
```
Exception while invoking method 'products/archiveProduct' Error: Mongo selector can't be an array.
    at Function.Mongo.Collection._rewriteSelector (packages/mongo/collection.js:377:11)
    at new CursorDescription (packages/mongo/mongo_driver.js:881:36)
    at [object Object].MongoConnection.find (packages/mongo/mongo_driver.js:807:11)
    at [object Object].MongoConnection.findOne (packages/mongo/mongo_driver.js:818:15)
    at Object.<anonymous> (packages/matb33_collection-hooks.js:819:16)
    at Object.collection.(anonymous function) [as findOne] (packages/matb33_collection-hooks.js:146:21)
    at [object Object]._.extend.findOne (packages/mongo/collection.js:334:29)
    at [object Object].productsArchiveProduct (server/methods/catalog.js:765:30)
    at packages/check.js:129:16
    at [object Object]._.extend.withValue (packages/meteor.js:1126:17)
    at Object.exports.Match._failIfArgumentsAreNotAllChecked (packages/check.js:128:41)
    at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1765:18)
    at packages/ddp-server/livedata_server.js:719:19
    at [object Object]._.extend.withValue (packages/meteor.js:1126:17)
    at packages/ddp-server/livedata_server.js:717:46
    at [object Object]._.extend.withValue (packages/meteor.js:1126:17)
```

### Steps to Reproduce the Behavior
I verified that this doesn't happen on `development`

1. Login as admin
1. Click on default product
1. Select "Clone product"
1. Select newly created product
1. Click on archive
1. Observe error on client and server

### Versions
```
Node: 7.10.0
NPM: 5.0.3
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.0-ce
Commit: b6e18a4b5a35b852bd75648ae395a2f4b77eae81
```

zenweasel:
Closed via #2650

PR #2650, opened 2017-08-08
Fix for error while archiving a product

foladipo:
This PR resolves [#2620](https://github.com/reactioncommerce/reaction/issues/2620). The said issue was a bug report for an error that occurred whenever a user tried to archive a product. These trials consequently failed. And it was all due to the wrong use of a certain function's parameter.

## Testing steps

1. Login as admin.
1. Click *once* on the default product (i.e `Basic Reaction Product`) or another product (if you've added any).
1. Select "Clone product" in the side menu.
1. *Double-click* the newly created product to open it (scroll the page to find it if necessary).
1. Click on "Archive" in the top menu.
1. The archiving operation should now be completed without any errors. Similarly, check the terminal/command prompt and the error reported in [#2620](https://github.com/reactioncommerce/reaction/issues/2620) should be absent.


zenweasel:
@foladipo a couple of notes 

1. Generally we want the PR title to be self-explanatory so whenever possible make it "Fix for error when archiving product" rather than just "Fix for issue".
1. You don't need to include things like "open your terminal prompt" in testing instructions. Basically your testing instructions could start with step 4. There are certainly worse things than being over-explicit but probably better to keep them shorter.
1. If this PR is ready you will want to assign a reviewer (for now, me) and also assign the PR to me. I will get a notification that you have requested a review and I can take a look at it.

foladipo:
Thanks for your feedback @zenweasel . I've resolved the first two points you raised. I will add you as a reviewer (and thus make you get a notification) as soon as I finish some checks today.

zenweasel:
@foladipo Not sure how this is happening since you didn't touch these files but we have 3 files that are failing ESLint. Let's fix those now while we catch them.

Issue #2195, opened 2017-05-01
[Marketplace] New permissions added to an existing RC app should update existing users permissions appropriately


spencern:
When new permissions are added to existing permission groups (e.g. through the registry), existing users should have their permissions updated appropriately.

This could possibly be accomplished by checking and updating a permission group for authorization rather than that each user has a specific permission if the user is in a specific group.

Currently when new permissions are added to `defaultRoles` existing users are not granted those permissions.

spencern:
@impactmass - I _think_ this is partially completed by your groups PR as changing the permissions in the dashboard will update all users with that group, correct?

What is not completed here is if you use the `addRolesToGroups` method I don't think that checks existing users.

impactmass:
Yes. Changing it in the dashboard updates the users. The pending part is `addRolesToGroups`.

spencern:
Closed by #2606

PR #2651, opened 2017-08-08
Inviting an existing Shop user as admin should promote that account to admin

impactmass:
Related to #2195

This is to update the "Add Admin User" form/process to handle inviting an existing shop user as admin. It should promote that account to admin. This will deprecate the "Add User To Group" section of the accounts dashboard UI.

To test:
_ Use the "Add Admin User" form on the upper right of Accounts dashboard
_ Invite an already existing user and confirm that the user is promoted to the new group

Changes introduced by this PR:
_ Updates "inviteShopMember" method to handle existing user emails. It previously throws a "user exists" message.
_ Now the "inviteShopMember" method called "groups/addUser" when it's given an existing user email, and then sends email
_ Also fixed a bug in getCurrentUserName method so it uses name of the user sending invite and not always"admin"
_ Removed AddGroupMembers component. The above replaces it.


kieckhafer:
A few things I've noticed... I know there isn't an outline as to what we are expecting here, but I think we should re-think these:

1) I shouldn't have to reset my password once I get the email inviting me to the shop. Perhaps I should have to enter my password when clicking the link in the email, but currently I'm sent to a screen that asks me to set my password.

2) My name changes to whatever the invitation was sent to. I was initially "Erik Kieckhafer", but when I send the invite, i put "John Peterson" as the name on the invite, and when I accepted the invite, the "Erik Kieckhafer" account was changed to "John Peterson" as a name.

3) My address book address was erased when I accepted the invitation. I had put my shipping address in already, and when I accepted, it reset to nothing.

imports/plugins/core/accounts/server/i18n/en.json
```diff
@@ -21,7 +21,8 @@
         },
         "groupsInvite": {
           "name": "John Smith",
-          "email": "johnsmith@reactioncommerce.com"
+          "email": "johnsmith@reactioncommerce.com",
```
aaronjudd: We really shouldn't hardcode examples like this, with @reactioncommerce, it's better to use @localhost if you need to.  We don't know where this UI will end up being used, and it's better to be general.

impactmass:
@kieckhafer Thanks for raising. It sure means we can't do without the "AddUserToGroup" part. I'll stick WIP on this PR while I take a look into that.

PS: I think I got another comment notification in my email for this PR that I can't find here anymore. Is that no more relevant here?

PR #2652, opened 2017-08-08
Router enhancements

mikemurray:
This update improves the router 404 handling. It used to redirect to a `/not-found` route. With this update, it will no longer redirect but instead display the `not-found` page on the current route.

This should bring the new React Router closer the functionality of the old router.

- Add React Router Switch component to App
- Add jsdoc to Router class
- Add a final, not-found route when no routes match
- Remove redirect to /not-found
- Cleanup code

### DIFF
--- a/imports/plugins/core/router/lib/router.js
+++ b/imports/plugins/core/router/lib/router.js
@@ -32,7 +32,7 @@ if (Meteor.isClient) {
   history = createMemoryHistory();
 }
 
-// Base router class (static)
+/** Class representing a static base router */
 class Router {
   /**
    * history
@@ -66,7 +66,7 @@ class Router {
 
   /**
    * Routes array
-   * @type {[Array]}
+   * @type {Array}
    * @param {Array} value An array of objects
    */
   static set _routes(value) {
@@ -178,7 +178,7 @@ class Router {
 
   /**
    * Watch path change. Is Reactive.
-   * @return {[type]} [description]
+   * @return {undefined}
    */
   static watchPathChange() {
     routerChangeDependency.depend();
--- a/imports/plugins/core/router/lib/router.js
+++ b/imports/plugins/core/router/lib/router.js
@@ -17,7 +17,6 @@ import { getComponent } from "@reactioncommerce/reaction-components/components";
 import BlazeLayout from "/imports/plugins/core/layout/lib/blazeLayout";
 import Hooks from "./hooks";
 
-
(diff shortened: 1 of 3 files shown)


imports/plugins/core/router/lib/router.js
```diff
@@ -35,12 +34,41 @@ if (Meteor.isClient) {
 
 // Base router class (static)
```
machikoyasuda: If you could change the syntax here to jsdoc style: `/** Class representing a static base router */`, jsdoc will render the rest of the class.

imports/plugins/core/router/lib/router.js
```diff
@@ -35,12 +34,41 @@ if (Meteor.isClient) {
...
+
+  /**
+   * Active classname for active routes
+   * @type {String}
+   */
   static activeClassName = "active";
 
+  /**
+   * Routes array
+   * @type {[Array]}
```
machikoyasuda: `{Array}` -- jsdoc can't parse the `[` and `]` inside `{` and `}`s

imports/plugins/core/router/lib/router.js
```diff
@@ -49,61 +77,109 @@ class Router {
...
       }
     }
 
+    // Update route
     Router.go(current.route.name, current.params, queryParams);
   }
 
+  /**
+   * Watch path change. Is Reactive.
+   * @return {[type]} [description]
```
machikoyasuda: jsdoc can't parse the `[` and `]` inside `{` and `}`s - so, {description}`? what does this method return?

mikemurray:
@machikoyasuda updated with your suggestions.

aaronjudd:
badurl_and_2__reaction_reset_-y____reaction__node_

mikemurray:
@aaronjudd The invalid tag is still a valid route because of the match `/tag/jdftfjshfgjsfg`; though not a tag, still matches route definition.

Think in that case the product grid has to determine what's valid, and display its own custom 404 page like the PDP does.

---

Otherwise, we need a routing table (of sorts) with all possible routes to match against to determine when one truly doesn't exist.

mikemurray:
Confirmed it's happening in master as well. Going to make a new PR with a not found page / view for the product grid.

screen shot 2017-08-08 at 2 14 56 pm

mikemurray:
PR for product grid not found view on broken tag #2656, currently WIP.

PR #2655, opened 2017-08-08
[WIP] Add package.json to all plugins

aaronjudd:
- adds package.json to all plugins
- migrates as many dependencies as possible into packages
- some shared/common dependencies remains 
- uses peerDependencies for react and react-dom warnings -> gets installed by main app
- updates .bithoundrc, remove package muting
- removes some packages that were deprecated or unused
- updates Reaction to 2.0.0
- enforces and removes duplicate imports of jquery. 
- updates migrations/version package import order

Testing:

switching to this branch will require a complete reset/restart as all packages will need to be rebuilt.

Should be tested with and without custom-theme packages installed.

`npm WARN` is currently expected for some packages, where common app level dependencies still exist.

When switching away from this branch, clean up your hidden node_modules that just got installed in all your package directories!

`find . -name "node_modules" -exec rm -rf '{}' +`


aaronjudd:
Closing, the [WIP] was just so everyone knew there was a branch where this work was going on (moving to packages for 2.0) and that this should be where that work proceeds.  Will re-open when we're closer to merge.

PR #2656, opened 2017-08-08
Added a not found view to product grid

mikemurray:
- Added new component NotFound that lets you do blank views
- Added not found view to product grid if tag was not found


zenweasel:
@mikemurray If I go to a non-existent tag I don't get a 404 I get a "No Products Found". Is that the correct behavior?

zenweasel:
Is it my imagination or does rendering of products now seem considerably slower?

zenweasel:
If I go to a tag with no products I just get nothing rather than the "No Products Found" page.

mikemurray:
I think I only did, `if the tag does not exist, then no products found`; which is more like a tag not found. Should extend this to, if the products array is empty, then no product found as well.

---

The load may be slower / feel slower.

zenweasel:
I feel like if I go to a tag with no products I should get the "No product" message rather than just a blank page which looks broken. What do you think?

That's the way it currently works on `master` (blank page) so it's not a bug/blocker

mikemurray:
I agree, would be easy to add I think.

imports/plugins/core/ui/client/components/notFound/notFound.js
```diff
@@ -0,0 +1,100 @@
...
+ */
+const NotFound = (props) => {
+  // ClassName for base wrapper,
+  // If one is provied in props, the default is not used
+  const baseClassName = classnames({
+    "container-fluid-sm": typeof props.className === "undefined"
+  }, props.className);
+
+  // ClassName for content wrapper
+  // If one is provied in props, the default is not used
```
zenweasel: typo in comment

imports/plugins/core/ui/client/components/notFound/notFound.js
```diff
@@ -0,0 +1,100 @@
...
+NotFound.propTypes = {
+  /**
+   * Children
+   * @type {ReactNode}
+   */
+  children: PropTypes.node,
+
+  /**
+   * Baase wrapper className
+   * @summary String className of `classnames` compatable object for the base wrapper
```
zenweasel: typo in comment

imports/plugins/core/ui/client/components/notFound/notFound.js
```diff
@@ -0,0 +1,100 @@
...
+        {props.children}
+      </div>
+    </div>
+  );
+};
+
+NotFound.propTypes = {
+  /**
+   * Children
+   * @type {ReactNode}
```
zenweasel: Should we make this more consistent? I feel like we call this different things different place (e.g. Node/ReactNode/JSX. Is there a difference?)

imports/plugins/core/ui/client/components/notFound/notFound.js
```diff
@@ -0,0 +1,100 @@
...
+   */
+  children: PropTypes.node,
+
+  /**
+   * Baase wrapper className
```
zenweasel: typo in comment

imports/plugins/core/ui/client/components/notFound/notFound.js
```diff
@@ -0,0 +1,100 @@
...
+   * @summary String className of `classnames` compatable object for the base wrapper
+   * @type {String|Object}
+   */
+  className: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+
+  /**
+   * Content className
+   * @summary String className of `classnames` compatable object for the content wrapper
```
zenweasel: typo in comment (compatible)

imports/plugins/core/ui/client/components/notFound/notFound.js
mikemurray: Maybe just go with `Node` prop type.

@machikoyasuda what do you think? Maybe we can set up a type in jsdoc that knows `Node` is react node when the docs are generated?

mikemurray:
@zenweasel Should now show the not-found view if the tag is valid, but no products were found still.

@machikoyasuda As a bonus, jsdoc'd all the things in `Products` component.

imports/plugins/core/ui/client/components/notFound/notFound.js
```diff
@@ -0,0 +1,100 @@
...
+NotFound.propTypes = {
+  /**
+   * Children
+   * @type {Node}
+   */
+  children: PropTypes.node,
+
+  /**
+   * Base wrapper className
+   * @summary String className or `classnames` compatable object for the base wrapper
```
zenweasel: compatible still spelled wrong isn't it?

zenweasel:
@mikemurray three lint errors that seem to come from nowhere

mikemurray:
@zenweasel wonder if something changed in our eslint config since those files were passing before, right?

mikemurray:
Fixed all lint issues

zenweasel:
@mikemurray yeah, I don't know where those lint errors came from but thanks for fixing them

zenweasel:
@mikemurray @machikoyasuda still two lint errors left

mikemurray:
ok, whats going on. I fixed all the lint errors and more keep poping up?

mikemurray:
All green!

PR #2662, opened 2017-08-10
[Marketplace] Inviting to groups should be scoped to user's roles + UI change

### LINKED ISSUE #2189: [Marketplace] Permission to grant scoped permissions should be grantable
Owners should be able to grant administrators permission to invite additional admins or grant users additional permissions. 

Permissions that can be granted by the administrator must be scoped to the shop or marketplace that the user is administrator for.

Perhaps this permission to grant should be limited to existing permissions that the admin has.

impactmass:
Resolves #2189 

**NB:** This PR:
* Changes to ensure Admin Invite can only be done into groups that an admin performing the invite has permissions for. These changes are done both in the methods and in the UI components
* Ensures that owner has all roles that are in other groups (plus the "owner" role). I found that roles such as "index" was not in the owner's list. We were giving owner general access based on the presence "owner" role alone.
* Includes some UI improvements (a little step closer to the zeplin mocks). 
* Freezes edits on Owner permissions list
* Adds pop-up confirmation before removing a user from a group

**How To Test:**
* As the shop owner, create a new group (sample name: Accounts Manager). Toggle on the `accounts` permission for that group.
* Invite a user to that group
* Login as the newly invited user, and confirm that he can only invite to that group (or any other group that have that user have it's permissions). 
* In contrast, a shop manager, for example, has more roles and can invite into that new "Accounts Manager" group.


server/api/core/core.js
```diff
@@ -60,7 +60,10 @@ export default {
     const registeredPackage = this.Packages[packageInfo.name] = packageInfo;
     return registeredPackage;
   },
+  defaultCustomerRoles: [ "guest", "account/profile", "product", "tag", "index", "cart/checkout", "cart/completed"],
```
spencern: Why are these back?

server/api/core/core.js
```diff
@@ -70,13 +73,17 @@ export default {
     }
 
     const shops = Shops.find(query).fetch();
-    const ownerRoles = Roles.getAllRoles().fetch().map(role => role.name);
-    const shopManagerRoles = ownerRoles.filter(role => role !== "owner");
+    let ownerRoles = Roles.getAllRoles().fetch().map(role => role.name);
+
+    // join all other roles with package roles for owner. Owner has all roles
+    ownerRoles = ownerRoles.concat(this.defaultCustomerRoles, this.defaultVisitorRoles);
```
spencern: Not sure about this. Feels pretty brittle

server/api/core/core.js
```diff
@@ -87,18 +94,44 @@ export default {
...
+   */
+  canInviteToGroup(options) {
+    const { group, user } = options;
+    const userPermissions = user.roles[group.shopId];
+    const groupPermissions = group.permissions;
+
+    // check that userPermissions array contains all groupPermissions
+    // _.difference(subset, superset).length === 0
+    // https://github.com/lodash/lodash/issues/1743#issuecomment-170598139
+    if (_.difference(groupPermissions, userPermissions).length === 0) {
```
spencern: As with this exact method above, let's inline the comments about this solution.

server/api/core/core.js
spencern: That's right.. I remember now, what triggered the question was the move and seeing them as additions.

impactmass:
@spencern I've fixed the breaking tests (after removing the `.only()` ). So you can take a look at the PR again.

You'll notice these two happen that I need to resolve:
* When creating a group, the list of roles/permissions available to be used should only be those the user have access to.
* Ensure user can only downgrade accounts at his level or below 

Considering that this PR at this point is already large (~20 files changed).... I'm going to create a separate ticket and tackle those there.

spencern:
@impactmass 

Just tried to invite a user to the main shop again and have this alert
image

impactmass:
@spencern let's double check this issue when you're back. I don't get that error as the shop owner. 
(perhaps you haven't pull the latest yet?)

spencern:
I had pulled the latest, but apparently not reset the server. Resetting fixed the issue.

PR #2664, opened 2017-08-11
Fix for variant option not updating when edited

### LINKED ISSUE #2582: Option name does not change when edited
### Expected behavior
When you change the option title (not the label) it should change elsewhere

### Actual Behavior
It does not change unless you delete the existing contents and then click away and then replace it.
basic_t-shirt

1. Login as admin
1. Create a new product
1. Set the Variant label
1. Change title, label, quantity and price.
1. Observe that the option title shown does not change but remains "Untitled Option" even when you publish

### Versions
```
Node: 7.10.0
NPM: 5.0.3
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.4.0
Reaction branch: development
Docker: 17.06.0-ce
```

foladipo:
Resolves #2582 

This PR fixes a bug with creating product variant options. Previously, editing a product did not update parts of UI. This bug fixes that.

### Testing steps
* Log in as admin.
* Open the panel for editing a product.
* Create a new variant option and try to edit its `option` or `price` form field. Immediately you hit `Enter`, click elsewhere on the page or hit `Tab`, the UI is updated with the text you added to the form field.

### DIFF
--- a/imports/plugins/included/product-variant/client/templates/products/productDetail/variants/variantForm/childVariant.html
+++ b/imports/plugins/included/product-variant/client/templates/products/productDetail/variants/variantForm/childVariant.html
@@ -73,23 +73,6 @@
               </div>
 
             <div class="rui item half">
-              <div class="rui textfield form-group {{hasErrorClassName 'optionTitle'}}">
-                <label for="option-{{childVariantFormId}}" data-i18n="productVariant.optionTitle">Option Title</label>
-                <input
-                  id="option-{{childVariantFormId}}"
-                  type="text"
-                  class="form-control input-sm child-variant-input"
-                  value="{{optionTitle}}"
-                  name="optionTitle"
-                  placeholder="Add option title"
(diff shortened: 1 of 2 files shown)


mikemurray:
This should be going into `marketplace`, right @aaronjudd ?

foladipo:
@mikemurray The initial bug report stated `development` as the branch on which the bug was observed, and that was why I branched off it.

zenweasel:
This fixes the bug but why did you change the titles and the placement of the fields? We would need to get design guidance before doing that.

zenweasel:
@foladipo Can you close this and rebranch from `marketplace`. It's not your mistake but we want all PR's to go into `marketplace` from now on. Thanks.

foladipo:
Ok. I'll rebranch, restore the order of that form and raise another PR.

PR #2667, opened 2017-08-13
Giving a user shopSettings permission does not allow them edit shop settings

### LINKED ISSUE #2294: Giving a user "Shop Settings" permission does not allow them to edit Shop settings
### Expected behavior
If they are granted permissions they should be able to edit these values? If not, then maybe it needs to show edit disabled?

### Actual Behavior
Receive a toast error message

reaction

1. Invite a user and grant them "Dashboard" and "Shop Settings" permissions
1. Log in as that user
1. Try to edit shop settings
1. Observe toast error message

### Versions
```
Node: 7.10.0
NPM: 4.2.0
Meteor Node: 4.8.2
Meteor NPM: 4.5.0
Reaction CLI: 0.8.2
Reaction: 1.2.0
Reaction branch: release-1.2.0
Docker: 17.03.1-ce
```

rhenshaw56:
**Resolves #2294** 

This happened because
- `shopSettings` permission was not amongst included permissions required for updates on Shops collection
- `Roles.userInRoles()` in email publication used to check if a user belongs to a particular group (with certain permissions) in a shop was returning false because it wasn't given a shopId required for checks. 

Test by
- Login as admin
- Create a new group called `customers` and Invite a user to that group
- Grant the `customers` group "Dashboard" , "Shop Settings" and "Email" permissions
- Log in as the user in step 2
- Try to edit shop settings by clicking on the shop icon on the toolbar menu
- Notice that a success toast message is shown and the set shop settings are changed
- Click on the email icon in the toolbar menu
- Notice that the email logs are visible but email settings are hidden


zenweasel:
@rhenshaw56 If you are going to submit a non-WIP PR, please make sure and assign reviewers.

zenweasel:
@rhenshaw56 Should I be seeing two entries for "dashboard"?

reaction

rhenshaw56:
@zenweasel i think that's the way the accounts dashboard is but i'll have to confirm with @impactmass . There's two of some permissions like accounts, checkout, connectors, dashboard.

impactmass:
re: those permission roles that seem like duplicates. They are parent and child roles. Those parent roles have child roles of the same name. We'll be cleaning those up at some point with Ryan, but those items would be nested within their respective parents when PR #2662 gets merged.

(PS. I'm testing the PR shortly)

rhenshaw56:
@impactmass i tested it and it seems to be something view related for multi-shops rather than the permissions not been assigned. Because when i do `Meteor.user()` in the console, i get to see `reaction-email` for that user but then even for the admin who's giving out that permissions the `Email` icon does not show. 
screen shot 2017-08-15 at 9 24 54 am

No email icon for an Admin.

impactmass:
@rhenshaw56 Due to the freezing of some tests that's happening on Marketplace, let's get back to this when #2662 gets merged. That way we are sure of tests as regards the fix.
I'll revisit this when that is done

Issue #2614, opened 2017-08-01
[Marketplace] Notification page not found


kieha:
### Expected behavior

When I click on the notifications icon on the navbar, Clicking on `View All` should take me to the notifications page.

### Actual Behavior

Clicking on `View All` takes me to the `Not found` page.

### Steps to Reproduce the Behavior

1. Make an order (as admin or non-admin)
2. Click on the notifications icon on the navbar
3. Click on the `View All` link
4. Observe that you're re-directed to the `Not Found` page

### Versions

Node: 6.2.2
NPM: 3.10.9
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 1.13.0

zenweasel:
Closed via #2673

PR #2673, opened 2017-08-14
[Marketplace] Fix notification page not found error #2614

jm-wanja:
Resolves issue #2614 

**NB:** This PR fixes the notification page not being found bug. This happened because the prefix given to the path for the `Link to` was wrong. 
It now works for both single-shop and multiple-shops scenarios, clicking the `View All` link on the notification icon will direct you to the `Notifications Page`.

**How To Test**
- Make an order as an admin or guest.
- Click on the Notifications Icon on the navbar
- Click on `View All`
- Observe that you are directed to the `notifications page`
- For multiple-shops scenario repeat the steps on the second shop.


imports/plugins/included/notifications/client/components/notificationDropdown.js
```diff
@@ -93,7 +87,7 @@ class NotificationDropdown extends Component {
           })}
         </ul>
         <div className="dropdown-footer text-center">
-          <Link to={`${this.prefix}/notifications`} data-i18n="notifications.body.viewAll">View All</Link>
```
zenweasel: The prefix is needed for when we have multiple shops, I think just removing it and hard-coding the link is going to break the link in some cases. Make sure you have tested this in both multi-shop and single shop scenarios. Ping me if you need help setting that up.

jm-wanja: Tested both scenarios and the `View All` link on the notification icon will direct you to the `Notifications Page`.

zenweasel:
Man, I hate to be a pain but I still feel like this is going to break something because we should be using the shop prefix.

PR #2679, opened 2017-08-15
fix className variable

kieckhafer:
fix className to allow variable from props, prior to this change it was literally displaying `{this.props.name}-edit-input` as class


kieckhafer:
Wrong branch... #2681 replaces this

PR #2683, opened 2017-08-15
Fix on ProductAdmin meta textfields throwing errors on state change

### LINKED ISSUE #2581: ProductAdmin meta textfields throwing errors on state change
Editing meta textfields in the ProductAdmin component is causes an issue with the field's state which throws an error in the console.

basic_reaction_product_and_reaction_ _working_copy__register-component-api_ _29_changed_files_

The state change that's throwing the error is here:

https://github.com/reactioncommerce/reaction/blob/development/imports/plugins/included/product-admin/client/containers/productAdminContainer.js#L48-L68

The updates appear to work (which is how it has gone unnoticed), but I suspect there's a decent chance something is broken somewhere that we're just not seeing on the surface.  Either way, the immutable `update` method being used there is deprecated anyway, so it should probably be replaced.  

That container should probably be consolidated into the `<ProductDetail/>` component that it wraps too.  It's a little too complicated to debug a component full of state and event handlers that all get passed to another component below it.  They could all be in one place and then only pass the data in from the container.  Perhaps that's out of scope for this issue, but it also may be one of those "while you're there..." kind of things.

- `reaction init -b development`
- log in as admin and go to product detail page
- change to edit mode
- click an edit pencil to open product admin panel
- expand "Details" tab and start typing in any meta text field
- note the errors shown above in your console after every keystroke

### Versions
```
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.4.0
```

rhenshaw56:
**Resolves #2581** 

The error thrown was been caused by `this.state` not having a `products` field. However the reason why it was still been saved was because saving happens `onBlur`. So this actually saves regardless of the console error been thrown. All i had to do was re-arrange things a bit and take away the `update()` method. @jshimko the `update()` method actually still works but not sure if its still in use with the latest react because its been used as a standalone as `react-addons-update`.

**Test by**
- Running reaction from the  marketplace branch
- Log in as admin and go to product detail page
- Change to edit mode
- Click an edit pencil to open product admin panel
- Expand "Details" tab and start typing in any meta text field
- Observe that there's no console errors


zenweasel:
@rhenshaw56 This is still marked WIP, is it ready to go?

rhenshaw56:
@zenweasel yeah, its ready to go, removing the WIP now.

Issue #2621, opened 2017-08-02
[Marketplace] Receive error when viewing homepage as incognito user


zenweasel:
### Expected behavior
Should not receive any errors when visiting the homepage

### Actual Behavior
Receive this error in the browser console
```
Exception in delivering result of invoking 'shop/getLocale': TypeError: Cannot read property 'currency' of undefined
    at http://localhost:3000/app/app.js?hash=42157a695a333e6013bb9333691cb188f3ad3908:103267:54
    at MethodInvoker._callback (http://localhost:3000/packages/meteor.js?hash=814eae5b938a9a89ce7cf9174c98be4982f953d8:1109:22)
    at MethodInvoker._maybeInvokeCallback (http://localhost:3000/packages/ddp-client.js?hash=14d966b7972bd95a1f7015fec9ac340f10508a44:3681:12)
    at MethodInvoker.receiveResult (http://localhost:3000/packages/ddp-client.js?hash=14d966b7972bd95a1f7015fec9ac340f10508a44:3701:10)
    at Connection._livedata_result (http://localhost:3000/packages/ddp-client.js?hash=14d966b7972bd95a1f7015fec9ac340f10508a44:4827:9)
    at onMessage (http://localhost:3000/packages/ddp-client.js?hash=14d966b7972bd95a1f7015fec9ac340f10508a44:3528:206)
    at http://localhost:3000/packages/ddp-client.js?hash=14d966b7972bd95a1f7015fec9ac340f10508a44:2908:9
    at Array.forEach (<anonymous>)
    at Function._.each._.forEach (http://localhost:3000/packages/underscore.js?hash=cde485f60699ff9aced3305f70189e39c665183c:149:11)
    at SockJS.self.socket.onmessage (http://localhost:3000/packages/ddp-client.js?hash=14d966b7972bd95a1f7015fec9ac340f10508a44:2907:43)
```

### Steps to Reproduce the Behavior
1. Visit the homepage as an incognito user

### Versions
```
Node: 7.10.0
NPM: 5.0.3
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.0-ce
commit: b6e18a4b5a35b852bd75648ae395a2f4b77eae81
```

zenweasel:
Closed via #2686

PR #2686, opened 2017-08-16
[Marketplace] Fix error when viewing homepage as an incognito user #2621

jm-wanja:
Resolves issue #2621 

**NB** This PR fixes the browser console error that appeared when visiting the homepage as an incognito user on the browser. This error was caused by the code that sets the shop currency to local storage being executed before the callback was done, so I added the if condition to ensure that the shop object is returned before setting it's currency to local storage.

 ** How To Test**
1. Visit the homepage as an incognito user


package.json
```diff
@@ -26,6 +26,7 @@
     "autoprefixer": "^7.1.2",
     "autosize": "^4.0.0",
     "babel-runtime": "^6.23.0",
+    "bcrypt": "^1.0.2",
```
zenweasel: Nope. Breaking `brcypt` break Docker.

jm-wanja: Alright let me correct this

PR #2687, opened 2017-08-16
Fix for variant option not updating when edited

### LINKED ISSUE #2582: Option name does not change when edited

foladipo:
Resolves #2582 

This PR fixes a bug with creating options for product variants. Previously, editing an option did not update parts of the UI. This PR fixes that bug.

### Testing steps
* Log in as admin.
* Open the panel for editing a product.
* Create a new variant option and try to edit its `option` or `price` form field. Immediately you hit `Enter`, click elsewhere on the page or hit `Tab`, the UI is updated with the text or price you added to the form field.


imports/plugins/included/product-variant/client/templates/products/productDetail/variants/variantForm/childVariant.js
```diff
@@ -49,9 +49,11 @@ Template.childVariantForm.helpers({
   Icon() {
     return Components.Icon;
   },
+
```
zenweasel: Can we not add in these line breaks for no reason?

zenweasel:
@foladipo So this is an example of where we want to make sure and loop back on these PR's quickly. Let's try to make the one change I requested and get this merged? You should almost always prioritize dealing with PR feedback over almost anything.

foladipo:
So sorry for the delay @zenweasel . I've now cleaned out the unnecessary line breaks and pushed the update.

PR #2692, opened 2017-08-16
React Permissions wrapper component

impactmass:
Updated:

This PR adds a withPermissions HOC for using to wrap components like so `registerComponent("Comp", Comp, withPermissions({ roles: ["accounts"] }));`. It returns the new wrapped component with a "hasPermissions" prop. 

I added an example with a component on the GroupsTable which already needs a wrapper like this.


imports/plugins/core/ui/client/components/permission/permission.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+/**
+ * withPermissions - Decorator to add permissions to any component
+ * withPermissions(["guest"])(MyComponent)
+ * @param  {Array|String} roles String or array of strings of permissions to check. default: roles=["guest", "anonymous"]
+ * @param  {String} group Slug title of a group to check against. Group option supercedes roles option. default: group="customer".
+ * Group access is given to users at that group level and above
+ * @return {function} Returns a compser higher order component
```
mikemurray: typo: `compser`

imports/plugins/core/ui/client/components/permission/permission.js
```diff
@@ -0,0 +1,57 @@
...
+          hasPermission = _.difference(grp.permissions, permissions).length === 0;
+        }
+      }
+
+      onData(null, { hasPermission });
+    }
+  });
+}
+
+registerComponent("Permission", withPermissions()(Permission));
```
mikemurray: should be
// Name, Base component, Decorators (withPermission)
`registerComponent("Permission", Permission, [ withPermission ])`

imports/plugins/core/ui/client/components/permission/permission.js
```diff
@@ -0,0 +1,57 @@
+import { Component, Children } from "react";
+import { Meteor } from "meteor/meteor";
+import _ from "lodash";
+import PropTypes from "prop-types";
+import { composeWithTracker, registerComponent } from "@reactioncommerce/reaction-components";
+import { Groups } from "/lib/collections";
+import { Reaction } from "/client/api";
+
+class Permission extends Component {
```
mikemurray: Add comment above the class like so:

```
/**
 * @summary React component for [REST OF YOUR SUMMARY]
 * @param {Object} props - React PropTypes
 * @property {Node} children - [THIS IS YOUR PROP TYPE DEFINITONS, works kike @param]
 * @property {Bool} hasPermission - [THIS IS YOUR PROP TYPE DEFINITONS, works kike @param]
 */
```

jshimko:
So, just a thought...

There's no markup needed here.  This would be way more flexible as a HOC.  This should just export that composer function so any other component can use it via `registerComponent`.  And it should probably go with [the rest of the HOC's in the components plugin](https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/core/components/lib/hoc.js).  Then people can use it wherever then need it within their wrapped component (instead of having it be all or nothing).

```js
// nothing more than this...

// HOC
export function withPermissions(roles = ["guest", "anonymous"], group = "customer") {
  return composeWithTracker((props, onData) => {
    const grpSub = Meteor.subscribe("Groups");

    if (grpSub.ready()) {
      let hasPermission = Reaction.hasPermission(props.roles || roles);

      if (group) {
        hasPermission = false;
        const grp = Groups.findOne({ slug: props.group || group });
        const user = Meteor.user();
        const permissions = user.roles[Reaction.getShopId()] || [];
        if (grp && grp.permissions) {
          // checks that userPermissions includes all elements from groupPermissions
          hasPermission = _.difference(grp.permissions, permissions).length === 0;
        }
      }

      onData(null, { hasPermission });
    }
  });
}
```

Then to use it in any other component...

```js
import { composeWithTracker, registerComponent, withPermissions } from "@reactioncommerce/reaction-components";

const MyComponent = ({ hasPermissions }) = (
  <div>
    {hasPermissions ? <span>You can see me!</span> : <span>Sorry, not allowed!</span>}
  </div>
);

// with default roles
registerComponent("MyComponent", MyComponent, withPermissions);

// with custom roles/group
registerComponent("MyComponent", MyComponent, withPermissions(['admin', 'other'], ['somegroup']));

// with other HOC's
registerComponent("MyComponent", MyComponent, [
  withPermissions,
  withCurrentUser
]);
```

Then you could have as much or little conditional UI as you want.  And allows other people to override your markup easily without touching the HOC options.

jshimko:
I also generally like to stick with the convention of stating the prop names in the HOC name.  So `withIsAdmin` injects an `isAdmin` prop, `withCurrentUser` injects `currentUser`, and so on.  So in this case `withHasPermissions` could inject `hasPermissions` (which I think should be plural).  But I also like the shorter option of `withPermissions` that adds `hasPermissions`.  So maybe it's worth altering the convention a little sometimes.

jshimko:
(updated some small typos/issues in my code example)

impactmass:
🎉 thanks @jshimko and @mikemurray, those are good upgrades. working on them. (adding WIP for now)

impactmass:
@mikemurray you can take a look at this again

imports/plugins/core/accounts/client/components/groupsTableButton.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Components, registerComponent, withPermissions } from "@reactioncommerce/reaction-components";
+
+const GroupsTableButton = ({ account, group, handleRemoveUserFromGroup, hasPermissions }) => {
```
mikemurray: Should have a @jsdoc comment. example for react components: https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/core/ui/client/components/notFound/notFound.js#L8

PR #2701, opened 2017-08-18
Combine order and cart transforms into one, replace usages

### LINKED ISSUE #2684: Refactor cart and order transforms into one
When I was working on Taxes I created a separate order transform because the cart transform methods all have names like 'cartSubtotal`. But we need to refactor these back into one because their functionality will be 95% the same I believe. It just means changing the name of all the references to the old method names.

zenweasel:
Resolves #2684  Prepare to move more logic into cart/order transforms (like `itemsByShop`)


spencern:
I think this is probably the way to go, what do you think about merging the stripe connect PR first?

zenweasel:
Six of one, half-dozen of the other to me.

zenweasel:
@spencer if you could take a look at this when you get a chance. Thanks.

spencern:
@zenweasel this is breaking the stripe connect tests

zenweasel:
@spencern Probably with this should be fixed.

aaronjudd:
I have an unsubstantiated feeling that this could be a dangerous PR, as I had some very specific reasons for the way transforms worked and were divided in their client/server functionality before.  In reviewing, testing (albeit quickly), I couldn't see anything amiss, and I don't really recall the original issue (just that there was a important requirement that made it so, back then).  It's possible that the architecture has changed enough that it's not an issue, and that we've moved on.   


The next thing I was concerned about, we renamed a lot of functions and added the `get` prefix to them. 
```
-      cartSubTotal: cart.cartSubTotal(),		 +      cartSubTotal: cart.getSubTotal(),
 -      cartCount: cart.cartCount(),		 +      cartCount: cart.getCount(),
 -      cartShipping: cart.cartShipping(),		 +      cartShipping: cart.getShippingTotal(),
 -      cartDiscount: cart.cartDiscounts(),		 +      cartDiscount: cart.getDiscounts(),
 -      cartTaxes: cart.cartTaxes(),		 +      cartTaxes: cart.getTaxTotal(),
 -      cartTotal: cart.cartTotal()		 +      cartTotal: cart.getTotal()

```

These are breaking changes to anyone who has implemented these.  We'll need documentation updates, and we should also probably explain the logic behind the change.  (this might be removal of any current references, and just let jsdoc handle the api documentation).

For me, I viewed these transforms as object properties, rather than getter methods. I'm not there's a perfect convention here, but if we're introducing breaking changes, it ought to be to follow some convention.

zenweasel:
@aaronjudd They aren't properties, and in a post-Blaze world you can't use them as such, which is why I changed them to method-naming to make it clearer which were which. Also, some of the method names conflicted with **actual** property names so adding the method-naming prevented any confusion between properties on the db object and computed values.

zenweasel:
I need to create an issue for documentation of code conventions and for a way of handling deprecation of existing method

spencern:
@zenweasel when you do followup with docs / deprecation will you tag that issue here?

zenweasel:
Added a PR for backwards compatibility and deprecation warnings here: https://github.com/reactioncommerce/reaction/pull/2756

PR #2704, opened 2017-08-18
[Marketplace] Adds the ability to set metafields on cart items

mikemurray:
This PR adds the ability to add metafields to a cart item while adding that item to cart. This could be useful to apply additional information to an item added like affiliate tracking or geolocation, etc.

**Examples:**

Adds metadata to a cart item only once, and does not get updated if the same item is added to the cart again.

```
const additionalOptions = {
  metafields: [
    { key: "KEY", value: "VALUE" }
  ]
};

Meteor.call("cart/addToCart", "PRODUCT_ID", "VARIANT_ID", QUANTITY, additionalOptions);
```

Adds metadata to a cart item while adding to cart, and will update every time that item is added to cart.

```
const additionalOptions = {
  overriteExistingMetafields: true,
  metafields: [
    { key: "KEY", value: "VALUE" }
  ]
};

Meteor.call("cart/addToCart", "PRODUCT_ID", "VARIANT_ID", QUANTITY, additionalOptions);
```

Final param, additionalOptions is optional.

```
Meteor.call("cart/addToCart", "PRODUCT_ID", "VARIANT_ID", QUANTITY);
```


server/methods/core/cart.js
```diff
@@ -314,12 +314,21 @@ Meteor.methods({
...
-  "cart/addToCart": function (productId, variantId, itemQty) {
+  "cart/addToCart": function (productId, variantId, itemQty, additionalOptions) {
     check(productId, String);
     check(variantId, String);
     check(itemQty, Match.Optional(Number));
+    check(additionalOptions, Match.Optional(Object));
+
+    // Copy additionalOptions into an options object to use througout the method
+    const options = {
+      overriteExistingMetafields: false, // Allows updating of metafields on quantity change
```
zenweasel: typo in method name

mikemurray: good catch

zenweasel:
My only concern here is that Metadata not turn into ActualData. We need to not store anything there that the app relies on to do any primary logic, otherwise it should be in the schema.

PR #2711, opened 2017-08-21
Fix PDP images not associated with variant/options

### LINKED ISSUE #2647: [Marketplace] PDP images not associated with variant/options
### Expected behavior
When a variant is selected, and you upload an image, it should be attached to the selected option, and visible.

### Actual Behavior

Images are uploading and attached to the top level product and are not attached to the selected variant.  The only way to upload an image to the variant is through the dashboard, and this seems like it should be a behavior of the media gallery

basic_reaction_product

Select test option, and upload image, correct behavior is that you should immediately see the image, and if you select another option, you should not (as images are displayed per variant/option)

### Versions
Node: 6.11.1
NPM: 3.10.10
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.0-ce

awadesmoline:
This fixes #2647 

# Steps to test
1. As admin, select a variant option in PDP
1. Click on the thumbnails to upload image for the variant.
basic_reaction_product

1. Observe that the image is attached to the selected variant. And if you select another option, you should not see the image
basic_reaction_product2


aaronjudd:
@zenweasel since you disagreed (in theory) with this behaviour, can you take a look and see if it makes sense or not to you?

zenweasel:
Yeah, I think it makes sense once you realize you have to select the top-level variant to be able to modify the "product" image and that just "deselecting" the options does not allow you to set the top-level images.

Issue #2355, opened 2017-05-26
Search orders by product, variant, or option title


spencern:
Users with permissions to view orders dashboard should be able to quickly search orders by the order items.

This search should be fuzzy and case insensitive.

Users should be able to type:
-   Product Title
-   Variant Title / Label
-   Option Title / Label

If there is more than one order with matching products - the search UI should replace the existing list of orders in the orders dashboard with the results of the search. Any existing custom sorting (e.g. sort by recently shipped, customer name, etc) or filtering (filter by status, between dates, etc) the user has done should persist.

If there is exactly one order with matching customer information, the matching orderId should be opened in the action view panel if the user has permission to view that order.

If there are no matching orders, there should be a "No orders found" message displayed inline to the user. The user should also be alerted if there are any filters enabled that might hide results and the alert should permit the user to turn off all filters and run the search again.

Extends #2152 with specifications.

Annotated Designs in Zeplin:
Web: https://zpl.io/1r0Vvm
App: zpl://screen?pid=58af8900737c137280322bd6&sid=58fa9b08c1ee69df600c0469

Related to: #2046

aaronjudd:
Closed in #2717

Issue #2354, opened 2017-05-26
Search orders by customer information


spencern:
Users with permissions to view orders dashboard should be able to quickly search orders by customer information.

Users should be able to type:
-  Customer first or last name
-  Customer phone number
-  Customer email address
-  Last 4 of customer credit card (Maybe useful?)

If there is more than one order with a matching customer information - the search UI should replace the existing list of orders in the orders dashboard with the results of the search. Any existing custom sorting (e.g. sort by recently shipped, customer name, etc) or filtering (filter by status, between dates, etc) the user has done should persist.

If there is exactly one order with matching customer information, the matching orderId should be opened in the action view panel if the user has permission to view that order.

If there are no matching orders, there should be a "No orders found" message displayed inline to the user. The user should also be alerted if there are any filters enabled that might hide results and the alert should permit the user to turn off all filters and run the search again.

This search should be case insensitive.

Extends #2152 with specifications.

Annotated Designs in Zeplin:
Web: https://zpl.io/1r0Vvm
App: zpl://screen?pid=58af8900737c137280322bd6&sid=58fa9b08c1ee69df600c0469

Related to: #2046

impactmass:
Update: WIP PR on this #2717. Needs #2688 to be merged first.

zenweasel:
Closed via #2717

PR #2688, opened 2017-08-16
[Marketplace] Order Filtering By Status

joykare:
Resolves #2357 

*Description*
Users with permissions to view orders dashboard should be able to quickly filter the list of orders by the order status.

*How to test*
- Clone branch
- Create multiple orders
- Have some orders in different stages of the order workflow ["New", "Approved", "Captured", "Completed", "Cancelled"]
- Use filters observe orders display according to filter
- Click on left-most icon to clear filter [Observe all your orders displayed]

PR #2717, opened 2017-08-22
Order dashboard search by orderId, customer info, order product info

### LINKED ISSUE #2353: Search orders by orderId
Users with permissions to view orders dashboard should be able to quickly search orders by OrderId. In small shops, the first 4-5 characters of an `orderId` should likely be sufficient to identify a single order, in large shops, it may require 6-8 characters, but still the most simple way to look them up.

Users should be able to type the first few characters of an orderId and hit enter:

If there is more than one order with a matching id `/^${searchString}/i` -  the search UI should replace the existing list of orders in the orders dashboard with the results of the search. **Any existing custom sorting (e.g. sort by recently shipped, customer name, etc) or filtering (filter by status, between dates, etc) the user has done should persist.**

If there is exactly one order with a matching id, the matching orderId **should be opened in the action view panel (i.e order summary panel)** if the user has permission to view that order.

If there are no matching orders, there should be a "No orders found with that ID" message displayed inline to the user. The user should also be prompted if there are any filters enabled that might hide results.

I think this search should be case insensitive. It does increase the possibility of false positives, but not by much. A single extra character is more than likely enough to permit case insensitive search at the same level of accuracy for most order sets.

Extends #2152 with specifications.

Annotated Designs in Zeplin:
Web: https://zpl.io/1r0Vvm
App: zpl://screen?pid=58af8900737c137280322bd6&sid=58fa9b08c1ee69df600c0469

Related to: #2046

impactmass:
Resolves #2353, #2354 and #2355 

## How To Test
- Create a few orders
- Search the orders using the Search Bar at the top with:
** OrderId
** Customer info (name, email, phone, card - last 4 digits)
** Product title, Variant title, Variant option title

- Confirm that all these search and any active filter takes effect on the results being displayed by the table.
- Confirm that clearing all filters and using search alone also works.

.
The only part of this pending is opening up the order summary action view if exactly one match/result is present in the table. Currently working on making Sortable not prevent the opening, but, I think reviews can start on this. I'll get the opening in before review completion.

**Update:**
There is a blocker to sliding out the table when there's only one matching item because the data fetching & rendering is happening within the table itself. Since we'll be moving the fetching #2768, that seem a better time to achieve this.


aaronjudd:
@impactmass now that #2688 is merged is this still a WIP? can you assign reviewers?

impactmass:
It got merged in after I left work. 
I'm updating this with main branch now, then test & fix any issues found, then assign reviewers.

impactmass:
I discovered places where there are [0] array references, so I had to update those in view of marketplace. With that done, this should be out of testing shortly.

imports/plugins/core/orders/client/components/orderDashboard.js
```diff
@@ -26,21 +30,74 @@ class OrderDashboard extends Component {
...
 
+  componentWillUnmount() {
+    this.subscription.stop();
+  }
+
+  // Extracted Tracker logic for the search subscription, to allow calling in both
+  // componentDidMount and componentWillReceiveProps
+  // This tracker is setup in the component because we need to re-subscribe when search input changes
+  setupTracker = () => {
+    Tracker.autorun(() => {
```
joykare: This should be happening in a container and passed down to the component as props. 
Rule of thumb: Components are presentational

impactmass: Well.... thanks.

I did make a comment just above the function to say why it's being done there. 
Or perhaps you meant that I should just move the same function (as is) into the container file? I'll do that now, but just as we're talking containers/components... the current orderdashboardcontainer probably needs more work to function as HOC really.

impactmass:
@joykare i've attended to your comment. @zenweasel I re-assigned you as well.

impactmass:
This needs a little design check too, so adding design label.

rymorgan:
There are a few weird UI issues that I'm noticing. When you search you should be able to filter the results. Instead, it seems to be resetting the results. The results should only reset when you hit clear. 
filter not working

rymorgan:
When clicking on an order to process it, the window closes when it should just open the details panel. 
filter not working

rymorgan:
translations

imports/plugins/core/orders/client/components/orderSearch.js
```diff
@@ -0,0 +1,72 @@
+import React, { Component } from "react";
+import { compose } from "recompose";
+import PropTypes from "prop-types";
+import { Components, withPermissions } from "@reactioncommerce/reaction-components";
+
+/**
+ * Class representing the Products React component
```
zenweasel: This represents the Products React component? Is this a copy/paste error? `load more products callback`?

imports/plugins/core/orders/client/components/orderSearch.js
```diff
@@ -0,0 +1,72 @@
...
+        <Components.TextField
+          className="search-input"
+          onChange={this.handleChange}
+          value={this.state.query}
+          i18nKeyPlaceholder="admin.dashboard.searchLabel"
+        />
+        <i className="fa fa-search fa-fw"/>
+        <Components.Button
+          className="search-clear"
+          i18nKeyLabel="search.clearSearch"
```
zenweasel: Missing a translation entry

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -141,8 +166,40 @@ class OrderDashboardContainer extends Component {
     });
   }
 
+  // Extracted Tracker logic for the search subscription, to allow calling in both
+  // componentDidMount and componentWillReceiveProps
+  // This tracker is setup in the class itself because we need to re-subscribe when search input changes
```
zenweasel: I see your comment there, but is there no way to do this in the container?

imports/plugins/included/search-mongo/server/methods/searchcollections.js
```diff
@@ -167,36 +167,52 @@ export function buildOrderSearchRecord(orderId) {
...
-  orderSearch.shippingName = order.shipping[0].address.fullName;
-  orderSearch.shippingPhone = _.replace(order.shipping[0].address.phone, /\D/g, "");
+  // get the billing object for the current shop on the order (and not hardcoded [0])
+  const shopBilling = order.billing.find(
+    billing => billing.shopId === Reaction.getShopId()
+  ) || {};
+
+  // get the shipping object for the current shop on the order
+  const shopShipping = order.shipping.find(
+    shipping => shipping.shopId === Reaction.getShopId()
```
zenweasel: I don't think this will work as the shipping record currently doesn't contain a shopId? Maybe we just fly with this for now since that should be added shortly, idk.

rymorgan:
In terms of the layout, it looks good to me. I tested the search which all seems to be working just not in tandem with the filtering as noted above.

imports/plugins/included/search-mongo/server/publications/searchresults.js
```diff
@@ -68,12 +71,29 @@ getResults.orders = function (searchTerm, facets, maxResults, userId) {
...
           $options: "i"
         } },
+        { billingCard: {
+          $regex: searchTerm,
+          $options: "i"
+        } },
+        // if parsed phone is empty, default to original, otherwise quering returns false matches
         { billingPhone: {
-          $regex: "^" + searchPhone + "$",
+          $regex: searchPhone || searchTerm,
```
zenweasel: I'm fairly certain removing those regex markers is going to cause a problem

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -1,11 +1,13 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import { Meteor } from "meteor/meteor";
+import { Tracker } from "meteor/tracker";
 import { composeWithTracker } from "@reactioncommerce/reaction-components";
-import { Media, Orders } from "/lib/collections";
+import { Media, Orders, OrderSearch as OrderSearchCollection } from "/lib/collections";
 import { Reaction, i18next } from "/client/api";
 import { Loading } from "/imports/plugins/core/ui/client/components";
```
zenweasel: Is this still needed? Can we/should we use the `Components.Loading` version? I thought that the `Loading` component was included by default?

imports/plugins/core/orders/client/components/orderSearch.js
```diff
@@ -0,0 +1,72 @@
...
+          label="Clear"
+          containerStyle={{ fontWeight: "normal" }}
+          onClick={this.handleClear}
+        />
+      </div>
+    );
+  }
+}
+
+export default compose(withPermissions({ roles: ["orders"] }))(OrderSearch);
```
zenweasel: Shouldn't we be doing `registerComponent` as well?

zenweasel:
It looks like search by email when there is an overlap is not working? This user checked out as anonymous and then added the email in the completed order screen. You may want to check on `marketplace` and see if this is an existing issue.

order_search

zenweasel:
Getting weird results for a phone number. I put in an order with a phone number with the + prefix and then I search for it (the first four number) and I get a completely unrelated match, and the actual match doesn't show up. Possibly related to removing the regex markers as I noted in code comments.

dashboard_orders

zenweasel:
Looks like search by name is also broken for anonymous users. Looking at the Order Search code it looks like names should be getting picked up correctly. Although it could be related to shipping records not having a shopId yet, but billing records do and so they should be picked up.

order_search2

zenweasel:
Looks like icons are messed up in a RTL layout. (Arabic is chosen here)

dashboard_orders

zenweasel:
Looks like certain characters confuse the search

dashboard_orders

zenweasel:
At certain widths the "No Results Found" gets sort of lost.

dashboard_orders

imports/plugins/core/orders/client/components/orderSearch.js
zenweasel: tfw you copy directly from the React docs but forget to change them 😈

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
impactmass: Hmmm.... I thought so too, but couldn't come up with an approach to achieve that. I'll take another stab at it.

imports/plugins/included/search-mongo/server/methods/searchcollections.js
impactmass: Or I can add leave as before on the zeroth index [0]; and add a TODO for when shipping update happens.

imports/plugins/included/search-mongo/server/publications/searchresults.js
impactmass: I'll review this.

impactmass:
Thanks for the reviews. 

Todo Items from current review:

- [ ] Search by email when there is an overlap is not working?
- [ ] Try out setting up the Tracker in composer and pass in handler
- [ ] Special characters search (re: certain characters confuse the search)
- [ ] Fix icons when switched to a RTL layout
- [ ] Investigate comment on anonymous user search
- [ ] Fix "Getting weird results for a phone number" comment
- [ ] Investigate "not finding order for when a user checked out as anonymous and then added the email in the completed order screen". Check on marketplace and see if this is an existing issue.
- [ ] Investigate "When clicking on an order to process it, the window closes when it should just open the details panel."
- Sync with @joykare on the translation issues mentioned on filters and order status

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
impactmass: Update: To be clear, we are saying we want to do this in the compose wrapper function. At the moment that code is in the OrderDashboardContainer.js file

impactmass:
Status: 
* I've cleared the issues with the search queries. Searches with email, name, phone, id, card, variant & product titles should be good now. [See screen cording](https://www.dropbox.com/s/vkbzk9zdkqjl03r/order_search.mov?dl=0)
* Remaining todos are as seen in the above list. 
* [Here](https://github.com/impactmass/reaction/commit/41de34be979173d98916552afcaac75d38ab7b7b), I started on moving to use container wrapper to replace the Tracker logic. Not quite done yet, but can help when this is picked up by someone.
* As also noted elsewhere, we'll need to breakdown the current OrderDashboardContainer file into smaller components/containers, to help when try to add new features.

joykare:
**Status**
- [] Tracker in composer instead of having in the component --- figuring that out [Purely code change for better optimization]

Adding the design review tag coz design changes are complete and works as specced out

rymorgan:
Icons are still messed up on right to left languages. There are some missing translations overall but I'll make another issue for that. 

right to left

rymorgan:
When you are in large view and you search for an order ID, you lost the item count number. 

search-loses-number

rymorgan:
Other than the two issues mentioned above I think this can be done from a UI perspective.

joykare:
@zenweasel ready for a re-review

imports/plugins/core/orders/client/components/orderSearch.js
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * handleChange - handler to call onchange of search input
+   * @param {string} event - event object
+   * @param {string} value - current value of the search input
+   * @return {null} -
+   */
+  handleChange = (event, value) => {
+    this.setState({ query: value });
+
+    if (this.props.handleChange) {
```
zenweasel: We should probably remove this conditional and make this property `isRequired`. There's no good reason to never handle this handler. If it's not there we should probably throw an error rather than just do nothing?

zenweasel:
The issue with anonymous user emails not being searchable is still present.

To reproduce:
1. Check out anonymously
1. On the completed order page add an email
1. Observe the order listed in the order list with the added email
1. Attempt to search for that email
1. Observe that it doesn't show up

anonymous_email

joykare:
@zenweasel @kieha Good to go

imports/plugins/included/search-mongo/server/methods/searchcollections.js
```diff
@@ -167,16 +171,23 @@ export function buildOrderSearchRecord(orderId) {
...
-    region: order.billing[0].address.region,
-    country: order.billing[0].address.country
+    address: shopBilling.address && shopBilling.address.address1,
+    postal: shopBilling.address && shopBilling.address.postal,
+    city: shopBilling.address && shopBilling.address.city,
+    region: shopBilling.address && shopBilling.address.region,
+    country: shopBilling.address && shopBilling.address.country
   };
   orderSearch.shippingAddress = {
     address: order.shipping[0].address.address1,
```
kieha: Remove hard coded shipping objects all over this file.

joykare:
@kieha made suggested fixes

zenweasel:
Testing this now and search doesn't work at all. As I type results are not being updated.

1. DId `reaction reset && reaction`
1. Put in a couple of orders
1. Type in a search term I know doesn't exist
1. Still shows all results even though none match

order_search

aaronjudd:
I can confirm the same results as @zenweasel. Search does not work.
reaction

kieha:
Had tested this before the marketplace merges and it worked. I can, however, also confirm that search doesn't work now.

screen shot 2017-09-14 at 1 42 36 am

joykare:
Fixed @zenweasel @kieha @aaronjudd

zenweasel:
Search by phone number is not working at all

basic_reaction_product

zenweasel:
Same with address
basic_reaction_product

joykare:
@zenweasel is searching by address a new spec for the order search? It wasn't a functionality of the search before.

zenweasel:
@joykare good point. Disregard that comment. I had assumed customer info included address.

joykare:
@zenweasel phone search works now
The only caveat is if you added a `+` on your query your order isn't found (even if the number did have a `+`). Tried a couple of fixes haven't come up with one that works yet. But I think it works right now and it shouldn't be a blocker for this PR merge.

aaronjudd:
cart_checkout Seeing the navigation header during checkout, haven't seen that elsewhere, haven't noticed it before. I suspect it's not related, but @zenweasel can you confirm?

aaronjudd:
I was unable search for testing emails, such as  `aaron+1@`, no result return after `+` , but the other search filtering seems to be working nicely.
reaction_and_order_dashboard_search_by_orderid__customer_info__order_product_info_by_impactmass_ _pull_request__2717_ _reactioncommerce_reaction
reaction_and_order_dashboard_search_by_orderid__customer_info__order_product_info_by_impactmass_ _pull_request__2717_ _reactioncommerce_reaction

zenweasel:
@aaronjudd I see the same header during checkout on an unrelated up-to-date marketplace branch. Is that what you mean?

cart_checkout

aaronjudd:
@zenweasel yes, do we have an issue documenting that?

aaronjudd:
@actually I take that back ... is that expected when logged in as admin? If so, then I'm good to go.

zenweasel:
Created this issue for emails and phone numbers with `+` signs. https://github.com/reactioncommerce/reaction/issues/2867

Otherwise ready to go.

Issue #2265, opened 2017-05-15
No shipping methods are configured. Configure now.


vanloc0301:
I using Reaction. When I tried to at step check out Product.

I can't complete at step: `Select Shipping Option`.
![screenshot_2017-05-15_18-21-25](https://cloud.githubusercontent.com/assets/4168981/26055530/8519e266-399b-11e7-8bbe-23ebe4f2c095.png)

I enabled `Fat Rate` optional.

zenweasel:
If you edit and save, or delete and readd the address, the shipping methods will show and show for all further checkouts.

vanloc0301:
Thanks. It's working.
Can you explain why do this the shipping option is enabled?

zenweasel:
The shipping address is not reactive to the shipping configuration being changed. This only affects the very first checkout for an admin when configuring the shipping methods. You only have to do this once. There is already issue #2232 to change it to be reactive.

aaronjudd:
Re-opening this issue as it's reproducible, almost always on the second purchase in the same session, but after a completed order. All current branches.
```
Node: 6.11.1
NPM: 3.10.10
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 2.0.0
Reaction branch: npm-manifests
Docker: 17.06.0-ce
```

aaronjudd:
![cart_checkout](https://user-images.githubusercontent.com/439959/29094702-1ce5c18e-7c43-11e7-9c87-7d72f69a162a.png)

I have placed an order previously, successfully, but cannot with this user.

zenweasel:
I can't replicate this on `development` using the following steps:

1. Login as admin
1. Place something in the cart and go to checkout
1. Enter address
1. Get "No shipping methods configured"
1. Enable one flat rate method (and enable payment, via Braintree for me)
1. Hit edit address and resave
1. Complete checkout
1. Place another item in cart
1. Checkout as normal and receive no errors

zenweasel:
Can't replicate on `marketplace` either using steps above

spencern:
@zenweasel Try checking out as anon in incognito window in place of step 2.

zenweasel:
@spencern How does that work then? Do I login as admin after I hit the "No Shipping Methods" or?

zenweasel:
Cannot replicate using anonymous user either

spencern:
[...]

PR #2719, opened 2017-08-22
Make order details panel visible.

### LINKED ISSUE #2617: [Marketplace] When using "collapsed" order view, order detail is not always visible.
### Expected behavior
Order detail panel should always be visible

### Actual Behavior
Order detail panel appears not to open because it is hidden. The main list view area should shrink to accomodate the expanded panel. Moving to "expanded" view shows the panel however.
order-detail-panel

1. Place an order
1. Make browser width around 1133px.
1. Go to order view
1. Click on an order
1. Observe that panel appears now to open

### Versions
```
Node: 7.10.0
NPM: 5.0.3
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.0-ce
```

foladipo:
Resolves #2617.

This PR resolves a bug with the orders page. Using the `Collapsed` mode to view the list of orders, the details panel for any specific order wasn't showing up because of some CSS issues.

## Test instructions
- Log in as admin and make one or two orders.
- Use the admin menu (on the right) to go to the `Orders` page.
- Switch to `Collapsed` view.
- Click the `>` beside any listed order and you'll notice that the details panel for that order now show up correctly in the browser.
- Try to resize the browser window. You'll notice that the order details panel stays visible throughout.


zenweasel:
When I first open the orders panel I get this big blank space where there should be none:

thfh7gymvdgsrzdms

zenweasel:
It seems like at wider widths we would want the main order grid to fill all available space and possibly add more fields rather than just having a blank space there.

thfh7gymvdgsrzdms

zenweasel:
At more narrow widths we don't want to lose the order selector there.

thfh7gymvdgsrzdms

@rymorgan Is there an already existing spec for how this order table should operate at all the varying widths? I feel like there is but I am not sure where one would look.

foladipo:
@zenweasel Thanks for the feedback. I have tested this PR and it truly displays the orders table in a weird way at wider screen widths. I'll fix that and update my PR.

As for the orders' table's behaviour at smaller widths, especially in the screenshot you shared, what happens is that the container for the table becomes scrollable. So, if you scroll to the right, the order selector is still visible.

PS: Other than the above, this PR also hides various columns as the screen size reduces in order to fit the orders' table to the screen.

aaronjudd:
@foladipo can you come back to this soon, and fix the merge conflicts, and get this ready for review, or can you close the PR until you are ready?

foladipo:
Thanks for the heads-up @aaronjudd . I put this in the icebox to work on #2265 because of how important it was to fix that bug.

I've now fixed the issues observed by @zenweasel and pushed the commits.

zenweasel:
This is what I am seeing upon opening the order dashboard for the first time

dashboard_orders

zenweasel:
@foladipo I thought the above corrupted view might be dependent on a particular width but that's the way it is at all widths, rendering the order panel unusable.

foladipo:
@zenweasel I've tried to reproduce this both on my computer's screen and a large monitor but can't. Can you tell me the steps you carried out to get the bug?

I also wonder why the order's detail panel (to the right) is actually duplicating itself.

zenweasel:
@foladipo I followed the instructions in the PR

rymorgan:
I'm seeing the same thing. There is something more going on with sidebar than just this. I created an issue for what I'm seeing in general with the sidebar UI in orders. The sidebar for orders is basically unusable in Marketplace right now. 

https://github.com/reactioncommerce/reaction/issues/2775

foladipo:
@rymorgan @zenweasel So, I've figured out where this happens, being guided by what Ryan reported in #2775. Apparently, you can reproduce this bug as follows:

- Add items to your cart and checkout (i.e pay for the products and get to the "Thank you" page.

- While on the checkout page, try to open the orders panel. It will open as a modal, as expected

- Click the blue, "manage order" button beside any new order. Instead of still have a modal, the orders panel is now a main page by itself, and the order details panel at the side is duplicated.

- If you go to the products' dashboard/grid and try to open the orders panel again, it functions as normal.

I'm guessing this was caused by some new code pushed to `marketplace`. I'll investigate it further, try to fix it and add that fix to this one.

foladipo:
@rymorgan and @zenweasel So, I've figured out that it doesn't even happen on the "Thank you" page alone. It happens on all pages that are *NOT* the index. So, if you go to `localhost:XXXX/product/XXXX` or some other non-index URL, you will see the orders panel misbehave.

I eventually traced the source down to [this section of code](https://github.com/reactioncommerce/reaction/pull/2719/commits/231d4007548eae32c496be62441c44e58ee4595d). So, I wish to know your corrections (e.g that that code isn't actually the root of the bug) or suggestions as to how to update that section to fix the bug *WHILE* not breaking other parts of the codebase.

rymorgan:
@mikemurray @kieckhafer Can one you two take a look at this. I can get you up to speed on the bug if needed.

mikemurray:
@foladipo The action view is split up into 2 sections.

[ actionView | actionViewDetail ]

You open the main action view with `Reaction.setActionView()` and the detail view `Reaction.setActionViewDetail()`, which opens a smaller sidepanel inside the big action view panel.

The button that opens the order detail should be doing `Reaction.setActionViewDetail()`.

foladipo:
@mikemurray Yes, the button that opens the order detail does call `Reaction.setActionViewDetail()`.

The problem is that that call eventually gets into [this function](https://github.com/reactioncommerce/reaction/blob/231d4007548eae32c496be62441c44e58ee4595d/client/modules/core/main.js#L625). What I don't understand is why we have to open either `admin/detailView` or `admin/actionView` based on whether or not we are on the index page.

In fact, if I simply open `admin/detailView` always, it fixes this new bug reported by Brent and Ryan.

Now, after doing some more digging, I found out that that section of code was written by Jeremy about two weeks ago in [this commit](https://github.com/reactioncommerce/reaction/commit/b2a192c3fd4076ee5bc12727b9cc6110062fff79).

@jshimko Please can you explain the aim of that section of code?

zenweasel:
@foladipo I don't know that Jeremy will know any thing particularly about that code as it was a merge commit. It was probably to fix a bug but that history has probably been lost with `development`.

Either way, does this current code fix this bug? I am going to re-test now because it's my assumption from your notes that that's the status and it's not marked [WIP].

zenweasel:
So it looks like this is still broken.

@mikemurray Can you help us understand why/when/if we would want to open `admin/adminView` for orders if there is any case for that? Is there no case for this (even for index right?). Maybe this is just some merge weirdness and we can get rid of that if branch.

mikemurray:
@zenweasel @foladipo I'm not sure why there is a check for the index page in the actionView function. It shouldn't care about that at all.

I think at some point that `index` was meant for the internal navigation stack of the action view itself, and not the actual Reaction Router routing/pages.

zenweasel:
So I just removed that branching since no one could understand why it's there and everything about orders seems to work great. A quick check of admin panels seems to be fine as well.

@mikemurray Would you mind taking one last look at this and make sure everything is operating as it should be?

foladipo:
@mikemurray Thanks for the heads up. That bug has now been fixed.

cc @zenweasel

PR #2722, opened 2017-08-23
added adjustments for navbar causing overflow in admin view using default theme

DisruptiveMind:
### Disclaimer
This is very minor but thought I'd contribute if I could. This is related to the default client theme. I'm open for correction if this is my slight of knowledge of _reaction_.

### Problem
When logged in as __admin__ and viewing a product there is a __double scroll-bar__. Please disregard if there is a similar fix.

screen shot 2017-08-22 at 10 15 28 pm

### Proposed Fix
This appears to be caused by the styles and removing them solves it in my isolated case:
- `/imports/plugins/included/default-theme/client/styles/navbar.less` => `.navbar .cart` with `right: -1px`
- `/imports/plugins/included/default-theme/client/styles/layouts/layout.less` => `.page` with `overflow-x: hidden`

#### References
- [navbar.less history](https://github.com/reactioncommerce/reaction/blob/bd61902943977e3470057d970c97c22f1dce8af5/imports/plugins/included/default-theme/client/styles/navbar.less)
- [layout.less history](https://github.com/reactioncommerce/reaction/blob/db4e8d2aa500a9da59f994e64df4f8999ebe6965/imports/plugins/included/default-theme/client/styles/layout/layout.less)

Cheers,

### DIFF
--- a/imports/plugins/included/default-theme/client/styles/layout/layout.less
+++ b/imports/plugins/included/default-theme/client/styles/layout/layout.less
@@ -14,7 +14,6 @@ body.admin {
   .flex-direction(column);
   .flex(1 1 auto);
   height: 100vh;
-  overflow-x: hidden;
 }
 
 .page #main {
@@ -23,7 +22,6 @@ body.admin {
   overflow: auto;
   -webkit-overflow-scrolling: touch;
 }
-
 
 .page.show-settings {
   min-width: 0;
--- a/imports/plugins/included/default-theme/client/styles/navbar.less
+++ b/imports/plugins/included/default-theme/client/styles/navbar.less
@@ -68,7 +68,6 @@
   .flex(0 0 auto);
   position: relative;
   cursor: pointer;
-  right: -1px;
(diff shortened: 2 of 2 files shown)


zenweasel:
@DisruptiveMind I think this **might** be fixed in `development` already. Have you tested it there?

DisruptiveMind:
@zenweasel I just tested on `development`, it seems the issue still exists there as well.

nbaldwin:
This was fixed as part of https://github.com/reactioncommerce/reaction/pull/2653 but the PR was only merged into marketplace, not development.

zenweasel:
Thanks for that @nbaldwin , I thought I remembered this from somewhere.

I doubt we will be moving this into `development` as it is currently deprecated. I'm not sure when we are going to do the next bugfix release.

Thanks very much for the PR @DisruptiveMind

DisruptiveMind:
Sounds good. I'll close this PR then. Thanks guys.

PR #2733, opened 2017-08-24
[Marketplace] Fix back button behaving strangely on PDP (#2660)

### LINKED ISSUE #2660: Back button behaves strangely on PDP
### Expected behavior
Back button should take you to the previous "page"

### Actual Behavior
If you have clicked on options, will take you through the history of options you clicked on

1. Go to "Basic Reaction Product"
1. Click on Red and add to cart
1. Click on Green and add to cart
1. Click on Back Button in browser
1. Notice Red is now selected rather than being taken to the Product Grid.

### Versions
```

Node: 7.10.0
NPM: 5.0.3
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.0-ce
```

jm-wanja:
Resolves #2660 

**NB** This PR fixes the bug from #2660 where the browser back button behaves strangely on PDP where it was directing to the previously selected variant option instead of product grid when clicked. 

**How To Test**

1. Go to "Basic Reaction Product"
2. Click on Red and add to cart
3. Click on Green and add to cart
4. Click on Back Button in browser
5. Notice that you are being taken to the Product Grid.


aaronjudd:
I did test this, it does work.  My gut says... `hack`!  Doesn't feel like this is quite the correct fix, and that we need to fix the way the variant is adding history.  @jshimko @mikemurray 

We've also agreed in the past, that clicking the variant should **not** change the url...

jm-wanja:
I had tried to not have the variant id on the route and I noticed that not having the id on the url when we click a variant results to not being able to add to cart. It raises the `Please choose options before adding to cart` warning.

mikemurray:
Yes, this is a hack, and could lead to other issues. We should come up with another solution that doesn't involve push, pop or replace state of the router when selecting variants.

1. If you come in with a url that is a variant, select that variant by default (usually from cart URLs)
2. Selecting a variant or a child variant does not change the url, only sets `Reaction.setCurrentVariant(varaintId) // It should do this now already`

With this solution we will have to ensure other react components related to the PDP are not accessing the variantId from the url, but instead get it from `Reaction.selectedProduct()` or `Reaction.selectedVariant()`

jm-wanja:
@aaronjudd mikemurray This is noted, I see how it could lead to other issues. I'm going to go back and implement a better solution following your recommendation.

jm-wanja:
@mikemurray  doesn't  `Reaction.Router.go("product", {
        handle: this.productHandle,
        variantId: variant._id
      }, {
        as: Reaction.Router.getQueryParam("as")
      });` change the url by adding the variant id to the history?

PR #2737, opened 2017-08-24
update groups to allow invitations

kieckhafer:
updating group roles for the `payments-stripe` package to fix / allow invitations using the dashboard accounts permissions forms.

### DIFF
--- a/imports/plugins/included/payments-stripe/server/startup/startup.js
+++ b/imports/plugins/included/payments-stripe/server/startup/startup.js
@@ -3,7 +3,7 @@ import { Reaction, Hooks } from "/server/api";
 Hooks.Events.add("afterCoreInit", () => {
   Reaction.addRolesToGroups({
     allShops: true,
-    groups: ["customer", "guest"],
+    groups: ["owner", "shop manager", "customer", "guest"],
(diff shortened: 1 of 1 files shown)


impactmass:
bitHound is also preventing merging this

spencern:
@impactmass that's a good idea as an option, especially when multiple plugins may be competing to add roles.

Another option would be to check to see if a user has `admin` or `owner` first and always permit them to add users to groups

spencern:
Not sure if we have this or not right now, but we should have a way to turn `admin` on or off for a particular group as well.

impactmass:
Yes, we can check to see if a user has `admin` or `owner` first, and permit them to add group. And also we can enable turning on/off `admin` for a group

A concern is we want to try and avoid another situation where have a lot of things dependent on hasAdmin. The we end up with the situation of once you grant people admin, then can do a lot
We have quite a number of actions that are dependent on hasAdmin.

spencern:
@impactmass That's a good point.
I think we should consider deprecating use of hasAdmin and moving any role checks that depend on that towards a more "group oriented" permission approach. This goes along with defining a few more sensible default groups that can take over the responsibility that "hasAdmin" currently plays.

Issue #2669, opened 2017-08-14
[Marketplace] Can't check out as newly invited shop owner


zenweasel:
### Expected behavior
Not sure if a new shop owner is supposed to configure shipping or not?

### Actual Behavior
When attempting to check out I get the "Shipping Method Not Configured" and clicking on the configure links takes me to a corrupted settings panel
![cart_checkout](https://user-images.githubusercontent.com/72819/29254992-c5d70be6-80cf-11e7-82a6-8dd4be733c22.png)

### Steps to Reproduce the Behavior
1. invite a new shop owner
1. As shop owner attempt to configure shipping

### Versions
```
Node: 7.10.0
NPM: 5.0.3
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.0-ce
```

spencern:
This is by design right now as my current understanding is that merchants should not be configuring their own shipping charges, but if that understanding is wrong then we can add the shipping package to the default merchant packages array. This can be configured per marketplace as it's stored in the database, but currently there is no UI for managing it.

zenweasel:
@spencern Sure but a) I don't get any shipping and b) that's probably not the UI that should show? Should I change the issue then to be "Correct shipping method not used when checking out" or something similar to reflect the actual problem?

zenweasel:
Closed via #2741

Issue #2713, opened 2017-08-22
[Marketplace] Checking out as merchant shop owner doesn't load shipping rates


spencern:
### Expected behavior
I should be able to checkout as any user, including merchant users.

### Actual Behavior
Checking out as a merchant shop owner, I receive no shipping methods configured error, clicking into the dashboard causes errors because merchant shops shipping packages are disabled by default.
![image](https://user-images.githubusercontent.com/1203639/29546486-ffa4bdd6-86b1-11e7-840f-17fa21d04178.png)

### Steps to Reproduce the Behavior
Invite a new shop owner
Accept invite and then add a product to cart and try to checkout as a shop owner

### Versions
```
Node: 7.8.0
NPM: 5.0.1
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: spencer-stripe-connect-merchant-signup
Docker: 17.06.0-ce
```

zenweasel:
Closed via #2741

PR #2741, opened 2017-08-25
Always grab shipping from primary shop regardless of current shop (Resolves #2732/#2713)

### LINKED ISSUE #2732: [Marketplace] Shipping methods don't show up when checking out in non-primary shop
**Note** I am not 100% clear on what are the valid use cases here. This is checking out on a second shop as Marketplace admin, but that's the only way you can see product from both Primary and non-Primary shops.

### Expected behavior
Shipping methods should show up

### Actual Behavior
They don't. If you switch to primary shop, they do

1. Full reset. Once you successfully check out it appears (?) it never happens again
1. Set up a second shop and add a product
1. As marketplace admin switch to the second shop
1. Place an item in your cart
1. Attempt to check out
1. Observe that you receive the "Shipping method not configured" error and the Shipping config window opens but is broken (which it shouldn't even be there).
1. Switch back to Reaction and re-save your address.
1. Observe shipping method now appear

### Versions
```
Node: 7.10.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.0-ce
```

zenweasel:
Resolves #2732 and #2713 

### Steps to Test
1. Full reset.
1. Set up a second shop and add a product as shop admin
1. Try to checkout as shop admin. Observe that shipping methods appear
1. As marketplace admin switch to the second shop
1. Place an item in your cart
1. Attempt to check out
1. Observe that the shipping methods appear

### DIFF
--- a/imports/plugins/core/shipping/client/templates/checkout/shipping.js
+++ b/imports/plugins/core/shipping/client/templates/checkout/shipping.js
@@ -46,7 +46,8 @@ function enabledShipping() {
   const enabledShippingArr = [];
   const apps = Reaction.Apps({
     provides: "shippingSettings",
-    enabled: true
+    enabled: true,
+    shopId: Reaction.getPrimaryShopId()
   });
   for (const app of apps) {
     if (app.enabled === true) enabledShippingArr.push(app);
@@ -65,7 +66,8 @@ Template.coreCheckoutShipping.onCreated(function () {
   const shippingOpts = {
     provides: "settings",
     name: "settings/shipping",
-    template: "shippingSettings"
+    template: "shippingSettings",
+    shopId: Reaction.getPrimaryShopId()
   };
 
   // If shipping not set, show shipping settings dashboard
--- a/imports/plugins/core/shipping/client/templates/checkout/shipping.js
+++ b/imports/plugins/core/shipping/client/templates/checkout/shipping.js
@@ -112,8 +112,19 @@ Template.coreCheckoutShipping.helpers({
     return false;
   },
 
+  /**
+   * Template helper that checks to see if the user has permissions for the shop
+   * responsible for shipping rates. This is the primary shop unless
+   * `merchantShippingRates` is enabled in marketplace
+   * @method isAdmin
+   * @return {Boolean} true if the user has admin access, otherwise false
+   */
   isAdmin() {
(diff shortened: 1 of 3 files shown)


zenweasel:
Also resolves #2669 (which is a dupe)

spencern:
We could make this conditional based on the [`merchantShippingRates` flag in marketplace](https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/included/marketplace/register.js#L61). That's how we'll need to do this once we permit each shop to define their own shipping options.

Since this does fix those issues and we're not ready to introduce that functionality yet, I'm going to add a TODO so that we remember to update this when we permit different shipping options per shop.

imports/plugins/core/shipping/client/templates/checkout/shipping.js
```diff
@@ -65,7 +66,8 @@ Template.coreCheckoutShipping.onCreated(function () {
   const shippingOpts = {
     provides: "settings",
     name: "settings/shipping",
-    template: "shippingSettings"
+    template: "shippingSettings",
+    shopId: Reaction.getPrimaryShopId()
```
spencern: Could this potentially permit a non-marketplace-owner to adjust the shipping settings for marketplace or does it just prevent the other shipping settings apps from showing up? 
AFAIK they shouldn't have been showing up anyway because they would be disabled, but these changes seem to indicate differently.

zenweasel: That's a good question. They only show up when you get the "No shipping methods configure. Configure now" with the link. Ideally this should only show up when you actually have the permissions to modify the shipping settings which is currently if you are a shop admin correct? But it's only currently checking for `["owner", "admin"]);`

I wonder if we can't make this only show up when you actually have permissions to configure shipping? i.e. are Marketplace Admin?

zenweasel: For now, I can just back out that change, and we can deal with that separately?

spencern: Yeah, I'm testing, I'll test with and without that, and approve one way or the other

PR #2742, opened 2017-08-25
Fix on Localized elements not updating on currency/language change

### LINKED ISSUE #2593: i18n and currency elements not all updating on change

rhenshaw56:
**Resolves #2593**

TEST BY:

- Cloning this branch
-  Setup `reaction.json` to have the open exchange rates API key or use the [standard `reaction.json`](https://github.com/reactioncommerce/rc-config-files/blob/master/reaction.json)
- Start reaction and login as admin
- Change currency or i18n in a page where there are multiple localized elements and notice that components update on currency/language change
- Test this also as a guest/non-admin user


client/modules/i18n/main.js
```diff
@@ -119,7 +119,7 @@ Meteor.startup(() => {
           locale.language = getBrowserLanguage();
           moment.locale(locale.language);
           // flag in case the locale currency isn't enabled
-          locale.currencyEnabled = locale.currencyEnabled;
+          locale.currencyEnabled = locale.currency.enabled;
```
zenweasel: I feel like @jm-wanja just changed this to fix another bug. Can you check with her?

client/modules/core/main.js
```diff
@@ -755,4 +751,4 @@ function createCountryCollection(countries) {
     Countries.insert(country);
   }
   return countryOptions;
-}
+}
```
zenweasel: Missing newline

zenweasel:
@rhenshaw56 This PR could probably use a better title.

client/modules/i18n/main.js
rhenshaw56: sure

rhenshaw56: @zenweasel  the reason for the change was because `locale` does not have a `currencyEnabled` field so it was always `undefined`. Just checked in with her and had her test if it affected the bug she worked on. She said it did not.

zenweasel:
@aaronjudd Can you take a look at this when you have a chance?

aaronjudd:
Verified that logged out / non admin currency and language do seem to be updating.  It's a bit hard to tell with translations at the moment though as there are still a lot I'm seeing all over that aren't translated here ("Shop Settings, and more, where I know there ~were~ translations).

Here's some clear examples though where I'm not seeing currency translating when I am logged in as an admin:
basic_reaction_product
basic_reaction_product

zenweasel:
So it was my (possibly incorrect) understanding that elements on the PDP should always be in the shop currency for an admin.

Either way I see that it's consistent, but want to make sure we correct to the right way.

aaronjudd:
@zenweasel let's open a separate issue for the  side bar currency issue.. and then merge this..

PR #2744, opened 2017-08-25
Order managers should be able to see a risk evaluation for an order

### LINKED ISSUE #2322: Order managers should be able to see a risk evaluation for an order
Users with permissions to see the order details view should be able to see a risk evaluation for an order.

Orders with a normal risk evaluation should not be highlighted abnormally.

Orders with an `elevated` or `high` risk evaluation should be highlighted and there should be a large notification present on the order details page as well as a badge on the order list page.

Orders with a `high` risk evaluation should require the order manager to acknowledge that the order is a high risk order before they may capture the payment.

#### Resources
* [Stripe Risk Evaluation Docs](https://stripe.com/docs/radar/risk-evaluation)

impactmass:
Resolves #2322. 

**How To Test:**
* Using Stripe as payment gateway, make an order. Use the card number `4000000000009235`. This is a test card from Stripe that succeeds with an elevated risk response. See comment below.
* Check the Order Dashboard for that order, you should see it flagged as elevated.
* Switch to Example Payment Method from Stripe, and make another order using the same card above. With this, the Example Payment flags the charge as high risk.
* Check the Order Dashboard for that order, you should see it flagged as high risk (red).
* Also confirm that there is a warning alert before capture of an elevated or high risk order in the Order summary side panel.
* Confirm that making orders with other normal cards (e.g 424242424242...) works as before without any side effect from this PR

NB: (i) The other included payment methods can be updated later based on how each processes risk evaluation. I updated only Stripe package & Example Method in this PR.
(ii) Bulk capturing doesn't popup an alert on risky orders capturing at the moment. That'll be handled in #2762 

**More Details about this PR:**
- Adds a risk level field on the paymentMethod schema (marked as optional). It will allow one of three values `normal`, `elevated` or `high`. So each payment package (when we come around to update them) will only need to set this field based on how they process risk levels. The example used for Stripe & Example payment here can be followed.
- Based on the value of this filed we display badges on the UI.
- We also show an alert when capturing an order that's flagged as risky.

Link
* [Stripe Risk Evaluation Docs](https://stripe.com/docs/radar/risk-evaluation)


impactmass:
Ready.
Updated look according to design
screenshot 2017-08-28 14 57 42

rymorgan:
@impactmass Looks good to me. I may decide to brighten up the alert box background color in the future but this is good to go as far as I'm concerned.

aaronjudd:
@impactmass can you give instruction for testing various risk scenarios, ie: how do I get a high risk order.  Also, Stripe Connect may only work for marketplaces, but we need to test against all payment methods..

impactmass:
Okay. I'll dig more into Stripe Connect's doc to see if there's a testing way to generate risk flag (last time I checked I didn't see that).
After resolving #2717

impactmass:
1). At the moment, I haven't found a way to make a payment flagged as risky by Stripe. 
2) Re the other payment methods getting updated for risk evaluation capturing, I'll make a separate issue to update our current payment methods. I only limited this to Connect to start.  
If it's preferable to get all of the payment methods updated before merging in the UI for the risk badges, I'll make this WIP (or temporarily close it).

aaronjudd:
I'm not looking for all other payment methods to enable risk evaluation -> because most do not have this functionality. What I'm looking is to make sure it's a generic enough implementation that they can add this functionality if needed -> but most importantly -> do the other methods all work 100% as they are and nothing breaks and there are no artifacts from this.  Since I don't have the instructions to test... I couldn't test for myself if the other methods would break.

aaronjudd:
If we do want to add to other payment methods.. that'd be a separate issue for each (we'd have to identify which support this).  It would be nice to have one method besides connect that a single shop implementation would be able to use.

impactmass:
Oh okay great. Thanks @aaronjudd

It's a generic enough implementation. I'll update the description to add what was done, but I've:

_ Added a risk level field on the paymentMethod schema (marked as optional). It will allow one of three values `normal`, `elevated` or `high`. So each payment package (when we come around to update them) will only need to set this field based on how they process risk levels. The example I used for Stripe can be followed.
_ Based on the value of this filed we display badges on the UI.

Also, the current stripe package (which I used here) works for single shops as well. The risk level is not a feature that's connected only to the Connect feature. (I'll add this to the description too).

spencern:
@impactmass See: https://stripe.com/docs/testing#cards for testign cards with non-standard responses.
Specifically: 
> `4000000000009235` Charge succeeds with a risk_level of elevated and placed into review.

Stripe's [Radar Docs](https://stripe.com/docs/radar/risk-evaluation) has some information about risk evaluation, but I'd imagine you've seen that. I don't see a way to trigger a high-risk eval with a card number. Might be able to do it with a token.

aaronjudd:
@impactmass will you have time resolve conflicts and bring this up to speed before next week, or should we postpone this for a following release?  If it's close, let's get it done..

impactmass:
yes, I'll bring this up and fix conflicts with base branch, and finish up on what is left asap

imports/plugins/core/orders/client/components/orderTable.js
```diff
@@ -155,6 +158,13 @@ class OrderTable extends Component {
             className="rui-order-avatar"
           />
           <strong>{order.shipping[0].address.fullName}</strong> | {emailAddress}
```
spencern: Not an issue created by this PR, and not a blocker for me on this specific PR, but very soon I think we should start rejecting PRs that don't address the `shipping[0]` or `billing[0]` issue as it's a strong indicator that that part of the code is not marketplace ready, and if we're in there, we should fix it anyway.

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -760,6 +760,7 @@ class OrderDashboardContainer extends Component {
         } else {
           // TODO: send these orders in batch as an array. This would entail re-writing the
           // "orders/capturePayments" method to receive an array of orders as a param.
+          console.log("capturePayments - OrderDashboardContainer");
```
spencern: no `console`

impactmass:
thanks for the initial comments. I'm still doing some final tests before removing the WIP, hence the random console.log, but I'll soon be done with that. I'll also add a quick doc update, and remove the WIP.

imports/plugins/core/orders/client/components/orderTable.js
impactmass: those [0] are being removed/updated in #2813

impactmass:
@spencern Ready to test. I updated the test instructions to include testing with Example Payment Provider as well.

imports/plugins/included/payments-example/server/methods/exampleapi.js
```diff
@@ -3,7 +3,7 @@ import { SimpleSchema } from "meteor/aldeed:simple-schema";
 import { Random } from "meteor/random";
 
 // Test card to use to add risk level flag for testing purposes only.
-export const TESTRISKYCARD = "4000000000009235";
+export const RISKY_TEST_CARD = "4000000000009235";
```
spencern: this is a great name for a test card

Issue #2740, opened 2017-08-25
Move accounts package from /client/modules/accounts into /imports/plugins/core/accounts


kieckhafer:
Some of our account related action components (Password reset, for instance) are located inside the `/client/modules/accounts` directory, and not in our `/imports/plugins/core/accounts` directory.

It seems that this causes an issue with load order, and the components are not available to be replaced using `replaceComponent` because of this.

We should move all of the accounts components into the imports directory.

jshimko:
Resolved by #2745

Issue #2493, opened 2017-06-22
no UI for users to update their email


jshimko:
### Expected behavior

Expect that any user can update their email.  

### Actual Behavior

There is nowhere in the UI to update your email (unless I'm totally missing it).  Editing the database is currently the only option. This is particularly a problem for initial shop admins where a random email can potentially be generated for them.  Once that has happened, there's no way for them to update it to their actual email.  They have to delete the user and force a new default admin creation with proper configs.

### Versions

Node: 8.1.2
NPM: 5.0.3
Meteor Node: 4.8.3
Meteor NPM: 4.6.1
Reaction CLI: 0.10.0
Reaction: 1.3.0
Reaction branch: development
Docker: 17.03.1-ce

jshimko:
Resolved by #2745

PR #2745, opened 2017-08-25
Moving accounts package into /imports, adding email update to profile

kieckhafer:
Resolves issue #2740 
Resolves issue #2493 
Resolves issue #2778

Move some account related components from Move accounts package from `/client/modules/accounts` into `/imports/plugins/core/accounts`. Follow up task in #2746 .

Add email update fields to profile page

Pull up `ReactionAvatar` component from client project into the core

Follow-up tasks:
~Email copy for new email: #2778~


server/methods/accounts/accounts.js
```diff
@@ -59,6 +60,74 @@ export function verifyAccount(email, token) {
...
+  const user = Meteor.user();
+
+  // Remove email address from user
+  MeteorAccounts.removeEmail(user._id, email);
+
+  // Verify new address
+  sendUpdatedVerificationEmail(user._id);
+
+  // Sync users and accounts collections
+  syncUsersAndAccounts();
```
kieckhafer: I'm not sure this is the best way to sync `users` and `accounts`... any input would be appreciated

server/methods/accounts/accounts.js
```diff
@@ -59,6 +60,74 @@ export function verifyAccount(email, token) {
...
 }
 
+
+/**
+ * updateEmailAddress
+ * @summary update a users email address
+ * @param {String} email - user email
+ * @returns {Boolean} - returns boolean
+ */
+export function updateEmailAddress(email) {
```
kieckhafer: After a bit of research, this seems to be the way people update email addresses on Meteor, first by adding a new one, then removing the old one. There is no callback on meteor's `Accounts.addEmail`, so I think you kind of have to send it back tot he client and do it this way.

aaronjudd:
bitHound - Dependencies — 1 failing dependency.

imports/plugins/core/accounts/client/components/email/updateEmail.js
```diff
@@ -0,0 +1,121 @@
...
+    email: PropTypes.string,
+    uniqueId: PropTypes.string
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      email: props.email,
+      newEmail: false,
```
jshimko: `this.state.newEmail` isn't used anywhere.

imports/plugins/core/accounts/client/components/email/updateEmail.js
```diff
@@ -0,0 +1,121 @@
...
+      showSpinner: false
+    };
+
+    this.handleFieldChange = this.handleFieldChange.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
```
jshimko: These bindings aren't needed because you defined those as arrow functions (accomplishes the same thing).

imports/plugins/core/accounts/client/components/email/updateEmail.js
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  componentWillReceiveProps() {
+    this.setState({ showSpinner: false });
+  }
```
jshimko: Is this needed?  I don't see how it does anything.

imports/plugins/core/accounts/client/components/email/updateEmail.js
```diff
@@ -0,0 +1,121 @@
...
+
+  handleSubmit = (event) => {
+    event.preventDefault;
+
+    this.setState({ showSpinner: true });
+
+    const { email } = this.state;
+
+    Meteor.call("accounts/validation/email", email, false, (result, error) => {
+      if (error.error) {
```
jshimko: All server communication or data fetching should always be in a container.  This should just be passed in as an event handler prop.  The UI component shouldn't need to know/care about how something gets done.  It just handles the state of the UI.

imports/plugins/core/accounts/client/components/email/updateEmail.js
```diff
@@ -0,0 +1,121 @@
...
+        Alerts.toast(i18next.t("accountsUI.error.invalidEmail", { err: error.reaston }), "error");
+        this.setState({ showSpinner: false });
+        return null;
+      }
+
+      Meteor.call("accounts/updateEmailAddress", email, (err) => {
+        if (err) {
+          Alerts.toast(i18next.t("accountsUI.error.emailAlreadyExists", { err: error.message }), "error");
```
jshimko: Wrong `error` object here in the Alert.  Should be `err` because `error` is from the upper scope.

imports/plugins/core/accounts/client/components/email/updateEmail.js
```diff
@@ -0,0 +1,121 @@
...
+          });
+        }
+      });
+    });
+
+    return null;
+  }
+
+  renderSubmitButton() {
+    if (this.state.showSpinner === true) {
```
jshimko: This button component could be simplified into a single instance.

```js
  render() {
    const { showSpinner } = this.state;

    return (
      <div>
        <Components.TextField
          i18nKeyLabel="accountsUI.emailAddress"
          label="Email Address"
          name="email"
          type="email"
          id={`email-${this.props.uniqueId}`}
          value={this.state.email}
          onChange={this.handleFieldChange}
        />
        <Components.Button
          bezelStyle={"solid"}
          icon={showSpinner && "fa fa-spin fa-circle-o-notch"}
          i18nKeyLabel={showSpinner ? "accountsUI.updatingEmailAddress" : "accountsUI.updateEmailAddress"}
          label={showSpinner ? "Updating Email Address" : "Update Email Address"}
          status={"primary"}
          onClick={this.handleSubmit}
          disabled={this.state.email === this.props.email}
        />
      </div>
    );
  }
```

imports/plugins/core/accounts/client/components/email/updateEmail.js
```diff
@@ -0,0 +1,121 @@
...
+          value={this.state.email}
+          onChange={this.handleFieldChange}
+        />
+        {this.renderSubmitButton()}
+      </div>
+    );
+  }
+}
+
+export default UpdateEmail;
```
jshimko: All internal components should be registered with the components API.

imports/plugins/core/accounts/client/templates/members/member.js
```diff
@@ -141,6 +143,15 @@ Template.memberSettings.helpers({
       component: ManageGroups,
       ...Template.currentData()
     };
+  },
+
+  ReactionAvatar() {
+    const user = Accounts.findOne(Meteor.userId());
```
jshimko: All of this data fetching logic should be in a container for the `ReactionAvatar` component, not the Blaze template.  That makes overriding the React component that much more complicated.

imports/plugins/core/accounts/client/templates/profile/profile.js
```diff
@@ -5,6 +5,9 @@ import { ReactiveVar } from "meteor/reactive-var";
 import { Reaction } from "/client/api";
 import { i18next } from  "/client/api";
 import * as Collections from "/lib/collections";
+import UpdateEmail from "/imports/plugins/core/accounts/client/components/email/updateEmail";
```
jshimko: This should be a registered component so it can be overridden.

imports/plugins/core/accounts/client/templates/profile/profile.js
```diff
@@ -26,6 +29,24 @@ Template.accountProfile.onCreated(() => {
  */
 Template.accountProfile.helpers({
 
+  UpdateEmail() {
```
jshimko: In both of these, the data fetching logic should be in a container, not the Blaze template.

imports/plugins/core/ui/client/components/avatar/avatar.js
```diff
@@ -0,0 +1,83 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import Avatar from "react-avatar";
+import classnames from "classnames/dedupe";
+import { registerComponent } from "@reactioncommerce/reaction-components";
+
+class ReactionAvatar extends Component {
```
jshimko: Any component that only has a render function should be a simple stateless function.  None of the React lifecycle methods or state handling are being used, so a Component class adds unnecessary overhead. 
This also depends on an email being passed in, so it should have a container wrapping it (currently being done by a Blaze template)

server/api/core/accounts/password.js
```diff
@@ -231,3 +231,118 @@ export function sendVerificationEmail(userId, email) {
...
+
+    if (!address) {
+      const msg = "No unverified email addresses found.";
+      Logger.error(msg);
+      throw new Meteor.Error("no-unverified-address", msg);
+    }
+  }
+
+  // make sure we have a valid address
+  if (!address || !_.includes(_.map(user.emails || [], "address"), address)) {
```
jshimko: `_.includes` should be `Array.includes` (I know you likely copied this from the other email verification function that I wrote.  Just trying to be do it right moving forward. :))

server/methods/accounts/accounts.js
jshimko: This seems sensible to me.

kieckhafer:
@aaronjudd The `react-measure v2.0` upgrade is a breaking change for us, something that I think would require a bit of a rewrite of the MediaGallery that's currently on the PDP. It might be better to hold off on updating just to get rid of that warning if we are planning on doing an overall update to MediaGallery: https://github.com/reactioncommerce/reaction/issues/2792#issuecomment-327322049

kieckhafer:
`react-measure` has been updated to 2.0.

Would love to get this merged so we're able to pull the changes into a client project.

PR #2747, opened 2017-08-25
Adds additional fields to tag schema

mikemurray:
- Adds optional type field for better categorization needs
- Adds optional metafields array for instances where schema may fall short
- Adds isVisible field, useful for hiding a tag from showing in a list
- Adds groups array, allowing for hiding tags by specific user groups

### DIFF
--- a/lib/collections/schemas/tags.js
+++ b/lib/collections/schemas/tags.js
@@ -1,5 +1,6 @@
 import { SimpleSchema } from "meteor/aldeed:simple-schema";
 import { shopIdAutoValue } from "./helpers";
+import { Metafield } from "./metafield";
 
 /**
  * Tag Schema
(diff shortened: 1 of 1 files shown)


lib/collections/schemas/tags.js
```diff
@@ -17,6 +18,14 @@ export const Tag = new SimpleSchema({
   slug: {
     type: String
   },
+  type: {
```
zenweasel: type seems sort of vague? Just reading this I am not clear how I would use it?

aaronjudd: `type` has been (my)/the preferred name for anything that determines functionality, and should be a fairly standard behavior expectation across our schemas. 👍

aaronjudd:
`isVisible` feels like it should actually be a security group related filter, but... I know we've used the same convention elsewhere.

jshimko:
After merging this into marketplace, I now see this error when restarting a shop with an existing database.

reaction

Missed migration or something?

mikemurray:
Probably the `isVisible` field, since it's not marked optional.

PR #2645, opened 2017-08-07
Completed orders should break items out by Shop

zenweasel:
Closes #2308 and #2245

My first React component replacement from scratch so I open to input on layout, structure, etc. Layout is responsive thanks to help from @kieckhafer. Also first time I have done any HTML of significance so there might be problems there.

#### Single Shop (for just the completed page functionality)

1. Place an order
1. Proceed to the completed order page

#### Multi-shop

**[Update]** Slightly shorter instructions now that you can see products from both shops w/o switching shops.

1. Login as admin (Marketplace admin)
1. Set up shipping (I've only been testing with Flat Rate but YMMV)
1. Enable/Configure Stripe (keys are in `reaction.json`)
1. Click on "Shop" settings in the Dashboard
1. Click on "Marketplace"
1. Enter a name and email in the "Invite Owner Form" and click "Send Invitation"
1. Copy the URL from the invite email
1. Open that URL in an incognito window (you will want to keep your logged in session)
1. Open "My Shop Settings"
1. Click on "Start Accepting Payment". You will be taken to Stripe's site
1. **Don't fill out this form**. Click on the "Skip this account form" link at the top, this will return you to your local site and Stripe is now connected 
1. While still a Shop Admin, create a product and publish
1. In your non-incognito browser window as Marketplace admin, place a Reaction item in your cart
1. Add your newly created Shop product to your cart
1. Checkout as normal

**Note that "Estimated Ship Date" is not implemented on the back end so that's why it doesn't display here.

**Another Note** There is currently an issue with the way that discounts are being stored on multi-shop orders so that discounts are doubled (once per shop).

Issue #2630, opened 2017-08-03
[Marketplace] Unable to change Email Settings


aaronjudd:
### Expected behavior

No Errors, Editable email settings

### Actual Behavior

There appears to be no errors in marketplace and you are unable to edit the email settings.

[marketplace]
![reaction](https://user-images.githubusercontent.com/439959/28933826-744285e8-7833-11e7-8884-d27378b789cd.png)
[seun-change-owner-2188]
![reaction](https://user-images.githubusercontent.com/439959/28933508-21c41512-7832-11e7-92a7-a9624f08b8cf.png)

### Steps to Reproduce the Behavior
Default install, with or without `reaction.json` settings, attempt to edit email settings.
### Versions
Node: 6.11.1
NPM: 3.10.10
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: seun-change-owner-2188 (and markeplace)
Docker: 17.06.0-ce

spencern:
@kieckhafer @impactmass this looks like the same type of issue that you guys solved for the other files that hadn't migrated to the new `Components.component` style

kieckhafer:
Closing via the PR's above

zenweasel:
Im retesting this now and can reconfigure mail w/o issue

Issue #1442, opened 2016-09-26
Email settings in ENV vars override those in dashboard


mikemurray:
### Expected behavior

If credentials from ENV vars will override the mail settings in the dashboard, then they should be displayed in the dashboard.

### Actual Behavior
- Credentials from ENV override settings in dashboard, but the dashboard does not reflect that.

### Steps to Reproduce the Behavior
- Export bad mail settings
- Reset and run reaction
- See errors in terminal when emails fail to send

```
export MAIL_URL="smtp://postmaster%40sandboxXXXX.mailgun.org:BAD_PASSWORD@smtp.mailgun.org:587"
```

rstegg:
And email form should probably be more consistent, keeping API Key and credentials [as an option or dully marked _not all inputs are always necessary_].  For example, send grid can use API key or user credentials.

aaronjudd:
Closing - insufficient details.

PR #2751, opened 2017-08-28
fix local SMTP email sending

jshimko:
Fixes the ability to use an SMTP provider on localhost (both UI and methods would throw errors without a username/password provided).  PR can be tested with [Maildev](http://danfarrelly.nyc/MailDev/).

```sh
npm i -g maildev

# Start mail server and inbox UI 
maildev

# SMTP server running at smtp://localhost:1025
# UI running at http://localhost:1080

# grab this branch
reaction init -b jshimko-fix-local-smtp-sending mail-fix
cd mail-fix
reaction
```
- Go to the shop's email settings
- pick `Maildev` from the dropdown
- click save 
- trigger an email send from somewhere in the Reaction UI 
- note the received email in Maildev's inbox UI (http://localhost:1080)

This also adds the ability to choose "custom" from the dropdown and enter a mail server host without user/password.  You obviously shouldn't ever do that in production, but at least now you can do things like run a custom SMTP server on localhost for development (although I highly recommend just using Maildev as the server and use the Maildev preset to connect to it).


aaronjudd:
Does this replace https://github.com/reactioncommerce/reaction/pull/2645/commits/29196aede286ba4ef8ec04f519083e8846cdc21c or should that get manually added to this?

jshimko:
Well, there's at least a small issue with that update.  It doesn't account for `settings.service` to be equal to `custom` (which would return nothing from `getServiceConfig`).  So the goal was sound, but the implementation needs some additional logic handling.  

That's also data getting passed into a UI component, so that should be happening in the container (where the [providers object already exists](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/email/client/containers/emailSettings.js#L9) anyway)

Give me a couple mins and I'll update my PR.

jshimko:
Actually, nevermind.  I had already done it (with the required logic for the custom setting that I mentioned).

https://github.com/reactioncommerce/reaction/blob/61c61305b5b3ab63031387546a570a6a91631443/imports/plugins/core/email/client/components/emailSettings.js#L31-L33

aaronjudd:
Could you also test for #1442 and #2630?

jshimko:
#1442 was definitely out of scope of this issue, so I'm positive that hasn't changed.  I could fix it, but it's definitely a bit more work and I was just putting out the immediate fire of "can't use a mail server on localhost".  Fixing that issue would require creating pub/sub to push down the parsed values of `MAIL_URL` and/or `Meteor.settings.MAIL_URL` so the UI could display them.  Not anything _too_ crazy, but not 5 mins of work for sure.

#2630 can't possibly still be broken or this PR wouldn't work at all (since it's a branch off of marketplace).  The console errors shown in that issue appear to be something with the components API that must've been fixed by someone.  Wasn't that closed at some point anyway?

zenweasel:
When I switch to maildev I get this:
reaction

and this error in the console:
```
  Error: Invalid login: 500 Error: command not recognized
      at Object.Future.wait (/Users/brenthoover/.meteor/packages/meteor-tool/.1.5.1.krqt03++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/dev_bundle/server-lib/node_modules/fibers/future.js:449:15)
      at packages/meteor.js:213:24
      at [object Object].emailVerifySettings (server/methods/email.js:49:14)
      at packages/check.js:129:16
```

Issue #2328, opened 2017-05-22
Weird behavior on checkout login/signup


kieha:
This happens when logging in/signing up on the checkout panel.

#### Scenario A

When logging in/signing up and I have entered the wrong credentials, an error message should pop up on the form.

#### Scenario B

When logging in/signing up and I have the correct credentials, I should be redirected back to the checkout panel immediately to continue filling out the checkout steps.

#### Scenario A

When logging in/signing up and I have entered the wrong credentials, the `Unauthorized` page pops up for a moment, then redirects back to the checkout page without displaying the error message on the form.

#### Scenario B

When logging in/signing up and I have the correct credentials, the `Unauthorized` page pops up for a moment, then the `Your cart is empty` page displays, then finally it redirects back to the checkout page.

### Steps to Reproduce the Behavior

* Add product to cart while not logged in
* Proceed to checkout while still not logged in
* Enter the wrong credentials on either signup/login to observe actual behavior in Scenario A
* Enter correct credentials on either signup/login to observe actual behavior in Scenario B

### Versions
Node: 6.2.2
NPM: 3.10.9
Meteor Node: 4.8.2
Meteor NPM: 4.5.0
Reaction CLI: 0.8.3
Reaction: 1.2.0
Reaction branch: development
Docker: 1.13.0

aaronjudd:
@mikemurray sounds like react router? however, we've had a similar behavior when the guest logins in the past as well...  where you see "oops your cart is empty" right after login, and then the cart loads..

mikemurray:
@aaronjudd @kieha The part about there being no error on wrong credentials was happening before React Router.

The second part is most likely React Router. When signing in the routes are re-rendered with updated permissions, and there's a moment where you have no permissions until the subscription is updated.

This update on login is how it used to work with Flow Router. We may have to think of a better approach for React Router.

aaronjudd:
@mikemurray @kieha is this still a valid issue in 1.3.0?

kieha:
@aaronjudd I've just tried reproducing it on 1.3.0 and it still happens, but the `unauthorized` page no longer shows up.

awadesmoline:
*Update*
[...]

PR #2752, opened 2017-08-28
Fix checkout login permissions

### LINKED ISSUE #2657: The arguments passed to userIsInRole() seems in the wrong order.
In master branch, release 1.4.0.
reaction/imports/plugins/core/checkout/client/templates/checkout/login/login.js
line 21:
const anonUser = Roles.userIsInRole(**"anonymous", Meteor.user()**, Reaction.getShopId());

According to the Role API document :
http://alanning.github.io/meteor-roles/classes/Roles.html#method_userIsInRole
_userIsInRole ( user  roles  [group] ) Boolean
Defined in roles/roles_common.js:254_

The order of 1st and 2nd argument should be reversed:
const anonUser = Roles.userIsInRole(**Meteor.user(), "anonymous"**, Reaction.getShopId());

The consequence of this issue is:  
   When I checkout without logged in and click "Continue as Guest" in first step, the  "Welcome, it's so good to see you." message will be displayed and go to 2nd step.(checkoutLoginCompleted returned true)

   If userIsInRole is invoked with correct argument order, the 1st step will keep its display and just go to 2nd step.(checkoutLoginCompleted returned false)

I am not sure which one is the expected behavior.

awadesmoline:
This resolves #2657 

### Steps to test
1. Checkout without logged in and click "Continue as Guest" in first step. The "Welcome, it's so good to see you." message will be displayed and go to 2nd step.(checkoutLoginCompleted returned true)
1. Checkout and Signin in the first step. The "Welcome, it's so good to see you." message will be displayed and go to 2nd step.(checkoutLoginCompleted returned true).


imports/plugins/core/checkout/client/templates/checkout/login/login.js
```diff
@@ -17,10 +17,11 @@ Template.checkoutLogin.helpers({
     const cart = Cart.findOne();
     if (cart && cart.workflow) {
       const currentStatus = cart.workflow.status;
-      const guestUser = Reaction.hasPermission("guest", Meteor.user());
-      const anonUser = Roles.userIsInRole("anonymous", Meteor.user(), Reaction.getShopId());
+      const guestUser = Reaction.hasPermission("guest", Meteor.userId(), Reaction.getShopId());
+      const anonUser = Reaction.hasPermission("anonymous", Meteor.userId(), Reaction.getShopId());
 
-      if (currentStatus !== self.template && guestUser === true && anonUser === false) {
+      // when user is guest, they are also anonymous.
```
impactmass: Can you describe this better? i.e to explain what the check below it does?

Seeing `when user is guest, they are also anonymous` doesn't particular help a reader to understand why we do this check.

impactmass:
Works as described in your steps. But when I register as a new user, the step 1 of checkout doesn't seem to recognize that someone is logged in. Can you confirm this?
screenshot 2017-08-28 14 19 29

imports/plugins/core/checkout/client/templates/checkout/login/login.js
awadesmoline: @impactmass I have fixed the changes you requested, and updated the comments. Please confirm it works well now

impactmass:
There's an error from the Circle CI tests. can you confirm why it failed?

awadesmoline:
@impactmass they're passing now.

imports/plugins/core/checkout/client/templates/checkout/login/login.js
```diff
@@ -17,10 +16,9 @@ Template.checkoutLogin.helpers({
     const cart = Cart.findOne();
     if (cart && cart.workflow) {
       const currentStatus = cart.workflow.status;
-      const guestUser = Reaction.hasPermission("guest", Meteor.user());
-      const anonUser = Roles.userIsInRole("anonymous", Meteor.user(), Reaction.getShopId());
+      const guestUser = Reaction.hasPermission("guest", Meteor.userId(), Reaction.getShopId());
```
aaronjudd: Don't need Meteor.userId() and getShopId(), those are supplied as default.

I worry about wether you should get rid of the permission check of "guest" and "anonymous".  I suspect this will have unexpected consequences in the `order/completed` page, and if you register a user in the Accounts UI-> not being in the checkout flow.

awadesmoline: I had to remove it because when clicking to continue as guest, the user still has the anonymous permission. So if we are checking to ensure that they are not anonymous, it fails. So that's why I just check if they are guest alone.

aaronjudd: When you click as "continue as guest" you ARE still `anonymous`.  you are only **NOT** anonymous if you have supplied an email, such as when you add an email at the order/completed stage.

awadesmoline: Yes that is how it is, but we were preventing `anonymous` from "continuing as guest". which is why I removed the permission for `anonymous`. I will retest and comment guest vs anonymous login view behavior as requested.

awadesmoline:
@aaronjudd I updated the comments. Please take a look.

aaronjudd:
See #2328 as well, may be related

zenweasel:
@aaronjudd Is there more to do on this? Should I have someone else pick it up or are we close to merging?

aaronjudd:
@rymorgan Note, while we closed this -> one side affect / repercussion -> the inline login in the checkout is supposed to show "REGISTER" as the default view, while the dropdown is supposed to show "SIGN IN" as the default view.

rymorgan:
@aaronjudd Got it. I've made an issue to correct it. https://github.com/reactioncommerce/reaction/issues/2944

PR #2756, opened 2017-08-29
Deprecate old cart transform methods

zenweasel:
Provides backwards compatibility and a deprecation message to anybody who might be using these cart methods


aaronjudd:
I'm feeling a little uncomfortable with this verbose of an approach.  Not that it's not right.. but it feels wrong.

Maybe we should at least have a plan / issue  to deprecate the deprecation notice..  say, this notice will be removed in the next release or something like that (ala React's various approaches to deprecating methods).  Still marinating on this.

zenweasel:
Yeah, I was thinking the same thing, I don't want this legacy stuff to hang around forever, but I don't know a good process for coming back to this. Would a `TODO` with some specific syntax maybe work?

lib/collections/transform/cartOrder.js
```diff
@@ -166,5 +175,30 @@ export const cartOrderTransform = {
       }
       return itemsByShop;
     }, {});
+  },
+  cartCount() {
+    Logger.warn("this method has been deprecated and is now replaced by the getCount method");
+    return this.getCount();
+  },
+  cartShipping() {
+    Logger.warn("this method has been deprecated and is now replaced by the getShippingTotal method");
```
aaronjudd: Why not say "cartShipping()" has been replaced instead of "this method" so we don't have to guess which method is now getShippingTotal.

Same comment applies to all.

aaronjudd:
@zenweasel sorry for the tardy response here. let's just clean up the conflicts and create an issue to deprecate in the next release.  we can merge after that.

spencern:
With the deprecation notice, we can also use the `@deprecated jsdoc tag instead of inline TODO comments. Not to add more work to this PR, but possible that it's worth it?
```
/**
 * @deprecated since version 1.5
 * /

machikoyasuda:
@spencern +1 on the `@deprecated` note

zenweasel:
@spencern @aaronjudd @machikoyasuda Added JSDoc deprecation notice, removed TODO and created issue #2882

Issue #2724, opened 2017-08-23
[Marketplace] Rethink implementation of order shipping fulfillment flow


kieha:
Referenced from #2613, there is need to rethink the implementation of the order shipping fulfillment flow to make it more flexible and extensible. 

We should also rethink the hard coded zeroth indexes of the `billing` and `shipping` arrays in the order object. This should be more flexible to allow for multiple billing and shipping objects that might be there in the marketplace implementation.

spencern:
Additionally we need to think about the flexibility of how we're progressing and regressing workflow steps.

From a comment I made on #2613 
>I'm always open to new ways to do things, but it seems like we should have "orders/progressStatus" and "orders/regressStatus" methods instead of one method per status.
>We could then have a set of workflows that identifies the next or previous status from the existing status and proceeds. There could be different states for items than for orders, and each should be overridable.
>This seems like it would make it easier to extend or replace functionality because if you replace the workflow set, then you replace the way that orders progress.

aaronjudd:
@kieha what's the estimated timeline on this? is it in progress? I consider this to be a (asap) critical issue.  Can you please schedule a time to review the approach with @spencern or myself on this before beginning.

kieha:
@aaronjudd this is not yet in progress. I'll reach out to you and/or @spencern ASAP before I start on it.

kieha:
A bit more discussion/comments on what should be done here to be found in this PR https://github.com/reactioncommerce/reaction/pull/2759.

In summary:
- In addition to rewriting the backend methods, a lot of rewrites need to happen in the front end code as well. I've added comments in the respective files, especially `imports/plugins/core/orders/client/containers/orderDashboardContainer.js` as to what would need to be done.

aldeed:
Done in 2.0.0

PR #2759, opened 2017-08-29
Fix broken marketplace orders dashboard

### LINKED ISSUE #2758: Marketplace orders dashboard broken
### Expected behavior

 - Can select multiple orders

### Actual Behavior

 - Cannot select multiple orders (Code missing due to merge issue)

- Make order
- Open orders dashboard and try select an order
- Observe that it doesn't happen

### Reaction Version
Node: 8.1.2
NPM: 5.0.3
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.10.0
Reaction: 1.5.0
Reaction branch: marketplace

kieha:
Resolves #2758 

Marketplace orders dashboard was broken, caused by the order filtering PR that was merged that had some code missing. This restores things back to order (pun intended 🤓).

### How to test

- Clone this branch
- Have at least 2 orders
- Visit orders dashboard
- Test it out: select orders, set shipping statuses, capture payments
- Observe that everything works as before

### DIFF
--- a/imports/plugins/core/orders/client/containers/orderDashboardContainer.js
+++ b/imports/plugins/core/orders/client/containers/orderDashboardContainer.js
@@ -436,8 +436,11 @@ class OrderDashboardContainer extends Component {
     let ordersToRegress = 0;
 
     selectedOrders.forEach((order) => {
+      // TODO: remove these hard-coded zero indexes to enable multiple shipments in marketplace
       const orderWorkflow = order.shipping[0].workflow;
       // check if the order(s) are in this state already or in the previous state
+
+      // TODO: model this with the assumption that there may be different workflows depending on the type of shop or product that a shop is selling.
       if (orderWorkflow.status === "new") {
         isNotPicked++;
       } else if (orderWorkflow.status === "coreOrderWorkflow/picked") {
@@ -481,8 +484,11 @@ class OrderDashboardContainer extends Component {
     let ordersToRegress = 0;
 
     selectedOrders.forEach((order) => {
+      // TODO: remove these hard-coded zero indexes to enable multiple shipments in marketplace
       const orderWorkflow = order.shipping[0].workflow;
       // check if the order(s) are in this state already or in one of the previous states
+
+      // TODO: model this with the assumption that there may be different workflows depending on the type of shop or product that a shop is selling.
       if (orderWorkflow.status === "new") {
         isNotPicked++;
       } else if (orderWorkflow.status === "coreOrderWorkflow/picked") {
@@ -534,8 +540,11 @@ class OrderDashboardContainer extends Component {
     let ordersToRegress = 0;
 
     selectedOrders.forEach((order) => {
+      // TODO: remove these hard-coded zero indexes to enable multiple shipments in marketplace
       const orderWorkflow = order.shipping[0].workflow;
       // check if the order(s) are in this state already or in one of the previous states
+
+      // TODO: model this with the assumption that there may be different workflows depending on the type of shop or product that a shop is selling.
       if (orderWorkflow.status === "new") {
         isNotPacked++;
         whichFalseState = shippingStates.picked;
@@ -588,8 +597,11 @@ class OrderDashboardContainer extends Component {
     let isShipped = 0;
 
     selectedOrders.forEach((order) => {
+      // TODO: remove these hard-coded zero indexes to enable multiple shipments in marketplace
       const orderWorkflow = order.shipping[0].workflow.status;
       // check if the order(s) are in this state already or in one of the previous states
+
+      // TODO: model this with the assumption that there may be different workflows depending on the type of shop or product that a shop is selling.
       if (orderWorkflow === "new") {
         isNotLabeled++;
         whichFalseState = shippingStates.picked;
--- a/imports/plugins/core/orders/client/containers/orderDashboardContainer.js
+++ b/imports/plugins/core/orders/client/containers/orderDashboardContainer.js
@@ -434,19 +434,20 @@ class OrderDashboardContainer extends Component {
     let ordersToRegress = 0;
 
     selectedOrders.forEach((order) => {
+      const orderWorkflow = order.shipping[0].workflow;
       // check if the order(s) are in this state already or in the previous state
-      if (order.shipping[0].workflow.status === "new") {
+      if (orderWorkflow.status === "new") {
         isNotPicked++;
-      } else if (order.shipping[0].workflow.status === "coreOrderWorkflow/picked") {
+      } else if (orderWorkflow.status === "coreOrderWorkflow/picked") {
         isPicked++;
       } else {
         // check if the selected order(s) are being regressed back to this state
-        if (order.shipping[0].workflow.workflow.includes("coreOrderWorkflow/picked")) {
+        if (orderWorkflow.workflow.includes("coreOrderWorkflow/picked")) {
           ordersToRegress++;
-        } else if (!order.shipping[0].workflow.workflow.includes("coreOrderWorkflow/picked") &&
-        (order.shipping[0].workflow.status === "coreOrderWorkflow/packed" ||
-        order.shipping[0].workflow.status === "coreOrderWorkflow/labeled" ||
-        order.shipping[0].workflow.status === "coreOrderWorkflow/shipped")) {
+        } else if (!orderWorkflow.workflow.includes("coreOrderWorkflow/picked") &&
+        (orderWorkflow.status === "coreOrderWorkflow/packed" ||
+        orderWorkflow.status === "coreOrderWorkflow/labeled" ||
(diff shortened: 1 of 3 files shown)


joykare:
Back to normal 💯  :1st_place_medal:

joykare:
@spencern **Urgent**  Fix to broken order dashboard due to earlier merge

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -194,6 +260,397 @@ class OrderDashboardContainer extends Component {
...
+  packedShippingStatus = (selectedOrders, status) => {
+    let isNotPicked = 0;
+    let isNotPacked = 0;
+    let isPacked = 0;
+    let ordersToRegress = 0;
+    const whichFalseState = shippingStates.picked;
+
+    selectedOrders.forEach((order) => {
+      // check if the order(s) are in this state already or in one of the previous states
+      if (order.shipping[0].workflow.status === "new") {
```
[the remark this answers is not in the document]
spencern: If we're going to leave this in for now, you could clean it up fairly easily with:
`const orderWorkflow = order.shipping[0].workflow.status;` and compare against that rather than using the longer version every time.

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -194,6 +260,397 @@ class OrderDashboardContainer extends Component {
...
+  }
+
+  labeledShippingStatus = (selectedOrders, status) => {
+    let isNotPacked = 0;
+    let isNotLabeled = 0;
+    let isLabeled = 0;
+    let ordersToRegress = 0;
+    let whichFalseState = "";
+
+    selectedOrders.forEach((order) => {
```
spencern: I've seen this bit of code 3 times now. Would like to have some comments and documentation explaining what the difference is between these very similar functions.

All the comments I've made on previous ones apply here too.

Additionally, worth adding a `TODO` task to write some helper functionality that can reduce the repetition and increase readability.

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -194,6 +260,397 @@ class OrderDashboardContainer extends Component {
...
+      // check if the order(s) are in this state already or in the previous state
+      if (order.shipping[0].workflow.status === "new") {
+        isNotPicked++;
+      } else if (order.shipping[0].workflow.status === "coreOrderWorkflow/picked") {
+        isPicked++;
+      } else {
+        // check if the selected order(s) are being regressed back to this state
+        if (order.shipping[0].workflow.workflow.includes("coreOrderWorkflow/picked")) {
```
spencern: Would like to have more comments explaining the different directions these `if..else if.. else` blocks can take.

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -194,6 +260,397 @@ class OrderDashboardContainer extends Component {
...
+      }
+    });
+    const selectedOrders = this.state.orders.filter((order) => {
+      return selectedOrdersIds.includes(order._id);
+    });
+
+    let orderCount = 0;
+
+    selectedOrders.forEach((order) => {
+      Meteor.call("orders/approvePayment", order, (err) => {
```
spencern: We should be batching these and performing a single method call rather than making a call for each order

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -194,6 +260,397 @@ class OrderDashboardContainer extends Component {
...
+        if (err) {
+          this.setState({
+            isLoading: {
+              capturePayment: false
+            }
+          });
+          Alerts.toast(`An error occured while approving the payment: ${err}`, "error");
+        } else {
+          Meteor.call("orders/capturePayments", order._id, (error) => {
```
spencern: Batch.

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -194,6 +260,397 @@ class OrderDashboardContainer extends Component {
...
+              this.setState({
+                isLoading: {
+                  capturePayment: false
+                }
+              });
+              Alerts.toast(`An error occured while capturing the payment: ${error}`, "error");
+            }
+
+            if (order.workflow.status === "new") {
+              Meteor.call("workflow/pushOrderWorkflow", "coreOrderWorkflow", "processing", order);
```
spencern: This should probably be performed on the server and not as an additional call from the client.

kieha:
@spencern all your comments on this are valid, i.e. this code is not marketplace ready. 

However, I'd like to point out that this is code that was already in marketplace, brought in from the [bulk actions PR](https://github.com/reactioncommerce/reaction/pull/2613), that got omitted when the [order filtering PR](https://github.com/reactioncommerce/reaction/pull/2688) was merged into marketplace, thus breaking stuff on the orders dashboard. I had already created an issue to make this code marketplace ready [here](https://github.com/reactioncommerce/reaction/issues/2724) that'll fix most of the issues with this.

That being said, I'll put comments and TODO blocks where they explain what the code at several places does and what needs to be done to make this marketplace ready.

spencern:
@kieha 
That was my understanding of this code as well, but I didn't remember seeing some of this to this point. We can merge this to fix the outstanding issues that were caused by the omission, but I needed to make note of the future incompatibility somewhere. 

I appreciate you taking the time to add TODOs and updating the issue #2724. It feels like updating the order code to be marketplace compatible is starting to become an epic in itself.

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -75,13 +83,52 @@ class OrderDashboardContainer extends Component {
       selectedItems: [],
       orders: props.orders,
       multipleSelect: false,
+      shipping: {
+        picked: false,
```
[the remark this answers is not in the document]
kieha: This part is not backend code to manage the shipping states. These are just state variables I'm using on the front end to change the UI (background color and icons on the lists) depending on what the user has clicked. 

E.g. before an order's status is set to `picked`, the state variable is as above, i.e. `this.state.shipping.picked = false`, which gives it this view (the order is not picked so the `picked` list item has a blue-ish background color and a circle as the icon):

screen shot 2017-08-30 at 2 59 09 am

After the order is set to `picked`, the state variable changes to `this.state.shipping.picked = true`, which triggers the UI change to this (the order has now been picked, this the list item background color and icon changes as below):

screen shot 2017-08-30 at 3 01 24 am

kieha:
@spencern I believe I've addressed most, if not all of the concerns raised around this PR by adding comments and TODO blocks around code that needs to be changed. I've also referenced this PR in this issue #2724.

You could take a second look at it.

spencern:
@kieha I'll take a look shortly.

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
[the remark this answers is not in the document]
spencern: I realize this, but doesn't this marry the front-end to our current backend workflow?

What happens if this frontend is used to manage a non-standard workflow?

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -194,6 +270,468 @@ class OrderDashboardContainer extends Component {
...
+        }
+      });
+    });
+  }
+
+  displayAlert = (selectedOrders, status, options) => {
+    // capitalize the first letter of the shipping status passed in
+    // e.g. 'shipped' becomes 'Shipped'.
+    // status[0].toUpperCase() capitalizes the first letter of the string
+    // status.substr(1).toLowerCase() converts every other letter to lower case
```
spencern: As an example for why this is important, but also why I'm an idiot, I was thinking that status[0] was the first item in an array during the last review not reading close enough to see that it's actually just pulling the first letter of the string.  Thanks for adding these comments.

kieha:
@spencern from your explanation,  I've added TODOs for the boolean shipping states as well. 
I think this is good to go now.

I'm starting on #2724 ASAP.

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -83,13 +83,15 @@ class OrderDashboardContainer extends Component {
...
       multipleSelect: false,
+      // TODO: model this with the assumption that there may be different stages to workflows
       shipping: {
         picked: false,
         packed: false,
         labeled: false,
         shipped: false
       },
       renderFlowList: false,
+      // TODO: model this with the assumption that there may be different stages to workflows
```
spencern: This is the right assumption 👍

PR #2772, opened 2017-08-30
Update group invitation check to always allow owner

impactmass:
Update after #2662 Scoped invitation.

**What Happened Earlier:**
* In #2189 we added check so that group invitation is scoped to the permissions that a user has. That means that when a user belongs to a group with role set `[a,b,c,d]`, she can only invite a user to groups with those permissions or a subset of them. We set up `Reaction.canInviteToGroup` method to do this check. 
* But, there is a case such as where an installed package adds a new role and the author didn't push the role to the owner group using `addRolesToGroups`. This will leave the owner inadvertently cut out by `canInviteToGroup`

**Changes in this PR:**
* To resolve the above, I added a check to grant access for a user with `owner` role
* I also added/copied `canInviteToGroup` to client Reaction, so it's possible to do Reaction.canInviteToGroup client side. This is useful when deciding dropdown options.
* I added more comments in the code around all the above

**How To test**
* Try multiple scenarios of trying to invite to group as an owner of a shop, either by:
** Adding a user to an existing group
** Adding a user to a new group created via the Accounts UI
** Adding a user to a new group created by Groups.insert(groupData) like in the case of a custom package
Confirm that an owner can invite, in all those cases.

**Confirm that previous functionality stays intact by:**
* Create a group with all Account roles toggled on. Add a user to that group. 
* Login as that user, go to the Accounts dashboard, confirm in the `Invite Form` that the user doesn't have dropdown for higher groups e.g "Shop Manager"

You can also test all these with a new shop other than the primary shop

### DIFF
--- a/server/methods/core/groups.js
+++ b/server/methods/core/groups.js
@@ -30,7 +30,7 @@ Meteor.methods({
     let _id;
 
     // we are limiting group method actions to only users with admin roles
-    // this also include shop owners, since they have the global roles
+    // this also include shop owners, since they have the `admin` role in their Roles.GLOBAL_GROUP
     if (!Reaction.hasPermission("admin", Meteor.userId(), shopId)) {
       throw new Meteor.Error(403, "Access Denied");
     }
@@ -80,7 +80,7 @@ Meteor.methods({
     check(shopId, String);
 
     // we are limiting group method actions to only users with admin roles
-    // this also include shop owners, since they have the global roles
+    // this also include shop owners, since they have the `admin` role in their Roles.GLOBAL_GROUP
     if (!Reaction.hasPermission("admin", Meteor.userId(), shopId)) {
       throw new Meteor.Error(403, "Access Denied");
     }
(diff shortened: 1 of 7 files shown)


server/methods/core/groups.js
```diff
@@ -124,14 +128,19 @@ Meteor.methods({
...
-        throw new Meteor.Error(403, "Access Denied");
-      }
+    // we are limiting group method actions to only users with admin roles
+    // this also include shop owners, since they have the `admin` role in their Roles.GLOBAL_GROUP
+    if (!Reaction.hasPermission("admin", loggedInUserId, shopId)) {
+      throw new Meteor.Error(403, "Access Denied");
+    }
+
+    // Users with `owner` and/or `admin` roles can invite to any group
+    // Also a user with `admin` can invite to only groups they have permissions that are a superset of
```
spencern: These two comment lines seem to contradict each other.
Can a user with `admin` permissions invite to any group, or only groups they have superset of permissions for?

spencern: @impactmass want to make sure you see this comment as I think these comments should be clarified, but going to merge anyway.

impactmass: I'll note to update the comment in next PR I open, but I meant the latter: a user with admin can invite to only groups they have superset of permissions for.

spencern: That's what I figured. Next PR is good

Issue #2760, opened 2017-08-29
[Marketplace] Routing issues with multiple shops


mikemurray:
### Expected behavior
- Router should be able to resolve routes for the primary shops as well as all marketplace shops using `Router.go` and `Router.pathFor`.
- Should be able to navigate to a marketplace shop's homepage. For example `/reaction1` and view its products.

### Actual Behavior
- The router can only resolve the current stop and the primary shop routes.
- There is no index route for shops. for example, if your shop URL was `/reacton1`, you would get a 404 page because that route doesn't exist

### Steps to Reproduce the Behavior
1) Add a new marketplace shop in settings
2) Go to slug for shop in address bar, see the 404

### Versions
Node: 7.2.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.1-ce

spencern:
It might be worth considering passing in either a shopId or shop slug into `Router.go` and `Router.pathFor`

spencern:
@mikemurray are you working to resolve this? Do you think it's a bug or an extension of existing features?

mikemurray:
@spencern Well, currently you get a 404 if you go to any presumed shop route, such as `/reaction1`, when you should get the index page for that shop. So Bug?

We need to add routes like the following options.

- `/:shopHandle` or `/shops/:shopHandle` & `/shops` for all shops?
- `/:shopHandle/tag/:id?` or `/shops/:shopHandle/tag/:id?`
- `/:shopHandle/product/:handle` or `/shops/product/shopHandle/:handle`

I am not currently working to resolve this issue.

spencern:
@mikemurray It does sound like a bug then.
Is this in your queue or is anyone else competent with your router to add these options?

Are you blocked by this right now?

mikemurray:
It is not in my queue and was punted to @jshimko to fix the router issue as he may have some ideas. As far as I know only you, @zenweasel and myself have worked on the router.

This issue is blocking progress in terms of properly testing filtering products by the shop on the grid in both marketplace and elsewhere.

aaronjudd:
@jshimko this is something that is a blocker, and we need to know if you're going to be able to work on this, this week?

spencern:
[...]

PR #2774, opened 2017-08-31
Fix product publications

spencern:
Fixes hacky product publications that got added to marketplace in haste a couple days ago.

Specifically:
**Security Fix**
* Fixes bug where users with permissions for any shop, were considered to have that permission for all shops by `Reaction.hasPermission`. This was code that was leftover from the original sellerShops implementation of marketplace.

**New Methods**
* Adds `Reaction.getShopsWithRoles` method which provides a list of shopIds that the user has at least one of the provided roles for.

**Updates**
* the `Products` publication now publishes admin products for all shops that a user has access for.
* The productMedia publication helper now publishes media for all published products, regardless of the shopId
* The Product publication was updated to check if a user has roles for the specific product that is being published, not the active shop. This fixes an issue where if a user had their shop active and browsed another's products, they could see them as if they were an admin.


spencern:
@mikemurray can you give this a review sometime soon?

mikemurray:
Yeah, reviewing now.

Question. I have 2 brand shops and the primary shop signed into three browsers. I on the homepage, while signed in as brand 2, I cannot see the products from brand 1. 

Should I only be able to see my products? Or, is this related to https://github.com/reactioncommerce/reaction/issues/2760.


reaction_and_reaction_and_reaction_and_fix_product_publications_by_spencern_ _pull_request__2774_ _reactioncommerce_reaction

spencern:
@mikemurray 
Are those products published? You should not be able to see hidden products from other brands. I'm also thinking we should integrate with the "edit mode" switch and only show products you have admin access for when you're in edit mode.

I thought this was publishing all public products as well as products you have admin access for, but it may be that the admin publication supersedes the other publication and results in the other products never getting sent.

mikemurray:
@spencern It seems that the admin publication is superseding the "normal user" part of that publication. As an anonymous user, I can see the published products from all shops. But as a brand, I can only see my own shops products on the main grid.

spencern:
What do you think about having the publication switch with the "edit mode" switch? Is that too confusing? Any thoughts for how we should we publish products that a user doesn't have admin access for when a user is an admin for at least one shop? (does that make sense?)

mikemurray:
The edit mode might be helpful, at least from the standpoint of being clear that these are my products to edit; you can get that from `Meteor.user().profile` on the server, or pass it in. Probably don't need to edit mode boolean, since the publication should know for which shop its publishing products for, and if I'm an admin of that shop. And edit mode on client mostly just hides the edit buttons, badges, etc.

---

On the main product grid, we could just adjust the publication to only show products that a normal user would see. Then on the shop's own product grid homepage, we can, of course, show all the editable products for that shop.

So basically...

**As shop owner, In edit mode**
- Homepage - sees grid like normal user
- My Shops page - **sees grid as admin**
- Other Shops page - sees grid like normal user

**As shop owner in, NOT in edit mode ( viewing as a customer )**
- Homepage - sees grid like normal user
- My Shops page - sees grid like normal user
- Other Shops page - sees grid like normal user

spencern:
I like this idea. Do we have a navigation link to "my shop" currently?

mikemurray:
We do not.

spencern:
Does the `my-shops` route exist? Are you referring to `/:shop-slug/products`

mikemurray:
They may exist if you physically type, or do a `location.href` to "/:shopPrefix/thing" but I don't think it'll give you a URL param to use to filter by the shop. That's what may be solved with https://github.com/reactioncommerce/reaction/issues/2760

spencern:
@mikemurray 
I've updated this PR to show all products when an administrator has edit mode turned off. I think updating it to be route aware for a route that doesn't exist yet is beyond the scope of this PR and should perhaps be reconsidered after #2760 is finished.


As shop owner, In edit mode
Any Product Grid page: See only products they have edit permissions for (owner/admin/createProduct)

As shop owner in, NOT in edit mode ( viewing as a customer )
See all products as a customer would. They will _not_ see their own products that are hidden.
I think this is probably preferable anyway as we've needed a way for admins to see what a given product grid looks like from a customer perspective.

Might be worth considering changing the language from "edit mode" to `customer --* admin` or something similar, but I think that would require additional Publication updates.

From my perspective this PR is ready to be fully reviewed and pulled in.

mikemurray:
@spencern checking now

PR #2777, opened 2017-08-31
[WIP] Allow users to udpate their email address through Profile UI

kieckhafer:
This update will allow users to update their email address through our Profile UI.

It will not allow a user to update their email to an address that is already in use by another user.


kieckhafer:
Already have a PR in the correct branch... #2745

PR #2793, opened 2017-09-05
Updated category tag find

Oleh-Pashchenko:
Added 'toLowerCase' to tag:
`const slug = Reaction.Router.getParam("slug").toLowerCase();`

Updated find Products by hashtags:
`const productCursor = Products.find({
    hashtags: { $in: tags.tags },
    ancestors: [],
    type: { $in: ["simple"] }
  });`

### DIFF
--- a/imports/plugins/included/product-variant/containers/productsContainer.js
+++ b/imports/plugins/included/product-variant/containers/productsContainer.js
@@ -101,7 +101,7 @@ function composer(props, onData) {
 
   let canLoadMoreProducts = false;
 
-  const slug = Reaction.Router.getParam("slug");
+  const slug = Reaction.Router.getParam("slug").toLowerCase();
   const tag = Tags.findOne({ slug: slug }) || Tags.findOne(slug);
   const scrollLimit = Session.get("productScrollLimit");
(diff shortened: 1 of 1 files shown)


spencern:
Hi @Oleh-Pashchenko 
Thank you for your PR! 

We don't accept pull requests directly to our master branch. We've also recently deleted our `development` branch in favor of feature release branches. Our current feature branch is `marketplace` however it's quite a bit ahead of `master` at this point. I think your code may still be relevant, but this PR will need to be made to the `marketplace` branch instead of `master`. You should be able to cherry pick this commit into a patch branch based on the current reaction marketplace branch and resubmit the PR.

In order to better understand what this update accomplishes, can you describe in a little more detail what issue you were seeing that prompted this code change and how it resolves that?

Thanks again for contributing! I hope this is the first of many contributions

aaronjudd:
This PR needs to be updated, and further explanation of the goal.  We can cherry pick this into marketplace (which will be in master soon). @Oleh-Pashchenko closing for now, if you want to re-open with more comments. (btw _ do think this is solving a real issue, just want to be sure what /how)

Issue #2690, opened 2017-08-16
Publish all without label doesn't let know which product is missing label


janat08:
### Expected behavior
If there are problems with a product during publish all, an informative error message would be displayed that indicates what needs to be done

### Actual Behavior
error is produced, but there's no immediate indication of where user attention is required

### Steps to Reproduce the Behavior
1. Login as Admin
1. Create a product but leave "Label" blank
1. Navigate to the product grid
1. Hit "Publish All"
1. Observe that the products don't publish, but there is no indicator what the problem is or which product is problematic.

### Versions
Node: 6.11.2
NPM: 3.10.10
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.4.1
Reaction branch: master

janat08:
Also spamming add products may block out publish all.

spencern:
Thanks for reporting this,
We've got an open issue to redesign the way we validate errors when editing or creating a product.
#2274 

Additionally #2404 gets into this issue a bit as well, though not specifically in regards to the label.

zenweasel:
@rymorgan It would be great to get some design input on how this error message should show up. My first idea was just a red X on the products that are problematic but maybe there is something better/more informative.

janat08:
Filtering out products that are okay, or not part of publication draft is good.

rymorgan:
@zenweasel What would be ideal is you show a red dot instead of the sea green dot on the products. And open the sidebar and, highlight the field and that needs is required with an inline alert that explains what the user needs to do to finish the publishing process. In the case of multiple products missing the label field, we would have to show that inline alert above the grid and say where they need to edit the fields.

aaronjudd:
@jm-wanja what's the status of this?

jm-wanja:
@aaronjudd I haven't worked on this yet, I've been dealing with #2533 in the past couple of days.

jm-wanja:
@rymorgan should the red button completely replace the sea green button(which shows changes made to the product that have not been published) or the red button should show that the product is missing the variant Label? @zenweasel

rymorgan:
@jm-wanja Yes, it should look like this, where teh sea green changes to a danger red (@rui-danger). 
![reaction-lable-indicator](https://user-images.githubusercontent.com/20409254/30079798-47b04782-9236-11e7-9cfa-6e0a718e815b.png)

zenweasel:
Reopened by #3021

[...]

PR #2795, opened 2017-09-06
Show products that fail validation when publishing.

jm-wanja:
Resolves issue #2690 

Previously we could identify which products are missing the label when publishing, the red button shows which products are missing the label.

**How to Test**

1. Login as Admin
2. Create a product but leave "Label" blank
3. Navigate to the product grid
4. Hit "Publish All"
5. Observe that the products don't publish, and there is a red button as an indicator of which product it is that needs changes. 
6. When the problem product is opened, observe that the `edit variant` panel is also opened and the label field flashes red to notify that the label is missing.


zenweasel:
@jm-wanja This PR could use a better name?

zenweasel:
So this works fine for just checking that the label exists but we want to flag any product that fails Simple Schema validation as I mentioned. So you will want to loop around where you are just checking the label length and call `isValid()` on the product and flag any product for which that is false. You can see the documentation for that here: https://github.com/aldeed/meteor-simple-schema#validating-data

jm-wanja:
@zenweasel I have made the requested changes. This now flags the products missing the variant label by calling `isvalid()`.

zenweasel:
So testing this it works fine. I wish it would take you to exactly what was broken (e.g. open the variant details panel when that's where the problem is) but that's probably beyond the scope of this ticket.

If we can just rename the variables/functions so they properly reflect what they are doing now then we should be able to merge.

rymorgan:
@jm-wanja @zenweasel I absolutely agree/think is critical that it should open panel/card and highlight the field specific label that is missing. This has become a pretty major UX issue.

jm-wanja:
@rymorgan @zenweasel in the case of multiple products missing the variant label field how can we handle opening the variant form on each product?

zenweasel:
@jm-wanja I think he means when you click on the product, that the PDP show what fields are invalid

rymorgan:
@zenweasel Yes! I do mean that. It can be a different PR if need be, but I think that would greatly help with something that's confusing to a lot of our users.

jm-wanja:
@rymorgan Is an inline alert needed or is opening the panel with the fields that is invalid and highlighting it enough? @zenweasel

zenweasel:
@jm-wanja I think opening the correct panel and showing the validation would be enough for now.

zenweasel:
Just remember that it's not just missing label that could cause the validation error. We need to handle any sort of validation error.

jm-wanja:
Alright, so currently this PR opens the edit variant panel when there is a validation error from a product variant. Also it flashes red on the field that is causing the error. This only works for fields on the product variant that are causing the validation error. @zenweasel @rymorgan

rymorgan:
@jm-wanja The validation on PDP isn't working for me. I made a different issue for the PDP label part. 

The missing label indicator on the grid seems to be working. So it might make sense to address the other issue in a different PR. @zenweasel that make sense?

rymorgan:
Ok, I take the previous comment back. I see that it does work. But it only works in the specific instance that: 
You create a product
Navigate to the grid
Publish product from the grid
Click back through product and you get the error 

So it's solving only one small fraction of the underlying issue. But again, I think we can address that in another issue.

zenweasel:
So I think if this PR:

1. Shows the red dot on products that fail schema validation in the product grid

I feel like it meets the criteria for the issue it's trying to solve. Handing how validation should be displayed in the PDP is super important but there are multiple open issues for that already.

In retrospect I should have removed the word "label" from the issue description to make it clearer that this should be all schema validation and not just missing labels.

zenweasel:
The Sold Out indicator overlaps with the status but @rymorgan has already noted that here: https://github.com/reactioncommerce/reaction/issues/2801
reaction

zenweasel:
I would like to do more with improving this UI and experience but it's beyond the scope here. After discussion with @rymorgan we decided to open a separate issue as this requires planning and resources. https://github.com/reactioncommerce/reaction/issues/2865

This works as expected for the original intent.

PR #2808, opened 2017-09-08
Framework for Shopify synchronization + Basic Inventory sync

spencern:
**This PR aims to do the following**  
* Provide an example for how to use webhooks in a plugin
* Extend the Shopify connector to be capable of syncing data from a Shopify store to a Reaction store.
* Provide a framework that sets up webhooks when a sync is started and removes the webhooks when the sync is turned off
* Provides a basic example sync which syncs inventory from Shopify to Reaction when an order is placed on Shopify

**This PR contains a new idea that should be reviewed extra carefully**
* Event logging on products and variants. Used to note that an external service reduced the inventory of an item, but is an informal proposal for how to log more data across Reaction.
  * New Schema file [schema/logs.js](https://github.com/reactioncommerce/reaction/blob/spencer-sync-shopify-orders/lib/collections/schemas/logs.js)
  * Used to note that inventory was updated for a particular variant in [sync/orders.js](https://github.com/reactioncommerce/reaction/blob/spencer-sync-shopify-orders/imports/plugins/included/connectors-shopify/server/methods/sync/orders.js#L62-L68)

**To Test:**  
Unfortunately, there's not an easy way to test this PR without deploying it to a real domain, or modifying some code and using a tunnel such as [ngrok](https://ngrok.com/).

1. Pull this branch & reset
2. Add Shopify Credentials.
3. Select Products and click Import - wait until the import is finished (You shouldn't need to wait, but it makes testing consistent)
4. Fire Up [ngrok](https://ngrok.com/) and determine what domain is tunneling to your localhost
you should have a terminal that looks something like this:
image
you'll see two urls `http://_______.ngrok.io` and `https://_______.ngrok.io` you want to use the `http` version as Shopify checks certificates on https webhooks.

5. **EDIT SOME CODE** - on line [48/49 of sync.js](https://github.com/reactioncommerce/reaction/blob/spencer-sync-shopify-orders/imports/plugins/included/connectors-shopify/server/methods/sync/sync.js#L48) - after integrations, update the Meteor call to look like this
*Note the slash `/` at the end of the ngrok url, it's necessary.*
    ```
      Meteor.call("connectors/shopify/createWebhook", {
        topic: topic,
        integrations: integrationsByTopic[topic],
        absoluteUrl: "http://yourdomain.ngrok.io/"
      });
    ```
    *Without adding this `absoluteUrl`, Shopify will attempt to add a webhook for your actual absoluteUrl which is `localhost` which will cause a failure as Shopify's webhook API will reject any attempt to create a webhook for a `localhost` url*
6. Select `Update Inventory when new Shopify order is placed` and click Sync to create webhooks webhooks.
7. Select a product that you want to test that's on both the Shopify store and the Reaction store. Note the inventory quantity for a particular variant or option.
8. Checkout with that particular variant or option.
9. The inventory available on the Reaction site should be reduced by the number of inventory that you checked out with on Shopify.


imports/plugins/included/connectors-shopify/client/settings/shopify.js
```diff
@@ -39,6 +39,97 @@ Template.shopifyProductImport.events({
...
+    });
+
+    if (pkg && pkg.settings && Array.isArray(pkg.settings.webhooks) && pkg.settings.webhooks.length > 0) {
+      return true;
+    }
+    return false;
+  },
+
+  integrationIsActive(integration) {
+    const pkg = Packages.findOne({
```
aaronjudd: Any reason this isn't updated to the pattern:

```
    const { settings } = Reaction.getPackageSettings( reaction-connectors-shopify");
```

imports/plugins/included/connectors-shopify/lib/collections/schemas/shopifyConnect.js
```diff
@@ -1,6 +1,37 @@
 import { SimpleSchema } from "meteor/aldeed:simple-schema";
 import { PackageConfig } from "/lib/collections/schemas/registry";
 
+const Webhook = new SimpleSchema({
+  shopifyId: {
+    type: Number,
+    label: "Shopify Webhook ID",
```
aaronjudd: Let's get our casing and language consistent here.

imports/plugins/included/connectors-shopify/server/endpoints/webhooks.js
```diff
@@ -0,0 +1,55 @@
+import crypto from "crypto";
+
```
aaronjudd: Is there a reason for this spacing, shouldn't lint have complained here?

imports/plugins/included/connectors-shopify/server/methods/api/products.js
```diff
@@ -0,0 +1,46 @@
...
+   * @async
+   * @method connectors/shopify/api/products/count
+   * @param {object} options An object of options for the shopify API call. Available options here: https://help.shopify.com/api/reference/product#count
+   * @returns {number} Number of products
+   */
+  async "connectors/shopify/api/products/count"(options) {
+    check(options, Match.Maybe(Object));
+
+    if (!Reaction.hasPermission(connectorsRoles)) {
+      throw new Meteor.Error(403, "Access Denied");
```
aaronjudd: Is this the correct param format for `Meteor.Error`?

imports/plugins/included/connectors-shopify/server/methods/api/products.js
```diff
@@ -0,0 +1,46 @@
+import Shopify from "shopify-api-node";
+
```
aaronjudd: Are we spacing here because we're grouping? That's hard to be consistent on.

lib/collections/schemas/products.js
```diff
@@ -301,6 +302,14 @@ export const ProductVariant = new SimpleSchema({
     type: Date,
     optional: true
   },
+  // TODO: REVIEW - Does this need to exist? Should we use workflow instead?
+  // Should it be called 'history' or something else instead?
+  // Should this go into the Logger instead? Is the logger robust enough for this?
```
aaronjudd: we use a `transactions` schema elsewhere as a general unstructured logging array.

aaronjudd:
@jshimko we should discuss https://github.com/reactioncommerce/reaction/blob/spencer-sync-shopify-orders/lib/collections/schemas/logs.js

imports/plugins/included/connectors-shopify/client/settings/shopify.html
```diff
@@ -10,13 +10,51 @@
...
+      {{> shopifySync}}
     </div>
   {{/if}}
 </template>
 
-<template name="shopifyProductImport">
-  <button class="btn btn-default" data-event-action="importProductsFromShopify">
-    <i class="fa fa-cloud-download"></i><span data-i18n="admin.shopifyConnectSettings.importProducts">Import Products</span>
-  </button>
+<template name="shopifyImport">
```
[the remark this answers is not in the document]
spencern: I strongly considered refactoring this to React before creating the PR, but wanted to get back to getting marketplace ready. It's high on my list too.

imports/plugins/included/connectors-shopify/server/endpoints/webhooks.js
spencern: No good reasons. Will fix.

imports/plugins/included/connectors-shopify/server/methods/api/products.js
spencern: yeah, I was grouping into Node, then Meteor, then Reaction, then local imports, but agreed and not sure it's worth the effort. I'll fix.

lib/collections/schemas/products.js
spencern: I've seen the `transactions` schema on billing, shipping, and discounts - is that the preferred method of logging these types of things. I think it could be useful to have a structured array of events that have occurred on products, inventory, etc - but I acknowledge this is a pretty big feature to sneak into this PR

imports/plugins/included/connectors-shopify/client/settings/shopify.html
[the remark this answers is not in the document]
spencern: Agreed, I strongly considered refactoring this to React, but felt it was better for me to push this out and get back to working on getting marketplace launched.

imports/plugins/included/connectors-shopify/server/methods/api/products.js
spencern: Good catch.
I think this is wrong by our style guide now as we don't throw codes anymore and it should be
```
throw new Meteor.Error("access-denied", "Access denied");
```

imports/plugins/included/connectors-shopify/server/methods/api/products.js
spencern: yes, I was grouping by Node > Meteor > Reaction > local, but agree that it's not necessarily consistent with the rest of the code base, I can remove the spacing.

imports/plugins/included/connectors-shopify/client/settings/shopify.js
spencern: Nope, will update.

imports/plugins/included/connectors-shopify/lib/collections/schemas/shopifyConnect.js
spencern: Sentence case for labels, correct?

lib/collections/schemas/products.js
spencern: Would love a discussion on how best to store logs for things like this. I think theres some value in having a structured set of data for events that have happened on products, orders, inventory, users, etc and was experimenting with that. Completely open to removing it entirely in favor of something else

spencern: Regarding this discussion, I also considered using a `history` object similar to what is found on orders here https://github.com/reactioncommerce/reaction/blob/marketplace/lib/collections/schemas/orders.js#L22-L35

PR #2813, opened 2017-09-08
Remove hardcoded billing and shipping objects

### LINKED ISSUE #2770: Remove hardcoded zeroth indexes of the billing and shipping arrays in the order object
Stemming from a conversation with @spencern, we've decided to break #2724 into two: Removing the hard coded zeroth indexes and making order shipping fulfillment flow more flexible and extensible.

This is part 1 of that: removing the hard coded zeroth indexes from the billing/shipping arrays in the order object.

In order for the current code to be marketplace ready, we need to remove hard coded indexes to allow for multiple shipment/billing support.

kieha:
Resolves #2770 

This PR aims to remove hard coded first billing and shipping objects from most code around orders.

### How to test

- Open browser console.
- Go through order creating/completing/processing-as-admin process.
- Make sure everything works as normal, and there are no errors in the browser console and/or server.


kieha:
@spencern this is ready to be reviewed.

imports/plugins/core/orders/client/components/orderSummary.js
```diff
@@ -1,8 +1,35 @@
...
+import { Reaction } from "/client/api";
 import { Badge, ClickToCopy } from "@reactioncommerce/reaction-ui";
 
+/**
+ * getBillingInfo
+ *
+ * @summary get proper billing object as per current active shop
+ * @param {Object} order - order object to check against
+ * @return {Object} proper billing object to use
+ */
```
spencern: a few notes on this function and use of `find`
- While it's perfectly valid javascript, as an organization, Reaction is moving towards enforcing the use of parens around your function args.  e.g. `billing.find(billing => x === y)` is not as clear as `billing.find((billing) => x === y)`
-  If you have to break your `find` into multiple lines, probably better for readability to use curly brackets and a `return`
- If you're falling back to an empty object, it's more readable if you put the result of the `find` into a variable. I didn't immediately realize what the ` || {}` was part of the return initially.

Example with all suggestions.
```
function getBillingInfo(order) {
  const billingInfo = order.billing.find((billing) => {
    return billing.shopId === Reaction.getShopId();
  }
  return billingInfo || {};
}
```

imports/plugins/core/orders/client/components/orderSummary.js
```diff
@@ -1,8 +1,35 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import moment from "moment";
+import { Reaction } from "/client/api";
 import { Badge, ClickToCopy } from "@reactioncommerce/reaction-ui";
 
+/**
+ * getBillingInfo
+ *
```
spencern: It's not quite canonized yet, but private methods/functions should use the `@private` tag in the jsdoc so they don't get rendered out to our docs.

imports/plugins/core/orders/client/components/orderSummary.js
```diff
@@ -1,8 +1,35 @@
...
+}
+
+/**
+ * getShippingInfo
+ *
+ * @summary get proper shipping object as per current active shop
+ * @param {Object} order - order object to check against
+ * @return {Object} proper shipping object to use
+ */
+function getShippingInfo(order) {
```
spencern: same comments on getBillingInfo apply here.

imports/plugins/core/orders/client/components/orderSummary.js
```diff
@@ -86,28 +113,28 @@ class OrderSummary extends Component {
             <div className="order-summary-form-group">
               <strong data-i18n="order.processor">Processor</strong>
               <div className="invoice-details">
-                {order.billing[0].paymentMethod.processor}
+                {getBillingInfo(order).paymentMethod.processor}
```
spencern: If we fall back to returning `{}` in getBillingInfo / getShippingInfo, this will throw an error here in that case. Might be worth building a few variables that won't throw errors before getting into the render function.

imports/plugins/core/orders/client/components/orderSummary.js
```diff
@@ -86,28 +113,28 @@ class OrderSummary extends Component {
...
               </div>
             </div>
 
             <div className="order-summary-form-group">
               <strong data-i18n="order.payment">Payment</strong>
               <div className="invoice-details">
-                {order.billing[0].paymentMethod.storedCard} ({order.billing[0].invoice.total})
+                {getBillingInfo(order).paymentMethod.storedCard} ({getBillingInfo(order).invoice.total})
```
spencern: will throw error if `getBillingInfo(order)` returns `{}`

imports/plugins/core/orders/client/components/orderSummary.js
```diff
@@ -86,28 +113,28 @@ class OrderSummary extends Component {
...
               </div>
             </div>
 
             <div className="order-summary-form-group">
               <strong data-i18n="order.transaction">Transaction</strong>
               <div className="invoice-details">
-                {order.billing[0].paymentMethod.transactionId}
+                {getBillingInfo(order).paymentMethod.transactionId}
```
spencern: will throw error if `getBillingInfo(order)` returns `{}`

imports/plugins/core/orders/client/components/orderSummary.js
```diff
@@ -86,28 +113,28 @@ class OrderSummary extends Component {
...
               </div>
             </div>
 
             <div className="order-summary-form-group">
               <strong data-i18n="orderShipping.carrier">Carrier</strong>
               <div className="invoice-details">
-                {order.shipping[0].shipmentMethod.carrier} - {order.shipping[0].shipmentMethod.label}
+                {getShippingInfo(order).shipmentMethod.carrier} - {getShippingInfo(order).shipmentMethod.label}
```
spencern: will throw error if `getShippingInfo(order)` returns `{}`

imports/plugins/core/orders/client/components/orderTable.js
```diff
@@ -34,6 +34,32 @@ const classNames = {
...
 };
 
+/**
+ * getBillingInfo
+ *
+ * @summary get proper billing object as per current active shop
+ * @param {Object} order - order object to check against
+ * @return {Object} proper billing object to use
+ */
+function getBillingInfo(order) {
```
spencern: Not sure what the best practices are for React around condensing helper functions into a lib file or how that should be done, but since this is exactly the same as the other getBillingInfo (and getShippingInfo below is also identical to the other file), perhaps it could exist in one place and get imported? @jshimko any best practice thoughts on this?

imports/plugins/core/orders/client/components/orderTable.js
```diff
@@ -118,7 +137,7 @@ class OrderTable extends Component {
           </span>
 
           <span className="order-data order-data-total">
-            <strong>Total: {formatPriceString(this.getBillingInfo(order).invoice.total)}</strong>
+            <strong>Total: {formatPriceString(getBillingInfo(order).invoice.total)}</strong>
```
spencern: will throw error if `getBillingInfo(order)` returns `{}`

imports/plugins/core/orders/client/components/orderSummary.js
spencern: After chatting with @jshimko another pattern that we're using in React if you don't want the curly brackets or return statement
```
order.billing.find((billing) => (
  billing.shopId === Reaction.getShopId()
)
```
I'm not 100% I love this outside of the React context and still prefer to either keep a single line or use curlies and a `return`, but it would be better .

imports/plugins/core/accounts/client/components/updateEmail.js
```diff
@@ -50,7 +50,7 @@ class UpdateEmail extends Component {
         />
         <Components.Button
           bezelStyle={"solid"}
-          icon={showSpinner && "fa fa-spin fa-circle-o-notch"}
+          icon={showSpinner ? "fa fa-spin fa-circle-o-notch" : ""}
```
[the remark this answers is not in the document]
aaronjudd: Style preference is the original `&&` versus ternary operators... (which aren't so cool, and generally dislike ternary operators because it results in hard to read code...). Let's revert this (and it has nothing to do with this PR).

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
```diff
@@ -500,11 +503,11 @@ class OrderDashboardContainer extends Component {
     // status of each order in regard to the other statuses
     // TODO: optimise this process to avoid having this similar repetitive block of code across 4 methods
     selectedOrders.forEach((order) => {
-      // TODO: remove these hard-coded zero indexes to enable multiple shipments in marketplace
-      const orderWorkflow = order.shipping[0].workflow;
+      const orderWorkflow = getShippingInfo(order).workflow;
```
aaronjudd: What if this `getShippingInfo(order)` returns null?

imports/plugins/core/orders/client/components/orderTable.js
```diff
@@ -143,24 +136,26 @@ class OrderTable extends Component {
   }
 
   renderShipmentInfo(order) {
-    const emailAddress = order.email || <Translation defaultValue={"Email not availabe"} i18nKey={"admin.orderWorkflow.ordersList.emailNotFound"} />;
+    const emailAddress = order.email ||
+    <Translation defaultValue={"Email not availabe"} i18nKey={"admin.orderWorkflow.ordersList.emailNotFound"} />;
```
zenweasel: Typo: available

imports/plugins/core/orders/client/templates/workflow/shippingTracking.js
```diff
@@ -44,7 +45,8 @@ Template.coreOrderShippingTracking.events({
   },
   "click [data-event-action=shipmentShipped]": function () {
     const template = Template.instance();
-    Meteor.call("orders/shipmentShipped", template.order, template.order.shipping[0], (err) => {
+    const shipment = getShippingInfo(template.order);
+    Meteor.call("orders/shipmentShipped", template.order, shipment, (err) => {
```
aaronjudd: not `(e)`, not `(err)`, but `(error)` is preferred style.

imports/plugins/core/orders/client/components/orderTable.js
```diff
@@ -205,14 +200,38 @@ class OrderTable extends Component {
...
+        },
+        "Shipping": {
+          accessor: row => getShippingInfo(row).workflow.status,
+          id: "shippingStatus"
+        },
+        "Status": {
+          accessor: "workflow.status",
+          id: "workflow.status"
+        },
+        "": {
```
jshimko: Can this empty string object key actually work?  I know it was already there, but I'm unsure how that didn't cause a problem before.

kieha: @jshimko yeah, it works. It creates a column with no header/label. That is the last column on the order table. But I can remove the empty string key/replace it with an actual string and find a workaround for creating the headerless column.

imports/plugins/core/accounts/client/components/updateEmail.js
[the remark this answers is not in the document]
kieha: @spencern @aaronjudd  I'd made this fix here because this throws this console error

`Failed prop type: Invalid prop 'icon' of type 'boolean' supplied to 'Reaction(Button)', expected 'string'.`

 if `showSpinner` is false. So it was passing an empty string to the `icon` prop instead.

I'll, however, revert this and probably open a separate issue for it as it's not related to this PR.

imports/plugins/core/orders/client/templates/workflow/shippingTracking.js
kieckhafer: Do we have a cascading preference or which should be used?

If we have a callback, and another method inside that callback, we can't use `error` again on the second callback since it's already defined (or at least the linter yells at you if you do). Do we do `error`, then `err`, then `e`?

kieha: Or scope out the errors depending on what the call is doing if you have nested calls, e.g. `productError` then `orderError`?

kieckhafer: ^^ Yes, that's a great idea.

spencern:
@kieha fyi, some conflicts we need to resolve here.

imports/plugins/core/orders/client/containers/orderDashboardContainer.js
kieha: The way `getShippingInfo(order)` is written, it'll either return the correct shipping object or an empty object.

kieha:
@spencern @aaronjudd this is ready for another look through.

aaronjudd:
@spencern this is ready for you to review again

imports/plugins/core/orders/client/components/orderSummary.js
```diff
@@ -149,15 +156,24 @@ class OrderSummary extends Component {
...
-          <br/><span>{profileShippingAddress.city}, {profileShippingAddress.region}, {profileShippingAddress.country} {profileShippingAddress.postal}</span>
+          <span>{profileShippingAddress && profileShippingAddress.fullName}</span>
+          <br/>
+          <span>{profileShippingAddress && profileShippingAddress.address1}</span>
+          {profileShippingAddress && profileShippingAddress.address2 &&
+            <span><br/>{profileShippingAddress.address2}</span>
+          }
+          <br/>
+          <span>
+            {profileShippingAddress && profileShippingAddress.city},&nbsp;
```
spencern: This works. 👍 

You _could_ have setup `profileShippingAddress` to pass an empty object if undefined and then just used `profileShippingAddress.property` everywhere, but that's not necessarily better, just different. Not blocking.

imports/plugins/core/orders/client/components/orderSummary.js
```diff
@@ -149,15 +156,24 @@ class OrderSummary extends Component {
...
+            {profileShippingAddress && profileShippingAddress.region},&nbsp;
```
spencern: I'm actually more curious about the need for `&nbsp;` here. Does React not respect spaces here?

kieha: @spencern I added the `&nbsp;` because the addition of `profileShippingAddress &&`s made the line too long, and I had to break it into different lines, therefore line 172 and 173 end up being two strings without space in between them.

But if I did this:
> You could have setup profileShippingAddress to pass an empty object if undefined and then just used profileShippingAddress.property everywhere.

then I wouldn't have to change the line/add `&nbsp;`s.

imports/plugins/core/orders/client/templates/workflow/shippingTracking.js
```diff
@@ -115,7 +123,7 @@ Template.coreOrderShippingTracking.helpers({
     const currentData = Template.currentData();
     const order = Template.instance().order;
 
-    const shippedItems = _.every(currentData.fulfillment.items, (shipmentItem) => {
+    const shippedItems = _.every(currentData.fulfillment && currentData.fulfillment.items, (shipmentItem) => {
```
spencern: This should use prototype `every` instead of `lodash` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

imports/plugins/core/orders/client/templates/workflow/shippingTracking.js
```diff
@@ -132,7 +140,7 @@ Template.coreOrderShippingTracking.helpers({
     const currentData = Template.currentData();
     const order = Template.instance().order;
 
-    const canceledItems = _.every(currentData.fulfillment.items, (shipmentItem) => {
+    const canceledItems = _.every(currentData.fulfillment && currentData.fulfillment.items, (shipmentItem) => {
```
spencern: Use prototype `every` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

imports/plugins/core/orders/client/templates/workflow/shippingTracking.js
```diff
@@ -149,7 +157,7 @@ Template.coreOrderShippingTracking.helpers({
     const currentData = Template.currentData();
     const order = Template.instance().order;
 
-    const completedItems = _.every(currentData.fulfillment.items, (shipmentItem) => {
+    const completedItems = _.every(currentData.fulfillment && currentData.fulfillment.items, (shipmentItem) => {
```
spencern: prototype instead of lodash

imports/plugins/core/orders/client/components/orderSummary.js
kieha: Yeah, this makes sense. Avoids having to repeat `profileShippingAddress &&` in a lot of places.

imports/plugins/core/orders/client/templates/workflow/shippingTracking.js
```diff
@@ -187,13 +196,15 @@ Template.coreOrderShippingTracking.helpers({
...
   },
   shipmentReady() {
     const order = Template.instance().order;
-    const shipment = order.shipping[0];
+    const shipment = getShippingInfo(order);
+    const shipmentWorkflow = shipment.workflow;
 
-    return _.includes(shipment.workflow.workflow, "coreOrderWorkflow/packed") && shipment.tracking ||
-    _.includes(shipment.workflow.workflow, "coreOrderWorkflow/packed");
+    return _.includes(shipmentWorkflow && shipmentWorkflow.workflow, "coreOrderWorkflow/packed") && shipment.tracking
```
spencern: use prototype `includes` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

imports/plugins/included/search-mongo/server/methods/searchcollections.js
```diff
@@ -185,8 +185,8 @@ export function buildOrderSearchRecord(orderId) {
 
   orderSearch.billingName = shopBilling.address && shopBilling.address.fullName;
   orderSearch.billingPhone = _.replace(shopBilling.address && shopBilling.address.phone, /\D/g, "");
-  orderSearch.shippingName = shopShipping.address.fullName;
-  orderSearch.shippingPhone = _.replace(shopShipping.address.phone, /\D/g, "");
+  orderSearch.shippingName = shopShipping.address && shopShipping.address.fullName;
+  orderSearch.shippingPhone = _.replace(shopShipping.address && shopShipping.address.phone, /\D/g, "");
```
spencern: Any reason we can't use prototype `replace` here? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

server/methods/core/orders.js
```diff
@@ -290,10 +300,13 @@ export const methods = {
       ordersInventoryAdjust(order._id);
     }
 
+    const billingRecord = order.billing.find((billing) => { return billing.shopId === Reaction.getShopId(); });
+    const shippingRecord = order.shipping.find((shipping) => { return shipping.shopId === Reaction.getShopId(); });
```
spencern: on a single line, you can use the shorthand find if you prefer.
```
order.shipping.find((shipping) => shipping.shopId === Reaction.getShopId());
```
Just trying to avoid using the shorthand for multi-line functions.

kieha:
@spencern I've addressed the most recent changes that you had requested. I believe this is good to go now.

spencern:
Only issue was when clicking 

image

Client Error:
```
Error invoking Method 'notification/send': Match failed [400]
```

Server Error:
```
Exception while invoking method 'notification/send' Error: Match error: Expected string, got null
    at exports.check (packages/check.js:55:15)
    at [object Object].notificationSend (imports/plugins/included/notifications/server/methods/notifications.js:21:5)
    at packages/check.js:128:16
    at packages/ddp-server/livedata_server.js:717:46
Sanitized and reported to the client as: Match failed [400]
```

spencern:
Turns out that issue is present in the `marketplace` branch.
Approved 👍

PR #2824, opened 2017-09-11
[WIP] Stripe Connect Capture charge

### LINKED ISSUE #2765: Capture Stripe Connect Direct Charges
The Stripe Connect implementation is currently incomplete.

The [stripe/payment/createCharges](https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/included/payments-stripe/server/methods/stripe.js#L169) method is capable of creating direct charges on behalf of another stripe account, but the [our stripe capture method](https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/included/payments-stripe/server/methods/stripe.js#L351) does not currently support capturing direct charges.

This ticket is for updating that capture payment method to support capturing payment for charges created by Stripe Connect on behalf of another stripe account (e.g. for a marketplace merchant).

I'm not 100% certain why this doesn't work as is, but my guess would be that we need to pass the merchant `stripe_account` into the charges/:id/capture call. 

I think this is a good time to update the existing Stripe capture method to use the node-stripe lib as well and you should be able to follow the patterns set in the `stripe/payment/createCharges` method to figure that part out.

**Resources and other notes**
* https://stripe.com/docs/connect/direct-charges - Stripe Direct Charges docs, note the `stripe_account` argument which is how we determine which stripe account/shop to create a charge on behalf of.
* https://stripe.com/docs/api#capture_charge - capture_charge docs (note: we're creating _direct_ charges not _destination_ charges, so the destination field does not apply here.
* Here's the Reaction stripe-payments code where we actually hit the stripe API to capture: https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/included/payments-stripe/server/methods/stripeapi.js#L101

zenweasel:
Resolves #2765 

Also makes Approvals happen and be by shop for each order.

**Note** This adds filtering by shop for an order. This feels a little bit not great but seems to work. Take a close look at the `reactiveAggregate.js` file that allows us to treat an aggregated result like a collection. If it **does** work then we can use it everywhere we need to filter collections by shop.

### To Test

1. Place a multi-shop order
1. Go to the orders panel for either shop
1. Approve the order
1. Open the other shop in incognito
1. Observe the same order for the second shop is still not approved
1. Select "capture" in the first shop
1. Observe amount is captured and order status in other shop remains the same


spencern:
Probably not the place for this, but I'm still pretty sure we decided we were going to remove the `Approval` step from our payment workflow. @aaronjudd any thoughts?

server/publications/collections/orders.js
```diff
@@ -63,10 +65,51 @@ Meteor.publish("CustomPaginatedOrders", function (query, options) {
...
+          sessionId: 1,
+          shopId: 1,
+          workflow: 1,
+          discount: 1,
+          tax: 1,
+          email: 1,
+          createdAt: 1
+        }
+      }
+    ], selector);
```
spencern: should this `selector` be `{ observeSelector: selector }` or similar?

server/publications/collections/orders.js
```diff
@@ -63,10 +65,51 @@ Meteor.publish("CustomPaginatedOrders", function (query, options) {
...
   }
+
+  // return any order for which the shopId is attached to an item
+  const selector = {
+    "items.shopId": shopId
+  };
   if (Roles.userIsInRole(this.userId, ["admin", "owner", "orders"], shopId)) {
-    Counts.publish(this, "order-count", Orders.find({ shopId: shopId }), { noReady: true });
-    return Orders.find({ shopId: shopId });
+    ReactiveAggregate(this, Orders, [
```
spencern: How is this filtered initially?

server/publications/collections/orders.js
```diff
@@ -63,10 +65,51 @@ Meteor.publish("CustomPaginatedOrders", function (query, options) {
...
+            $filter: {
+              input: "$shipping",
+              as: "shipping",
+              cond: { $eq: ["$$shipping.shopId", shopId] }
+            }
+          },
+          cartId: 1,
+          sessionId: 1,
+          shopId: 1,
+          workflow: 1,
```
spencern: We probably should make sure that any give shop can't move the overall order workflow forward. The overall workflow should only be progressed/regressed by some type of system process or hook that knows when all shipments/payments/items have been progressed.

Also discussion topic, maybe we don't use this field at all? Maybe we keep it for marketplace?

server/publications/collections/orders.js
```diff
@@ -63,10 +65,51 @@ Meteor.publish("CustomPaginatedOrders", function (query, options) {
...
+          discount: 1,
+          tax: 1,
```
spencern: Let's look into this tax and make sure it's marketplace compatible.

server/publications/collections/orders.js
```diff
@@ -63,10 +65,51 @@ Meteor.publish("CustomPaginatedOrders", function (query, options) {
...
+  // return any order for which the shopId is attached to an item
+  const selector = {
+    "items.shopId": shopId
+  };
   if (Roles.userIsInRole(this.userId, ["admin", "owner", "orders"], shopId)) {
-    Counts.publish(this, "order-count", Orders.find({ shopId: shopId }), { noReady: true });
-    return Orders.find({ shopId: shopId });
+    ReactiveAggregate(this, Orders, [
+      {
+        $project: {
```
spencern: Overall, I think using a projection like this is definitely the right way to filter orders for marketplaces.

server/publications/collections/orders.js
```diff
@@ -63,10 +65,51 @@ Meteor.publish("CustomPaginatedOrders", function (query, options) {
...
+          items: {
+            $filter: {
+              input: "$items",
+              as: "item",
+              cond: { $eq: ["$$item.shopId", shopId] }
```
spencern: Let's consider using `$in` and passing in an array of shopIds instead of checking against one shopId. That would permit marketplace owners or multi-shop admins to see several shops rather than just one.

That being said, it might make the UI far more complicated, so it's possible that it's better to permit those types of users to switch between the shops they are viewing the order as.

@rymorgan any thoughts on this?

imports/plugins/included/payments-stripe/server/methods/stripeapi.js
```diff
@@ -43,118 +43,80 @@ const expectedErrors = [
...
-        };
-        return { error: normalizedError, result: null };
-      }
-      Logger.error("Received unexpected error type: " + e.rawType);
-      Logger.error(e);
-
-      // send raw error to server log, but sanitized version to client
-      const sanitisedError = {
-        details: "An unexpected error has occurred"
+StripeApi.methods.createCharge = function ({ chargeObj, apiKey }) {
```
spencern: I don't think this method is used anymore, replaced by `createCharges`.

imports/plugins/included/payments-stripe/server/methods/stripeapi.js
```diff
@@ -43,118 +43,80 @@ const expectedErrors = [
...
-  }).validator(),
-  run({ transactionId, apiKey }) {
-    let stripe;
-    if (!apiKey) {
-      const dynamicApiKey = StripeApi.methods.getApiKey.call();
-      stripe = require("stripe")(dynamicApiKey);
-    } else {
-      stripe = require("stripe")(apiKey);
-    }
+StripeApi.methods.captureCharge = function ({ transactionId, captureDetails, apiKey }) {
```
spencern: Once all of this "works" for both direct and regular charges, I'd like to see us use a consistent style from method to method. The `stripe/payment/createCharges` method uses the `export const methods` style vs these using the `StripeApi.methods`let's figure out which direction we want to go here and update the other ones to match.

It seems like the rest of the app uses the `export const methods` style.

@aaronjudd or @machikoyasuda any preference on method naming conventions?

server/publications/collections/reactiveAggregate.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/**
+ * Create a Reactive collection from a Mongo aggregate pipeline
+ * @param {Object} sub - The publication we are creating
+ * @param {Collection} collection - the collection we are operating on
+ * @param {Array} pipeline - The mongo aggregation pipeline to run
+ * @param {Object} options - an object of options
+ * @returns {Cursor} A mongo cursor for subscription
+ * @constructor
```
spencern: Should this be using `@constructor`? My understanding of that tag would be that it's used for classes.
 @machikoyasuda ?

server/publications/collections/reactiveAggregate.js
```diff
@@ -0,0 +1,102 @@
...
+ */
+export function ReactiveAggregate(sub, collection, pipeline, options) {
```
spencern: Should this be on the Reaction namespace? and where should it live?

imports/plugins/included/payments-stripe/server/methods/stripeapi.js
zenweasel: Yeah, the `StripeApi` was a convention created before we had import/exports. I think we can eliminate it and combine `stripeapi.js` and `stripe.js` into one file. Most of this was to allow for testing but I question the value of doing it that way now.

server/publications/collections/orders.js
zenweasel: In Mongo3.2, `$in` is not supported in projections. It is in 3.4

zenweasel:
Closing this until I have finished all stages of Stripe Connect.

Issue #2743, opened 2017-08-25
Inventory tracking can't be disabled properly


janus-reith:
The tracking of available items cannot be disabled, even when `inventoryPolicy` and `inventoryManagement` are are set to false, and the inventory plugin is disabled.

### Expected behavior
When inventory is disabled, approving an order in the backend should not alter inventoryQuantity.

### Actual Behavior
A product can be ordered with inventoryQuantity 0, and the checkout can be completed.
As soon as the the Admin clicks Approve in the backend, the inventorQuantity gets decremented -1, no matter which values `inventoryPolicy` and `inventoryManagement` have.
Further tries to add this product to cart will fail, as the inventoryQuantity will go below 0 at some point, so the Product can't be validated against the schema anymore - The addtoCart method throws an error.

### Steps to Reproduce the Behavior
Create a variant, use a GUI like Robomongo to monitor its inventoryQuantity field.
Disable inventory plugin, and inventory tracking of the variant.
Set its quantity to 0, add one to cart, checkout and approve.
Notice that inventoryQuantity changed to -1, and adding it to Cart again will fail.

### Versions
Latest - 1.4.1

janus-reith:
Hey @aaronjudd, I just noticed the `needs-more-detail` label.
Were you unable to reproduce the issue? - Which further information can I give to help out?

To sum it up, from what I see atm there is no proper way to disable inventory and stop products from running out of stock.

**The main issue:**
**The method that is triggered when you click Approve for an order does not check for inventory tracking, and will always decrement inventoryQuantity by 1.
Instead, if the product/variant has inventoryPolicy/inventoryManagement set to false, it should not alter that value at all.**

There are those three possible situations;

- If Inventory tracking  is enabled in the product, it will run out of stock, display an inline alert for the user and can't be added to cart anymore(like expected).

- Furthermore, if the inventory plugin is enabled, and Quantity is zero, no matter if the product has inventory tracking, approving the order in the Backend throws an error and won't succed (inventoryQuantity smaller than 0 => Invalid Schema).

- If inventory plugin is removed, and the product has inventory tracking disabled, Approving an order in the backend works, but will still decrement inventoryQauntity, so addtoCart will throw an error as soon as Quantity goes below zero.(=> Invalid Schema).

aaronjudd:
[...]

PR #2825, opened 2017-09-11
Added inventoryPolicy check and min removed from Schema

Akarshit:
Fix for #2743 

1. Removed that `min: 0` check from `Product` schema.
2. Added a condition to check for the inventoryPolicy before rejecting a order.

But this would cause the inventoryQuantity to be as negative as the no. of backorders on that variant. Is this the required behaviour?

### DIFF
--- a/imports/plugins/included/inventory/server/methods/statusChanges.js
+++ b/imports/plugins/included/inventory/server/methods/statusChanges.js
@@ -235,9 +235,9 @@ Meteor.methods({
     check(backOrderQty, Number);
     this.unblock();
 
-    // this use case could happen then mergeCart is fires. We don't add anything
+    // this use case could happen when mergeCart is fired. We don't add anything
     // or remove, just item owner changed. We need to add this check here
-    // because of bulk operation. It thows exception if nothing to operate.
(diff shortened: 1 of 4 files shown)


aaronjudd:
@Akarshit this looks good, but can you resubmit to marketplace (or we can release if there is a 1.4.3, but I'd rather just bring it into marketplace now as we're doing a fixes to the PDP).  @spencern let's keep an eye and cherry-pick if needed to move this along.

Akarshit:
@spencern Changed base and updated the PR.

spencern:
This PR raises a few issues that are related but not caused by this PR. I'll create separate issues for these, but also list them here.

1. Toggling inventoryPolicy / "Allow Backorder" for a variant does _not_ cascade that choice into it's options
2. Options do not have a switch that will permit toggling the inventory policy. I had to edit the inventoryPolicy for the options in the database directly in order to permit backordering on the default product.

image

spencern:
https://github.com/reactioncommerce/reaction/issues/2911 is the new issue that I created to address the additional issues noticed here.

Issue #2820, opened 2017-09-09
Possible error with meteor -> reaction integration


mikicho:
### Expected behavior
`reaction init` will succeed
### Actual Behavior
Failed with following log: [link](https://gist.github.com/mikicho/1ee2f0434034ef07779eac7cf0e8831d)
### Steps to Reproduce the Behavior
Follow after instructions in [requirements](https://docs.reactioncommerce.com/reaction-docs/development/requirements)
### Versions
Node: 6.11.3
NPM: 3.10.10
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
windows 10 x64

Installation on my machine working only if I remove `meteor` from `meteor npm install` (i.e run npm of my machine instead the one that bundle with meteor) in `initInstallModules` function in `\npm\node_modules\reaction-cli\dist\utils\node_modules.js` line 38

the problem is, that didn't help because then reaction didn't find any module.

mikicho:
Update: this probably because I didn't had openSSH installed on my machine, I think it's should insert into requirements for windows, cause there is no ssh executable by defaul, personally I installed from [here](http://www.mls-software.com/opensshd.html).
Now I'm encounter allocation problem:
```
PS path\reaction> reaction
Using settings file at settings/dev.settings.json
Setting up plugin imports...
Setting up style imports...
[[[[[ ~\pathe\reaction ]]]]]
=> Started proxy.
=> Meteor 1.5.2 is available. Update this project with 'meteor update'.
=> Started MongoDB.
cfs:tempstore: updating npm dependencies -- combined-stream...
cfs:gridfs: updating npm dependencies -- mongodb, gridfs-stream...
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory
```

pascalprat:
I get the same errors in my browser console after a fresh install.

pascalprat:
By running `reaction debug`, I discovered there was a missing module named `react-addons-shallow-compare`.
`meteor npm install --save react-addons-shallow-compare` solved the problem.
It should maybe be fixed in the `reaction init` script.

aaronjudd:
@machikoyasuda we should review our windows docs for this for clarity, as well as , in general when npm packages are missing, it's because there was a failure to download/find/install/dependency, and `meteor npm install` is usually the fix.. (paying attention to remote source installation failures)

PR #2829, opened 2017-09-11
Adding react-addons-shallow-compare dependency

josx:
Fix https://github.com/reactioncommerce/reaction/issues/2820

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "react": "15.4.2",
     "react-addons-create-fragment": "15.4.2",
     "react-addons-pure-render-mixin": "15.4.2",
+    "react-addons-shallow-compare": "^15.6.0",
     "react-autosuggest": "^9.0.0",
     "react-avatar": "^2.3.0",
     "react-color": "^2.11.4",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "reaction",
   "description": "Reaction is a modern reactive, real-time event driven ecommerce platform.",
-  "version": "1.4.1",
+  "version": "1.4.2",
   "main": "main.js",
   "directories": {
     "test": "tests"


zenweasel:
@josx When I do `meteor npm install` on my local installation I get this package automatically because it is a dependency of `react-avatar`. Can you confirm you are having this problem on a clean install?

josx:
yes, i followed the documentation to start a clean installation, and i got that error.

I have done the following:
```
npm install -g reaction-cli
reaction init
cd reaction
reaction
```

zenweasel:
@josx Yeah, testing with a fresh install I am seeing the same issue. I wonder if we need to bump up `react-avatar` or something though? I hesitate to hardcode a dependency in like that.

jshimko:
The output of `reaction init` spits this out deep in the middle of the npm install...

```
npm WARN react-avatar@2.4.0 requires a peer of react-addons-shallow-compare@^0.14.0 || ^15.0.0 || ^16.0.0 but none was installed.
```

This is because it is only [declared as a peer dependency](https://github.com/Sitebase/react-avatar/blob/master/package.json#L32).  It's [also in the devDependencies](https://github.com/Sitebase/react-avatar/blob/master/package.json#L64) but those don't get installed when you install a package in your app.  

So any time this hasn't been a problem in the past was likely because something else _did_ have it as a hard dependency.  So maybe that has changed in whatever package installed it for us.  Either way, that is a legit dependency for us now, so I guess we need to move forward with this.  On the upside, a peer dependency allows us to pick any version satisfied by `^0.14.0 || ^15.0.0 || ^16.0.0` (which is way better than being stuck with whatever the 3rd party package uses.

josx:
This is a hotfix (right now, you cant install the app smoothly).

[react-addons-shallow-compare](https://www.npmjs.com/package/react-addons-shallow-compare) is legacy react code, so any library using must be replaced with [React.PureComponent](https://facebook.github.io/react/docs/react-api.html#react.purecomponent)

jshimko:
Lovely.  So that means the real fix is a PR to someone else's package.

josx:
Yes i think that.

Also I have checked [readme on react-avatar](https://github.com/Sitebase/react-avatar)  and you need to install react-addons-shallow-compare aside

josx:
BTW I have oponed an issue on react-avatar about this  https://github.com/Sitebase/react-avatar/issues/66

mikemurray:
At the very least, this issue seems to have been fixed in a PR to the marketplace branch. https://github.com/reactioncommerce/reaction/pull/2841

josx:
Ok, so if that code will be on master soon, we can close this PR

josx:
Anyways you have to know that you cant install the app from current master branch

PR #2830, opened 2017-09-11
update verbiage for better Backorder display / understanding

### LINKED ISSUE #2828: Backorder on product seems to be working opposite of how it should
### Expected behavior
When `Allow Backorder` is true, products should show the backorder label.
When `Allow Backorder` is false, products should show a sold-out label`
- Create a product with inventory tracking.
- Make inventory 0.
- Toggle `Allow Backorder` to see badges.

### Versions
Node: 4.2.6
NPM: 5.4.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.2-ce

basic_reaction_product

basic_reaction_product 2

kieckhafer:
Resolves #2828 

The terminology used on the PDP in relation to Backorders / inventory is confusing, seemingly displaying the opposite of what is actually happening.

This update simply reverses the display of a toggle switch to show the opposite boolean value, which helps decipher what's actually going on for the end-user.


rymorgan:
This makes sense to me.

mikemurray:
Maybe instead of reversing the value, we can ensure that when the switch is `on` it's  `true` regardless of what ever the initial boolean value was that was supplied to it?

Not sure re-reversing the value the switch provided, which was already reversed, makes this less confusing.

spencern:
@mikemurray that's how it was intended to work originally, but would involve changing the language back to something similar to "Deny when out of stock" as is still listed as the label for the schema.

mikemurray:
@spencern makes sense. 👍

imports/plugins/included/product-variant/components/variantForm.js
```diff
@@ -152,6 +152,29 @@ class VariantForm extends Component {
...
+    /*
+    Due to some confusing verbiage on how inventoryPolicy works / is displayed, we need to handle this field
+    differently than we handle the other checkboxes in this component. Specifically, we display the opposite value of
+    what the actual field value is. Because this is a checkbox, that means that the opposite value is actually the
+    field value as well, not just a display value, so we need to reverse the boolean value when it gets passed into
+    this function before we send it to the server to update the data. Other than reversing the value, this function
+    is the same as `handleCheckboxChange`.
+    */
+
+    const reverseValue = !value;
```
spencern: Nitpicky but I'd prefer `inverseValue`

Issue #2804, opened 2017-09-07
Update email copy for inviting a user to create a shop


spencern:
We need to update the copy for the email that goes out when a marketplace admin invites a new user to create/manage a new shop on their marketplace platform.

This issue can serve as the base for the discussion as to what this email should contain.

Here is the current email:
![image](https://user-images.githubusercontent.com/1203639/30175180-e285ba46-93ba-11e7-9088-8f1aaa154eea.png)

At a minimum, the updated copy should clarify:
- Who/which marketplace is doing the inviting, right now you're being invited by the new shop that you're asked to create. This doesn't make sense.
- What you'll need to do to accept this invitation
- That this is _not_ marketing email and you've been invited directly by the marketplace manager.

We can - and perhaps should - include the new name of the shop in the form when the invite is created rather than appending a "1" or leaving off the shop name?

Other thoughts on this?

kieckhafer:
I think there are 6 places where we need to decide which shop information is being applied. I think all 6 should use the main marketplace shop, instead of the newly created / invited shop. 4 & 5 might be removed if we get new copy, but the rest will probably stay.

<img width="985" alt="airmail" src="https://user-images.githubusercontent.com/4482263/30338208-a6797f04-979f-11e7-9b31-3628fb7b31f8.png">

Thoughts?

spencern:
I'm for updating all of them to use the main marketplace. Possible that as we update the invite flow, we might be able to use the `ShopName` they are invited to manage somewhere (`Admin has invited you to manage a shop` feels so bleh)

rymorgan:
@spencern @kieckhafer -- Here's my take. It's a bit generic b/c I would imagine most marketplaces would want to add their own style to it. 

Subject line: $Marketplace has invited you to join as a shop

Email:
Hi, $username 

You’ve been invited by $Marketplace to set up a shop. Click below to configure your shop and start adding products. 

Get Started CTA

—————
You received this email because $Marketplace invited you to manage a shop. Questions and suggestions? Email us at info@$marketplace.com

spencern:
I think if we get the language consistent it's definitely better than what we have now.
I think I lean what if we used "create a store"? Of the three you've got I like "manage a store best" but not sure that fully conveys the idea.

![image](https://user-images.githubusercontent.com/1203639/30412038-1cef4c1c-98d2-11e7-8e40-87348f2c45eb.png)

rymorgan:
@spencern Fair point. 

[...]

PR #2835, opened 2017-09-12
add primary shop data to dataForEmail object

kieckhafer:
Add primary shop info to the `dataForEmail` object to get marketplace shop info for use in emails, in addition to the current store info.

Changes the send from email address / name on the `Invite New Shop` email, to send from main marketplace shop instead of newly created shop.

Resolves updating of information for #2804 (does not resolve full issue, copy for emails is still needed).


server/methods/accounts/accounts.js
```diff
@@ -580,7 +580,7 @@ export function inviteShopOwner(options) {
 
   Reaction.Email.send({
     to: email,
-    from: `${shop.name} <${_.get(shop, "emails[0].address")}>`,
+    from: `${_.get(dataForEmail, "primaryShop.name")} <${_.get(dataForEmail, "primaryShop.emails[0].address")}>`,
```
kieckhafer: @spencern I have no reason for using `_.get` vs. just `dataFormEmail.primaryShop.name` here, I just followed the convention we were currently using. If it's better not to invoke `lodash`, I will change it.

spencern: I'd prefer not to use lodash here, I don't think `_.get` is very clear unless you're recently versed in lodash terminology, so if you don't mind gutting it, that'd be better.

server/methods/accounts/accounts.js
```diff
@@ -580,7 +580,7 @@ export function inviteShopOwner(options) {
 
   Reaction.Email.send({
     to: email,
-    from: `${_.get(dataForEmail, "primaryShop.name")} <${_.get(dataForEmail, "primaryShop.emails[0].address")}>`,
+    from: `${dataForEmail.primaryShop.name} <${dataForEmail.primaryShop.emails[0].address}>`,
```
spencern: One thing that `_.get` does here is guard against errors, so we need to be 100 percent certain that primaryShop.name and primaryShop.emails[0].address always exist in this scenario, if there's a chance that's not the case, we should probably build `const`s that guard for us before calling send.

server/methods/accounts/accounts.js
kieckhafer: reverted back to using _.get() as discussed

spencern: For anyone following along at home, decided to keep `_.get` for now because it automatically guards against throwing an TypeError when trying to access a property of something that is undefined.

We're currently using `_.get` throughout the app like this and we should discuss what lodash features are permitted within Reaction and either standardize on using _.get, or eliminate it.

PR #2843, opened 2017-09-13
Fix Shopify issue importing multiple images for top level products

spencern:
Fixes an issue where when importing products from a Shopify store, the importer would only import the first image of the top level product.


kieckhafer:
I'm guessing this is how their shop is set up, so I don't think this is a comment on how this code is working - which it seems to be. However, all the images seem to be attached to a variant, and not a variant option. So I select a size, and i see all the pictures:

bondage_bralette

But then I select a color, and only one picture is displayed

bondage_bralette

Is that something we should be trying to remedy? Or do we just chalk it up to "that's how your shop is set up, that's how we import"?

spencern:
So yeah, this is the way their shop is setup. If you're browsing the product in question on shopify in admin, you can append `.json` to the url and see exactly what we import.

In this user's case, almost every product is attached to the top level product, and very few are attached to variants.

Shopify doesn't actually have a concept of multiple levels of variants or "options" as we call them, so we just attach the images to whatever the lowest level of variant happens to be, in this case it would be the option.

We actually don't attach images to the variant level if there are options available, though I could hear an argument for attaching images to both. There's not really a 1-to-1 comparison here.

tldr; unless a product is attached to the variant in Shopify, they won't be attached to a variant in Reaction.

spencern:
In regards to your first question about all the images being attached to a variant. It's actually showing all images because there are _no_ images attached to a variant.

I can think through that a little bit and try to come up with a better way to import images for products that have variants and options.

PR #2855, opened 2017-09-14
Create reactive aggregate

zenweasel:
At Spencer's suggestion I am creating a separate PR for this.

This adds new class called `ReactiveAggregate` that allows you to use Mongo aggregates as a normal Meteor publication. The primary use case for this now is allowing us to filter things like Orders where we want to filter out items within an order rather than the whole order.

I'm including the Order filtering here so we can see an example of it in use (and because I think that's the primary use case right now). As a store admin you should only see the billing/shipping/items on orders attached to your store. So to a store admin, it will appear that only your parts of the order are the entire order.

### DIFF
--- a/server/publications/collections/orders.js
+++ b/server/publications/collections/orders.js
@@ -12,8 +12,9 @@ import { Reaction } from "/server/api";
  * @param {Object} sort - An object containing a sort
  * @param {Number} limit - An optional limit of how many records to return
  * @returns {Array} An array of projection operators
+ * @private
  */
-export function createAggregate(shopId, sort = { createdAt: -1 }, limit = 0) {
+function createAggregate(shopId, sort = { createdAt: -1 }, limit = 0) {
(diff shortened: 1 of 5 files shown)


zenweasel:
Changed this back to WIP as tests are failing

server/publications/collections/orders.js
```diff
@@ -1,13 +1,63 @@
...
+
 /**
- * orders
+ * A shared way of creating a projection
+ * @param {String} shopId - shopId to filter records by
+ * @param {Object} sort - An object containing a sort
+ * @param {Number} limit - An optional limit of how many records to return
+ * @returns {Array} An array of projection operators
  */
+export function createAggregate(shopId, sort = { createdAt: -1 }, limit = 0) {
```
spencern: is this just for creating an orders aggregation?
Would createOrdersAggregate be more descriptive or is it generic enough to be used for more publications?
In that case would it be better for this to live in another file?

Issue #2821, opened 2017-09-11
Receive error on server when a new user logs on for the first time


zenweasel:
### Expected behavior
No errors

### Actual Behavior
Receive the following error in the server console:
```
Exception in onLogin callback: Error: Match error: Failed Match.OneOf, Match.Maybe or Match.Optional validation
    at exports.check (packages/check.js:56:15)
    at [object Object].cartMergeCart (server/methods/core/cart.js:114:5)
    at packages/check.js:129:16
    at [object Object]._.extend.withValue (packages/meteor.js:1126:17)
    at Object.exports.Match._failIfArgumentsAreNotAllChecked (packages/check.js:128:41)
    at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1765:18)
    at packages/ddp-server/livedata_server.js:1686:15
    at [object Object]._.extend.withValue (packages/meteor.js:1126:17)
    at packages/ddp-server/livedata_server.js:1684:36
    at [object Object]._.extend.applyAsync (packages/ddp-server/livedata_server.js:1683:12)
    at [object Object]._.extend.apply (packages/ddp-server/livedata_server.js:1622:26)
    at [object Object]._.extend.call (packages/ddp-server/livedata_server.js:1604:17)
    at server/startup/accounts.js:237:21
    at runAndHandleExceptions (packages/callback-hook.js:152:24)
    at packages/callback-hook.js:159:12
    at packages/accounts-base/accounts_server.js:164:5
```

### Steps to Reproduce the Behavior
1. Login as admin
1. Invite a user to a shop
1. Open the invite email and copy the URL
1. Open that URL in an incognito browser
1. Observe error

### Versions
```
Node: 7.10.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: brent-fix-issue-2765
Docker: 17.06.2-ce
```

zenweasel:
Closed via #2857

PR #2857, opened 2017-09-14
Fix "Error on server when a new user logs on for the first time" bug 

efalayi:
Resolves [#2821](https://github.com/reactioncommerce/reaction/issues/2821)
Server displays the error when new user logs in for the first time.

The error occurred because currentSessionId (set as a boolean value) was parsed as a parameter to `cart/mergeCart` function. However, currentSessionId was supposed to be set as a string.

###### Previous definition for currentSessionId
```
const currentSessionId = options.methodArguments &&
options.methodArguments.length === 1 &&
options.methodArguments[0].sessionId;
```

###### Fix
options is an object which returns the following
```
{ type: 'password',
  allowed: true,
  methodName: 'resetPassword',
  methodArguments:
   [ 'dRqG4aeaPvhb4kP9N',
     { digest: '4ad63f4d1273a10b3697a173cb991e5378eadd558564c4a8b24099069b4cdaf7',
       algorithm: 'sha-256' } ],
  user:
   { _id: 'xo45Pba8cbKJigs4C',
     createdAt: Thu Sep 14 2017 11:12:02 GMT+0100 (WAT),
     services: { password: [Object], resume: [Object] },
     emails: [ [Object] ],
     profile: {},
     roles: { J8Bhq3uTtdgwZx3rz: [Object] },
     name: 'Mayowa Makinde' },
  connection:
   { id: 'Mm4GKcDu6Cymi3Y4D',
     close: [Function],
     onClose: [Function],
     clientAddress: '127.0.0.1',
     httpHeaders:
      { 'x-forwarded-for': '127.0.0.1',
        'x-forwarded-proto': 'ws',
        host: 'localhost:3000',
        'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.12; rv:55.0) Gecko/20100101 Firefox/55.0',
        'accept-language': 'en-US,en;q=0.5' } } }
```

solution 1: set currentSessionId to equal options.methodArguments[0]
solution 2: set currentSessionId to equal options.connection.id

Test
```
Login as admin
Invite a user to a shop
Open the invite email and copy the URL
Open that URL in an incognito browser
Observe console
```


zenweasel:
@efalayi So let's give this a more descriptive name that will look good in release notes?

server/startup/accounts.js
```diff
@@ -229,9 +229,9 @@ export default function () {
       // in current version currentSessionId will be available for anonymous
       // users only, because it is unknown for me how to pass sessionId when user
       // logged in
-      const currentSessionId = options.methodArguments &&
-        options.methodArguments.length === 1 &&
-        options.methodArguments[0].sessionId;
+
+      // currentSessionId should be a string but the former declaration assign currentSessionId a boolean value.
```
zenweasel: Generally we don't need comments in code to explain why a change was made. That's for commit messages and PR notes. This comment won't mean anything to anybody trying to understand this code.

server/startup/accounts.js
```diff
@@ -229,9 +229,9 @@ export default function () {
...
+      const currentSessionId = options.methodArguments[0];
```
zenweasel: So, what this previous code was probably _trying_ to do was ensure that both `options.methodArgument` existed and that it was a non-zero-length array. We probably need to dig a little deeper and determine if this is a value that sometimes appears (so we need to wrap this in a conditional before assigning it) or it should always be there so we are okay with throwing an uncaught error if it doesn't exist.

efalayi: Ok. I'm guessing that it's a value that may sometimes be there, as evidenced in `server/methods/core/cart.js`, line 123, where it's expected to be undefined or falsy sometimes. I haven't been able to reproduce those instances though. So, I've done a conditional assignment for its value.

PR #2860, opened 2017-09-14
Order filtering by date

### LINKED ISSUE #2359: Filter orders by created at date
Users with permissions to view orders dashboard should be able to quickly filter the list of orders by the created date.

Users should be able to set either or both `beforeDate` and `afterDate` fields and return orders that were created between those dates and times.

If the `beforeDate` is not set, the filter should return all orders that were created after the date provided to the `afterDate` field

If the `afterDate` is not set, the filter should return all orders that were created before the date provided to the `beforeDate` field

If both dates are set, the filter should return all orders created between `afterDate` and `beforeDate`

Any other filters, such as status, should also apply.

Searches performed with this filter active should take it into account.

From a UI perspective, this might be best accomplished via a calendar range selector of sorts that could live persistently on the orders page.

joykare:
Resolves #2359 

### How to test
- Make a couple of orders (better if you change the time orders are created to observe that the date filter works)
- Pick a range of dates
- Observe that orders are filtered correctly
- Try filtering by status and searching for a particular group of orders
- Observe that the date filter works in tandem with the other filter and search


rymorgan:
@joykare here are my thoughts on the date picker styling. https://zpl.io/bz8M3ME

joykare:
@rymorgan I did not add question-label at the bottom of the calendar because it does not have any functionality (the keyboard shortcuts on react-dates don't work with this particular implementation).

Otherwise good for a quick check

rymorgan:
@joykare I'll check it out. I only added the question mark b/c I thought the react calendar had that to explain how to use the calendar. But, I think, not including it is fine. I'll check out your PR.

rymorgan:
Some small design tweaks. Looking pretty good though! 

date-pick-feedback-2
date-pick-feedback-1

rymorgan:
Also, the date picker doesn't seem to work for end-date. In the following video I can see my order for the 19th when it's my end date, but the 18th. 

date-pick

rymorgan:
Also, one more thing for the rollover state it looks can you use the hover color from here: 
https://styleguide.reactioncommerce.com/reaction/guide/colors

This one:
@rui-info-hover

imports/plugins/core/orders/client/components/orderActions.js
```diff
@@ -1,112 +1,168 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import classnames from "classnames/dedupe";
-import { Button, DropDownMenu, MenuItem } from "@reactioncommerce/reaction-ui";
+import { Components } from "@reactioncommerce/reaction-components";
 
 class OrderActions extends Component {
   static propTypes = {
-    className: PropTypes.string,
+    className: PropTypes.object,
```
zenweasel: This variable name sounds like it should be a string but it appears it's not now. Can we give it a more descriptive name?

imports/plugins/core/orders/client/components/orderDashboard.js
```diff
@@ -18,6 +18,7 @@ class OrderDashboard extends Component {
     handleSelect: PropTypes.func,
     isLoading: PropTypes.object,
     multipleSelect: PropTypes.bool,
+    onDatesChange: PropTypes.func,
```
zenweasel: Is there some sort of pattern you are using for naming functions `onThisChange` as opposed to `handleThisChange`? If not I would like to standardize on one or the other. If so, let's codify this in a style guide somewhere so we can be consistent.

imports/plugins/core/ui/client/components/calendarPicker/calendarPicker.js
```diff
@@ -0,0 +1,119 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { registerComponent } from "@reactioncommerce/reaction-components";
+import { DayPickerRangeController } from "react-dates";
```
zenweasel: import order

imports/plugins/core/ui/client/components/calendarPicker/calendarPicker.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  onDatesChange = ({ startDate, endDate }) => {
+    this.setState({
+      startDate,
+      endDate
+    });
+
+    if (this.props.onDatesChange) {
+      this.props.onDatesChange(startDate, endDate);
```
zenweasel: Why are we wrapping this in a conditional? Will this function not always exist? And if not shouldn't we throw an error? I feel like this is an anti-pattern that has crept into our code somehow.

Also, maybe this is unavoidable but it seems like having this function and the one passed in via props both having the same name is a recipe for a bug down the road?

imports/plugins/included/default-theme/client/index.js
```diff
@@ -3,3 +3,4 @@ import "./favicons";
 
 // Scripts
 import "bootstrap/dist/js/npm.js";
+
```
zenweasel: Let's not add a blank line for no reason

rymorgan:
Also, it's not working responsively. I think for now we should hide shipping status on phones too. 

orders-responsive-issue

zenweasel:
Ran into one problem with filter but doesn't seem to be related to this PR. (#2902). Otherwise this seems to work well. Just had some minor style things to address and this should be ready to go.

imports/plugins/core/ui/client/components/calendarPicker/calendarPicker.js
joykare: The reason we have that conditional is b/c you dont have to have a prop function for when dates change; but in case you want to add some operations/ or change some stuff onDateChange specific to the component that extends the calendarPicker you could do that under this function.
Other example: https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/ui/client/components/menu/menu.js#L7

joykare:
@zenweasel @kieha good for another look

imports/plugins/core/ui/client/components/calendarPicker/calendarPicker.js
zenweasel: Ok, makes sense to me

rymorgan:
date-picker 1

rymorgan:
In anticipation of merging this PR I made a bug for the mobile responsive issue https://github.com/reactioncommerce/reaction/issues/2910

zenweasel:
@kieha We're still waiting for your official blessing on this

joykare:
@rymorgan fixed the select. Works both ways now.

zenweasel:
@joykare Is making the filter disappear at more narrow width a quick fix you can do here rather than having it linger in another ticket?

joykare:
@zenweasel not as quick a fix as I thought initially, so I asked Ryan to have that as a separate ticket so that we can move along with getting all the filters in.

zenweasel:
Because BH is not working running ESLint gives me these errors
```
/Users/brenthoover/Projects/js/reaction/imports/plugins/core/ui/client/components/calendarPicker/calendarPicker.js
  40:1  error  Expected indentation of 4 spaces but found 2  indent
  42:1  error  Expected indentation of 4 spaces but found 2  indent
  43:1  error  Expected indentation of 6 spaces but found 4  indent
  44:1  error  Expected indentation of 6 spaces but found 4  indent
  45:1  error  Expected indentation of 6 spaces but found 4  indent
  46:1  error  Expected indentation of 6 spaces but found 4  indent
  47:1  error  Expected indentation of 4 spaces but found 2  indent
  49:1  error  Expected indentation of 4 spaces but found 2  indent
  50:1  error  Expected indentation of 6 spaces but found 4  indent
  51:1  error  Expected indentation of 8 spaces but found 6  indent
  52:1  error  Expected indentation of 8 spaces but found 6  indent
  53:1  error  Expected indentation of 8 spaces but found 6  indent
  54:1  error  Expected indentation of 8 spaces but found 6  indent
  55:1  error  Expected indentation of 8 spaces but found 6  indent
  56:1  error  Expected indentation of 8 spaces but found 6  indent
  57:1  error  Expected indentation of 8 spaces but found 6  indent
  58:1  error  Expected indentation of 8 spaces but found 6  indent
  59:1  error  Expected indentation of 8 spaces but found 6  indent
  60:1  error  Expected indentation of 6 spaces but found 4  indent
  61:1  error  Expected indentation of 4 spaces but found 2  indent
  62:1  error  Expected indentation of 2 spaces but found 0  indent
```

Issue #2514, opened 2017-06-30
Translation/i18n of tags and product info


cusspvz:
Hey guys,

I was messing around with reaction for a little bit and it is awesome. I already knew it for a long time but you guys are taking it to a next level and thank you for that.

Inspite, I have a doubt on a feature, since I don't have completely sure if it is possible: **Are the tags and product information supporting i18n / locales?**

Just to let it clear, I was able to change the locale of the interface but I didn't found a way to internationalize tags and/or product info (such as title, description, etc).

If it's not possible, I'm willing to help and contribute on it. Thanks!

spencern:
Hi @cusspvz 

Currently Reaction does not support locales or i18n for user generated content. Contributions would be very welcome.

cusspvz:
I'm interested on developing it. Is there a way we can schedule a Skype talk to discuss a PR approach?

aaronjudd:
@cusspvz we have public group video chats on a regular schedule, see: https://docs.reactioncommerce.com/reaction-docs/master/community-channels (core/components would be a good place to discuss..)

aaronjudd:
This is actually part of a series of epics in the publishing workflow that are in progress. 

Conceptually, I've been discussing supporting multiple languages for content, products by storing independent entries for each language in a normalized "catalog" collection.  The translations process being handled via our publishing workflow, where the scope of the "Publish" button would trigger an event flow allowing translation packages (lingohub, yandex, gengo, transifex,etc) to return translated content, synchronously updating to the normalized catalog when translations completed.

janat08:
I imagine #2641 may have to be reliable, if you were to let admin translate himself where app would need to register language change to reconfigure field bindings. I suppose google with it's AI is better translator.

PolGuixe:
Any progress in this field? What is the current state?

aldeed:
I'm going to move this to the feature request repo. I believe the current status is that this would be implemented as alternative catalogs, but further discussion would be needed.

PR #2862, opened 2017-09-14
[WIP] Internationalization of products and categories content

cusspvz:
Directly related with #2514

**This PR is a Work-In-Progress**

I believe the first stage to include i18n over the product's and category's content is to let the admin change it manually. Then we can accomplish automatic i18n by using a plugin for that effect.

@aaronjudd I would like to hear from you where do you think we should implement the `value => intl object` at the frontend side. Should we apply at the textfield or somewhere into the data schema? If you prefer the last option can you please point where?

Thanks in advance.

### DIFF
--- a/imports/plugins/core/i18n/server/i18n/index.js
+++ b/imports/plugins/core/i18n/server/i18n/index.js
@@ -2,6 +2,7 @@ import { loadTranslations } from "/server/startup/i18n";
 
 // import ar from "./ar.json";
 // import bg from "./bg.json";
+// import cs from "./cs.json";
 // import de from "./de.json";
 // import el from "./el.json";
 import en from "./en.json";
@@ -9,6 +10,7 @@ import en from "./en.json";
 // import fr from "./fr.json";
 // import he from "./he.json";
 // import hr from "./hr.json";
+// import hu from "./hu.json";
 // import it from "./it.json";
 // import my from "./my.json";
 // import nb from "./nb.json";
@@ -29,4 +31,4 @@ import en from "./en.json";
 // automated translation software
 //
 loadTranslations([en]);
-// loadTranslations([ar, bg, de, el, en, es, fr, he, hr, it, my, nb, nl, pl, pt, ro, ru, sl, sv, tr, vi, zh]);
+// loadTranslations([ar, bg, cs, de, el, en, es, fr, he, hr, hu, it, my, nb, nl, pl, pt, ro, ru, sl, sv, tr, vi, zh]);
--- a/lib/collections/schemas/intlfield.js
+++ b/lib/collections/schemas/intlfield.js
@@ -1,0 +1,64 @@
+import { SimpleSchema } from "meteor/aldeed:simple-schema";
+
+/**
+* Reaction Schemas Intlfield
+*/
+
+const defaultAttributeOptions = {
+  type: String,
+  optional: true,
+  defaultValue: ""
+}
+
+const langs = [
+  'ar', 'bg', 'cs', 'de', 'el', 'en', 'es', 'fr', 'he', 'hr', 'hu', 'it', 'my',
+  'nb', 'nl', 'pl', 'pt', 'ro', 'ru', 'sl', 'sv', 'tr', 'vi', 'zh',
+]
+
+function createI18nObject ( string = "" ) {
+  const obj = {}
+
+  for ( let lang of langs ) {
+    obj[lang] = string
+  }
+
+  return obj
+}
+
+function autoValue () {
+  const baseLang = 'en'
+
+  const { value } = this
+
+  if ( value && typeof value == 'object' ) {
+    return value
+  }
+
+  if ( value && typeof value == 'string' ) {
+    this.value = {
+      [baseLang]: createI18nObject(value)
+    }
+  }
+
+  return createI18nObject()
+}
+
+export function schemaBuilder ( i18nAttributeOptions = {}, extraAttributes = {} ) {
+  const attributes = {}
+
+  for ( let lang of langs ) {
+    attributes[lang] = { ...defaultAttributeOptions, ...i18nAttributeOptions }
+  }
+
+  return new SimpleSchema({ ...attributes, ...extraAttributes })
+}
+
+export const Intlfield = schemaBuilder()
+
+export function i18nify ( extraOptions = {}, i18nAttributeOptions, extraAttributes ) {
+  return {
+    type: schemaBuilder( i18nAttributeOptions, extraAttributes ),
+    autoValue,
+    ...extraOptions,
+  }
+}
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "autoprefixer": "^7.1.1",
     "autosize": "^3.0.20",
     "babel-runtime": "^6.23.0",
+    "bcrypt": "^1.0.3",
```
spencern: We've removed `bcrypt` because it's been causing errors in our build process, so while it complains when you start up, it's been removed on purpose.

lib/collections/schemas/intlfield.js
```diff
@@ -0,0 +1,64 @@
...
+  if ( value && typeof value == 'string' ) {
+    this.value = {
+      [baseLang]: createI18nObject(value)
+    }
+  }
+
+  return createI18nObject()
+}
+
+export function schemaBuilder ( i18nAttributeOptions = {}, extraAttributes = {} ) {
```
spencern: needs jsdoc

lib/collections/schemas/intlfield.js
```diff
@@ -0,0 +1,64 @@
...
+  for ( let lang of langs ) {
+    attributes[lang] = { ...defaultAttributeOptions, ...i18nAttributeOptions }
+  }
+
+  return new SimpleSchema({ ...attributes, ...extraAttributes })
+}
+
+export const Intlfield = schemaBuilder()
+
+export function i18nify ( extraOptions = {}, i18nAttributeOptions, extraAttributes ) {
```
spencern: needs jsdoc

lib/collections/schemas/intlfield.js
```diff
@@ -0,0 +1,64 @@
...
+  const obj = {}
+
+  for ( let lang of langs ) {
+    obj[lang] = string
+  }
+
+  return obj
+}
+
+function autoValue () {
```
spencern: needs jsdoc

lib/collections/schemas/intlfield.js
```diff
@@ -0,0 +1,64 @@
...
+  optional: true,
+  defaultValue: ""
+}
+
+const langs = [
+  'ar', 'bg', 'cs', 'de', 'el', 'en', 'es', 'fr', 'he', 'hr', 'hu', 'it', 'my',
+  'nb', 'nl', 'pl', 'pt', 'ro', 'ru', 'sl', 'sv', 'tr', 'vi', 'zh',
+]
+
+function createI18nObject ( string = "" ) {
```
spencern: needs jsdoc

package.json
```diff
@@ -150,7 +151,8 @@
   },
   "babel": {
     "plugins": [
-      "lodash", [
```
spencern: `package.json` should not ever be committed unless for some reason you're adding core functionality that adds a new npm package.

imports/plugins/core/i18n/server/i18n/index.js
```diff
@@ -2,13 +2,15 @@ import { loadTranslations } from "/server/startup/i18n";
 
 // import ar from "./ar.json";
 // import bg from "./bg.json";
+// import cs from "./cs.json";
```
spencern: what's going on in this file? Why add commented lines?

package.json
cusspvz: Yeah, this was a mistake, gonna remove it on a next commit and when everything is cool, we just squash all commits into one and those changes will disappear.

package.json
cusspvz: Same here! :D

imports/plugins/core/i18n/server/i18n/index.js
aaronjudd: @spencern this is the piece we were just discussing: re, might be nice to move to the registry, ie: `translations: ["ar", bg"], etc and the use the plugin loader to import the translations, rather than requiring this loading process in every plugin.

cusspvz: When I was studying the structure of the reaction, I've realised that i18n didn't had all the languages on the comments, and as I saw the comments I though they weren't removed for some reason, so I've added them as a reminder to ask you: There are a bunch of i18n with commented imports, do they matter? Are you guys saving them with any purpose?

lib/collections/schemas/intlfield.js
cusspvz: I'm gonna add JSDocs as soon as everything is working flawless. I can give this case as an example, I don't even know if we're gonna stick with this `autoValue` method.

spencern:
@cusspvz are you waiting on me for anything on this issue at this point?

cusspvz:
I just need to know where is the best place to write down the `intl object <-> string` handler so we can make it work on the frontend side without much hassle. My idea was to have this working with the active language at the time. So if we are editing a product in `en`, it should read and write into the `en`, and the same for others. But all the methods I've tried failed somehow. If you could give me some lights here it would be awesome!!

spencern:
Hey @cusspvz 

I think the eventual vision for us with this would be that for each language we could have a clone of a product, this clone could inherit all fields from the primary product, but any field that it defines (title/label/description/etc) would override the main product.

@aaronjudd has done a lot more thinking on this than I have, so he might have some thoughts on this as well

cusspvz:
@spencern me and @aaronjudd already discussed on the way to store the data at the database. We need intl not only on products but on categories as well, and maybe on other schemas in the future. So the backend itself already works, both the `title` and `description` returns an object with languages as keys and the values as the language value.

The only thing is that I don't know quite how to handle the value at the inputs when saving, because it is trying to save as a `String` and we have to handle it before to send the right object definition.

spencern:
I believe that we'd still want to save and validate the schema normally. On the server, you should be able to check the language of the user (`Meteor.user().profile.lang`). Check to see if that lang is different than the default language and then if it's different, store the description/title/etc for the language.

If I'm still misunderstanding the problem, perhaps it would be more helpful to use code examples to illustrate the challenge.

lib/collections/schemas/intlfield.js
zenweasel: Should be double-quotes everywhere

lib/collections/schemas/products.js
```diff
@@ -270,22 +271,24 @@ export const ProductVariant = new SimpleSchema({
     optional: true
   },
   taxDescription: {
-    type: String,
+    type: Intlfield,
```
zenweasel: This is more of an internal field, do we actually need this to be translated if customers never see it? Was your thinking to just change everything that's a `String` to `Intlfield`?

zenweasel:
Just for means of discussion, not that I am necessarily opposed to the approach but at least one problem I can see is that you are going to be pulling a **lot** of extra data that will in most cases not be displayed. Another approach may be to store just these fields in separate documents so that we could just fetch the appropriate language. Trading off a little db overhead with shipping a lot of extra data to the client (something we are trying to scale down). So more of a language overlay file.

This is sort of counter to our "one document" pattern we have followed so far so that's one downside as well.

spencern:
Closing because this is a work in progress issue.
@cusspvz feel free to re-open this once it's ready for a review.

Issue #2874, opened 2017-09-15
The main RC App container component should be fully customizable.


patrick-tavares:
While the current Components API allows customization of almost all of the UI, the highest level App container does not allow for customization. The page structure built from imports/plugins/core/router/client/app.js locks users into the default Reaction UI with the admin tool bar at the very top of the page and shortcuts bar on the right. While this is certainly a good default UI, customization should be available at the highest page structure level for users that want to completely customize the UX.

PR #2875, opened 2017-09-15
Patrick customizable app container

patrick-tavares:
This PR addresses the new functionality requested in issue #2874.

Ultimately the core <App> component building code was moved out from imports/plugins/core/router/client/app.js into a registerable component in imports/plugins/core/ui/clients/components/apps and imports/plugins/core/ui/clients/containers (though I am open to guidance if there is a better place for this to live). In addition, defaults.js is used to specify what the overriding AppContainer should be.

### DIFF
--- a/imports/plugins/core/ui/client/components/index.js
+++ b/imports/plugins/core/ui/client/components/index.js
@@ -1,6 +1,7 @@
 // export ButtonGroup from "./buttonGroup/buttonGroup";
 
 export { Alerts, Alert } from "./alerts";
+export { App } from "./app";
 export { default as Icon } from "./icon/icon";
 export { default as CircularProgress } from "./progress/circularProgress";
 export { default as Divider } from "./divider/divider";
--- a/imports/plugins/core/ui/client/containers/index.js
+++ b/imports/plugins/core/ui/client/containers/index.js
@@ -1,5 +1,6 @@
 export { default as EditContainer } from "./edit";
 export { default as Alerts } from "./alerts";
+export { default as App } from "./appContainer";
 export { default as ReactionAvatar } from "./avatar";
 export { default as SortableItem } from "./sortableItem";
 export { default as MediaGalleryContainer } from "./mediaGallery";
--- a/imports/plugins/core/router/client/browserRouter.js
+++ b/imports/plugins/core/router/client/browserRouter.js
@@ -9,10 +9,10 @@ import { isEqual } from "lodash";
 import queryParse from "query-parse";
 import { Session } from "meteor/session";
 import { Tracker } from "meteor/tracker";
-import App from "/imports/plugins/core/router/client/app";
 import { Router } from "../lib";
 import { MetaData } from "/lib/api/router/metadata";
 import { TranslationProvider } from "/imports/plugins/core/ui/client/providers";
+import { Components } from "@reactioncommerce/reaction-components";
 
 const history = Router.history;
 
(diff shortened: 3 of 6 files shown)


patrick-tavares:
So one thing I just noticed in going through another round of UI tests is that, while double-clicking a product grid item (while logged in as admin) properly displays the PDP page with Product Settings ActionView in FireFox, Safari and Chrome keep showing the Product Grid page with the Grid Settings ActionView. I could use another set of eyes here as the productGridItemsContainer.js code wasn't touched but seems to not like my changes.

client/config/defaults.js
```diff
@@ -7,6 +7,10 @@ import { AutoForm } from "meteor/aldeed:autoform";
  * A place to put misc. package configurations
  */
 
+// Set the default React app container inside which the rest of the UI will live
+// This enables the ability to completely override the look and feel of RC
+export const DEFAULT_APP_CONTAINER = "CoreReactionApp";
```
mikemurray: What's the need to set a session variable for the default app container?

Could you not just override it by using `replaceComponent("CoreReactionApp", MyCustomApp);`

patrick-tavares: I don't think I have a good response here. I'd gotten used to seeing registerComponent in digging through the code. Going the replaceComponent route simplifies things a bit so I'm onboard.

imports/plugins/core/ui/client/components/app/coreReactionApp.js
```diff
@@ -0,0 +1,106 @@
+import { Switch } from "react-router-dom";
```
mikemurray: Simply call is app.js and class name App. core and reaction are not necessary. Everywhere its called CoreReactionApp should just be App.

imports/plugins/core/router/client/app.js
```diff
@@ -1,116 +1,15 @@
...
+import { Session } from "meteor/session";
+import { getComponent } from "@reactioncommerce/reaction-components";
+
+const App = (props) => {
+  // Make sure a default value is set.
+  const defaultContainer = Session.get("DEFAULT_APP_CONTAINER") || "CoreReactionApp";
+  const AppContainer = getComponent(defaultContainer);
+
+  return (
+    <AppContainer {...props} />
```
mikemurray: This file doesn't need to exist anymore. In `browserRouter.js` file in the same folder, you could just do the following.

```
import { Components } from "@reactioncommerce/reaction-components";

<Components.App>
```

client/config/defaults.js
```diff
@@ -22,6 +26,7 @@ export const INDEX_OPTIONS = {
 // default load qty for product grid
 export const ITEMS_INCREMENT = 24;
 
+Session.setDefault("DEFAULT_APP_CONTAINER", DEFAULT_APP_CONTAINER);
```
mikemurray: Don't use session vars in `default.js` we will most likely move away from this solution in the future. And you can override components using `replaceComponent`

imports/plugins/core/ui/client/components/index.js
```diff
@@ -1,6 +1,7 @@
 // export ButtonGroup from "./buttonGroup/buttonGroup";
 
 export { Alerts, Alert } from "./alerts";
+export { CoreReactionApp } from "./app";
```
mikemurray: export as `App`

imports/plugins/core/ui/client/containers/index.js
```diff
@@ -1,5 +1,6 @@
 export { default as EditContainer } from "./edit";
 export { default as Alerts } from "./alerts";
+export { default as CoreReactionAppContainer } from "./coreReactionAppContainer";
```
mikemurray: We can just call it App, no need for the container in the name.

PR #2877, opened 2017-09-15
Implement babel-preset-env

aaronjudd:
- Remove es2015 preset
- Remove browerstack-local (not in use)
- Remove npm-shrinkwrap (not in use, deprecated)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7001 @@
+{
+  "name": "reaction",
+  "version": "1.5.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@braintree/wrap-promise": {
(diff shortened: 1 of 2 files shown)


aaronjudd:
Just a note on this, I tested removing only using `babel-preset-meteor`, only `babel-preset-env`, `env`+`preset:react`, without installing `babel-preset-react` and without `babel-preset-stage2`.. well, we need them all (preset + reset + stage2), so I opted for `env` over `meteor` as the base preset package.

spencern:
Any way to get bithound to rerun?

aaronjudd:
No idea what's going on there. Looks, their status page says all is good, but looks like our analysis' are just stuck in some queued land..  I'm good with this going, if you want to hit the button.

spencern:
I'll hit the button. I've had the bitHound link open for a while and it's still printing off nerdspirational messages.

Seems like perhaps the queue is borked or super backed up.

image

PR #2880, opened 2017-09-17
[WIP] Fix duplicate shipping methods in first checkout

### LINKED ISSUE #2866: Duplicated shipping options in first checkout
### Expected behavior
Shipping Options should be listed normally (once per options)

### Actual Behavior
Each shipping method is listed twice

1. `reaction reset`
1. Login as admin
1. Enable Flat Rate shipping methods
1. Add a product to cart and proceed to checkout
1. After entering address observe that checkout methods are listed twice

**Note** this only seems to happen the first time you checkout. All subsequent checkouts work fine. But if you open an incognito window it happens again.

### Versions
```
Node: 7.10.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace (I really was on marketplace this time)
Docker: 17.06.2-ce
```

foladipo:
Resolves #2866.
<!--Please provide the following information to help us approve your PR. Failure to provide this
information may result in your PR being closed without comment.

- [ ] Clearly explain what your PR is trying to accomplish. A link to an issue is best
->

### DIFF
--- a/imports/plugins/included/shipping-rates/server/hooks/hooks.js
+++ b/imports/plugins/included/shipping-rates/server/hooks/hooks.js
@@ -34,6 +34,23 @@ function getShippingRates(previousQueryResults, cart) {
       return previousQueryResults;
     }
   }
+  if (!(cart.shipping && cart.shipping[0] && cart.shipping[0].address)) {
+    const errorDetails = {
+      requestStatus: "error",
+      shippingProvider: "shippo",
+      message: "The 'shipping' property of this cart is either missing or incomplete."
+    };
+    rates.push(errorDetails);
+    return [rates, retrialTargets];
+  }
+  if (!(cart.items && cart.items[0] && cart.items[0].parcel)) {
+    const errorDetails = {
+      requestStatus: "error",
+      shippingProvider: "shippo",
+      message: "This cart has no items, or the first item has no 'parcel' property."
+    };
+    return [[errorDetails], []];
+  }
 
   let merchantShippingRates = false;
   const marketplaceSettings = Reaction.getMarketplaceSettings();


zenweasel:
This works and I am going to merge it but maybe you can help me understand the problem and why this fixes it better.

imports/plugins/included/shipping-rates/server/hooks/hooks.js
```diff
@@ -34,6 +34,23 @@ function getShippingRates(previousQueryResults, cart) {
       return previousQueryResults;
     }
   }
+  if (!(cart.shipping && cart.shipping[0] && cart.shipping[0].address)) {
```
spencern: One of our goals was to eliminate these hard-coded 0th indexes - I'm a bit concerned about introducing these here as it seems like adding technical debt immediately

foladipo:
The reason why this fixes the bugs is that, on the first ever checkout, `cart.shipping` will be undefined/empty. So, a shipping plugin should not attempt, in that case, to add shipping methods to the cart because, since `cart.shipping` is undefined/empty, there won't be `cart.shipping[0].address`, which is necessary for any shipping activity. The `reaction-shippo` plugin took care of this, which was why it wasn't duplicated on the first checkout. So, I add to handle this case for the flat rate shipping pluggin too.

Much of the same argument also applies to when `cart.items` is undefined or empty.

zenweasel:
Unfortunately I don't think that assumption is always correct. If a user has checked out and has selected a default address before it immediately attaches the address to the cart as soon as it is created.

foladipo:
Are you saying a user can have a saved address, and thus have a non-empty `cart.shipping` BEFORE they make their first ever purchase?

PR #2884, opened 2017-09-18
Fix Login dropdown not showing user account even when signed in

### LINKED ISSUE #2803: Login dropdown shows "Sign In" even when you are signed in
### Expected behavior
After you sign in you should see your username/profile

### Actual Behavior
Dropdown shows "Sign In" but if you refresh you are shown as logged in

I have seen this other places but this is how I was seeing it most consistently
1. Invite a shop owner
1. Open the invite email and copy the link
1. In an incognito session open the invite link
1. Supply a password
1. Observe that even though you are now signed in the dropdown says "Sign In"
1. Hard refresh
1. Observe that your profile info shows

not_signed_in

### Versions
```
Node: 7.10.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.0-ce
```

impactmass:
Resolves #2803.
Username and avatar should consistently be shown when a user is logged in, but this doesn't happen. Sometimes a user is logged in, and the dropdown still shows "Sign In".

## How To Test:
* Log In and confirm that the username and avatar is show in the account dropdown.
* Log out and log in again to confirm that the user details always get shown (Repeat this a few times).

The fix:
_ Check to ensure that Accounts is ready before doing Accounts.findOne

### DIFF
--- a/imports/plugins/core/components/lib/hoc.js
+++ b/imports/plugins/core/components/lib/hoc.js
@@ -40,14 +40,16 @@ export function withCurrentAccount(component) {
       return null;
     }
 
-    // shoppers should always be guests
-    const isGuest = Roles.userIsInRole(user, "guest", shopId);
-    // but if a user has never logged in then they are anonymous
-    const isAnonymous = Roles.userIsInRole(user, "anonymous", shopId);
+    const accSub = Meteor.subscribe("Accounts", user._id);
+    if (accSub.ready()) {
+      // shoppers should always be guests
+      const isGuest = Roles.userIsInRole(user, "guest", shopId);
+      // but if a user has never logged in then they are anonymous
+      const isAnonymous = Roles.userIsInRole(user, "anonymous", shopId);
+      const account = Accounts.findOne(user._id);
 
-    const account = Accounts.findOne(user._id);
-
-    onData(null, { currentAccount: isGuest && !isAnonymous && account });
+      onData(null, { currentAccount: isGuest && !isAnonymous && account });
+    }
   })(component);
 }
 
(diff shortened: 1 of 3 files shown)


jshimko:
While it's a little tough to consistently reproduce #2803, I can do it and this does seem to stop that bug from happening.  There are still an insane amount of re-renders in the dropdown component (3 on initial page load and about 15-20 after login/logout), but there are now less than before and the UI experience is no longer broken.  So... 👍

jshimko:
reaction

imports/plugins/core/components/lib/hoc.js
```diff
@@ -40,14 +40,17 @@ export function withCurrentAccount(component) {
...
-    // shoppers should always be guests
-    const isGuest = Roles.userIsInRole(user, "guest", shopId);
-    // but if a user has never logged in then they are anonymous
-    const isAnonymous = Roles.userIsInRole(user, "anonymous", shopId);
+    const accSub = Meteor.subscribe("Accounts", user._id);
+    if (accSub.ready()) {
+      // shoppers should always be guests
+      const isGuest = Roles.userIsInRole(user, "guest", shopId);
+      // but if a user has never logged in then they are anonymous
+      const isAnonymous = Roles.userIsInRole(user, "anonymous", shopId);
```
aaronjudd: Should be using `Reaction.hasPermissions`

jshimko: Yep @impactmass =, just knock those two role check instances out and we can merge this.

impactmass: @jshimko I've updated the use Roles.userIsInRole for the guest check. At the moment, we can not replace the anonymous check with hasPermission. I added comments describing why. That will be worked in #2895.

jshimko: Great, thanks!

impactmass:
This should be good to go cc @aaronjudd / @jshimko

Issue #1647, opened 2016-12-20
PDP UI Fixes


rymorgan:
1. Move public/private buttons for variants to sidebar
2. Move Toggle up/down indicator in panels to the middle. 
3. Notification for edited fields -- either a stroke/dot or some other indicator 
4. Look at cleaning up images - should the thumbnails be forced into squares. Do we need show images for variants. 
5. Move settings "cog" to the publish bar

aaronjudd:
<img width="1611" alt="basic_reaction_product" src="https://cloud.githubusercontent.com/assets/439959/21333959/87ab6f88-c608-11e6-839b-a36c09e93747.png">

rymorgan:
Here's some design work that I reviewed with @aaronjudd yesterday. There are a few elements I'm still working on but some of it can be implemented now as an iteration. 

1. Publish bar can go to top and buttons can change formats. 
2. Clean up of sidebar
3. Moving visible toggle to the sidebar variant
4. Making all sidebar section collapsible panels with arrow toggle switch
5. Drag/drop style change
6. Panel headers great for panel/white for nested panels. 

Still refining: 
1. Over/down states for editable fields if we move away from the circle buttons always being visible
2. Clean up of detail panel 
3. State for what has been edited 
4. Color of store buttons vs admin buttons
 
![pdp desktop](https://cloud.githubusercontent.com/assets/20409254/21461130/88a91176-c903-11e6-8fd3-608341452966.png)

mikemurray:
Moving of the toolbar and some design iteration:
https://github.com/reactioncommerce/reaction/tree/1624-move-pdp-toolbar

rymorgan:
Updated mock with cleaned up variant details panel. Let's discuss - There's some that we should start on now and some that needs a little more tweaking. 

Moving this to the top of the waffleboard. I think we should get these into a branch as soon as possible to facilitate the UI cleanup. 

![pdp ui updates](https://cloud.githubusercontent.com/assets/20409254/21736887/e3c600d4-d428-11e6-8d5f-4962b20878b4.png)

rymorgan:
Attached are some new revisions of the PDP. First one has notes. Whoever takes this, let's discuss. 
![sidebar_-_pdp_reorg_-_edit_variant_notes](https://cloud.githubusercontent.com/assets/20409254/23086639/13058afc-f523-11e6-854c-930fbdf5f333.png)
[...]

PR #2893, opened 2017-09-19
Update Variant Edit form

### LINKED ISSUE #2851: Convert variant options card on PDP to react.
Convert variant options card on PDP to react.

### LINKED ISSUE #2853: Inventory card opens at product details card on PDP
### Expected behavior
The card in the PDP that I am clicking on should only open the card I click. 

### Actual Behavior
When I click on the 'taxable' or 'inventory tracking' card it also open the 'product details' card

Navigate to the PDP
Close the product details card
Click on either taxable or inventory tracking card to open
See that the product details card also opens

### Versions
inventory card opens variant detaisl card

### LINKED ISSUE #2854: On the PDP the sidebar should open and show where which variant or label is required
It's a usability issue that when you publish you can't see which label is missing. When you try to publish without a label the sidebar should open and focus on the variant or variant option that is missing the label. 

If there are multiple labels missing, it should start at the top and each subsequent time you publish should open and focus on the field that is missing. 

We may also want to consider a red indicator dot on the actual PDP page in place of the info dot (for not published). That will give them two forms of feedback to guide them to correct the error. 

lable-validation and error

kieckhafer:
Various updates to the Admin panel of the PDP page to enhance editing of product variants.

- Update Variant Option section to use React form (the same form top-level variants use, with less fields displayed) ( Resolves #2851 , fixes portions of #1647 )
- Add `Compare at price` (MSRP) UI to variant options
- Change language on some field titles to match functionality a little better
- Add `helpText` to all form fields in the variant form
- Introduce `helpMode` to the aforementioned help text. As of now, nothing changes with how anything is displayed, but laying the foundation for globally toggled help text in the future. (https://github.com/reactioncommerce/reaction/pull/2893/files#diff-bee363c8fbac612bc278e9c9ced881c3R33)
- Better error visibility to help with hard to find errors
- Fix errors where clicking edit buttons / panels open and close various other panels at (seemingly) random ( Resolves #2853 , resolves #2854 )

basic_reaction_product


kieckhafer:
Seems to be an infinite loop of state resetting somewhere in this PR. I am trying to work through it at the moment, but if we can get some more eyes on this it could be helpful.

kieckhafer:
This is ready to be tested, however I have been able to unreliably, unintentionally, reproduce a strange issue, and I'm hoping someone else can take a look and (possibly) reproduce while testing, and maybe have any ideas as to what's going on, because I'm at a point where I don't really know anymore.

What seems to happen is the `actionView` will get caught in some sort of infinite loop of reloading, only when the Action View panel is already open, one of the variants inside the already opened actionView panel has a validation error, and then you make a file change that reloads meteor.

This seems to be the only time I can reproduce the issue, and even with that specificity, I can't seem to make it happen every time.

spencern:
I like this - preventing the user from trying to save an invalid state.
http://g.recordit.co/ODO3KKmeuX.gif

spencern:
The price fields are truncating the `.00` from the price if it's a whole number. I think we'd want to keep that as it's how prices are displayed throughout the app.

http://g.recordit.co/vX059ic1pq.gif

spencern:
There's an issue where a server error is thrown when trying to publish an invalid option without a label, but the action view does not open, and the action view does not focus on the invalid product/option.

This video is a little bigger, so click through to see http://recordit.co/4oRFSOtFD1

spencern:
Another "issue" that may just be me trying to use it "wrong" is that I'd expect that by clicking the red dot, that I'd be taken to the exception for that option/variant, but clicking it does nothing. I'd consider this to be a UIX issue that should be fixed, even if it currently works as intended.

spencern:
I see an error toast "price is required" when I tab away from "price" for a new option that has a label. It doesn't invalidate (turn red) the option header or indicate that the price field is in an invalid state though. Considering how well the UI works to notify you that a label is invalid, I'd expect similar behavior for other fields that are required.

The red dot does stay, indicating that there is some awareness the option is in an invalid state.






video: http://recordit.co/3CLtfAGaQ1

spencern:
The help text is very nice to have and I think it will help resolve a lot of issues. 👍 
image

spencern:
This might not be the exact issue to fix this on, but we I think we need to add most fields that exist on variant to the options as well. 

Weight, Dimensions, Warn At, etc should be configurable at whatever the bottom level is, IMO.

image

spencern:
Could we consider having a different icon for a variant/option/(maybe product?) that hasn't been published? Additionally, it might be worth considering not showing labels such as `backordered` until the product has been published at least once.

image

imports/plugins/included/product-variant/components/variantForm.js
```diff
@@ -53,21 +54,26 @@ class VariantForm extends Component {
 
   componentWillReceiveProps(nextProps) {
     const nextVariant = nextProps.variant || {};
-    const currentVariant = this.props.variant || {};
+    const currentVariant = this.state.variant || {};
 
-    if (!isEqual(nextVariant, currentVariant)) {
+    if (_.isEqual(nextVariant, currentVariant) === false) {
```
spencern: There's not really a good `isEqual` replacement in prototype or ES6 at this point, so I'm okay continuing to use lodash in this circumstance

private/data/i18n/en.json
```diff
@@ -86,7 +86,51 @@
         "pinProduct": "Pin Product",
         "toggleSize": "Toggle Size",
         "publish": "Publish",
-        "moreOptions": "More Options"
+        "moreOptions": "More Options",
```
spencern: Some case disparities in this file.

private/data/i18n/en.json
```diff
@@ -86,7 +86,51 @@
...
+        "details": "Details",
+        "addVariantFail": "Unable to create variant for {{title}}.",
+        "cloneVariantFail": "Unable to clone variant for {{title}}.",
+        "removeVariantConfirm": "Are you sure you want to delete {{title}}",
+        "restoreVariantConfirm": "Are you sure you want to restore {{title}} from the archive?",
+        "restoreVariantFail": "Unable to restore {{title}}.",
+        "archiveVariantConfirm": "Are you sure you want to archive {{title}}",
+        "thisVariant": "this variant",
+        "thisOption": "this option",
+        "archiveProduct": "Archive Product",
```
spencern: casing

private/data/i18n/en.json
```diff
@@ -86,7 +86,51 @@
...
+        "restoreProduct": "Restore Product from archive",
+        "archiveThisProduct": "Archive this product?",
+        "archiveSelectedProducts": "Archive selected products?",
+        "noSelectedProducts": "No selected products. Use the ",
+        "orShiftCtrlCmd": " or shift|ctrl|cmd + click on products to select multiple. With shift you can select a contiguous range of products.",
+        "makeVisible": "Make Visible",
```
spencern: casing

private/data/i18n/en.json
```diff
@@ -86,7 +86,51 @@
...
+        "optionTitle": "Option title",
+        "barcode": "Barcode",
+        "compareAtPrice": "Compare at price",
+        "fulfillmentService": "Fulfillment service",
+        "weight": "Weight",
+        "length": "Length",
+        "width": "Width",
+        "height": "Height",
+        "taxable": "Taxable",
+        "taxCode": "Tax Code",
```
spencern: casing here and below

private/data/i18n/en.json
```diff
@@ -86,7 +86,51 @@
...
+        "taxDescription": "Tax Description",
+        "inventoryManagement": "Inventory Tracking",
+        "inventoryManagementLabel": "Register product in inventory system and track its status",
+        "inventoryPolicy": "Allow backorder",
+        "inventoryPolicyLabel": "Do not sell the product variant when it is not in stock",
+        "lowInventoryWarningThreshold": "Warn At",
+        "lowInventoryWarningThresholdLabel": "Warn customers that the product will be close to sold out when the quantity reaches the next threshold",
+        "originCountry": "Origin Country",
+        "selectTaxCode": "Select Tax Code"
```
spencern: more casing

imports/plugins/included/product-variant/server/i18n/en.json
```diff
@@ -3,11 +3,24 @@
   "ns": "reaction-product-variant",
   "translation": {
     "reaction-product-variant": {
+      "admin": {
+        "helpText": {
+          "compareAtPrice": "Original price or MSRP",
+          "optionInventoryQuantity": "Inventory",
+          "optionTitle": "Displayed on Product Detail Page",
```
spencern: Is the Product Detail Page a proper noun in Reaction? Could probably use some docs on what we consider to be proper nouns that should be upper cased @machikoyasuda

kieckhafer:
```
The price fields are truncating the .00 from the price if it's a whole number. I think we'd want to keep that as it's how prices are displayed throughout the app.
```

Not saying this isn't an "issue" / something that can be fixed, but this is not new, it happens in the current `marketplace` branch as well.

```
There's an issue where a server error is thrown when trying to publish an invalid option without a label, but the action view does not open, and the action view does not focus on the invalid product/option.

I think we should probably not be calling server methods with data we know is invalid, this should get intercepted before it gets to that point.

I also think we should be opening the action view to the first product/variant/option that is invalid if a user attempts to publish an invalid product
```

Makes sense to not call the server if the data is invalid, I will look into working around that. As for opening the sidebar if something is invalid, I purposely did not do that based on not wanting to open to a certain field if there were multiple fields that were invalid. No real reason other than I didn't like the flow that it opened to this seemingly random field when multiple things are there. I just didn't know which to focus on, as "first" in a vertical list looks different than "first" in the horizontal buttons on the actual PDP, not as easy to correlate as clicking on the actual variant.

```
Another "issue" that may just be me trying to use it "wrong" is that I'd expect that by clicking the red dot, that I'd be taken to the exception for that option/variant, but clicking it does nothing. I'd consider this to be a UIX issue that should be fixed, even if it currently works as intended.
```
Clicking will do nothing, it's "working" as expected, however I like your idea, it should open to the variant, I will get that updated.

```
I see an error toast "price is required" when I tab away from "price" for a new option that has a label. It doesn't invalidate (turn red) the option header or indicate that the price field is in an invalid state though. Considering how well the UI works to notify you that a label is invalid, I'd expect similar behavior for other fields that are required.
```
Weird. Will look into.

```
This might not be the exact issue to fix this on, but we I think we need to add most fields that exist on variant to the options as well.

Weight, Dimensions, Warn At, etc should be configurable at whatever the bottom level is, IMO.
```
The variants and options both use the same form now (with different fields visible / not visible), so this will be very easy to do. Don't think it's a blocker for this ticket if we don't get it done, but we can make a new ticket to discuss which fields should be there and easily update.

```
Could we consider having a different icon for a variant/option/(maybe product?) that hasn't been published? E.g. this product is "invalid", but also possible that I'm still working on it, or want to not save it. Additionally, it might be worth considering not showing labels such as backordered until the product has been published at least once.

Should products/variants/options that never get published get actually deleted, or still archived?
```
@rymorgan what do you think about another icon for never published products? I think it might get a little confusing having that many different icons, but it wouldn't be too hard to get used to if they are all there.

spencern:
> The variants and options both use the same form now (with different fields visible / not visible), so this will be very easy to do. Don't think it's a blocker for this ticket if we don't get it done, but we can make a new ticket to discuss which fields should be there and easily update.

This is great to hear! I think this will make iterating on this a lot faster/easier

spencern:
> No real reason other than I didn't like the flow that it opened to this seemingly random field when multiple things are there. I just didn't know which to focus on, as "first" in a vertical list looks different than "first" in the horizontal buttons on the actual PDP, not as easy to correlate as clicking on the actual variant.

Good argument. Could we at least force the action view open for the "first" variant? Or is that too much too?

spencern:
> @rymorgan what do you think about another icon for never published products? I think it might get a little confusing having that many different icons, but it wouldn't be too hard to get used to if they are all there.

This part is definitely not a blocker for me by the way.

kieckhafer:
```
Good argument. Could we at least force the action view open for the "first" variant? Or is that too much too?
```

Yeah I think we could open it. I'll look into that.

imports/plugins/included/product-variant/server/i18n/en.json
kieckhafer: If it's not officially... I think it should be.

kieckhafer:
@spencern @rymorgan did we decide sentence casing everywhere? Or sentence casing in most of the app, with title casing on headers / card headers?

If the latter is the case, I think we need to see these translations in context before making any changes, perhaps a Hack-a-Translation-a-thon for a few hours next time we're all in the same room?

spencern:
deferring to @rymorgan on this one. Title casing on headers makes sense to me

rymorgan:
**Casing** 
Yes, sentence casing everywhere. 

**Unpublished icon**
I think we have too many icons now and are moving towards a world where we have fewer icons and more editing happening in the action view. I might rather use our dot indicator as a way to show a variant or option as unpublished vs an icon. The dot indicator now is the one that reflects the publishing status so it makes logical sense to use that. I know this is a change, but would it make sense to use the info color for never been published, warning color for changes that need be published and red for error? 

**More fields**
As far as the fields. This UI that we've made huge improvements to gets more confusing/overwhelming if we add the height/weight to each variant option. I know this is useful for saving money on shipping but I think we can be smarter about the way we add it. Can we add a preference or checkmark farther up the tree that says "Apply to all variant options"? As I think this is the 80% or more use case. If you have that unchecked then maybe there's some sort of expandable UI element that allows you to add this per variant option. Otherwise, our UI is going to feel harder to use than is necessary. This, of course, can be another issue. 

**Clicking on red dots**
Love this idea. 

**Not showing back order until it's been published**
Love this idea too @kieckhafer, is that doable?

kieckhafer:
`The price fields are truncating the .00 from the price if it's a whole number. I think we'd want to keep that as it's how prices are displayed throughout the app.`

This issue seems to go deeper than just the form fields, it looks like we truncate the data somewhere along the way and actually save it without the `.00`'s in the database. Created a new ticket - #2964 - to look into.

kieckhafer:
```
Not showing back order until it's been published
Love this idea too @kieckhafer, is that doable?
```
Think it might be a little much to do for this PR, but in the future, yes.

spencern:
@kieckhafer is this ready for another review?

kieckhafer:
@spencern still working on the price is required toast, but I think everything else is ready to go

spencern:
Fix for the Title bug is now detailed in #2985

PR #2898, opened 2017-09-19
Update email for inviting a user to create a shop

### LINKED ISSUE #2804: Update email copy for inviting a user to create a shop

impactmass:
Resolves #2804. 
_ Updates email copy for new shop owner and uses the primary shop details (name, address etc) when sending email to new owner (instead of the newly created shop details as before).
_ Also updates the email copy sent to a user when added to a group (see ticket for the copy #2804)

**How To Test:**
- Invite a new owner/merchant under Shop > Marketplace in dashboard
- Confirm updated copy in the email sent to the merchant
- Invite a new user to a group in Accounts > Add Admin User in dashboard in Primary Shop
- Confirm updated copy in the email sent to the account
- Invite a new user to a group like above but in a merchant shop
- Confirm that email copy sent is also updated (and it reflects the right shop name)

### DIFF
--- a/.meteor/release
+++ b/.meteor/release
@@ -1,1 +1,1 @@
-METEOR@1.5.2
+METEOR@1.5.1
--- a/client/modules/core/main.js
+++ b/client/modules/core/main.js
@@ -47,46 +47,42 @@ export default {
     Tracker.autorun(() => {
       let shop;
       if (this.Subscriptions.PrimaryShop.ready()) {
-        // if we've already set the primaryShopId, carry on.
-        // otherwise we need to define it.
-        if (!this.primaryShopId) {
-          // There should only ever be one "primary" shop
-          shop = Shops.findOne({
-            shopType: "primary"
-          });
+        // There should only ever be one "primary" shop
+        shop = Shops.findOne({
+          shopType: "primary"
+        });
 
-          if (shop) {
-            this.primaryShopId = shop._id;
-            this.primaryShopName = shop.name;
+        if (shop) {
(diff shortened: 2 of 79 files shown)


kieckhafer:
This looks good overall, everything works as it should. My one comment is I think we should update the copy to read something like:

{{currentUserName}} has invited you to join *the group* {{groupName}} in *the store* {{shopName}}

And update the subject to read something similar. What we have now doesn't flow very well or explain what `Shop Manager` or `Reaction1` are.

airmail

rymorgan:
Agree with @kieckhafer

impactmass:
Ok. I'll make that update shortly

impactmass:
Updated based on the comment. cc @kieckhafer

kieckhafer:
@impactmass I made one small change that just removes the quotes from surrounding the shop name, but aside from that this is good! Going to merge as soon as the tests pass.

Issue #2766, opened 2017-08-30
Stripe List Refunds should work with Stripe Connect charges


spencern:
The Stripe Connect implementation is currently incomplete.

The [stripe/payment/createCharges](https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/included/payments-stripe/server/methods/stripe.js#L169) method is capable of creating direct charges on behalf of another stripe account, but [our stripe list refunds method]https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/included/payments-stripe/server/methods/stripe.js#L422) does not currently support listing refunds of direct charges.

This ticket is for listing refunds associated with a given transaction id, this method does _not_ currently work with Stripe Connect direct charges.

A generic method that depends on this has also been disabled in the email template pending this getting updated: https://github.com/reactioncommerce/reaction/blob/marketplace/server/methods/core/orders.js#L503 so that should be added back as part of this ticket as well.

My best guess as to where to start would be that we need to pass the merchant `stripe_account` into the list refunds call. 

I think this is a good time to update the existing Stripe list refunds method to use the node-stripe lib as well and you should be able to follow the patterns set in the `stripe/payment/createCharges` method to figure that part out.

**Resources and other notes**
* https://stripe.com/docs/api#list_refunds - Stripe List Refunds docs
* Here's the Reaction stripe-payments code where we actually hit the stripe API to list refunds: https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/included/payments-stripe/server/methods/stripeapi.js#L142

spencern:
Resolved by #2912

PR #2912, opened 2017-09-21
Make approve/capture/refund/listRefunds multishop-aware

### LINKED ISSUE #2767: Stripe Create Refund for Stripe Connect charges
The Stripe Connect implementation is currently incomplete.

The [stripe/payment/createCharges](https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/included/payments-stripe/server/methods/stripe.js#L169) method is capable of creating direct charges on behalf of another stripe account, but [our stripe create refund method](https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/included/payments-stripe/server/methods/stripe.js#L378) does not currently support creating a refund of direct charges.

This ticket is for creating a refund for a stripe charge, this method does _not_ currently work with Stripe Connect direct charges.

My best guess as to where to start would be that we need to pass the merchant `stripe_account` into the create refund call. 

I think this is a good time to update the existing Stripe create refund method to use the node-stripe lib as well and you should be able to follow the patterns set in the `stripe/payment/createCharges` method to figure that part out.

**Resources and other notes**
* https://stripe.com/docs/api#create_refund - Stripe Create Refund docs
* Here's the Reaction stripe-payments code where we actually hit the stripe API to create a refund: https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/included/payments-stripe/server/methods/stripeapi.js#L122

zenweasel:
Close #2767, #2766, #2765

### To Test

1. Full reset
1. Create a shop and a marketplace product and enable Stripe Connect in the shop
1. Create an order with Basic Reaction Product and your created product
1. As the marketplace admin Approve and then Capture the order 
1. In another browser as the merchant observe the order and approve and capture the order
1. Observe that these captures are done separately
1. Issue a refund for this order in the merchant and marketplace admins respectively
1. Observe the refund is issued against the correct charge and is reflected in the dashboard correctly.


spencern:
@zenweasel I know you were pushing to get this ready before you finished up last week, but as it's still [WIP] and no reviewers have been assigned, I'd imagine it's not quite done yet?

zenweasel:
@spencern Yes. The methods are working but tests are still broken. Should have this ready for review today

zenweasel:
Got all tests working but when I pulled in changes all the order filtering/dashboard is broken. Trying to figure out what is going on

zenweasel:
Still working on fixing up orders but hope to have this ready tomorrow morning some time

zenweasel:
Ok @spencern this should be ready to look at now.

spencern:
Getting a shipping methods are not configured error when trying to checkout right now. I'll debug that a bit and follow up shortly.

image

imports/plugins/included/payments-stripe/server/methods/stripe.js
```diff
@@ -28,17 +25,33 @@ function unformatFromStripe(amount) {
   return (amount / 100);
 }
 
+export const utils = {};
+
+utils.getStripeApi = function (paymentPackageId) {
```
spencern: Thanks for abstracting this. Needed to happen

imports/plugins/included/payments-stripe/server/methods/stripe.js
```diff
@@ -28,17 +25,33 @@ function unformatFromStripe(amount) {
...
+
+  const stripeKey = utils.getStripeApi(paymentMethod.paymentPackageId);
+  const stripe = stripeNpm(stripeKey);
+
   try {
-    const captureResult = StripeApi.methods.captureCharge.call({
-      transactionId: paymentMethod.transactionId,
-      captureDetails: captureDetails
-    });
+    const capturePromise = stripe.charges.capture(paymentMethod.transactionId, captureDetails);
```
spencern: Do you think we should change my current `createCharges` method to use Promises instead of async/await? Feels weird to me to do this in different ways within the same plugin.

imports/plugins/included/search-mongo/server/methods/searchcollections.js
```diff
@@ -185,7 +185,10 @@ export function buildOrderSearchRecord(orderId) {
   orderSearch.billingName = shopBilling.address && shopBilling.address.fullName;
   orderSearch.billingPhone = shopBilling.address && shopBilling.address.phone.replace(/\D/g, "");
   orderSearch.shippingName = shopShipping.address && shopShipping.address.fullName;
-  orderSearch.shippingPhone = shopShipping.address && shopShipping.address.phone.replace(/\D/g, "");
+  if (shopShipping.address && shopShipping.address.phone) {
```
spencern: Did this sneak in from a different PR? Or just fixing an undocumented issue that you came across?

server/methods/core/orders.js
```diff
@@ -64,6 +64,43 @@ export function ordersInventoryAdjust(orderId) {
   });
 }
 
+// TODO: Marketplace: Is there a reason to do this any other way? Can admins reduce for more
+// than one shop
+/**
+ * ordersInventoryAdjustByShop
+ * adjust inventory for a particular shop when an order is approved
```
spencern: As part of removing the "approval" step, I think that inventory should be reduced when an order is placed (after copyCartToOrder) so that we avoid the possibility of multiple orders for the same inventory.

spencern:
Successfully captured a payment from a marketplace shop
image

Successfully issued partial refund!
image

zenweasel:
@spencern Have you pulled recently? I fixed that Icon issue on another PR (it was unrelated to this one and just a bad import) and it should have been merged in.

I have opened an issue for the problems with viewing orders as a merchant admin here: https://github.com/reactioncommerce/reaction/issues/2976

server/methods/core/orders.js
zenweasel: I think removing Approval should probably be separate from this issue.

imports/plugins/included/search-mongo/server/methods/searchcollections.js
zenweasel: I was running into this throwing a `can't find replace of undefined` when trying to place an order. Probably just something that hadn't come up in testing before? Sort of just a workaround but required to keep moving.

zenweasel:
@spencern Yeah, these other (I believe unrelated) issues are why this took me longer than expected (to fix some and document others). But I think this PR does what it's supposed to without introducing any new errors.

zenweasel:
Refunds are not stored anywhere for any payment methods. That's essentially why we have the `listRefunds` methods

spencern:
Regarding the transaction being wrapped in a "response" object. It appears that other payment providers do this sometimes, but it's the raw response from the payment provider. Since each provider returns different data in different formats, I'm going to say this is not a blocker and we'll merge and reconcile later if necessary.

PR #2914, opened 2017-09-21
Order filtering shipping status

### LINKED ISSUE #2906: Order filtering by shipping status
Users with permissions to view orders dashboard should be able to quickly filter the list of orders by the shipping status.

From a dropdown, users should be able to filter orders by shipping status. 
Shipping statuses:
- Picked
- Packed
- Labeled
- Shipped
- Delivered

If users search while a filter is active, the search should take the filter into account.

Users should be able to clear all filters easily as well (#2358)

joykare:
Resolves #2906 

### How to test
- Make a couple of orders 
- Mark different shipping statuses for these orders
- Observe that orders are filtered correctly by shipping status
- Test that the shipping status filter works in tandem with the other filters and the order search


rymorgan:
When I filter by a shipping status, I get a blank row that shows up. 

blank-row

rymorgan:
### Expected behavior

I should be able to add two filters and unselect one filter and see the results show up again. 

### Actual Behavior

When I filter by status, then add a shipping status filter it narrows the results as expected. But when I unfilter status 

Place orders some orders.
Partially complete orders to get various statuses and shipping statuses.
Filter by status.
Filter by shipping status.
Click filter icon to un-filter by status. 
See that results don't un-filter and show in results. 

filter-unfilter

joykare:
@rymorgan  on the extra row issue on #2952 https://github.com/reactioncommerce/reaction/issues/2952#issuecomment-332236691

So, do you want to have it not be the default on this particular table or is having three rows minimum okay

rymorgan:
@joykare To be clear the minimum of 3 rows is for when there are no results. What I pointed out is a different issue. It looks broken when there are extra rows when you have results. Seems this may be a larger bug that affects all our tables?

joykare:
@rymorgan I've seen what the issue is, but since its a bug on marketplace I'll have it as a separate issue from this. See https://github.com/reactioncommerce/reaction/issues/2952#issuecomment-332256781 

Otherwise, I fixed the other issue you had, you can test it out.

joykare:
@kieckhafer 
- The shipping states, right now are #2906 (Right now there isn't a way to set an order to `delivered` so I left that one out) 
- The filter icon is meant to clear filters.

kieckhafer:
I don't think that's the best UI.

@rymorgan should make the final call, but I'd suggest having the filter icon and the arrows show the dropdown, and adding a new `x` or `clear` button that shows only when the filter is in use.

joykare:
@rymorgan @kieckhafer since the UI works as the designs right now prescribe, I don't think it should be a blocker for this PR. We could add in issues to change the designs of the filters if need be.

@zenweasel review on this.

rymorgan:
@joykare Yes, I agree. I'll make a separate issue for the refinement of the UI.

zenweasel:
These labels are probably not supposed to be in all lowercase?

cart_checkout

zenweasel:
Mostly this works well. Just need to fix the status headers and one tiny typo. There are other issues with the order dashboard but I will create separate issues for those.

joykare:
@zenweasel all fixed. @kieckhafer any additional thoughts...

zenweasel:
Verified issues fixed

PR #2929, opened 2017-09-22
Marketplace routing

### LINKED ISSUE #2760: [Marketplace] Routing issues with multiple shops

spencern:
Resolves #2760 
Paired with @mikemurray to create this branch

* Adds `slug` to Shops schema
* New migration to add slug to Shops schema
* Automatically creates a `marketplace` route for each route defined in the schema. Marketplace routes are of the format `/shop/:shopId/route`
  * Route resolver is pretty dumb right now, we check to see if there is a `shopId` param, and if so, map to the `/shop/:shopId` version of the route. Otherwise we map to the standard route.
* Updates the Product Grid and Product subscriptions to use the `shopId` to filter if present.
* Removes old `/shop/:shopId` route in favor of dynamically created version based on the `index` route.
* Adds new list of `merchantShops` to the primary shop. This array contains the shopId, slug, and name of the shop. This permits building urls for merchant shops without having to publish all shops to the client.


spencern:
@jshimko I've added you as the reviewer here since both @mikemurray and I had a hand in creating this, would be good to have an additional set of eyes on it. It's definitely not the final place we need to take the router, but it should be a good fix for marketplace to get this 1.5.0 release out

spencern:
@mikemurray will you throw a 👍 on here so I can get this merged?

PR #2930, opened 2017-09-22
Update `provides` in the registry accept an array of strings

### LINKED ISSUE #2897: Registry Provides should support an array of options
**Summary**: The `provides` property in the registry should support an array options.

Currently the `provides` option permits only a single string as the value. This string determines which Reaction "App" each particular registry entry is used for (see: https://github.com/reactioncommerce/reaction/blob/marketplace/client/modules/core/helpers/apps.js#L9-L44). If `provides` permitted using an array of strings as the value, we could have a single registry entry be used for to multiple reaction apps.

In order to maintain backwards compatibility, provides should support both an array of strings or a string value, potentially deprecating the string value in a future version. This would be a breaking change for any existing plugins, so it should be done only with a major version bump. Resolving this breaking change in any existing plugins would be somewhat simple, as plugin authors would merely need to update their registry to provide an array and update anywhere they reference the provides value in registry entries, if any exist.

spencern:
This PR resolves #2897 by enabling the use of multiple `provides` values in an array, making registry entries such as `provides: ["paymentSettings", "marketplacePaymentSettings"]` valid. 

Going forward we will use a method based on this to determine if a given plugin is for ready for marketplace. This PR does not contain any functionality to identify marketplace ready packages itself, but it does provide the structure for that.

It _does not_ accept both a String and an Array as was suggested in the issue. That would have required using a `blackbox` type in Simple Schema which felt like the wrong direction. Instead a migration is provided which will upgrade any packages that currently exist on the system from using (using "dashboard" for example) `provides: "dashboard"` to `provides: ["dashboard"]`.

The update to Reaction Apps which filters packages to determine if a given plugin registry entry is a match has been updated to  check against the simple schema type of a registry property. If the schema property type and the existing registry property are both `array`s then we will check to see if the filter matches an element in the array rather than checking to see if there is an exact match.

This maintains backwards compatibility to any package that manages to avoid getting upgraded.


aaronjudd:
@reactioncommerce/core any idea why `imports/plugins/included/discount-rates/disabled.register.js` is disabled?

aaronjudd:
Ah, nevermind... that's the "custom rates", not "discount-codes"... and I did that.. because we don't yet have "custom rates"....  (which is pricing)

aaronjudd:
Are we expecting that there are still some packages running deprecated tags? (warnings working great!).:

```20:24:48.378Z  WARN Reaction: Plugin reaction-accounts is using a deprecated version of the provides property for the account/verify registry entry. Since v1.5.0 registry provides accepts an array of strings.
20:24:48.463Z  WARN Reaction: Plugin reaction-checkout is using a deprecated version of the provides property for the cart/checkout registry entry. Since v1.5.0 registry provides accepts an array of strings.
20:24:48.463Z  WARN Reaction: Plugin reaction-checkout is using a deprecated version of the provides property for the cart/completed registry entry. Since v1.5.0 registry provides accepts an array of strings.
20:24:48.729Z  WARN Reaction: Plugin reaction-orders is using a deprecated version of the provides property for the dashboard/pdf/orders registry entry. Since v1.5.0 registry provides accepts an array of strings.
20:24:49.001Z  WARN Reaction: Plugin reaction-ui is using a deprecated version of the provides property for the dashboard/uiThemeDetails registry entry. Since v1.5.0 registry provides accepts an array of strings.
20:24:49.336Z  WARN Reaction: Plugin reaction-marketplace is using a deprecated version of the provides property for the shop registry entry. Since v1.5.0 registry provides accepts an array of strings.
20:24:49.369Z  WARN Reaction: Plugin reaction-notification is using a deprecated version of the provides property for the notifications registry entry. Since v1.5.0 registry provides accepts an array of strings.
20:24:49.477Z  WARN Reaction: Plugin reaction-paypal is using a deprecated version of the provides property for the /paypal/done registry entry. Since v1.5.0 registry provides accepts an array of strings.
20:24:49.477Z  WARN Reaction: Plugin reaction-paypal is using a deprecated version of the provides property for the /paypal/cancel registry entry. Since v1.5.0 registry provides accepts an array of strings.
20:24:49.522Z  WARN Reaction: Plugin reaction-stripe is using a deprecated version of the provides property for the /stripe/connect/authorize registry entry. Since v1.5.0 registry provides accepts an array of strings.
20:24:49.620Z  WARN Reaction: Plugin product-detail-simple is using a deprecated version of the provides property for the product registry entry. Since v1.5.0 registry provides accepts an array of strings.
20:24:49.672Z  WARN Reaction: Plugin reaction-product-variant is using a deprecated version of the provides property for the tag registry entry. Since v1.5.0 registry provides accepts an array of strings.```

spencern:
@aaronjudd interesting - do you have any packages in your `custom` folder?

spencern:
I think this may have be an issue where the warning is triggering on any registry entry that does not have a  `provides` definition. Building a fix now

spencern:
@aaronjudd Good catch on that review, looks like I was catching instances where `provides` was undefined in addition to instances where provides was a string. Fixed now.

spencern:
Interesting thought on the warnings being a one-time thing. They are only triggered when we find a registry entry with a string value for `provides` Do we have a mechanism we could use to persist an error in this case? 

I see the error on the UI, and that does seem like it could be triggered by a `provides` issue. I'll try your reproduction steps and see what I can fix.

spencern:
@aaronjudd I've updated the message to include 
> Please update any plugins you maintain to use an array for provides. You will not see this message again for these plugins.

I've also identified and fixed the issue that was causing the dashboard panel to render in the small view. There are a few other spots in our code that were checking for `provides` equality. I've updated all of those instances to use prototype `includes` and to guard against `provides` being undefined.

Array.prototype.includes has the side-affect of checking for string inclusion as well as array inclusion. I think in this case this is a positive thing, though there are instances where it could be detrimental.

`"dashboard".includes("dashboard")` => `true`
`["dashboard"].includes("dashboard")` => `true`

the downside here would be any cases where something like this could occur:

`"marketplace-dashboard".includes("dashboard")` => `true`

We don't currently have any instances where this would be possible and since we're forcibly moving to arrays, I think it's okay, but we could do an equality check on string types, and use `includes` only on Arrays if there is concern.

Issue #2812, opened 2017-09-08
User profile name changes when adding a billing address with a different name. 


rymorgan:
### Expected behavior
User profile name should stay the same. 

### Actual Behavior
When you change a user's billing address in checkout, it also changes the user profile name. 

### Steps to Reproduce the Behavior
Create an address with different billing name
See that the user profile name also changes. 

![user profile bug](https://user-images.githubusercontent.com/20409254/30227395-693bc71e-948e-11e7-890c-19c079c63998.gif)

zenweasel:
@machikoyasuda Is there a status on this? Was this resolved by #2932? If you aren't going to be working on it could you unassign yourself?

machikoyasuda:
Unassigning myself

PR #2932, opened 2017-09-22
Username only updated on first address and when there is 1 address

Akarshit:
### Fix for #2812 
Username not updated everytime the billing address is changed. It is only updated when
* This is the first billing address the user is entering.
* The user is editing the only billing address.
### To test the PR
1. Sign-up as a new user
2. Checkout a product, the username will be changed from `Guest` to the name you use in `Billing Address`
1. Create an address with different billing name, the profile name will not change.
2. Delete the second address
1. Edit the billing address and change name, the profile name will change.


server/methods/accounts/accounts.js
```diff
@@ -356,24 +359,30 @@ export function addressBookAdd(address, accountUserId) {
...
-      name: address.fullName,
       userId: userId
     },
     $addToSet: {
       "profile.addressBook": address
     }
-  });
+  };
+
+  if (!account.name || _.get(account, 'profile.addressBook.length', 0) === 0) {
```
mikemurray: Double quotes on strings.

server/methods/accounts/accounts.js
```diff
@@ -469,22 +478,29 @@ export function addressBookUpdate(address, accountUserId, type) {
...
-  }, {
+  const accountsUpdateQuery = {
     $set: {
-      "name": address.fullName,
       "profile.addressBook.$": address
     }
-  });
+  };
+  // update the name when there is no name or the user updated his only shipping address
+  if (!account.name || _.get(account, 'profile.addressBook.length', 0) <= 1) {
```
mikemurray: Double quotes.

spencern:
@mikemurray can you fix those quote issues in this PR so we can merge?

PR #2936, opened 2017-09-25
Added Units of Length

lcampanis:
Just like Units of Measure we are now adding units of length `uol` and `baseUOL`

We have also changed all units to be lowercase by default and changed grams to be`g` instead of `gr` as per official documentation: http://abacus.bates.edu/~ganderso/biology/resources/writing/HTWabbr.html

If accepted, this change will require a migration by RC to change all units across all shops and user profiles.

Credit @stevaras


spencern:
@stevaras @lcampanis This seems like a sensible PR, can one of you add a migration, or were you hoping to get sign-off that this is a direction we're hoping to go first?

lcampanis:
@spencern migration is now up.

imports/plugins/included/shippo/server/methods/shippo.js
```diff
@@ -409,10 +409,9 @@ export const methods = {
       const shippoAddressFrom = createShippoAddress(shop.addressBook[0], shop.emails[0].address, purpose);
       // product in the cart has to have parcel property with the dimensions
       if (cart.items && cart.items[0] && cart.items[0].parcel) {
-        const unitOfMeasure = shop && shop.unitsOfMeasure && shop.unitsOfMeasure[0].uom || "KG";
-        // at the moment shops don't have a kind of unitOfMeasure for distance
-        // so we put CM...
-        shippoParcel = createShippoParcel(cart.items[0].parcel, unitOfMeasure, "CM");
+        const unitOfMeasure = shop && shop.baseUOM || "kg";
+        const unitOfLength = shop && shop.baseUOL || "cm";
```
spencern: A little weird that the fallback units are different than the default base units, but I don't see any real problem with it and cm/kg are more widely used.

impactmass:
@spencern @lcampanis This PR introduced migration 9 into the marketplace branch, but that migration never runs because it was not included in the import file. Looks like that wasn't caught during review.

I came across this because I'm working on #2938. I've added it to the import file now, and working through some errors thrown.

lcampanis:
Thank you @impactmass!

spencern:
Thanks @impactmass

PR #2937, opened 2017-09-25
Fix alerts for confirmation when inviting new user

### LINKED ISSUE #2928: Inline alerts appear in the wrong place in accounts admin
### Expected behavior

The inline alert should happen above the form element and not overlap with card header. 

### Actual Behavior

Overlaps with the card header. 
screen_shot_2017-09-22_at_9_36_50_am

Invite a new user and see the alert overlaps with the card header. 

### Versions
Node: 4.5.0
NPM: 5.4.1
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.12.0
Reaction: 1.5.0
Reaction branch: marketplace

jm-wanja:
Resolve #2928 

**How To Test**

1. Invite a new user from the accounts admin dashboard and the invite owner form in the shop admin.
2. Observe that the success alert is a toast and the error alert is an inline alert

screen shot 2017-09-27 at 12 51 43 pm
screen shot 2017-09-27 at 12 51 23 pm


rymorgan:
@jm-wanja Now that I've thought about this a little more, we should use a toast alert in this instance.  It's a temporary alert and confirmation that the email has been sent.

jm-wanja:
@rymorgan this is noted. Should we also use a toast alert for the `invite owner form` that is on the Shop admin?

rymorgan:
@mikemurray @kieckhafer -- I'm changing my mind again on this. Do y'all have an opinion on the best UX here? Inline alert or toast alert confirmation when an invite email has been sent.

kieckhafer:
I'm a fan of the toasts, as it's less intrusive and doesn't alter any of the existing display when it shows.

mikemurray:
The toast is probably better in this situation since it's a success, and you only need to see that it succeed for a second to "get it".

rymorgan:
@jm-wanja So, yes then let's go to toast for both here and invite owner. The caveat is that if there is an error, that needs to happen inline so that they know where the error is and how to correct it.

jm-wanja:
@rymorgan I have implemented the new changes, toasts alerts for success and inline alert for error. I have done this for both invite admin and invite owner.

Issue #2942, opened 2017-09-25
Update babel-eslint to 8.0.0


spencern:
We need to update babel-eslint to v8.0.0 and make sure that it doesn't cause any errors.

![image](https://user-images.githubusercontent.com/1203639/30832593-dbe3fda4-a208-11e7-916c-33977165f56a.png)

spencern:
Fixed in `marketplace`/`1.5.x` by #2948 and in `1.4.3` by #2945

Issue #2933, opened 2017-09-22
Upgrade Meteor version to either 1.5.x or 1.6.x


spencern:
in #2901 I reverted Meteor version to 1.5.1 because app rebuilds were taking ~3+ minutes to occur.

In retrospect, there should have been a lot more discussion around that downgrade, at the time I was just trying to remove a blocker that was causing significant delays in app rebuilds after a code change.

*To reproduce the issue that 1.5.2 was causing:*
0. Pull `marketplace`
1. Change versions in your `.meteor/release` file to `1.5.2`
2. Start reaction with the `METEOR_PROFILE=1 METEOR_LOG=debug meteor run --verbose` command instead of using the `reaction-cli`
3. Wait until meteor has started the first time.
4. Add a new line in any server file.
5. Observe that Meteor rebuilds some things in CFS and every time, causing a 2-3 minute or more delay in restarting the server.

Specifically I'm seeing the restart hang with the following messages
```
> Downloading missing packages
> Preparing to build package cfs:worker
```

Notable that `cfs:worker` is not listed in the list of packages that Local package version is up-to-date.

The profiler looks like:
```
  Preparing project                         -
| ProjectContext prepareProjectForBuild.......................102,015 ms (1)
| ├─ _initializeCatalog............................................37 ms (1)
| │  ├─ LocalCatalog#initialize....................................33 ms (1)
| │  │  ├─ addPatternsToList                                        9 ms (2)
| │  │  ├─ LocalCatalog#_computeEffectiveLocalPackages              1 ms (1)
| │  │  └─ LocalCatalog#_loadLocalPackages.........................23 ms (1)
| │  │     ├─ PackageSource#initFromPackageDir.....................20 ms (4)
| │  │     │  ├─ optimistic statOrNull..............................3 ms (4)
| │  │     │  │  └─ safeWatcher.watch                               3 ms (1)
| │  │     │  ├─ runJavaScript package.js                           1 ms (4)
| │  │     │  ├─ sqlite query                                      10 ms (3)
| │  │     │  └─ other PackageSource#initFromPackageDir             5 ms
| │  │     └─ other LocalCatalog#_loadLocalPackages                 3 ms
| │  └─ other _initializeCatalog                                    4 ms
| ├─ _resolveConstraints...........................................44 ms (1)
| │  ├─ Select Package Versions....................................19 ms (1)
| │  │  ├─ new CS.Input                                            18 ms (1)
[...]

Issue #2586, opened 2017-07-23
Remove jQuery dependency


lil5:
If you are using es6 and react js

Aren't these helpful enough to remove jquery?

_Sent from my Wileyfox Swift using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

aaronjudd:
Reaction does not have a direct jquery dependency, but Meteor Blaze does. We'd love to see it gone, but it's a dependency of Blaze templates. Once we've converted the entire UI to React (almost there) there won't be any need for it, but I'm betting we have some Meteor dependencies that will still be including jquery - and we'll work through updating/removing as we identify those dependencies.

PR #2948, opened 2017-09-26
Meteor, Babel and deps maintainence

aaronjudd:
- updates babel presets to use env #2942
- updates to Meteor 1.5.2.1 #2933 
- resolve a few blockers for updating to 1.6.x
- remove duplicate bundling of jquery #2586
- removes circular import of migrations/aka versions
- restore bit hound checks 
- some code linting


imports/plugins/core/ui-navbar/client/components/navbar.js
```diff
@@ -8,7 +8,7 @@ async function openSearchModalLegacy(props) {
   if (Meteor.isClient) {
     const { Blaze } = await import("meteor/blaze");
     const { Template } = await import("meteor/templating");
-    const { default: $ } = await import("jquery");
+    const { default: $ } = await import("meteor/jquery");
```
spencern: good find

Issue #2921, opened 2017-09-21
Double lines and weird spacing in the permissions/account admin 


rymorgan:
### Expected behavior
The cards in the action panel shouldn't have extra space or double lines. And the detail view, in general, should be working properly. 

### Actual Behavior
They have extra space and double lines. And the panel lines don't extend all the way down the page. 
![accounts](https://user-images.githubusercontent.com/20409254/30719683-40d48b62-9ed9-11e7-8043-344eeb2e8b3d.png)

### Steps to Reproduce the Behavior
Navigate to accounts. 

### Versions
Node: 4.5.0
NPM: 5.4.1
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.12.0
Reaction: 1.5.0
Reaction branch: marketplace

rymorgan:
Also this: 

![accounts-2](https://user-images.githubusercontent.com/20409254/30721149-6b262280-9edf-11e7-9baf-18824d02e213.png)

mikemurray:
This should function in much the same way the orders panel does where the left & right sides are scrollable independently of one another.

zenweasel:
@rymorgan Was this issue closed by the above PR #2949?

rymorgan:
Yea, and the leftover/underlying problems were put into another issue https://github.com/reactioncommerce/reaction/issues/2963

PR #2949, opened 2017-09-26
Fix "Card not properly implemented" bug

efalayi:
Resolves [#2921](https://github.com/reactioncommerce/reaction/issues/2921)
Cards in action panel have extra space and double lines. And the panel lines don't extend all the way down the page.

`.accounts-table .col-md-3` had padding-left of 5px
`.accounts-table .col-md-9` also had border-right defined as: 1px solid #ccc;

###### Fix
- Remove padding left from `.accounts-table .col-md-3`
- Remove border-right from `.accounts-table .col-md-9`

###### Test
```
Sign in
Navigate to accounts.
```


rymorgan:
permissions-side-view

mikemurray:
The issue @rymorgan is pointing at is as follows.

The way the orders panel works is how this split between the "big" portion of the action view and the "smaller" detail view.

In the orders panel, you have an independent scrollable area for the "big" and "small" sides. See the below picture.

reaction

Rough example below of how to open.

"big" **Reaction.setActionView**
```
Reaction.setActionView({
  label: "Label",
  i18nKeyLabel: "some.i19n.key.label",
  template: "NameOfRegisteredTamplate",
  data: {}
});
```

"small" **Reaction.setActionViewDetail**
```
Reaction.setActionViewDetail({
  label: "Permissions",
  i18nKeyLabel: "admin.settings.permissionsSettingsLabel",
  data: {}, // Any data you want to pass in, like the context of the user
  template: "memberSettings"
});
```

impactmass:
@mikemurray It's fuzzy memory now....., but I remember when adding the accounts UI running into an issue with "reactivity" that created this whole situation in the first place. I think the issue was around data (either group / account or the combination) getting modified in the small actionViewDetail and not immediately reflecting big actionView. This might not help much but just throwing it out there in case it speeds this up.
I just went with this then, as the accounts work was holding up other things.

rymorgan:
I made a separate issue for the further fixes here: https://github.com/reactioncommerce/reaction/issues/2963

mikemurray:
@impactmass When you pass `data` into the action view, it's not meant to be reactive. The registered component/container component passed into `template` is then responsible for setting up its own tracker to then take that data and use it to query a reactive data source.

mikemurray:
So essentially, if you clicked on a group to the right, you'd pass in 

```
Reaction.setActionViewDetail({
  template: "groupSidebar"
  data: { groupId: "1234" }
});
```

Then in `groupSideba`r, you would have a composer, that subscribed to groups, takes in that data as a prop, and does a query for that actual group data.

spencern:
@mikemurray you've got comments here, but no review? Should this be merged?

mikemurray:
@spencern Those comments are better served in https://github.com/reactioncommerce/reaction/issues/2963. This PR is fine to merge.

Issue #2946, opened 2017-09-26
Address issue found in imports/plugins/included/product-detail-simple/client/containers/productDetailContainer.js [Opened by bitHound]


aaronjudd:
bitHound discovered 1 LintIssue in [imports/plugins/included/product-detail-simple/client/containers/productDetailContainer.js](https://github.com/reactioncommerce/reaction/blob/04e3fe45826b743ef09c50189cf00959bc1a2596/imports/plugins/included/product-detail-simple/client/containers/productDetailContainer.js). For details go to [bitHound](https://www.bithound.io/github/reactioncommerce/reaction/blob/04e3fe45826b743ef09c50189cf00959bc1a2596/imports/plugins/included/product-detail-simple/client/containers/productDetailContainer.js).

spencern:
Resolved by #2950

PR #2950, opened 2017-09-26
Fix lint issue found  in "imports/plugins/included/product-detail-simple/client/containers/productDetail.js" 

efalayi:
Resolves [#2946](https://github.com/reactioncommerce/reaction/issues/2946)
bitHound discovered 1 Lint Issue in imports/plugins/included/product-detail-simple/client/containers/productDetail.js. For details go to bitHound.

Seems `state` was directly manipulated on line 136

###### Fix
- Pass a param to setState to return mutation
- Rename `click` to `productClick`

```
this.setState(({ productClick }) => ({
  productClick: productClick + 1
}));
```


mikemurray:
If you are mutating state, with data from the previous state or props, it's recommended to pass a function as the first param into `setState` that returns your mutation like the following examples.

This will ensure that the state you are pulling is the most up-to-date it can be, as props and state are mutated asynchronously.

```
// Full definition
this.setState((prevState, props) => {
  return {
    click: prevState.click + 1
  };
});

// Using the setState function in the case of the click
  return {
    click: prevState.click + 1
  };
});

// Short hand arrow function with implicit return
this.setState((prevState) => ({
  click: prevState.click + 1
}));

// Even more shorter-hand with param descructuring and implicit return
this.setState(({ click }) => ({
  click: click + 1
}));
```

efalayi:
@mikemurray I've made the requested changes. I just learnt something new. Thank you very much.

Issue #2938, opened 2017-09-25
Create migration files for upgrade from 1.0 to 1.5


impactmass:
Extends / replaces #2604 with specifics.

The following are changes to be accounted for:

- *Primary Shop*: Coming to Marketplace setting requires one shop to be labelled as primary. We need a migration to add a "shopType": "primary" to the main shop. (introduced in #2577)
- *Order billing*: We need a migration to add shopId field to each billing object in Order.billing array. (introduced in #2682) (Also, confirm if needed on Cart too)
- *Order shipping*: Similar to billing above. (Introduced in #2791)
- *Rebuild OrderSearch*: OrderSearch collection needs to be rebuilt to match the changes in the Order Code (see existing [similar migration](https://github.com/reactioncommerce/reaction/blob/marketplace/imports/plugins/core/versions/server/migrations/1_rebuild_account_and_order_search_collections.js), might just need to promote that exiting one). 
- *Cart Product*: Whole Product data is required on Carts (introduced in #2610). Hence, we need a migration to add that in for old orders.

impactmass:
I got close to a WIP PR on this, but I'm currently resolving issue mentioned in "Added Units of Length" PR above

spencern:
Thanks for the update, let's get this knocked out today if possible :)

impactmass:
I'm about to open for review, a PR that address the 5 things listed on this ticket, but resolving these 5 revealed more 1.0-incompatible changes already in marketplace. Two new ones I just realized are around Order workflow and Refunds. I'll look into these and write up details for them description above, and continue with migration for them too.

For now, I'll write out "How To Test" section on the PR #2962, so anyone can jump in to start testing.

mikemurray:
I get this error when going from 1.0 to 1.5 when you try to invite a shop owner.

```
20:30:29.721Z ERROR Reaction: Failed to shop/createShop
  Error: Units of length is required
      at getErrorObject (packages/aldeed_collection2-core.js:480:15)
      at [object Object].doValidate (packages/aldeed_collection2-core.js:462:13)
      at [object Object].Mongo.Collection.(anonymous function) (packages/aldeed_collection2-core.js:214:25)
      at [object Object].Mongo.Collection.(anonymous function) [as insert] (packages/dispatch_run-as-user.js:325:19)
      at [object Object].shopCreateShop (server/methods/core/shop.js:88:25)
      at packages/check.js:128:16
[...]

PR #2962, opened 2017-09-27
Create migration files for upgrade from 1.0 to 1.5

impactmass:
Resolves in part #2938

**How To Test:**
- On an updated marketplace branch, git checkout to commit `91e18f26b7bc`. That's v1.0.0 commit. (You can also checkout into the tag if you have fetched it earlier).
- After `reaction reset -y && reaction`, create a few orders, and also add a product to cart without checking out.
- After that, checkout back to marketplace. Then run `reaction pull marketplace` (I do this just to use cli to re-instal modules). Then, fetch+checkout into branch `seun-migration-1.0-1.5`.
- Start `reaction` again. Confirm that migrations run up to version 15.
- Visit the Order dashboard, visit the PDP, and other pages. The goal is to have everything run smoothly as it should. This PR at the moment doesn't fully solve that. It gets us to the point of loading the app successfully, and being able to see Order dashboard (with shipping working).

Next up to resolve will be *Refunds* (which is the error you'll be seeing in your console). This PR as is can still be merged will I start working separately to fix the other issues. 

Other issues found (if related to migration) can be reported on the main ticket #2938.


imports/plugins/core/versions/server/migrations/15_update_shipping_status_to_workflow.js
```diff
@@ -0,0 +1,78 @@
+import { Migrations } from "meteor/percolate:migrations";
+import { Orders } from "/lib/collections";
+
+Migrations.add({
+  version: 15,
+  // Reaction v1.0 had 3 shipping boolean states (packed, shipped, delivered). Shipping workflow is not managed with a
+  // workflow object that keeps track of previous state.
+  up: function () {
+    Orders.find().forEach((order) => {
+      const currentShipping = order.shipping[0];
```
kieha: Should this be the hard-coded first shipping object or the shipping object associated with the current active shop?

impactmass: I did that to staying consistent with 1.0 (or below 1.5) codebase. It's all x[0] in there. Since this migration file is interacting with data created from such code, I referenced it that way. 
I think it's more recently that we started referencing shipping it by shopId (i.e marketplace).

spencern:
One issue I'm seeing that could be resolved in the next ticket is that routes are different between 1.0 and 1.5.

I'm not really sure the best way to resolve this, as it will probably need some type of redirect, and it will be different depending on the shop, especially if any shops in the 1.0 -> 1.4.3 have I think what we need is some way to define redirects within the app. This really will only be an issue for SEO, as pathfor, etc have been updated internally, but that's a pretty huge issue for any live shop.

image

spencern:
Also seeing this issue here (same as what @mikemurray reported in #2938), but tested and it's also present in `marketplace right now, so not blocking for me.

```
23:39:46.379Z ERROR Reaction: Failed to shop/createShop
  Error: Units of length is required
      at getErrorObject (packages/aldeed_collection2-core.js:480:15)
      at [object Object].doValidate (packages/aldeed_collection2-core.js:462:13)
      at [object Object].Mongo.Collection.(anonymous function) (packages/aldeed_collection2-core.js:214:25)
      at [object Object].Mongo.Collection.(anonymous function) [as insert] (packages/dispatch_run-as-user.js:325:19)
      at [object Object].shopCreateShop (server/methods/core/shop.js:88:25)
      at packages/check.js:128:16
      at packages/ddp-server/livedata_server.js:1686:15
      at packages/ddp-server/livedata_server.js:1684:36
      at [object Object]._.extend.applyAsync (packages/ddp-server/livedata_server.js:1683:12)
      at [object Object]._.extend.apply (packages/ddp-server/livedata_server.js:1622:26)
      at [object Object]._.extend.call (packages/ddp-server/livedata_server.js:1604:17)
      at [object Object]._inviteShopOwner (server/methods/accounts/accounts.js:557:29)
```

Issue #2931, opened 2017-09-22
Sign in and Sign out not working properly in dropdown. 


rymorgan:
### Expected behavior

I should be able to sign in and sign out smoothly. You should sign in have spinner if necessary and go directly to your store with no flashing weirdness. 

When I sign out it should simply return me to the signed out version of the store.

### Actual Behavior

When I sign in with the dropdown, the full sign in page flashes behind it producing a very rough looking sign in. 

When I sign out it flashes the full sign out page (this should all just happen in the dropdown). Then it takes me the store. After a longish pause it flashes the sign out page again, then starts spinning infinitely. 

![sign in sign out](https://user-images.githubusercontent.com/20409254/30762354-9399108a-9f96-11e7-8635-94d37140e6d2.gif)

### Steps to Reproduce the Behavior

Sign in using the dropdown as an admin.
Sign out using the dropdown as an admin. 

Happens in regular browser but does not happen in incognito mode. 

### Versions
Node: 4.5.0
NPM: 5.4.1
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.12.0
Reaction: 1.5.0
Reaction branch: marketplace

rymorgan:
Likely related: https://github.com/reactioncommerce/reaction/issues/2328

efalayi:
![rc-sign-out](https://user-images.githubusercontent.com/28438068/30806918-9112873c-a1ef-11e7-8608-5abf3bfb6421.gif)
@rymorgan, I also observed that when the user tries to sign out from profile page, the user is not redirected to the home page.

rymorgan:
I'll reset my install and see if I can reproduce the error. If I can't, then I'll close the issue.

kieckhafer:
Not seeing all the issues described, but one thing I am noticing is the loader in the navbar. We should remove that loader icon from the sign in. There is already an icon on the page, and if it's in the navbar too, it pushes everything down awkwardly.

rymorgan:
I'm still seeing it after resetting my install, Mike can reproduce as well. Sometimes you have to sign in/ sign out multiple times.

mikemurray:
It doesn't happen every time, but after a couple of tries, I get an infinite spinner when I sign in / sign out on the PDP page.

Also, on routes other than `/` I get a flash of the "sign in" form. in the page as my route permissions are changing between user groups. 

[...]

Issue #2750, opened 2017-08-28
[Marketplace] Get multiple console errors when switching shops


zenweasel:
### Expected behavior
No errors

### Actual Behavior
Get a cascade of errors, probably all due to not waiting on the `Shop` subscription
```
TypeError: Cannot read property 'brandAssets' of undefined
    at Brand.getLogo (brand.js:23)
    at Brand.render (brand.js:33)
    at modules.js?hash=09e8e2dfa58d9bd3fd253c6052a7e955435c914c:107773
    at measureLifeCyclePerf (modules.js?hash=09e8e2dfa58d9bd3fd253c6052a7e955435c914c:107053)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (modules.js?hash=09e8e2dfa58d9bd3fd253c6052a7e955435c914c:107772)
    at ReactCompositeComponentWrapper._renderValidatedComponent (modules.js?hash=09e8e2dfa58d9bd3fd253c6052a7e955435c914c:107799)
    at ReactCompositeComponentWrapper._updateRenderedComponent (modules.js?hash=09e8e2dfa58d9bd3fd253c6052a7e955435c914c:107723)
    at ReactCompositeComponentWrapper._performComponentUpdate (modules.js?hash=09e8e2dfa58d9bd3fd253c6052a7e955435c914c:107701)
    at ReactCompositeComponentWrapper.updateComponent (modules.js?hash=09e8e2dfa58d9bd3fd253c6052a7e955435c914c:107622)
    at ReactCompositeComponentWrapper.receiveComponent (modules.js?hash=09e8e2dfa58d9bd3fd253c6052a7e955435c914c:107524)
```

### Steps to Reproduce the Behavior
Because this is a race-condition type thing you won't always see it, but usually after a couple of switches you can.

1. Login as admin
1. Create a second shop
1. As marketplace admin switch to second shop
1. Observe errors

### Versions
```
Node: 7.10.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.0-ce
```

mikemurray:
@zenweasel @aanchirinah Made a PR here that should address this https://github.com/reactioncommerce/reaction/pull/2966

impactmass:
closed by #2966

PR #2966, opened 2017-09-27
Fixes issues related to missing `shop` on sign in / sign out

mikemurray:
Fixes https://github.com/reactioncommerce/reaction/issues/2931 https://github.com/reactioncommerce/reaction/issues/2750

What was happening...

Signing in and out of the app sometimes causes the screen to be blank or causing with some components to not render at all due to errors. This is because shop data is no longer available on the client and many components that assumed shop would always be available. 

This PR, add guards around the initial set of components that were complaining about missing `shop`. It further guards against the shop not being available at the router level and replaces subscription manager with local subscriptions.

For some reason, using the subscription manager in the router `startup.js` was causing issues with sign in and sign out. Perhaps this is due to how it caches subscriptions or otherwise. By switching to local subscriptions, it appears this problem is resolved or at least lessened its frequency.

I added a `TODO` to revisit with subscription manager. This is a really good case for using `Redux` or something more transparent for debugging purposes. :)

### DIFF
--- a/imports/plugins/core/router/client/startup.js
+++ b/imports/plugins/core/router/client/startup.js
@@ -10,7 +10,7 @@ Meteor.startup(function () {
   loadRegisteredComponents();
 
   // Subscribe to router required publications
-  // Note: Although these are subscribed to by the subscription mamager in "/modules/client/core/subscriptions",
+  // Note: Although these are subscribed to by the subscription manager in "/modules/client/core/subscriptions",
   // using the subscriptions manager sometimes causes issues when signing in/out where you may seee a grey screen
   // or missing shop data throughout the app.
   // TODO: Revisit subscriptions manager usage and waiting for shops to exist client side before rendering.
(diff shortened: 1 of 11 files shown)


zenweasel:
I'm still seeing the flashes of different screens that @rymorgan noted in his ticket.

imports/plugins/core/ui-navbar/client/containers/navbar.js
```diff
@@ -1,11 +1,15 @@
...
 import { Reaction } from "/client/api";
 import NavBar from "../components/navbar";
+import { Media, Shops } from "/lib/collections";
 
 function composer(props, onData) {
+  const shop = Shops.findOne(Reaction.getShopId());
   const searchPackage = Reaction.Apps({ provides: "ui-search" });
   let searchEnabled;
   let searchTemplate;
+  let bandMedia;
```
impactmass: Typo, I believe this should be `brandMedia`.

imports/plugins/core/router/client/startup.js
```diff
@@ -2,25 +2,34 @@ import { loadRegisteredComponents } from "@reactioncommerce/reaction-components"
...
-import { Reaction } from "/client/api";
+import { Shops } from "/lib/collections";
 import { initBrowserRouter } from "./browserRouter";
 import { Router } from "../lib";
 
 Meteor.startup(function () {
   loadRegisteredComponents();
 
+  // Subscribe to router required publications
+  // Note: Although these are subscribed to by the subscription mamager in "/modules/client/core/subscriptions",
```
impactmass: Typo `manager`

PR #2971, opened 2017-09-28
Router Performance fix to include theme in route's context

lcampanis:
Layout's theme is now included in each route when Router initialises all routes `Router.initPackageRoutes` so it can be easily included in as a body-class.

Previously, in order to get the theme, the layout was refetched in each route change using `getRouteLayout()` which used the same mechanics with Router for layout resolve, with fix it is now more performant.

Credit @prosf


spencern:
@mikemurray can you review this one? You've got more router experience/wrote it

mikemurray:
@spencern reviewing now.

mikemurray:
Not related to this PR, but there are some more issues with routing and grey screen logged here: https://github.com/reactioncommerce/reaction/issues/2975

Will review once routing is reliable again.

PR #2973, opened 2017-09-28
Fixed Tax Cloud API not not loading tax codes on PDP

### LINKED ISSUE #2533: TaxCloud: Tax codes dropdown is empty after api configuration.
### Expected behavior
Tax codes from tax cloud should load into the `tax code` dropdown in the product detail page after the tax cloud API has been configured.

### Actual Behavior
This does not happen.

1. Start reaction
2. Configure taxcloud API
3. Access tax codes dropdown in product detail page

### Versions
???

[Updated 7/10/2017 by @spencern]

aanchirinah:
Resolves #2533

Tax Cloud API methods was not fetching the tax codes , so I added an `onSuccess` method in the Tax cloud Autoform hook to fetch tax codes.

##### HOW TO TEST 
- Configure Tax Cloud API
- Access tax codes dropdown from PDP


imports/plugins/included/product-variant/containers/variantFormContainer.js
```diff
@@ -89,6 +89,21 @@ const wrapComponent = (Comp) => (
         }
       });
       const taxCodesArray = [];
+      // if (provider.name === "taxes-taxcloud") {
```
spencern: Don't commit commented out code -

aanchirinah: Fixed requested changes

imports/plugins/included/taxes-taxcloud/client/settings/taxcloud.js
```diff
@@ -2,6 +2,7 @@ import { Template } from "meteor/templating";
 import { AutoForm } from "meteor/aldeed:autoform";
 import { Packages } from "/lib/collections";
 import { Reaction, i18next } from "/client/api";
+import { Meteor } from "meteor/meteor";
```
zenweasel: Import order

imports/plugins/included/taxes-taxcloud/client/settings/taxcloud.js
```diff
@@ -20,8 +21,21 @@ Template.taxCloudSettings.helpers({
 AutoForm.hooks({
   "taxcloud-update-form": {
     onSuccess: function () {
-      return Alerts.toast(i18next.t("admin.taxSettings.shopTaxMethodsSaved"),
-        "success");
+      Meteor.call("taxcloud/getTaxCodes", (err, res) => {
+        if (err) {
+          throw new Meteor.Error(Number, "description");
```
zenweasel: Why are you are throwing a JS type as the first parameter to `Meteor.Error`?

aanchirinah: Will fix that , I was not sure the error code I wanted to use at that time so this is an oversight.

kieha:
When I select a tax code from the drop-down I get this console error:

screen shot 2017-09-29 at 1 55 05 pm

kieha:
I think you should safeguard against someone clicking the `Save Changes` button twice. I just clicked it twice and got this exception:

screen shot 2017-09-29 at 1 58 05 pm

THis happens because the DB is already populated with these values so when you click the `Save Changes` button again, it tried to populate the DB again with the same values, thus throwing the error.

aanchirinah:
@kieha , I have fixed the requested changes and errors you experienced

imports/plugins/included/taxes-taxcloud/client/settings/taxcloud.js
```diff
@@ -20,8 +22,23 @@ Template.taxCloudSettings.helpers({
 AutoForm.hooks({
   "taxcloud-update-form": {
     onSuccess: function () {
-      return Alerts.toast(i18next.t("admin.taxSettings.shopTaxMethodsSaved"),
-        "success");
+      if (!TaxCodes.findOne({ taxCodeProvider: "taxes-taxcloud" })) {
+        Meteor.call("taxcloud/getTaxCodes", (err, res) => {
+          if (err) {
+            throw new Meteor.Error(500, "description");
```
spencern: We no longer recommend using number error codes. Also "description should be an actual description of the error.

The suggested template for errors is
```
throw new Meteor.Error("short-description", "Longer description")
```

E.g. 
```
throw new Meteor.Error("operation-failed", "client safe message about why")
```

imports/plugins/included/taxes-taxcloud/client/settings/taxcloud.js
```diff
@@ -20,8 +22,23 @@ Template.taxCloudSettings.helpers({
...
+          } else if (res && Array.isArray(res)) {
+            Alerts.toast(i18next.t("admin.taxSettings.shopTaxMethodsSaved"),
+              "success");
+            res.forEach((code) => {
+              Meteor.call("taxes/insertTaxCodes", Reaction.getShopId(), code, "taxes-taxcloud", (error) => {
+                if (error) {
+                  throw new Meteor.Error("Error populating TaxCodes collection", error);
```
spencern: This error uses a different format than your previous error.

As noted above, this will not alert the user via the UI, perhaps a Toast alert with the "error" style would be more appropriate here? I'm not too familiar with this tax plugin, so I'll defer to @zenweasel on what exactly the correct process is here.

imports/plugins/included/taxes-taxcloud/client/settings/taxcloud.js
```diff
@@ -20,8 +22,23 @@ Template.taxCloudSettings.helpers({
 AutoForm.hooks({
   "taxcloud-update-form": {
     onSuccess: function () {
-      return Alerts.toast(i18next.t("admin.taxSettings.shopTaxMethodsSaved"),
-        "success");
+      if (!TaxCodes.findOne({ taxCodeProvider: "taxes-taxcloud" })) {
+        Meteor.call("taxcloud/getTaxCodes", (err, res) => {
+          if (err) {
+            throw new Meteor.Error(500, "description");
+          } else if (res && Array.isArray(res)) {
```
spencern: I'm a bit skeptical that this code is doing what you want.

There are a few problems I want to point out:

1. You're toasting success here, and then looping through the responses and potentially throwing an error. It seems contradictory to alert the user of success if the error processing hasn't happened yet.

2. If you throw an error in this `forEach` loop, the loop will stop executing in the middle. You're not catching the error anywhere, and the error will not `toast` alert the user, it will just stop the app from continuing and log the error in the console.

imports/plugins/included/taxes-taxcloud/client/settings/taxcloud.js
aanchirinah: @spencern , thanks for pointing it out, I will fix it

imports/plugins/included/taxes-taxcloud/client/settings/taxcloud.js
```diff
@@ -20,8 +22,27 @@ Template.taxCloudSettings.helpers({
...
+            Alerts.toast(
+              `${i18next.t("admin.taxSettings.shopTaxMethodsFailed")} ${err}`, "error"
+            );
+          } else if (res && Array.isArray(res)) {
+            Alerts.toast(i18next.t("admin.taxSettings.shopTaxMethodsSaved"),
+              "success");
+            res.forEach((code) => {
+              Meteor.call("taxes/insertTaxCodes", Reaction.getShopId(), code, "taxes-taxcloud", (error) => {
+                if (error) {
+                  Alerts.toast(
```
spencern: This is better, but the issue of toasting success and then toasting failure remains

imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
```diff
@@ -16,9 +16,9 @@ const taxCalc = {};
 
 taxCalc.getPackageData = function () {
   const pkgData = Packages.findOne({
```
aaronjudd: Is this the correct pattern here?

spencern:
What's the status on this?

imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
zenweasel: I think we want to use `Reaction.getPackageSettings`?

aanchirinah:
@spencern , fixing requested changes and pending reviews.

imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
aanchirinah: What is supposed to be the correct pattern? @aaronjudd

imports/plugins/included/taxes-taxcloud/client/settings/taxcloud.js
aanchirinah: So should I scrape the toast of failure inside `onSuccess` hook? @spencern

imports/plugins/included/taxes-taxcloud/client/settings/taxcloud.js
aanchirinah: I realised it wasn't necessary to be checking for errors while looping so I have refactored my code to work better.

imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
spencern: you should use the Reaction.getPackageSettings method:
https://github.com/reactioncommerce/reaction/blob/marketplace/server/api/core/core.js#L432

zenweasel: Also, is this change to Avalara required to make TaxCloud work? Something seems fishy?

aanchirinah: No it was to make Avalara work, and I have made the correct pattern change in my latest commit, @spencern @aaronjudd @zenweasel

imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
```diff
@@ -15,11 +15,8 @@ const requiredFields = ["username", "password", "apiLoginId", "companyCode", "sh
 const taxCalc = {};
 
 taxCalc.getPackageData = function () {
-  const pkgData = Packages.findOne({
-    name: "taxes-avalara",
-    shopId: Reaction.getShopId(),
-    enabled: true
-  });
+  const pkgData = Reaction.getPackageSettings("taxes-avalara");
+  pkgData.settings.avalara = true;
```
zenweasel: Help me understand why we need to set a value in the package settings?

zenweasel:
@aanchirinah Although I guess these test steps work, in the future I would like to see more detailed steps so that someone who is not super familiar with the project could be able to test it.

imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
aanchirinah: During my sync with @kieha , she made me aware the `enabled` boolean variable wasn't working properly and that `settings.avalara` is rather being used in the implementation of Avalara.

imports/plugins/included/taxes-taxcloud/client/settings/taxcloud.js
```diff
@@ -20,8 +22,19 @@ Template.taxCloudSettings.helpers({
...
+          if (res && Array.isArray(res)) {
+            Alerts.toast(i18next.t("admin.taxSettings.shopTaxMethodsSaved"),
+              "success");
+            res.forEach((code) => {
+              Meteor.call("taxes/insertTaxCodes", Reaction.getShopId(), code, "taxes-taxcloud");
+            });
+          }
+        });
+      } else {
+        Alerts.toast("Taxcodes from Tax Cloud already loaded");
```
zenweasel: This needs a translation

imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
zenweasel: Do we need to open an issue to make this work correctly if we are working around a bug elsewhere? We don't just hack something and move on, we would expect a `TODO` item at least.

zenweasel:
@aanchirinah You've introduced a lint error now that you need to resolve.

imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
aanchirinah: Duly noted, I would put a a `TODO` above it then

zenweasel:
I fixed lint error and TODO comment so that we could go ahead and merge as it appears to work and fix this bug

zenweasel:
@aanchirinah Actually it looks like we have failing tests

stunjiturner:
I am using 1.13 reaction, when I use the dropdown, after selecting the taxable toggle. the dropdown is empty

PR #2979, opened 2017-09-29
Save LocalStorage user currency in user.profile

lcampanis:
LocalStorage currency set based on the user's location, should also be set in the user's profile for consistency, which can then be applied throughout the app.

Credit @stevaras


spencern:
This seems like a good idea to me. Looks like you guys were responsible for the original code, so no need to go hunting for the original author to figure out why we were using localstorage to begin with.

@stevaras @lcampanis do you think there's still a need for saving the currency in localstorage? Should it be stored exclusively in profile?

stevaras:
@spencern Basically it's used as a fallback for the user. In the callback of shop/getLocale where we initialize the user's preferences we give this hierarchy:
1) User's profile ( which is the most personal )
2) LocalStorage which is more related to the geographical position of the user
3) Primary's shop default value

For example let's say a Reaction Commerce site is in China. The primary shop has it's currency in Yen. A user visits the site from UK. We want:
*) The registered user to see everything in his preferred currency ( the one he chose from the currency dropdown at some point)
*) An anonymous user to see currencies according to his geographical position ( in this example GBP )
*) If a user doesn't have profile preference and for some reason we couldn't get his position, then we use as a fallback the Primary's Shop currency which is Yen.  

Basically now that I think about it we could just set the profile by default instead of localStorage,

spencern:
@stevaras I think that using profile as a default makes more sense to me than the localstorage option - it also scopes the setting to the user rather than to the browser/machine which seems better to me.

kieckhafer:
Just echoing @spencern , this is a good idea, and seems like it should be the primary way to do it:

```
In your example, of a user visiting a Reaction site, where does localstorage fit in? It's not going to help determine preference unless the user has visited before, and if the user has visited before then the profile should be available.

Seems to me that your current PR contains the best way to do it, but also that we could remove the localstorage check entirely
```

spencern:
I'm okay to merge this in as is, but would prefer to remove the localStorage option now if that's a direction we can all agree on.

lcampanis:
@spencern @kieckhafer Agreed thank you! We'll remove LocalStorage and push the update.

spencern:
@lcampanis I'd like to merge this today if possible so it can be in our `1.5.0` release. When do you think you'll have time to update?

lcampanis:
@spencern No problem, I'll get the update in the next few hours for you.

spencern:
@lcampanis I'm updating the target branch here to `release-1.5.0` it shouldn't affect you, but wanted to give you the heads up.

lcampanis:
All set @spencern

spencern:
Great! Thanks @lcampanis

lcampanis:
No problem @spencern. Updated it, let me know if there's something else ;)

Issue #2773, opened 2017-08-31
Discounts not displayed properly in Orders panel


zenweasel:
### Expected behavior
Discounts should show as a line item and be reflected in totals

### Actual Behavior
"Add Discount" is displayed but the actual discount is not

### Steps to Reproduce the Behavior
1. Login as an admin and set up shipping and payment methods
1. Add a discount code
1. Check out and use that code
1. Go to orders panel
1. Observe that discounts are not displayed

### Versions
```
Node: 7.10.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.0-ce
```

spencern:
May be related to #2763

aaronjudd:
@zenweasel self-assigned and this hasn't been updated in 14 days. I'm assuming that this is a blocker as it probably affecting a single shop instance as well?  If it is working then we can at least remove the "marketplace" label.

zenweasel:
As far as I remember (and as listed in the test steps) this affects single shop instances as well. I thought I was going to knock this out with the other order panel changes but apparently not. Unassigning myself for now and will grab it again when I have cleared the other stuff from my plate.

zenweasel:
Tested again and verified that this happens in a single-shop instance. Discounts work fine just not displayed in the dashboard. Might just be a UI thing that got lost.

zenweasel:
Closed via #2980

PR #2980, opened 2017-09-29
Fix "Discounts not displayed properly in Orders panel" bug #2773

efalayi:
Resolves [#2773](https://github.com/reactioncommerce/reaction/issues/2773)
"Add Discount" is displayed but the actual discount amount is not

**Add Discount** is a tag link which when clicked displays a form or list of discount codes
Total displayed does not reflect discount (if included)

###### Fix
- Register component that displays discount lists/form
```
registerComponent("DiscountList", discountListComponent);
```

- Replace `Add Discount` tag with *discount amount*
```
 {formatPriceString(invoice.discounts)}
```

- Update `getTotalByShop` function to include discount amount
```
const discount = parseFloat(this.getDiscounts());
const discountTotal = Math.max(0, shopSubtotal - discount);
const shopTotal = discountTotal + shopTaxes + shipping;
```

###### Test
```
Login as an admin and set up shipping and payment methods
Add a discount code
Check out and use that code
Go to orders panel
Observe that discount amount is displayed
```


lib/collections/transform/cartOrder.js
```diff
@@ -203,14 +203,16 @@ export const cartOrderTransform = {
     const subtotals = this.getSubtotalByShop();
     const taxes = this.getTaxesByShop();
     const shipping = parseFloat(this.getShippingTotal());
+    const discount = parseFloat(this.getDiscounts());
```
zenweasel: I know I said we should be reflecting discount everywhere but this is actually the exception. We haven't finalized how we are handling discounts in a multi-shop scenario and I think this is going to give us odd results, so let's back out this change for now.

joykare:
@efalayi ready for a review?

lib/collections/transform/cartOrder.js
efalayi: @zenweasel when I add a discount, it's not reflected in `Total` displayed on Orders page.

zenweasel:
@efalayi Can you add a little more context to your comment. In a multi-shop scenario or? Where on the orders page using what payment provider? When I tested it in a single shop I was seeing the discount show up

efalayi:
@zenweasel here are screenshots of an order I just made
admin-screen
completed-order

Observe total displayed. They do not match.

zenweasel:
Added this issue #3005 for Stripe, will deal with separately.

Issue #2958, opened 2017-09-27
Shippo does not work with newly created products


zenweasel:
### Expected behavior
Should receive shipping quotes

### Actual Behavior
Receive no shipping quotes

### Steps to Reproduce the Behavior
1. Login as administrator
1. Enable Shippo and enable carriers
1. Create a product (don't use the built-in product)
1. Add **only** the newly created product to cart
1. Attempt to checkout
1. Observe that "No shipping methods are configured. Configure now." is displayed

*Note* this happens on both `master` and `marketplace`. I believe this is because the shippo method is expecting a `parcel` object on the product but it's stored on the variant. This works with the "Basic Reaction Product" because it's coded into the JSON.

### Versions
```
Node: 7.10.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.12.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.2-ce
```

zenweasel:
@spencern / @aaronjudd  Not sure if this is something we need to triage first, but it seems like `marketplace` blocker?

spencern:
Does seem like this could be a blocker. Even if it's not a regression, it should probably be resolved asap.

zenweasel:
So I think the solution is simple but twofold.

1. Restore/re-enable setting dimensions/weight at the product level on the PDP (you can currently only set it on the top-level variant)
1. Have shippo use either the variant or the product dimensions/weight when getting quotes for Shippo

zenweasel:
So it actually looks like you can't even set parcel at the variant level. When you set them you get `height`, `weight`, etc. on the top level variant. Not sure where `parcel` is coming from.

spencern:
This is something that @rymorgan and I have discussed in https://github.com/reactioncommerce/reaction/pull/2893#issuecomment-332406765 a little bit.

rymorgan:
@spencern I'm not sure I'm following exactly what's going on here. I definitely see that I'm getting the no shipping methods configured when I follow the steps outlined by @zenweasel -- if we need to allow/force height weight etc at option level to get this to work, then we can just add those fields. We can figure out a better UX down the line if that's the case.

spencern:
Resolved by #2992

PR #2992, opened 2017-10-02
Allow settings of "parcel" object from PDP Options and store on cart

zenweasel:
This closes https://github.com/reactioncommerce/reaction/issues/2958

I debated just **moving** these dimensions from the variant to the option but tried to make it backwards compatible.

1. Login as administrator
1. Enable Shippo and enable carriers
1. Create a product (don't use the built-in product) and set weight and dimensions
1. Add **only** the newly created product to cart
1. Attempt to checkout
1. Observe that shipping methods are displayed


spencern:
@rymorgan this gets into your territory a little bit as well.

PR #2995, opened 2017-10-02
Missing translations on order filters

joykare:
### Problem
- Missing translations on order filters.

### Fix
- Added translations


spencern:
@joykare @kieckhafer yes, we should be moving translations for plugins into the plugin translations file.
App-wide translations are fine to stay in the `private/data/i18n` file, but any definitions that are specific to a plugin should be in the plugin.

spencern:
@joykare @kieckhafer After looking through the i18n file in `/private/data` and seeing how many orders definitions and other definitions that should be elsewhere, I'm punting on that move to #3009  so that we can move on from this.

PR #3004, opened 2017-10-02
sync inventory policy toggle between top-level variants and child variants

### LINKED ISSUE #2911: Inventory policy switch on variant does update options
image
1. Toggling inventoryPolicy / "Allow Backorder" for a variant does _not_ cascade that choice into it's options

2. Options do not have a switch that will permit toggling the inventory policy. I had to edit the inventoryPolicy for the options in the database directly in order to permit backordering on the default product.

kieckhafer:
Resolves #2911 

- Introduce `ReactionProduct.getSiblings()` to get documents of a variants `siblings`, who have the same ancestor tree
- Introduce `ReactionProduct.getParent()` to get document of a variants direct parent
- Updates the way the inventory policy toggle updates.

If a variant has no child variants, the toggle functionality is unchanged.

If a variant has child variants, the toggle is disabled on the top-level variant, and inherits the following value, based on it's children:

1. If **_any_** of the children are set to `inventoryPolicy === false`, meaning backorders are allowed, then the top-level variant will also be set to `inventoryPolicy === false`
2. If _**all**_ of the children are set to `inventoryPolicy === true`, then the top-level variant will also be set to `inventoryPolicy === true`

### DIFF
--- a/imports/plugins/included/product-variant/components/variantForm.js
+++ b/imports/plugins/included/product-variant/components/variantForm.js
@@ -6,6 +6,7 @@ import Velocity from "velocity-animate";
 import "velocity-animate/velocity.ui";
 import { Components } from "@reactioncommerce/reaction-components";
 import { formatPriceString } from "/client/api";
+import { Reaction } from "/lib/api";
 
 const fieldNames = [
   "title",
@@ -129,6 +130,7 @@ class VariantForm extends Component {
   }
 
   handleFieldBlur = (event, value, field) => {
+    console.log("field", field);
     if (this.props.onVariantFieldSave) {
       this.props.onVariantFieldSave(this.variant._id, field, value, this.state.variant);
     }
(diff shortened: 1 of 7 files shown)


rymorgan:
If we are showing the inventory control on the options, don't we also need to show the warn at field? 
inventory-warn-at

imports/plugins/included/product-variant/containers/variantFormContainer.js
```diff
@@ -195,6 +199,44 @@ const wrapComponent = (Comp) => (
       Meteor.call("products/updateProductField", variant._id, "isVisible", !variant.isVisible);
     }
 
+    /**
+     * @method updateInventoryPolicyIfChildVariants
+     * @description update parent inventory policy if variant has children children
```
zenweasel: typo? "children children"

imports/plugins/included/product-variant/containers/variantFormContainer.js
```diff
@@ -195,6 +199,44 @@ const wrapComponent = (Comp) => (
...
+     * @return {undefined} return nothing
+     */
+    updateInventoryPolicyIfChildVariants = (variant) => {
+      // Get all siblings, including current variant
+      const options = ReactionProduct.getSiblings(variant);
+      // Get parent
+      const parent = ReactionProduct.getVariantParent(variant);
+
+      // If this is not a top-level variant, update top-level inventory policy as well
+      if (parent) {
```
zenweasel: Could we combine this into one conditional e.g. `if (parent && options && options.length)` ?

imports/plugins/included/product-variant/components/variantForm.js
```diff
@@ -6,6 +6,7 @@ import Velocity from "velocity-animate";
 import "velocity-animate/velocity.ui";
 import { Components } from "@reactioncommerce/reaction-components";
 import { formatPriceString } from "/client/api";
+import { Reaction } from "/lib/api";
```
zenweasel: Imported but not used.

zenweasel:
So is the default behavior supposed to be that if I create an option for a product it's "disabled" at the top level forever more?

kieckhafer:
@zenweasel 

> So is the default behavior supposed to be that if I create an option for a product it's "disabled" at the top level forever more?

yes

PR #3013, opened 2017-10-03
Added missing Mail parameter for reaction.json.example

### LINKED ISSUE #2516: mail parameter in example reaction.json is missing an option
Not a big issue, but still I want to share it.

### Expected behavior
When setting the mail parameter in reaction.json correctly the mail server should be working after starting reaction up and show a green dot in the dashboard.

The example reaction.json says:
```
            "mail": {
                "user": "",
                "password": "",
                "host": "",
                "port": ""
            },
```

### Actual Behavior
Settings are put correctly in the dashboard, but the red stays red. Obviously the "service" option is missing.  The "service" parameter (e.g. "service": "custom") has to be set in reaction.json additionally. 

Like this it works fine for me:
```
            "mail": {
                "service": "custom",
                "user": "",
                "password": "",
                "host": "",
                "port": ""
            },
```

### Versions
Reaction: 1.3.0

aanchirinah:
Resolves #2516 

Added a `service` parameter to the mail object in `reaction.json.example`.

#### HOW TO TEST
- Rename the `reaction.json.example` in `/private/settings/` to `reaction.json`.
- Perform a `reaction reset` and then `reaction` in your terminal.
- Log in as admin and configure your email SMTP , you will now see that the status changes to green.

### DIFF
--- a/private/settings/reaction.json.example
+++ b/private/settings/reaction.json.example
@@ -18,7 +18,8 @@
                 "user": "",
                 "password": "",
                 "host": "",
-                "port": ""
+                "port": "",
+                "service": ""
             },
             "cart": {
                 "cleanupDurationDays": "older than 3 days"
--- a/private/settings/reaction.json.example
+++ b/private/settings/reaction.json.example
@@ -19,7 +19,7 @@
                 "password": "",
                 "host": "",
                 "port": "",
-                "service": ""
+                "service": "custom"
             },
             "cart": {
                 "cleanupDurationDays": "older than 3 days"


zenweasel:
@aanchirinah So what am I supposed to be seeing when I test this PR?

aanchirinah:
@zenweasel , I have added more context now on how to test.

zenweasel:
So I am still unclear what is supposed to be accomplished here. (the original ticket is a little confusing to me as well)

If I follow the instructions and click on the settings icon I get an error. If I "Log in as admin and configure your email SMTP" of course it goes green but that was true before this PR. Help me understand what is supposed to be different. Obviously this change is fine but I want to ensure we are actually solving the issue.

Issue #3023, opened 2017-10-04
Allocation failed - process out of memory


joshuacox:
### Expected behavior
Build a custom image from a default checkout of reaction with no modifications

### Actual Behavior
CircleCI documents my build failure [here](https://circleci.com/gh/OpenBusinessBox/reaction/17)

that is  on the same commit as [8984 here](https://circleci.com/gh/reactioncommerce/reaction/8984) which passed, yet mine fails.  

I have another [branch](https://circleci.com/gh/OpenBusinessBox/reaction/5) which is actually passing. However, it is using meteor 1.5.1

There are many other builds there [too](https://circleci.com/gh/OpenBusinessBox/reaction).  I believe the only ones that have passed so far have been using meteor 1.5.1, like this [one](https://circleci.com/gh/OpenBusinessBox/reaction/16), which is identical to this [one](https://circleci.com/gh/OpenBusinessBox/reaction/14) except the latter is 1.5.2.1

### Steps to Reproduce the Behavior

Fork reaction, make a new circleCI account, add docker env vars, build from master 

-or-

```
reaction init
cd reaction
reaction build mycustom
```

### Versions
```
Node: 8.6.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.14.2
Reaction: 1.5.0
Reaction branch: reaction-8984
Docker: 17.07.0-ce
```

There is also a [forum post](https://forums.reactioncommerce.com/t/1-4-2-fatal-error-call-and-retry-last-allocation-failed-process-out-of-memory/532/27)  where I've been documenting this issue for a few weeks now.

I've got a good number of machines running ubuntu zesty and xenial, debian stretch, and arch linux that all exhibit this behavior (most of which have tons of available RAM), with circleCI also doing the same.

### CirclleCI
branch  | 1.5.1 | 1.5.2.1 | 1.5.2.2 |
 ------------ | :-----------: | -----------: | -----------: |
master       |   [**success**](https://circleci.com/gh/OpenBusinessBox/reaction/18)    |         [**fail**](https://circleci.com/gh/OpenBusinessBox/reaction/20) |     [**fail**](https://circleci.com/gh/OpenBusinessBox/reaction/22) |
[...]

PR #3033, opened 2017-10-04
Gitlab ci.yml

joshuacox:
I am trying to have very reproducible issues when it comes to stuff like this [out of memory error](https://github.com/reactioncommerce/reaction/issues/3023), so I'm integrating a few other CI platforms aside from circleCI

You should be able to push this fork into a gitlab instance (*that has a CI runner enabled), and it will attempt to build reaction as closely as possible to how the circleCI method does.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,9 +1,9 @@
 image: node:7
 
 variables:
-  DOCKER_VERSION: 17.03.1-ce
-  DOCKER_COMPOSE_VERSION: 1.13.0
-  METEOR_ALLOW_SUPERUSER: true
+  - DOCKER_VERSION: 17.03.1-ce
+  - DOCKER_COMPOSE_VERSION: 1.13.0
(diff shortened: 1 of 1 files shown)


joshuacox:
of note this has gone through a few iterations already and should be considered a WIP!  at this point in time it dies on jest:

https://gitlab.webhosting.coop/reactioncommerce/reactiongitlabtest/-/jobs/160

and unfortunately I must travel west, and will most likely not get back to this a few hours while on the road.  I welcome comments.

aaronjudd:
@joshuacox see (recently added): https://github.com/reactioncommerce/reaction/blob/master/.reaction/ci/config.yml

This is the gitlab CI file we're using for our own implementation, but we can testify we're regulary building with GitLab.

The jest errors made me think maybe you are installing node modules with the wrong version of npm (meteor npm vs npm?)...  those are dev dependencies though, so I wouldn't think those should be required when building.

PR #3035, opened 2017-10-04
Enable anonymous users to checkout with Stripe Connect

### LINKED ISSUE #2832: Handle anonymous users with Stripe Connect
@spencern has noted this a couple of places but it needs to get resolved somehow before Marketplace is launched.

With Stripe Connect users **must** have an email. Right now if you try to check out as anonymous you just receive the error `[Customer does not have default email]` when trying to checkout.

Probably we just need to not allow anonymous checkout in a Stripe Connect scenario.

kieha:
Resolves #2832 

### How to test
- As marketplace admin set Stripe Connect as payment method
- Load Reaction in a new incognito window. Make sure this is a new incognito session
- Place an item in the cart, and proceed to checkout
- Click on `Continue As Guest` button
- Enter an email address in the form
- Continue with checkout steps. Observe that the checkout process is completed and you get the `orders/completed` page.


machikoyasuda:
great jsdocs ✅ 🎊 

just added some notes on `@method` and copy casing.

imports/plugins/core/accounts/client/components/loginInline.js
```diff
@@ -0,0 +1,98 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      email: ""
+    };
+  }
+
+  /**
+   * @summary handle setting state whenever the field on the form change
```
machikoyasuda: Start with `@method handleFieldChange`

example: https://github.com/reactioncommerce/reaction-jsdoc/blob/master/README.md#documenting-a-function

imports/plugins/core/accounts/client/components/loginInline.js
```diff
@@ -0,0 +1,98 @@
...
+              <Components.TextField
+                name="email"
+                type="email"
+                tabIndex="1"
+                value={this.state.email}
+                onChange={this.handleFieldChange}
+              />
+              <Components.Button
+                type="submit"
+                label="Add Email"
```
machikoyasuda: We are sentence-casing buttons, too. So: `Add email`, `Continue as guest`

server/methods/core/cart.js
```diff
@@ -1011,5 +1011,31 @@ Meteor.methods({
     }
 
     return Collections.Cart.findOne(selector);
+  },
+
+  /**
+   * cart/setAnonymousUserEmail
```
machikoyasuda: `@method cart/setAnonymousUserEmail`

server/methods/core/cart.js
```diff
@@ -1011,5 +1011,31 @@ Meteor.methods({
...
+
+    if (!currentUser.emails.length) {
+      newEmail = {
+        address: email,
+        provides: "default",
+        verified: false
+      };
+    }
+
+    return Collections.Accounts.update(userId, { $push: { emails: newEmail } });
```
spencern: This method of solving this problem opens the possibility of having several `Accounts` documents that all have the same default email address. We really need to solve this by attaching the email address to the `cart` and not by attaching an email address to an anonymous user account.

See how it's done in `orders/AddOrderEmail` for an example.
https://github.com/reactioncommerce/reaction/blob/kieha-stripe-anon-checkout-2832/server/methods/core/orders.js#L797-L824

As you can see in the [cart schema](https://github.com/reactioncommerce/reaction/blob/kieha-stripe-anon-checkout-2832/lib/collections/schemas/cart.js#L126) cart schema already supports orders.

joykare:
It works as advertised and the code looks :ok_hand:

Just one qn: What do you mean  `I reused the email form code from the orders/completed page`

imports/plugins/core/accounts/client/templates/inline/inline.js
```diff
@@ -1,18 +1,14 @@
...
-  /**
-   * Continue as guest.
-   * @param  {Event} event - jQuery Event
-   * @return {void}
-   */
-  "click .continue-guest": (event) => {
-    event.preventDefault();
-    Meteor.call("workflow/pushCartWorkflow", "coreCartWorkflow", "checkoutLogin");
+Template.loginInline.helpers({
+  component() {
```
zenweasel: Maybe we could make this name a little less generic?

kieha:
A bit of a loophole I’ve found with the new approach of attaching the email to the cart: when checking out the first time as an anonymous user, you get the form to set the email, and that works okay. 

But when checking out the second time with the same anonymous user in the same session, since the `checkoutLogin` and `checkoutAddressBook` steps were already completed the first time, you don’t get the form, and the email is set to `undefined` here https://github.com/reactioncommerce/reaction/pull/3035/files#diff-2941bb8c7406a5744bd46716065eb1b4R253. 

But the checkout process is complete eventually because `customerEmail` is set to a value albeit `undefined`, and on `order/completed`, the email form shows up there.

spencern:
Good catch on the loophole here.
A couple options I can think of:
1. We could copy the email address to the newly created cart?
 - The major downfall of this option that I can think of is that now their email is attached to the cart and since they aren't logged in, I'm not sure if they can remove it without creating an account.

2. We could reset the checkout workflow to the beginning for an anonymous user when they complete checkout.
  - The downside to this option is that if a user decides to complete another transaction, they have to start over, but the more I think about it, that seems like it should actually be the expected behavior because they are _not_ opting in to having their email saved by creating an account.

kieha:
Update: went with the resetting cart workflow for anonymous users.
> We could reset the checkout workflow to the beginning for an anonymous user when they complete checkout.

However, this then becomes the default behavior for all anonymous users with any payment method. Not sure if this is the intended approach for all the other payment methods.

kieha:
@spencern this is ready for another look.

aaronjudd:
re: the loophole -> this was part of what the older `customer`, `guest`, and `anonymous` user roles were implemented to signify user states:

I've been seeing these move/changed/ but for clarity, there are three states of user/account (I think):

`anonymous` - no email , no password
`guest` - supplied email, no password
`customer` - supplied email, password

I suspect one of the customer login state issues are because the handling for these states have changed and aren't being consistently used.

spencern:
@aaronjudd I think we should audit this behavior as I'm also not sure that these states are being respected consistently.

@kieha I think this is the correct behavior for anonymous accounts. I'll review asap.

PR #3040, opened 2017-10-05
Fixed archived variants being also cloned when top level variant is cloned

### LINKED ISSUE #2814: Duplicated products show archived variants
## Expected behavior

When I duplicate a product from a product that has archived variants options, I shouldn't see the archived variant options on the duplicate. I should see the variant options with their duplicated option title and title. 

### Actual Behavior

When I duplicate a product it shows me the variant options that have been archived from the product I duplicated it from. Plus, the option title, title, and price are missing.  

1. Create a product
1. Add a variant option
1. Add a second variant option
1. Archive and publish the second variant option
1. Duplicate the top-level variant 
1. Observe that all the variant options (including the previously archived ones) show up and they don't  copy the option title, title, and price.

aanchirinah:
Resolves #2814 

I noticed what was causing this bug was as a result of the variants collection being fetched for the Product to be cloned didn't have a constraint to check whether the variant has been archived or not. So I added a constraint to check that ([find it here](https://github.com/reactioncommerce/reaction/blob/audax-fix-issue-2814/server/methods/catalog.js#L354)). I also added an `optionTitle` parameter with `- copy` appended to it to the `Object.assign` function to handle having "Untitled option" in the text fields after duplicating.

#### How to test
- Create a product
- Add a variant option
- Add a second variant option
- Archive and publish the second variant option
- Duplicate the top-level variant
- You would observe there are no archived variant options among the duplicated variants and option title for cloned variants is also present now.


aanchirinah:
@rymorgan and @kieckhafer can you guys please test this PR out to see if it fixes the bugs you both experienced

aanchirinah:
@kieckhafer , I have fixed the requested change.

rymorgan:
When you clone a variant option, we don't need to append the ' - copy' language to the option title. This make it harder for a user to quickly propagate variant options. (We should, however, keep the behavior for the top level variant.)

lunar_force

rymorgan:
Also, it should copy the price from the variant option vs the price from the top level variant. 

prices

aanchirinah:
@rymorgan , implementing your requested changes , sorry for the late response.

aanchirinah:
@rymorgan , so I have removed the `- copy` appending for variant option title and also resolved the price of cloned variants not having the top level variant price.

rymorgan:
@aanchirinah yep, works as I would expect. 

@spencern Will you check this out from a UX perspective? This issue is about cloning a variant and it's options properly. So this PR will make it so when you create a variant with options, and clone that variant it will: 

1. Will keep any variant options archived on clone
2. Will append -copy to the top level variant but not to it's options
3. Will keep the prices from variant and it's options 

My UX thinking here is that this makes it the easiest for a user to create the exact top-level variant they want say a color as the top level variant and sizes as the options. Then when they have all the sizes as the options they clone that top-level variant, create a new color and all the options are ready to go.

spencern:
@aanchirinah we've got failing tests here.

aanchirinah:
@spencern , When I run `reaction test` of this branch  locally on my Mac it passes all tests and the error reading on CircleCI can be seen below ,
screen shot 2017-10-06 at 8 02 31 pm

spencern:
@kieckhafer are you gonna have a chance to give this a final review today?

kieckhafer:
Yes, sorry, didn't realize the `[WIP]` had been removed.

kieckhafer:
We want the word "Copy" to show only on the top-level variant, not on any of the options.

new_product

The price also doesn't seem to be copying to the duplicated product as it should.

aanchirinah:
@kieckhafer , I have added reinforcements to make sure price is displayed and the `- copy` is not displayed on options when Top-level variant is duplicated.

kieckhafer:
@aanchirinah Price and copy are there, but still missing the Width / length / height / weight. They are still being copied from the top level, not the option variant.

aanchirinah:
@kieckhafer can you elaborate more what is being copied from the top-level variant instead of the option variant?

kieckhafer:
The dimensions are being copied from the top-level variant to the option variant. They should be copied from the option variant to the new option variant that's the copy.
screenshot_10_9_17__10_46_am

aanchirinah:
@kieckhafer , reinforced the acquisition of the height/width/weight/length parameters for cloned variants in my recent commit.

zenweasel:
@aanchirinah Looks like "MSRP/Compare At Price: is still not being copied over. Quantity is also being copied over though I am not certain it should.

aanchirinah:
@zenweasel , I have reinforced the copy of MSRP and also added a `delete clone.lowInventoryWarningThreshold` to prevent quantity from also being copied.

zenweasel:
This is working for me. Ready to merge once we have approval from @rymorgan and @kieckhafer (and 1.5.0 is merged)

PR #3041, opened 2017-10-05
Enable discounts for single shop checkout when using a multi-shop enabled payment provider (e.g Stripe)

### LINKED ISSUE #3005: Stripe should work with discounts for single shop
See #2773 and #2763 

It appears that a bug was introduced during the Stripe Connect build out that prevents discounts working even when in a single shop instance.

impactmass:
Resolves #3005 

**Added / Changed:** 
- Updated `getTotalByShop()` to subtract discount from shopTotal (if only one shop is participating in the order)
- Updated the `discounts/codes/apply` method to check that discounts are only applied to single shop carts
- Updated the Add Discount UI to show error alert to user when trying to add discount on a multi-shop checkout
- Added two extra tests for discounts/codes/apply method

## How To Test:
- Create a discount code (under Dashboard > Payment)
- While still under Payment, enable Stripe as payment provider
- Add an item to cart and proceed to checkout. Apply the discount code created earlier, then add card details and pay.
- Confirm that the discount is applied to the order on both the "Thank you page" and in the Order Dashboard

Second test
- Add two items from separate shops into a cart. (you'll need to make a second shop for this)
- Proceed to checkout and try to apply a discount code
- Confirm that the code is not applied and an error message to explain why is shown


imports/plugins/included/discount-codes/server/methods/methods.js
```diff
@@ -193,6 +193,24 @@ export const methods = {
...
+    const uniqueShopObj = items.reduce((shopObj, item) => {
+      if (!shopObj[item.shopId]) {
+        shopObj[item.shopId] = true;
+      }
+      return shopObj;
+    }, {});
+    const participatingShops = Object.keys(uniqueShopObj);
+
+    if (participatingShops.length > 1) {
+      throw new Meteor.Error(400, "discounts.multiShopError", "Discounts cannot be applied to a multi-shop cart or order");
```
zenweasel: We don't pass a number to `Meteor.Error` any more. I think we should probably pass `not-implemented` there as I have been using that whenever there is a marketplace feature that's not yet done.

impactmass: Cool. Adding that now

spencern:
Works well for checking out from single shop. Tested both credit and discount modes successfully.

Issue #2878, opened 2017-09-15
Sortable table extends awkwardly


rymorgan:
### Expected behavior
Sortable table in orders admin panel should expand remain contained and not extend outside of the panel. It should all stay inside the table width. We can allow them to move the info inside the table to see a column that has extra data. In that case, we can let the other column hide beyond the edge of the table width. 

### Actual Behavior
It expands outside of the panel. 

![table-extending](https://user-images.githubusercontent.com/20409254/30503200-3fc94470-9a1e-11e7-9fbf-53b12adb1c21.gif)

### Steps to Reproduce the Behavior
Place some orders
Grab and expand one of the column headers 
See that it expands outside of the width of the table. 

### Versions
Node: 4.5.0
NPM: 5.4.1
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace

efalayi:
@rymorgan when a column header is resized, where should the additional column be displayed? I'm thinking we'll probably need to listen for a drag event, then hide some stuff, and display the new column when the event occurs. CSS might not be our only fix here.

rymorgan:
@efalayi, In this case, I don't think you need to hide anything. 

When expanded, the sortable table in orders admin panel should remain contained and not extend outside of the panel. It should all stay inside the table width. We can allow them to move the info inside the table to see a column that has extra data. In that case, we can let the other column hide beyond the edge of the table width. I would squish down the columns as it's expanded but limit it a size where you can still grab a column. 

If you are the user and you are expanding the table, that means you are likely doing something like, looking at someone's full email address or name. I want it to remain contained though, so it doesn't look broken. 

@kieckhafer Can you give some thoughts/feedback on this issue?

kieckhafer:
`we can let the other column hide beyond the edge of the table width`

If you mean allow the table to be horizontally scrollable, then yes, this is the way to go. SortableTable should already be set up to do that.

Take a look at the example here, you can see that the table is scrollable, without the whole window being scrollable.

https://react-table.js.org/#/story/custom-column-widths

rymorgan:
Good example. That's what I mean, but we should put a min-width on them, maybe of 20px so you can't lose columns entirely off the page.

efalayi:
@kieckhafer thanks for that helpful link. I've been trying to figure out how react-table works. @rymorgan I'll give you an update in an hour.

[...]

PR #3044, opened 2017-10-05
[WIP] Fix "Sortable table extends awkwardly" bug

efalayi:
Resolves [#2878](https://github.com/reactioncommerce/reaction/issues/2878)
Table column expands outside of the panel.

Some changes were made in -https://github.com/reactioncommerce/reaction/commit/0c465809a27bae8b5dcbffefa60c2033b12e842a#diff-1462e7801c36243a6542abe7414fe30bL36
-https://github.com/reactioncommerce/reaction/commit/0c465809a27bae8b5dcbffefa60c2033b12e842a#diff-1462e7801c36243a6542abe7414fe30bL163. These are the default style of `sortableTable` component.

###### Fix
```
- Add overflow property to `rt-table` class
- Fix border styles
```

###### Test
```
Place some orders
Login in as admin
Go to orders
Grab and expand one of the column headers
See that column does not extend out of the table.
```


efalayi:
@kieckhafer this is not a permanent fix. What do you think about the approach? I could create a function that handles setting min widths for the different columns.

rymorgan:
Yea, it would great if this fix would work across all the react tables, I'm guessing this issue happens everywhere this table is used. 

pepke6cbyb

efalayi:
@rymorgan at the moment, i'm implementing this fix for Orders table. We might have to create a separate issue to accommodate changes to other the instances of sortableTable.

spencern:
@efalayi Can you explain why the fix only works for one table and not all sortableTable components? If this PR is hacky and there's a "right" way to do it, I'd much prefer we just do it right the first time.

rymorgan:
Yea, I am hoping we are building it modular enough to work elsewhere.

efalayi:
@spencern I observed that min and max widths were specified for columns in Orders table. This fix handled it. But we can just specify the min and max widths in `sortableTable` (the main component which renders react table in RC) directly. Should I do that in this PR?

spencern:
> @kieckhafer this is not a permanent fix. What do you think about the approach? I could create a function that handles setting min widths for the different columns.

I don't think we should be doing temporary fixes. If you're not sure what to do, please reach out to someone for help.

rymorgan:
The current change isn't fixing the issue. 

ljnujmcv4p

efalayi:
@rymorgan just observed that happens after a certain width. I'll look into it.

spencern:
@efalayi any updates on this? I'd like to get this merged today.

imports/plugins/core/orders/client/components/orderTable.js
```diff
@@ -371,7 +372,7 @@ class OrderTable extends Component {
           />
         }
         <SortableTable
-          tableClassName={`rui order table -highlight ${this.props.selectedItems.length > 0 ?
+          tableClassName={`rui order table -striped -highlight ${this.props.selectedItems.length > 0 ?
```
mikemurray: Whats the purpose of the `-`(dash) in front of the class names for `striped` and `highlight`. Seems unnecessary to me? @kieckhafer thoughts?

kieckhafer: That's just how ReactTable does it, it's not on our end.

efalayi:
@spencern changes I've made don't seem to override the styles set. I'm still working on it.

joykare:
@efalayi adding https://github.com/reactioncommerce/reaction/commit/0c465809a27bae8b5dcbffefa60c2033b12e842a#diff-1462e7801c36243a6542abe7414fe30bL36
https://github.com/reactioncommerce/reaction/commit/0c465809a27bae8b5dcbffefa60c2033b12e842a#diff-1462e7801c36243a6542abe7414fe30bL163
those two lines back fixes the issue. I'm not sure whether it breaks anything though. @foladipo any input?

foladipo:
@joykare  @efalayi I tested this and it currently works fine for me. That is, when I expand any of the columns, the overall table becomes horizontally scrollable to accomodate the extra width, which I think is acceptable behaviour.

The only problem I found is that the horizontal dividers in the table don't expand with it. See SS below for details:
screen shot 2017-10-12 at 11 58 22 am-highlighted

rymorgan:
I'm looking at this PR and it's still not working the intended way. I think there is some confusion about how this should work.

I want the columns to never move outside the table, but rather condense down to a min width of around 20pixels.

So you can expand the columns to see hidden information w/out losing the rest of the columns.

Here is what I'm seeing now. 
qgmbvmukfz

spencern:
@efalayi why was this PR closed?

efalayi:
@spencern I made an error. I'll reopen it right away.

spencern:
Hi @efalayi 
Closing a PR that isn't ready yet isn't a problem as long as you leave a note explaining why it was closed and what steps you'll be taking to get it ready to open again.

zenweasel:
@efalayi Please update this with status and next steps.

efalayi:
@zenweasel at the moment I've reverted to previous styles implemented on the table. Next step is find a way to improve `SortableTable` component to use different styles.

efalayi:
@joykare I've reverted to the old styles on this table. We might have to implement a different style of `sortableTable` component for ordersTable. What do you think?

rymorgan:
@kieckhafer @joykare It doesn't seem like we need a whole new table to make the changes I requested. Can y'all help @efalayi out or let me know that my request is harder than I think to implement.

spencern:
To echo @rymorgan. Let's either get this resolved and merged or close it and let someone else work on it. This PR has been [WIP] for a long time.

efalayi:
@spencern this issue will resolved before the end of the week. I got caught with another issue.  @joykare @kieckhafer could we have a sync on this issue?

spencern:
@joykare I think you may be best suited to pair with @efalayi as @kieckhafer has a pretty long list of things to get done right now.

efalayi:
I read the source code of `react-table` package and I discovered that:
1. `min-width` can be hard-coded.
1. `max-width` is dynamically set when a resize function is called.
1. passing `maxWidth` and `minWidth` as props doesn't override the previous css styles.

Looking at the table for the orders dashboard, I noticed that some columns do not really need to be resizable. Since the resize function can be disabled in `react-table`, what if we disabled the resizing of some (or all) columns and then display extra text with a tooltip? @joykare  @kieckhafer @spencern what do you think?

kieckhafer:
What kind of extra text would you be displaying with a tooltip? How would that work?

I think a big issue here is screen-size, the columns are being hidden / funky on smaller screens. You can't use tooltips on mobile, at least not to be the only display of information.

rymorgan:
@efalayi I don't think tooltip is a good solution here. It'll be a pain for people to use and won't work on mobile. @joykare was working on a solution the rows were horizontally scrollable. Which could be the best approach and would work on mobile. We'd have to fix the first column and last columns so that they aren't scrollable and the rest of the columns scroll underneath them in both directions. @joykare Should be able to help you as she had this partially working when she was in Los Angeles.

joykare:
@rymorgan @efalayi I was working on this got caught up in the foray sprint and travelling. But I understand what @rymorgan wants so we can sync on this @efalayi.

efalayi:
@rymorgan @kieckhafer you're right. Tooltip wouldn't work for small screens. @joykare when would you like us to have a sync on this?

efalayi:
@joykare and I will have a sync on this today. Updates will be given later today.

efalayi:
### Update
@joykare  and I concluded that the first and last columns would be fixed (not resizable) but every other column in between them would be resizable. This would be implemented for all instances of `sortableTable`.

rymorgan:
@efalayi I'm not sure we want to implement this in every react table we have. We'd definitely need to look at that before rolling it out. I'd rather just apply this to this one table.

efalayi:
@rymorgan @joykare I updated this PR some days ago. Could you check it out to see the changes I've made.

rymorgan:
@efalayi Did you get together with @joykare on this issue? I'm still not happy with this solution. Especially on mobile. You can currently expand the table out and not be able pull the fields back after you expand them. This will amount to a frustrating UX. What happened the idea we discussed with @joykare with fixing the first and last columns and allowing the middle columns to be scrollable? 

A few things to note on this: 
1. We can lose columns completely once they are expanded. 
2. The button to process the orders should be fixed b/c the most important action a user will take on this page is to hit that button and process an order. 
3. We can't lose name associated with the order b/c they will be using that find orders. 

It seems like we are well past the time we should scratch this PR and reapproach with a clear solution.

rymorgan:
The last column is still not fixed in place nor are the rows scrollable. 

qdgqxy6jwd

Issue #2864, opened 2017-09-14
A way to archive and disable a shop


cindyfirestone:
As the marketplace owner, I should have the ability to: 
1. Archive or disable a shop - which includes removing permissions/access for all users of that shop
2. Archive or un-publish all of the products from that shop (make it invisible) 
3. Remove that shop from the navigation

aaronjudd:
This is a duplicate, extension, or just a more focused version of https://github.com/reactioncommerce/reaction/issues/1578 -> but that's important to understanding this issue.

aaronjudd:
@spencern we should have an architectural discussion about this. I'd suggest that we should use a `workflow` status, to accomodate: 'archived', 'visibility', and other possible publishing states for a shop at a global level.  We can add to the various publications (products) for filtering/validation of these, and also using permissions.

spencern:
@aaronjudd 
I agree that we should have various states for shops. We've discussed the desire to have visibility and publication status both be more flexible than booleans on Products and I don't see any reason not to work with that mindset for Shops as well.

Here are some initial thoughts I've got on a shops workflow

**Shop Workflow**
This may be different for merchant shops than for primary shops.
I see this as the approval and publication workflow for a "managed" merchant shop. My initial thoughts would have the workflow looking something like this:
**Shop Workflow**
  * **Pending** - a shop that has been created for a user, but the user hasn't accepted managing that shop is in this state
  * **New** - if users can create a shop without an invite, they start here, otherwise when they have accepted their invite, their shop is set to this state.
  * **Valid** - All essential information has been filled out and is in good standing (address, payment info, tax info, shipping rates, etc - anything that is required to be submitted has been submitted)
  * **Invalid** - Could be shared with "new" but might be good to have a way to distinguish shops that are in an invalid state because of neglect, vs those that haven't been set up yet.
  * **Active** - Shop is in a valid state and has been approved/enabled/published by the marketplace owner if there is a an approval queue
  * **Archived** - May need sub status or "reason" that could determine whether the user can re-open or if the marketplace must re-open - e.g. Closed, Paused, Banned, Suspended, 

A few questions:
1. Is there a need to have a visibility property in addition to these workflows? Should it be a part of the workflow or separate.
2. Are there use cases that are not covered by any of these workflows? Do we need an "inactive" status?
3. Some of these states are meant to be moved between while others are steps in a process, is there any good reason to separate those concepts?

[...]

PR #3049, opened 2017-10-06
Adds the ability to set shops to active / disabled

mikemurray:
Fixes items 2 and 3 for https://github.com/reactioncommerce/reaction/issues/2864

This PR adds a simple table to set marketplace shops to `active` / `disabled`

- As a customer
If a shop is disabled, its products are de-listed from the product grid if using the default `Products` publication.

- As an admin
You can see all shops and all products from all shops at the moment.

**This is the start of a full-on marketplace shop management tools.**


imports/plugins/included/marketplace/client/components/marketplaceShops.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  renderShopsTable() {
+    const fields = ["name", "emails", "workflow"];
+
+    const columnMetadata = fields.map((field) => {
+      return {
+        Header: <Components.Translation i18nKey={`marketplaceShops.headers.${field}`} defaultValue={field} />,
+        accessor: field,
+        Cell: (data) => ( // eslint-disable-line
```
spencern: What's disabled here? It's a little more clear if we can use `eslint-disable-line rule-name` so future us/others can understand why we disabled and what for

imports/plugins/included/marketplace/server/methods.js
```diff
@@ -1,23 +1,43 @@
...
+  SHOP_WORKFLOW_STATUS_ACTIVE,
+  SHOP_WORKFLOW_STATUS_DISABLED
+];
+
+export function marketplaceUpdateShopWorkflow(shopId, workflowStatus) {
+  check(shopId, String);
+  check(workflowStatus, String);
+
+  if (shopId === Reaction.getPrimaryShopId()) {
+    throw new Meteor.Error(403, "Cannot change shop status");
```
spencern: We're no longer using error codes here, this should be `Meteor.Error("access-denied", "Cannot change shop status");

imports/plugins/included/marketplace/client/components/marketplaceShops.js
spencern: I'm actually not seeing any lint errors when removing this, so perhaps not needed?

mikemurray: It's usually a react issue of defining more than one component per file.

mikemurray:
@spencern ninja fixed it before I could 👍

Issue #2935, opened 2017-09-23
New Chinese Locale


reidlai:
I am now looking for a way to implement Traditional Chinese instead of default Simplified Chinese.  Traditional Chinese is the standard Chinese variant of Taiwan and Hong Kong.  After my research of recent browser capabilities, RFC 4647 tag has already defined locale lookup priority as below:

1. zh-Hant-TW
2. zh-Hant
3. zh

My suggestion is that I'll create new PR which will rename default zh to zh-Hans and create additional language zh-Hant.

aaronjudd:
@reidlai 👍  I would have sworn we'd implemented zh as the default before, but supporting the language variants has always been "the next item todo".  Happy to have this functionality and welcome your contributions.

reidlai:
I nearly finished new zh-hans and zh-hant but I found that there is a minor problem.  Even changing language from English to Simplified Chinese or Traditional Chinese, the admin console still stays in English.  But there is no problem when changing to French or other default languages.  Did I miss any setting in mongo or mapping between browser locale and i18next?

spencern:
Hi @reidlai 
Have you tried resetting the app - `reaction reset`

Another option would be to run:
Try `Meteor.call("i18n/flushTranslations")` from your browser console.

aldeed:
There does not seem to be much point in tracking this request any longer, but if anyone needs this and would like to re-implement the `zh` split in the current Reaction release, please do so and submit a PR.

PR #3050, opened 2017-10-06
Reidlai fix issue 2935

reidlai:
change to use locale zh-Hant and zh-Hans instead of zh

resolves https://github.com/reactioncommerce/reaction/issues/2935


client/modules/core/helpers/templates.js
```diff
@@ -62,8 +62,9 @@ Template.registerHelper("monthOptions", function (showDefaultOption = true) {
   //   parentLocale: "zh-cn"
   // });
   let lang = i18next.language;
+  console.log(lang)
```
aaronjudd: You'll want to remove the console.log..

reidlai: console.log has been removed

spencern:
Hi @reidlai 

Would love to accept this PR, unfortunately there are some conflicts with the latest version of Master. If you would be able to resolve these conflicts, we'd be happy to merge this PR. 👍 👍

PR #3056, opened 2017-10-06
Fix console errors after placing an order for a single shop in multi-shop mode

### LINKED ISSUE #3024: Receive error in server console after placing an order for a single shop in multi-shop mode
### Expected behavior
No errors

### Actual Behavior
Receive this error in console:
```
Exception while invoking method 'orders/refunds/list' TypeError: Cannot read property 'paymentMethod' of undefined
    at [object Object].ordersRefundsList (server/methods/core/orders.js:953:27)
    at packages/check.js:128:16
    at packages/ddp-server/livedata_server.js:717:46
```

This seems to be because this function is returning the paymentMethod for the wrong shop (only the current shop)
https://github.com/reactioncommerce/reaction/blob/0059da1f76b8c51930eedc94a4fd4e9b0c1806ce/server/methods/core/orders.js#L18

1. Create a marketplace shop and create a product for that shop
1. Place only that product in the cart and checkout
1. Observe the error

### Versions
```
Node: 7.10.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.13.1
Reaction: 1.5.0
Reaction branch: brent-fix-issue-2836
Docker: 17.09.0-ce
```

kieha:
Resolves #3024

### How to test

- Create a marketplace shop and create a product for that shop
- In admin shop, place only that product in the cart and checkout
- Observe no server errors

### DIFF
--- a/imports/plugins/core/orders/client/templates/workflow/shippingTracking.js
+++ b/imports/plugins/core/orders/client/templates/workflow/shippingTracking.js
@@ -187,7 +187,7 @@ Template.coreOrderShippingTracking.helpers({
         view = true;
       }
       // TODO modularize tracking more, editable to settings
-      if (view && shipment.shipmentMethod.carrier === "Flat Rate") {
+      if (view && (shipment.shipmentMethod && shipment.shipmentMethod.carrier) === "Flat Rate") {
         return true;
       }
     }
--- a/server/methods/core/orders.js
+++ b/server/methods/core/orders.js
@@ -20,7 +20,7 @@ export function orderCreditMethod(order) {
   const billingRecord = creditBillingRecords.find((billing) => {
     return billing.shopId === Reaction.getShopId();
   });
-  return billingRecord;
+  return billingRecord || {};
 }
 // helper to return the order debit object
 export function orderDebitMethod(order) {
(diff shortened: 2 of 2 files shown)


zenweasel:
I think this is a complicated problem that I am not immediately sure what the solution is, because when we had only one payment we knew what to look for but when we have multiple shops we can't just use `getShopId` because the current shop might not be the shop that the charge was against. Theoretically we should be passing in `paymentMethod` to this like we do with `orders/refunds/create` and then grabbing the information based on that rather than just assuming we can get the payment method by current shop.

zenweasel:
Looking at the code this gets called a couple of times in two different contexts. In the context of the user it should cycle through all payments and return an array of refunds for all shops because a user doesn't care which shop, just wants the total (and that's what's throwing the error). The code is already expecting an array of refunds.

Working out how it will work in the order dashboard might be something we could figure out post-alpha release.

I do think we need to back out some of this code where we are failing silently.

zenweasel:
Ok, last thought. If we use the order passed in only (don't re-grab it from the DB), which should be filtered already, then we should just be able to always query for all the billing records on the order and return an array because orders from dashboard should *already* be filtered so that should fix both situations I think.

zenweasel:
Rather than modify @kieha's PR I opened a new PR with an alternative fix here: https://github.com/reactioncommerce/reaction/pull/3064

kieha:
Closing this in favor of #3064

PR #3058, opened 2017-10-07
Revert "changed user collection to Accounts instead"

spencern:
Reverts the change that switched from using Meteor.users to using Reaction Accounts for currency. The Accounts update caused the currency dropdown to stop working and display multiple currencies in the navbar.

@lcampanis I'd still like to update to use the Account instead of the Meteor user, but had to back this change out as it was breaking.


spencern:
Resolves issue seen here: 
image

spencern:
This should resolve #3060 once merged

lcampanis:
@spencern have a look at #3061 better

spencern:
Closed in favor of #3061

Issue #3060, opened 2017-10-07
Currency dropdown displaying multiple values and not working


spencern:
![image](https://user-images.githubusercontent.com/1203639/31305590-9edfd09a-aafa-11e7-8e22-0a89e8934180.png)

This was introduced when we merged #2979

kieckhafer:
This is how it's saved in the database, so it's not "wrong", per se. Perhaps we just need to split it and display USD only?

@mikemurray mentioned it's saved this way in the `Countries` collection, which then sends it to the shops.

![screenshot_10_9_17__10_24_am](https://user-images.githubusercontent.com/4482263/31350474-2aeb3fc4-acdc-11e7-91de-1ca59fc6da47.png)

spencern:
@kieckhafer thanks for digging this up. @lcampanis had mentioned similarly in his PR. If we have been splitting the currencies, a la `currency.split(",")[0]` up to this point, I guess it's not a regression to continue to do that, but we also should figure out how we'd like to display/save currencies for countries that have multiple currency codes. I'll create another issue for this particular task (which may or may not be important) after doing another review of #3061

aaronjudd:
`currency.split(",")[0]` is fine for now.  I wrote it this way because the vast majority doesn't use anything but that first indicator.. seems like an issue to tackle in future (and the future isn't now). Do we have a case where someone actually needs multiple currencies yet?

spencern:
Not that I'm aware of, should I just leave that issue off the table until someone requests it?

spencern:
Resolved by #3061

PR #3061, opened 2017-10-07
Save user currency server-side in Accounts.profile.currency based on locale or default to shop's currency

lcampanis:
@spencern Needed to add currency in the profile schema for it to work correctly.

However, since this is now saved in mongo there's no reason to process it on the client side, so moved the logic server side. This should solve the issue of multiple currencies in the dropdown, as well as not saving the chosen currency.

Let me know if any issues.


spencern:
@lcampanis thanks for jumping right on this! I'll review it now

lcampanis:
@spencern Impressive, I didn't come across any of these. Looking into it now.

lcampanis:
@spencern I have a fresh install, new modules, clean mongo and wasn't able to see any issues. The line with the error tells me that you have a currency that doesn't exist in the shop at all, more so enabled (ie. an array or something else). I also tested in incognito just in case, but no issues there.

To satisfy this case which I'm not able to replicate, I have added an extra check in my last commit, so that if the currency doesn't exist, it will at least fallback to the shop's currency.

Let me know if I'm missing something and how I can replicate this.

spencern:
@lcampanis I'll see if I can reproduce on a clean install. It may have been a migration issue if didn't fall back to the shop's currency by default.

spencern:
@lcampanis I was able to reproduce both the server error and the missing prices after a full restart before applying the current patch suggesting that we may need a migration and some additional default data.

After applying the patch, the prices appeared, and now I'm doing a full reset with the patch applied.

spencern:
After a quick chat with @lcampanis it seems we're not able to reproduce and git similar results. @kieckhafer would you mind booting this up and seeing if you can reproduce

kieckhafer:
I am seeing the issue. On first load after reset, as an anon user, I see the correct currency.

When I log in as an admin, I see the issue:
reaction

If I then set my currency when I'm logged in, and then log out, I see the issue when I am logged out: 
reaction

But when I log back in to the same account I just set the currency for, the currency is still set as it should be. If I log out again, I still see the above issue.

lcampanis:
@spencern @kieckhafer Thanks for the feedback. 
No matter how hard I try I just don't get the error as I get one currency on UK (GBP). I think though that if you enter from the US, you get those 3 US currencies. I can't check easily cause Hola doesn't work on localhost. I added the `split(",")[0]` on shop.currency as we used to have before, so that will ensure that you only get USD from that list. 
If you don't mind trying once more. Thanks!

spencern:
@lcampanis that's an interesting observation re: GBP vs US currencies.
I was not aware that USN and USS were currency codes for the US until now, so I've learned something:

> The US dollar actually has three codes assigned: USD, USS (same day) and USN (next day). According to UN/CEFACT recommendation 9, paragraphs 8-9 (ECE/TRADE/203, 1996)

I don't love the '0th index' solution as it seems like it could generate the incorrect currency code for a different locale if the correct code was not the first item in the array, I think we really need to figure out why this is happening now and fix the root of the problem.

lcampanis:
@spencern With the last fix, we are where we were before in terms of UI, the split is currently in production. I am not sure why the US has three currencies in a string. 
image
But it should be an internal RC decision as to how to treat this. I personally don't think anyone knows what USN and USS currencies are, so I'd remove it.

spencern:
Doing a quick search reveals that there are 12 different countries that we have multiple currencies listed for.

image

spencern:
@lcampanis even with the latest code including the split for currency, I'm still seeing some issues with the display of the currency.

After signing in and out a time or two, I'm seeing these errors in the console and the sign-in is hanging and requires a browser refresh, suggesting that maybe there's a missing guard (or two) in the currency check. I've done a complete reset and restart multiple times as well as logging in from an incognito window so as to make sure it's a clean session.

here's a video: http://recordit.co/G0j5XuTGkv

I'm working through reviewing a couple other PRs right now, but I'll jump on this issue or pair with anyone who wants to try to identify the problem here today if we can't figure it out soon.

Aaron's feedback in #3060 is confirmation that we're still good to use the `split(",")[0]` solution, but I would like to eliminate these random occurrences of showing the entire currency string for countries like the US that have multiple currencies as well as to fix these console errors.

```
Exception from Tracker recompute function:
meteor.js?hash=6d285d84547b3dad9717a7c89c664b61b45ea3d8:942 TypeError: Cannot read property 'profile' of undefined
    at composer (currencyDropdown.js:51)
    at tracker.js:19
    at Tracker.Computation._compute (tracker.js?hash=997515fa2d5b0530ba07741da556c4b36963ef3b:339)
    at new Tracker.Computation (tracker.js?hash=997515fa2d5b0530ba07741da556c4b36963ef3b:229)
    at Object.Tracker.autorun (tracker.js?hash=997515fa2d5b0530ba07741da556c4b36963ef3b:613)
    at tracker.js:17
    at Object.Tracker.nonreactive (tracker.js?hash=997515fa2d5b0530ba07741da556c4b36963ef3b:640)
    at tracker.js:16
    at Container._subscribe (compose.js:85)
    at Container.componentWillReceiveProps (compose.js:31)
meteor.js?hash=6d285d84547b3dad9717a7c89c664b61b45ea3d8:942 Exception from Tracker recompute function:
meteor.js?hash=6d285d84547b3dad9717a7c89c664b61b45ea3d8:942 TypeError: Cannot read property 'profile' of undefined
    at findCurrency (currency.js:27)
    at _formatPriceString (currency.js:73)
    at ProductGridItems.renderGridContent (productGridItems.js:103)
    at ProductGridItems.render (productGridItems.js:144)
    at modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108665
    at measureLifeCyclePerf (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:107945)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108664)
    at ReactCompositeComponentWrapper._renderValidatedComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108691)
    at ReactCompositeComponentWrapper.performInitialMount (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108231)
    at ReactCompositeComponentWrapper.mountComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108127)
meteor.js?hash=6d285d84547b3dad9717a7c89c664b61b45ea3d8:942 Exception from Tracker recompute function:
meteor.js?hash=6d285d84547b3dad9717a7c89c664b61b45ea3d8:942 TypeError: Cannot read property '_currentElement' of null
    at ReactCompositeComponentWrapper._updateRenderedComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108614)
    at ReactCompositeComponentWrapper._performComponentUpdate (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108593)
    at ReactCompositeComponentWrapper.updateComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108514)
    at ReactCompositeComponentWrapper.receiveComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108416)
    at Object.receiveComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:100967)
    at ReactCompositeComponentWrapper._updateRenderedComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108623)
    at ReactCompositeComponentWrapper._performComponentUpdate (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108593)
    at ReactCompositeComponentWrapper.updateComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108514)
    at ReactCompositeComponentWrapper.receiveComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108416)
    at Object.receiveComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:100967)
meteor.js?hash=6d285d84547b3dad9717a7c89c664b61b45ea3d8:942 Exception from Tracker recompute function:
meteor.js?hash=6d285d84547b3dad9717a7c89c664b61b45ea3d8:942 TypeError: Cannot read property '_currentElement' of null
    at ReactCompositeComponentWrapper._updateRenderedComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108614)
    at ReactCompositeComponentWrapper._performComponentUpdate (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108593)
    at ReactCompositeComponentWrapper.updateComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108514)
    at ReactCompositeComponentWrapper.receiveComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108416)
    at Object.receiveComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:100967)
    at ReactCompositeComponentWrapper._updateRenderedComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108623)
    at ReactCompositeComponentWrapper._performComponentUpdate (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108593)
    at ReactCompositeComponentWrapper.updateComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108514)
    at ReactCompositeComponentWrapper.receiveComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:108416)
    at Object.receiveComponent (modules.js?hash=6659d28924f4a2c3173a0b0af5eac2e050ff7178:100967)
```



```
Uncaught TypeError: Cannot read property 'trim' of undefined
    at AuthContainer._this.handleFormSubmit (app.js:47538)
    at SignIn._this.handleSubmit (app.js:45963)
    at Object.ReactErrorUtils.js.ReactErrorUtils.invokeGuardedCallback (modules.js:99347)
    at executeDispatch (modules.js:99128)
    at Object.executeDispatchesInOrder (modules.js:99151)
    at executeDispatchesAndRelease (modules.js:98541)
    at executeDispatchesAndReleaseTopLevel (modules.js:98552)
    at Array.forEach (<anonymous>)
    at forEachAccumulated (modules.js:99454)
    at Object.processEventQueue (modules.js:98752)
```

PR #3066, opened 2017-10-09
Fixed other admin users not being able to add/edit/remove product tag

### LINKED ISSUE #2426: Non-admin user not able to add/edit/remove product tag in Product details
### Expected behavior
Non-admin user who can manage the product should be able to add/edit/remove product Tags in Product details page

### Actual Behavior
Non-admin users are not able to manage(add/edit/remove) the Product Tags for new or existing products

1) Clone https://github.com/reactioncommerce/reaction-example-plugin.git
2) Create new Account and provide all permissions to that account
3) Verify Email for new account
4) Login with new account
5) Edit existing product
5) Click Tags on right side panel
6) Only the Tags are read-only and not able to add/edit/remove any tags.

### Versions
$ reaction -v

Node: 4.4.7
NPM: 2.15.8
Docker: 1.12.6
Reaction CLI: 0.7.9
Reaction: 1.2.0

aanchirinah:
Resolves #2426 ,

New users with some administrative permissions couldn't access `TagList` component on the PDP because an `editable` prop that checks whether a user is a customer was not passed to the component . So I added the missing prop ie.`editable`' in the `TagList` component being used in product settings view.

#### How to test
- Sign in as admin.
- Invite a new shop manager or admin.
- Open a private tab in your browser and follow the link in the invitation mail
- Once you are signed in , open a product and view it's PDP panel.
- Notice that on the Tags Card now you can add/edit/remove a tag.


aanchirinah:
@zenweasel , Is my solution acceptable?

zenweasel:
@mikemurray @impactmass I feel like there is an intent there for passing in `props.permission` that I don't want to just undo. Does anybody understand the context of what this `permission` value is supposed to be?

impactmass:
Yes, if I remember correctly, the `isEditable` prop is based on the "Edit mode" switch at the top left (which controls general product management). Meanwhile the `permission` prop is used specifically to check if the user has access to create tags or not. 

I think the two work together, and outrightly removing one would cause issues. @aanchirinah you should test with Edit mode turned on and off (with different user roles).

aanchirinah:
@impactmass , I noticed `props.editable` and `props.permission` all are `undefined` when provided to the `TagList` component. So I added a permissions array to `Reaction.hasPermission` function to fix the issue now.

aanchirinah:
After pairing with @impactmass , we noticed the `Taglist` component for product settings was not passed any prop for `editable` , that was resulting the `props.editable` being undefined.

zenweasel:
Looks like we are getting an unwanted side-effect where I am always in edit mode for tags
**On Master**

reaction

**On this branch**
reaction

efalayi:
I pulled this branch and tested. `Edit` seems to work fine.

zenweasel:
Yeah, my mistake. I pulled the latest changes and everything seems to be working properly.

PR #3067, opened 2017-10-09
Fix overlapping badges on archived product variants

### LINKED ISSUE #2961: "Archived" badge overlapping.
### Expected behavior
The badges should not overlap.
### Actual Behavior
Archived badge is overlapping with the edit badge.
screen shot 2017-09-27 at 6 46 16 pm

Archive a product with sub-variant.
### Versions
```
Node: 8.5.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.12.0
Reaction: 1.5.0
Reaction branch: marketplace
```

### Solution
Instead of using "Achived" we should use a icon that means "Archived".

foladipo:
Resolves #2961 .

This PR fixes a UI bug that caused the "Archived" badge above the buttons for product variants to overlap other buttons.

## Test Instructions
- Log in as admin and navigate to the PDP page of one of your products.
- Archive one of the variant options of the product you're viewing.
- Close the side panel you used to archive the product variant(s). You should be on the main PDP page now.
- Open your browser's developer tools (use `Cmd + Alt + I` for a Mac).
- Reduce the screen width to <= 480px and >= 260px.
- Observe that the buttons for the product variant options become larger and are now arranged in a two-buttons-per-row grid.
- Observe that the `Archived` badge doesn't overlap the buttons for other variant options.

*Note*: this PR doesn't handle screens <= 260px wide. That's something for the future.


zenweasel:
@foladipo You requested review but this still says WIP. Is this ready to go?

foladipo:
@zenweasel Yes it is. Forgot to remove the WIP tag yesterday, but I have done so now.

zenweasel:
This does not appear to be working as described.
dumb_product

foladipo:
@zenweasel 
Status update: I've updated the test instructions. Please review.

rymorgan:
@foladipo While I appreciate the effort to try and solve this issue, I don't think this is the right solution. The archive icon inside the circle makes it look like an icon button (implying that it's actionable) but in fact, this is an indicator, not a button. I'd rather revisit this when we play around with moving away from all the inline editing and overall mobile experience, which are both on my UI/UX roadmap.

foladipo:
Thanks @rymorgan . While we prepare to do that migration in the future, is it possible/advisable to implement any interim solution? For example, part of the fix in this PR is that, on screens with a width of <= 480px, the buttons for the variant options take 50% of their container, as opposed to previously taking 33%. So can we have that or some other solution, or would it just be most efficient to wait?

PS: At the end of the day, we could have something like this:
screen shot 2017-10-11 at 9 29 16 pm

rymorgan:
@foladipo While the 50% spacing solves the one issue for viewing archived. It doesn't solve the issue for other languages that have longer words, nor does it make it a better user experience for customers b/c going 2 across instead of 3 make the page even longer. 

By the way, the archived label shouldn't be red. We can make it an almost black how about -  @black80

foladipo:
@rymorgan The backgrounds for the "Archived" badges have been changed. So do you think we are now set for merging?

rymorgan:
After a little more thought, the badge should be badge-warning -- we don't want to introduce further status colors unneccesarily.  

Also the sidebar status color should match the one in the page. 
air_max_and_developer_tools_-_http___localhost_3000_product_air-max_s8ew99terh8e2p8gr

foladipo:
@rymorgan I have two questions:
1. Using `badge-warning` will make the `Archived` badge look like the `Edited` badge to its right (with the pencil icon). Do you think I should use `badge-danger` instead?
1. Should I change the `Archived` badge's background on the PDP page alone or everywhere else too e.g on the product grid?

On my machine, I've already changed the badge to `badge-warning` everywhere there's an `Archived` badge. I'm only seeking these explanations before I commit and push them.

rymorgan:
@foladipo - 

1. Yes, use badge-danger instead of badge-warning
2. Yes, change it everywhere 

Thanks!

zenweasel:
@foladipo What's the status on this?

foladipo:
@zenweasel Yes it is, but I'll like to sync/update it with the base branch and test before you merge it. I'll do that as soon as I get back to my computer.

foladipo:
@zenweasel This PR's source branch has been updated and a new bug due to the update has been fixed. So everything is good to go!

zenweasel:
I have followed the test instructions as carefully as I can but it still doesn't appear that this is working properly.

basic_reaction_product_and_brenthoover

zenweasel:
@foladipo Can you put a comment on this clarifying what the current status is?

foladipo:
@zenweasel This PR is now ready for final review/testing.

zenweasel:
I guess this seems to be doing what it's supposed to in the testing instructions. @rymorgan Could you take a look at this one last time and double-check it's up to snuff? 👍 ?

rymorgan:
@zenweasel The desktop versions look ok. When I use the xcode simulator (in this case the iphone 8). The options are stacking instead of fitting across. 
iphone_8_-_ios_11_0_and_my_lastpass_vault

rymorgan:
here's some direction from @mikemurray 
screen shot 2017-10-20 at 11 20 24 am

zenweasel:
@foladipo What's the status on this?

imports/plugins/included/default-theme/client/styles/badge.less
```diff
@@ -63,7 +63,6 @@
   border: 1px solid @brand-warning;
 }
 
-
```
zenweasel: Let's not add/blank lines for no reason

foladipo: Ok. I just pushed a commit that fixes that.

zenweasel:
@foladipo Have you addressed Mike's and Ryan's comments above?

foladipo:
@zenweasel No, I haven't, but I will soon push a commit that does address those comments.

foladipo:
@zenweasel I have pushed some changes that addresses their concerns.

PS:
@mikemurray I did use `justify-content: space-between`. I also used `width: 47.5%` so as to fill the container as much as possible while minimizing the risk of overflowing it.

imports/plugins/included/default-theme/client/styles/badge.less
zenweasel: Still wasn't changed

spencern:
Bithound's integration stopped again, but the code was still checked, just need to follow the link.
No lint errors or other breaking bitHound code errors.

PR #3068, opened 2017-10-09
Fix language select issue

lcampanis:
**Expected behaviour** 
As a seller I can change language from the language select.

**Actual Behaviour**
As a seller I change language from the language select but nothing happens.

**Steps to Reproduce the Behaviour**

1. Enable the allowMerchantSignup in marketplace package
2. Register
3. Become a seller
4. Change language select

Credit @stevaras


zenweasel:
Ready to go just waiting until release is complete to merge.

PR #3074, opened 2017-10-10
Re-wrap in variant form in a blaze wrapper to prevent focus loss on tab

mikemurray:
Wraps the new React variant edit form in a blaze wrapper top prevent an issue with fields loosing focus when tabbing between them.

This is a temporary fix until we can discover the root of the issue.


kieckhafer:
Bithound hooks seem to be off, but on their site the files look good, so merging anyway: https://www.bithound.io/github/reactioncommerce/reaction/mm-fix-form-focus-loss

PR #3076, opened 2017-10-10
add error message for inactive shops trying to connect to Stripe

kieckhafer:
We currently do not show any error for `new` or `inactive` shops, when a user tries to click the "Start Accepting Payment" button on the Stripe connect settings. A user would click the button and nothing would happen.

This update adds an error message to direct the user to contact the marketplace admin.

I've also moved a few translations into this package, as they were in a different package and didn't seem to be working.

### DIFF
--- a/imports/plugins/included/marketplace/client/templates/stripeConnectSignupButton/stripeConnectSignupButton.js
+++ b/imports/plugins/included/marketplace/client/templates/stripeConnectSignupButton/stripeConnectSignupButton.js
@@ -29,6 +29,10 @@ Template.stripeConnectSignupButton.events({
 
     const shop = Shops.findOne({ _id: shopId });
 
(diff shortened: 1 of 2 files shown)


imports/plugins/included/marketplace/client/templates/stripeConnectSignupButton/stripeConnectSignupButton.js
```diff
@@ -29,6 +29,10 @@ Template.stripeConnectSignupButton.events({
 
     const shop = Shops.findOne({ _id: shopId });
 
+    if (!shop || !shop.workflow || shop.workflow.status !== "active") {
```
spencern: I think this is probably a good intermediate fix. Long run we should be building up the shops workflow and aligning UI with that so that buttons that are inactive are not even available on the UI.

PR #3084, opened 2017-10-11
Add jsdoc to CI build +  templates and configuration [#3094 #3095]

aaronjudd:
- Adds jsdoc and config as a part of CircleCI
- Adds new jsdoc validation rules to eslint
- Updates jsdoc site w/ custom Reaction style
- Fixes lots of jsdoc warnings
- TODO install npm or use local dev deps

- Resolves https://github.com/reactioncommerce/reaction/issues/3094
- Resolves https://github.com/reactioncommerce/reaction/issues/3095


spencern:
@aaronjudd did you want to get this PR into the `1.5.0` release?

aaronjudd:
We might want to ignore the eslint/jsdoc warnings this first pass unless we want to fix all the eslint warnings before merging this (but I think it's easier to ignore, merge, then begin linting all the docs).

I think, in theory, the CI changes I just added, should sync successful master builds to an S3 bucket now being served at http://api.docs.reactioncommerce.com/

The AWS credentials (for `api-docs-publisher`) added to CircleCI credentials.

@jshimko we may want to install `awscli` in the base instead.

.reaction/jsdoc/build.sh
```diff
@@ -0,0 +1,12 @@
...
+## master branch docs will be published via aws s3 sync when all test pass
+set -e
+
+## install awscli
+apt-get -y -qq install awscli
+## install jsdoc
+npm install -g jsdoc
+# build new jsdocs
+echo "Running jsdocs on the reaction codebase"
+jsdoc . --verbose --configure .reaction/jsdoc/jsdoc.json
```
machikoyasuda: do we want to use `node_modules/.bin/jsdoc` instead of `jsdoc` here?

jshimko: Nope.  Should be fine as it is since `jsdoc` is being installed globally on line 9.

package.json
```diff
@@ -131,6 +131,7 @@
     "eslint-plugin-react": "^7.1.0",
     "jest": "^21.2.1",
     "js-yaml": "^3.9.0",
+    "jsdoc": "^3.5.5",
```
jshimko: Since this is already installed globally, this can be removed.

jshimko: Actually, I take that back.  If our tests rely on jsdoc being there, we should probably leave that in `devDependencies` so users won't need the global install to be able to get tests to pass.  So ignore the previous comment.

machikoyasuda:
tested the `npm run docs` command locally and works like a charm.

.eslintrc
```diff
@@ -181,7 +181,20 @@
...
+        "MethodDefinition": false,
+        "ClassDeclaration": false,
+        "ArrowFunctionExpression": false
+        }
+    }], // https://eslint.org/docs/rules/require-jsdoc
+    "valid-jsdoc": [2, {
+      "prefer": {
+        "arg": "param",
+        "argument": "param"
+        }
```
machikoyasuda: Currently, adding the `require-jsdoc` as a `1` causes BitHound to reject the PR with 180 files with `warn` errors: https://www.bithound.io/github/reactioncommerce/reaction/jsdoc-c-integration-release-1.5.0/files 

Example: https://www.bithound.io/github/reactioncommerce/reaction/blob/627d214dd6dbbc27fc09431d45747058cc390333/imports/plugins/included/payments-stripe/client/checkout/stripe.js#filter-lint

When I run ESLint locally, I get: `✖ 410 problems (0 errors, 410 warnings)`

How do I add the `require-jsdoc` rule as an ignore-able rule for now?

machikoyasuda: @aaronjudd Should I just change the warn level to `0` for this rule?

PR #3111, opened 2017-10-13
This one weird trick will fix your Shopify import problems

spencern:
This seems to fix the Shopify importer for me.

Not sure why that would be the case, but putting this up here for now, I'm gonna pull it down and run it through on a fresh install in a bit.

### DIFF
--- a/imports/plugins/included/connectors-shopify/package.json
+++ b/imports/plugins/included/connectors-shopify/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "reaction-connectors-shopify",
+  "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


spencern:
@mikemurray this fixes the problems with the importer.. but it doesn't make sense to me as why it's fixing it. There are issues with the importer running afoul of Meteor fibers when the package.json file is not present, but this doesn't really feel like a clean fix to me. 
I'm gonna keep looking at it, but this might do to let us keep experimenting with the onboarding flow for now.

spencern:
Adding the WIP flag here because I want to re-write the way this works so that it's not dependent on whatever is happening with the package.json - it should work in either scenario.

Issue #3018, opened 2017-10-03
Use "Completed Order" component in Profile page


zenweasel:
We use the old completed order Blaze template in the profile page but we should reuse the Completed Order component that was converted to React and made marketplace-aware.

janus-reith:
Maybe we could even take this a step further?
The orders are shown on the completed page, the profile page, the orders dashboard panel and the eMail template.
It would be great to generalize this somehow, to make customization easier.

zenweasel:
@janus-reith I would agree except maybe for the orders dashboard as that's an entirely different view with different permissions (especially as it pertains to marketplace). But I think sharing this code for everywhere a user looks at their own order makes a lot of sense.

rymorgan:
This hasn't been fully reviewed or approved, but that's my plan with the work I've done on our revamped user profile views. https://scene.zeplin.io/project/58af8900737c137280322bd6/screen/59b7272087fcafc969870826

zenweasel:
@rymorgan Do I need to wait for any designs for the task in the title or can I have someone start on that?

rymorgan:
@zenweasel You can start this now. As you can see from the zeplin designs, I want us to make the user profile pages more user-friendly in the near future. Those new designs assume we use the complete order template here as well so it'd be good to do it now so it's ready for the redesign.

rymorgan:
@efalayi Yes, that's the point of this issue. Let's bring it into the profile as it is now (the new react component that Brent built). It'll probably need some design adjustments, but this should use the same component.

spencern:
Resolved by #3112

PR #3112, opened 2017-10-15
Use "Completed Order" component in Profile page

efalayi:
Resolves [#3018](https://github.com/reactioncommerce/reaction/issues/3018)
Displayed orders list on user profile page uses the old completed order Blaze template  but we should reuse the Completed Order component that was converted to React and made marketplace-aware.

###### Fix
- Create a template for userOrdersList to manage list display
- Create an helper function for userOrdersList template that return the `CompletedOrder` react component
- Create a container component for userOrdersList
- Fix CSS style errors
- Update profile template with new template

###### Test
```
Login as admin
Create an order
Navigate to user profile page
Scroll to displayed user order
Observe that displayed list matches the completedOrder template
```


zenweasel:
@efalayi As I've mentioned before, when we want Ryan to review something let's add the `design-review` label, not add him as a reviewer.

zenweasel:
So you will need to make some adjustments to the completed order component because some of those elements don't make sense within the context of the profile page, e.g. the giant "Thank you".

account_profile

imports/plugins/core/accounts/client/templates/profile/userOrdersList.js
```diff
@@ -0,0 +1,82 @@
+import { Template } from "meteor/templating";
+import { Orders, Media } from "/lib/collections";
+import CompletedOrder from "/imports/plugins/core/checkout/client/components/completedOrder";
+
+/**
+ * @method handleDisplayMedia
```
zenweasel: Isn't there a way we could be rendering the Completed Order container rather than duplicating all this code?

imports/plugins/core/accounts/client/templates/profile/userOrdersList.html
```diff
@@ -0,0 +1,11 @@
...
+  {{#each orders data=data}}
+    <div>
+      {{> React completedOrder }}
+    </div> 
+  {{else}}
+    <div class="alert alert-info">
+      <span data-i18n="cartCompleted.noOrdersFound">No orders found.</span>
+    </div>
+  {{/each}}
+</template>
```
zenweasel: Missing CR/LF

rymorgan:
Two things: 

1. Let's tweak the layout so that the items and cart portion stretch across the full box
2. Also, on the user profile, the title should change from 'Your cart' to 'Order summary'

complete orders view - user profile

efalayi:
@ryan should the tweak in layout be specific to just the user profile page?

rymorgan:
@efalayi Yep, only on the user profile page.

efalayi:
@ryan I just observed that the layout above is different from the one here: https://scene.zeplin.io/project/58af8900737c137280322bd6/screen/59b7272087fcafc969870826. Do we still intend to use the layout on zeplin?

rymorgan:
@efalayi The layout in Zeplin is where we are going in the future w/ a refreshed design of our user profile views. The issue we are working in this PR is using the component consistently. So use the designs I provided here vs the Zeplin ones, which we will tackle when we work on the rest of the user profile updates.

efalayi:
@zenweasel @rymorgan I've implemented all requested changes on this PR.

rymorgan:
@efalayi I'm seeing some formatting problems and after that, this PR should be good form my perspective. 

Need to vertically center align the copy in the rows for 'Your items'. 

account_profile

Need to fix casing 
account_profile-2


On mobile - 
You can reduce the padding by a half. 
Stack shipping address and payment method. 
Make sure the containers are all centered. (email/password, your orders, ad address) and not getting cut off. 

iphone_x_-_ios_11_0

efalayi:
@rymorgan I've updated the PR with the changes specified.

rymorgan:
A few of my comments didn't get addressed yet. 

Match the spacing between outer box and inner box. The Your orders box should match the address box. So, we need to reduce the spacing to match.

iphone_x_-_ios_11_0_and_inbox_ _morgan_ry_gmail_com

I know this wasn't in the original ticket, but while were on this page, I'd like to fix the casing issue on the address book title. Should read 'Address book' 

account_profile-4

I also noticed two other spacing issues see below: 

The spacing is inconsistent between all of the boxes. All the boxes w/ title should have even spacing between them. The Your items box has more space both above and below than it should. This is happening on mobile and desktop. 

account_profile-5

There's an awkward unevenness between the shipping and payment boxes on desktop. 

account_profile-3

efalayi:
@zenweasel I created an `OrdersList` component as you suggested and refactored the code. @rymorgan the css issues have also been fixed.

zenweasel:
@rymorgan Are we will saying "Your cart" in the completed order page? Should we just say "Your items"?

cart_completed

zenweasel:
@efalayi I spoke with @rymorgan and we agreed we should make it "Your items" (rather than You cart) on the completed order page as well as the profile page. Also we want to add Order ID on the Account version. Ryan has updated the comp.

rymorgan:
Here's the profile. 

complete orders view - user profile

efalayi:
I'll get to it and update the PR asap.

efalayi:
@zenweasel @rymorgan after making the change, this is what completed order page looks like:

completed-order-page

`Your items` is repeated twice. I'm not sure what the first does though.

rymorgan:
31971759-1f9edd8e-b915-11e7-9bac-4a59b87252ba_png__2560x1450_

efalayi:
@zenweasel I just updated the PR with the changes. I noticed that no message was displayed if the user has no orders (that's now fixed in the update).

imports/plugins/core/accounts/client/templates/profile/userOrdersList.js
```diff
@@ -0,0 +1,46 @@
...
+import { Reaction } from "/client/api";
+import CompletedOrder from "/imports/plugins/core/accounts/client/containers/userOrdersListContainer";
+
+/**
+ * @method getOrders
+ * @summary returns user's orders
+ * @since 1.5.1
+ * @return {Array} - an array of orders
+ */
+function getOrders() {
```
zenweasel: I'm confused as to why we are doing anything at all in the template. If we still need the template at all the only thing it should be doing is rendering the component.

imports/plugins/core/accounts/client/templates/profile/userOrdersList.html
```diff
@@ -0,0 +1,11 @@
+<template name="userOrdersList">
+  {{#if hasData data=data}}
+    <div>
+      {{> React completedOrders }}
+    </div> 
+  {{else}}
+    <div class="alert alert-info">
```
zenweasel: The component should be handling both the found and not found cases.

imports/plugins/core/checkout/client/components/completedOrder.js
```diff
@@ -27,8 +28,15 @@ const CompletedOrder = ({ order, orderId, shops, orderSummary, paymentMethods, h
       />
     );
   }
-  return (
-    <div className="container order-completed">
+
+  let headerText;
```
zenweasel: This component should not need to be aware of what route it is in. It should be handled by passing in a prop. Components should be as dumb as possible.

imports/plugins/core/checkout/client/components/completedShopOrders.js
```diff
@@ -10,16 +9,24 @@ import CompletedOrderItem from "./completedOrderItem";
  * @property {String} shopName - The name of the shop
  * @property {Array} items - an array of individual items for this shop
  * @property {Function} handleDisplayMedia - A function for displaying product images
+ * @property {boolean} isProfilePage - Checks if current page is Profile Page
```
zenweasel: Consistent casing for Javascript types

imports/plugins/core/checkout/client/components/completedShopOrders.js
```diff
@@ -10,16 +9,24 @@ import CompletedOrderItem from "./completedOrderItem";
...
+    </span>
+  ) : `${shippingMethod.carrier} - ${shippingMethod.label}`;
   return (
     <div className="order-details-shop-breakdown">
       {/* This is the left side / main content */}
       <div className="order-details-info-box">
         <div className="store-detail-box">
           <span className="order-details-store-title">{shopName}</span>
-          <span className="order-details-shipping-name">{shippingMethod.carrier} - {shippingMethod.label}</span>
+          <span className="order-details-shipping-name">{shippingName}</span>
```
zenweasel: Why was this changed?

imports/plugins/core/checkout/client/components/shopOrderSummary.js
```diff
@@ -6,9 +6,11 @@ import { Components, registerComponent } from "@reactioncommerce/reaction-compon
  * @summary Displays the order summary for each shop
  * @param {Object} props - React PropTypes
  * @property {Object} shopSummary - An object representing the summary information for this Shop
+ * @property {boolean} isProfilePage - Checks if current page is Profile Page
```
zenweasel: Consistent casing

zenweasel:
Still not exactly right on mobile devices

iphone_x_-_ios_11_0_and_applications_and_cart_completed

zenweasel:
Seems like this order break could be delineated better. Really gets lost on mobile.

account_profile

rymorgan:
@zenweasel Agree, let's add a rule between different orders. Two things: 

The orders ID text should be 16px not 14px font. Right now it's smaller than everything else. 
Let's add a rule between the orders. The rule **should not** extend across the whole container. It should be the width of the order box. Here's a mock. 

orders detail

imports/plugins/core/accounts/client/templates/profile/userOrdersList.html
efalayi: Okay. I'll create functions to render found and not found cases in the component.

imports/plugins/core/checkout/client/components/completedShopOrders.js
efalayi: `shippingName` varies depending on whether the current page is profile page. It is defined in  line 16.

efalayi:
@zenweasel at what screen size does this occur?

iphone_x_-_ios_11_0_and_applications_and_cart_completed

I just tested with chrome tools.

PR #3123, opened 2017-10-18
jsdoc: document registerSchema, getSchemas in collections package

machikoyasuda:
An example of how to document core methods and add references to the docs.

Screenshot:
screen shot 2017-10-17 at 5 11 48 pm

Sidebar:
screen shot 2017-10-17 at 5 12 41 pm


imports/plugins/core/collections/lib/registerSchema.js
```diff
@@ -1,11 +1,19 @@
 export const Schemas = {}; // populated with all Schemas
 
 /**
- * Register a schema
+ * @file **Reaction Schemas** - Use these methods to register and fetch Reaction Core schemas.
+ * See {@link https://docs.reactioncommerce.com/reaction-docs/master/simple-schema full documentation}.
+ *
+ * @module collections
```
aaronjudd: I wonder if we should group `schemas` and `collections` independently at the same prominence?

PR #3126, opened 2017-10-18
jsdoc: document all Meteor.call() methods

machikoyasuda:
## What this PR does
1. Adds a `@namespace` called `Meteor/Accounts`.
2. Adds `@memberof Meteor/Accounts` in the jsdoc definition to methods that are called with `Meteor.call("accounts/...")` across 3 different files
3. Cleans up syntax: Use `@summary`, over `@description`. Specify `@method` when using `@memberof Meteor/Accounts`
4. Adds `@namespace` for all of the Meteor.call() methods:
- `Meteor/Accounts`
- `Meteor/Cart`
- `Meteor/Email`
- `Meteor/Group`
- `Meteor/Inventory`
- `Meteor/Notification`
- `Meteor/SMS`
- `Meteor/Workflow`
- `Meteor/Product`
- `Meteor/Shop`
- `Meteor/Reaction`

## How that renders:
1. Sidebar:
screen shot 2017-10-17 at 7 40 20 pm

2. One page `Meteor/Accounts`: All methods appear on 1 page

screen shot 2017-10-17 at 7 40 36 pm

## How to do it:
1. In one file: Create `@namespace` at the top of the file. A namespace only needs to be declared once across the whole repository:

```
/**
 * @file Meteor methods for Accounts
 * Reaction extends {@link https://github.com/meteor/meteor/tree/master/packages/accounts-base MeteorAccounts} with Reaction-specific behavior and user interaction.
 * @namespace Meteor/Accounts
 */
```

2. In all methods: Above the method, add:

```
@name NameOfMethod
@method
@memberof NameOfNamespace
@summary Method summary goes here
@params {Object} shop - current shop
@returns {Object} shop - new shop
```


machikoyasuda:
@aaronjudd @spencern What do you think about grouping `Meteor.call()` methods this way? I'm open to suggestions.

spencern:
I think this could be a good way to subgroup the methods.

I'd propose that instead of something like `addressBookAdd` we use `account/addressBook/add` or `account/addresses/add`, but realize that we'd be changing the API at that point. I think at the minimum we should have a documented standard for how we name our methods, both private / internal methods as well as external methods that can be accessed via `Meteor.call()`

This would give us natural, alphabetical organizing for methods as well.

- `shops/members/invite`
- `account/verify`
- `account/permissions/set`
- `account/emails/update`
- `/emails/account/welcome/send` or `onboarding/emails/welcome/send` ?
- `account/tokens/create` (does it need to be a fallback token specifically? 

etc.. 

We can alias old ways of naming methods to keep our API backwards compatible. Open to all the reasons this is a bad idea as well..

zenweasel:
I also agree about making these names more consistent. I think we are pretty good already and there are just a few outliers. @spencern Should we make an issue? I've been adding some "cleanup" tasks already.

spencern:
@zenweasel Probably should make an separate issue for discussion, this is really a core API / architecture discussion topic I think. I'm not sure we're really all that consistent with our method naming yet and I'm also not convinced that we should keep our current naming structure just because it's what we've got, especially if it makes it harder to understand, find, or document.

machikoyasuda:
screen shot 2017-10-19 at 3 52 39 pm
screen shot 2017-10-19 at 3 52 47 pm
screen shot 2017-10-19 at 3 52 53 pm
@spencern @zenweasel 

Glad that listing out all the methods like this in jsdoc has surfaced inconsistencies and spurred discussion :)

Here are all the `Meteor.call()` methods and their names. Doesn't include `.publish()` methods or anything else.

aaronjudd:
@machikoyasuda are there any blockers here? like to see this in 1.5.4

machikoyasuda:
@aaronjudd @spencern Ready now:

screen shot 2017-10-24 at 7 41 22 am
screen shot 2017-10-24 at 7 37 43 am

PR #3132, opened 2017-10-18
Disable shops

### LINKED ISSUE #3131: Disable shops
An admin should be able to: 

1. Archive or disable a shop - which includes removing permissions/access for all users of that shop
2. Hide from view all of the products from that shop
3. Remove that shop from the navigation

Linked to issue on Foray repo

joykare:
Resolves #3131

### How to test
1. Create multiple shops
2. Create a few products in each shop
3. Disable a shop
4. Observe you cannot navigate to the shop anymore
5. Also observe you cannot see products related to disabled shop
6. Repeat 4 & 5 for non-admin users (ensure you cant navigate to shop or see products associated to shop)

### DIFF
--- a/imports/plugins/included/product-variant/containers/productsContainer.js
+++ b/imports/plugins/included/product-variant/containers/productsContainer.js
@@ -166,16 +166,16 @@ function composer(props, onData) {
     window.prerenderReady = true;
   }
 
-  const shops = Shops.find({
+  const activeShops = Shops.find({
     "workflow.status": {
-      $ne: "disabled"
+      $in: ["active", "new"]
     }
   }).fetch();
 
   const productCursor = Products.find({
     ancestors: [],
     type: { $in: ["simple"] },
-    shopId: { $in: shops.map(shop => (shop._id)) }
+    shopId: { $in: activeShops.map(activeShop => (activeShop._id)) }
   });
 
   const products = productCursor.map((product) => {
--- a/imports/plugins/included/product-variant/containers/productsContainer.js
+++ b/imports/plugins/included/product-variant/containers/productsContainer.js
@@ -166,17 +166,17 @@ function composer(props, onData) {
     window.prerenderReady = true;
   }
 
-  const activeShops = Shops.find({
+  const activeShopsIds = Shops.find({
     $or: [
       { "workflow.status": "active" },
       { _id: Reaction.getPrimaryShopId() }
     ]
-  }).fetch();
+  }).fetch().map(activeShop => activeShop._id);
 
   const productCursor = Products.find({
     ancestors: [],
     type: { $in: ["simple"] },
-    shopId: { $in: activeShops.map(activeShop => (activeShop._id)) }
+    shopId: { $in: activeShopsIds }
   });
 
   const products = productCursor.map((product) => {
--- a/imports/plugins/included/product-variant/containers/productsContainer.js
+++ b/imports/plugins/included/product-variant/containers/productsContainer.js
@@ -167,9 +167,10 @@ function composer(props, onData) {
   }
 
   const activeShops = Shops.find({
-    "workflow.status": {
-      $in: ["active", "new"]
-    }
+    $or: [
+      { "workflow.status": "active" },
+      { _id: Reaction.getPrimaryShopId() }
+    ]
   }).fetch();
 
   const productCursor = Products.find({
--- a/imports/plugins/included/product-variant/containers/productsContainer.js
+++ b/imports/plugins/included/product-variant/containers/productsContainer.js
@@ -8,7 +8,7 @@ import { Reaction } from "/client/api";
 import { ITEMS_INCREMENT } from "/client/config/defaults";
 import { ReactionProduct } from "/lib/api";
 import { applyProductRevision } from "/lib/api/products";
-import { Products, Tags } from "/lib/collections";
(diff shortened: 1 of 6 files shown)


joykare:
@zenweasel review on this

zenweasel:
I think we should probably make this the inverse. That is, look for the "active" status and filter all those that are not active? Or maybe an array of visible states (to be named later, but just "active" for now). I know we are going to add more workflow states to shops really soon.

imports/plugins/included/product-variant/containers/productsContainer.js
```diff
@@ -166,9 +166,16 @@ function composer(props, onData) {
     window.prerenderReady = true;
   }
 
+  const shops = Shops.find({
```
zenweasel: Maybe we could name this something like `activeShops` to make it clearer what we are filtering for below.

zenweasel: If we filter for just the `_id` field we could also avoid using the `map` below.

joykare: We'd still have to use map coz after:
```
const activeShopIdsArray = Shops.find({
    "workflow.status": {
      $in: ["active", "new"]
    }
  }, { fields: { _id: 1 } }).fetch();
```
It returns an array of objects that looks like this:
```
[ { _id: "J8Bhq3uTtdgwZx3rz" }, { _id: "Xw8zPBgTmSF7ZdFZS" } ]
```
So eventually I'd need to get just the values

zenweasel: True. I forgot that it doesn't just return an array of ids.

joykare:
another review @zenweasel

zenweasel:
Hold your horses. I'm testing as fast as I can.

zenweasel:
Is "new" a status we want to show products for? That's the default status when they are created so I think not? @mikemurray ?

joykare:
Its also the constant status of the default/primary shop. It doesn't change to "active". So to show products in the primary shop we need it.

zenweasel:
Hmmm, maybe we should change that state then? That seems like it's going to trip us up elsewhere. Is that just set in the `Shops.json`?

zenweasel:
Except for that question (do we want to display the `new` state) this seems to be working exactly as advertised.

zenweasel:
Alternately we could change that query to be "shops that are active or equal to the primaryShopId". I don't think we need the option to deactivate the primary shop. Looks like the "new" state is defined by the `workflow` schema so I don't think we want to just change that by default.

zenweasel:
@spencern Your input would be appreciated

mikemurray:
@zenweasel `new` is the default status because of the workflow schema default value like you pointed out.

If we had onboarding then `new` could be "not yet onboard", and `active` be "onboarding complete". But as it stands, only the super admin can make a shop "active".

Status should have no effect on the primary shop.

joykare:
NB. Changed it to use the `primaryShop` in the meantime

joykare:
More scope on this, just for clarity: @mikemurray and @spencern 
To get the active shops I was checking for the workflow statuses; `new` and `active`. Coz the default status of a primary shop is `new`.
Right now I changed the query to search for `active` status or if the shop is primary.

spencern:
That sounds correct to me. I think the first solution is to only display products from "Active" shops, in the near term we'll be building out a more complex, extensible workflow for shops, but since a shop shouldn't display products until their payment, address, etc are setup, then I think for now a new shop should _not_ be displayed.

I agree that the exception for the primary shop should exist.

zenweasel:
It just occurred to me that we are doing this all on the client-side and we didn't add any filtering to the publications? We need to do that as well. We shouldn't be leaking unpublished Shops/Products out of the product/shop publications in the first place rather than filtering them on the client.

joykare:
@zenweasel 
The Shop/Products publications had a filter already; 
- What has changed now is that; the filter is moved up on the product pub so that it works for both admins and non-admins it initially just applied to non-admins
https://github.com/reactioncommerce/reaction/pull/3132/files#diff-937ffc563dcc5f313ce5a2ceecb2441aL520
- The filters on the client side are still there coz every time you change something e.g. `disable a shop` the subscription is only called if you refresh the page.

server/publications/collections/products.js
```diff
@@ -95,6 +95,15 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
     return this.ready();
   }
 
+  // Get disabled shop id's to use for filtering
+  const disabledShopIds = Shops.find({
```
zenweasel: Shouldn't we do this the same way we did above? e.g. `activeShops` rather than `disabledShops`?

zenweasel:
Data from the Product/Shop publications should be updating when the data changes, that's a pretty core Meteor thing. If the components aren't updating when the data changes that's a different thing (and surely beyond the scope of this PR, but we should open an issue)

joykare:
Yeah the backend is reactive its the components that don't update before a refresh.
Made an issue for this #3146 

Its ready now

joykare:
Findings, so far:
- Running the products test suite, by itself === passing tests
- Shop is deleted at some point by another test file/suite (after trying Shop.find() the shop created in the products test is not part of the shops found in the products pub...)

PR #3139, opened 2017-10-19
Rerun circle tests on test failure

spencern:
This PR creates two new files in `.circleci/`
 - `tests.sh` which runs tests as normal and pipes output (stdout and stderr) to a file. This file runs the tests with the `set +e` flag so that CircleCI doesn't fail if this set of tests fail.
 - `backup-tests.sh` which checks the output file (`testoutput`) for indications of failure and reruns tests in that case. If the tests passed, we can skip re-running. If the tests fail a second time, we let the CI fail


jshimko:
I simplified this a bit.  I moved everything to one file and made use of exit code checks instead of writing temporary files.

The whole thing is now...

```sh
# allow first one to potentially fail
set +e

reaction test

# if it worked, we're done here.  
# if not, run it one more time and don't swallow failures

if [ $? -eq 0 ]; then
  echo "Reaction tests have passed!"
else
  echo "Reaction tests failed. Trying one more time..."
  set -e
  reaction test
fi
```

spencern:
Cleaner, thanks!

jshimko:
And it worked.  First time failed, second time passed.

continuous_integration_and_deployment

aaronjudd:
@spencern we should add this to 1.5.3 (can do now)?

spencern:
@aaronjudd yeah, I'd like to

PR #3140, opened 2017-10-19
jsdoc: remove all {[type]} - causes parsing errors.

machikoyasuda:
The jsdoc Atom package adds `{[type]}` for you as you add doc tags. If you don't fill it out, the `{[type]}` will cause a parse error!

I went through all the instances of `{[type]}` in master and fixed them so we don't get parse errors.


machikoyasuda:
jk! closing this b/c the errors were already fixed in `release-1.5.3` branch.

PR #3141, opened 2017-10-19
jsdoc: format Invoice React component into @module

machikoyasuda:
## What this PR does
1. Add `@module` definition at the top of the file
2. Add `@type {propTypes}`
3. Add `@name` in front of the methods so they get rendered.

## How it renders

screen shot 2017-10-19 at 4 23 49 pm


aaronjudd:
Feels like the components should be a higher level group, at the same level as "method". This might just be a styling issue?
invoice_-_documentation_and_reactioncommerce_reaction

PR #3150, opened 2017-10-21
Add AWS region for api publishing

aaronjudd:
- While working w/o locally, region is needed on circleci 
- tested via ci ssh
- adds `API_DOC_BUCKET`, `API_DOC_BUCKET_REGION` to env variables for CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ jobs:
           name: Deploy to S3 if tests pass and branch is Master
           command: |
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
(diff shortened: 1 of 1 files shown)


aaronjudd:
@jshimko @spencern I believe this is good to merge direct to master.  While `aws s3 sync` works locally, on CircleCI it needed to be supplied as an option `--region` otherwise the command returns a warning. Updated both the bucket and region to be env variables as well. Tested via CI SSH.

spencern:
@aaronjudd how can I test this? Code looks good and makes sense, do we need to put ENV vars into CircleCI or do I supply them from my local?

aaronjudd:
Really only one way to truly test.

Issue #3120, opened 2017-10-17
Order shipped email total undefined


aaronjudd:
Order Total is `undefined` in the order completion email.  

![inbox__10_099_messages__164_unread_](https://user-images.githubusercontent.com/439959/31642104-0f9ea612-b29e-11e7-885d-8b4a2191de4b.png)

```
Node: 6.11.1
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.17.1
Reaction: 1.5.1
Reaction branch: release-1.5.1
Docker: 17.09.0-ce
```

spencern:
Resolved by #3153

PR #3151, opened 2017-10-23
Casting strings to ints

Akarshit:
Fix for #3120 

The problem was that the prices are being stored as Strings. So while summing casting has to be done.

### To Test
1. Place a order and checkout
2. In the mail received, the total would be a float.


zenweasel:
While this solves the problem I think we want to look for the core of the problem which is why are these getting written to strings in the first place.

PR #3153, opened 2017-10-23
Fix Order shipped email total undefined

### LINKED ISSUE #3120: Order shipped email total undefined

iakhator:
Resolves #3120 
This PR resolve Order shipped total price is undefined to Float. The cause of the issue was that prices was Stored as Strings which causes a NaN error. Before summing, the prices have to be cast to a Float to convert them to Numbers.

#### Test
* Login as Admin.
* Place an order and checkout.
* Check the email received to see the total price shown.


zenweasel:
Opened an issue #3215 for the underlying problem but I think we should get this workaround in in the meantime because having a broken order confirmation email is bad.

PR #3155, opened 2017-10-23
Fixes reaction issue #3154

### LINKED ISSUE #3154: SKIP_FIXTURES=true starts infinite loop on initialization
### Expected behavior
There should be a hook that runs functions before the core initialization where Shop data can be imported when a user skips default data load.

### Actual behavior
Since there are no "beforeCoreInit" hook operations, if SKIP_FIXTURES=true then the applications starts an infinite loop searching for a shopId.

Delete everything from Mongo database from terminal:
``` bash
reaction reset -n
```
Then run reaction skipping fixtures:
```bash
SKIP_FIXTURES=true reaction
```

### Versions
Node: 6.11.4
NPM: 3.10.10
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.12.0
Reaction: 1.5.3
Reaction branch: master

sgobotta:
Resolves #3154 

This PR provides a new hook in the very first line of initialization function, right before the shopId validation.
Version 1.5.0 came with a new validation in server/startup/load-data that uses SKIP_FIXTURES environment variable to skip default data load.
With this new hook a user could load default data from a custom plugin before the shopId validation.

**Test**
+ In the server/init.js file from a custom plugin add a new function to the hook
```javascript
Hooks.Events.add("beforeCoreInit", () => {
  CustomDataImportModule.startJob();
});
```
+ Then run reaction skipping default data
```bash
SKIP_FIXTURES=true reaction
```

### DIFF
--- a/server/api/core/core.js
+++ b/server/api/core/core.js
@@ -21,6 +21,10 @@ const { Jobs, Packages, Shops, Groups } = Collections;
 export default {
 
   init() {
+
+    // run beforeCoreInit hooks
+    Hooks.Events.run("beforeCoreInit");
+
     // make sure the default shop has been created before going further
     while (!this.getShopId()) {
       Logger.warn("No shopId, waiting one second...");


FacundoMainere:
This would help me too.

Issue #2797, opened 2017-09-06
Break Password setting Template into separate "initial/onboarding" and "reset" flows


kieckhafer:
Currently, we use the same Template (Blaze wrapped around React) for setting a password both the first time a user is able to - after they receive their initial welcome email - and anytime they are resetting their password.

We should:

1) Convert this fully to React (Issue #2746 )
2) Break this out to have two separate workflows depending on if this is an "onboarding" password set, or a password reset.

The Blaze templates are rendered here (currently in a PR branch, look for this same file in `master` when this gets pulled in:
https://github.com/reactioncommerce/reaction/blob/ek-accountsPatch/imports/plugins/core/accounts/client/templates/updatePassword/updatePassword.js#L16-L34

kieckhafer:
After diving in, I think the best course of action on this is to not split up the password reset flow into two completely separate templates / components, but rather, at least for now, add a `type` prop that will allow for subtle changes between the two templates.

Feedback is welcome.

spencern:
Resolved by #3166

PR #3166, opened 2017-10-25
Determine if password reset overlay is setting or resetting password

kieckhafer:
This PR adds a prop, `type`, which determines what type of password overlay we are seeing: a password reset, or an initial password setting overlay, which only happens when a new user is invited via email.

This will allow us to update the text shown for each type of display independently, as the language is currently confusing, only allowing the same text for both types of forms.

This is step one to resolving #2797 

To test:
1) Invite a user, via the Accounts dashboard
2) See the password modal displays `Set password` text
3) Log out, and send a password reset email to the same user
4) See the password modal displays `Update password` text


mikemurray:
@spencern ready to merge into the next release

PR #3175, opened 2017-10-25
jsdoc: add PropTypes namespace for PropType validating methods

machikoyasuda:
## What this PR does
1. Adds `PropType` namespace
2. Adds methods to this namespace

## How that renders

screen shot 2017-10-25 at 3 26 36 pm

### DIFF
--- a/lib/api/prop-types.js
+++ b/lib/api/prop-types.js
@@ -3,7 +3,7 @@ import { check } from "meteor/check";
 import * as Schemas from "/lib/collections/schemas";
 
 /**
- * @file **PropTypes** - React Component PropType methods for validating Tags
+ * @file **PropTypes** - React Component PropTypes methods for validating Tags
  *
  * @namespace PropTypes
  */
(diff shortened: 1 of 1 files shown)


lib/api/prop-types.js
```diff
@@ -2,12 +2,21 @@ import _ from "lodash";
...
+ */
+
 const TagSchema = Schemas.Tag.newContext();
 
 export const PropTypes = {};
 
 /**
- * React Component propType validator for a single Tag
+ * @name Tag
+ * @summary React Component propType validator for a single Tag
```
mikemurray: Should `propType` be `PropType` to match the `@file` description above?

machikoyasuda: yes! fixed!

spencern: In line 17 here it's still showing as `propType` to me

machikoyasuda:
Fixed again and updating master to 1.5.5

PR #3186, opened 2017-10-27
Fix Issue #3185: Expand Featured Images to fit container in Product Detail view

### LINKED ISSUE #3185: Expand Featured Images to fit container in Product Detail view
### Expected behavior
Product Images in the Product detail view fit the image container.

### Actual behavior
If the image is smaller than the container it displays in its actual size and aligned to the left.

Add a featured image to a product with a small resolution (lets say 250px x 250px)

### Versions
- Node: 8.1.3
- NPM: 5.0.3
- Meteor Node: 4.8.4
- Meteor NPM: 4.6.1
- Reaction CLI: 0.14.3
- Reaction: 1.5.3

I'll submit a PR right away.

glmaljkovich:
Resolves #3185 

This way child images will expand appropriately when they are too small.


zenweasel:
This works as advertised but I am not that's the desired behavior (stretching to fit). That's going to look pretty awful for some images. Thoughts @kieckhafer @mikemurray?

glmaljkovich:
@zenweasel It should be noted that this fix doesn't stretch the image _but it's container._ The image conserves its aspect ratio. I haven't been able to quick test it with tall images (using placeholder.com) because of CORS, but try changing the width of the **rui item half** div wrapping the image on the browser inspector to something like **max-width: 25% !important;** and you'll see it scales gracefully.

zenweasel:
@glmaljkovich Yeah, I was referring to scaling up the image. I tried a 200x200 image and it created a pretty blurry image when scaled up to the default size for the Product Detail Page.

rymorgan:
I agree with @kieckhafer that it shouldn't be cropping vertical images like this. Also, agree that it's probably on the store owner to upload a high enough resolution image to not get the crappy expanded blurriness. 

Having square product images is 80% use case but what if you had vertical images w/ a white background and a user wanted the whole area to appear white. I think a vertical should center to the box and have a transparent background. A media manager (on the roadmap should address some of these issues in the future too.

zenweasel:
@glmaljkovich Thanks a lot for this PR. I think this approach is not going to work for enough use cases for us to approve this PR. If you have another solution that also deals with the issues that @kieckhafer and @rymorgan noted I would love to look at it

Issue #3019, opened 2017-10-03
Add settings panel to define dimensions for default parcel to shipping dashboard


spencern:
With #2992 we now automatically calculate the parcel size for cart items from whichever variant was added to the cart. 

We need to add a settings panel within the shipping dashboard that permits defining default parcel sizes for products and variants that don't have dimensions defined. 
We should add sensible defaults for the default parcel size so that it's never null which could cause Shippo or other shipment providers to fail to calculate shipping correctly.

![image](https://user-images.githubusercontent.com/1203639/31150232-10d8c754-a850-11e7-95f8-e9ec962718bd.png)

This section should permit setting the default parcel size with the fields: width, height, length, and weight. 
Other app-wide shipping settings can also live here.

spencern:
This should default to a flat rate priority box dimensions and a weight of 1lb for all reaction.
This should also permit users to update the default.

zenweasel:
@rymorgan Do we actually need design for a settings panel?

rymorgan:
@zenweasel I think we might at this stage. But something on my upcoming agenda is to work on some standards for settings cards that will enable us to build this w/out design.

efalayi:
@rymorgan is there layout for this section that I can use?

zenweasel:
@efalayi So these values should be used like this:
1. When a new product is created it should use these values as defaults.
1. When shipping is calculated if no values are set (i.e. they are 0) then it should use those values to send to Shippo.
1. When the store owner sets the values it should ask them if they want to reset all existing products, and if so it will overwrite all values with this new default.
1. If the product is not set and the default is not set it should fall back to the dimensions of a flat rate priority and a weight of 1 so that we never get a situation where we don't have a weight for a product.

zenweasel:
from @spencern 
USPS Priority Mail Medium Flat Rate Box - 11 1/4" x 8 3/4" x 6"
Not sure what a good default weight would be, maybe 7 or 8 lbs?

Reference for dimensions:
https://www.usps.com/ship/priority-mail.htm

efalayi:
@zenweasel should the user also have the option of not resetting existing products? 
Also, should the minimum values for size be 0 or 1?

efalayi:
@zenweasel could you please explain this part:
```
If the product is not set and the default is not set it should fall back to the dimensions of a flat rate priority and a weight of 1 so that we never get a situation where we don't have a weight for a product.
```

zenweasel:
[...]

PR #3193, opened 2017-10-30
Add default parcel size setting

efalayi:
Resolves [#3019](https://github.com/reactioncommerce/reaction/issues/3019)
Add a settings panel within the shipping dashboard that permits defining default parcel sizes for products and variants that don't have dimensions defined. Also include sensible defaults for the default parcel size so that it's never null which could cause Shippo or other shipment providers to fail to calculate shipping correctly.

###### Fix
- Create component and container for new setting
- Create new package `reaction-shipping-parcel-size` to manage parcel size setting
- Add default values. Reference: [USPS](https://www.usps.com/ship/priority-mail.htm) 
- Update `Shops.json` with default parcel size values
- Update Shops and Product Variant schemas
- Update core translation
- Write unit tests for new setting
- Update `cart-create.app.test`

Test
- Reset and start reaction
- Run `meteor mongo` and 'use meteor in terminal (or start Robomongo)
- `db.Shops.find({}, {"defaultParcelSize": 1, "_id": 0})` should give `{ "defaultParcelSize" : { "weight" : 7, "length" : 11.25, "width" : 8.75, "height" : 6 } }`
- Login as admin and enable payment
- Go to shipping setting
- Enable Shippo (refresh methods if no values are shown)
- Check values shown in `Default parcel size` form fields. They should be the same values in the database
- Add a product
- Check the product dimensions. They should be the same values in the database
- Publish product, add to cart, and checkout (Shipping rates should be displayed)
- Check `Orders` in database. `items.parcel` should be the same value of product dimensions
- Change product dimensions and repeat step 10. 
- Check `Orders` in database and verify that `items.parcel` is equal to the new dimensions saved for the product
- Create another product and add a variant option. Variant dimensions should be equal to the values in the database
- Publish product, add variant option to cart, and checkout
- Check Orders in database. `items.parcel` should be the same value of variant option dimensions
- Change variant option dimensions and repeat steps 15 - 16
- Add a new product and edit dimension values
- Repeat steps 10 - 13
- Add a new product
- Repeat step 18 with zero dimension values (i.e: weight = 0, length = 0, width = 0, height = 0)
- Check `Orders` in database and verify that `items.parcel` is equal to the value for default parcel size
- Remove defaultParcelSize value in `Shops.json`
- Reset and run reaction
- Repeat above steps


rymorgan:
@zenweasel Should we take the extra time and convert this to React? Also, based on the title I don't think it's clear what this panel is meant to do. In terms of translations, this is too long but it should be something more like "Default parcel size." 

@spencern Do you have any ideas of a shorter clearer way to say that?

rymorgan:
Also, translations aren't working for the title. 

shipping panel

spencern:
One note based on that screen shot is that I don't think parcel size should have an "enable/disable" switch on it. 

Also have failing tests and a failing bithound file

efalayi:
@rymorgan I've fixed the translation issue. It now works.

rymorgan:
- Translation still not working. 
- It needs to be sentence case. 
- I think this should say "Default parcel size" to make it's functionality clear. @spencern Can you confirm that that makes the most sense. Parcel size doesn't really say what 

basic_reaction_product_and_cart_checkout

rymorgan:
- Labels are screwed up. Weight says inches and With say lbs so they have the wrong unit of measure. 
- The units of measure shouldn't be hardcoded and it seems like it might be. 
- Remove the save button it's unnecessary in this application. 

basic_reaction_product-2

rymorgan:
This should be 7 not 70

basic_reaction_product-3

rymorgan:
The weight isn't right on the product level. 

basic_reaction_product-4

efalayi:
@rymorgan the translation files have to be added in `/imports/plugins/included/default-parcel-size/server/i18n` and updated `/imports/plugins/included/default-parcel-size/server/i18n/index.js`

efalayi:
@rymorgan the default values show when a new product is created. The `basic reaction product` already has a weight set.

zenweasel:
@efalayi You've requested my review on this but it still says [WIP]. Is this ready for review?

efalayi:
@zenweasel no, it's not ready for a full review yet.

efalayi:
@zenweasel this PR is ready for review

zenweasel:
@efalayi Help me understand the idea behind creating a new plugin for these settings rather than putting this inside the `core/shipping` plugin.

efalayi:
@zenweasel I observed sub-settings (flat rates and shippo) in Shipping were created as plugins. I used them as a guide.

zenweasel:
When I try to checkout using Shippo I get this error and I cannot complete checkout:

```
00:14:37.425Z  WARN Reaction: Failed to get shipping methods from these packages: [ { packageName: 'shippo', fileName: 'shippo.js' } ]
00:15:03.300Z  WARN Reaction: Failed to get shipping methods from these packages: [ { packageName: 'shippo', fileName: 'shippo.js' } ]
```

zenweasel:
@efalayi True but those are shipping methods and this is not. It's just a general setting for all shipping methods. I don't think for simplicity sake we want to create a new plugin just for a few settings.

efalayi:
I've been seeing that error since last month. However, it disappears when  I refresh methods in Shippo.

zenweasel:
I test this almost every day and I do not see it. And I have refreshed methods and it's still erroring.

efalayi:
@zenweasel what if we decide to add unit calculations in future?

zenweasel:
@efalayi How would that make a difference? Would that make this a shipping method?

zenweasel:
@efalayi Also, these test instructions could **really** use some work. They should test every part of the app that you touched (for example, seeing the default settings in the PDP). Not just open a settings panels and look at the fields. This is in your own best interest to have these as complete as possible.

efalayi:
I read the source code of `react-table` package and I discovered that:
1. `min-width` can be hard-coded.
1. `max-width` is dynamically set when a resize function is called.
1. passing `maxWidth` and `minWidth` as props doesn't override the previous css styles.

Looking at the table for the orders dashboard, I noticed that some columns do not really need to be resizable. Since the resize function can be disabled in `react-table`, what if we disabled the resizing of some (or all) columns and then display extra text with a tooltip? @joykare  @kieckhafer @spencern what do you think?

efalayi:
#### Update
- Writing unit test to ensure that Shippo actually receives the default values.

efalayi:
@zenweasel requested changes on the PR have been implemented.

imports/plugins/core/shipping/client/components/parcelSizeSettings.js
```diff
@@ -0,0 +1,131 @@
...
+   * @memberof ParcelDimensionSettings
+   */
+  handleFieldFocus() {
+    this.setState({
+      isEditing: true
+    });
+  }
+  /**
+   * @param {Object} event
+   * @memberof ParcelDimensionSettings
```
zenweasel: I think these JSDocs are incomplete for most these functions

imports/plugins/core/shipping/client/containers/parcelSizeSettingContainer.js
```diff
@@ -0,0 +1,99 @@
+import { compose, withProps } from "recompose";
+import { registerComponent, composeWithTracker } from "@reactioncommerce/reaction-components";
+import { Meteor } from "meteor/meteor";
+import { Reaction } from "/client/api/";
+import { isEmpty } from "lodash";
```
zenweasel: Import order. Please check the style guide

imports/plugins/core/shipping/client/containers/parcelSizeSettingContainer.js
```diff
@@ -0,0 +1,99 @@
...
+import ParcelSizeSettings from "../components/parcelSizeSettings";
+
+/**
+ * @method isNumber
+ * @summary checks if an input is an integer or a decimal
+ * @param {String} input - input string
+ * @since 1.5.5
+ * @return {Boolean} - returns a boolean result
+ */
+const isNumber = (input) => {
```
zenweasel: The description and the name don't match. Does this tell you if something is a number or if it's an integer or a decimal (both numbers)? If it's checking if it's a number or not I would prefer we use `_.isNumber` or `_.isFinite` rather than writing our own one-off batch of regex.

imports/plugins/core/shipping/client/containers/parcelSizeSettingContainer.js
```diff
@@ -0,0 +1,99 @@
...
+  };
+};
+
+/**
+ * @method saveDefaultSize
+ * @summary calls "shipping/size/save" method if size is a valid object
+ * @param {String} shopId - current shopId
+ * @param {Object} size - size to be saved
+ * @since 1.5.5
+ * @return {Function} callback
```
zenweasel: Parameter callback missing from JSDoc

imports/plugins/core/shipping/client/containers/parcelSizeSettingContainer.js
```diff
@@ -0,0 +1,99 @@
...
+ */
+const saveDefaultSize = (shopId, size, callback) => {
+  const { isValid, sizeInt } = validateInput(size);
+
+  if (isValid === true) {
+    Meteor.call("shipping/updateParcelSize", shopId, sizeInt, (error) => {
+      if (error) {
+        Alerts.toast("An error occured", "error");
```
zenweasel: Need i18n. (Here and on all Alerts)

zenweasel: Since we have an error object, should we output it somewhere?

imports/plugins/core/shipping/server/methods/methods.app-test.js
```diff
@@ -0,0 +1,62 @@
...
+  weight: 10,
+  length: 10,
+  width: 10,
+  height: 10
+};
+
+let shop;
+let initialDefaultParcelSize;
+before(function () {
+  this.timeout(10000);
```
zenweasel: Do you really need those? If so, why?

imports/plugins/core/shipping/server/methods/methods.app-test.js
```diff
@@ -0,0 +1,62 @@
...
+    sandbox.restore();
+  });
+
+
+  describe("shipping/updateParcelSize", function () {
+    it("should throw 403 error with shipping permission", function (done) {
+      sandbox.stub(Roles, "userIsInRole", () => false);
+      // this should actually trigger a whole lot of things
+      expect(() => Meteor.call("shipping/updateParcelSize", "dummyShopId", parcelSize)).to.throw(Meteor.Error, /Access Denied/);
+      return done();
```
zenweasel: You only need to `return done()` when testing asynchronous functions

lib/collections/schemas/helpers.js
```diff
@@ -99,3 +99,67 @@ export function shopDefaultCountry() {
     return this.value;
   }
 }
+
+export function shopDefaultParcelWeight() {
```
zenweasel: This needs some commenting to explain why it exists and what it does

lib/collections/schemas/helpers.js
```diff
@@ -99,3 +99,67 @@ export function shopDefaultCountry() {
...
+    });
+    if (shop && shop.defaultParcelSize) {
+      return shop.defaultParcelSize.weight;
+    }
+    return this.value;
+  }
+  return this.unset();
+}
+
+export function shopDefaultParcelLength() {
```
zenweasel: Rather than exporting/importing these all individually, it might be better to create a namespace and just export/import that

server/methods/core/cart.js
```diff
@@ -434,6 +434,10 @@ Meteor.methods({
     // we need to get the parent of the option to check if parcel info is stored there
     const immediateAncestors = variant.ancestors.filter((ancestor) => ancestor !== product._id);
     const immediateAncestor = Collections.Products.findOne({ _id: immediateAncestors[0] });
+    // Get default parcel size
+    const shop = Collections.Shops.findOne({
+      _id: Reaction.getShopId
```
zenweasel: should `getShopId` be a function there? Has this code ever been run? Doesn't look like it has.

zenweasel:
What's with the red outline here?:

reaction

zenweasel:
If I blank out an output and hit save I get a red toast that says "Invalid Input". That's not good. I should be getting standard web validation
1. Outline the errant field in red
1. Show a meaningful message e.g. "Weight is required"

zenweasel:
I am not sure this supposed to open by default the first time. Even if it is, it's not retaining it's open/close status even after I have closed it, moved to another panel and move back. That's a lot of real estate to waste on something that theoretically gets set one time.

zenweasel:
If I have Flat Rate shipping enabled I get this error and "No Shipping Methods Configured" appears. I can't checkout.

```
09:25:34.738Z  WARN Reaction:
  Failed to get shipping methods from these packages: [ { packageName: 'flat-rate-shipping', fileName: 'hooks.js' },
    { packageName: 'flat-rate-shipping', fileName: 'hooks.js' } ]
```

imports/plugins/core/shipping/client/containers/parcelSizeSettingContainer.js
efalayi: @zenweasel the function was written to check that the input is not a string. Initially, I tried `_.isNumber` but it also returned true for a string. I'll try the `_.isFinite` method.

zenweasel:
Looks much better but when I get validation warnings and then correct them, the validation messages are not being cleared. (I hit save here already and got the green toast but these messages remained)
cart_checkout

server/methods/core/cart.js
```diff
@@ -444,6 +448,18 @@ Meteor.methods({
     if (variant.weight || variant.height || variant.width || variant.length) {
       parcel = { weight: variant.weight, height: variant.height, width: variant.width, length: variant.length };
     }
+
+    // Use default settings for parcel weight, length, height, and width
```
zenweasel: So these values should only be used when there is no parcel set on the product but it looks like these are going to be used in all cases. This needs to be wrapped in some sort of `if (!parcel)` conditional

server/methods/core/cart.js
```diff
@@ -434,6 +434,10 @@ Meteor.methods({
     // we need to get the parent of the option to check if parcel info is stored there
     const immediateAncestors = variant.ancestors.filter((ancestor) => ancestor !== product._id);
     const immediateAncestor = Collections.Products.findOne({ _id: immediateAncestors[0] });
+    // Get default parcel size
+    const shop = Collections.Shops.findOne({
+      _id: Reaction.getShopId()
```
zenweasel: I think we need to look more closely at which shop we should be grabbing here. I think these settings are set at the Marketplace level, not at the individual shop level (at least not yet)

efalayi: @zenweasel to fix this, I used `Reaction.getPrimaryShopId` method.

zenweasel: That's still not correct. This could from the primary shop or from the individual shops depending on marketplace settings.

imports/plugins/core/shipping/client/containers/parcelSizeSettingContainer.js
```diff
@@ -0,0 +1,90 @@
+import { Meteor } from "meteor/meteor";
+import { compose, withProps } from "recompose";
```
zenweasel: Import order

imports/plugins/core/shipping/client/containers/parcelSizeSettingContainer.js
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * @method onCardExpand
+ * @summary set "edit/focus" in current Reaction state
+ * @param {String} cardName - card name to be set
+ * @since 1.5.5
+ * @return {Function} callback
+*/
+const onCardExpand = (cardName) => {
+  Reaction.state.set("edit/focus", cardName);
```
zenweasel: I don't see this returning anything including a callback. Do we need to be doing `return Reaction.state.set`?

imports/plugins/core/shipping/client/containers/parcelSizeSettingContainer.js
```diff
@@ -0,0 +1,90 @@
...
+ * @since 1.5.5
+ * @return {String} - value in current Reaction state
+*/
+const getEditFocus = () => {
+  return Reaction.state.get("edit/focus");
+};
+
+/**
+ * @method composer
+ * @summary composer
```
zenweasel: I think we can probably say something better in the @ summary than "composer"

imports/plugins/core/shipping/client/containers/parcelSizeSettingContainer.js
```diff
@@ -0,0 +1,90 @@
...
+    onCardExpand,
+    size,
+    saveDefaultSize,
+    validation
+  });
+};
+
+registerComponent("ParcelSizeSettings", ParcelSizeSettings, [
+  composeWithTracker(composer),
+  withProps({})
```
zenweasel: If you don't have any props to pass you don't need to include `withProps` both here and below

imports/plugins/core/shipping/lib/collections/schema/parcelsize.js
```diff
@@ -0,0 +1,36 @@
...
+    decimal: true
+  },
+  height: {
+    type: Number,
+    min: 1,
+    decimal: true
+  },
+  weight: {
+    type: Number,
+    min: 1,
```
zenweasel: We discussed this and it's unlikely that any other parameter would be >0 but <1 but weight certainly could be, especially if their UOM was kg.

imports/plugins/core/shipping/server/methods/methods.app-test.js
```diff
@@ -0,0 +1,57 @@
...
+  beforeEach(function () {
+    sandbox = sinon.sandbox.create();
+  });
+
+  afterEach(function () {
+    sandbox.restore();
+  });
+
+  describe("shipping/updateParcelSize", function () {
+    it("should throw 403 error with shipping permission", function () {
```
zenweasel: I assume you actually mean **without** shipping permission?

imports/plugins/core/shipping/server/methods/methods.app-test.js
```diff
@@ -0,0 +1,57 @@
...
+      sandbox.stub(Reaction, "getShopId", () => {
+        return shop._id;
+      });
+      sandbox.stub(Shops, "update", () => {
+        shop.defaultParcelSize = parcelSize;
+      });
+      Meteor.call("shipping/updateParcelSize", shop._id, parcelSize);
+      const updateParcelSize = shop.defaultParcelSize;
+      expect(initialDefaultParcelSize).to.not.equal(updateParcelSize);
+      return done();
```
zenweasel: There's no reason for this `done` unless you are dealing with calling async functions. https://mochajs.org/

lib/collections/schemas/helpers.js
```diff
@@ -99,3 +99,103 @@ export function shopDefaultCountry() {
     return this.value;
   }
 }
+
+/**
+ * shopDefaultParcelSize
+ * @namespace
```
zenweasel: This is just blank?

private/data/Shops.json
```diff
@@ -3211,5 +3211,11 @@
     "isCommercial": true,
     "isShippingDefault": true,
     "isBillingDefault": true
-  }]
+  }],
+  "defaultParcelSize": {
```
zenweasel: I wonder if in addition to doing this if we shouldn't create auto-values on Shop? We sort of talked about this but now I am seeing want to also deal with UOM that aren't lbs.

zenweasel:
If I remove `defaultParcel` from the `Shops` collection (as people might do) I get this error:
```
TypeError: Cannot read property 'weight' of undefined
    at ParcelSizeSettings.render (parcelSizeSettings.js:129)
    at modules.js?hash=578d148ad84cf37bad8eb892026c14c06707c234:108571
    at measureLifeCyclePerf (modules.js?hash=578d148ad84cf37bad8eb892026c14c06707c234:107851)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (modules.js?hash=578d148ad84cf37bad8eb892026c14c06707c234:108570)
    at ReactCompositeComponentWrapper._renderValidatedComponent (modules.js?hash=578d148ad84cf37bad8eb892026c14c06707c234:108597)
    at ReactCompositeComponentWrapper.performInitialMount (modules.js?hash=578d148ad84cf37bad8eb892026c14c06707c234:108137)
    at ReactCompositeComponentWrapper.mountComponent (modules.js?hash=578d148ad84cf37bad8eb892026c14c06707c234:108033)
    at Object.mountComponent (modules.js?hash=578d148ad84cf37bad8eb892026c14c06707c234:100794)
    at ReactCompositeComponentWrapper.performInitialMount (modules.js?hash=578d148ad84cf37bad8eb892026c14c06707c234:108146)
    at ReactCompositeComponentWrapper.mountComponent (modules.js?hash=578d148ad84cf37bad8eb892026c14c06707c234:108033)
```

efalayi:
@zenweasel I've updated the PR with the requested changes. However, I have a blocker in creating a custom validation with SimpleSchema.

zenweasel:
@efalayi I added custom validator functions to the `ParcelSize` schema. Let me know if that unblocks you.

efalayi:
Thanks @zenweasel.

zenweasel:
Let's change all the `since` notes in the JSDocs too

zenweasel:
If I change the dimensions/weight for a parcel in the PDP, when I purchase that item it is still using the default parcel size. This would be a **huge** problem if this were to get into production and I had specifically mentioned that we need to test that the correct values are being used.

server/methods/core/cart.js
```diff
@@ -434,16 +434,45 @@ Meteor.methods({
...
+    // Set parcel
+    let parcel = {
+      weight: immediateAncestor ? immediateAncestor.weight : variant.weight,
+      height: immediateAncestor ? immediateAncestor.height : variant.height,
+      width: immediateAncestor ? immediateAncestor.width : variant.width,
+      length: immediateAncestor ? immediateAncestor.length : variant.length
+    };
+
+    // Set default parcel size as parcel if parcel is not defined
+    if (!parcel) {
```
zenweasel: When would this ever be false since you just set the value right above?

server/methods/core/cart.js
```diff
@@ -434,16 +434,45 @@ Meteor.methods({
...
-      parcel = { weight: variant.weight, height: variant.height, width: variant.width, length: variant.length };
+
+    // Get default parcel size from primary shop
+    const { defaultParcelSize }  = Collections.Shops.findOne({
+      _id: shopId
+    });
+
+    // Set parcel
+    let parcel = {
+      weight: immediateAncestor ? immediateAncestor.weight : variant.weight,
```
zenweasel: The logic here is wrong. If the immediate ancestor exists you use that weight even though that weight might be 0. Probably more correct (but still wrong) would be
```
weight: immediateAncestor.weight ? immediateAncestor.weight : variant.weight
```
but even that would be wrong because if the `variant.weight` exists we want to use that (that's why in my code it was applied last) so the correct logic would be

```
    let parcel = {
      weight: variant.weight ? variant.weight : immediateAncestor.weight,
      height: variant.height ? variant.height : immediateAncestor.height,
      width: variant.width ? variant.width : immediateAncestor.width,
      length: variant.length ? variant.length : immediateAncestor.length
    };
```

And a quick test of adding that to my cart shows that the correct get used instead of always using the default, but the default is still used when values are not defined on the variant.

packages/gridfs/.npm/package/npm-shrinkwrap.json
```diff
@@ -1,89 +1,90 @@
 {
+  "lockfileVersion": 1,
   "dependencies": {
     "bson": {
       "version": "0.5.7",
       "resolved": "https://registry.npmjs.org/bson/-/bson-0.5.7.tgz",
-      "from": "bson@>=0.5.4 <0.6.0"
+      "integrity": "sha1-DRH+CTbB/uAp4R9wY/XQqyQi6j4="
```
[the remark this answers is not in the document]
efalayi: @zenweasel I think these changes happen when I pull master before push my update.

zenweasel:
I spoke with @efalayi a few days ago and she is on vacation but will resume work when she returns. Marking blocked for now

zenweasel:
@efalayi Are you still planning on working on this?

zenweasel:
@Akarshit Can you take over this PR and see if we can get it over the finish line?

efalayi:
@zenweasel I tried using the `requiringShipping` value for a product but it's undefined. I checked the database and it wasn't defined too. Also, I'm unsure as to what the check defined in [https://github.com/reactioncommerce/reaction/blob/49deec7e0be54c4a40e8a41c507c9d9258940a4b/lib/collections/schemas/products.js#L190-L191](https://github.com/reactioncommerce/reaction/blob/49deec7e0be54c4a40e8a41c507c9d9258940a4b/lib/collections/schemas/products.js#L190-L191) does.

efalayi:
@Akarshit are you working on this presently?

zenweasel:
@efalayi The value is `requiresShipping` not `requiringShipping`

https://github.com/reactioncommerce/reaction/blob/8338c21a26bfc8b69068688dc7f3bb61848240d9/lib/collections/schemas/products.js#L515

efalayi:
@zenweasel yeah. That was a typo. It's undefined

efalayi:
@zenweasel could you try creating a new product and the check the database?

Akarshit:
@efalayi Ya I will be working on this.

Akarshit:
Steps I followed for testing

1. `reaction reset`
1. `reaction`
### Testing default value.
1. Run 'meteor mongo` and 'use meteor` in terminal
1. db.Shops.find({}, {"defaultParcelSize": 1, "_id": 0}) should give `{ "defaultParcelSize" : { "weight" : 7, "length" : 11.25, "width" : 8.75, "height" : 6 } }`
1. Login as admin
1. Go to example product and click on the variant, its dimensions should *not* be the default ones.
1. Add a product. 
1. Check its variant's dimension, they should be the default ones.
1. Publish this and buy this.
1. Check the parcel size in the `Orders` (`db.Orders.find({}, { 'items.parcel': 1, _id: 0 })`), should give the default parcel size.
1. Update the parcel size of the newly added product manually, and again buy the product and check the size in `Orders`
1. Change the defaultParcelSize in the Shipping side menu and then repeat the above steps to check consistency.

lib/collections/schemas/helpers.js
```diff
@@ -99,3 +99,82 @@ export function shopDefaultCountry() {
...
+
+  /**
+   * A function in shopDefaultParcelSize (shopDefaultParcelSize.weight).
+   * @function weight
+   * @memberof shopDefaultParcelSize
+   * @summary Function to get weight value from shop default parcel size
+  */
+  weight: function () {
+    const shop = new shopDefaultParcelSize.currentShop();
+    if (shop && shop.defaultParcelSize) {
```
Akarshit: @zenweasel could this be `false` anywhere?

private/data/Shops.json
Akarshit: I didn't understand what you meant here.

server/methods/core/cart.js
```diff
@@ -434,16 +434,37 @@ Meteor.methods({
...
+    const parcel = {
+      weight: variant.weight ? variant.weight : immediateAncestor.weight,
+      height: variant.height ? variant.height : immediateAncestor.height,
+      width: variant.width ? variant.width : immediateAncestor.width,
+      length: variant.length ? variant.length : immediateAncestor.length
+    };
+
+    // Check if any parcel property is set to zero and set to corresponding value on default parcel size
+    Object.keys(parcel).forEach(function (key) {
+      if (parcel[key] === 0) {
```
Akarshit: How can this be when in the schema it's defined to have minimum value of 1?

lib/collections/schemas/helpers.js
efalayi: @Akarshit yes. It could be false if defaultParcelSize is not defined in [Shops.json](https://github.com/reactioncommerce/reaction/blob/efb71a757a9a2c5058af3b55dcedb65854b4f786/private/data/Shops.json). After a discussion with @zenweasel we decided to also set autoValues for defaultParcelSize in [shops.js](https://github.com/reactioncommerce/reaction/blob/efb71a757a9a2c5058af3b55dcedb65854b4f786/lib/collections/schemas/shops.js).

Akarshit:
@efalayi Sorry for the confusion. I am not working on this, please continue.

zenweasel:
@efalayi Before I am going to willing to approve this PR I really need to see some in-depth test instructions that cover all the various aspects and scenarios covered by this ticket (no values in Shop.json,etc. and non-US UOM). This has the potential to introduce critical bugs into checkout so I need to have some really good means for understanding when this is actually working correctly and not having an impact on anything else. Something along the lines of what @Akarshit had started there

efalayi:
@zenweasel  I just updated the branch.

##### Test Steps
1. Reset and start reaction
2. Run `meteor mongo` and 'use meteor in terminal (or start Robomongo)
3. `db.Shops.find({}, {"defaultParcelSize": 1, "_id": 0})` should give `{ "defaultParcelSize" : { "weight" : 7, "length" : 11.25, "width" : 8.75, "height" : 6 } }`
4. Login as admin and enable payment
5. Go to shipping setting
6. Enable Shippo (refresh methods if no values are shown)
7. Check values shown in `Default parcel size` form fields. They should be the same values in the database
8. Add a product
9. Check the product dimensions. They should be the same values in the database
10. Publish product, add to cart, and checkout (Shipping rates should be displayed)
11. Check `Orders` in database. `items.parcel` should be the same value of product dimensions
12. Change product dimensions and repeat step 10. 
13. Check `Orders` in database and verify that `items.parcel` is equal to the new dimensions saved for the product
14. Create another product and add a variant option. Variant dimensions should be equal to the values in the database
15. Publish product, add variant option to cart, and checkout
16. Check Orders in database. `items.parcel` should be the same value of variant option dimensions
17. Change variant option dimensions and repeat steps 15 - 16
18. Add a new product and edit dimension values
19. Repeat steps 10 - 13
20. Add a new product
21. Repeat step 18 with zero dimension values (i.e: weight = 0, length = 0, width = 0, height = 0)
22. Check `Orders` in database and verify that `items.parcel` is equal to the value for default parcel size
23. Remove defaulParcelSize value in `Shops.json`
24. Reset and run reaction
25. Repeat above steps

###### Observation
Some shipping providers have max values for parcel dimensions (for instance [DHL](https://www.dhlparcel.be/en/business/service/sending-packages/size-and-weight-specifications).

###### Suggestions
- To ensure that users enter the right values, a chart for UOM conversion could be provided
- For downloadable products with zero dimension values, an option could be provided to update `requiresShipping`. Since these products do not need to be shipped, `requiresShipping` would be set to `false` and users can checkout (without selecting shipping rates). However for now, products with zero dimension values set to `defaultParcelSize` value.

@Akarshit could you please help test this branch. Thank you.

zenweasel:
@aldeed I think all four values are/were required because Shippo requires all four to give a shipping quote.

aldeed:
@zenweasel But why do they need to be required in the defaults object? Example:

I want to supply only a default weight, so I save the defaults form with only that field populated. It should succeed, resulting in defaults object of `{ weight: 10 }`

Then when I create a product, its `weight` prop should default to 10 but the other three props can default to `0` like before.

I suppose making all required in the defaults object with default values of 0 is pretty much the same, however it results in more data being stored unnecessarily.

zenweasel:
@aldeed I guess we were thinking that the primary use case would be forcing someone to set these values to be non-zero so we are protecting them from failing during checkout. This is really mostly only relevant in the case of using Shippo though, when using Flat Rate (or other shippers that allow dimensional weight) it doesn't make sense to force anything but weight to be required and if you are using Flat Rate, you don't need any of these values.

aldeed:
@zenweasel ah ok. Knowing that, I feel like there is something confusing about how this is implemented. You are entering these values in one place, but they're being used for two different purposes: (1) The defaults when you create a new product/variant, and (2) the `parcel` values in the cart for shipping purposes.

In reality, I think one is more of a product setting (create each new product with these size values because they are common), and the other is more of a shipping setting (for checkout/processing, use these size values if a product does not have them, because shipping needs them). Should there be two different sets of defaults?

Somewhat related, the docs and UI should be clearer about how inheritance works. If any one value is set on the variant, then it will take all 4 values from the variant rather than the parent, even if some are not set. It's possible the store owner expects per-field inheritance, and I didn't see any documentation about how it works.

zenweasel:
@aldeed Fair. Possibly one version of these settings could be moved into the Shippo settings (where we **could** force all four values to be set) and another in the "Catalog" section where they could all be optional.

Also, I think that @spencern's original requirement was that there be a "default default" so that even if you didn't set the defaults there would still be some sort of value there. I think this is where we got into all the stuff with the Shop schemas. But maybe these can be set at the shipper level rather than at the shop?

zenweasel:
I've spoke with @spencern about this and although there's been some good work done here I am very tempted to close this PR and rethink the implementation from the ground up again.

zenweasel:
I am closing this PR pending a rethinking of how this should be implemented. This seemed like a simple enough change but was clearly not thought through enough (by me).

Issue #3036, opened 2017-10-04
Empty cart flashing when I register a new user


rymorgan:
### Expected behavior
When I register a new user, I should stay on the same page, see a spinner (if there is a delay or pause) and not see a flashing empty cart. 

### Actual Behavior
When I register a user in checkout I hit submit and an empty cart screen flashes. 

![jujabxxuuq](https://user-images.githubusercontent.com/20409254/31205064-f6ecfde0-a923-11e7-84f9-b5badd03e918.gif)

### Steps to Reproduce the Behavior

Using an incognito window checkout as a guest 
See that an empty cart flashes. 

### Versions
[ALL]-1.5.0-rc.1

rymorgan:
I confirmed it's not flashing anymore, but it's a jarring experience for small action like registering a user to have a spinner happen on the whole page. If a spinner is necessary it should happen on the component level.

aaronjudd:
@zenweasel @spencern can we review this again, not sure that this PR to fix this actually does anything.

zenweasel:
@aaronjudd Are there steps to replicate this in core as I cannot using the instructions in this ticket?

Akarshit:
No "empty card flashing" screen when registering user. 
Checked in `release-1.8.0`

PR #3194, opened 2017-10-30
fix empty cart flashing when registering user

iakhator:
Resolve issue #3036 

This PR fix ```empty cart``` showing up while registering new user,  The reason for the empty cart showing up is that the cart.items is empty at the that window it tries to register the user. What I did was to create a helper method in `checkout.js` called `cartCount` where it checks if the `cart.item` exists and also check if the length is greater than 0.

#### Test
* Place an order.
* Checkout order.
* Next step will show telling you to `register`. 
* Input details and `register`.
* It takes you to the next step.

### DIFF
--- a/imports/plugins/core/checkout/client/templates/checkout/checkout.html
+++ b/imports/plugins/core/checkout/client/templates/checkout/checkout.html
@@ -1,5 +1,5 @@
 <template name="cartCheckout">
-  {{#if cart.items}}
+  {{#if cartCount}}
 
     <div class="checkout-progress">
       <div class="container">
--- a/imports/plugins/included/product-variant/containers/gridItemControlsContainer.js
+++ b/imports/plugins/included/product-variant/containers/gridItemControlsContainer.js
@@ -44,10 +44,20 @@ const wrapComponent = (Comp) => (
     // check whether all required fields have been submitted before publishing
     checkValidation = () => {
       // this returns an array with a single object
-      const variants = ReactionProduct.getVariants(this.props.product._id).map((variant) => this.validation.validate(variant));
-      this.setState({
-        validProduct: Object.assign({}, this.props.product, { __isValid: variants[0].isValid })
-      });
+      const variants = ReactionProduct.getVariants(this.props.product._id);
+
+      // should validate variants if they exist to determine if product is Valid
+      if (variants.length !== 0) {
+        const validatedVariants = variants.map((variant) => this.validation.validate(variant));
+        this.setState({
+          validProduct: Object.assign({}, this.props.product, { __isValid: validatedVariants[0].isValid })
+        });
+      } else {
+        // if variants do not exist then validation should pass
+        this.setState({
+          validProduct: Object.assign({}, this.props.product, { __isValid: true })
+        });
+      }
     }
 
     checked = () => {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "reaction",
   "description": "Reaction is a modern reactive, real-time event driven ecommerce platform.",
-  "version": "1.5.10",
+  "version": "1.5.11",
   "main": "main.js",
   "directories": {
     "test": "tests"
--- a/imports/plugins/core/checkout/client/templates/checkout/checkout.js
+++ b/imports/plugins/core/checkout/client/templates/checkout/checkout.js
@@ -15,6 +15,13 @@ Template.cartCheckout.helpers({
       return Cart.findOne();
     }
     return {};
+  },
+  cartCount() {
+    const cart = Cart.findOne();
+    if (cart.items && cart.items.length > 0) {
+      return true;
+    }
+    return false;
   }
 });
 


iakhator:
cc @rymorgan

rymorgan:
I confirmed it's not flashing anymore, but it's a jarring experience for small action like registering a user to have a spinner happen on the whole page. If a spinner is necessary it should happen on the component level.

iakhator:
@rymorgan, the sad thing about this is that it is written in blaze so  it will be better the spinner is removed in my opinion or what do you suggest I do?

rymorgan:
@zenweasel Thoughts? We should really be working on converting this to react. I don't want to sink a bunch of work into blaze components. This solution looks less broken thane before but I don't think is good UX yet.

zenweasel:
I definitely don't want this bug fix ticket to upscope into rewriting the checkout process in React. If this fixes the original bug let's merge and move on and if we need to open a new issue for the spinner then let's do that but probably we will end up revisiting the entire UI.

rymorgan:
Ok, we can merge this, but I'm not happy with this UX. But it's worth merging because it looks fixes it from looking broken. I'll make an issue to improve this UX upon register.

zenweasel:
If I am on the Completed Order page and I log out I get a cascade of errors:
```
Invariant Violation: Route.render(): A valid React element (or null) must be returned. You may have returned undefined, an array or some other invalid object.
    at invariant (http://localhost:3000/packages/modules.js?hash=7ee907e6d17784321fe48b45f09bcaa0f5d0b366:85884:15)
    at ReactCompositeComponentWrapper._renderValidatedComponent (http://localhost:3000/packages/modules.js?hash=7ee907e6d17784321fe48b45f09bcaa0f5d0b366:108730:143)
    at ReactCompositeComponentWrapper.performInitialMount (http://localhost:3000/packages/modules.js?hash=7ee907e6d17784321fe48b45f09bcaa0f5d0b366:108261:30)
    at ReactCompositeComponentWrapper.mountComponent (http://localhost:3000/packages/modules.js?hash=7ee907e6d17784321fe48b45f09bcaa0f5d0b366:108157:21)
    at Object.mountComponent (http://localhost:3000/packages/modules.js?hash=7ee907e6d17784321fe48b45f09bcaa0f5d0b366:100918:35)
    at ReactCompositeComponentWrapper.performInitialMount (http://localhost:3000/packages/modules.js?hash=7ee907e6d17784321fe48b45f09bcaa0f5d0b366:108270:34)
    at ReactCompositeComponentWrapper.mountComponent (http://localhost:3000/packages/modules.js?hash=7ee907e6d17784321fe48b45f09bcaa0f5d0b366:108157:21)
    at Object.mountComponent (http://localhost:3000/packages/modules.js?hash=7ee907e6d17784321fe48b45f09bcaa0f5d0b366:100918:35)
    at Object.updateChildren (http://localhost:3000/packages/modules.js?hash=7ee907e6d17784321fe48b45f09bcaa0f5d0b366:107721:51)
    at ReactDOMComponent._reconcilerUpdateChildren (http://localhost:3000/packages/modules.js?hash=7ee907e6d17784321fe48b45f09bcaa0f5d0b366:107243:32)
```

zenweasel:
Verified does not happen on `master`

iakhator:
ok, I will see to this

iakhator:
@zenweasel , I have fixed all errors happening on logout and also glitches happening when signing in and signing out.

spencern:
@zenweasel @iakhator 
I'm seeing the following error message when attempting to change the base branch to `release-1.5.10` 

>There are no new commits between base branch 'release-1.5.10' and head branch 'itua-fix-issue-3036'

is it possible that this PR is a duplicate of something else?

zenweasel:
@spencern Even if the code was duplicated, wouldn't the commits still be unique?

imports/plugins/core/checkout/client/templates/checkout/checkout.html
```diff
@@ -25,7 +25,15 @@
       </div>
     </div>
   {{else}}
-    {{> emptyCheckoutCart}}
+    {{#if currentUser}}
```
spencern: image

While this does seem to work to remove the flash of "your cart is empty" I'm seeing a side effect of causing an "endless" spinner to be rendered for any logged in user who happens to be on the `cart/checkout` route with an empty cart.

iakhator: @spencern, @zenweasel 
The reason for this is happens is the way I implemented the checks. The user is shown with an infinite spinner which was a wrong implementation regarding the fact that every logged in user should be able to see the `empty cart page` if there are not items in the cart. 

#### Fix
- I created a helper method which checks if the items in the cart is available and if it has a value.

#### Test
- Login
- go to `cart/checkout` route.
- `empty cart error` should show now
- Checks works fine now. It register the user without flashing an empty cart

imports/plugins/included/product-variant/containers/gridItemControlsContainer.js
```diff
@@ -44,10 +44,20 @@ const wrapComponent = (Comp) => (
     // check whether all required fields have been submitted before publishing
     checkValidation = () => {
       // this returns an array with a single object
-      const variants = ReactionProduct.getVariants(this.props.product._id).map((variant) => this.validation.validate(variant));
-      this.setState({
-        validProduct: Object.assign({}, this.props.product, { __isValid: variants[0].isValid })
-      });
+      const variants = ReactionProduct.getVariants(this.props.product._id);
```
zenweasel: What's happening here?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "reaction",
   "description": "Reaction is a modern reactive, real-time event driven ecommerce platform.",
-  "version": "1.5.10",
+  "version": "1.5.11",
```
zenweasel: What?

PR #3195, opened 2017-10-30
Make all Meteor.Error calls consistent

### LINKED ISSUE #3122: Make all `Meteor.Error` calls consistent
This is just a "cleanup" task

`Meteor.Error` should be `Meteor.Error("short-message", "Long message with more detail");`

Currently we have a hodgepodge of different styles. Specifically we want to eliminate versions that use a number instead of a message as these don't really convey any helpful info.

codejockie:
Resolves #3122 

## How To Test
In an editor, search for all instances of `Meteor.Error` check that it conforms with this style `Meteor.Error("short-message", "Long message with more detail")`

### DIFF
--- a/server/methods/accounts/accounts.js
+++ b/server/methods/accounts/accounts.js
@@ -655,7 +655,7 @@ export function inviteShopMember(options) {
   if (!shop) {
     const msg = `accounts/inviteShopMember - Shop ${shopId} not found`;
     Logger.error(msg);
-    throw new Meteor.Error("shop-not-found", msg);
+    throw new Meteor.Error("not-found", msg);
   }
 
   if (!Reaction.hasPermission("reaction-accounts", this.userId, shopId)) {
@@ -667,11 +667,11 @@ export function inviteShopMember(options) {
 
   // check to ensure that invitee has roles required to perform the invitation
   if (!Reaction.canInviteToGroup({ group, user: Meteor.user() })) {
-    throw new Meteor.Error("access-denied", "cannot invite to group");
+    throw new Meteor.Error("access-denied", "Cannot invite to group");
   }
 
   if (group.slug === "owner") {
-    throw new Meteor.Error("bad-request", "cannot directly invite owner");
+    throw new Meteor.Error("bad-request", "Cannot directly invite owner");
   }
 
   const currentUser = Meteor.users.findOne(this.userId);
--- a/server/methods/accounts/accounts.js
+++ b/server/methods/accounts/accounts.js
@@ -327,7 +327,7 @@ export function addressBookAdd(address, accountUserId) {
   if (typeof accountUserId === "string") { // if this will not be a String -
     // `check` will not pass it.
     if (!Reaction.hasAdminAccess()) {
-      throw new Meteor.Error(403, "Access denied");
+      throw new Meteor.Error("access-denied", "Access denied");
     }
   }
   this.unblock();
@@ -422,7 +422,7 @@ export function addressBookUpdate(address, accountUserId, type) {
   if (typeof accountUserId === "string") { // if this will not be a String -
     // `check` will not pass it.
     if (!Reaction.hasAdminAccess()) {
-      throw new Meteor.Error(403, "Access denied");
+      throw new Meteor.Error("access-denied", "Access denied");
     }
   }
   this.unblock();
@@ -539,7 +539,7 @@ export function addressBookRemove(addressId, accountUserId) {
   if (typeof accountUserId === "string") { // if this will not be a String -
     // `check` will not pass it.
     if (!Reaction.hasAdminAccess()) {
-      throw new Meteor.Error(403, "Access denied");
+      throw new Meteor.Error("access-denied", "Access denied");
     }
   }
   this.unblock();
@@ -667,11 +667,11 @@ export function inviteShopMember(options) {
 
   // check to ensure that invitee has roles required to perform the invitation
   if (!Reaction.canInviteToGroup({ group, user: Meteor.user() })) {
-    throw new Meteor.Error(403, "cannot invite to group");
+    throw new Meteor.Error("access-denied", "cannot invite to group");
   }
 
   if (group.slug === "owner") {
-    throw new Meteor.Error(400, "cannot directly invite owner");
+    throw new Meteor.Error("bad-request", "cannot directly invite owner");
   }
 
   const currentUser = Meteor.users.findOne(this.userId);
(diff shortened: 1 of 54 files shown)


zenweasel:
I feel like trying to make these messages fit within the confines of HTTP response code is a leaky abstraction. Quickly you get lots of things that don't fit in that mold or sort of do but the HTTP message is not _quite_ what you want. I think the main thing is that the short messages need to be consistent. So every access denied needs to be `access-denied` not sometimes `forbidden`.

For me keeping the short messages consistent is about making the log files easily grep-able.

server/methods/accounts/accounts.js
```diff
@@ -667,11 +667,11 @@ export function inviteShopMember(options) {
 
   // check to ensure that invitee has roles required to perform the invitation
   if (!Reaction.canInviteToGroup({ group, user: Meteor.user() })) {
-    throw new Meteor.Error(403, "cannot invite to group");
+    throw new Meteor.Error("access-denied", "cannot invite to group");
   }
 
   if (group.slug === "owner") {
-    throw new Meteor.Error(400, "cannot directly invite owner");
+    throw new Meteor.Error("bad-request", "cannot directly invite owner");
```
zenweasel: Should start with capital letter

zenweasel:
I think some of these that are `invalid-thing` or `cannot-find-thing` could be switched to `invalid-parameter` with the long message being more descriptive. All the `access-denied`'s that are server-side can be made more explicit stating what access what denied, but leave the client-side ones vague.  ("You do not have access to modify products")

That's my opinion at least. Open to more suggestions and hopefully something that we can codify into a style rule.

server/methods/catalog.js
```diff
@@ -1394,12 +1394,12 @@ Meteor.methods({
         });
       } else {
         Logger.debug("invalid product visibility ", productId);
-        throw new Meteor.Error("Forbidden", "Variant is required");
+        throw new Meteor.Error("forbidden", "Variant is required");
       }
```
zenweasel: this isn't really forbidden. It seems like invalid parameters.

server/methods/catalog.js
```diff
@@ -1421,7 +1421,7 @@ Meteor.methods({
       return res === 1 && !product.isVisible;
     }
     Logger.debug("invalid product visibility ", productId);
-    throw new Meteor.Error("Bad Request");
+    throw new Meteor.Error("bad-request", "Bad Request");
```
zenweasel: should be `invalid-parameters`

server/methods/core/cart.js
```diff
@@ -348,7 +348,7 @@ Meteor.methods({
     const cart = Collections.Cart.findOne({ userId: this.userId });
     if (!cart) {
       Logger.error(`Cart not found for user: ${ this.userId }`);
-      throw new Meteor.Error(404, "Cart not found",
+      throw new Meteor.Error("cart-not-found",
```
zenweasel: Probably should be `invalid-parameter`

server/methods/core/cart.js
```diff
@@ -374,12 +374,12 @@ Meteor.methods({
     // const variant = Collections.Products.findOne(variantId);
     if (!product) {
       Logger.warn(`Product: ${ productId } was not found in database`);
-      throw new Meteor.Error(404, "Product not found",
+      throw new Meteor.Error("product-not-found",
         "Product with such id was not found!");
```
zenweasel: Let's remove exclamation points. Stay calm everyone.

server/methods/core/cart.js
```diff
@@ -845,14 +845,14 @@ Meteor.methods({
...
-        throw new Meteor.Error("An error occurred adding the address");
+        throw new Meteor.Error("server-error", "An error occurred adding the address");
       }
     }
     // refresh shipping quotes
     Meteor.call("shipping/updateShipmentQuotes", cartId);
 
     if (typeof cart.workflow !== "object") {
-      throw new Meteor.Error(500, "Internal Server Error",
+      throw new Meteor.Error("internal-server-error",
```
zenweasel: This should be just `server-error` like the rest.

server/methods/core/groups.js
```diff
@@ -55,13 +55,13 @@ Meteor.methods({
...
     if (groupExists) {
-      throw new Meteor.Error(409, "Group already exist for this shop");
+      throw new Meteor.Error("conflict", "Group already exist for this shop");
     }
     try {
       _id = Groups.insert(newGroupData);
     } catch (error) {
       Logger.error(error);
-      throw new Meteor.Error(400, "Bad request");
+      throw new Meteor.Error("bad-request", "Bad request");
```
zenweasel: Let's make this `invalid-parameter`

server/methods/core/groups.js
```diff
@@ -100,7 +100,7 @@ Meteor.methods({
 
     // prevent edits on owner. Owner groups is the default containing all roles, and as such should be untouched
     if (group.slug === "owner") {
-      throw new Meteor.Error(400, "Invalid request");
+      throw new Meteor.Error("bad-request", "Invalid request");
```
zenweasel: Let's make this `invalid-parameter` and the long explanation a little more helpful if possible.

server/methods/core/groups.js
```diff
@@ -216,11 +216,11 @@ Meteor.methods({
     // we are limiting group method actions to only users with admin roles
     // this also include shop owners, since they have the `admin` role in their Roles.GLOBAL_GROUP
     if (!Reaction.hasPermission("admin", Meteor.userId(), shopId)) {
-      throw new Meteor.Error(403, "Access Denied");
+      throw new Meteor.Error("access-denied", "Access Denied");
     }
 
     if (!user) {
-      throw new Meteor.Error(404, "Could not find user");
+      throw new Meteor.Error("not-found", "Could not find user");
```
zenweasel: invalid-parameter

spencern:
@zenweasel @codejockie 
My thinking on this seems to be inline with Brent here. I'd like to stay consistent with error messages where it makes sense, but agree that we shouldn't be dogmatic about sticking to HTTP status code style messages when it doesn't fit.

What I'd like to do is provide some type of style-guide for these messages so that people writing code don't have to stop to figure out what their error message should say, it can be open-ended in the sense that we can always add a new `short-message` as the need arises, but we can attempt to keep things consistent when possible.

@machikoyasuda let's figure out if there are some patterns we can adopt from this PR that we can use to start a style guide for this (and perhaps look around to see if any other open-source repos have good examples)

zenweasel:
@spencern @machikoyasuda Should we just approve/merge this PR for now and revisit when we have a standard? I feel like this is definitely a big improvement over the existing implementation

PR #3196, opened 2017-10-30
[WIP] Fix Reaction.hasPermissions for anonymous users

### LINKED ISSUE #2895: Update Reaction.hasPermission's owner pass and replace instances of Roles.userIsInRole
Reaction.hasPermission currently gives global access for users with "owner" role (i.e it always returns true for owners). This means that a check like `Reaction.hasPermission("anonymous")` will return `true` when we want it to return `false`.

This happens because `Reaction.hasPermission` joins "owner" to the roles passed to it. We need to remove that join, so that when we check for particular role(s), we are sure that the response is really based on if user has that role.

Then we need to ensure that all current usage of Reaction.hasPermission still works with this change.

After that, we proceed to replace instances of `Roles.userIsInRole` with `Reaction.hasPermission`.

foladipo:
Resolves #2895.

This PR removes the behaviour of `hasPermission` (on both the client and server) where it auto-appends the `owner` permission to the list of permissions supplied to it. Before now, if an `owner` user was logged in and the app ran `hasPermission("anonymous")`, the call would return `true` even though `owner` users don't have the `anonymous` permission. This PR fixes that by removing the auto-appending behaviour. Now, you have to explicitly list all the permissions you want `hasPermission` to check for whenever you call it.

## Test instructions
- For the server-side `hasPermission`, I have written some unit tests for it in `server/api/core/hasPermission.app-test.js`. You can check out the tests and then run them with `reaction test` to see whether they fail or not.
- For the client-side `hasPermission`, you can:
  - add [these logging statements](https://github.com/reactioncommerce/reaction/pull/3196#issuecomment-341572933) to any client-side file e.g the `render` method of `imports/plugins/core/ui/client/components/app/app.js`.
  - login as admin and navigate to `localhost:PORT/`.
  - open your browser's console and observe different calls to `hasPermission` and their results.

### DIFF
--- a/client/modules/core/main.js
+++ b/client/modules/core/main.js
@@ -192,9 +192,6 @@ export default {
       if (!checkPermissions) {
         return false;
       }
-      if (typeof checkPermissions !== "string" && !Array.isArray(checkPermissions)) {
-        return false;
-      }
 
       if (Roles.userIsInRole(userId, checkPermissions, group)) {
         return true;
--- a/client/modules/core/main.js
+++ b/client/modules/core/main.js
@@ -159,14 +159,19 @@ export default {
   },
 
   /**
-   * hasPermission - client
-   * client permissions checks
-   * hasPermission exists on both the server and the client.
-   *
-   * @param {String | Array} checkPermissions -String or Array of permissions if empty, defaults to "admin, owner"
-   * @param {String} checkUserId - userId, defaults to Meteor.userId()
-   * @param {String} checkGroup group - default to shopId
-   * @return {Boolean} Boolean - true if has permission
+   * hasPermission - Performs client-side permission checks. Note that
+   * hasPermission exists on both the server and the client. So this is
+   * the client-side version.
+   * @param {String | Array} checkPermissions - permissions to check for. If
+   * empty/undefined, this function returns false.
+   * @param {String} checkUserId - user ID. If empty/undefined, this function uses
+   * the value returned by Meteor.userId().
+   * @param {String} checkGroup - the group that the user being checked
+   * belongs to. If empty/undefined, this function uses the ID of the
+   * current shop or Roles.GLOBAL_GROUP.
+   * @since 0.14
+   * @return {Boolean} - returns true if the said user has any of the permissions
+   * in checkPermissions, and false if s/he has none of them.
    */
   hasPermission(checkPermissions, checkUserId, checkGroup) {
     let group;
--- a/client/modules/core/main.js
+++ b/client/modules/core/main.js
@@ -195,15 +195,6 @@ export default {
       } else if (typeof checkPermissions === "string") {
         permissions = [checkPermissions];
       } else {
-        /* TODO: Consult and add the code below if it is thought
-        of as useful.
-        if (!Array.isArray(checkPermissions)) {
-          return false;
-        }
-        if (checkPermissions.length === 0) {
-          return false;
-        }
-        */
         permissions = checkPermissions;
       }
       // if the user has owner permissions we'll always check if those roles are enough
--- a/client/modules/core/main.js
+++ b/client/modules/core/main.js
@@ -180,14 +180,10 @@ export default {
     let permissions = ["owner"];
     let id = "";
     const userId = checkUserId || Meteor.userId();
-    //
-    // local roleCheck function
-    // is the bulk of the logic
-    // called out a userId is validated.
-    //
-    function roleCheck() {
-      // TODO: There're some outdated comments in this function. Clean them out.
 
+    // This is declared as a function for the sake of reuse. Note that it is
+    // only called after userId is validated.
+    function roleCheck() {
       // permissions can be either a string or an array
       // we'll force it into an array and use that
       if (checkPermissions === undefined) {
@@ -197,46 +193,16 @@ export default {
       } else {
         permissions = checkPermissions;
       }
-      // if the user has owner permissions we'll always check if those roles are enough
-      // By adding the "owner" role to the permissions list, we are making hasPermission always return
-      // true for "owners". This gives owners global access.
-      // TODO: Review this way of granting global access for owners
 
-      // TODO: Find out why were we ever joining "owner" to the permissions
-      // array in the first place.
-
-      //
-      // return if user has permissions in the group
-      //
       if (Roles.userIsInRole(userId, permissions, group)) {
         return true;
       }
 
-      // global roles check
-      // TODO: Review this commented out code
-      /*
-
-      const sellerShopPermissions = Roles.getGroupsForUser(userId, "admin");
-      // we're looking for seller permissions.
-      if (sellerShopPermissions) {
-        // loop through shops roles and check permissions
-        for (const key in sellerShopPermissions) {
-          if (key) {
-            const shop = sellerShopPermissions[key];
-            if (Roles.userIsInRole(userId, permissions, shop)) {
-              return true;
-            }
-          }
-        }
-      }*/
-      // no specific permissions found returning false
       return false;
     }
 
-    //
-    // check if a user id has been found
-    // in line 156 setTimeout
-    //
+    // If userId has been found, this calls roleCheck and then returns the
+    // result of that call. Returns false otherwise.
     function validateUserId() {
       if (Meteor.userId()) {
         Meteor.clearTimeout(id);
@@ -246,21 +212,17 @@ export default {
       return false;
     }
 
-    //
-    // actual logic block to check permissions
-    // we'll bypass unecessary checks during
-    // a user logging, as we'll check again
-    // when everything is ready
-    //
+    /*
+      Checks for permissions. We'll skip performing some checks if the user is
+      currently logging in, as those will be done when log in is complete.
+    */
     if (Meteor.loggingIn() === false) {
-      //
       // this userId check happens because when logout
-      // occurs it takes a few cycles for a new anonymous user
+      // occurs. It takes a few cycles for a new anonymous user
       // to get created and during this time the user has no
       // permission, not even guest permissions so we
       // need to wait and reload the routes. This
-      // mainly affects the logout from dashboard pages
-      //
+      // mainly affects the logout from dashboard pages.
       if (!userId) {
         id = Meteor.setTimeout(validateUserId, 5000);
       } else {
--- a/client/modules/core/main.js
+++ b/client/modules/core/main.js
@@ -189,7 +189,7 @@ export default {
      * @method roleCheck
      * @summary check whether or not a user is in a list of roles.
      * @private
-     * @since 1.5.5
+     * @since 0.15.0
      * @return {Boolean} - returns true if a user is in one or more of the roles
      * listed and false if not.
      */
@@ -210,7 +210,7 @@ export default {
      * @summary verifies that Meteor.userId() returns a value, after which
      * it calls some functions, especially roleCheck.
      * @private
-     * @since 1.5.5
+     * @since 0.15.0
      * @return {Boolean} - if userId has been found, this calls roleCheck and
      * then returns the result of that call. Returns false otherwise.
      */
--- a/client/modules/core/main.js
+++ b/client/modules/core/main.js
@@ -177,33 +177,20 @@ export default {
       group = this.getShopId() || Roles.GLOBAL_GROUP;
     }
 
-    let permissions = ["owner"];
     let id = "";
     const userId = checkUserId || Meteor.userId();
 
     // This is declared as a function for the sake of reuse. Note that it is
     // only called after userId is validated.
     function roleCheck() {
-      // permissions can be either a string or an array
-      // we'll force it into an array and use that
-      // TODO: Should we also cater for checkPermissions === "" and checkPermissions === []?
-      if (checkPermissions === undefined) {
-        permissions = ["owner"];
-      } else if (typeof checkPermissions === "string") {
-        if (checkPermissions === "owner") {
-          return Roles.userIsInRole(userId, "owner", group);
-        }
-
-        permissions = [checkPermissions];
-      } else {
-        if (checkPermissions.length === 1 && checkPermissions[0] === "owner") {
-          return Roles.userIsInRole(userId, checkPermissions, group);
-        }
-
-        permissions = checkPermissions;
+      if (!checkPermissions) {
+        return false;
+      }
+      if (typeof checkPermissions !== "string" && !Array.isArray(checkPermissions)) {
+        return false;
       }
 
-      if (Roles.userIsInRole(userId, permissions, group)) {
+      if (Roles.userIsInRole(userId, checkPermissions, group)) {
         return true;
       }
 
--- a/client/modules/core/main.js
+++ b/client/modules/core/main.js
@@ -180,8 +180,14 @@ export default {
     let id = "";
     const userId = checkUserId || Meteor.userId();
 
-    // This is declared as a function for the sake of reuse. Note that it is
-    // only called after userId is validated.
+    /**
+     * @method roleCheck
+     * @summary check whether or not a user is in a list of roles.
+     * @private
+     * @since 1.5.5
+     * @return {Boolean} - returns true if a user is in one or more of the roles
+     * listed and false if not.
+     */
     function roleCheck() {
       if (!checkPermissions) {
         return false;
@@ -197,8 +203,15 @@ export default {
       return false;
     }
 
-    // If userId has been found, this calls roleCheck and then returns the
-    // result of that call. Returns false otherwise.
+    /**
+     * @method validateUserId
+     * @summary verifies that Meteor.userId() returns a value, after which
+     * it calls some functions, especially roleCheck.
+     * @private
+     * @since 1.5.5
+     * @return {Boolean} - if userId has been found, this calls roleCheck and
+     * then returns the result of that call. Returns false otherwise.
+     */
     function validateUserId() {
       if (Meteor.userId()) {
         Meteor.clearTimeout(id);
--- a/client/modules/core/main.js
+++ b/client/modules/core/main.js
@@ -186,6 +186,8 @@ export default {
     // called out a userId is validated.
     //
     function roleCheck() {
+      // TODO: There're some outdated comments in this function. Clean them out.
+
       // permissions can be either a string or an array
       // we'll force it into an array and use that
       if (checkPermissions === undefined) {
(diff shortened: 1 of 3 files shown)


foladipo:
@zenweasel @impactmass

I have found and written a fix for the problem reported in #2895. However, I want more info/history about why we wrote `hasPermissions` the way it was. In summary, why were we auto-appending `owner` to the list of permissions passed to `hasPermissions`? You may check [this link](https://github.com/reactioncommerce/reaction/pull/3196/files#diff-e0509424f4672c76deaab12707ba478d) for details about the code I'm talking about.

Similarly, I want to know if I can add some checks that handle when `hasPermissions` is called with an array of permissions. See [this link](https://github.com/reactioncommerce/reaction/blob/34b53cf33cff5ba8b90860caeb863b170d8f0eec/client/modules/core/main.js#L198) for more details.

impactmass:
Going through this now. A quick note on why `owner` role gets appended is so that the owner always have permission to perform an action regardless of the specific role being checked for. That is, when checking if a user `hasPermissions("createProduct")`, if that user is the owner of the shop, it should return `true`.

impactmass:
Another thing to note is that we have two `hasPermission`s one on the client (which you're working on right now), and one for the server in `/server/api/core/core.js` which you'll need to modify as well

client/modules/core/main.js
```diff
@@ -186,21 +186,33 @@ export default {
...
       // permissions can be either a string or an array
       // we'll force it into an array and use that
       if (checkPermissions === undefined) {
         permissions = ["owner"];
       } else if (typeof checkPermissions === "string") {
         permissions = [checkPermissions];
       } else {
+        /* TODO: Consult and add the code below if it is thought
+        of as useful.
+        if (!Array.isArray(checkPermissions)) {
```
impactmass: While it doesn't do any harm, I _think_ we can drop this check. I say that because the final method called -[Roles.userIsInRole](http://alanning.github.io/meteor-roles/files/roles_roles_common.js.html#l254)- can handle both a string and an array

client/modules/core/main.js
```diff
@@ -186,21 +186,33 @@ export default {
...
+          return false;
+        }
+        if (checkPermissions.length === 0) {
```
impactmass: The same comment above should apply to this too

client/modules/core/main.js
```diff
@@ -186,21 +186,33 @@ export default {
     // called out a userId is validated.
     //
     function roleCheck() {
+      // TODO: There're some outdated comments in this function. Clean them out.
```
impactmass: For the outdated comments, you can make that change in this PR (i.e remove them). That will start a discussion here about if it needs to change or not

client/modules/core/main.js
foladipo: Ok. Will do that.

client/modules/core/main.js
foladipo: Roger that. I've done some, but left with a few.

foladipo: But what should I do to [this commented code](https://github.com/reactioncommerce/reaction/blob/18ac96a140173753d12fbaf7b9d911a86503d31b/client/modules/core/main.js#L216)? I wasn't the one who wrote it, so I don't know it still is/will be useful.

foladipo:
@impactmass Roger that. I'll investigate more about what effects of not auto-appending `owner`. 

And about the server-side version of `hasPermissions`, I agree that we will have to work on that one too after this client-side version.

client/modules/core/main.js
impactmass: Looking at that piece of commented code, it's checking to see if a user has `admin` role for shops then based on that, it decides hasPermissions access (i.e return `true` or `false`). So, if a user has `admin` role, they get a pass. 
That can lead us to the same situation we're in with the `owner` role being concatenated in (i.e what you're fixing). I think that's why it was commented out. So, it's fine to remove it, since there are other ways to perform admin check.

foladipo: Done.

impactmass: Cool. Do you want me to do another review? or still working on other parts?

foladipo:
@impactmass I now understand why we were auto-appending `owner` to the array of permissions passed to `hasPermissions`. So, in order to still allow users with the `owner` permission to perform any action, I have pushed some changes. Please review and let me know if they satisfy the preceding condition WITHOUT introducing new problems.

client/modules/core/main.js
foladipo: Yes, you can perform another review. I've pushed changes to address the remaining part I was working on (see comments on main thread).

foladipo:
@impactmass I tested the commit `feb7a9c` above by adding these logging statements to the `render` method of [this component](https://github.com/reactioncommerce/reaction/blob/feb7a9c5f2d2bb1cd82db4370a7ef6263a88e63d/imports/plugins/included/product-detail-simple/client/components/childVariant.js) and then visiting the PDP page:
```javascript
console.log("Reaction.hasPermission('owner'): ", Reaction.hasPermission("owner"));
console.log("Reaction.hasPermission('anonymous'): ", Reaction.hasPermission("anonymous"));
console.log("Reaction.hasPermission(): ", Reaction.hasPermission());
console.log("Reaction.hasPermission(''): ", Reaction.hasPermission(""));
console.log("Reaction.hasPermission([]): ", Reaction.hasPermission([]));
console.log("Reaction.hasPermission(['owner']): ", Reaction.hasPermission(["owner"]));
console.log("Reaction.hasPermission(['owner', 'random']): ", Reaction.hasPermission(["owner", "random"]));
console.log("Reaction.hasPermission(['random', 'owner']): ", Reaction.hasPermission(["random", "owner"]));

```

Then I viewed the results in the browser console.

*NB*: Don't forget to import `Reaction` in the component above.

client/modules/core/main.js
```diff
@@ -180,60 +180,38 @@ export default {
...
-    // is the bulk of the logic
-    // called out a userId is validated.
-    //
+
+    // This is declared as a function for the sake of reuse. Note that it is
+    // only called after userId is validated.
     function roleCheck() {
       // permissions can be either a string or an array
       // we'll force it into an array and use that
+      // TODO: Should we also cater for checkPermissions === "" and checkPermissions === []?
```
impactmass: It doesn't hurt, but it can also be deferred to final Roles method check just like this other [comment](https://github.com/reactioncommerce/reaction/pull/3196#discussion_r148066088)

client/modules/core/main.js
```diff
@@ -243,21 +221,17 @@ export default {
...
     }
 
-    //
-    // actual logic block to check permissions
-    // we'll bypass unecessary checks during
-    // a user logging, as we'll check again
-    // when everything is ready
-    //
+    /*
+      Checks for permissions. We'll skip performing some checks if the user is
```
impactmass: I think we can leave this part of the comment as is. It seems valid and clear (IMO)

client/modules/core/main.js
```diff
@@ -243,21 +221,17 @@ export default {
...
+      currently logging in, as those will be done when log in is complete.
+    */
     if (Meteor.loggingIn() === false) {
-      //
       // this userId check happens because when logout
-      // occurs it takes a few cycles for a new anonymous user
+      // occurs. It takes a few cycles for a new anonymous user
```
impactmass: do you mean to use `,` here? seems like it should be a `,` instead of `.`

client/modules/core/main.js
```diff
@@ -180,60 +180,38 @@ export default {
...
+    // only called after userId is validated.
     function roleCheck() {
       // permissions can be either a string or an array
       // we'll force it into an array and use that
+      // TODO: Should we also cater for checkPermissions === "" and checkPermissions === []?
       if (checkPermissions === undefined) {
         permissions = ["owner"];
       } else if (typeof checkPermissions === "string") {
+        if (checkPermissions === "owner") {
+          return Roles.userIsInRole(userId, "owner", group);
```
impactmass: adding a detailed comment about this on main thread.

impactmass:
Just to be sure we are both on the same page, the goal here is to remove hard coding of 'owner' inside this function. So that way when it's called to check for a specific role, it is clear what role it's checking for. 
`Reaction.hasPermission("x")` should check if the user has role x. `Reaction.hasPermission(["x", "y"])` should check if the user has either one or both roles x, y.
etc. No implicit adding of `owner`. 

When there's need to give access based on role x, and we want to give owner access as well, we then just write `Reaction.hasPermission(["x", "owner"])` ... passing `owner` in explicitly. Are we on same page on that? I'm asking this because I still see some specific owner checks there. (or is there something else you're trying to achieve there?)

Those test cases you've written can also be fleshed out into a test file for the /server/api/core/core part (permissions.app-test.js maybe?)

client/modules/core/main.js
```diff
@@ -180,60 +180,38 @@ export default {
...
     let id = "";
     const userId = checkUserId || Meteor.userId();
-    //
-    // local roleCheck function
-    // is the bulk of the logic
-    // called out a userId is validated.
-    //
+
+    // This is declared as a function for the sake of reuse. Note that it is
+    // only called after userId is validated.
```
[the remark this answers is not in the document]
foladipo: Done.

client/modules/core/main.js
foladipo: Noted.

client/modules/core/main.js
foladipo: Ok. I'll restore the former comment.

client/modules/core/main.js
foladipo: Thanks, I'll correct that.

foladipo: Done.

foladipo:
@impactmass Yeah, I agree that we should require calls to `hasPermission` to explicitly list the permissions to check, and I have pushed commits with the code for that. Please review and share your thoughts.

About writing tests based off of [this comment](https://github.com/reactioncommerce/reaction/pull/3196#issuecomment-341572933), I'm on it.

foladipo:
## UPDATE

About converting [these logging statements](https://github.com/reactioncommerce/reaction/pull/3196#issuecomment-341572933) to tests, we can't do that. "Why?", you say. Because I created a `hasPermission.app-test.js` spec file and tried to run it, but Meteor didn't pick it up. So I consulted with @zenweasel and he told me that there is no way to run tests for client-side methods right now.

impactmass:
About the tests, note what I said (on the server):
```
Those test cases you've written can also be fleshed out into a test file for the /server/api/core/core part (permissions.app-test.js maybe?)
```

foladipo:
Oh, I see. I missed that later part. I will work on writing tests for the server part.

zenweasel:
@foladipo What's the status on this? You said you were going to work on writing tests 10 days ago? Should I assign this to someone else to take over?

foladipo:
@zenweasel Sorry for the delay. I've pushed the tests now.

foladipo:
@impactmass I have written tests for the server-side version of `hasPermissions` and I have also modified its code to stop auto-appending `owner` to the list of permissions it is supplied with. Now, the server-side `hasPermissions` also requires anyone calling it to explicitly list all the permissions it is to check for.

server/api/core/hasPermission.app-test.js
```diff
@@ -0,0 +1,63 @@
...
+    expect(hasPermissions).to.be.false;
+  });
+
+  it.only("should return false when `permissions` is an empty array", function () {
+    sandbox.stub(Meteor, "userId", () => "random-user-id");
+    const hasPermissions = Core.hasPermission([]);
+    expect(hasPermissions).to.be.false;
+  });
+
+  it.only("should return false for anonymous users when `permissions === [owner]`", function () {
```
impactmass: This test should be less about whether the user is anonymous, and more about whether they have the role/permission being checked for. You can make the test description clearer.

We expected it to return false in this case because anonymous users don't have the `owner` role.

server/api/core/hasPermission.app-test.js
```diff
@@ -0,0 +1,63 @@
...
+    const anonymousUser = Factory.create("anonymous");
+    sandbox.stub(Meteor, "userId", () => anonymousUser._id);
+    const hasPermissions = Core.hasPermission(["owner"]);
+    expect(hasPermissions).to.be.false;
+  });
+
+  it.only("should return false for registered users when `permissions === [owner]`", function () {
```
impactmass: comment above (about test description) can be applied to this too

server/api/core/hasPermission.app-test.js
```diff
@@ -0,0 +1,63 @@
+import { Meteor } from "meteor/meteor";
+import { expect } from "meteor/practicalmeteor:chai";
+import { sinon } from "meteor/practicalmeteor:sinon";
+import { Factory } from "meteor/dburles:factory";
+import Core from "./core";
+
+describe.only("hasPermission", function () {
```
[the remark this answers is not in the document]
impactmass: And all other `it.only`s

foladipo:
OK. I'm working on the comments now.

client/modules/core/main.js
```diff
@@ -177,63 +177,38 @@ export default {
...
-    // local roleCheck function
-    // is the bulk of the logic
-    // called out a userId is validated.
-    //
+
+    /**
+     * @method roleCheck
+     * @summary check whether or not a user is in a list of roles.
+     * @private
+     * @since 1.5.5
```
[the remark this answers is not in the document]
foladipo: `roleCheck` was added in `0.15.0`. You can check the changes made to `client/modules/core/main.js` in [this commit](https://github.com/reactioncommerce/reaction/commit/da4a379586c347a39adb3984d842490a210c7392#diff-e0509424f4672c76deaab12707ba478d) to verify. Now, you'll notice that that commit was made on `18 August, 2016`, and the next release after that, according to [this release history](https://github.com/reactioncommerce/reaction/releases?after=v0.18.0), was on `23 August, 2016`. So, `roleCheck` was added in `0.15.0`.

You can check if my findings above and let me know whether or not you think they are accurate.

client/modules/core/main.js
```diff
@@ -177,63 +177,38 @@ export default {
...
-    //
-    // check if a user id has been found
-    // in line 156 setTimeout
-    //
+    /**
+     * @method validateUserId
+     * @summary verifies that Meteor.userId() returns a value, after which
+     * it calls some functions, especially roleCheck.
+     * @private
+     * @since 1.5.5
```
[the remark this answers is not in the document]
foladipo: Done.

server/api/core/hasPermission.app-test.js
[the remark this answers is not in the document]
foladipo: I have tried removing the `.only`s and, after running `reaction test`, I saw some of the tests I wrote for this PR fail. What I am trying to figure out right now is why they fail solely when I remove the `.only`s and run all the tests of this project. That's where I'm at right now. And if you have suggestions about the cause, do share them.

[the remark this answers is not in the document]
zenweasel: These `only` are for local debugging only and should never be committed/pushed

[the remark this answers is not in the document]
foladipo: Roger that.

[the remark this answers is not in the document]
foladipo: @impactmass I have removed the `.only`s and fixed the failing tests.

foladipo:
@zenweasel @impactmass I am done with the last outstanding review. I have also added appropriate test instructions for this PR and removed the WIP tag. So, please review and share your thoughts.

server/api/core/hasPermission.app-test.js
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  it("should return true for anonymous users when `permissions === ['anonymous', 'guest']`", function () {
+    const user = Factory.create("anonymous");
+    sandbox.stub(Meteor, "userId", () => user._id);
+
+    // It is necessary to explicity supply shopId to the Core.hasPermission call
+    // below. If we don't, the said function defaults to using Core.getShopId
+    // which, when this test is run, will give a shop ID that will be different
+    // from the one used by Factory.create("anonymous").
```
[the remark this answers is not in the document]
impactmass: @zenweasel @spencern any thoughts?

server/api/core/hasPermission.app-test.js
```diff
@@ -0,0 +1,68 @@
...
+    const shopId = Object.keys(user.roles)[0];
```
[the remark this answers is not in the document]
foladipo: I'm not sure it's possible for that index to be 1. I based my decision to use index 0 off of the declaration of `Factory.define("anonymous")` in [this file](https://github.com/reactioncommerce/reaction/blob/b97b0757d7465e17166aaf7142fc0a2e829829e2/server/imports/fixtures/users.js). More specifically, if you check [this line](https://github.com/reactioncommerce/reaction/blob/b97b0757d7465e17166aaf7142fc0a2e829829e2/server/imports/fixtures/users.js#L102) in that file, you will see that `user.roles` can only have one key in it.

So, I do think this is the best approach. Or do you have another approach in mind?

server/api/core/hasPermission.app-test.js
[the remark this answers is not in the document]
foladipo: Well, being explicit with `shopId` too wouldn't hurt. The question is: is it necessary? And what are the scenarios that exist right now or may happen that will make it necessary?

Consider this scenario: we have users X, Y and Z they are respectively the admins of shops X, Y and Z. If user X has shop X open, is there a button somewhere that s/he can use to delete a product in shop Z? Because then, even though s/he is logged into shop X, this app will need to check if s/he has the `deleteProduct` permissions in a different shop, shop Z.

If scenarios similar to the above exist or can easily happen, then we definitely should require `shopId` to be explicitly passed to `hasPermission`.

server/api/core/hasPermission.app-test.js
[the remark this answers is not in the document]
impactmass: I meant that the fixture can change. A time will come when we make the fixtures multi-shop aware, then there will be more shopId keys (0,1 etc) there

server/api/core/hasPermission.app-test.js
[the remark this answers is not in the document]
impactmass: Thinking about this again, I think it's fine as we have it currently. With default values for shopId (getShopId) and user (Meteor.user). The edge cases seem rare and each can be handled.
(so let's take this comment thread as resolved).

server/api/core/hasPermission.app-test.js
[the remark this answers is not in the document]
foladipo: I think it will be better to wait till that change is made then. Coz until then, we can't know the structure of the new fixtures and the index of the right shopId to use above.

zenweasel:
@impactmass Is this ready to go now?

impactmass:
I think there's a bit more planning to be done about how this will get merged in, because this might cause some side effects in places where hasPermission is in use.
- how we can test that there is no breaking change in other parts of core using this method
- how we plan to release this out. considering other folks might have built on hasPermission, and this might be a breaking change for them.

This discussion should continue from the issue (and we can probably mark this as WIP?)

spencern:
@impactmass I appreciate that you're thinking about these things. This might need to be considered a breaking change in the sense that additional plugins or apps relying on `hasPermission` may no longer work as expected

aaronjudd:
@spencern @zenweasel @impactmass can we get a status update here? if this is not moving forward.. how about some closure?

aaronjudd:
I am closing this with extreme prejudice.

[WIP] Fix Reaction.hasPermissions for anonymous users

Re-open if this is a real thing, or give up...

PR #3218, opened 2017-11-01
Fix "Failed to get currency exchange rates" error in server

### LINKED ISSUE #3103: "Failed to get currency exchange rates" error in server
I get the `Failed to get currency exchange rates` error even though I don't have an Open Exchange Rates app ID set up. If I don't have that set up it should just not be fetching rates and not giving me that error.

### Steps to reproduce

1. Start Reaction without an open exchange rate ID loaded (e.g. don't use `reaction.json`).
1. Navigate around Reaction some (click on products, etc)
1. Observer the errors in the console

codejockie:
Resolves #3103 

This PR fixes the error logged on the server when the open exchange appId is not set. It ensures that the error message doesn't show up if appId isn't set.

## Test

1. Unset the Open Exchange API App ID by removing it from the `reaction.json` file
2. Do a `reaction reset` of the app
3. Run app after reset
4. Visit the home route `http://localhost:3000`
5. Check that on the terminal, you no longer see the error message "Failed to get currency exchange rates."


spencern:
eslint passes locally

Issue #3215, opened 2017-10-31
Order summary values being stored as strings instead of decimals


zenweasel:
### Expected behavior
Numbers should be stored as numbers

### Actual behavior
Numbers stored as Strings

### Steps to reproduce the behavior
1. Place an order
1. Open Robomongo or similar
1. Go to billing/invoice and observe the values are stored as string
<img width="678" alt="robo_3t_-_1_1" src="https://user-images.githubusercontent.com/72819/32242485-2ff1bca4-be30-11e7-83cf-472a5270e0f6.png">

Additionally, the Schemas is set for validate these values are set to be decimals, however the entire billing/payment record is set to `blackbox: true` so the schema is ignored. There may be some items in payment we need to blackbox (like transactions) but we should reenabled as much of these schema validations as possible.

### Versions
```
Node: 7.10.0
NPM: 5.4.2
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.19.1
Reaction: 1.5.4
Reaction branch: master
Docker: 17.09.0-ce
```

PR #3222, opened 2017-11-02
fix order values store as string

iakhator:
Resolve issue #3215 

Fix Order values storing as `String` in the database even when the schema is set as a `Number`.

#### Fix
- In the Cart Schema there is a `billing` property which has a validation of blackbox set as true is remove.
- In the PaymentMethod Schema  the createAt property with validation of denyUpdate set to true is also remove.

#### Test.
- Place an order
- Go to billing/invoice and observe the values are stored as Double
screen shot 2017-11-02 at 4 00 05 pm


zenweasel:
It seems like removing the `denyOnUpdate` is going to mean that every time you update the record the `createdAt` record will be set which is not what we want. Can you verify this will not happen?

iakhator:
okay, I will check it out

iakhator:
I figured out that `denyUpdate` is set in the paymentMethod schema which only needs a created date. Because payment is made once after checking out. removing the `denyUpdate` for the paymentMethod schema does not update.

zenweasel:
@iakhator You have some tests failing. Lots of them.

iakhator:
Right. I will attend to the failing test.

iakhator:
Resolved.

Issue #3223, opened 2017-11-02
Components.Alert reconciliation issue when more than two instances are displayed


prinzdezibel:
Ensure that three Alert boxes are displayed on page.

Wait for approximately 10 seconds to see the boxes disappearing again.

### Expected behavior:

All 3 boxes disappear.

### Actual behaviour:

Only 2 boxes are disappearing. The 2nd box will remain visible.

prinzdezibel:
![screen shot 2017-11-02 at 17 43 34](https://user-images.githubusercontent.com/1733229/32338546-63e7cc8a-bff5-11e7-83ec-5377ed9a883c.png)

prinzdezibel:
Fixed with PR #3224

PR #3224, opened 2017-11-02
Clear all alerts after time has elapsed

prinzdezibel:


zenweasel:
@prinzdezibel Please include in your PR comments the following:
1. "Closes #Issue" (this will automatically link the issue in GH)
1. A short summary of what the PR fixes/changes when relevant (which is usually)
1. Steps to take to test the PR
1. Since you are a Contributor you can branch directly off Reaction, which makes it easy for us to make small changes (like lint errors, etc) w/o need to go for another round of review/changes.

prinzdezibel:
Closes #3223 

Summary: Don't use index as key for React components. This will not work when the position of the node in the DOM changes (e.g. in a list), because the reconciliation process can't know that the identity of the node actually changed. Instead we're using a mongo database id to unambiguously designate node identity.

Testing this PR: See #3223

PR #3228, opened 2017-11-02
Tags.updatedAt should set when not specified

zenweasel:
This should be an auto-populated field when not specified so that Tags can be easily populated from fixtures w/o specifying a value.

### DIFF
--- a/lib/collections/schemas/tags.js
+++ b/lib/collections/schemas/tags.js
@@ -86,7 +86,10 @@ export const Tag = new SimpleSchema({
     }
   },
   updatedAt: {
-    type: Date
+    type: Date,
+    autoValue: function () {
+      return new Date;
+    }
   }
 });
 


lib/collections/schemas/tags.js
```diff
@@ -86,7 +86,10 @@ export const Tag = new SimpleSchema({
     }
   },
   updatedAt: {
-    type: Date
+    type: Date,
+    autoValue: function () {
+      return new Date;
```
spencern: should this be `new Date();`?

zenweasel: I don't believe so. I don't see us doing that elsewhere. e.g. in `revisions.js`
```
  updatedAt: {
    type: Date,
    autoValue: function () {
      return new Date;
    },
    optional: true
  },
```

although the [collection2 docs](https://github.com/aldeed/meteor-collection2#autovalue) show `new Date()` so I am guessing both forms are correct?

spencern:
bitHound issue is resolved in master and not caused by this branch.

PR #3231, opened 2017-11-03
Allow for the addition of OAuth Providers

pmn4:
**Goal**: use an OAuth Provider other than Facebook/Google/Twitter/Instagram
**Problem**: [the set of providers used to create the settings dashboard is inaccessible](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/accounts/client/helpers/util.js#L10-L15)
**Proposed Solution**: offer a way to add to the list without exposing it outright

To test, I created a tiny plugin which adds Untappd (Foursquare for beer) and looks something like this:
``` bash
meteor add brewline:accounts-untappd
```

plugins/custom/untappd-oauth/client/index.js:
``` javascript
import { ServiceConfigHelper } from "../../../core/accounts/client/helpers/util";
// (unrelated) is there a better path to use here? going up the directory tree feels unfortunate

ServiceConfigHelper.addProvider("Untappd", [
  { property: "clientId", label: "Client ID" },
  { property: "secret", label: "Client Secret" }
]);
```


pmn4:
(bitHound shows one failing file, but when you go to the site, the list of failing files is empty... :thinking_face:)

imports/plugins/core/accounts/client/helpers/util.js
```diff
@@ -102,4 +102,12 @@ export class ServiceConfigHelper {
       return service;
     });
   }
+
```
spencern: need jsdocs here

zenweasel:
Running `eslint` on this locally shows no errors so I don't know what's up with Bithound. 👎

machikoyasuda:
Woot @pmn4 🎊

Issue #2917, opened 2017-09-21
Deprecate aldeed:simple-schema and implement npm simpl-schema


aaronjudd:
Deprecate https://github.com/aldeed/meteor-simple-schema and replace  with https://github.com/aldeed/node-simple-schema

We've tested a few times and there were some issues, but it's time to revisit, as we need this in place for our react components to use for validation, and also in some new registry requirements we're working on.   This would also allow us to deprecate meteor autoform as soon as the remaining dashboard blaze templates are converted.

zenweasel:
fwiw I played with this recently and the main issue I ran into was that `node-simple-schema` does not support things like `[MyCustomSchema]` which we have a fair amount of.

spencern:
I've gone ahead and replaced all instances of `type: [String]` or any other arrays that are built like that in our current code base.
See my branch `spencer-node-simple-schema-update-schemas`

Current issues I'm running into:

it appears that the node version does not support our `index` property

```
Error: Invalid definition for userId field: "index" is not a supported property
    at /Users/sn/dev/reaction-dev/node_modules/simpl-schema/dist/SimpleSchema.js:965:13
    at Function._.each._.forEach (/Users/sn/dev/reaction-dev/node_modules/underscore/underscore.js:158:9)
    at checkAndScrubDefinition (/Users/sn/dev/reaction-dev/node_modules/simpl-schema/dist/SimpleSchema.js:963:24)
    at /Users/sn/dev/reaction-dev/node_modules/simpl-schema/dist/SimpleSchema.js:498:9
    at Function._.each._.forEach (/Users/sn/dev/reaction-dev/node_modules/underscore/underscore.js:158:9)
    at SimpleSchema.extend (/Users/sn/dev/reaction-dev/node_modules/simpl-schema/dist/SimpleSchema.js:486:28)
    at new SimpleSchema (/Users/sn/dev/reaction-dev/node_modules/simpl-schema/dist/SimpleSchema.js:107:10)
    at meteorInstall.lib.collections.schemas.accounts.js (lib/collections/schemas/accounts.js:74:25)
    at fileEvaluate (packages/modules-runtime.js:333:9)
    at require (packages/modules-runtime.js:228:16)
=> Exited with code: 1
```

From @mikemurray, we need to be aware of how our [Validation](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/collections/lib/validation.js#L5) works with this. I've not looked into this at all yet.

[...]

PR #3250, opened 2017-11-07
Use Registered Schemas instead of Schema Source

pmn4:
## Allow for projects to modify Schemas

Creating a new database schema is simple enough, however, in order for the frontend to honor it (through Validation), we need a way for core components to use the updated schemas.
If core components refer to the Schema Registry instead of importing from the Schema definition files, this can be achieved.

Take, for instance, the `VariantFormContainer`. In it, [a `Validation` object is initialized](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/included/product-variant/containers/variantFormContainer.js#L24) using the `ProductVariant` schema.

If one were to add a field to the schema*, however, that Validation would reject changes to the new properties. (I understand that for this simple example, I could extend the Container and simply overwrite the constructor, but, you know, I'd rather not!).

If we replace:
``` JavaScript
import { ProductVariant } from "/lib/collections/schemas/products";
```
with 
``` JavaScript
import { Schemas } from "@reactioncommerce/reaction-collections";
const ProductVariant = Schemas.ProductVariant;
```
then a new `ProductVariant` schema can replace of the original, and the Container could remain untouched.

let me know what you think!

_(If you are comfortable with these changes, I will write some tests)_

---
 \* - assumption is that Schema changes use the `registerSchema` pattern to replace existing schemas, not monkey-patching the original schema.

### DIFF
--- a/imports/plugins/core/discounts/lib/collections/schemas/config.js
+++ b/imports/plugins/core/discounts/lib/collections/schemas/config.js
@@ -3,14 +3,12 @@ import { Schemas } from "@reactioncommerce/reaction-collections";
 import { Discounts } from "./discounts";
 import { registerSchema } from "@reactioncommerce/reaction-collections";
 
-const PackageConfig = Schemas.PackageConfig;
-
 /**
 * DiscountsPackageConfig Schema
 */
 
 export const DiscountsPackageConfig = new SimpleSchema([
-  PackageConfig, {
+  Schemas.PackageConfig, {
     "settings.rates": {
       type: Object,
       optional: true
--- a/imports/plugins/core/discounts/lib/collections/schemas/config.js
+++ b/imports/plugins/core/discounts/lib/collections/schemas/config.js
@@ -1,7 +1,9 @@
 import { SimpleSchema } from "meteor/aldeed:simple-schema";
-import { PackageConfig } from "/lib/collections/schemas/registry";
+import { Schemas } from "@reactioncommerce/reaction-collections";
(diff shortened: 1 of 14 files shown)


zenweasel:
@pmn4 I like the idea of this approach (essentially creating a global that can be overridden) but at first glance I wonder if all instance of schema use have been updated. I think pretty much every included plugin (at least all the ones that have settings) use a schema. So we would want to make sure that change is made everywhere.

spencern:
This approach sounds reasonable, but I'm not certain it will be necessary after #2917 is finished.

Either way, we should wait to implement this until after @aldeed has finished the conversion to using the npm version of `simpl-schema`

pmn4:
cool. looking forward to it!

aldeed:
Yeah this overlaps the changes I've made, but I can rebase/adjust my branch. I've finished all changes I think, but I'm still investigating and resolving a few bugs that it created. Will hopefully have a PR in the next few days.

spencern:
@pmn4 We're about ready to merge #2917 - once that happens will you verify if it resolves your needs or if we need to update and revisit this PR?

aaronjudd:
@reactioncommerce/community @aldeed Closing this PR, we should re-open, re-review for 1.9 release

zenweasel:
I wonder if this isn't superceeded by: https://github.com/reactioncommerce/reaction/issues/3494. I feel like that ticket covers what this PR is trying to accomplish?

aldeed:
Agree that new schema extension model should address this

PR #3252, opened 2017-11-08
Added customers import to shopify connector

wayann:
@zenweasel as said earlier in the chat I've tested successfully this import with 22k customers and 620 products.
6c36d4fc-eb3a-4e1d-a5da-40294cb3c14c


kieha:
@spencern @wayann there's no major overlap here with what I'm working on. 

As you said, instead of having 2 import buttons, the checkboxes should be used to decide what to import, then the single import button handles all the selected options.

As for the React conversion, the only thing that I'd need to add UI-wise is the `customers`checkbox, since I've tied the import button to a single method that's called with the selected checkboxes, and it calls the respective import methods (needs to be refined).

wayann:
@spencern @kieha thanks for considering my PR
glad I can contribute to this amazing project.  👍 

Single import button [commit] (https://github.com/reactioncommerce/reaction/pull/3252/commits/46e607c75083c05d8e9d39b0fc642f5578c22e01) 

I hope this won't break React conversion from @kieha 

it crashed several times when importing 22k customers, 
so at this point makes totally sense to handle products, customers with jobs.

just need to understand how @spencern did it with the images jobs, 
I'm fairly new to Meteor and Reaction.

wayann:
@kieckhafer 🥇 👍  ;)

PR #3255, opened 2017-11-08
Remove unused Blaze templates.

prinzdezibel:
The following two Blaze templates are no longer needed.

      1) layoutHeader
         This template was used for most cases (coreLayout). But since
         coreLayout.js does include the new Components.NavBar React
         component, it's obsolete
      2) checkoutHeader
         This was a customized version of the header for the checkout
         workflow. Nowadays the checkout workflow also uses the standard
         NavBar component. Therefore it's obsolete, too.


mikemurray:
@prinzdezibel which I think you may have fixed in: https://github.com/reactioncommerce/reaction/pull/3257/files#diff-9ebade6744b20e86dd07f325d59075d0R22 

It seems like coreLayout should be pulling a template for the header from the layout.

prinzdezibel:
@mikemurray I think this is already broken in master. Should I fix it as part of this ticket or do you think it makes sense to have a new issue for that?

prinzdezibel:
@zenweasel Hi Brent, @mikemurray made the suggestion to fix this behaviour (the checkout navbar should be a dedicated one without tag bar) right away as part of this ticket. If you agree, I could do it tomorrow. Alternatively, we could create a new issue, because it's broken in master anyway. Please decide and let me know. thanks.

zenweasel:
@prinzdezibel Yes, If we can fix this quickly in this PR let's do it. But if it turns into more than a couple hours of extra work let's break it off into another issues, especially if it's something already broke in `master`.

prinzdezibel:
I've searched the code base for an visual component (React or Blaze) that represents what formerly was known as `checkoutHeader`. But I can't find anything. Probably it would make sense to just add options to the existing NavBar and strip it down when using in checkout context. But I don't know how this is supposed to look. @rymorgan: Is it just what the regular navigation bar (called NavBar) does, except for the tag-bar not visible.

prinzdezibel:
@zenweasel Maybe it makes more sense to have this in a separate ticket, because I can't it fix straight away without further help.

rymorgan:
Having the full navigation bar show up at the checkout is definitely a regression and an important one to fix. I'm not sure where or how the navigation started appearing here. The only thing that should show once someone is in the checkout is the logo area in the upper left. 

checkout

prinzdezibel:
@mikemurray  @zenweasel  : This PR is obsolete and is superseded by this PR:
https://github.com/reactioncommerce/reaction/pull/3257

I've implemented @rymorgan 's design into the other PR


I'm going to close this one.

PR #3261, opened 2017-11-09
Dropdown aligns with the button

Akarshit:
### Current view
screen shot 2017-11-09 at 5 12 04 pm

### View after this fix
screen shot 2017-11-09 at 5 10 00 pm

### Work done
Removed the padding and increased the width of button, so that all the parent components adjust their width.

### To test
1. Run `reaction` click on the `Language select` dropdown and `Currency select` dropdown.
2. Note that they have consistent dropdown position now.


imports/plugins/included/default-theme/client/styles/navbar.less
```diff
@@ -340,3 +338,7 @@ html:not(.rtl) .rui.navbar .languages .dropdown-menu {
   border: solid rgba(129, 122, 122, 0.62) 2px;
   box-sizing: content-box !important;
 }
+
+.dropdown-toggle button.rui.btn.btn-default.flat {
```
mikemurray: Selector seems way too specific. Can it not be simplified?

Akarshit: It can be, but I chose it to be that specific so that nothing breaks in some other place later. As such the selector can be just `.dropdown-toggle button`.

Should I modify it?

zenweasel: @mikemurray Can you reply to @Akarshit here?

zenweasel: @mikemurray Hey Mike, could you reply to @Akarshit here? Second call.

mikemurray: @Akarshit could you use`.rui.navbar .dropdown-toggle .btn`? Just so it's not specific to button states?

Akarshit: @mikemurray Ok will do.

spencern:
@Akarshit is this ready for another review?

Akarshit:
@spencern Yes.

PR #3267, opened 2017-11-09
Checking if the variant belongs to current product

Akarshit:
### Problem
The variantId was not getting updated when selecting a new product.

### Solution
In the check for updating the product/variant instead of only checking if `variantId == null` also check that `variantId` belong to the selected product

### Another solution
Maybe a better solution is to call `ReactionProduct.setCurrentVariant(null)` as soon as some PDP page is opened.

### To test 
1. Run `reaction`
2. Add image to example-product and publish.
3. Add another product with image(say test-product).
4. Go to home and then select example-product(image should show).
5. Go to home and select test-product(image should show).
6. Do a refresh(image should show).

### DIFF
--- a/lib/api/products.js
+++ b/lib/api/products.js
@@ -128,7 +128,7 @@ ReactionProduct.setProduct = (currentProductId, currentVariantId) => {
   if (product) {
     // set the default variant
     // as the default.
-    if (!variantId) {
+    if (!variantId || !variantIsSelected(variantId)) {
       const variants = ReactionProduct.getTopVariants(productId);
       variantId = Array.isArray(variants) && variants.length &&
         variants[0]._id || null;


zenweasel:
I think this is a workaround, as we shouldn't be getting a selectedVariant that's not associated with the product, but this fixes several pretty serious bug and fixing the source of the issue may involve rewriting some core PDP stuff.

janus-reith:
Could you test again?
This is broken for me, but could also be related to my pdp variant modifications.
Accesing the main page works, but if I try to directly browse to a variant by handle, like
`http://localhost:3000/product/handle/id`, the pdp just shows the loading spinner and console throws:

```
meteor.js?hash=cbcc712d51de4298c275e8dcf25c66c29914f19a:992 TypeError: Cannot read property 'indexOf' of undefined
    at _variantIsSelected (products.js:84)
    at ReactiveDict.ReactionProduct.setProduct (products.js:169)
    at composer (productDetail.js:652)
    at tracker.js:37
    at Tracker.Computation._compute (tracker.js?hash=997515fa2d5b0530ba07741da556c4b36963ef3b:339)
    at Tracker.Computation._recompute (tracker.js?hash=997515fa2d5b0530ba07741da556c4b36963ef3b:358)
    at Object.Tracker._runFlush (tracker.js?hash=997515fa2d5b0530ba07741da556c4b36963ef3b:532)
    at onGlobalMessage (meteor.js?hash=cbcc712d51de4298c275e8dcf25c66c29914f19a:448)
```

zenweasel:
@janus-reith I believe we dropped support for browsing to particular variant by URL a while ago. Is that what you mean?

Since this is already released, if you have a problem can you open a new issue?

PR #3272, opened 2017-11-09
Remove extraneous blank line

spencern:
Removes an extraneous blank line. Wouldn't be a big deal, but I updated this in a downstream project, so it should also get updated in Core to avoid conflicts. 
🤦‍♂️


mikemurray:
masterpiece

PR #3278, opened 2017-11-10
Checking for AWS setting before deploying

Akarshit:
### Solution
Before deploying to `AWS_DOC_BUCKET` check if those env. variables are defined. This would prevent the build from failing.

### To test
1. Fetch this branch and merge it with master(on local system)
2. Install the CI command locally and manually run CI on master(circleci build)
3. The build will not fail.


.circleci/config.yml
```diff
@@ -62,7 +62,11 @@ jobs:
           name: Deploy to S3 if tests pass and branch is Master
           command: |
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
-              aws s3 sync /tmp/reaction-docs ${API_DOC_BUCKET} --delete --region ${API_DOC_BUCKET_REGION}
+              if [[ "${API_DOC_BUCKET}" && "${API_DOC_BUCKET_REGION}" ]]; then
```
spencern: I'm not certain this will ensure that `${API_DOC_BUCKET}` and `${API_DOC_BUCKET_REGION}` are set in bash.

https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash

Akarshit:
@spencern I tested in by setting the variables in `environment:` of `config.yml`. Let me try setting these variables in the bash too.

.circleci/config.yml
aaronjudd: so
```
if [ "${CIRCLE_BRANCH}" == "master" ] && [[ -v API_DOC_BUCKET ]] ;  then
```

jshimko: That `-f` in the second example checks if the variable resolves to a path to a file that exists.  

https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html

Side note on that... it would also require quotes because of the single brackets.  If it ever resolved to `undefined`, that becomes an invalid bash expression.  

So, when not set, this...

```sh
if [ -f $API_DOC_BUCKET ];
```

is evaluated as...

```sh
if [ -f ];
```

Which is invalid and would tank the script.

Anyway, for the sake of simplicity and flexibility, I usually prefer the double brackets with nothing more than the var names.  Quotes are optional and you can do any amount of conditions within one set of brackets.  That said, his original code is valid.

```bash
if [[ $API_DOC_BUCKET && $API_DOC_BUCKET_REGION ]]; then
```
You can give it a try by creating a test script and set/unset the values...

```bash
#!/bin/bash

MY_VAR="stuff"
MY_VAR2="things"

if [[ $MY_VAR && $MY_VAR2 ]]; then
  echo "Both vars are set!"
else
  echo "Nope! One or both vars isn't set"
fi
```

So anyway, I think we're good as far as logic goes. I guess my only nitpick is the message when the vars aren't set.  It could use a little more detail.

Maybe something like...

```sh
echo "S3 bucket configuration not found for jsdocs"
echo "Set API_DOC_BUCKET and API_DOC_BUCKET_REGION to build and deploy jsdocs to S3"
```

jshimko: That `-f` in the second example checks if the variable resolves to a path to a file that exists.  

https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html

Side note on that... it would also require quotes because of the single brackets.  If it ever resolved to `undefined`, that becomes an invalid bash expression.  

So, when not set, this...

```sh
if [ -f $API_DOC_BUCKET ];
```

is evaluated as...

```sh
if [ -f ];
```

Which is invalid and would tank the script.

Anyway, for the sake of simplicity and flexibility, I usually prefer the double brackets with nothing more than the var names.  Quotes are optional and you can do any amount of conditions within one set of brackets.  That said, his original code is valid.

```bash
if [[ $API_DOC_BUCKET && $API_DOC_BUCKET_REGION ]]; then
```
You can give it a try by creating a test script and set/unset the values...

```bash
#!/bin/bash

MY_VAR="stuff"
MY_VAR2="things"

if [[ $MY_VAR && $MY_VAR2 ]]; then
  echo "Both vars are set!"
else
  echo "Nope! One or both vars isn't set"
fi
```

So anyway, I think we're good as far as logic goes. I guess my only nitpick is the message when the vars aren't set.  It could use a little more detail.

Maybe something like...

```sh
echo "S3 bucket configuration not found for jsdocs"
echo "Set API_DOC_BUCKET and API_DOC_BUCKET_REGION to build and deploy jsdocs to S3"
```

Akarshit: @jshimko Also I would like to say that the following scipt
```bash
#!/bin/bash

MY_VAR="stuff"
MY_VAR2="things"

if [[ $MY_VAR && $MY_VAR2 ]]; then
  echo "Both vars are set!"
else
  echo "Nope! One or both vars isn't set"
fi
```
will output `Nope! One or both vars isn't set`, which is what we need.

I will update the message.

jshimko:
This would also fail if AWS credentials weren't provided, so we probably want to add those to the check as well.

jshimko:
Before we merge this, I think we need to add a check for the AWS credential env vars too.  Otherwise, this would fail tests where `API_DOC_BUCKET` and `API_DOC_BUCKET_REGION` were set, but credentials weren't.  And the logs probably wouldn't give you useful enough feedback about how to fix that.

I only mention it because those credential environment variables don't exist anywhere in the CI scripts, so there's no easy way for users to find what variables are supposed to be set without having to go search through the AWS docs.  And there are no code comments mentioning what values you should be setting in your CircleCI dashboard.  For example, like I did here for Docker...

https://github.com/reactioncommerce/reaction/blob/master/.circleci/deploy.sh#L3-L12

Since this is an external branch from an outside contributor, I think we could also just merge this to the release branch and make those updates there.

spencern:
I've already merged this into the release-1.6.0 branch on Aaron's approval. Is this a quick check you can write before we merge the release branch into master and tag a new release?

aaronjudd:
I tested with and without variables set, and with only one variable set, and if you have done that much then I suspect that you understand what this **internal**  mechanism is doing.

Issue #3227, opened 2017-11-02
Image for variant not showing when clicking on a product


arsen3d:
### Expected behavior
Product images to show
### Actual behavior
Placeholder images show
### Steps to reproduce the behavior
Click between products
https://hello.reactioncommerce.com/
Do not use back button
Do not click refresh.
Clicking refresh brings the image in.
### Versions
1.5
### Notes
I think the problem is at ReactionProduct in /lib/api/products.js.
The bug can be seen at function composer(props, onData) in imports/plugins/included/product-detail-simple/client/containers/social.js
//selectedProduct returns correct product
    const product = ReactionProduct.selectedProduct() || {};
//selectedVariant return wrong variant.
    const selectedVariant = ReactionProduct.selectedVariant() || {};

arsen3d:
![screen recording 2017-11-02 at 12 09 53 pm](https://user-images.githubusercontent.com/298235/32345356-27b258e4-bfc7-11e7-82fd-303dad714c10.gif)

arsen3d:
![screen recording 2017-11-02 at 12 09 53 pm 2](https://user-images.githubusercontent.com/298235/32345484-9006f76a-bfc7-11e7-8283-9dba379c85cd.gif)

arsen3d:
Can someone label this as a bug with critical impact.

arsen3d:
Note the file name in debugger, the function being called, and that varientId in console does not change as I click between products.
![screen recording 2017-11-02 at 1 47 53 pm](https://user-images.githubusercontent.com/298235/32349580-c045efdc-bfd4-11e7-870d-55b1d2e5750a.gif)

arsen3d:
Looks like the video is too blurry. Here is a screenshot
![screen recording 2017-11-02 at 1 47 53 pm](https://user-images.githubusercontent.com/298235/32349679-212eec18-bfd5-11e7-926d-85c99aec3353.gif)

arsen3d:
As you can see in the commented code, I've attempted  to resolve the problem by waiting for subscription to be ready. This did not solve the problem as it appears that ReactionProduct is being synchronized with autopublish in the background. Though it's not entirely clear to me about exactly how the underlying components are working.

speedingdeer:
Same here

arsen3d:
@speedingdeer I have not tested that. I suggest you make a new issue out of it. Make a video/gif if you can.

speedingdeer:
[...]

PR #3279, opened 2017-11-10
Get proper variantId so you can show proper image

### LINKED ISSUE #3263: Cart items picture displays placeholder (sometimes) when product has variants
### Expected behavior
Cart displays the selected product variant's picture
### Actual behavior
Cart display nothing (sometimes) instead of selected product variant's picture
Create few product variants and add pictures for each of them. Next as user select few and them to cart. Click on the cart image and check if the pictures in cart match the ordered variant pictures.

screen shot 2017-11-09 at 12 53 34

### Versions
Node: 8.1.4
NPM: 5.0.3
Meteor Node: 4.8.5
Meteor NPM: 4.6.1
Reaction CLI: 0.20.0
Reaction: 1.5.6
Reaction branch: master
Docker: 17.09.0-ce

zenweasel:
Resolves #3263 and I believe #3227 as well

### Probems

1. Media query was not searching for image from both the variant and the product. So when the appropriate image was at the product level it would not show. (this may not affect images that are uploaded, but when importing images it wasn't working)
1. `ReactionProduct.setProduct` was constantly just calling itself with the wrong variantId, which would never change until you refreshed so `selectedVariant` was `null` so no `Media` was found. I just eliminated this parameter since getting it from the router worked 100% of the time.

### To Test

1. Add several products with a single image at the product level
1. Navigate to various products.
1. Observe that the proper image is always being shown without having to refresh

### DIFF
--- a/imports/plugins/included/product-detail-simple/client/containers/productDetail.js
+++ b/imports/plugins/included/product-detail-simple/client/containers/productDetail.js
@@ -322,7 +322,10 @@ function composer(props, onData) {
           // Loop through ancestors in reverse to find a variant that has media to use
           for (const ancestor of selectedVariant.ancestors.reverse()) {
             const media = Media.find({
-              "metadata.variantId": ancestor
+              $or: [
+                { "metadata.variantId": ancestor },
+                { "metadata.productId": ancestor }
+              ]
             }, {
               sort: {
                 "metadata.priority": 1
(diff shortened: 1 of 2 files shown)


Akarshit:
@zenweasel If we are going for this solution instead of #3267, may be we should also change [here](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/included/product-detail-simple/client/containers/productDetail.js#L287)(and the logic around it) so that there is no confusion.

zenweasel:
@Akarshit I didn't see your PR for this. I do think my solution gets more to the root of the problem but agree we should probably look at where it's being set.

mikemurray:
@zenweasel Removed myself as a reviewer as I will not be available to review within the next week.

mikemurray:
one

PR #3282, opened 2017-11-13
Outbound Shopify Order Sync

### LINKED ISSUE #3322: Outbound Shopify Order Sync
1. Configure/Enable outbound hook in Connector settings
1. On order/create send the order to shopify.
1. Additionally ???

Part of Phase two of #2718

zenweasel:
Resolve #3322 

### How to test

**single shop**
1. Go to the Connectors dashboard panel and Enable export hooks
1. Place an order
1. Observe that the order is exported to Shopify
1. Observe that the order is marked exported

**Multi-shop**
1. Create a second shop and a product within that shop.
1. Go to the second shop and configure the slug for that shop and enable hooks
1. Place an order with products from both the primary and secondary shops
1. Observe that the order is exported to each individual shop separately


zenweasel:
putting this comment in to see if GH will start recording my commits properly

zenweasel:
Making progress. Off for the holiday now but hoping to have this done by c.o.b. Monday

zenweasel:
Complete except for JSDocs

imports/plugins/included/connectors-shopify/client/settings/shopify.js
```diff
@@ -91,6 +91,23 @@ Template.shopifySync.helpers({
 
     // If no webhooks are found, this integration is inactive
     return "";
+  },
+
+  hookIsActive(hook) {
+    const { settings } = Reaction.getPackageSettings("reaction-connectors-shopify");
+    const { synchooks } = settings;
```
spencern: `synchooks` has different casing than `syncHook`

imports/plugins/included/connectors-shopify/lib/collections/schemas/shopifyConnect.js
```diff
@@ -86,6 +102,11 @@ export const ShopifyConnectPackageConfig = new SimpleSchema([
       type: [Webhook],
       label: "Registered Shopify webhooks",
       optional: true
+    },
+    "settings.synchooks": {
```
spencern: Why are we calling them `synchooks` in some place and `pushooks` in others?

imports/plugins/included/connectors-shopify/server/hooks/orders.js
```diff
@@ -0,0 +1,278 @@
...
+
+/**
+ * @private
+ * @summary build a new object for export to Shopify
+ * @param {Object} doc - the order to convert
+ * @param {Number} index - the sequence number of billing/shipping records we are processing
+ * @param {String} shopId - the Id of the shop we are processing for
+ * @returns {Object} shopifyOrder - the converted order
+ */
+function convertRcOrderToShopifyOrder(doc, index, shopId) {
```
spencern: I'd prefer using `convertReactionOrderToShopifyOrder` unless that's so long that it's breaking something

imports/plugins/included/connectors-shopify/server/hooks/orders.js
```diff
@@ -0,0 +1,278 @@
...
+  shopifyOrder.customer = convertCustomer(billingAddress, order);
+  shopifyOrder.email = order.email;
+  const paymentType = order.billing[index].method;
+  if (paymentType === "credit" && order.billing[index].mode === "authorize") {
+    shopifyOrder.financial_status = "authorized";
+  } else {
+    shopifyOrder.financial_status = "paid";
+  }
+  shopifyOrder.id = order._id;
+  const itemsForShop = order.getItemsByShop()[shopId];
```
spencern: This would be easier to read if all of the variables that are not part of `shopifyOrder` were together and the `shopifyOrder` properties were all set together

imports/plugins/included/connectors-shopify/server/hooks/orders.js
```diff
@@ -0,0 +1,278 @@
...
+ * @param {Number} weight - the original weight value
+ * @param {String} shopId - the shop we are converting for
+ * @returns {Number} the normalized value
+ */
+function normalizeWeight(weight, shopId) {
+  // if weight is not grams, convert to grams if is grams just return
+  const shop = Shops.findOne(shopId);
+  const { baseUOM } = shop;
+  // Could've used switch, this was easier to read
+  // should these be a utilities module somewhere?
```
spencern: Yes! Should definitely add these to a util module somewhere. I've thought the same thing several times.

imports/plugins/included/connectors-shopify/server/hooks/orders.js
```diff
@@ -0,0 +1,278 @@
...
+    lineItem.fulfillment_service = "manual";
+    lineItem.fullfillment_status = null;
+    if (item.parcel && item.parcel.weight) {
+      lineItem.grams = normalizeWeight(item.parcel.weight, item.shopId);
+    }
+    lineItem.id = item._id;
+    lineItem.product_id = item.productId;
+    lineItem.quantity = item.quantity;
+    lineItem.requires_shipping = item.product.requiresShipping;
+    // lineItem.sku = ??? Not sure what should be the SKU here
```
spencern: Reaction has a SKU, but I'm not sure it's used or set currently.
https://github.com/reactioncommerce/reaction/blob/master/lib/collections/schemas/products.js#L300

imports/plugins/included/connectors-shopify/server/hooks/orders.js
```diff
@@ -0,0 +1,278 @@
...
+
+
+Orders.after.insert((userId, doc) => {
+  const pkgData = Reaction.getPackageSettings("reaction-connectors-shopify");
+  if (pkgData) {
+    const { settings } = pkgData;
+    const { synchooks } = settings;
+    if (synchooks) {
+      synchooks.forEach((hook) => {
+        if (hook.topic === "orders" && hook.event === "orders/create") {
```
spencern: seems like maybe you could just filter all the synchooks with a specific topic / event / type and eliminate some if these if blocks here.

imports/plugins/included/connectors-shopify/server/hooks/orders.js
```diff
@@ -0,0 +1,278 @@
...
+          if (hook.syncType === "exportToShopify") { // should this just be dynamic?
```
spencern: I think it should be dynamic. Right now it seems like the idea of a synchook exists in theory, but you'd still need to add code to this hook to add a different hook.

imports/plugins/included/connectors-shopify/server/methods/sync/hooks.js
```diff
@@ -0,0 +1,25 @@
+import { Meteor } from "meteor/meteor";
+import { check } from "meteor/check";
+import { Reaction } from "/server/api";
+import { Packages } from "/lib/collections";
+
+export const methods = {
+  "synchooks/shopify/addHook": (hook) => {
```
spencern: Security?

imports/plugins/included/connectors-shopify/server/methods/sync/hooks.js
```diff
@@ -0,0 +1,25 @@
...
+    check(hook, { name: String, checked: Boolean });
+    const [ topic, event, syncType ] = hook.name.split(":");
+    const hookSetting = { topic, event, syncType };
+    Packages.update({ name: "reaction-connectors-shopify", shopId: Reaction.getShopId() },
+      { $push: { "settings.synchooks": hookSetting }
+      });
+  },
+  "synchooks/shopify/removeHook": (hook) => {
+    check(hook, { name: String, checked: Boolean });
```
spencern: security?

zenweasel:
@spencern re: inventory Which options are you talking about?

imports/plugins/included/connectors-shopify/client/settings/shopify.html
```diff
@@ -44,7 +46,8 @@ <h4><span data-i18n="admin.shopifyConnectSettings.headingSync">Sync Settings</sp
     {{else}}
       <form data-event-action="setupShopifySync">
         <div>
-          <input type="checkbox" name="orders/create:updateInventory" {{integrationIsActive "orders/create:updateInventory"}}/>
+          <input type="checkbox" name="webhooks:orders/create:updateInventory" {{integrationIsActive
```
spencern: @zenweasel orders/create:updateInventory - option here

zenweasel: That's for incoming order from Shopify isn't is?

imports/plugins/included/connectors-shopify/server/hooks/orders.js
zenweasel: Hmmm, I disagree but okay

zenweasel:
@spencern Ready for another look

imports/plugins/included/connectors-shopify/lib/collections/schemas/shopifyConnect.js
```diff
@@ -51,6 +51,24 @@ const Webhook = new SimpleSchema({
 
 registerSchema("Webhook", Webhook);
 
+export const Synchook = new SimpleSchema({
```
spencern: jsdoc

imports/plugins/included/connectors-shopify/server/methods/sync/hooks.js
```diff
@@ -0,0 +1,32 @@
...
+    if (!Reaction.hasPermission(connectorsRoles)) {
+      throw new Meteor.Error("access-denied", "Access Denied");
+    }
+    const [ topic, event, syncType ] = hook.name.split(":");
+    const hookSetting = { topic, event, syncType };
+    Packages.update({ name: "reaction-connectors-shopify", shopId: Reaction.getShopId() },
+      { $push: { "settings.synchooks": hookSetting }
+      });
+  },
+  "synchooks/shopify/removeHook": (hook) => {
```
spencern: is there jsdoc for this method anywhere?

imports/plugins/included/connectors-shopify/server/methods/sync/hooks.js
```diff
@@ -0,0 +1,32 @@
+import { Meteor } from "meteor/meteor";
+import { check } from "meteor/check";
+import { Reaction } from "/server/api";
+import { Packages } from "/lib/collections";
+import { connectorsRoles } from "../../lib/roles";
+
+export const methods = {
+  "synchooks/shopify/addHook": (hook) => {
```
spencern: jsdoc?

imports/plugins/included/connectors-shopify/client/settings/shopify.js
```diff
@@ -91,6 +91,23 @@ Template.shopifySync.helpers({
 
     // If no webhooks are found, this integration is inactive
     return "";
+  },
+
+  hookIsActive(hook) {
```
spencern: jsdoc?

spencern:
Had some dialogue with @zenweasel offline because he'd been able to put orders through that did use the same information during his testing.

What we discovered is that when the phone number matches an existing customer but the email is different, it throws an error. If the phone and email are identical, then it seems to let the order through and tack it on to the existing customer.

spencern:
@zenweasel any progress on this PR?

zenweasel:
@spencern No. Will try to wrap up tomorrow (Wednesday/Thursday)

zenweasel:
Trying to resolve the issue with existing customers I have run into an problem with what I think is a bug in the node shopify module. Opened an issue on that library [here](https://github.com/MONEI/Shopify-api-node/issues/156). For now will be trying to fallback to a raw HTTP request to get the data. Will try to submit a PR to the Shopify lib if I can figure out what the core issue is.

zenweasel:
Found problem with Shopify lib. Was pilot error as `params` need to be passed in as string, not object. Still working on dealing with existing vs. new customers.

zenweasel:
@spencern
**Duplicate Customer** I believe I've solved the customer problem by checking for an existing customer and using that if it exists.

**Phone Number** I have fixed the formatting problem using `libphonenumber-js` but since we don't do any validation on phone numbers during checkout, if the customer inputs a completely invalid number (like 0 for example) there's nothing I can do about it. The good news is that Shopify seems just accept it

**Variant Info**
 it looks like variant info is actually being passed if you try to fulfill the order you can see it. Don't why it doesn't show up in the order summary but it's there



**Shipping Info** Fixed

Also, thinking about this I think it's probably best to move this entire process into a job so that it doesn't block checkout and so it can be retried in case of intermittent failure. It could probably be retested w/o that change however.

@prinzdezibel Michael, can you also code review this and give it some testing? @spencern is really pressed for bandwidth and I'm sure it could use your eyes on it. PM me to get details on how to login into our test shopify store.

zenweasel:
Also debating whether I want to store the order along with the job to ensure that the original order gets exported or just store the order._id so that it pulls the current version of the order. Thinking order._id is the way to go so that if the merchant "fixes" something on the order this updated version gets exported.

Also what UI do we need for managing these jobs?

PR #3291, opened 2017-11-15
Fixed browser console displaying errors when email/accounts action panel is viewed

### LINKED ISSUE #3219: Invalid prop `noDataText` browser console error
### Expected behavior
No errors in console when setting is clicked

### Actual behavior
Show `Invalid prop noDataText` in browser console

dashboard-error

- Pull master branch
- Start application and refresh page
- Click on `Accounts` setting
- Observer error in browser console

### Versions
```
Node: 8.3.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.10.0
Reaction: 1.5.1
```

aanchirinah:
Resolves #3219,

I noticed that `noDataText` prop for React-Table was passed an arrow function that returns a HTML span element as seen below,
```
noDataText={() => <span className="sortableTable-noDataText">{this.props.noDataMessage}</span>}
```
So I resolved the bug by declaring a function `displayNoResultsFound` and put the definition of the arrow function passed into it there rather and invoked it as seen below, 
```
  const displayNoResultsFound = () => {
        const displayText = <span className="sortableTable-noDataText">{this.props.noDataMessage}</span>;
        return displayText;
    };
.....
.....
.....
noDataText={displayNoResultsFound()}
```

#### HOW TO TEST

1. Sign in as admin
1. Refresh page
1. Click on Accounts setting in the side action panel
1. Observe no error in browser console`


imports/plugins/core/ui/client/components/table/sortableTable.js
```diff
@@ -272,6 +271,13 @@ class SortableTable extends Component {
   render() {
     const { ...otherProps } = this.props;
     const defaultClassName = "-striped -highlight";
+    const displayNoResultsFound = () => {
+      let displayText = "";
+      if (this.getTableData() === 0) {
+        displayText = <span className="sortableTable-noDataText">{this.props.noDataMessage}</span>;
+        return displayText;
```
zenweasel: Should this return be defined outside that `if` so that a result is return regardless? I assume that's why you defined it outside the `if`?

zenweasel: Also needs JSDocs?

aanchirinah: Oh yes the idea was to return display text even when the `if` condition isn't satisfied, I have rectified that now. Also, I have fixed the requested changes.

imports/plugins/core/ui/client/components/table/sortableTable.js
```diff
@@ -253,6 +252,21 @@ class SortableTable extends Component {
...
+   * @name displayNoResultsFound()
+   * @method
+   * @summary This function displays a 'No Results Found' when there is no data to populate the table
+   * @return {node} returns a JSX node or empty string
+   */
+  displayNoResultsFound() {
+    let displayText = "";
+    if (this.getTableData() === 0) {
+      displayText = <span className="sortableTable-noDataText">{this.props.noDataMessage}</span>;
+      return displayText;
```
zenweasel: You don't need the return inside the `if` block.

aanchirinah: Changes requested fixed @zenweasel

PR #3295, opened 2017-11-15
Stripe Connect Form Pre-Fills (one fix, a few additions)

pmn4:
Code was referencing `address.state` instead of `address.region` when passing data to Stripe Connect. Not a bug necessarily, but it did require the user to complete the state field in Stripe after they had just done so in Reaction.

To test, create a Seller Shop, give it an address, then go through the Stripe Connect process. You will find that the State you entered in Reaction does not pre-fill the Stripe form.

There is a second commit in this PR which adds some additional data to the form. It is optional, but convenient.

### DIFF
--- a/imports/plugins/included/marketplace/client/templates/stripeConnectSignupButton/stripeConnectSignupButton.js
+++ b/imports/plugins/included/marketplace/client/templates/stripeConnectSignupButton/stripeConnectSignupButton.js
@@ -1,4 +1,3 @@
-import _ from "lodash";
 import { Template } from "meteor/templating";
 import { Reaction } from "/lib/api";
 import { i18next } from "/client/api";
@@ -42,7 +41,6 @@ Template.stripeConnectSignupButton.events({
       return Alerts.toast(`${i18next.t("admin.connect.shopAddressNotConfigured")}`, "error");
     }
 
-    const email = shop.emails[0].address;
     const country = shop.addressBook[0].country;
     const phoneNumber = shop.addressBook[0].phone;
     const businessName = shop.addressBook[0].company;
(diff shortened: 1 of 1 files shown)


imports/plugins/included/marketplace/client/templates/stripeConnectSignupButton/stripeConnectSignupButton.js
```diff
@@ -47,11 +48,15 @@ Template.stripeConnectSignupButton.events({
     const businessName = shop.addressBook[0].company;
     const streetAddress = shop.addressBook[0].address1;
     const city = shop.addressBook[0].city;
-    const state = shop.addressBook[0].state;
+    const state = shop.addressBook[0].region;
     const zip = shop.addressBook[0].postal;
 
+    const user = Meteor.user() || {};
+    const { address: email } = _.first(user.emails) || {};
+    const [firstName, lastName] = _.split(user.name, " "); // split on " " isn't perfect, but we use it elsewhere, so it's consistent. to that point, consider adding a helper to User
```
spencern: a couple issues here:
1. I don't think we need to use lodash here, as [string.prototype.split](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split) should work fine here.
2. The bigger issue here is that I think there are some edge cases where this will leave off parts of someone's name. E.g. my name is `Steven Spencer Norman Jr.` and if I have all of that here, this will split first name and last name into `Steven Spencer` and not `Steven Norman` - I think the easy solution here would be to use spread into the rest of the array.. e.g.
```
const [firstName, ...last] = user.name.split(" ");
const lastName = last.join(" ");
```

imports/plugins/included/marketplace/client/templates/stripeConnectSignupButton/stripeConnectSignupButton.js
```diff
@@ -47,11 +48,15 @@ Template.stripeConnectSignupButton.events({
     const businessName = shop.addressBook[0].company;
     const streetAddress = shop.addressBook[0].address1;
     const city = shop.addressBook[0].city;
-    const state = shop.addressBook[0].state;
+    const state = shop.addressBook[0].region;
     const zip = shop.addressBook[0].postal;
 
+    const user = Meteor.user() || {};
+    const { address: email } = _.first(user.emails) || {};
```
spencern: A couple issues here:
Object destructuring doesn't work as you're intending here I don't think. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assigning_to_new_variable_names

We should be finding the default email here rather than just taking the first email available.. I'd use something like

```
const defaultEmail = user.emails.find((email) => email.provides === "default");
const defaultEmailAddress = defaultEmail.address;
```

imports/plugins/included/marketplace/client/templates/stripeConnectSignupButton/stripeConnectSignupButton.js
```diff
@@ -47,11 +48,15 @@ Template.stripeConnectSignupButton.events({
...
+    const [firstName, lastName] = _.split(user.name, " "); // split on " " isn't perfect, but we use it elsewhere, so it's consistent. to that point, consider adding a helper to User
+
     const stripeConnectAuthorizeUrl = `https://connect.stripe.com/oauth/authorize?response_type=code&state=${shopId}&client_id=${clientId}&scope=read_write`;
-    const autofillParams = `&stripe_user[email]=${email}&stripe_user[country]=${country}&stripe_user[phone_number]=${phoneNumber}&stripe_user[business_name]=${businessName}&stripe_user[street_address]=${streetAddress}&stripe_user[city]=${city}&stripe_user[state]=${state}&stripe_user[zip]=${zip}`; // eslint-disable-line max-len
+    const autofillParams = `&stripe_user[email]=${email}&stripe_user[country]=${country}&stripe_user[phone_number]=${phoneNumber}&stripe_user[business_name]=${businessName}&stripe_user[street_address]=${streetAddress}&stripe_user[city]=${city}&stripe_user[state]=${state}&stripe_user[zip]=${zip}&stripe_user[email]=${email}&stripe_user[first_name]=${firstName}&stripe_user[last_name]=${lastName}`; // eslint-disable-line max-len
```
spencern: If we change `email` to `defaultEmailAddress` we just need to update this to be ${defaultEmailAddress}

imports/plugins/included/marketplace/client/templates/stripeConnectSignupButton/stripeConnectSignupButton.js
```diff
@@ -41,17 +41,25 @@ Template.stripeConnectSignupButton.events({
       return Alerts.toast(`${i18next.t("admin.connect.shopAddressNotConfigured")}`, "error");
     }
 
-    const email = shop.emails[0].address;
```
pmn4: the business form does not have an email input, so we grab the user's default email address instead

zenweasel:
@pmn4 Hey Pat, looks like we have a lint error here. I would just fix it but I can't push to your repo. You've been made a collaborator so in the future you can just branch directly off Reaction. I think this PR is ready to go except the lint. Thanks!!

pmn4:
thanks for the review @zenweasel. updated.
(and will branch straight from here in the future!)

pmn4:
I have addressed all of @spencern's comments, and rebased from master (since I couldn't make sense of the bitHound report... 🤞 for green!)

zenweasel:
Just need @spencern to give this a thumbs up and it should be ready to go

PR #3302, opened 2017-11-16
Fix cannot cancel order processed using Authorize.net

### LINKED ISSUE #3216: Cannot cancel orders processed using Authorize.net
### Expected behavior
Should be able to cancel an order processed using any payment provider

### Actual behavior
Clicking on cancel does nothing and an error is shown in the server console
```
 Attempt for refund transaction failed WSoTNKWh4waDujT8P 60033194958 Reaction does not yet support direct refund processing from Authorize.net. Please visit their web portal to perform this action. https://account.authorize.net/
```
Basically, if a payment processor does not support refunds, it should not attempt to refund when cancelling.

1. Place an order using Authorize.NET
1. Approve and Capture the order
1. Click on "Cancel"
1. Observe that nothing happens except an error is thrown on the server

### Versions
```
Node: 7.10.0
NPM: 5.4.2
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.19.1
Reaction: 1.5.3
Reaction branch: itua-fix-issue-3075 (master)
Docker: 17.09.0-ce
```

iakhator:
Remove exception error  for Authorize.Net payment method when user click on cancel order.

#### Fix
since refund cannot happen with a payment method that does not support refund. `isRefundable` check is place to make sure a payment method  refund is enabled before performing a refund action.

#### Test
- Place an order using Authorize.NET
- Approve and Capture the order
- Click on "Cancel"
- Observe that nothing happens and no exception error thrown in the server.

### DIFF
--- a/server/methods/core/orders.js
+++ b/server/methods/core/orders.js
@@ -348,7 +348,16 @@ export const methods = {
     });
 
     // refund payment to customer
-    Meteor.call("orders/refunds/create", order._id, paymentMethod, Number(invoiceTotal));
+    const paymentMethodId = paymentMethod && paymentMethod.paymentPackageId;
+    const paymentMethodName = paymentMethod && paymentMethod.paymentSettingsKey;
+    const paymentMethods = Packages.findOne({ _id: paymentMethodId });
+    const isRefundable = paymentMethods && paymentMethods.settings && paymentMethods.settings[paymentMethodName]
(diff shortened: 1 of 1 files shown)


zenweasel:
@iakhator Remember that the PR title is what will go in the release notes, so make sure it's meaningful and doesn't start with your name.

iakhator:
ok

PR #3312, opened 2017-11-17
Fix flash 403/404 views when auth state change

### LINKED ISSUE #3241: Flash 403 / 404 views when auth state changes
Due to the way the router was implemented, there is a flash of 403 / 404 views when signing in or signing out.

This is in part due to the Trackers in the router startup that re-render the entire browser router when there is new data from the publications.

Also, the reactivity of these trackers means that, as you are transitioning from an authenticated user to an unauthenticated user, your permissions are changing and the UI is trying to keep up with it.

### Expected behavior

UI should wait for the auth state to be set before trying to update based views on permissions.

### Actual behavior

UI flashes back and forth between authentication states when signing in and out.

Sign in and sign out on various pages like the PDP page.

### Versions

Node: 7.2.0
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.20.0
Reaction: 1.5.5
Reaction branch: master
Docker: 17.09.0-ce

iakhator:
Resolve #3241 

This PR fix flashing 403/404 page  when user try to sign out or sign in.
#### Fix
This is fixed by checking if the reaction router grants permission to the user who is currently entering the page if not reveal the unauthorized. 

#### Test
- Click on product to reveal the product detail page or any other page aside the home page.
- Sign in and Sign out from the product detail page or any of the pages.
- Check to see that 403/404 page does not flash when performing sign in and out.


iakhator:
@mikemurray

spencern:
@iakhator @mikemurray is on vacation for the next week, so you may need to find someone else to review.

Issue #3314, opened 2017-11-18
After accessing a  product from the search page  , I am not able to continue scrolling down the single product page


samyilias:
When I hit the navbar search button , I select a product and  I am suppose to be able to continue scrolling down the single product page
but the page is completely stuck , I have to refresh my page to go back to normal 
**Steps to reproduce the behavior**
here is how you can reproduce the bug:
1) hit the search icon 
2) search for any product
3) select a product
4) try scrolling down the page
**Versions:**
Node: 7.7.1
NPM: 4.1.2
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.22.5
Reaction: 1.5.8

wackywombaat12:
Are you happy for me to work on this @zenweasel ?

zenweasel:
@samyilias @wackywombaat12 I am not able to replicate this error. I've attached a gif of what I am doing to test and tell if something is wrong.

![search_scroll](https://user-images.githubusercontent.com/72819/33363081-d7d347da-d519-11e7-8208-6322b244ef73.gif)

wackywombaat12:
@zenweasel maybe because you are logged in? I think this happens when you aren't signed in to an account

zenweasel:
@wackywombaat12 Yeah, that was it. Thanks!

PR #3315, opened 2017-11-18
Fix: Not being able to scroll after using search function #3314

wackywombaat12:
This is a fix for issue #3314.

**Fix**

You can now scroll as normal after searching for a product or exiting the search function. This involved removing the the 'overflow: hidden' style tag from the body HTML element after a search is complete.

**Test**

1. Click on the search icon
2. Search for a product
3. Click on the product
4. Observe that you can now scroll as intended on the page.


zenweasel:
@wackywombaat12 @samyilias I am not able to replicate the original error this is supposed to fix. Please see my notes on the ticket.

Issue #3260, opened 2017-11-09
Newly added shops don't show up under "Marketplace shops" until you refresh


zenweasel:
### Expected behavior
When you create a shop it should automatically show up under "Marketplace Shops"

### Actual behavior
Shops only show up after you refresh

### Steps to reproduce the behavior
1. Enable marketplace
1. Invite a shop owner (aka create a shop)
1. Open "Marketplace Shops"
1. Observe that the newly created shop is not there
1. Refresh
1. Observe that the shop now shows up

### Versions
```
Node: 7.10.0
NPM: 5.5.1
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.20.0
Reaction: 1.5.6
Reaction branch: master
Docker: 17.09.0-ce
```

PR #3316, opened 2017-11-18
Fix MarketplaceShops table update

efalayi:
Resolves [#3260](https://github.com/reactioncommerce/reaction/issues/3260)
Newly created shops only show up on the table after you refresh

##### Fix
```
- Subscribe to `MerchantShops` and fetch shops when subscription is ready.
```

##### Test
```
Login as admin
Enable Marketplace
Invite a shop owner (aka create a shop)
Open "Marketplace Shops"
Observe that the newly created shop is displayed
```


imports/plugins/included/marketplace/client/containers/marketplaceShops.js
```diff
@@ -9,17 +9,22 @@ const onWorkflowChange = (shopId, value) => {
...
 
 const composer = (props, onData) => {
-  // Get all shops, excluding the primary shop
-  const shops = Shops.find({
-    _id: {
-      $nin: [Reaction.getPrimaryShopId()]
-    }
-  }).fetch();
+  // Subscribe to MerchantShops
+  const merchantShopSub = Meteor.subscribe("MerchantShops");
```
impactmass: Can you use `Reaction.Subscriptions.MerchantShops` instead? I believe we get caching benefits when the subscriptions manager is used.
Example:

https://github.com/reactioncommerce/reaction/blob/efe64c6d3b292ce8cd9344bf7daa76db4f4fe1ee/imports/plugins/core/i18n/client/containers/languageDropdown.js#L18-L22

efalayi: @impactmass, I made change as requested. However, I observed that I get the cached value of `Reaction.Subscriptions.MerchantShops` unless I refresh the page. I've pushed the update so you can test it.

#### Steps
1. Do a reaction reset
2. Create a shop
3. Check Marketplace shops table. Observe that no shop is displayed
4. Create another shop and repeat step 3. Observe that no shop is still displayed
5. Refresh page and check marketplace shop table again. Observe that shops are now displayed
6. Create another shop and repeat step 3
7. Observe that table is updated with the new shop

impactmass: I just tested it too. I see the behavior you talked about. I can also see that once the subscription kicks in, then you don't need to refresh again. Any idea why this happens?

You can revert back, so that this doesn;t block your PR. We also have other places where we do Meteor.subscribe, so it's okay.

PR #3319, opened 2017-11-20
Consideration for Marketplace Shops on Different URLs

pmn4:
This PR is to address Merchant Shops which live on a different URL than the Primary Shop.
For example, my Primary Shop is hosted at https://reserve.brewline.io and a Merchant Shop at https://twolights.brewline.io. Visiting the latter shows the branding and products of the Primary Shop.
Opening developer tools on the Merchant Shop and running `ReactionCore.getPrimaryShopId() === ReactionCore.getShopId()` yields true (and the merchant shop has its `domains` attribute set properly).

Digging into the code yielded that `Meteor.absoluteUrl` relies on `$ROOT_URL` which is a problem when using the domain as an identifier.

First step was to wrap `Meteor.absoluteUrl` (in `Reaction.absoluteUrl`) passing a `rootUrl` option, and doing a find/replace throughout (there are many files changed, but few actual changes).

Next, once the active shop was correctly determined by the hostname, there were a few places which I believe were using the wrong shop.
I went through each use of `Shops.find`/`Shops.findOne` and replaced with `Reaction.getShop()` (or `Reaction.getPrimaryShop()`, where applicable).

Lastly, I added documentation to any method I created/modified.

(now that I write this, it's pretty clear that I should have broken this PR up. I will work on that tomorrow, but will leave this here if you'd like to start commenting).

### DIFF
--- a/client/modules/core/main.js
+++ b/client/modules/core/main.js
@@ -23,6 +23,7 @@ const reactionState = new ReactiveDict();
  */
 export default {
   _shopId: new ReactiveVar(null), // The active shop
+  _shopDomain: new ReactiveVar(null), // The active shop's domain (a shop may have multiple domains)
   _primaryShopId: new ReactiveVar(null), // The first shop created
   marketplace: { _ready: false }, // Marketplace Settings
 
@@ -48,9 +49,7 @@ export default {
       let shop;
       if (this.Subscriptions.PrimaryShop.ready()) {
         // There should only ever be one "primary" shop
-        shop = Shops.findOne({
-          shopType: "primary"
-        });
+        shop = this.getPrimaryShop();
 
         if (shop) {
           this.primaryShopId = shop._id;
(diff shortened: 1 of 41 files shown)


pmn4:
btw, I learned A TON about both Reaction and Meteor doing this!

client/modules/core/main.js
```diff
@@ -23,6 +23,7 @@ const reactionState = new ReactiveDict();
  */
 export default {
   _shopId: new ReactiveVar(null), // The active shop
+  _shopDomain: new ReactiveVar(null), // The active shop's domain (a shop may have multiple domains)
```
pmn4: just to confirm, a `ReactiveVar` is sort of a thread-safe variable? that is, no two connection share the same instance of this?

client/modules/core/main.js
```diff
@@ -375,39 +396,69 @@ export default {
...
+   * primaryShopId
+   * @summary setter method for the primary shop.
+   * (should probably only be used at startup)
+   */
   set primaryShopId(shopId) {
     this._primaryShopId.set(shopId);
   },
 
-  getPrimaryShopId() {
-    return this.primaryShopId;
```
pmn4: _(this is just some strange GH highlighting... the code is not removed, it is below)_

client/modules/core/main.js
```diff
@@ -431,36 +484,91 @@ export default {
     return this._shopId.get();
   },
 
-  getShopId() {
```
pmn4: _(again, this is not removed, it is below)_

client/modules/i18n/helpers.js
```diff
@@ -45,7 +45,7 @@ Template.registerHelper("currencySymbol", function () {
   });
   const profileCurrency = user.profile && user.profile.currency;
   if (profileCurrency) {
-    const shop = Shops.findOne();
+    const shop = Reaction.getPrimaryShop();
```
pmn4: not positive this should be `getPrimaryShop`, but for now the end result is the same as it was

server/api/core/core.js
```diff
@@ -230,14 +278,14 @@ export default {
...
+   */
   getPrimaryShopCurrency() {
     const primaryShop = this.getPrimaryShop();
 
-    if (primaryShop && primaryShop.currency) {
-      return primaryShop.currency;
-    }
-
-    return "USD";
+    return shop && shop.currency || "USD";
```
pmn4: _(refactored to match the client version of this method)_

server/api/core/core.js
```diff
@@ -274,12 +322,27 @@ export default {
...
+   *                  when unset
+   * @return {String} name for the primary shop
+   */
   getShopId(userId) {
     check(userId, Match.Maybe(String));
     const activeUserId = Meteor.call("reaction/getUserId");
     if (activeUserId || userId) {
       const activeShopId = this.getUserPreferences({
-        userId: activeUserId || userId,
+        userId: activeUserId || userId, // should this be reversed? if we are trying to get the shop for a given order, this would return "my" shop, not the user's shop
```
pmn4: _(comment unrelated to this PR)_

pmn4:
closing in favor of: https://github.com/reactioncommerce/reaction/pull/3332

PR #3321, opened 2017-11-20
Add email updated translation

### LINKED ISSUE #3307: Missing translation for email Updation
### Expected behavior

The translation to be present.

### Actual behavior
The translation is missing.

The i18n code used [here](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/accounts/client/containers/updateEmail.js#L24) is missing in the translation files [here](https://github.com/reactioncommerce/reaction/blob/master/private/data/i18n/en.json#L658)

iakhator:
Resolve #3307 

- Added email updated translation
- Whenever user try to update email it shows `accountUI.info.emailUpdated` instead of `emailUpdated`

#### Test
- Login as Admin or User
- Click  profile to reveal the profile page.
- Change email and click on update email address.
- Notice the toastr message 'Email Updated'.


zenweasel:
@iakhator So it's not really testing instructions to tell me to just look at the code. Surely this must manifest in the UI somewhere right?

iakhator:
yes @zenweasel , this manifest when the user is updating the email. I will re-write the test instruction.

iakhator:
@zenweasel  how to test updated

Akarshit:
@zenweasel Do you think that it makes sense that when adding a new i18n key we should translate it all the languages we support and then add it.
So basically just use google translate(or something) and add translations for all the languages.
Wouldn't this prevent a lot of `xxxx is not translated for yyyy language` issues?

zenweasel:
@Akarshit No, we use LingoHub which basically automates that process

Issue #3317, opened 2017-11-20
Tax is not shown in Completed Order page when using Avalara


zenweasel:
### Expected behavior
Taxes should display as part of the order summary

### Actual behavior
They do not

### Steps to reproduce the behavior
1. Enable Avalara as Tax provider
1. Checkout as normal with an address that would be charged tax
1. Observe tax charged during checkout
1. Observe tax does not display on completed order page

### Versions
```
Node: 8.9.1
NPM: 5.5.1
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.22.5
Reaction: 1.5.8
Reaction branch: brent-fix-issue-2718 (master)
Docker: 17.09.0-ce
```

Akarshit:
Did a quick and dirty fix for this in #3326. Have some doubts that I asked there.

PR #3326, opened 2017-11-20
Using the default tax rate

Akarshit:
## Fix for #3317 
Well this kind of feels dirty to me.
I figured the cause of the issue was missing `taxRatesByShop` field [here](https://github.com/reactioncommerce/reaction/blob/master/lib/collections/transform/cartOrder.js#L137). This was because when the tax is being calculated by Alavara API, we don't set the`taxRatesByShop` instead we just set the `taxRate` in `tax` field of cart [here](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/included/taxes-avalara/server/hooks/hooks.js#L36) which eventually does a call [here](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/taxes/server/methods/methods.js#L68).
Instead of this I think we should calculate the `taxRatesByShop` and set it up in the cart itself like it is being done [here](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/taxes/server/methods/methods.js#L157-L187). Though I am not 100% sure what would be the best way to do this. May be the people here can help.

### To check
1. Enable Avalara as Tax provider
1. Checkout as normal with an address that would be charged tax
1. Observe tax charged during checkout
1. Observe tax does display on completed order page


zenweasel:
Yeah, this is probably not the right approach. More feedback to come.

zenweasel:
We don't store taxRatesByShop because Avalara is more complex than that. The tax is stored `per item` and that needs to override any other method of storing tax.

`taxRatesByShop` is also a bit of a misnomer because as far as I know we don't have any way of configuring tax rates by shop.

PR #3337, opened 2017-11-21
(feat): Update PublishControls React component to use Components API

impactmass:
The PublishControls component in `core/ui/components` was not "registered" using the components API, and that means it cannot be replaced in a custom plugin.

This PR changes it so that it is registered with the components API.

All functionality by this component should be untouched.

### How To Test:
- On this branch, login as an admin (only admins see publish controls)
- Confirm that you can see the PublishControls (toolbar) containing all the icons should be there
e.g 

screenshot 2017-11-21 19 33 44

### Some PR notes:
- I noticed that the PublishControls component was created as a class with the named `PublishControls`, but in a file named `toolbar.js`. 
- At first, I registered it as toolbar component i.e `registerComponent("Toolbar")` but that failed to render.  I later saw that there is another component called toolbar, so I renamed the file for PublishControls to be `publishControls.js` (so you'll noticed file name changes)


imports/plugins/core/dashboard/client/components/index.js
```diff
@@ -1,4 +1,4 @@
 export { default as ActionView } from "./actionView";
 export { default as PackageList } from "./packageList";
-export { default as Toolbar } from "./toolbar";
+export { default as PublishControls } from "./publishControls";
```
impactmass: Relevant comment about this file name change:

- I noticed that the PublishControls component was created as a class with the named `PublishControls`, but in a file named `toolbar.js`. 
- At first, I registered it as toolbar component i.e `registerComponent("Toolbar")` but that failed to render when called as `Components.Toolbar`.  I later saw that there is another component called toolbar, so I renamed the file for PublishControls to be `publishControls.js`

impactmass:
closing in favor of #3339

PR #3358, opened 2017-11-29
#3354 Fix fulfilling part of a multi-merchant order removes other parts of order

### LINKED ISSUE #3354: Fulfilling part of a multi-merchant order removes other parts
### Expected behavior
Fulfilling a single merchant's part of a multi-merchant order should only mark that part of the order as completed and other merchant's parts of an order should never be modified.

### Actual behavior
When a part of an order is `packed`, the other parts of that order are deleted from the database. My guess is that this is happening because somewhere there's a $set going on with the items in the order from a client that is unaware of the other parts of the order. I've not verified that hunch, and work needs to be done to assess the exact cause of this bug.
See video
http://recordit.co/xw7xI8dgdP

1. Setup a multi-merchant marketplace (primary shop + at least one merchant shop)
2. Enable the merchant shop(s)
3. Setup shipping and Stripe payments in the primary shop
4. Setup stripe payments in the merchant shop
5. Create and publish products from each shop
6. As a guest, add products from each merchant to your cart and checkout
7. Fulfill a single merchant's part of the order, from either the merchant or the primary shop.
8. Observe that the other parts of the order are deleted when items are marked as packed

### Versions
```
Node: 8.9.0
NPM: 5.5.1
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.23.2
Reaction: 1.5.11
Reaction branch: master
Docker: 17.06.2-ce
```

codejockie:
Resolves #3354 

Fixes order been deleted when a part of a multi-merchant order is `packed` in the Order dashboard.

## Steps to Test

1. Setup a multi-merchant marketplace (primary shop + at least one merchant shop)
2. Enable the merchant shop(s)
3. Setup shipping and Stripe payments in the primary shop
4. Setup stripe payments in the merchant shop
5. Create and publish products from each shop
6. As a guest, add products from each merchant to your cart and checkout
7. Fulfill a single merchant's part of the order, from either the merchant or the primary shop.
8. Observe that the other parts of the order are not deleted when items are marked as packed


zenweasel:
I mentioned this on the original ticket but I can't reproduce this error on `master`. Since you said that you were able to, maybe you can help me understand what I am missing

codejockie:
@zenweasel the issue was that given two or more shop owners, when a customer/user makes an order with items from different shop owners; the other parts of the order are deleted when one of the shop owners approves the order and packs it. This is more like a first come first serve.

Say, shop owner A goes first to pack an order, then the other parts of the order that were bought from say shop owner B are deleted from the items list.

zenweasel:
@codejockie Yes, I get that. I am seeking some specificity around what "other parts of the order" are as I do not see anything missing from the order. I want to know exactly what part of the order object are missing. Preferably some shots from robomongo or something. Because I am not seeing it.

codejockie:
@zenweasel  I'd send you a screenshot now.

codejockie:
screen shot 2017-11-30 at 08 16 51

The first Order `items` prop has just a single item in it as it was what I used to reproduce the bug.
The other orders were used to test out this PR.

imports/plugins/core/checkout/server/methods/workflow.js
```diff
@@ -335,6 +335,39 @@ Meteor.methods({
     check(order, Object);
     check(itemIds, Array);
 
+    if (status === "coreOrderItemWorkflow/packed") {
+      const shopId = Reaction.getShopId();
+
+      const existingOrder = Orders.find({ _id: order._id }).fetch();
```
zenweasel: You want to just do `findOne` there.

imports/plugins/core/checkout/server/methods/workflow.js
```diff
@@ -335,6 +335,39 @@ Meteor.methods({
     check(order, Object);
     check(itemIds, Array);
```
zenweasel: This doesn't seem like the right approach to this problem. Firstly, we should really try to avoid hardcoding in individual statuses like this as it's really brittle. Also, this method is already getting an array of item Ids, it seems we should be getting the correct values there and be able to update the workflow based on those values. Basically here we are just ignoring those and rewriting the entire order items. 

Basically this is a hack around the problem rather than a real solution.

zenweasel: This is a problem with the existing code as well. It's not using `itemIds` at all.

codejockie: Yes, I noticed that `itemsIds` wasn't used in the method at all also, was surprised.

imports/plugins/core/checkout/server/methods/workflow.js
codejockie: Yes, I needed just a specific order.

zenweasel: Right so just `const existingOrder = Orders.findOne({ _id: order._id })` which eliminates the need for the `existingOrder[0]`

codejockie: Wow, cool.

zenweasel:
@codejockie What's the status on this? It's been two days since an update

codejockie:
@zenweasel I will push today

codejockie:
@zenweasel please take a look at the recent update.

imports/plugins/core/checkout/server/methods/workflow.js
zenweasel: So this seems like just a slightly fancier way of hard-coding in the process. This method should not need to be aware of any specific status such as `isPack` and I don't understand why we are doing that.

zenweasel:
@codejockie What's the status on this?

codejockie:
It's fixed.

zenweasel:
Tested. Verified that after fulfilling part of the order from a merchant shop the other items remained present and the workflow was properly updated for just the appropriate items

spencern:
Failed tests, rerunning.

codejockie:
@spencern tests passed locally.

spencern:
I've rerun tests several times on this PR and they are consistently failing. It seems to be timeout related issues. My guess is that because you are now calling a db update inside of a `for` loop you've introduced a significant performance penalty. I think the reviewer (@zenweasel) and you should discuss whether this performance hit is worth it or if there is potentially another solution to this problem that doesn't involve a db update call within a `for` loop.

If it's decided that the loop is worth the performance penalty, we should probably document that code a little better and explain why it's the right way to resolve this issue. Then the tests should be updated to account for the new, slower tests and the timeout should be updated.

But that should only be done if there isn't a satisfactory solution that doesn't involve a nested db update call.

zenweasel:
FWIW: These tests are failing for me locally as well.

The test is failing due to a timeout. I believe if the test was actually failing because of the extra time taken by the database call (which would mean the db call is taking more than 5 seconds) then it would be an assertion error. Regardless we need to look more closely at this test and understand why it's failing.

@codejockie and I did discuss different approaches. The problem is that this method does not have the entirety of the order so it needs to just modify the order items it gets info about. We discussed pulling in the order, modifying the individual items and then rewriting all the items. This approach has the problem of not being atomic (the order could be modified between the time it's read and rewritten and then this would cause data corruption, avoiding which I was willing to pay the performance penalty). It's not an edge case that another shop could be modifying this order at the exact same time.

If there's another approach that I am not aware of I would be very interested to know about it because as far as I know this has been an issue since Mongo 2.6.

spencern:
As this call is happening on the server, I do wonder if it could be made more nuanced. AFAIK the original issue here is that we have a merchant shop (A) that doesn't necessarily have visibility into the entire order that needs to update the items associated with it but _not_ the items associated with other shops (B, C, D, ...)

It seems like perhaps this update could be done using the shopId of the items in question, but I haven't looked into exactly how that mongo query would need to be formed.

If that's not possible in an array within mongo, then perhaps pulling the items and updating them as was happening before, but only updating the items from the particular shop would be the way to go. As this is on the server, I can't think of any reasons that this method shouldn't have access to the full order as long as the only items that are mutated are from the shopId that the user has `"order"` or whatever the relevant permissions are for that shop.

I understand you may have discussed some of these approaches offline, but it'd be good to have some of these approaches discussed on the ticket.

zenweasel:
> If that's not possible in an array within mongo, then perhaps pulling the items and updating them as was happening before, but only updating the items from the particular shop would be the way to go. 

Because we are just given itemId's I don't know how it would be possible.

zenweasel:
@codejockie Since they are all getting the same value could we use an `$in` for an array of ID's and make all the updates in one call?

zenweasel:
Really we should be looking at:

`read a record`
`modify the record in memory`
`write the record out again`

as an anti-pattern to be eliminated as it's not safe for highly concurrent systems.

codejockie:
@zenweasel I don't quite understand what you meant by "they are all getting the same value".

zenweasel:
They are get set to the same value so something like:

```
      result = Orders.update(
        { "_id": order._id, "items._id":  {$in: [id1, id2, id3] }},
        {
          $set: {
            "items.$.workflow.status": status,
            "items.$.workflow.workflow": workflow
          }
        });
```

and eliminating the `for` loop. 

Not sure if Mongo allows for using `$in` in an array update query but it's worth a try.

codejockie:
Actually Mongo does allow `$in` in update query.

zenweasel:
@codejockie Yes of course in an `update`, I said an `array update` query.

codejockie:
@zenweasel I would revisit the code and give it a try.

zenweasel:
Yes, that is what I am asking you to do

codejockie:
I have tried doing this without the loop, using the suggested code above but it turned out it wouldn't work since we are dealing with an array of `items`. This `items` array also contains the `workflow` that will be used in the update.

I could try using the `shopId` as suggested above.

codejockie:
I'm open to suggestions on using `shopId` option as I tried with that and it didn't work. `items` is an array containing the different orders placed.

spencern:
The `$in` may not work to update multiple items in an array within a single document. I know it works to update multiple documents, but that's a different use case. 

@zenweasel I'll see if I can put some code together to demonstrate what I'm thinking.

spencern:
This ticket needs to be merged asap, I think I'm the blocker right now and will work on this after the community call tomorrow am.

spencern:
Update, I'm working on this now.

spencern:
@zenweasel can you re-review this? I've optimized the code that was looping the Mongo updates and that has also resolved the test timeout issue.

Issue #3355, opened 2017-11-28
"event is not defined" error in browser console in Firefox


zenweasel:
### Expected behavior
No errors

### Actual behavior
Receive error in console
```
ReferenceError: event is not defined
[Learn More]
productGridItemsContainer.js:52:8
ProductGridItemsContainer/_this.onPageClick
productGridItemsContainer.js:52:8
<anonymous>
```

### Steps to reproduce the behavior

1. Launch Reaction with FF

### Versions
```
Node: 8.9.1
NPM: 5.5.1
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.23.1
Reaction: 1.5.11
Reaction branch: master
Docker: 17.09.0-ce
```

codejockie:
Fixed.

spencern:
Resolved in release [v1.6.4](https://github.com/reactioncommerce/reaction/releases/tag/v1.6.4)

PR #3359, opened 2017-11-30
[issue #3355]

AshtonHarding:
This PR is intended to fix the issue mentioned in #3355 for firefox
Here is an image of my results: https://i.imgur.com/b6hCPGT.png

The reason the error was occurring is because Firefox doesn't have the global `window.event` available. I looked into alternatives for `window.event` and this deals with the problem.

I know it's minor, but I hope this helps. :)


AshtonHarding:
I'll figure out why it's not passing the checks real quick.

AshtonHarding:
Ok, it appears bitHound isn't a requirement to pass. However, over the next few days, when I have the time, I will be working on getting my fork to completely pass these checks.

zenweasel:
The linting error looks to be related to the solution. Not sure if we maybe need to disable this check?

```
/Users/brenthoover/Projects/js/reaction/imports/plugins/included/product-variant/containers/productGridItemsContainer.js
  52:9  error  Read-only global 'event' should not be modified  no-native-reassign
```

AshtonHarding:
The latest push has made the error stop occurring. (This is the first time I've been using ESLint, pretty neat that they use comments!)

spencern:
@zenweasel @AshtonHarding I'm wary of reassigning the global event here. Should the event be coming in the function args here? Are we sure that the global event is what we're needing to reassign?

@mikemurray any thoughts on this?

zenweasel:
Taking the solution in #3429 over this since we don't need to overwrite the global event.

PR #3361, opened 2017-11-30
Fix flash empty cart flashing when registering a user

iakhator:
Resolve issue #3036

This PR fix empty cart showing up while registering new user, The reason for the empty cart showing up is that the cart.items is empty at the that window it tries to register the user. What I did was to create a helper method in checkout.js called cartCount where it checks if the cart.item exists and also check if the length is greater than 0.

#### Test
- Place an order.
- Checkout order.
- Next step will show telling you to register.
- Input details and register.
- Notice that it doesn't show the `empty cart page`
- It takes you to the next step.


imports/plugins/core/checkout/client/templates/checkout/checkout.js
```diff
@@ -15,6 +15,13 @@ Template.cartCheckout.helpers({
       return Cart.findOne();
     }
     return {};
+  },
+  cartCount() {
+    const cart = Cart.findOne();
+    if (cart.items && cart.items.length > 0) {
```
impactmass: There is a brief time when cart doesn't exist (this happens when a cart is transitioned to an order and another cart is being created for the user). In that brief time, an error is thrown in the browser console by this line.

This line will be better as:
`if (cart && cart.items && cart.items.length > 0) {`

PR #3365, opened 2017-12-04
[WIP] Introduce workflow for paymentmethods

prinzdezibel:
Warning: This involves significant changes to processing of
    paymentmethods. It's not completed, but  it should work for these
    two paymentmethods:  payments-cod & example-payment

    For testing, payments-cod must be installed (same branch name)

    Changes:
      - Generic mechanism for defining workflows through
      javascript-state-machine
      - Allows to customize transisitions between states via Strategy API
      - Connect workflow states to workflow-steps in registry.
      - Helper to render visual representation of current state
      - Stateflows can be used client and server side (haven't tested server
      side yet)
      - Generic save method for all Stateflows (workflow/setState)
      - probably a couple of other things..


prinzdezibel:
@zenweasel To get a high-level overview, how this is supposed to work, please have a look at the changes in invoiceContainer.js. and invoiceActions.js

prinzdezibel:
I think in general one can say: A stateflow  is a blueprint of a workflow in a specific context. E.g. payments-cod stateflow defines how workflow "PaymentMethod" should look like in payments-cod.

Also I think that the property `collection` should go away from Packages.layout and instead be moved to Stateflows. Stateflows can optionally know on which data they operate (not mandatory though), layouts should not know at all.

zenweasel:
@prinzdezibel With the upcoming architecture changes I wonder if this PR is no longer relevant?

prinzdezibel:
@zenweasel  I think the PR itself is still relevant, because it's not dealing with Meteor specific things. It's rather about providing more flexible workflows and decouple the UI from the business logic that drives/trigger workflow states. 

@aaronjudd, @zenweasel: But I'm aware that the planned architectural changes may shift the priorities. So even if I'd love to see this landing in core, I'm not sure it's worth to invest time into it. Also, @spencer was thinking about more flexible workflows, if I remember correctly. I do believe this PR can weigh in on it.  Maybe we put this PR to the icebox for later?

spencern:
For big architectural changes like this, I think we'll want to start by expressing what the goals are rather than jumping straight into the code.

Ultimately, this type of architectural decision should be made considering _all_ of our workflows. I don't think it helps to start out with a specific workflow such as payments that don't involve a credit card. 

Any solution we architect for solving this problem should also be able to solve similar issues with any other type of workflow, such as, but not limited to:
1. Onboarding workflows, including having different workflows for different user types
2. Merchant shop activation workflows (new -> pending -> active -> suspended -> closed, etc)
3. Order fulfilment workflows

While I think you've got some interesting thoughts regarding payment providers, most of our current payment providers are credit card processors, so I don't think we can consider this without understanding how those payment providers will work.

I think that the @reactioncommerce/architecture may want to define more specifically how these types of workflows should work.

I agree that being able to customize workflows in the app is an important issue to resolve, and I don't want to discourage ideation. 

That being said, I think we may have the cart before the horse a bit here by writing so much code without having a clear direction or understanding of what the specs will be or what all of the use-cases are.

prinzdezibel:
I certainly don't want to die on that hill, but I think this PR has deserved more substantial feedback.

It not only solves a very specific problem I had with the COD plugin, but also solves (or at least tries to solve) a much more general problem in our code base. Namely, that it's very hard and sometimes impossible to extend/customize existing workflows, for instance the `cartWorkflow`.

>  I don't think it helps to start out with a specific workflow such as payments that don't involve a credit card.

This was my specific problem I was facing and it's solved through a concrete manifestation of this PR's essence, if you will. By no means the concept of this PR is constrained to non-credit-card workflows or even payment workflows in general.

The core idea is to have workflows definitions in a general and consistent manner throughout our system, save their state consistently in storage and being able to hook into workflow transitions without being forced to reuse existing UI components.

I certainly don't have a problem with a PR that gets rejected because it doesn't align with architectural objections or because the priorities have shifted. I just ask for good explanations.

PR #3366, opened 2017-12-04
(fix): Determine denormalized product inventory status during Shopify import

spencern:
Fixes an unreported issue where denormalized product inventory status was not set during import of products from Shopify.

This PR resolves this by setting `isSoldOut`, `isLowQuantity`, and `isBackorder` flags correctly on the top level product during import.

### DIFF
--- a/imports/plugins/included/connectors-shopify/server/methods/import/products.js
+++ b/imports/plugins/included/connectors-shopify/server/methods/import/products.js
@@ -412,14 +412,14 @@ export const methods = {
 
                         // Save all relevant variant images to our option
                         const optionImages = findVariantImages(shopifyOption.id, shopifyProduct.images);
-                        for (const optionImage of optionImages) {
+                        for (const [index, optionImage] of optionImages.entries()) {
                           saveImage(optionImage.src, {
                             ownerId: Meteor.userId(),
                             productId: reactionProductId,
                             variantId: reactionOptionId,
                             shopId: shopId,
                             priority: 1,
-                            toGrid: 0
+                            toGrid: index === 0 ? 1 : 0 // We save the first of each variant image to the grid
                           });
                         }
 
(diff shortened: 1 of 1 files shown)


impactmass:
Looks good generally, but I have some questions:
- Looks like there's no equivalent `isLowQuantity` value from the Shopify data?
- How does this change affect anyone that already used the current version to import products?

impactmass:
@spencern looks like you missed the comment above

spencern:
arg. I thought I responded to this already, but probably have a draft mode somewhere.

`isLowQuantity` does not have an equivalent on Shopify, so we just set it to false.

Re: this change affecting anyone that already used the current version to import products, that's a great question. I'm pretty sure that top-level field gets updated every time inventory gets changed, but it might be worth including a migration here that loops through all products and updates them anyway?

impactmass:
I think if we can confirm that the top-level field gets updated every time inventory gets changed, this should be good to approve then.

spencern:
@impactmass Those do get updated as part of this `denormalize` function that gets called when variants are updated. See `catalog.js` https://github.com/reactioncommerce/reaction/blob/master/server/methods/catalog.js#L188

It's then called on L564 by updateVariant, and on L965 by updateProductField.

spencern:
I've added one more update in here to fix the way variant images are imported. To this point only the first image imported was added to the grid via `toGrid: 1` but after experimenting with how Reaction assigns product media to the grid when creating a product and variants, I've updated the Shopify importer to now add the first image for each variant to the grid as well. This should resolve the situation we've seen in #3368 where grid item layouts do not show additional product media that has been imported from Shopify.

It will not affect existing images from Shopify that have been imported.

I'm not certain if we want to add a migration here, as it's possible that people have built their apps to account for the way that the media was imported originally, but a migration could be added to set the first image of each variant to have `toGrid: 1`

I lean towards this being a fix and not adding a migration, but could understand an argument either way. @impactmass what do you think?

spencern:
@impactmass ready for another review here.

impactmass:
Yes, doing a migration in this case can cause a break in the way someone else already built on earlier imports. So, I tend towards the way it's setup here.

PR #3367, opened 2017-12-04
(fix): Use denormalized properties on product for gridItemNotices

spencern:
Resolves unreported issue where gridItemNotices were calculating product inventory status on the fly by (essentially) reducing the status of that product's variants.

This PR achieves this by doing the following
- Updates the GridItemNotice component to use boolean propTypes for `isBackorder`, `isLowQuantity`, and `isSoldOut`.
- Eliminates the need for GridItemNoticeContainer as we can pass the boolean propTypes directly into the GridItemNotice component
- Registers the GridItemNotice component at the bottom of gridItemNotice.js


spencern:
This PR was reverted out of the upcoming release when updating from release-1.7.0 to release-1.6.2.

There were significant obstacles encountered when attempting to resolve denormalized product inventory statuses - `isSoldOut`, `isLowQuantity`, and `isBackorder` - which will require changes or removal of the Product Revisions system.

PR #3377, opened 2017-12-07
[WIP] Feat: Add dashboard panel to manage marketplace shop packages

impactmass:
WIP on 3372

Todo;

- [ ] Fix selected table row (issue with setActionView detail causing un-needed re-render)

### DIFF
--- a/imports/plugins/included/marketplace/server/publications.js
+++ b/imports/plugins/included/marketplace/server/publications.js
@@ -2,7 +2,7 @@ import _ from "lodash";
 import { Meteor } from "meteor/meteor";
 import { check, Match } from "meteor/check";
 import { Reaction } from "/lib/api";
-import { Shops } from "/lib/collections";
+import { Shops, Packages } from "/lib/collections";
 
 
 Meteor.publish("SellerShops", function (shopIds) {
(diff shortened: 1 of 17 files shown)


imports/plugins/included/marketplace/server/publications.js
```diff
@@ -73,3 +73,46 @@ Meteor.publish("SellerShops", function (shopIds) {
...
+/**
+ * MarketplaceShopPackages
+ * @summary function to publish packages available for a marketplace shop
+ * @param {string} shopID
+ * @return {object | boolean }
+ */
+Meteor.publish("MarketplaceShopPackages", function (shopId) {
+  check(shopId, String);
+
+  if (!Reaction.hasPermission("admin", this.userId, Reaction.getPrimaryShopId())) {
```
pmn4: @impactmass should this be `shopId` instead of `Reaction.getPrimaryShopId()`, and if not, could you explain why? (I'm not super clear on all of the roles... i.e., does a Marketplace Shop Owner have an admin role in the Primary Shop?)
thanks!

impactmass:
@pmn4, yes, it should be getPrimaryShopId() here because the action in this
method is to be performed on the main marketplace shop.

`getPrimaryShopId()` - returns the id of the main marketplace shop i.e
primary shop (which has controls/settings for the marketplace - itself and
the other associated merchant or affiliate shops).
`getShopId()` - returns the id of the current shop based on the app
context. So this can return a different value at different times.

A marketplace shop owner (i.e the owner of merchant or affiliate shop) does
not get admin permissions in the primary shop. Except if explicitly granted
by the primary shop admin.

spencern:
@impactmass I'd like to get this issue resolved if it's close or closed if we're postponing it for a while. Will you update this ticket with any progress that has been made and/or additional work that would be needed to resolve this issue?

impactmass:
It's not close to completing because there are some specs to clear up on the ticket. I'll write a status on the ticket and close this for now.

PR #3380, opened 2017-12-08
(fix): 3379 - sweetAlerts2 ^7.0.0 update compatibility

### LINKED ISSUE #3379: sweetAlert2 ^7.0.0 breaks existing pop up alerts
### Expected behavior
`Alerts.alert` should do something when an alert window is confirmed, or cancelled

### Actual behavior
`Alerts.alert` silently fails when either button is clicked

- Find any pop-up alert (for instance, deleting a picture from the PDP), and click either the cancel or confirm buttons
- See the pop-up closes, but nothing happens

### Versions
sweetAlert2 `7.0.0`+

kieckhafer:
Fixes #3379 

When we updated package.json, it bumped sweetAlerts2 to a `^7.0.0` version, which had breaking changes.

This PR follows [their directions](https://github.com/limonte/sweetalert2/releases/tag/v7.0.0) to add backwards compatibly to our `Alerts.alert` so the pop-up alerts continue to work as they did.

To Test:
- Find any pop-up alert (for instance, deleting a picture from the PDP), and click either the cancel or confirm buttons
- See that the action you select actually happens, instead of just closing the pop-up silently

### DIFF
--- a/imports/plugins/core/layout/client/templates/layout/alerts/reactionAlerts.js
+++ b/imports/plugins/core/layout/client/templates/layout/alerts/reactionAlerts.js
@@ -72,6 +72,7 @@ Object.assign(Alerts, {
   alert(titleOrOptions, messageOrCallback, options, callback) {
     if (_.isObject(titleOrOptions)) {
       return swal({
+        useRejections: true,
         type: "info",
         ...titleOrOptions
       }).then((isConfirm) => {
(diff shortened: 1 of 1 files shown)


spencern:
@kieckhafer how can I test this, what should I be looking for?

kieckhafer:
@spencern updated description with testing instructions. Basically all pop ups Alerts were just closing silently without actually doing what they were supposed to do. So just find a pop-up and make sure clicking the button does what it should instead of just closing the pop-up.

duanhong169:
Same issue and @kieckhafer 's fix works.

Issue #3373, opened 2017-12-06
Refactor paymentMethod.supports into an array of booleans rather than string


zenweasel:
For some reason this was implemented as an array of strings (and I approved it) and this is not what we want and doesn't represent the data well and makes auto-generating forms from the schema impossible.

This shouldn't be complicated but means touching each payment method and finding all places where `supports` is used and testing each method thoroughly.

This change is currently blocking the rewriting of the payment method settings panels into React using the `<Form>` component.

zenweasel:
Closing as this was to resolve an issue with the `Form` component but that would not appear to be relevant. Can be handled when we encounter an actual problem to solve.

PR #3381, opened 2017-12-08
[WIP] Refactor paymentMethod.supports into an array of booleans rather than string

iakhator:
Working on #3373

This is a work in progress  PR that does:
- Refactor payment schema with an array in `support` property
- Change the `array of String` to `object of Boolean`
- Make sure payment work properly.

#### Test 1
- Log in
- Open the payment settings panel
- Enable either `Braintree or Authnet` payment method.
- Go to `Payment provider supported methods` check  or uncheck any of the supported methods the click on `save changes`. Notice the toast with a message of updated.
- Go to packages open `settings-reaction-braintree.support` or `settings-reaction-auth-net.support` depending on the current payment method  notice the type of `Boolean` and the changed from either true to false or false to true.

#### Test 2
- Log in
- Enable either `Braintree or Authnet` payment method.
- Place an order
- Checkout using any of the payment method stated above
- Notice the payment worked as normal.

### DIFF
--- a/imports/plugins/core/orders/client/containers/invoiceContainer.js
+++ b/imports/plugins/core/orders/client/containers/invoiceContainer.js
@@ -224,8 +224,7 @@ class InvoiceContainer extends Component {
     const paymentMethodName = orderBillingInfo.paymentMethod && orderBillingInfo.paymentMethod.paymentSettingsKey;
     const paymentMethod = Packages.findOne({ _id: paymentMethodId });
     const isRefundable = paymentMethod && paymentMethod.settings && paymentMethod.settings[paymentMethodName]
-      && paymentMethod.settings[paymentMethodName].support.includes("Refund");
-
+      && paymentMethod.settings[paymentMethodName].support.refund;
     return isRefundable;
   }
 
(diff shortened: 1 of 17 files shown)


imports/plugins/core/orders/client/containers/invoiceContainer.js
```diff
@@ -224,8 +224,7 @@ class InvoiceContainer extends Component {
     const paymentMethodName = orderBillingInfo.paymentMethod && orderBillingInfo.paymentMethod.paymentSettingsKey;
     const paymentMethod = Packages.findOne({ _id: paymentMethodId });
     const isRefundable = paymentMethod && paymentMethod.settings && paymentMethod.settings[paymentMethodName]
-      && paymentMethod.settings[paymentMethodName].support.includes("Refund");
-
+      && paymentMethod.settings[paymentMethodName].support.refund;
```
zenweasel: I think we need to move this check to a method or something so we can give a safer check. This makes a lot of assumptions about the entire object being there. Or maybe just use `_.get` or equivalent

zenweasel:
This seems like the right idea. Let's just go ahead and get it finished up

Issue #3382, opened 2017-12-08
Replacing ProductGridContainer via Components.API throws error.


prinzdezibel:
Trying to replace ProductGridContainer will result in Javascript errors when rendering ProductGrid (LandingPage), because `this.state.productIds` of ProductGridContainer is not defined. The page will not load, but render the spinner indefinitely.

### Expected behavior
Overriding of ProductGridContainer via `replaceComponent` should work without errors.

### Actual behavior
Error is thrown. Spinner is shown  forever.

### Steps to reproduce the behavior

Somewhere in code: 
```
import { replaceComponent, getRawComponent, getHOCs } from "@reactioncommerce/reaction-components";

replaceComponent("ProductGrid", getRawComponent("ProductGrid"), getHOCs("ProductGrid"));
```

Navigate to `http://localhost:3000`

See errors in browser console.

### Versions
```
Node: 8.2.1
NPM: 5.3.0
Meteor Node: 8.8.1
Meteor NPM: 5.4.2
Reaction CLI: 0.23.1
Reaction: 1.6.0
Reaction branch: master
Docker: 17.03.1-ce
```

jshimko:
Is there a real world use case for your above example?  It's replacing the `ProductGrid` component with itself and then re-wrapping it with the same HOC's.

prinzdezibel:
thats only a show case. i m doing it with another component thats really overridden. the result is the same, though

prinzdezibel:
@jshimko You were right. PEBCAB. The "re-wrapping part" of your question was the culprit.

Stupid me. I've read the Components.API docs, but obviously not close enough, which is why I did:
I was not aware that the original HOCs are already mixed into when replacing a component. I've read about the Component.API in our docs, but obviously not close enough, which is why I did:

```
replaceComponent("ProductGrid", MyProductGrid, getHOCs("ProductGrid"));
```
instead of
```
replaceComponent("ProductGrid", MyProductGrid);
```

I was not aware that the original HOCs gets merged in automatically when calling `replaceComponent`.

Sorry for wasting your time :(

Going to close this issue.

PR #3383, opened 2017-12-08
Fix for error when replacing ProductGridContainer via Components.API 

prinzdezibel:
This change fixes issue 3382.

The reason for the error is that a re-instantiation through the
Meteor.Tracker reruns the constructor of ProductGridContainer without
passing the properties `props.productsByKey` and `props.productIds`
which results in the error:

>   TypeError: Cannot read property 'map' of undefined

This is because Tracker instantiation does not occur via code in
products.js, line 69 :
```
<Components.ProductGrid
        productsByKey={productsByKey || {}}
        productIds={getIds({ tags: products })}
        canEdit={Reaction.hasPermission("createProduct")}
        products={products}
      />
```
which is used the first time the component is created.

The solution is to use a reactive composer that injects the properties
during each tracker run.


prinzdezibel:
@spencern  Thanks for the reactivity hint! Haven't been thinking about reactivity at all, but you're right. As there's no reactive variable/func, I don't need it. Replaced it with plain `compose`.

I'm still using the composer function itself, because otherwise I think I would have to pass the properties down the hierarchy, i.e. from containers/productContainer ->  components/products -> containers/productGridContainer very similar to what has been done currently with `products`.

@mikemurray The motivation for this change is really the fact, that passing productIds  and productsByKey through JSX 
```
<Components.ProductGrid
        productsByKey={productsByKey || {}}
        productIds={getIds({ tags: products })}
```
is not sufficient, because the ProductGrids constructor may be called through another code path:
 Namely as a consequence of the [calling `onData`](https://github.com/reactioncommerce/reaction/blob/50035306461edb84bd8fa5d3f935b6638c03913d/imports/plugins/included/product-variant/containers/productsContainer.js#L198) in productContainer's composer function (which I admittedly don't understand entirely, as it happens within React library due to state changes).

But this behaviour only is visible , when overriding the component like explained in issue #3382.
(Note, that the `replaceComponent` code is just there for demonstration purposes. I'm facing a real use case when implementing the swag shop, though.)

prinzdezibel:
@spencern @mikemurray  Any news on this? Would be great if you could check this, because another PR in reaction-swag-shop depends on it.

prinzdezibel:
@spencern, @mikemurray   Digging into it again, I need to revise the following statement from me:

> I'm still using the composer function itself, because otherwise I think I would have to pass the properties down the hierarchy, i.e. from containers/productContainer -> components/products -> containers/productGridContainer very similar to what has been done currently with products.

Actually it turns out hierarchy is rather flat, it just consists of one higher order component. So I was wrong with my assumption, I would have to pass the properties down deeply.

Therefore the whole PR gets considerably sleeker. There's no need for a separate composer function as it is completely superfluous. Going to push an improved version.

prinzdezibel:
Going to close, because this is a PEBCAK issue :(

PR #3384, opened 2017-12-08
Fix: Undefined property: Reaction.Router.current().queryParam

### LINKED ISSUE #3496: Product filtering via URL query parameter does not work any longer
### Expected behavior
In the past it was possible to filter products according to predefined filter criteria, e.g.
`http://localhost:3000/?price.min=10&price.max=20`
would find only products between 10 $ and 20 $.

### Actual behavior
This doesn't work any longer, because the URL params are not evaluated at all, because they're accessed in a wrong way (`Reaction.Router.current().queryParams` instead of ` Reaction.Router.current().query`)

- Create one product between 10 and 20 $
- Create a second product over 20 $
- paste this URL into address bar of browser: http://localhost:3000/?price.min=10&price.max=20
- see both products being displayed instead of only the first

prinzdezibel:
Filtering via URL query param did not work any more. This is because the Router's current route has no property called `queryParams`. Correct would be `query`.

Closes #3496 

## To test
- Create one product between 10 and 20 $
- Create a second product over 20 $
- paste this URL into address bar of browser: http://localhost:3000/?price.min=10&price.max=20
- only the product with a price lower than $20 should be displayed, not both.


zenweasel:
@prinzdezibel All PR's should have 1) An issue they are resolving and 2) Instructions for how to test.

Having an issue helps us to have the proper understanding and context to approve a PR

prinzdezibel:
This PR does not have an issue, because I found it accidentally. I just noticed that at one point in the past, it obviously was possible to pass a filter for the products subscription via URL query string, e.g. http://localhost:3000/?price.min=12.00

But I think it's not used any more, because the old code Router.current().queryParams can't possibly work, as the Router does not have a property called "queryParam". Correct would be "query". The current code just returns undefined, regardless of the URL query string.

@spencer: AFAIK, Router.current().params  refers to parameters in URL slugs, e.g. /tags/:name:/

zenweasel:
@prinzdezibel So this issue still needs 1) An issue it is resolving (even if you open it yourself) and 2) Instructions how to test

prinzdezibel:
@zenweasel Done. Find instructions on how to test in #3496.

mikemurray:
@prinzdezibel I think what @zenweasel  means is putting what was fixed, and how to test should go in the original post of this PR, in addition to the ticket being fixed. That way all the information needed and any additional concerns are captured in one place for testers.


**Example**
---

Fixes #ticketNumber

What I fixed

### To Test
1. step 1
1. step 2
1. step 3

spencern:
This PR looks pretty straightforward, let's get some test steps and get it merged or closed. :)

prinzdezibel:
As written above:

> Find instructions on how to test in #3496.

prinzdezibel:
Copied here verbatim to have it side-by-side:

> ## Steps to reproduce the behavior
> Create one product between 10 and 20 $
> Create a second product over 20 $
> paste this URL into address bar of browser: http://localhost:3000/?price.min=10&price.max=20
> see both products being displayed instead of only the first

prinzdezibel:
@spencern Can we get this merged? Lately, somebody on Gitter asked for this specifically.
Thank you !

spencern:
@prinzdezibel I’d like to.

prinzdezibel:
@spencern Is there something I can do? Test instructions are added meanwhile.

PR #3393, opened 2017-12-11
(fix): 3392 - remove shopId check on payment methods for merchent orders

### LINKED ISSUE #3392: Package check mismatch on payment methods
When a merchant shop is processing an order that was created using a marketplace payment method, there is an error in the orders panel caused by a mismatch between the packageId and the current shop ID - the package belongs to the marketplace, but the current shop is the merchant, so a package would never be found, causing data to be missing.

```
shippingInvoice.js:136 Uncaught TypeError: Cannot read property 'settings' of undefined
    at Object.clickDataEventActionCancelOrder (shippingInvoice.js:136)
    at blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3775
    at Function.Template._withTemplateInstanceFunc (blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3744)
    at Blaze.View.<anonymous> (blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3774)
    at blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:2617
    at Object.Blaze._withCurrentView (blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:2271)
    at Blaze._DOMRange.<anonymous> (blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:2616)
    at HTMLButtonElement.<anonymous> (blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:863)
    at HTMLDivElement.dispatch (jquery.js?hash=e5fc9cebb2e37c48ddfaf76e0ab0b77cb6ce6e52:4721)
    at HTMLDivElement.elemData.handle (jquery.js?hash=e5fc9cebb2e37c48ddfaf76e0ab0b77cb6ce6e52:4389)
```

shippinginvoice_js_ ___projects_reaction_foray-reaction

kieckhafer:
Fixes #3392 

An error was occurring when trying to process orders as a merchant shop, using a payment method that belonged to the marketplace owner shop.

The package was checking for a packageId && a shop ID, however the shop ID was the active shop, and the packageId belongs to the marketplace shop, so if you were a merchant processing an order, there would never be a match here, causing an error.

the packageIds are specific enough where we don't need this extra shopId check.

### DIFF
--- a/imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
+++ b/imports/plugins/core/orders/client/templates/workflow/shippingInvoice.js
@@ -130,8 +130,7 @@ Template.coreOrderShippingInvoice.events({
     const settingsKey = paymentMethod && paymentMethod.paymentSettingsKey;
     // check if payment provider supports de-authorize
     const checkSupportedMethods = Packages.findOne({
-      _id: packageId,
-      shopId: Reaction.getShopId()
+      _id: packageId
     }).settings[settingsKey].support;
 
     const orderStatus = paymentMethod && paymentMethod.status;
--- a/server/methods/core/orders.js
+++ b/server/methods/core/orders.js
@@ -1077,8 +1077,7 @@ export const methods = {
     const settingsKey = paymentMethod.paymentSettingsKey;
     // check if payment provider supports de-authorize
     const checkSupportedMethods = Packages.findOne({
-      _id: packageId,
-      shopId: Reaction.getShopId()
+      _id: packageId
     }).settings[settingsKey].support;
 
     const orderMode = paymentMethod.mode;


spencern:
@zenweasel will you look at this PR as well. I think you reviewed the PR that these lines came in from and I want to be sure that I'm not missing anything important.

zenweasel:
Yes, there's absolutely no reason to check for shopId if you are grabbing the package by packageId (like @spencern said). I feel pretty confident it was probably a copy and paste artifact.

PR #3403, opened 2017-12-13
(fix): orders date filter not appearing

### LINKED ISSUE #3402: Cannot show date picker filter in orders panel
### Expected behavior
Clicking the arrow to show the date filters in the orders panel should show a popover with a date picker.

### Actual behavior
The popover is not shown and an error appears in the console.

reaction

With all npm packages up to date on the `master` branch. Open the orders panel and try to reveal the date picker.

## Research

It seems like in the latest version of the `react-dates` package we need to initialize before use.

This is how the imports should look in `calendarPicker.js`.

```
import "react-dates/initialize";
import { DayPickerRangeController } from "react-dates";
```

While the above should allow the date picker to show up, **all** of the CSS is broken. Seems they refactored all of the class names. Seems most of them seem similar, but with removed hyphens and added camelCasing.

**summary**
- Need to initialize before use `import "react-dates/initialize";`
- Need to update all CSS in `calendarPicker.less`

### Versions

Node: 7.2.0
NPM: 5.3.0
Meteor Node: 8.8.1
Meteor NPM: 5.4.2
Reaction CLI: 0.24.2
Reaction: 1.6.1
Reaction branch: master
Docker: 17.09.0-ce

mikemurray:
Fixes #3402 

This PR updates the `CalendarPicker` component as a recent update of `react-dates` introduced some major braking changed

## Highlights of this change

- React dates must be Initialized react dates before importing
- CSS classNames and some properties re-worked to be compatible `react-dates`
- Fixes RTL support for the calendar picker.

## To Test

With a copy of this branch and a fresh `meteor npm install`...

- Open the orders dashboard and attempt to use the date filters. It should work as it used to.
- Add some orders to test if the date filtering still works.
- Switch to an RTL language, see that the date picker is now RTL enabled

### DIFF
--- a/imports/plugins/core/ui/client/components/calendarPicker/calendarPicker.js
+++ b/imports/plugins/core/ui/client/components/calendarPicker/calendarPicker.js
@@ -20,6 +20,23 @@ class CalendarPicker extends Component {
     };
   }
 
+  /**
+   * Getter returning a value of `true` if the current drection is set to RTL
+   * @return {Boolean} true if RTL false if LTR
+   */
+  get isRTL() {
+    if (typeof this.props.isRTL === "boolean") {
+      // If isRTL is set from props, then use
+      return this.props.isRTL;
+    } else if (document) {
+      // Otherwise try to determine RTL status from the html class name "rtl"
+      return document.getElementsByTagName("html")[0].className.includes("rtl");
+    }
+
+    // Return false if above matches fail
+    return false;
+  }
+
   onDatesChange = ({ startDate, endDate }) => {
     this.setState({ startDate, endDate });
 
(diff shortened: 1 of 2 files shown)


spencern:
@mikemurray this still has the [WIP] tag, but I see you moved it to review/qa is it ready for a review?

mikemurray:
@spencern I think it's automatically moved when its a PR for some reason. I didn't personally move it.

mikemurray:
The thing I'm not sure about is it's only December, but all other months make the container resize properly. Perhaps it has something to do with the number of physical weeks in a month.

mikemurray:
@joykare I added a min-height to the transition container so that long months like this December aren't partially cut off. The caveat of this, is now the container is longer than necessary for all the other months.

PR #3343, opened 2017-11-22
Zoom image in PDP

Akarshit:
## Fix for #3100

This uses the [react-image-magnify](https://github.com/ethanselzer/react-image-magnify) component to render the main image in PDP.

I have tuned some small things like the delay in showing the zoomed image, etc. on a _as I like basis_. Would like to get view of others on this.

### To review
We don't support multiple res. of same image right now. So I am basically zooming the image and showing it to the user. Is this the intended behaviour?

### To test
1. Run `reaction`
2. Upload some image in the `example-product`
3. Hover over the image and see it zoomed.
4. Also try the mobile version which responds to long press instead of hover.
5. Also turn off the edit mode `on` and hover on the image, there should be no zoomed image visible.
5. Log out and log in as a customer, you should be able to see the zoomed image on hover.

PR #3329, opened 2017-11-20
Removing check for the tax being 0

Akarshit:
## Fix for #3328 

This basically keep the same tax flow even if the tax value is 0. And doesn't trigger an exception.

PR #3406, opened 2017-12-14
Release 1.7.0

spencern:
# v1.7.0

## Breaking Changes
 - (breaking): (perf): improve products publication performance (#3368)

    **Version 1.7.0 of Reaction contains one potentially breaking change.**

    **tl;dr** This PR stops publishing product variants to the product grid.

    #3368 is a performance focused PR which improves publication performance for the product grid by reducing the number of documents sent to the client. Combined with #3367, we no longer rely on having the variants present on the client to determine the inventory status which was the primary reason that the variants were being published at this point. This will reduce the number of documents published to the client from hundreds or thousands in some shops with products that have several variants and options down to 24 (or whatever number is set for your `SCROLL_LIMIT`.

    This PR also fixes a bug that would cause the admin products publication to occasionally publish all products and variants to the client at once. This PR also introduces an updated media publication for the product grid that publishes only the media necessary to display on the product grid.

    **If your plugin or store relied on product variant documents existing on the client for the product grid, you will need to create a new publication and subscription for your plugin or shop.**

## Fixes
 - (fix): orders date filter not appearing (#3403)
 - Using the default tax rate (#3326)
 - Site will open in a new tab (#3327)
 - Removing check for the tax being 0 (#3329)
 - Polish currency formatting (#3330)
 - (fix): Determine denormalized product inventory status during Shopify import (#3366)
 - (fix): Use denormalized properties on product for gridItemNotices (#3367)
 - Fix: Not being able to scroll after using search function #3314 (#3315)

## Features
 - (feat): allow getShopName() to accept a shopId parameter (#3400)
 - (feat): zoom image in PDP (#3343)

## Updates
-  Upgrade to Meteor v1.6.0.1 (#3407)

## Contributors
Thanks to @Akarshit and @wackywombaat12 for contributing to this release


zenweasel:
@spencern So I am seeing a product I added flagged `backorder` even though I have stock and I have set my "warn at" set to 0. Is there something I am missing?

spencern:
@zenweasel backorder should be set only if the product is `sold out` and the product has `inventoryPolicy` set to `false`

You can see that defined here: https://github.com/reactioncommerce/reaction/blob/release-1.7.0/server/methods/catalog.js#L282-L294

As for what would cause that, it sounds like perhaps the inventory status (qty available and inventoryPolicy) are not being denormalized properly. Do you have reproduction steps for this?

zenweasel:
Reproduction steps for "backorder"

1. Create a new product with a top level variant and set nothing about inventory
1. Create an option and set inventory quantity
1. Publish
1. Go to the product grid, observe the "Backorder" badge is showing

spencern:
Confirmed that this is happening when creating an option after not setting inventory on the variant. Investigating

spencern:
@zenweasel This issue is deeper than I was hoping. It's caused by an issue connected to variants and revisions and if the variant is published at the time that the denormalize method runs.

Have a good start on it, but nothing that's working for all edge cases yet.

spencern:
@zenweasel After some more experimentation, I think we may just need to pull out the performance elements of this and release most of this as 1.6.2 and continue to work on the product publication PR. 
My PR assumed that the denormalization of inventory, low quantity, and backorder status was being done correctly, and it does not appear that is the case. The revision status that products can have, as well as the visibility switch makes this more complicated than I was expecting and there are a lot of edge cases to consider. 

E.g. when a top level product is published and visible but contains unpublished qty changes to a variant's qty that would affect the denormalized status. Or same scenario but the product is hidden.
I've got some ideas as to how to simplify this process, but right now it's a little too messy for me to feel good about pushing out a release.

zenweasel:
@spencern Isn't it #3367 that was the problem? Or the combination?

spencern:
@zenweasel you are right, it's a bit of both. #3367 is a problem because that depends on the denormalization working correctly and displays the grid item notices from the denormalized properties on the product. #3368 is a problem because it removes the part of the products publication that publishes variants. #3368 is dependent on #3367 working correctly.

spencern:
Closing in favor of #3416. 

I've changed the PR and the version on this upcoming release to reflect the fact that we're not introducing any major features or breaking changes. #3367 and #3368 have been reverted from the `release-1.6.2` branch

PR #3407, opened 2017-12-14
METEOR@1.6.0.1

aaronjudd:
METEOR 1.6.0.1

Node has been upgraded to version 8.9.3, an important security release.
The npm package has been upgraded to version 5.5.1.


.meteor/packages
```diff
@@ -33,7 +33,7 @@ service-configuration@1.0.11
 mdg:validated-method
 shell-server@0.3.0
 dynamic-import@0.2.0
-standard-minifier-js@2.2.3
```
spencern: Any idea why it dropped the version for standard minifier?

aaronjudd: Looking into it..

aaronjudd: `The specified packages are at their latest compatible versions.`. Tested-docker builds still function and this was the main deal we had with the migration to this minifier.

spencern: ok, so we're good to merge you think?

aaronjudd: I do 👍

PR #3410, opened 2017-12-15
(fix): resend email from dashboard even if completed

### LINKED ISSUE #3388: Cannot resend emails as an admin if email was successfully sent on the first try
### Expected behavior
As an admin I should be able to resend emails from the email panel.

### Actual behavior

As an admin, I cannot resend emails from the email panel. This is happening exclusively with emails that have already sent successfully once.

I can only resend emails that were unsuccessful on the first try.

- Do not setup email config
- Perform action that sends email, observe that it fails to send
- setup email config 
- resend email, see that it succefully sends
- on the same email try resend it, observe that it fails (status red)

spencern:
Resolves #3388 

Enables admins to resend an email from the email dashboard even if the email was previously successfully sent.

This was not working because of the way that the [job-collection](https://atmospherejs.com/vsivsi/job-collection) treats completed jobs. Only jobs that have failed or been canceled can be restarted. Completed jobs must be rerun. This PR determines what status an email job is, and then reruns or restarts it appropriately.

Side Effect: when you resend an email that was previously successful, an additional entry will be added to the emails table rather than updating the existing entry. This is how the job collection works. Restarting a failed email will update that entry.

### DIFF
--- a/server/methods/email.js
+++ b/server/methods/email.js
@@ -117,10 +117,25 @@ Meteor.methods({
     }
 
     check(jobId, String);
+    let emailJobId = jobId;
 
     Logger.debug(`emails/retryFailed - restarting email job "${jobId}"`);
 
-    Jobs.update({ _id: jobId }, {
+    // Get email job to retry
+    const job = Jobs.getJob(jobId);
+    // If this job was never completed, restart it and set it to "ready"
+    if (job._doc.status !== "completed") {
+      job.restart();
+      job.ready();
+    } else {
+      // Otherwise rerun the completed job
+      // `rerun` clones the job and returns the id.
+      // We'll set the new one to ready
+      emailJobId = job.rerun(); // Clone job to rerun
+    }
+
+    // Set the job status to ready to trigger the Jobs observer to trigger sendEmail
+    Jobs.update({ _id: emailJobId }, {
       $set: {
         status: "ready"
       }


server/methods/email.js
```diff
@@ -117,10 +117,25 @@ Meteor.methods({
...
     check(jobId, String);
+    let emailJobId = jobId;
 
     Logger.debug(`emails/retryFailed - restarting email job "${jobId}"`);
 
-    Jobs.update({ _id: jobId }, {
+    // Get email job to retry
+    const job = Jobs.getJob(jobId);
+    // If this job was never completed, restart it and set it to "ready"
+    if (job._doc.status !== "completed") {
```
impactmass: We probably should add why we're adding this logic (to help future debugging), what do you think?
I mean the fact that we're doing this because jobs weren't getting resent successfully?

spencern: @impactmass Do you not think the inline comments are enough, or just need more explanation overall?

spencern:
@impactmass Re: potential issue with having multiple emails showing up. I thought about that, but also thought if a user was having trouble receiving emails from a shop that it might be better for the shop owner to be able to say how many emails had been sent and at what times. They do have different timestamps.

Could go either way on it really, and it'd be a simple change.

server/methods/email.js
impactmass: The current comment is good.... but I meant adding something around the reason why those steps are now needed. Something like what you have in the PR description.
`.....because of the way that the job-collection treats completed jobs. Only jobs that have failed or been canceled can be restarted. Completed jobs must be rerun.`

impactmass:
Re the _duplicates_: Okay, makes sense. Let's keep them since the different timestamps on them give them a valid reason to be displayed.

PR #3411, opened 2017-12-15
[WIP] #3373 (breaking change) Convert paymentMethod.supports into an object

iakhator:
Working on #3373

This PR that does the following:
- Refactor payment schema with an array in `support` property
- Change the `array of String` to `object of Boolean`
- Make sure payment work properly.
- Migration of modified schema

#### Test 1
- Log in as an admin
- Open the payment settings panel
- Enable any payment method.
- Go to `Payment provider supported methods` check  or uncheck any of the supported methods the click on `save changes`. Notice the toast with a message of updated.
- Go to packages open `settings-> paymentMethodname->support` depending on the current payment method  notice the type of `Boolean` and the changed from either true to false or false to true.

#### Test 2
- Log in admin
- Open the payment settings panel
- Enable any payment method.
- Place an order and checkout.
- You will see a successful item purchase.
- Go to packages open `settings-> paymentMethodname->support` depending on the current payment method  notice the type of `Boolean` and the changed from either true to false or false to true.


zenweasel:
@iakhator You have some lint errors there to fix

zenweasel:
Just heads-up that you will probably need to write a migration to migrate people's current settings to this new format so we don't break the site for people with an existing registry. Look in `import/plugins/core/versions/server/migrations` for some examples.

zenweasel:
@prinzdezibel Could you take a look at this ticket as well? Testing it would involve going through every payment method and verifying that checkout/approve/refund/ship all work for each method.

imports/plugins/core/versions/server/migrations/19_add_object_of_boolean_to_payments_schema.js
```diff
@@ -0,0 +1,120 @@
...
+        // Loop through all packages and update support to be an object
+        packages.forEach((pkg) => {
+          // Lopp through all paypalsupport of payment to get each paymentmethod
+          Object.keys(PaypalSupport).forEach((paypal) => {
+            support = pkg.settings[PaypalSupport[paypal]].support;
+            // Checking to make sure support is an array
+            if (Array.isArray(support)) {
+              const supportObject = {};
+              support.forEach((method) => {
+                supportObject[method.toLowerCase()] = true;
```
prinzdezibel: @iakhator You'd also need to deal with the dash in the old value `De-authorize` which will be migrated to `de-authorize`. Expected by schema is `de_authorize`. This results in potential settings losses during migration as seen in screenshot:
robo_3t_-_1_1_and_men_and__3373_convert_paymentmethod_supports_into_an_object_by_iakhator_ _pull_request__3411_ _reactioncommerce_reaction

imports/plugins/core/orders/client/containers/invoiceContainer.js
```diff
@@ -223,9 +223,9 @@ class InvoiceContainer extends Component {
     const paymentMethodId = orderBillingInfo.paymentMethod && orderBillingInfo.paymentMethod.paymentPackageId;
     const paymentMethodName = orderBillingInfo.paymentMethod && orderBillingInfo.paymentMethod.paymentSettingsKey;
     const paymentMethod = Packages.findOne({ _id: paymentMethodId });
+    const refund = _.get(paymentMethod.settings[paymentMethodName].support, "refund");
```
prinzdezibel: There's another problem in Line 273:
invoiceactions_js_-_reaction_-____github_reactioncommerce_reaction__and_reactioncommerce_reaction_-_gitter

prinzdezibel: @iakhator  Also in shippingInvoice.js, line 130 ff

prinzdezibel: @iakhator See also exampleSettingsForm.js from payments-example:
examplesettingsform_js_-_reaction_-____github_reactioncommerce_reaction_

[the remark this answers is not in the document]
zenweasel: Because you risk causing an error that would kill checkout I would rather we be extra cautious when testing for deeply nested properties like this. (although if `supports` doesn't exist you probably have other issues, but a runtime error is not the way to handle it.)

iakhator:
@prinzdezibel so all changes have been made accordingly. please review.

zenweasel:
It looks like if I migrate I lose all my settings

### Steps to reproduce

1. Start at current `master`
1. Start Reaction and set all methods to only support "de-authorize"
1. Checkout this branch and start
1. Notice that all the settings are either completely different or not set

imports/plugins/included/payments-paypal/client/templates/settings/express.html
```diff
@@ -21,6 +21,10 @@
     {{>afQuickField name='settings.password' type="password"}}
     {{>afQuickField name='settings.signature'}}
     {{>afQuickField name="settings.express.support" options="allowed" noselect=true}}
+     {{>afQuickField name="settings.express.support.authorize" type="boolean-checkbox" options="allowed" noselect=true}}
+        {{>afQuickField name="settings.express.support.de_authorize" type="boolean-checkbox" options="allowed" noselect=true}}
```
zenweasel: Is there a reason these are indented?

imports/plugins/included/payments-stripe/register.js
```diff
@@ -11,11 +11,11 @@ Reaction.registerPackage({
     "api_key": "",
     "reaction-stripe": {
       enabled: false,
-      support: [
-        "Authorize",
-        "Capture",
-        "Refund"
-      ]
+      support: {
+        authorize: true,
```
zenweasel: Why isn't de-authorize included here?

server/methods/core/orders.js
```diff
@@ -383,14 +383,14 @@ export const methods = {
     const paymentMethodId = paymentMethod && paymentMethod.paymentPackageId;
     const paymentMethodName = paymentMethod && paymentMethod.paymentSettingsKey;
     const getPaymentMethod = Packages.findOne({ _id: paymentMethodId });
+    const refund = _.get(getPaymentMethod.settings[paymentMethodName].support, "refund");
```
zenweasel: This variable should be named using a boolean name like `isRefundEnabled`

imports/plugins/included/payments-stripe/register.js
iakhator: I was following the pattern of the initial `Array` since this is an object now it is a good to make de-authorize `false`

iakhator:
@zenweasel , I have done the necessary  clean up and for the migration, following your steps above to reproduce, after checking `de-authorize` as true on master and returning to my branch my `de-authorize` remain true for all methods.

iakhator:
@prinzdezibel, eslint was flagging the underscore. That is the reason for using the bracket and quotes.

zenweasel:
Changed Example-Payment to not support refund since it does not. Otherwise tested all payment methods and seems to be working correctly.

prinzdezibel:
I've also tested also all payments methods (actually I've forgotten example plugin). Everything worked as expected.

spencern:
Before merging this, I'd like to see how it plays with custom payment providers that aren't included.
Adding WIP tag so we don't accidentally merge it for now.

prinzdezibel:
Good catch, @spencern! This PR introduces breaking changes which I missed to declare as such. Settings forms for custom payment providers need to be adapted to conform to this change.

zenweasel:
I agree with @spencern that instead of using the hard-coded values the migrations should be iterating over all registered payment packages so that we don't break custom payment packages.

I also agree with @prinzdezibel's note that we probably shouldn't allow users to enable methods that their payment provider doesn't support so we can just remove these from the settings panels. I like the functionality of being able to disable refunds but we should either only do that, or just let people modify the values in the registry.

So I am going to close this PR for now and I will take this over and make the changes and reopen when it's ready.

PR #3417, opened 2017-12-18
(fix): Import PropTypes from "prop-types" package

mikemurray:
Fix what seems to be the last instance of importing PropTypes from the "react" package. Now imports it from the "prop-types" package as it should.

### To Test

- start reaction
- see no warning in console
- rejoice

### DIFF
--- a/imports/plugins/core/ui/client/providers/dragDropProvider.js
+++ b/imports/plugins/core/ui/client/providers/dragDropProvider.js
@@ -6,23 +6,6 @@ import { registerComponent } from "@reactioncommerce/reaction-components";
 
 const defaultManager = new DragDropManager(HTML5Backend);
 
-// /**
-//  * This is singleton used to initialize only once dnd in our app.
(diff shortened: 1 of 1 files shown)


imports/plugins/core/ui/client/providers/dragDropProvider.js
```diff
@@ -1,27 +1,11 @@
-import React, { Component, PropTypes, Children } from "react"; // eslint-disable-line
+import { Component, Children } from "react";
+import PropTypes from "prop-types";
 import { DragDropManager } from "dnd-core";
 import HTML5Backend from "react-dnd-html5-backend";
 import { registerComponent } from "@reactioncommerce/reaction-components";
 
 const defaultManager = new DragDropManager(HTML5Backend);
 
-// /**
```
zenweasel: Is this comment no longer relevant?

mikemurray: No, as it applied to that commented out function below it.

PR #3419, opened 2017-12-18
LingoHub based on release-1.6.2

aaronjudd:
Pull request created by LingoHub based on branch release-1.6.2


spencern:
Closing, I think we need to do another LingoHub update before we can trust it's PRs again.

PR #3423, opened 2017-12-19
Fix for #3431. Correct style applied for guest users.

Akarshit:
### What was wrong
I didn't apply the CSS for image height when the user was not the admin user.

### To test ->
The point is basically to load a big image that would take some time and test. These instruction are for testing in local
1. Run `reaction`.
2. Copy the URL of a image(from somewhere on the internet).
3. Add `return <image URL>;` [here](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/ui/client/components/media/media.js#L87) (just return the URL from the function)
4. Add `BrowserPolicy.content.allowOriginForAll("*");` after [this](https://github.com/reactioncommerce/reaction/blob/master/server/security/policy.js#L54)
5. Go to the PDP of example-product. The image should load and show.


zenweasel:
@Akarshit This first link for where to make the `return <image URL>` change is broken.

Akarshit:
@zenweasel Sorry. Updated now.

imports/plugins/included/default-theme/client/styles/media.less
```diff
@@ -164,6 +164,21 @@
 
 }
 
+.gallery-image.no-fade-on-hover {
```
pmn4: any sense in making this a mixin to use here and above (`.admin-gallery-image.no-fade-on-hover`)?

PR #3428, opened 2017-12-21
fix: ‘order’ should be translated to '订单' rather than ‘命令’ in Chinese

unaheidi:
In Chinese, we use "订单" to express ’order for goods‘ rather than ’命令‘.
And when Chinese people use ’命令‘, we want to express ‘command’ .
So I suggest that 'order' in ecommerce should be translated to "订单" in Chinese.

### DIFF
--- a/imports/plugins/core/orders/server/i18n/zh.json
+++ b/imports/plugins/core/orders/server/i18n/zh.json
@@ -14,7 +14,7 @@
       },
       "admin": {
         "shortcut": {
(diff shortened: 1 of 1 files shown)


zenweasel:
@unaheidi Thanks for so much for the contribution! Please make sure and enter our hack-a-bug-a-thon [here](getrxn.io/hackabugathon) for a chance to win gift cards or swag.

PR #3430, opened 2017-12-22
(feat): getProduct method

kieckhafer:
Add `getProduct` method to ReactionProduct API.


pmn4:
Hi @kieckhafer, this change seems reasonable to me, but I'm curious, when would you expect to use this method?

I was thinking that to get a product you would subscribe to the `"Product"` Meteor event.

Of course, this only gives you the "current" product, as defined by the URL, so I understand if you are looking to find other products. That said -- and, assuming I understand things correctly! -- in the client-side code `Products.findOne(id);` does not go to the database, and will return `null` unless 1) the `"Products"` event has triggered and 2) the Product you are looking for is contained in the subset of products returned in the [`"Products"` query](https://github.com/reactioncommerce/reaction/blob/master/server/publications/collections/products.js#L440-L443)

Even after that, I am not saying this is not worthy of adding, I just think it could cause some headaches when called from the client... so maybe just add some additional documentation?

further, please let me know if I am misunderstanding anything!

kieckhafer:
@pmn4 I think you're understanding correctly.

This method stemmed from a client project, where we needed product info of a `related` product on our product grid. Since it was on the product grid, the related product was also published via the publication, so the data was available.

When building it out, we realized we already had `ReactionProduct` methods for getting `getVariant` and `getVariantParent` and `getTopLevelVariant`, and wanted to add this related method to make it easier to directly get product info.

PR #3431, opened 2017-12-22
Handle null Addresses when sending emails

pmn4:
I tried to send a reset password email today and the action failed due to a missing address (which is strange, since the current shop definitely _has_ an address... I'll tackle that issue next). This code adds some safety to grabbing details about the first address.

To reproduce, try to send a reset password in a store with no addresses (tbh, I'm not sure how I wound up in this state... you may have to update the document in the DB).

---
_(the code, btw, is copied from [accounts.js](https://github.com/reactioncommerce/reaction/blob/pmn4%2Fphysical-address/server/methods/accounts/accounts.js#L968))_

### DIFF
--- a/server/api/core/accounts/password.js
+++ b/server/api/core/accounts/password.js
@@ -74,12 +74,12 @@ export function sendResetPasswordEmail(userId, optionalEmail) {
     homepage: Meteor.absoluteUrl(),
     emailLogo: emailLogo,
     copyrightDate: moment().format("YYYY"),
-    legalName: shop.addressBook[0].company,
+    legalName: _.get(shop, "addressBook[0].company"),
     physicalAddress: {
-      address: shop.addressBook[0].address1 + " " + shop.addressBook[0].address2,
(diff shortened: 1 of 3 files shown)


zenweasel:
@pmn4 Does this branch still exist in your repo. I am not able to pull it
```
 »» git pull https://github.com/pmn4/reaction.git physical-address
fatal: Couldn't find remote ref physical-address
```

pmn4:
@zenweasel I moved the branch to this repository (the prefix is probably throwing you off... it's a habit I developed working on larger teams).

the branch name is `pmn4/physical-address`, so `git pull origin pmn4/physical-address`

zenweasel:
@pmn4 Pat, if you haven't already, make sure you enter the Hack-a-bug-a-thon [here](getrxn.io/hackabugathon)

zenweasel:
Tested by setting `Shops.address` to `[]` in the database while on `master` and the app would not load. Switched to this branch and the app loaded and seemed to function.

Issue #3426, opened 2017-12-20
Add Event.Hooks hooks to all relevant events


zenweasel:
We need to add `Event.Hooks` to add the Events as defined in this document created by @spencern 
 https://www.lucidchart.com/documents/edit/77b3a316-cfce-4645-935f-2770f09fe1ac/0

zenweasel:
This is covered in multiple other issues

PR #3439, opened 2018-01-03
#3426 Add hooks to cart and order events

codejockie:
This PR adds hooks to cart and order events. It currently have hooks setup for just the cart methods.

Closes ##3426

### DIFF
--- a/server/methods/catalog.js
+++ b/server/methods/catalog.js
@@ -616,7 +616,7 @@ Meteor.methods({
     const productId = toDelete[0].ancestors[0];
     toDenormalize.forEach(field => denormalize(productId, field));
 
-    Hooks.Events.run("onProductVariantDelete", deleted);
+    Hooks.Events.run("onProductVariantDelete", toDelete);
     return typeof deleted === "number" && deleted > 0;
   },
 
@@ -757,7 +757,7 @@ Meteor.methods({
       }
     }
 
-    Hooks.Events.run("onProductClone", products);
+    Hooks.Events.run("onProductClone", { ...products, source: productOrArray });
     return results;
   },
 
@@ -800,7 +800,6 @@ Meteor.methods({
           title: "",
           type: "variant" // needed for multi-schema
         });
-        Hooks.Events.run("onProductCreate", result);
       }
     });
   },
--- a/server/methods/catalog.js
+++ b/server/methods/catalog.js
@@ -6,7 +6,7 @@ import { Meteor } from "meteor/meteor";
 import { copyFile, ReactionProduct } from "/lib/api";
 import { ProductRevision as Catalog } from "/imports/plugins/core/revisions/server/hooks";
 import { Media, Products, Revisions, Tags } from "/lib/collections";
-import { Logger, Reaction } from "/server/api";
+import { Hooks, Logger, Reaction } from "/server/api";
 
 /* eslint new-cap: 0 */
 /* eslint no-loop-func: 0 */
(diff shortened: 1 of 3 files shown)


server/methods/catalog.js
```diff
@@ -613,6 +616,7 @@ Meteor.methods({
     const productId = toDelete[0].ancestors[0];
     toDenormalize.forEach(field => denormalize(productId, field));
 
+    Hooks.Events.run("onProductVariantDelete", deleted);
```
zenweasel: `deleted` here is just the results of the `remove` call. I think you want to at least pass in `toDelete` or some info about what product is being deleted

server/methods/catalog.js
```diff
@@ -793,6 +800,7 @@ Meteor.methods({
           title: "",
           type: "variant" // needed for multi-schema
         });
+        Hooks.Events.run("onProductCreate", result);
```
zenweasel: All of these product hooks should have some consistency. `result` here is just going to be the results of the db call, not any information about the call.

server/methods/catalog.js
```diff
@@ -752,6 +756,8 @@ Meteor.methods({
         results.push(result);
       }
     }
+
+    Hooks.Events.run("onProductClone", products);
```
zenweasel: `onProductClone` should probably included both the source and the destination products.

server/methods/core/cart.js
```diff
@@ -203,6 +203,8 @@ Meteor.methods({
             items: { $each: mergedItems, $slice: -(mergedItems.length) }
           }
         });
+
+        Hooks.Events.run("onCartMerge", sessionCart, [cartId]);
```
zenweasel: Why is this being result as an array?

server/methods/core/cart.js
```diff
@@ -426,6 +429,8 @@ Meteor.methods({
         Logger.debug(`cart: increment variant ${variantId} quantity by ${
           quantity}`);
 
+        Hooks.Events.run("onCartAdd", cart);
```
zenweasel: The hooks for add to cart should also include what is being added

server/methods/core/cart.js
```diff
@@ -574,6 +582,7 @@ Meteor.methods({
     Meteor.call("workflow/revertCartWorkflow", "coreCheckoutShipping");
     // reset selected shipment method
     Meteor.call("cart/resetShipmentMethod", cart._id);
+    Hooks.Events.run("onCartRemove", cart);
```
zenweasel: The event for remove Cart should include what is being removed.

server/methods/core/cart.js
```diff
@@ -636,6 +645,7 @@ Meteor.methods({
     // update or insert method
     try {
       Collections.Cart.update(selector, update);
+      Hooks.Events.run("onCartShipmentMethodSet", cart);
```
zenweasel: The hook for this should include the shipment method that is being set. This applies to almost all of these cart methods.

server/methods/core/cartToOrder.js
```diff
@@ -173,6 +173,8 @@ export function copyCartToOrder(cartId) {
   const orderId = Collections.Orders.insert(order);
 
   if (orderId) {
+    Hooks.Events.run("onOrderCreate", order, [orderId]);
```
zenweasel: Why would we pass both order and orderId? And why is orderId being passed as the only member of an array?

zenweasel:
Based on conversations with @spencern and others I think we need to make sure we do this more methodically from @spencern's chart, and we may want to change/extend the implementation of Event.Hooks. So closing this PR but leaving the original issue open.

PR #3441, opened 2018-01-04
(fix): Product images missing from confirmation emails

### LINKED ISSUE #3391: Product images missing from confirmation emails
### Expected behavior
Product images should be displayed in order emails

### Actual behavior
A placeholder image is displayed in the an order email

- Upload an image to a product
- Create an order with that product
- See that the product image is not displaying in the order email received

### Versions
Node: 9.2.1
NPM: 5.6.0
Meteor Node: 8.8.1
Meteor NPM: 5.4.2
Reaction CLI: 0.24.2
Reaction: 1.6.0
Reaction branch: master
Docker: 17.11.0-ce

impactmass:
Closes #3391 (see last comment on the issue)

How To Test
- Create a new product with an image. (or use the default Example Product, but first, upload an image for it)
- As a customer, make an order with the above product
- Open the order confirmation email sent to you, and confirm the image for the product shows up (It should not be blank, and it should not use the default placeholder image, since there's a picture on the product.
  
Note: If you tested this on localhost and used Gmail as your email sending service, Gmail (sometimes) mask out urls containing localhost or 127.0... This will make it appear like the image is missing, but the correct link is present (just masked).

### DIFF
--- a/private/email/templates/orders/new.html
+++ b/private/email/templates/orders/new.html
@@ -265,6 +265,8 @@
                                                       <td valign="middle" align="left">
                                                         {{#if variantImage}}
                                                           <img src="{{variantImage}}"  width="50" height="50" alt="" />
+                                                        {{else if productImage}}
+                                                          <img src="{{productImage}}"  width="50" height="50" alt="" />
                                                         {{else}}
                                                           <img src="{{placeholderImage}}" width="50" height="50" alt="" />
                                                         {{/if}}
--- a/server/methods/core/orders.js
+++ b/server/methods/core/orders.js
@@ -699,12 +699,12 @@ export const methods = {
           });
           // variant image
           if (variantImage) {
-            orderItem.variantImage = path.join(Meteor.absoluteUrl(), variantImage.url());
+            orderItem.variantImage = Meteor.absoluteUrl(variantImage.url());
           }
           // find a default image
           const productImage = Media.findOne({ "metadata.productId": orderItem.productId });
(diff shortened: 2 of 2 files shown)


private/email/templates/orders/new.html
```diff
@@ -265,6 +265,8 @@
                                                       <td valign="middle" align="left">
                                                         {{#if variantImage}}
                                                           <img src="{{variantImage}}"  width="50" height="50" alt="" />
+                                                        {{else if productImage}}
+                                                          <img src="{{productImage}}"  width="50" height="50" alt="" />
```
impactmass: In some cases, the variantImage is not available, but a productImage is available, and can be used instead of falling to the default.

server/methods/core/orders.js
```diff
@@ -699,12 +699,12 @@ export const methods = {
           });
           // variant image
           if (variantImage) {
-            orderItem.variantImage = path.join(Meteor.absoluteUrl(), variantImage.url());
+            orderItem.variantImage = Meteor.absoluteUrl(variantImage.url());
```
impactmass: I found that Path.join (and [node path module](https://nodejs.org/api/path.html)) is used for file/system paths, and less useful for urls as we have here. (introduced in #1501). The fact that [absoluteUrl](https://docs.meteor.com/api/core.html#Meteor-absoluteUrl) can take a string argument which it joins with the root_url solves this problem better.

PR #3451, opened 2018-01-06
(feat): Split shop invite email templates for existing or new user accounts

### LINKED ISSUE #3449: User group invite emails do not consider whether a user has an account already
### Expected behavior
When an existing user is invited to a user group in the Accounts dashboard, the email should not include a "Get Started" button that links to the password setting page.

### Actual behavior
Inviting an existing user generates an email with a link to set the user's password. Because the user already exists, this process hangs

1. Create a guest account.
2. From the Admin, invite a user with the just registered email to become a shop manager.
3. Follow the link from the email.

### Versions
Node: 9.2.1
NPM: 5.6.0
Meteor Node: 8.9.3
Meteor NPM: 5.5.1
Reaction CLI: 0.24.2
Reaction: 1.6.3
Reaction branch: master
Docker: 17.11.0-ce

kieckhafer:
Fixes #3449 

When an existing user was invited to become a shop manager / invited to a group on a shop, the user was sent an email that included our enrollment link, which would hang when clicked as the user already existed.

This update makes it so there are two separate emails sent out for invitations, one for existing user accounts, and one for new user accounts. The new user account will work in the same way as the app previously worked. The existing user email will include a link to the landing page, instead of the enrollment link.

**Testing**
- As a shop admin, invite a _NEW_ user to become a shop manager (or other group member) of your shop.
- Receive an email that will direct you to create a password when you click the "Get Started" link in the email

- Create a new user account
- As a shop admin, invite this user you just created to become a shop manager (or other group member) of your shop.
- Receive an email that will direct you to the home / landing page of Reaction, and not require a password setting

### DIFF
--- a/private/email/templates/accounts/inviteShopMember.html
+++ b/private/email/templates/accounts/inviteShopMember.html
@@ -75,7 +75,7 @@ table[class=wrap1001], td[class=wrap1001] { width:96% !important; margin:0 !impo
                   <tr>
                     <td align="left" valign="top"><table width="100%" border="0" cellspacing="0" cellpadding="0">
                       <tr>
-                        <td width="223" height="40" align="center" valign="middle" bgcolor="#1999dd" style="font-family: 'Source Sans Pro', sans-serif; font-size:18px; font-weight:normal; line-height:18px;"><a href="{{url}}" style="text-decoration:none; color:#ffffff;">Vist {{shopName}}</a></td>
+                        <td width="223" height="40" align="center" valign="middle" bgcolor="#1999dd" style="font-family: 'Source Sans Pro', sans-serif; font-size:18px; font-weight:normal; line-height:18px;"><a href="{{url}}" style="text-decoration:none; color:#ffffff;">Visit {{shopName}}</a></td>
                         <td>&nbsp;</td>
                       </tr>
                     </table></td>
--- a/private/email/templates/accounts/inviteShopMember.html
+++ b/private/email/templates/accounts/inviteShopMember.html
@@ -75,7 +75,7 @@ table[class=wrap1001], td[class=wrap1001] { width:96% !important; margin:0 !impo
                   <tr>
                     <td align="left" valign="top"><table width="100%" border="0" cellspacing="0" cellpadding="0">
                       <tr>
-                        <td width="223" height="40" align="center" valign="middle" bgcolor="#1999dd" style="font-family: 'Source Sans Pro', sans-serif; font-size:18px; font-weight:normal; line-height:18px;"><a href="{{url}}" style="text-decoration:none; color:#ffffff;">Get Started</a></td>
+                        <td width="223" height="40" align="center" valign="middle" bgcolor="#1999dd" style="font-family: 'Source Sans Pro', sans-serif; font-size:18px; font-weight:normal; line-height:18px;"><a href="{{url}}" style="text-decoration:none; color:#ffffff;">Vist {{shopName}}</a></td>
                         <td>&nbsp;</td>
                       </tr>
                     </table></td>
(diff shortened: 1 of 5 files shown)


server/methods/accounts/accounts.js
```diff
@@ -678,17 +678,31 @@ export function inviteShopMember(options) {
...
 
+  // If the user already has an account, send informative email, not "invite" email
   if (user) {
-    userId = user._id; // since user exists, we promote the account
+    // The user already exists, we promote the account, rather than creating a new one
+    userId = user._id;
     Meteor.call("group/addUser", userId, groupId);
+
+    // do not send token, as no password reset is needed
+    token = null;
```
kieckhafer: @spencern when reviewing, please take a look at the way I did the token / url here, not sure if we need both of these.

Initially, I just set the token to be null here, which would in turn return just the `Meteor.absoluteUrl()` in the `getEmailUrl` function on line 1017.

After doing this, I decided to add the extra url `const` here, in case we would ever want to pass another URL directly in. I think this was is better, however I wasn't sure about removing the update to the `token` here, because it doest make sense to leave it in case another URL isn't provided.

private/email/templates/accounts/inviteShopMember.html
```diff
@@ -75,7 +75,7 @@
                   <tr>
                     <td align="left" valign="top"><table width="100%" border="0" cellspacing="0" cellpadding="0">
                       <tr>
-                        <td width="223" height="40" align="center" valign="middle" bgcolor="#1999dd" style="font-family: 'Source Sans Pro', sans-serif; font-size:18px; font-weight:normal; line-height:18px;"><a href="{{url}}" style="text-decoration:none; color:#ffffff;">Get Started</a></td>
+                        <td width="223" height="40" align="center" valign="middle" bgcolor="#1999dd" style="font-family: 'Source Sans Pro', sans-serif; font-size:18px; font-weight:normal; line-height:18px;"><a href="{{url}}" style="text-decoration:none; color:#ffffff;">Visit {{shopName}}</a></td>
```
kieckhafer: "Visit {{shopName}}" isn't the most thrilling language. Should we leave it how it was, to say "Get Started"? Or change this to something else? Or keep it as "Visit {{shopName}}"?

spencern: I'd agree about the language here. @rymorgan or @aaronjudd any thoughts?

"Get Started" could still work since there's a new link. 
There's no acceptance process right now I don't think, so there's nothing this user has to accept and this email is more confirmation that their permissions have been upgraded for the specified shop. Visit {shopName} kinda works in that sense too.

server/methods/accounts/accounts.js
```diff
@@ -994,7 +1011,7 @@ function getDataForEmail(options) {
     user: Meteor.user(), // Account Data
     currentUserName,
     invitedUserName: name,
-    url: getEmailUrl(token)
+    url: url || getEmailUrl(token)
```
spencern: I see how this works now, passing in a null token might be the simplest solution, though if you're also passing in a url, it doesn't seem like it should matter.

kieckhafer:
One future concern with the way these emails are set up is there is no way to provide custom data to the emails without touching core code.

For future consideration, we should look into a way to pass in data.

server/methods/accounts/accounts.js
```diff
@@ -678,17 +678,31 @@ export function inviteShopMember(options) {
...
+    userId = user._id;
     Meteor.call("group/addUser", userId, groupId);
+
+    // do not send token, as no password reset is needed
+    token = null;
+    const url = Meteor.absoluteUrl();
+
     // use primaryShop's data (name, address etc) in email copy sent to new shop manager
-    dataForEmail = getDataForEmail({ shop: primaryShop, currentUserName, name, token, emailLogo });
+    dataForEmail = getDataForEmail({ shop: primaryShop, currentUserName, name, token, emailLogo, url });
```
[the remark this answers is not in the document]
kieckhafer: Nope. Good catch. It's removed.

kieckhafer:
@spencern I've updated a couple things based on our discussion and notes here. I think this should be ready to go, with a note added above concerning the future customization of the data.

PR #3331, opened 2017-11-20
Use NPM SimpleSchema rather than Meteor 

aldeed:
**BREAKING CHANGES**

This PR updates the `aldeed:simple-schema` Meteor package dependency to instead depend on the `simpl-schema` NPM package, which is the newest release of the same library. As part of this change, there are several breaking changes and other gotchas to be aware of.

- **IMPORTANT!** The NPM package does not play nice with the previous Meteor package. After updating to this Reaction release, run the app one time, and then look at the `.meteor/versions` file. Make sure that `aldeed:simple-schema` is not listed. If it is there, that is because you depend on another Meteor package that depends on `aldeed:simple-schema`. You will have to update or remove any such packages (with `meteor remove` / `meteor add`) until `aldeed:simple-schema` disappears from your `.meteor/versions` file.
- Search your app for any `import { SimpleSchema } from "meteor/aldeed:simple-schema"` lines that you have added in your custom code, and replace them with `import SimpleSchema from "simpl-schema"`
- Be aware that the package name does not have the "e" on "simpl". (There is a different NPM package called `simple-schema` with the "e", and that is NOT the one you want.)
- If you have your own custom schemas, refer to the SimpleSchema changelog to update them for the breaking changes: https://github.com/aldeed/meteor-simple-schema/blob/master/CHANGELOG.md#200
- If you use `attachSchema` in your code, be aware that passing an array to `attachSchema` will no longer work. You should first merge all the schemas and then pass the combined schema to `attachSchema`

PR #3461, opened 2018-01-10
s/Schemas.Shops/Shops.simpleSchema()

pmn4:
1) Modify Shop's Schema. for sake of argument:
``` JavaScript
export const ExternalShop = new SimpleSchema({
  externalId: {
    type: Number,
    optional: true
  }
});

Shops.attachSchema(ExternalShop);
```

2) create a shop with `Meteor.call("shop/createShop")`, passing shop data which includes the new externalId.

issue: createShop fails validation because External Id is not known to a Shop's _original_ schema (`Schemas.Shop`).
fix: When a Shop's _current_ schema is used (`Shops.simpleSchema()`), validation passes.

---
I will write tests before merging, but hope to get your thoughts before going too far.

### DIFF
--- a/server/methods/core/shop.js
+++ b/server/methods/core/shop.js
@@ -27,7 +27,7 @@ Meteor.methods({
    */
   "shop/createShop": function (shopAdminUserId, shopData) {
     check(shopAdminUserId, Match.Optional(String));
-    check(shopData, Match.Optional(Schemas.Shop));
+    check(shopData, Match.Optional(Collections.Shops.simpleSchema()));
 
     // Get the current marketplace settings
     const marketplace = Reaction.getMarketplaceSettings();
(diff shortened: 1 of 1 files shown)


server/methods/core/shop.js
```diff
@@ -92,11 +92,6 @@ Meteor.methods({
     seedShop._id = Random.id();
     seedShop.name = seedShop.name + count;
 
-    // We trust the owner's shop clone, check only when shopData is passed as an argument
```
pmn4: (this was redundant)

aldeed:
@zenweasel @pmn4 We will have to discuss how schema extension/override should be done going forward after the upgrade to latest SimpleSchema. It might be best to recommend something like `Schemas.Shop.extend(ExternalShop)` instead of calling `attachSchema`, in which case both `Schemas.Shop` and `Shops.simpleSchema()` would return the mutated schema.

However, if anybody were trying to attach multiple schemas by selector, then they would have to do `attachSchema`. So, even if the recommendation is to do `Schemas.Shop.extend(ExternalShop)` whenever possible, it would still be safest to retrieve the schemas using `.simpleSchema()`.

So I think this should be merged, except it would be best to change the check to `check(shopData, Match.Optional(Collections.Shops.simpleSchema(shopData)));` (pass shopData to `.simpleSchema()`) because then it will be sure to select the correct schema if someone has attached multiple.

(I can make this same change in my PR if it gets merged.)

aldeed:
The changes from this PR are now included in https://github.com/reactioncommerce/reaction/pull/3331. Closing this one.

PR #3463, opened 2018-01-10
User prefs stored in Accounts

### LINKED ISSUE #3404: Action view - Grid Settings reactivity error
Opening the `Grid Settings` panel from the product grid and several performance issues appear:
These are experience items, hard to capture in screenshots, video, so you'll need to reproduce:

**Actual:**

1 - select product
2- Panel opens VERY sluggish
3- clicking alternate layouts provides no response (updates in console occur)
4 - grid isn't updated to new layout (but refreshing the page will show the correct it.

**Expected**

Reactive changes to grid should happen when layouts are selected.
Grid settings should be responsive when opening, selecting additional products.

```
Node: 6.11.1
NPM: 5.6.0
Meteor Node: 8.8.1
Meteor NPM: 5.4.2
Reaction CLI: 0.24.2
Reaction: 1.6.1
Reaction branch: master
Docker: 17.09.0-ce
```

Reproduced in #3368 as well.

Akarshit:
### Closes #3404 

The issue is happening because of the following reason
1. Whenever some sub-menu is opened(or any action is taken), `Meteror.user().profile.preferences` is updated.
2. This causes all the `Tracker.autorun()` that depend on Meteor.user() to run again. This takes huge amount of time and is the main reason for the UI to be lagging.
An example would be the i18n startup script(https://github.com/reactioncommerce/reaction/blob/master/client/modules/i18n/startup.js#L41), which uses Meteor.user() and is triggered every time some thing is updated in `Meteor.user()`

To fix this, the user preference will be now be stored in`Accounts` collection.

### To test

1. Run `reaction`
2. Log in as admin, click on a product, notice the grid panel opens almost instantaneously.
3. Open social settings from Action menu, open/close some sub-menu. Notice there is no lag in opening of the menus now.
4. Take a note of the sub-menus, opened and closed.
5. Log out and Log in as admin
6. Reopen the social settings and only the same sub-menu should be open that were left opened before logout.
7. Change the shop from marketplace and then logout.
8. Browse to RC in incognito mode, login as admin.
9. The same shop should open that was opened on last login.

I have tried my best to check all the places where `preferences` are being used. If anyone can think of anything else, please let me know.


client/modules/core/main.js
```diff
@@ -387,10 +387,9 @@ export default {
   },
 
   getUserPreferences(packageName, preference, defaultValue) {
-    const user = Meteor.user();
-
+    const user = Accounts.findOne(Meteor.userId());
```
pmn4: consider adding a comment so that this change is not reversed in the future

pmn4:
nice find @Akarshit! lgtm

client/modules/core/main.js
Akarshit: @pmn4 Thanks, I have added a comment where the user preferences are being saved.

Akarshit:
Will do something more to solve this completely.

Akarshit:
Ready for review.

client/modules/core/main.js
```diff
@@ -9,14 +9,15 @@ import { Roles } from "meteor/alanning:roles";
 import Logger from "/client/modules/logger";
 import { Countries } from "/client/collections";
 import { localeDep } from  "/client/modules/i18n";
-import { Packages, Shops } from "/lib/collections";
+import { Packages, Shops, Accounts } from "/lib/collections";
 import { Router } from "/client/modules/router";
+import store from "store";
```
zenweasel: import order

package-lock.json
```diff
@@ -16910,6 +16910,11 @@
       "resolved": "https://registry.npmjs.org/stealthy-require/-/stealthy-require-1.1.1.tgz",
       "integrity": "sha1-NbCYdbT/SfJqd35QmzCQoyJr8ks="
     },
+    "store": {
```
zenweasel: I think we need to remove the entry for "amplify-store"

client/modules/core/main.js
```diff
@@ -387,25 +388,35 @@ export default {
...
-
-    if (user) {
-      const profile = Meteor.user().profile;
-      if (profile && profile.preferences && profile.preferences[packageName] && profile.preferences[packageName][preference]) {
-        return profile.preferences[packageName][preference];
+    getDep(`${packageName}.${preference}`).depend();
+    if (Meteor.user()) {
+      const packageSettings = store.get(packageName);
+      // packageSettings[preference] should not be undefined or null.
+      if (packageSettings && packageSettings[preference] != undefined) { // eslint-disable-line eqeqeq
```
zenweasel: Let's not do an equality comparison on `undefined`

client/modules/core/main.js
```diff
@@ -387,25 +388,35 @@ export default {
...
 
   getUserPreferences(packageName, preference, defaultValue) {
-    const user = Meteor.user();
-
-    if (user) {
-      const profile = Meteor.user().profile;
-      if (profile && profile.preferences && profile.preferences[packageName] && profile.preferences[packageName][preference]) {
-        return profile.preferences[packageName][preference];
+    getDep(`${packageName}.${preference}`).depend();
+    if (Meteor.user()) {
```
prinzdezibel: - Is `Meteor.user()` still necessary?

Akarshit: @prinzdezibel It's just an optimisation to get settings only when the user is actually logged in.
As we have removed all the dependencies on by not storing anything in `Meteor.user().profile.preferences` this shouldn't cause components to re-render.
Why did you think it would be a good idea to remove it? I might be missing something.

Akarshit: @prinzdezibel It's just an optimisation to get settings only when the user is actually logged in.
As we have removed all the dependencies on by not storing anything in `Meteor.user().profile.preferences` this shouldn't cause components to re-render.
Why did you think it would be a good idea to remove it? I might be missing something.

prinzdezibel:
@Akarshit  Good start! I think it's a great thing to tackle performance. But somehow I have the feeling that this PR can be improved. I have to admit that I can't present you a better solution out of my head, just some ideas/thoughts on why I think there could exist a better approach to our problem.

- Redundancy: Having settings from namespace `reaction` persistent in a redundant manner (local storage AND Account collection), seems a bit unnecessary. Also it requires you to think about syncing between both data sources.
- Consistency: Having settings from namespace `reaction` persistent in Accounts collection & local storage, but settings from other namespaces only in local storage is a bit confusing for the developer.
- I believe that purging local storage would not sync the settings from Accounts selection into local storage without a page reload. Which may or may not be a problem.
- We would need to think about migration of existing settings. This would require a database migration from Users collection to Accounts collection
- Meteor.users.profile is somewhat special, because it's added by default when a new user registers and is automatically writable from the client and it's published automatically (which is not necessarily a good thing, but I'm not sure if this fact has implications when moving to Accounts.)

Akarshit:
@prinzdezibel 
> Redundancy, Consistency

We actually need the `shopId` in the Accounts collection, because serves uses that to process requests [here](https://github.com/reactioncommerce/reaction/blob/master/server/api/core/core.js#L380) so the `reaction` package settings have to synced to a collection(`Accounts` in this case).
Splitting preferences between Accounts and localStorage gives a little performance boost.

> I believe that purging local storage would not sync the settings from Accounts selection into local storage without a page reload
Yes, it won't.

> We would need to think about migration of existing settings.
These settings(which menu's are open etc.) are more of a "good to have" thing, so even if we miss these on a particular login I don't think anyone is going to miss these.
But I might be missing something here. Is there any case, absence of these settings would cause problem?

> Meteor.users.profile is somewhat special, because it's added by default when a new user registers and is automatically writable from the client and it's published automatically

We have decided not to store the `preferences` in database. So we don't have to worry about the publish part. Also one more thing that `Meteor.user()` provides is reactivity, which we are implementing by ourselves over `localStorage`.

But yes I get your point, let's wait for suggestion from more people. 
One possible solution can be to not use localStorage at all, just use `Accounts` collection implement controlled reactivity(which is already done) over `Accounts` collection. The only downside would be the DB call on every change in the UI.

zenweasel:
So if I open a panel (e.g. Social Settings) and logout and back in again, I am not getting that panel reopened again. I know there was some discussion about whether this was a must have but it doesn't seem to be worked and is noted in the test instructions.

zenweasel:
@Akarshit Could there be any security implications of storing the Account data in `localstorage`?

Akarshit:
@zenweasel When I open a panel (e.g. Social Settings) and logout and back in again, I am **getting** that panel reopened again.
This won't work if the panel is opened in a incognito window.
> Could there be any security implications of storing the Account data in localstorage

We are only storing the stuff inside the `profile.preferences` in the localStorage. I couldn't find any place where we were leaking something important.

client/modules/core/main.js
```diff
@@ -387,25 +388,35 @@ export default {
...
+    // to prevent all autorun() with dependency on Meteor.user() to run again.
     if (Meteor.user()) {
-      return Meteor.users.update(Meteor.userId(), {
-        $set: {
-          [`profile.preferences.${packageName}.${preference}`]: value
-        }
-      });
+      // "reaction" package settings should be synced to
+      // the Accounts collection.
+      if (packageName in ["reaction"]) {
```
zenweasel: So if we only sync Reaction package, how will plugin-creators packages differ? Will it not be reactive when their settings change or will it not save the state of their panel? What's the thinking behind segregating them like this?

client/modules/core/main.js
```diff
@@ -842,3 +853,10 @@ function createCountryCollection(countries) {
   }
   return countryOptions;
 }
+
+function getDep(key) {
```
zenweasel: JSDoc

client/modules/core/startup.js
```diff
@@ -19,6 +20,24 @@ Meteor.startup(() => {
   return Tracker.autorun(() => {
     const userId = Meteor.userId();
 
+    // Load data from Accounts collection into the localStorage
+    Tracker.nonreactive(() => {
```
zenweasel: Using `Tracker.nonreactive` probably always warrants an explanation of why

client/modules/core/startup.js
```diff
@@ -19,6 +20,24 @@ Meteor.startup(() => {
...
+          Object.keys(user.profile.preferences).forEach((packageName) => {
+            const packageSettings = user.profile.preferences[packageName];
+            Object.keys(packageSettings).forEach((preference) => {
+              Reaction.setUserPreferences(packageName, preference, packageSettings[preference]);
+            });
+          });
+        }
+      } catch (err) {
+        Logger.error("Problem in loading user preferences from Accounts collection");
+        Logger.error(err);
```
zenweasel: Dumping an un-sanitized error out client-side is probably a bad practice

client/modules/core/main.js
Akarshit: > So if we only sync Reaction package, how will plugin-creators packages differ

I didn't understand this, maybe because I am not very clear how the plugin-creators packages work.
The reason for segregation here is that the `activeShopId` is stored in the `reaction` package, this is also used on the server side([here](https://github.com/reactioncommerce/reaction/blob/master/server/api/core/core.js#L379)).
So only if the `reaction` package is synced to backend, `activeShopId` would be the correct one.

spencern:
Just tested this myself and seems to be _way_ faster. Nicely done.

pmn4:
🎉
very excited for this one! for any marketplace, we need to be sure our shop owners stay impressed!

Akarshit:
Thanks :)

PR #3465, opened 2018-01-10
ProductsContainer HOC should pass unused properties to child component.

### LINKED ISSUE #3464: Refactoring proposal: HOC should pass unused properties to child component.
In current ProductsContainer implementation (product-variant/productsContainer.js), the HOC function `wrapComponent` needs to know exactly which properties it's supposed to pass along to its child component.

This is unfortunate, because it means that this HOC needs to be overridden if one decides to pass a new, unknown property to the child component. This leads to quite a bit boilerplate code, as it is shown [here](https://github.com/reactioncommerce/reaction-swag-shop/blob/c6f01913ae055fa8125e5d5499781aba8ba7499b/client/containers/product-variant/productsContainer.js#L19) with the `tags` property.

It would be nice if the wrapComponent HOC would be agnostic about the properties of its children.

prinzdezibel:
Resolves #3464 

Allows the HOC to be agnostic about the properties of the child
component.


prinzdezibel:
@mikemurray Hi Mike, could you have a look into this? It's somewhat important for the swag shop. thx.

jshimko:
Couldn't this...

```js
      return (
        <Comp
          ready={this.ready}
          productsSubscription={this.props.productsSubscription}
          loadMoreProducts={this.loadMoreProducts}
          loadProducts={this.loadProducts}
          showNotFound={this.props.showNotFound}
          {...passedProps}
        />
      );
```
just be this?...
```js
      return (
        <Comp
          ready={this.ready}
          loadMoreProducts={this.loadMoreProducts}
          loadProducts={this.loadProducts}
          {...this.props}
        />
      );
```

prinzdezibel:
@jshimko Yes, that would work too. The question is also, if we wouldn't care passing superfluous properties like `canLoadMoreProducts`, which is only used in wrapper component and doesn't **(shouldn't ?)** actually need to be passed to child.

prinzdezibel:
@jshimko, @spencern Can we get this decided?  I would need to know, because I'd like to get the exact same thing for ProductGridContainer approved (no PR yet).

jshimko:
Not really my call, but my personal preference is usually to just pass everything through a HOC with `{...this.props}` because it's simple and doesn't break anything.  Since most users are usually importing/using our HOC's directly, I think it'd be reasonable for them to assume anything passed to the wrapper component would also be available in their wrapped component (whether used there or not).  But I also realize we haven't standardized that at this point and that's not always a safe assumption.  That _is_ how `composeWithTracker` works though.  See [the render method](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/components/lib/composer/compose.js#L95-L117) of the wrapper that fetches data.  Fetched data is merged with any other upstream props that were passed into the wrapper and then they're all passed down to the child component.  Also note that it uses [`hoist-non-react-statics`](https://github.com/mridgway/hoist-non-react-statics) to pull any non-React statics up into the parent from the child component (see [why you should care about that](https://reactjs.org/docs/higher-order-components.html#static-methods-must-be-copied-over) in the React docs).

I guess more importantly, I find the props filtering solution a little hard to quickly read through.  I know it works fine and I do get what's happening there, but I don't know if I'd ever want to do that for every HOC that needs to filter passed props.  If you really want to keep the filtering and improve the readability a little, you could do a simple one-liner with Lodash that accomplishes the same thing.  

So this...

```js
const ownProps = Object.keys(ProductsContainer.propTypes);
const passedProps = Object.keys(this.props).reduce((obj, key) => (
  ownProps.includes(key) ?  { ...obj } : { ...obj, [key]: this.props[key] }
), {});
```

Could be shortened to this...

```js
const passedProps = _.omit(this.props, Object.keys(this.propTypes));
```
(which is generic enough to be used in _any_ HOC)

But that's just me.  I don't think any of the discussed options are _wrong_.  I just tend to prefer syntax simplicity and readability.

prinzdezibel:
Agree fully. If compose() does pass all props, our own wrappers should do this too for consistency reasons.

Also this documentation recommends this as best practice (first scentence):
https://reactjs.org/docs/higher-order-components.html#refs-arent-passed-through

PR #3468, opened 2018-01-10
Fix #3371 Test Shopify credentials before saving.

### LINKED ISSUE #3371: Incorrect Shopify credentials displays a success message
### Expected behavior
If incorrect information is entered into the Shopify connector, I should receive an error message that tells me my credentials do not match or is incorrect. 

### Actual behavior
Currently if I enter in incorrect information into the Shopify connector, I receive a success message. 
Note: the success message appears to be saying that I have successfully entered and saved my Shopify credential, not that I've entered a valid Shopify credentials. 

1. Create Reaction shop
2. Go to Connectors
3. Enter in Shopify API information, purposely put in incorrect information in any of the API fields
4. See success message upon hitting save changes.

Akarshit:
### To test
1. Reset database to load new transaltions `reaction reset -n`
1. Run `reaction`
2. Open the Connector from Action menu.
3. Enter invalid API credentials and hit the `Save` button. You should get a message that setting was *not* saved.
4. Enter valid API credentials and hit the `Save` button. You should get a message that setting was saved.

### Review
Should the message on invalid credetials be something like "Invalid Credentials" or just say "Setting not saved"(when behind the scene they actually have been saved)?


zenweasel:
@Akarshit I think you should get a regular validation error on the field like you would if you entered any other invalid value.

Also, this newline problem is popping up again. Please try to review your IDE configuration and see if you can figure out what's going on. This seems to be only happening to you.

zenweasel:
@Akarshit It can be that "credentials were invalid" rather than "settings not saved". I  think that "settings not saved" would be confusing because you aren't saying why. But true that you can't show a field-specific error.

Akarshit:
@zenweasel Updated the PR and test instructions.

imports/plugins/included/connectors-shopify/client/settings/shopify.js
```diff
@@ -143,7 +143,12 @@ Template.shopifySync.events({
 AutoForm.hooks({
   "shopify-connect-update-form": {
     onSuccess: function () {
-      return Alerts.toast(i18next.t("admin.settings.saveSuccess"), "success");
```
pmn4: my only thought here is that the return values are different now (previously: a Promise from Alerts; currently: `undefined`). does anything depend on this returning a promise?

pmn4:
lgtm

imports/plugins/included/connectors-shopify/client/settings/shopify.js
Akarshit: @pmn4 I don't think so because even before, that Promise from Alert wasn't being used anywhere. But let's also wait for smarter people(than me) to see this.

Akarshit:
There is no new-line error now, it's just something with *Codacy* that it's showing the error.

imports/plugins/included/connectors-shopify/client/settings/shopify.js
zenweasel: @Akarshit I don't think @pmn4 means the promise from Alerts but the promise from the Stripe call.

zenweasel:
If I change shared secret to an invalid value I am still getting the "Successfully saved" message

zenweasel:
If I hit "Save Changes" when a field is blank I am getting the following error on the server
```
Exception while invoking method 'connectors/shopify/api/credentials/test' Error: Missing or invalid options
    at new Shopify (/Users/brenthoover/Projects/js/reaction/node_modules/shopify-api-node/index.js:36:11)
    at Promise.asyncApply (imports/plugins/included/connectors-shopify/server/methods/api/credentials.js:28:21)
    at /Users/brenthoover/.meteor/packages/promise/.0.10.0.ak8m4i++os+web.browser+web.cordova/npm/node_modules/meteor-promise/fiber_pool.js:43:40
```
I still do get the error toast on the frontend, but there should be no error thrown on the backend.

Akarshit:
@zenweasel Updated.

zenweasel:
@Akarshit Conflicts that need to be resolved here

spencern:
We can't merge with codacy errors. Need to figure out what it's complaining about and resolve.

Akarshit:
@zenweasel Resolved conflicts.
@spencern resolved the errors(just merged with the `master`).

zenweasel:
@Akarshit Maybe the success message could be "Settings validated and saved"?

zenweasel:
So If I save the form with no slug I get "Invalid API key/password". Can we separate messages about verifying the API key with just the form being valid. I feel like people are going to get confused.

Akarshit:
@zenweasel We can do a "Invalid credentials" when the API key, password, shop-slug combination is not correct.

Akarshit:
@zenweasel Had a discussion about this with @aldeed, he said that the blackboxes stuff would be validated in SimpleSchema 2.So when his PR is merges the form validation should start working.

PR #3479, opened 2018-01-12
Fix #3478 register variantList

### LINKED ISSUE #3478: Bug VariantList not registered.
### Expected behavior
[VariantList](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/included/product-detail-simple/client/containers/variantList.js) should be registed via registerComponent() so that it could be replaced by users.
### Actual behavior
Since the component is not registered, replacing it with custom component throws error `Component 'VariantList' not found. Use registerComponent to create it.`

Akarshit:
###Fix for #3478 

### To test
1. Run `reaction` 
1. Open the PDP of example-product, everything should be fine in the variant section.
1. Try to add custom code by extending the `VariantList` Component, and then replacing the native component with the `<CustomVariantListComponent>` using `replaceComponent`.
1. There should be no error and the `<CustomVariantListComponent>` should be rendered.


pmn4:
lgtm

PR #3493, opened 2018-01-16
Fix 3473 fix typo and logic

Akarshit:
### To test
1. Run `reaction`
1. Enable Avalara's address validation.
2. Add the `Example product` to the cart and go to checkout step.
3. For the checkout address enter the address as shown below in the screenshotscreen shot 2018-01-16 at 3 29 07 pm
4. On `Save change` the address will be validated and saved.


pmn4:
lgtm

PR #3504, opened 2018-01-18
fix #3474 making validatedAddress object to prevent exception

### LINKED ISSUE #3474: Address validation in plugins bound to fail due to a logic issue in code.
https://github.com/reactioncommerce/reaction/blob/master/server/methods/accounts/accounts.js#L285
 
We call the plugin's address validation method (as acquired further up in the code). We check the resulting address to see if it is a *truthy* variable. If it isn't, we then set the validation to failed - but then we try setting the `validatedAddress.failedValidation` prop to false (which, why are we setting it to false?). Unfortunately the `validatedAddress` variable is *not* an object, as any object - even an empty one, would be truthy. So it is impossible for `validatedAddress` to ever get to that point as an object, which means it should always fail at:
https://github.com/reactioncommerce/reaction/blob/master/server/methods/accounts/accounts.js#L289

The taxes-avalara default plugin also should fail because of this. I have been unable to test it, however examining the code and mocking it in my tests reveals that it should fail as well. 
https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/included/taxes-avalara/server/methods/taxCalc.js#L271
We set the `validatedAddress` to an empty string. The reasoning is so that it is a falsy value. If you examine the code you'll find that if no addresses are returned, the `validatedAddress` variable will be returned falsy (which will fail as mention in the above paragraph).

I haven't examined the code fully, but I would propose a solution as setting validatedAddress to an empty object.

```javascript
    } else {
      // No address, fail validation
      validated = false;
      validatedAddress.failedValidation = false;
    }
```

becomes

```javascript
    } else {
      // No address, fail validation
      validated = false;
      validatedAddress = {
        failedValidation: false
      };
    }
```

Akarshit:
### To test
1. Run `reaction`
2. Buy `Example Product` and go to the checkout sceen.
3. Enter a invalid **US** address and click `Save and Continue`.
4. There should be no exception in the console.

### DIFF
--- a/server/methods/accounts/accounts.js
+++ b/server/methods/accounts/accounts.js
@@ -286,13 +286,13 @@ export function validateAddress(address) {
       validationErrors = compareAddress(address, validatedAddress);
       if (validationErrors.totalErrors || formErrors.length) {
         validated = false;
-        validatedAddress.failedValidation = false;
+        validatedAddress.failedValidation = true;
       }
     } else {
       // No address, fail validation
       validated = false;
       validatedAddress = {
-        failedValidation: false
+        failedValidation: true
       };
     }
   }
--- a/server/methods/accounts/accounts.js
+++ b/server/methods/accounts/accounts.js
@@ -291,7 +291,9 @@ export function validateAddress(address) {
     } else {
       // No address, fail validation
       validated = false;
-      validatedAddress.failedValidation = false;
+      validatedAddress = {
+        failedValidation: false
+      };
     }
   }
   const validationResults = { validated, fieldErrors: validationErrors, formErrors, validatedAddress };


server/methods/accounts/accounts.js
```diff
@@ -291,7 +291,9 @@ export function validateAddress(address) {
     } else {
       // No address, fail validation
       validated = false;
-      validatedAddress.failedValidation = false;
+      validatedAddress = {
+        failedValidation: false
+      };
     }
   }
   const validationResults = { validated, fieldErrors: validationErrors, formErrors, validatedAddress };
```
pmn4: previously, this method returned the `address` passed to it, but with your change, it could be replaced with `{ failedValidation: false }`. This could be fine (I have no idea what the downstream code looks like), but feels like a potentially breaking change.
To be safe, consider checking that `validatedAddress` is null/undefined,
``` JavaScript
// line 294:
if (!validatedAddress) {
  validatedAddress = {};
}
validatedAddress.failedValidation = false; // same as it was
```

zenweasel:
I know this isn't exactly in the scope of this bug, but it would seem to me like both places where `failedValidation` is set there it should be set to `true`. Probably some mixup when looking at `validated` and `failedValidation` which probably could have better named so as to not have similar but opposite names.

server/methods/accounts/accounts.js
pmn4: oops! my bad. I thought this was the `else` for `if (validator)`

PR #3506, opened 2018-01-18
Remove all local packages

aldeed:
This speeds up reloads during development. I diffed all the local packages, made all changes in the official packages, republished those, and bumped versions here. Here are my testing results:

screenshot 2018-01-18 11 24 42

*(average of 2 restarts each)*

A lot of things can affect reload times, so I would not get too excited about the percentages, but I think it is safe to say that most reloads for most developers will now be at least 5 seconds faster each time they change files and save.

Here is my CollectionFS commit: https://github.com/CollectionFS/Meteor-CollectionFS/commit/8db94a68c686f425796db39be456189a80d41ffc

The only file changed is the versions file. The rest of the diff is from deleting the whole packages folder.

This can be tested by running through any test plans involving making sure images are properly uploaded and displayed. It would affect only CollectionFS things, and should have no effect since the code should be the same if I did it right.

NOTE: We are planning to replace CFS eventually, but this will help all developers work faster in the short term.


zenweasel:
@aldeed Normally you should assign a reviewer (just one and tag anybody else)

aldeed:
FYI @aaronjudd

Issue #3507, opened 2018-01-18
Memoize `getShopId` to Reduce Queries


pmn4:
### Expected behavior
Once the "current" shop id is determined, return a cached version to avoid querying the DB again.

### Actual behavior
On startup, we call `getShopId` more than 900 times
Per request, it's about 30.
(this method makes one or two calls to the database, depending on whether you are logged in and have an "active shop" selected).

### Steps to reproduce the behavior
place a `console.log` statement [here](https://github.com/pmn4/reaction/blob/master/server/api/core/core.js#L313) and [here](https://github.com/pmn4/reaction/blob/master/server/api/core/core.js#L322) to count the number of times we go to the database to get the current shop.

### Versions
master

spencern:
@zenweasel are you or @jshimko going to review this one?

Akarshit:
This might be fixed in #3463 [here](https://github.com/reactioncommerce/reaction/pull/3463/files#diff-e0509424f4672c76deaab12707ba478dR465).
Basically the setShopId is not done when the newId is same as the old one, hence all the multiple calls to `getShopId`(because of reactiveness) are prevented.
But not 100% sure.

zenweasel:
My mistake. Assigning to @jshimko to review

PR #3510, opened 2018-01-18
[#3507] Memoize/Cache getShopId to Reduce DB Load

pmn4:
`getShopId` makes one or two queries to the database. It is called >900 times at startup and ~15 times per request. Caching the result of `getShopId` would reduce these queries to 1 for startup, and 1 per request.

---
is the breaking change ok to keep?
(specifically, removing the `userId` from `getShopId(userId)`... I replaced it with `getUserShopId(userId)` and updated the 3/4 places it was used in this repo, however any third-party plugins would have to do the same)


server/api/core/core.js
```diff
@@ -364,29 +363,93 @@ export default {
     return null;
   },
 
+  /**
+   * @name connectionDataStore
+   * @summary a construct to store per-connection data. Once a connection is
+   * closed, the data is lost. This is server-side companion to Sessions in the
+   * client. connectionDataStore has two methods: get and set
+   * @memberof Core
+   */
```
pmn4: before finalizing, this code should be moved to a separate file.

server/api/core/core.js
```diff
@@ -364,29 +363,93 @@ export default {
...
+  connectionDataStore: (function () {
+    const inMemoryCache = {};
+    const connectionKey = "connection-data";
+
+    function connectionData() {
```
pmn4: implementation reflects that of [`Meteor.userId()`](https://github.com/meteor/meteor/blob/f551a56f93246b0a8251426d8f8eb9964cbdd865/packages/accounts-base/accounts_server.js#L69-L80), with a fallback for using it outside of a connection (i.e., on startup)

server/api/core/core.js
```diff
@@ -364,29 +363,93 @@ export default {
...
-   * @todo This should intelligently find the correct default shop Probably whatever the main shop is or marketplace
-   * @param  {String} userId User ID String
+   * @summary Get shop ID, first by checking the current user's preferences
+   * then by getting the shop by the current domain.
+   * @todo should we return the Primary Shop if none found?
    * @return {StringId}        active shop ID
    */
-  getShopId(userId) {
-    check(userId, Match.Maybe(String));
+  getShopId() {
```
pmn4: ^ this is the breaking change.
In order to simplify things a bit, I removed the `userId` parameter/feature. There were only a handful of uses of that feature, which have been updated in this PR. Of course, I could not account for all of the plugins that exist in the world, so your call on whether to keep this change.

server/api/core/core.js
```diff
@@ -364,29 +363,93 @@ export default {
...
+
+    if (shopId) {
+console.log(`debugging: no database. skipped ${this.debuggingSavedQueryCount++} queries`);
+      return shopId;
+    }
+
     const activeUserId = Meteor.call("reaction/getUserId");
-    if (activeUserId || userId) {
-      const activeShopId = this.getUserPreferences({
-        userId: activeUserId || userId,
```
pmn4: (no longer an issue, but this was a bug, right? if you pass `userId` as a parameter, it should be used, with `activeUserId` as the fallback... right??)

server/api/core/core.js
```diff
@@ -364,29 +363,84 @@ export default {
...
+  getShopId() {
+    // is there a stored value?
+    let shopId = this.connectionDataStore.get("shopId");
+
+    if (shopId) {
+console.log(`debugging: no database. skipped ${this.debuggingSavedQueryCount++} queries`);
+      return shopId;
+    }
+
     const activeUserId = Meteor.call("reaction/getUserId");
```
pmn4: _(is it frowned upon to call `Meteor.userId()`?)_

server/api/core/core.js
```diff
@@ -396,9 +450,30 @@ export default {
...
 
+  /**
+   * @name getUserShopId
+   * @method
+   * @memberof Core
+   * @summary Get a user's shop ID, as stored in preferences
+   * @todo This should intelligently find the correct default shop Probably whatever the main shop is or marketplace
+   * @return {StringId}        active shop ID
+   */
+  getUserShopId(userId) {
```
pmn4: the `userId` feature of `getShopId` has been moved here to it's own method.

server/api/core/core.js
```diff
@@ -364,29 +363,84 @@ export default {
...
+console.log(`debugging: database query #${this.debuggingQueryCount++}`);
     }
 
+    // store the value for faster responses
+    this.connectionDataStore.set("shopId", shopId);
+
+    return shopId;
+  },
+
+  getShopIdByDomain() {
```
pmn4: This was pulled out of `getShopId` to clean things up. No functional reason for doing so

pmn4:
If this PR/concept it approved, I plan to apply the same logic to all `getShop*` methods (`getShopName`, `getShopPrefix`, etc.)

client/modules/core/main.js
```diff
@@ -483,10 +483,12 @@ export default {
...
-    if (id) {
-      this.shopId = id;
-      this.setUserPreferences("reaction", "activeShopId", id);
-    }
+    if (!id) { return; }
+
+    this.shopId = id;
+    this.setUserPreferences("reaction", "activeShopId", id);
+
+    Meteor.call("shop/resetShopId");
```
pmn4: to maintain consistency between client and server, I added the `"shop/resetShopId"` method to effectively clear cache, and re-determine the shopId. I chose not to do a _set_ shopId method to avoid having to check whether the user is permitted to do so (and presumably duplicating logic which already exists).

server/api/core/ConnectionDataStore.js
```diff
@@ -0,0 +1,42 @@
...
+  if (!hasDataKey) {
+    connection[connectionKey] = {};
+  }
+
+  return connection[connectionKey];
+}
+
+const ConnectionDataStore = {
+  get(key) {
+    return connectionData()[key];
```
pmn4: we want to call `connectionData()` each time we try to access data since `ConnectionDataStore` is a singleton, and we want to be sure we are using the _current_ connection at all times.

pmn4:
_(wow! I broke a lot of tests!)_

server/api/core/ConnectionDataStore.js
```diff
@@ -0,0 +1,44 @@
...
+  if (!hasDataKey) {
+    connection[connectionKey] = {};
+  }
+
+  return connection[connectionKey];
+}
+
+export default {
+  get(key) {
+    return connectionData()[key];
```
pmn4: we want to call `connectionData()` each time we try to access data since `ConnectionDataStore` is a singleton, and we want to be sure we are using the _current_ connection at all times.

zenweasel:
@pmn4 Sorry for neglecting these for a while. Asking @jshimko to take a look at this (and your others)

aaronjudd:
@reactioncommerce/community can we move this forward? Can we shepherd this into 1.9.0?

zenweasel:
@pmn4 Just a couple of little code changes and resolve the conflicts and I think we can move this forward. Really appreciate your patience with these

zenweasel:
@jshimko I looked at these code conflicts and I think we should be able resolve these as most are just the upstream linting changes. In the interest of getting this into `release-1.9` can you resolve and make any other changes necessary? I can help out reviewing/testing when you're done if you need.

pmn4:
@zenweasel I'll be at the computer tomorrow morning early (EST) and will resolve if you haven't beaten me to it

zenweasel:
@jshimko Can you take another look at this?

zenweasel:
@jshimko @pmn4 Can we get a doc ticket for this new API. I agree with Jeremy that this is probably something that can get re-used

PR #3512, opened 2018-01-18
ProductGridContaintainer should pass all properties to child component

### LINKED ISSUE #3511: HOC returning ProductGridContainer should pass all properties to child component
For HOCs it's best practice to pass all properties down to the child component.

See: https://reactjs.org/docs/higher-order-components.html#refs-arent-passed-through
(first sentence)

Also I face this need, because otherwise it's not possible to reuse the HOC without copying a lot of boilerplate code.

prinzdezibel:
Closes #3511.

Per convention, all properties should be passed to child component:
https://reactjs.org/docs/higher-order-components.html#refs-arent-passed-through

This is enables us to use the HOC child component agnostic.

### DIFF
--- a/imports/plugins/included/product-variant/containers/productGridContainer.js
+++ b/imports/plugins/included/product-variant/containers/productGridContainer.js
@@ -13,12 +13,10 @@ import ProductGrid from "../components/productGrid";
 const wrapComponent = (Comp) => (
   class ProductGridContainer extends Component {
     static propTypes = {
-      canEdit: PropTypes.bool,
       isSearch: PropTypes.bool,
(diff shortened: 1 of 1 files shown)


prinzdezibel:
@spencer @jshimko  This is the second HOC that needs to pass the properties. Would be great to get this reviewed in a timely manner. Thanks!

spencer:
I don't think you're meaning to '@' me on this. #justsayin ;-)

prinzdezibel:
@spencer: Ooops, sorry :-)  
Was missing an 'n' as in @spencern.

PR #3513, opened 2018-01-18
(fix): collection hooks caused cancelOrder to fail

spencern:
This fixes an unreported issue where canceling an order and choosing not to restock would sometimes, with custom payment providers, cause the `orderCancel` method to fail to run all the way through. This failure is connected to the way the `collectionHooks` run, as using `Products.direct.update` works as does moving the Products.update loop out of it's own function.

I've chosen to move the Products.update call inline in the orderCancel function as I don't want to break all of the `collectionHooks` that are being called within the cancellation method.

### DIFF
--- a/server/imports/fixtures/orders.js
+++ b/server/imports/fixtures/orders.js
@@ -68,6 +68,7 @@ export function getShopId() {
  */
 
 export default function () {
+  const shopId = getShopId();
   Factory.define("order", Orders, {
     // Schemas.OrderItems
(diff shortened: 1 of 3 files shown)


spencern:
There's an annoying issue here where adding the permission check and verifying the shopId of the product is causing one of our tests to fail because the Factory isn't using the same shopId between the Order items and the products it's creating. Debugging but then should be ready to go.

spencern:
@mikemurray added you as the reviewer here in place of @impactmass

PR #3520, opened 2018-01-19
Move back to current Authorize.net version and primary source of library

zenweasel:
We were using our own fork of this library because it had some vulnerabilities that have now been fixed and to resolve another PR https://github.com/ongoworks/node-authorize-net/pull/2 that's also been fixed upstream. 

Testing this I ran into an issue when entering invalid cards but it turns out that's happening on `master` so created a separate issue for it

### To Test
1. `reaction reset -y` (libs have changed)
1. Login and enable Authorize.net
1. Check, Approve and Capture (Authorize does not support refund)

Once this is approved we can probably eliminate the `ongoworks` version until we move to the "official" lib


pmn4:
is it expected that `package-lock.json` would change as much as it did?

zenweasel:
Changing to WIP to investigate the vulnerabilities that Snyk pointed out

zenweasel:
My PR  to patch the vulnerability was accepted upstream so this is now clean

zenweasel:
@pmn4 I think the `package-lock.json` changed so much because that library has a ton of dependencies so the whole tree got pulled out and then a nearly identical tree inserted to replace this.

PR #3524, opened 2018-01-19
(breaking, feat) CoreLayout should probe for React component as fallback

### LINKED ISSUE #3523: Not possible to overwrite layouts and/or INDEX_OPTIONS without being forced to use Blaze
Issue:
coreLayout.js renders the content of &lt;main&gt; through Blaze in an unconditional manner.

This forces users who wish to write customized views to write deprecated Blaze templates. That's not good.

An intermediary solution would be to render a Blaze , if a Blaze template with the passed name is available. If not, coreLayout could probe for a React component and use that as fallback.

That way the old behaviour is guaranteed, even if there's a Blaze template and a React component with the same name (which should not happen anyway, because Blaze templates should be camelcased and React component name start with upper letter).

prinzdezibel:
Closes #3523

This PR enables users to overwrite layouts and/or INDEX_OPTIONS without
being forced to use deprecated Blaze templates.

Through rendering Blaze templates in a precendential manner, the old behaviour
should be guaranteed, even if there's a React component with the same
name.

# How to test
Put this somewhere in code
```
const myMain = () => (
  <h1>My custom main</h1>
);
registerComponent("products", myMain);
```

BREAKING CHANGE: This will break the user's plugins if he has named a React component in the same way we named Blaze templates in core.

### DIFF
--- a/imports/plugins/core/layout/client/components/coreLayout.js
+++ b/imports/plugins/core/layout/client/components/coreLayout.js
@@ -17,14 +17,13 @@ const CoreLayout = ({ actionViewIsOpen, structure }) => {
   const footerComponent = layoutFooter && getComponent(layoutFooter);
 
   let mainNode = null;
-  if (Template[template]) {
-    mainNode = <Blaze template={template} />;
-  } else {
-    try {
-      const mainComponent = getComponent(template);
-      mainNode = React.createElement(mainComponent, {});
-    } catch (error) {
-      // no-op
+  try {
+    const mainComponent = getComponent(template);
+    mainNode = React.createElement(mainComponent, {});
+  } catch (error) {
+    //  Probe for Blaze template (legacy)
+    if (Template[template]) {
+      mainNode = <Blaze template={template} />;
     }
   }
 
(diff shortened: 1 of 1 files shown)


imports/plugins/core/layout/client/components/coreLayout.js
```diff
@@ -16,17 +16,28 @@ const CoreLayout = ({ actionViewIsOpen, structure }) => {
...
 
+  let mainNode = null;
+  if (Template[template]) {
+    mainNode = <Blaze template={template} />;
+  } else {
+    try {
+      const mainComponent = getComponent(template);
+      mainNode = React.createElement(mainComponent, {});
+    } catch (error) {
+      // no-op
```
mikemurray: We should be checking for the React component first, and in the `catch` statement try getting the blaze component instead. This way we don't have a blank `catch` statement, and we assume that most components in the app are React.

prinzdezibel: @mikemurray The ordering was intentionally to ensure that the current behavior is retained and minimizing risk of breaking changes. That's the reason the Blaze template was probed first. But I've also been thinking that we should make the React component the first class citizen, not vice versa. But this would introduce a slightly higher risk to break user plugins (E.g. if a user has registered a React component with the same name like the Blaze template name).

PR #3528, opened 2018-01-20
[Feature Request] Prioritize Primary when multiple Shops match domain

pmn4:
#### Feature Request:

When multiple shops match the searched domain, prioritize the Primary Shop. This is important when searching for "shop settings" which are only stored on the Primary Shop (e.g., smtp settings).

This is unlikely to even be noticed because Mongo _usually_ returns documents in the order in which they were stored (and the Primary Shop is stored first), but there are a few scenarios which could disrupt this.
I noticed this recently in development, and figured I would suggest the change now long before it is needed.

### DIFF
--- a/imports/plugins/core/core/server/Reaction/core.js
+++ b/imports/plugins/core/core/server/Reaction/core.js
@@ -430,15 +430,21 @@ export default {
    */
   getShopIdByDomain() {
     const domain = this.getDomain();
+    const primaryShop = this.getPrimaryShop();
+
+    // in cases where the domain could match multiple shops, we first check
+    // whether the primaryShop matches the current domain. If so, we give it
+    // priority
+    if (primaryShop.domains.includes(domain)) {
+      return primaryShop._id;
+    }
+
     const shop = Shops.find({
       domains: domain
     }, {
       limit: 1,
       fields: {
         _id: 1
-      },
-      orderby: {
-        merchantShops: -1 // give preference to PrimaryShop if multiple results
       }
     }).fetch()[0];
 
--- a/imports/plugins/core/core/server/Reaction/core.js
+++ b/imports/plugins/core/core/server/Reaction/core.js
@@ -436,6 +436,9 @@ export default {
       limit: 1,
       fields: {
         _id: 1
+      },
+      orderby: {
+        merchantShops: -1 // give preference to PrimaryShop if multiple results
       }
     }).fetch()[0];
 


server/api/core/core.js
```diff
@@ -394,6 +394,9 @@ export default {
       limit: 1,
       fields: {
         _id: 1
+      },
+      orderby: {
+        merchantShops: -1 // give preference to PrimaryShop if multiple results
```
pmn4: if this were SQL, I would have done something like:
``` SQL
SELECT _id
FROM shops
WHERE domain = 'domain'
ORDER BY CASE shopType WHEN 'primary' THEN 0 ELSE 1 END # <--
LIMIT 1
```

but I couldn't find an equivalent construct in Mongo.
Instead, I sort on whether a Shop has Merchant Shops, assuming only the Primary Shop does.

zenweasel: @pmn4 Since `merchantShops` is an array, I wonder if this is going to always give you the results you want? Seems like we would want to sort on `merchantShop.lengrth` somehow?

zenweasel: Wow, we call this function > 900 times on startup.

pmn4: we sure do! and about 15/request (I have [a PR](https://github.com/reactioncommerce/reaction/pull/3510) for that too!)

pmn4: as for the sort, it is definitely not ideal. It is basically a null/not-null sort.
seems like sorting on length would be an expensive query, but since you are not likely to have too many shops, I guess it would be ok.
I am definitely out of my depth when it comes to Mongo, so open to any suggestions

zenweasel: Since `orderBy` is basically happening **after** the find, I wonder how much slower sorting it in memory would be? Although I wonder why this is doing a `find` and then a `fetch[0]`. Seems like we could be doing a `findOne`? If there's 1000 shops in a marketplace, that's a lot of data to pull across the network and then immediately throw away?

pmn4: there is a `limit: 1` in the `find`, so it shouldn't be pulling much/any more data across the wire than `findOne` would. Doesn't seem like getting a cursor here helps at all (since we immediately get the first record), so happy to change it to `findOne`

jshimko:
I feel like the bigger problem here is that you could possibly have more than one shop with the same domain.  Should we maybe consider just fixing that instead?  Then you would also know for sure that the query with `limit: 1` or a `findOne()` would always be safe.  And of course, you wouldn't have the issue of trying to access two shops with the same domain.

Just a thought.  Not sure if there was originally some reason for allowing that.  Or maybe I don't totally understand how domains are being used in a marketplace context.  Opinion @spencern?

pmn4:
@jshimko this is a tough one because _all_ shops will have the Primary Shop's domain to allow the Marketplace/Primary Shop owners to toggle between the shops in Admin mode.

This is a total edge case... Mongo sorts in such a way that this will likely never be an issue. The use-case for this change is if your site gets big enough that you start sharding your DB, or if for some reason, you delete and re-insert your primary shop.

I hit this while tinkering with the DB directly, so don't think it's a super important change, but one that could avoid a headache somewhere way down the line

aaronjudd:
@reactioncommerce/community can we move this forward?

zenweasel:
@jshimko What's holding us back from moving forward with this?

pmn4:
I completely understand the concern. If it helps you get more comfortable, this is a `$type` sort, and an argument can be made that this _is_ a [supported mongo feature](https://docs.mongodb.com/manual/reference/bson-type-comparison-order/#bson-types-comparison-order). LMK what you think.

Another solution would be something like this, like you hinted at earlier:
``` JavaScript
getShopIdByDomain() {
  const domainShops = Shops.find({
    domains: domain
  }, {
    fields: {
      _id: 1,
      shopType: 1
    }
  }).fetch();

  const shop = domainShops.find((s) => s.shopType === "primary") || domainShops[0];
  return shop._id;
}
```
If it is safe to assume that we have already fetched the primary shop (which seems likely), we could do:
``` JavaScript
getShopIdByDomain() {
  const primaryShop = Reaction.getPrimaryShop();

  if (primaryShop.domains.includes(domain)) {
    return primaryShop._id;
  }

  // the rest of the method as it is currently defined
}
```

spencern:
@zenweasel can you take a look at this?

zenweasel:
@pmn4 So I have been working with optimisations lately and I think this might have a performance implication that hasn't been noted in that it's impossible to index on that sort and in situations where a user may have a lot of shops it's going to cause this query to do a full-table scan.

Just thinking out loud but could we leverage the `primaryShop` flag somehow to order? If we can't sort on it maybe do queries for with and without and concatenate them?

Maybe I am missing something coming at this again after so long.

pmn4:
thanks for following up on this @zenweasel.
I am a little out of my depth in mongoland, and given your concerns with indexing, I re-wrote it a bit. lmk what you think.

imports/plugins/core/core/server/Reaction/core.js
```diff
@@ -414,9 +414,19 @@ export default {
    * @method
    * @memberof Core
    * @summary returns the shop which should be used given the current domain
+   * @return {StringId} shopId for the current domain
    */
   getShopIdByDomain() {
     const domain = this.getDomain();
+    const primaryShop = this.getPrimaryShop();
```
pmn4: I wanted to avoid making two queries for this method, but
1. if the condition on line 426 is satisfied, the second query will not execute
2. it is probably ok to assume that `getPrimaryShop` will be called at some other point and could be cached

loanlaux:
A client's project needed this one and I had to manually integrate @pmn4's fix as a temporary bandaid. Would be great to see it merged, as I think I'm not the only one who had to do this.

zenweasel:
I'm ok with this approach if we can get the conflicts resolved

pmn4:
@zenweasel rebased.

PR #3529, opened 2018-01-20
[Feature Request] Allow for ShopId when adding Brand Assets

pmn4:
#### Feature Request

I am working on a Shop Importer. i.e., there exists a data feed for all of my customers and with one click, an automated process pulls in their information, their products, and their brand assets.

I do not wish to rewrite code, so I moved the method body for `updateBrandAssets` to a function (and `export`ed it, for server-side reuse), added the `shopId` parameter (which remains `Reaction.getShopId()` when called from the client).

As a result of pulling the code out into a separate function, the diff is not great. I will highlight the changes.

let me know what you think.

### DIFF
--- a/private/data/i18n/en.json
+++ b/private/data/i18n/en.json
@@ -125,6 +125,8 @@
         "options": "Options",
         "paymentMethods": "Payment methods",
         "externalServices": "External services",
+        "shopBrandAssetsSaved": "Shop brand assets saved",
+        "shopBrandAssetsFailed": "Shop brand assets failed",
         "shopLocalizationSettingsSaved": "Shop localization settings saved.",
         "shopLocalizationSettingsFailed": "Shop localization settings update failed.",
         "shopPaymentMethodsSaved": "Shop Payment Methods settings saved.",
--- a/server/methods/core/shop.js
+++ b/server/methods/core/shop.js
@@ -11,6 +11,62 @@ import * as Collections from "/lib/collections";
 import * as Schemas from "/lib/collections/schemas";
 
 /**
+ * @name updateShopBrandAssets
+ * @method
+ * @param {Object} asset - brand asset {mediaId: "", type, ""}
+ * @param {String} shopId - the shop id coresponding to the shop for which
+ *                 the asset should be applied (defaults to Reaction.getShopId())
+ * @param {String} userId - the user id on whose behalf we are performing this
+ *                 action (defaults to Meteor.userId())
+ * @return {Int} returns update result
+ */
+export function updateShopBrandAssets(asset, shopId = Reaction.getShopId(), userId = Meteor.userId()) {
+  check(asset, {
+    mediaId: String,
+    type: String
+  });
+  check(shopId, String);
+
+  // must have core permissions
+  if (!Reaction.hasPermission("core", userId, shopId)) {
+    throw new Meteor.Error("access-denied", "Access Denied");
+  }
+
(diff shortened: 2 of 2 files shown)


server/methods/core/shop.js
```diff
@@ -10,6 +10,62 @@ import { Reaction } from "/lib/api";
 import * as Collections from "/lib/collections";
 import * as Schemas from "/lib/collections/schemas";
 
+/**
```
pmn4: this is mostly copy/pasted from below. changes are highlighted.

server/methods/core/shop.js
```diff
@@ -10,6 +10,62 @@ import { Reaction } from "/lib/api";
...
+ * @name shop/updateBrandAsset
+ * @method
+ * @memberof Methods/Shop
+ * @param {Object} asset - brand asset {mediaId: "", type, ""}
+ * @param {String} shopId - the shop id coresponding to the shop for which
+ *                 the asset should be applied
+ * @return {Int} returns update result
+ */
+export updateShopBrandAssets(asset, shopId) {
```
pmn4: added `shopId` param

server/methods/core/shop.js
```diff
@@ -10,6 +10,62 @@ import { Reaction } from "/lib/api";
...
+  check(asset, {
+    mediaId: String,
+    type: String
+  });
+  check(shopId, String);
+
+  // must have core permissions
+  if (!Reaction.hasPermission("core", null, shopId)) {
```
pmn4: added `shopId` to the permissions check.
I don't know the permissions code well, but I read the code and this feels right.

pmn4: the rest of this method is unchanged from the previous

server/methods/core/shop.js
```diff
@@ -855,44 +911,8 @@ Meteor.methods({
...
-      _id: Reaction.getShopId()
-    }, {
-      $push: {
-        brandAssets: {
-          mediaId: asset.mediaId,
-          type: asset.type
-        }
-      }
-    });
+    return updateShopBrandAssets(asset, Reaction.getShopId());
```
pmn4: since this can be called from the client, do not expose the `shopId` parameter, and instead, use the current shop

server/methods/core/shop.js
```diff
@@ -10,6 +10,62 @@ import { Reaction } from "/lib/api";
 import * as Collections from "/lib/collections";
 import * as Schemas from "/lib/collections/schemas";
 
+/**
+ * @name shop/updateBrandAsset
```
zenweasel: JSDoc does not match method name

zenweasel:
@pmn4 I am getting an error when making an uploaded image the brand image. No error in the client or server console just the red toast that says `shopSettings.shopBrandAssetsFailed`

pmn4:
@zenweasel ready for re-review

private/data/i18n/en.json
```diff
@@ -125,6 +125,8 @@
         "options": "Options",
         "paymentMethods": "Payment methods",
         "externalServices": "External services",
+        "shopBrandAssetsSaved": "Shop brand assets saved",
+        "shopBrandAssetsFailed": "Shop brand assets failed",
```
pmn4: Is it ok that I only know the `"en"` version?

zenweasel:
@jshimko Can you take over getting this one resolved?

PR #3539, opened 2018-01-22
closes #3540 Avalara: Return options.result on MethodHooks.after("cart/copyCartToOrder") call

### LINKED ISSUE #3540: Avalara: Fix bug on after hook call
The MethodHooks.after("cart/copyCartToOrder") call in the `hooks.js` file return `options` instead of `options.result`. This causes problems when you do subsequent after hooks, since the return value is not the one from the original method call but the options object itself.

**To Test**
```js
MethodHooks.after("cart/copyCartToOrder", (options) => {
   Logger.info("After cart/copyCartToOrder:", options);
   return options.result;
});
```

glmaljkovich:
closes #3540 
This is causing a bug on subsequent calls to the after hooks since the result is getting a copy of the options object instead of the original return value of the methods.

To see the problem add in any file:

```js
MethodHooks.after("cart/copyCartToOrder", (options) => {
   Logger.info("After cart/copyCartToOrder:", options);
   return options.result;
});
```


zenweasel:
@glmaljkovich Looks like this is causing some tests to fail

glmaljkovich:
@zenweasel I changed the PR / Issue to focus solely on the "cart/copyCartToOrder" hook which I experienced concrete problems with. Tests are now passing. Just as a comment, the previous commit (with all hooks returning options.result) failed because of a two seconds timeout. I leave the stack trace below. I still think that all MethodHooks.after() calls should return `options.result` by default so I find it weird that some tests fails.

**Stack Trace**

```
I20180123-02:40:06.428(0)?   300 passing (1m)
I20180123-02:40:06.429(0)?   2 failing
I20180123-02:40:06.429(0)? 
I20180123-02:40:06.429(0)?   1) orders test orders/refunds/refundItems should update the order as partially refunded if not all of items in the order are refunded:
I20180123-02:40:06.430(0)?      Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
I20180123-02:40:06.430(0)?   
I20180123-02:40:06.430(0)? 
I20180123-02:40:06.430(0)?   2) orders test orders/refunds/refundItems should update the order as refunded if all items in the order are refunded:
I20180123-02:40:06.431(0)?      Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```

zenweasel:
@glmaljkovich You are correct, these should all return `options.result`. I don't know why the tests were failing, I would need to dig in more since it doesn't make sense at first glance. Anyway I have created an issue https://github.com/reactioncommerce/reaction/issues/3583, for fixing the rest of the instances of these.

PR #3549, opened 2018-01-23
Remove server operation callbacks (make sync)

### LINKED ISSUE #3514: Remove callbacks from most server db operations
There are a lot of server code blocks that look something like this:

```
Cart.update({ _id: cartId }, update, function (error) {
    if (error) {
      Logger.warn(`Error in setting shipping query status to "pending" for ${cartId}`, error);
      return;
    }
    Logger.debug(`Success in setting shipping query status to "pending" for ${cartId}`, rates);
  });
```

Providing a callback to insert/update/remove makes it asynchronous on the server, but yet there seem to be sequences of events that are assuming the previous line has already finished its update. So I would expect that this patterns leads to random timing errors a lot. It would be much better IMO to remove all the callbacks and wrap in try/catch instead.

```
// Synchronous
const id = Product.insert(doc);
const prod = Product.findOne(id); // prod will definitely be defined

// Asynchronous
const id = Product.insert(doc, callback);
const prod = Product.findOne(id); // prod may or may not be defined
```

There isn’t necessarily anything wrong with intentionally making it async for good reasons, but usually I default to everything being synchronous, and then if I make something async for performance reasons, I put a comment above it explaining that it was intentional.

In hooks I’d say it’s definitely better to be synchronous. In certain methods, where you want the method to return quickly and the user doesn’t really care whether the operation was successful, then those places could use a callback.

Along with searching for and fixing all instances of this issue, it would be a good idea to document these recommendations and explanation in the docs. Meteor docs reference: http://docs.meteor.com/api/collections.html#Mongo-Collection-insert

> On the server, if you don’t provide a callback, then insert blocks until the database acknowledges the write, or throws an exception if something went wrong. If you do provide a callback, insert still returns the ID immediately. Once the insert completes (or fails), the callback is called with error and result arguments. In an error case, result is undefined. If the insert is successful, error is undefined and result is the new document ID.

Performance implications: If operations within methods or method hooks are made synchronous, they may be a little slower to return from the server. But they will also be less error prone, which is part of perceived speed.

aldeed:
Fix #3514

### DIFF
--- a/imports/plugins/included/marketplace/server/methods.js
+++ b/imports/plugins/included/marketplace/server/methods.js
@@ -22,17 +22,17 @@ export function marketplaceUpdateShopWorkflow(shopId, workflowStatus) {
   }
 
   if (status.includes(workflowStatus)) {
-    return Shops.update({
-      _id: shopId
-    }, {
-      $set: {
-        "workflow.status": workflowStatus
-      }
-    }, function (error) {
-      if (error) {
-        throw new Meteor.Error("server-error", error.message);
-      }
-    });
+    try {
+      return Shops.update({
+        _id: shopId
+      }, {
+        $set: {
+          "workflow.status": workflowStatus
+        }
+      });
+    } catch (error) {
+      throw new Meteor.Error("server-error", error.message);
+    }
   }
 
   throw new Meteor.Error("server-error", "Workflow status could not be updated, should be 'active' or 'disabled'");
--- a/server/methods/catalog.js
+++ b/server/methods/catalog.js
@@ -425,28 +425,22 @@ Meteor.methods({
       delete clone.lowInventoryWarningThreshold;
 
       copyMedia(productId, oldId, clone._id);
-      return Products.insert(clone, {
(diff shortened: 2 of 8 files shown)


Akarshit:
@aldeed 
Noob question: How does converting callbacks to try/catch improve performance. 
Is it because the callbacks were getting executed even in the case of success, and we were only concerned with error case; using try/catch, catch will only be executed in error case.

aldeed:
@Akarshit It actually probably makes the methods take longer to return because we now wait for the database ops to be confirmed, so it will be slower. However, it improves "performance" in that it should reduce random confusing errors due to something expecting the operation to be done already when it is actually still in progress.

In Meteor it's best to have everything in server code be written synchronous since it's within Fibers. The equivalent concept for non-Meteor would be making sure that you wait for a Promise result, by using `then` or async/await.

(See http://docs.meteor.com/api/collections.html#Mongo-Collection-insert for an explanation of how adding a callback makes it run async on the server.)

PR #3552, opened 2018-01-23
Remove unused code and packages

aldeed:
Removed several packages and one code file that did not appear to be used anywhere.


package.json
```diff
@@ -38,7 +38,6 @@
     "copy-to-clipboard": "^3.0.8",
     "core-js": "^2.5.1",
     "country-data": "^0.0.31",
-    "create-react-class": "^15.6.2",
```
kieckhafer: I'm still seeing `create-react-class` in the `package-lock.json`
 file, should it be removed form there as well?

aldeed: Looks like it is a dependency of "react" package, so it will remain in the lock file.

PR #3554, opened 2018-01-23
Fix not working crop feature

nicebro:
This PR resolves issue when images doesn't get cropped and resized versions after uploading.
How to test.
1. Run reaction server with given changes.
2. Update image to any product
3. With chrome devtools get link of updated image and open it in a new tab.
4. In address bar change url parameter &store=large to &store=small
5. You should see cropped square version of the product image

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,6 +11,7 @@
     "Assets": true, // Meteor global
     "Package": true, // Meteor global
     "FS": true, // gridFS,
+    "gm": true, // cfs:graphicsmagick
     "Alerts": true,  // this needs to get cleaned up, but ignoring for now
     "jest": true,
     "expect": true,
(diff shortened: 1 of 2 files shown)


nicebro:
Here when it happens It looks like accidental change. https://github.com/reactioncommerce/reaction/commit/5bcd994dfae415a8aff571406d45a4fcd1d3eb8a#diff-7cc0a2710f06d7093d38125e5bb5506f

impactmass:
Thanks for linking to that release. Looking further I can see that the change happened in this PR #2450 (which was when we updated our ESLint setup and had to remove some global references. My guess is that `gm` was available on `this` was the change was made, but not anymore)

Thanks for the fix. I tested it with @nnnnat and it worked on this PR.

PR #3556, opened 2018-01-23
Add eslint rules

aldeed:
Added many rules from Airbnb config, commented out for now pending later issues to fix, and changed a few existing rules. Fixed some issues caught by the rule changes.

@aaronjudd @spencern or anyone, let me know if you see anything you disagree with here. There will be about 3600 errors needing fixing from these rules. You can pull down, uncomment the rules, and run `npm run lint` to see.

After this is merged, I will write probably around 10 issues for uncommenting, aiming for ~360 errors per issue. It looks like a large percentage of these can be fixed with `eslint --fix`, but we need to do the commits in small, reviewable chunks.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,76 +1,80 @@
 {
-  "parser": "babel-eslint", // now for the support of allowImportExportEverywhere
+  "parser": "babel-eslint",
+  // "extends": [
+  //   "plugin:jsx-a11y/recommended"
+  // ],
   "env": {
     "browser": true,
-    "node": true,
     "es6": true,
-    "mocha": true
+    "jest/globals": true,
+    "mocha": true,
+    "node": true
   },
   "globals": {
-    "browser": true,
     "Assets": true, // Meteor global
     "Package": true, // Meteor global
     "FS": true, // gridFS,
-    "Alerts": true,  // this needs to get cleaned up, but ignoring for now
-    "jest": true,
-    "expect": true,
-    "test": true
+    "Alerts": true  // this needs to get cleaned up, but ignoring for now
+  },
+  "settings": {
+    "import/resolver": "meteor"
   },
   "parserOptions": {
-     "ecmaVersion": 6,
-     "sourceType": "module",
-      "ecmaFeatures": {
-        "arrowFunctions": true,
-        "blockBindings": true,
-        "classes": true,
-        "defaultParams": true,
-        "destructuring": true,
-        "forOf": true,
-        "generators": false,
-        "modules": true,
-        "objectLiteralComputedProperties": true,
-        "objectLiteralDuplicateProperties": false,
-        "objectLiteralShorthandMethods": true,
-        "objectLiteralShorthandProperties": true,
-        "spread": true,
-        "superInFunctions": true,
-        "templateStrings": true,
-        "jsx": true
-      },
-     "allowImportExportEverywhere": true
-  },
-  "plugins": ["react"],
-  // NOTE: We're now using eslint-4
-  "rules": {
+    "ecmaVersion": 2017,
+    "sourceType": "module",
+     "ecmaFeatures": {
+       "arrowFunctions": true,
+       "blockBindings": true,
+       "classes": true,
+       "defaultParams": true,
+       "destructuring": true,
+       "forOf": true,
+       "generators": false,
+       "modules": true,
+       "objectLiteralComputedProperties": true,
+       "objectLiteralDuplicateProperties": false,
+       "objectLiteralShorthandMethods": true,
+       "objectLiteralShorthandProperties": true,
+       "spread": true,
+       "superInFunctions": true,
+       "templateStrings": true,
+       "jsx": true
+     },
+    "allowImportExportEverywhere": true
+ },
+ "plugins": ["import", "jest", "jsx-a11y", "react"],
+ "rules": {
     /**
-     * Strict mode
-     * babel inserts "use strict"; for us
-     * http://eslint.org/docs/rules/strict
-     */
+    * Strict mode
+    * babel inserts "use strict"; for us
+    * http://eslint.org/docs/rules/strict
+    */
     "strict": 0,
 
     /**
-     * ES6
-     */
+    * ES6
+    */
     "object-curly-spacing": ["error", "always", { "objectsInObjects": true }],
     "no-var": 2, // http://eslint.org/docs/rules/no-var
+
     /**
-     * Variables
-     */
+    * Variables
+    */
     "no-shadow": 2, // http://eslint.org/docs/rules/no-shadow
     "no-shadow-restricted-names": 2, // http://eslint.org/docs/rules/no-shadow-restricted-names
     "no-const-assign": 2, // http://eslint.org/docs/rules/no-const-assign
     "no-unused-vars": [2, { // http://eslint.org/docs/rules/no-unused-vars
-      "vars": "local",
-      "args": "after-used"
+      "vars": "all",
+      "args": "after-used",
+      "ignoreRestSiblings": true
     }],
     "no-use-before-define": [2, "nofunc"], // http://eslint.org/docs/rules/no-use-before-define
     "no-implicit-globals": 2, // http://eslint.org/docs/rules/no-implicit-globals
 
     /**
-     * Possible errors
-     */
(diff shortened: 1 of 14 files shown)


.eslintrc
```diff
@@ -1,76 +1,80 @@
 {
-  "parser": "babel-eslint", // now for the support of allowImportExportEverywhere
+  "parser": "babel-eslint",
+  // "extends": [
+  //   "plugin:jsx-a11y/recommended"
```
aldeed: This pulls in default accessibility recommendations for React components, which I think is a requirement for Reaction, right? This, like the rules, can be uncommented in a later ticket. There are ~100 errors from it, and they'll take manual work to solve.

.eslintrc
```diff
@@ -1,76 +1,80 @@
...
-      },
-     "allowImportExportEverywhere": true
+  "settings": {
+    "import/resolver": "meteor"
   },
-  "plugins": ["react"],
-  // NOTE: We're now using eslint-4
-  "rules": {
+  "parserOptions": {
+    "ecmaVersion": 2017,
```
aldeed: Diff looks weird here, but the only actual change to `parserOptions` is changing ecmaVersion to 2017

imports/plugins/core/accounts/client/helpers/templates.js
```diff
@@ -26,35 +26,3 @@ Template.registerHelper("displayName", function (displayUser) {
     }
   }
 });
-
-/*
- * registerHelper fName
- */
-Template.registerHelper("fName", function (displayUser) {
```
aldeed: Note that (a) the exact same helper was registered in two places, and (b) I chose to remove it rather than fix the lint error in it. I do not see where this helper is used anywhere.

server/startup/accounts.js
```diff
@@ -124,49 +123,52 @@ export default function () {
...
-            user.username = user.services[service].username;
-            user.name = user.services[service].full_name;
-            additionals.name = user.services[service].full_name;
-            additionals.profile.picture = user.services[service].profile_picture;
-            additionals.profile.bio = user.services[service].bio;
-            additionals.profile.name = user.services[service].full_name;
-            additionals.profile.username = user.services[service].username;
+        const userServices = user.services;
+        for (const service in userServices) {
+          if ({}.hasOwnProperty.call(userServices, service)) {
```
aldeed: Main lint fix here was wrapping in this `if`, but I also added the `serviceObj` const below so that there isn't so much dot property accessing.

.eslintrc
```diff
@@ -253,6 +256,91 @@
...
     "space-in-parens": [2, "never"], // http://eslint.org/docs/rules/space-in-parens
     "spaced-comment": [2, "always"], // http://eslint.org/docs/rules/spaced-comment
-    "arrow-spacing": [2, { "before": true, "after": true }] // https://eslint.org/docs/rules/arrow-spacing
+    "arrow-spacing": [2, { "before": true, "after": true }], // https://eslint.org/docs/rules/arrow-spacing
+
+    /**
+    * Additional rules to enable one by one
+    */
+    // "array-bracket-spacing": ["error", "never"],
+    // "array-callback-return": ["error", { "allowImplicit": true }],
```
spencern: this should be super helpful. 👍

.eslintrc
```diff
@@ -253,6 +256,91 @@
...
+    // "arrow-body-style": ["error", "as-needed", { "requireReturnForObjectLiteral": false }],
+    // "arrow-parens": [ "error", "as-needed", { "requireForBlockBody": true }],
```
spencern: This may need to be "always" here based on our prior rules.

.eslintrc
```diff
@@ -253,6 +256,91 @@
...
+    // "block-spacing": ["error", "always"],
+    // "computed-property-spacing": ["error", "never"],
+    // "dot-location": ["error", "property"],
```
spencern: I prefer this to "object" as well. 👍

.eslintrc
```diff
@@ -253,6 +256,91 @@
...
+    // "function-paren-newline": ["error", "multiline"],
+    // "import/export": "error",
```
spencern: @aldeed Do these `import` rules come from benmosher's `eslint-plugin-import` plugin? https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules

.eslintrc
```diff
@@ -253,6 +256,91 @@
...
+    // "import/first": ["error", "absolute-first"],
```
spencern: How will this play with dynamic imports?

.eslintrc
```diff
@@ -253,6 +256,91 @@
...
+    // "import/newline-after-import": "error",
+    // "import/no-duplicates": "error",
+    // "import/no-mutable-exports": "error",
+    // "import/no-named-default": "error",
+    // "new-parens": "error",
+    // "newline-per-chained-call": ["error", { "ignoreChainWithDepth": 4 }],
```
spencern: 4 is sensible

.eslintrc
```diff
@@ -253,6 +256,91 @@
...
+    // "no-await-in-loop": "error",
+    // "no-bitwise": "error",
+    // "no-case-declarations": "error",
+    // "no-confusing-arrow": ["error", { "allowParens": true }],
+    // "no-empty-pattern": "error",
+    // "no-lonely-if": "error",
+    // // This rule copied from Airbnb's config
+    // "no-mixed-operators": ["error", {
```
spencern: 👍

.eslintrc
```diff
@@ -253,6 +256,91 @@
...
+    //     ["&", "|", "^", "~", "<<", ">>", ">>>"],
+    //     ["==", "!=", "===", "!==", ">", ">=", "<", "<="],
+    //     ["&&", "||"],
+    //     ["in", "instanceof"]
+    //   ],
+    //   "allowSamePrecedence": false
+    // }],
+    // "no-multi-assign": ["error"],
+    // "no-multi-spaces": ["error", { "ignoreEOLComments": false }],
+    // "no-plusplus": "error",
```
spencern: I'm good with this one.

.eslintrc
```diff
@@ -253,6 +256,91 @@
...
+    // "no-unneeded-ternary": ["error", { "defaultAssignment": false }],
+    // "no-unsafe-finally": "error",
+    // "no-useless-computed-key": "error",
+    // "no-useless-concat": "error",
+    // "no-useless-constructor": "error",
+    // "no-useless-escape": "error",
+    // "no-void": "error",
+    // "object-curly-newline": ["error", { "ObjectExpression": { "multiline": true, "consistent": true }, "ObjectPattern": { "multiline": true, "consistent": true } }],
+    // "object-property-newline": ["error", { "allowAllPropertiesOnSameLine": true }],
+    // "object-shorthand": ["error", "always", { "ignoreConstructors": false, "avoidQuotes": true }],
```
spencern: This will be a big cleanup job but will make a lot of code more consistent. What is the thought behind using `avoidQuotes: true`?

.eslintrc
aldeed: I'm cool with that

.eslintrc
aldeed: Yes. The plugin packages all begin with `eslint-plugin-` in the package.json, and then the part of the name after `eslint-plugin-` is referenced in the "plugins" array in this file. Any rule in `import/` namespace is from the import plugin package, etc.

I pulled in the rules Airbnb uses for `import/`, but then changed/removed a few to match Meteor/Reaction conventions. These will mostly help catch consistency things (like order of import statements) and potential issues (like mutable exports).

I added the `eslint-import-resolver-meteor` package, too, which makes it understand Meteor's way and be able to resolve absolute import paths.

.eslintrc
aldeed: Per the docs, it only looks at `import` lines, not `require()`, so I don't think it would look at dynamic imports either. Which is probably what we want.

.eslintrc
aldeed: Just copied this one from Airbnb, so I'm ok with any number we agree on

.eslintrc
aldeed: I copied this one from Airbnb without looking closely, but now that you bring it up, I think I'd actually prefer `avoidQuotes: false`. I typically use shorthand even if a property has to have quotation marks for special characters in it.

PR #3559, opened 2018-01-23
Product images should be reactive

### LINKED ISSUE #3475: Product images do not publish reactively
### Expected behavior
When I make a product visible and publish it, any connected client that has that product loaded should also have the appropriate images loaded as well.

### Actual behavior
The product is published, but the images are not loaded until the view is reloaded (hard refresh works).

1. Setup and activate a shop
2. Import or create several products for the shop
3. Connect an unauthenticated client as a "consumer"
4. Make visible and publish a previously hidden product.

### Versions
```
Node: 8.9.0
NPM: 5.5.1
Meteor Node: 8.8.1
Meteor NPM: 5.4.2
Reaction CLI: 0.24.2
Reaction: 1.6.1
Reaction branch: master
Docker: 17.09.0-ce
```

prinzdezibel:
Closes #3475

### DIFF
--- a/server/publications/collections/products.js
+++ b/server/publications/collections/products.js
@@ -545,6 +545,27 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
   const mediaProductIds = productCursor.fetch().map((p) => p._id);
   const mediaCursor = findProductMedia(this, mediaProductIds);
 
+  const handle = productCursor.observe({
+    added: (product) => {
+      let productId;
+      if (product.type === "variant") {
+        productId = product.ancestors[0];
+      } else {
+        productId = product._id;
+      }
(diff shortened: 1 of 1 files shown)


zenweasel:
@kieckhafer wasn't this made non-reactive on purpose? I thought that one of the things that broke the image import for a while was it pushing out updates over the wire to the client.

kieckhafer:
@zenweasel that could very well be the case, will approve assuming so.

spencern:
@kieckhafer @zenweasel I'm not sure if we made the images for imports non-reactive on purpose, I don't remember doing so personally. That's a much more minor issue to me though, and this seems to resolve the major issue. The image import issue could be due to the images getting imported via job after all of the products are finished importing.

PR #3564, opened 2018-01-24
(fix): add guard to for loop to fix lint error

kieckhafer:
Fixes the lint error at `server/startup/accounts.js 127:9`:
`The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in`

### DIFF
--- a/server/startup/accounts.js
+++ b/server/startup/accounts.js
@@ -125,37 +125,39 @@ export default function () {
         additionals.groups = [group._id];
         // also add services with email defined to user.emails[]
         for (const service in user.services) {
-          if (user.services[service].email) {
-            const email = {
-              provides: "default",
-              address: user.services[service].email,
-              verified: true
-            };
-            user.emails.push(email);
-          }
-          if (user.services[service].name) {
-            user.username = user.services[service].name;
-            additionals.profile.name = user.services[service].name;
-          }
-          // TODO: For now we have here instagram, twitter and google avatar cases
-          // need to make complete list
-          if (user.services[service].picture) {
-            additionals.profile.picture = user.services[service].picture;
-          } else if (user.services[service].profile_image_url_https) {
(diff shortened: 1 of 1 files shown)


server/startup/accounts.js
```diff
@@ -125,37 +125,39 @@ export default function () {
...
-          // Correctly map Instagram profile data to Meteor user / Accounts
-          if (user.services.instagram) {
-            user.username = user.services[service].username;
-            user.name = user.services[service].full_name;
-            additionals.name = user.services[service].full_name;
-            additionals.profile.picture = user.services[service].profile_picture;
-            additionals.profile.bio = user.services[service].bio;
-            additionals.profile.name = user.services[service].full_name;
-            additionals.profile.username = user.services[service].username;
+          if (user.services.hasOwnProperty(service)) {
```
spencern: This should be
```
{}.hasOwnProperty.call(user.services, service)
```
Because we should not use prototype builtins. This is a rule that we will be enabling in the next release.
see: https://eslint.org/docs/rules/no-prototype-builtins

mikemurray: @spencern would it be better for it to be `Object.hasOwnProperty(user.services, service)` then vs using `{}`?

kieckhafer:
@spencern updated

server/startup/accounts.js
spencern: We should probably work to remove the use of `for..in` in favor of `Object.keys(object).forEach` as it does not have the problem that `for..in` can exhibit and does not need to be guarded.

I'll merge this one as it's fixing a lint error, and we don't have anything explicit about avoiding `for..in` yet, but worth considering.

spencern:
not sure how I missed this, but looks like another PR resolved this already in 1.6.6. Closing.

PR #3576, opened 2018-01-24
[WIP] (fix): eslint rule updates

### LINKED ISSUE #3566: Enable and fix eslint rules (1)
Uncomment the following rules in .eslintrc, run `npm run lint`, and fix all errors, either manually or using `eslint --fix`

array-bracket-spacing
array-callback-return
arrow-body-style
arrow-parens

kieckhafer:
Resolves #3566 

`array-bracket-spacing`
`array-callback-return`
`arrow-body-style`
`arrow-parens`

Please take an extra hard look at these two commits, as I'm not 100% comfortable all the returns are what we want them to be (some `undefined` vs `null`):
https://github.com/reactioncommerce/reaction/pull/3576/commits/37e7d1fdafc692039993320379071a9482518b02
https://github.com/reactioncommerce/reaction/pull/3576/commits/fca582c8c6d31c517c4bd164481b287341d8e715 (there are some fixes to the above commit in this one)

**Notes**

Enabling `array-callback-return` has resulted in many cases where a semi-colon is removed. While it's not explicitly required, in the past we have required them internally, so perhaps something to look into.

Enabling `array-callback-return` has also resulted in a few cases where the line length exceeded our 160 character limit, which in turn produced a new linting error, `max-len`. I've disabled the 160 character limit on these lines, as that seems to be a less important lint rule than `array-callback-return`. I am still seeing this warning here, even though I have the `eslint-disable-line` on the line, so I'm not sure what to do there: `imports/plugins/included/product-variant/components/productGridItems.js 69:1`

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -263,7 +263,7 @@
     * Additional rules to enable one by one
     */
     "array-bracket-spacing": ["error", "never"],
-    // "array-callback-return": ["error", { "allowImplicit": true }],
+    "array-callback-return": ["error", { "allowImplicit": true }],
     "arrow-body-style": ["error", "as-needed", { "requireReturnForObjectLiteral": false }],
     "arrow-parens": [ "error", "always", { "requireForBlockBody": true }],
     // "block-spacing": ["error", "always"],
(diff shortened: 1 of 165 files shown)


kieckhafer:
Refer to #3578

PR #3577, opened 2018-01-24
refactor: Remove unused hook

aldeed:
This hook has an error in the logic making it probably not work as intended, and I don't see any update calls anywhere that have a modifier that would match. So it seems safe to remove this.

Assigning @aaronjudd since it seems you wrote this years ago. Is this a necessary feature that we should reimplement elsewhere in a better way?


aaronjudd:
I have no idea what that was originally intended to do, but regardless doesn't seem that what it could have done, would be appropriate in a hook.. so I'd say the could be deprecated, and if there are any unintended consequences, it'd be more appropriately handled elsewhere.

PR #3578, opened 2018-01-24
(fix): eslint rule updates (1)

### LINKED ISSUE #3566: Enable and fix eslint rules (1)

kieckhafer:
Resolves #3566 

`array-bracket-spacing`
`arrow-body-style`
`arrow-parens`
`array-callback-return`

There are a lot of changes that aren't fixed by `eslint --fix` in the PR, including some rewrites of `map` to `filter` and `find`. Please pay particular attention to commit https://github.com/reactioncommerce/reaction/pull/3578/commits/3fdf607476148778ce890d9642c0d0752520c46e to make sure switching `forEach` in for `map` was used correctly in these cases where we didn't need the return value.

Also, these commits have some changes to taxCalc and display, @zenweasel since you wrote these initially, I'm tagging you so you can make sure it works how it did before: https://github.com/reactioncommerce/reaction/pull/3578/commits/88837f717da95bdf98a33e30faee7a28fae0355a https://github.com/reactioncommerce/reaction/pull/3578/commits/27b3182828a53788819c11e7806fc56196c84be0

As noted in #3589, this PR has also pushed a few lines over our 160 character limit, which introduced new linting errors, so I've disabled the line limit on those lines, as that seems like the less important lint.


imports/plugins/included/product-variant/components/productGridItems.js
```diff
@@ -66,9 +66,9 @@ class ProductGridItems extends Component {
       if (this.props.isMediumWeight()) {
         return (
           <div className={`product-additional-images ${this.renderVisible()}`}>
-            {this.props.additionalMedia().map((media) => {
-              return <span key={media._id} className="product-image" style={{ backgroundImage: `url('${media.url({ store: "medium" })}')` }} />;
-            })}
+            {/* eslint-disable max-len */}
+            {this.props.additionalMedia().map((media) => <span key={media._id} className="product-image" style={{ backgroundImage: `url('${media.url({ store: "medium" })}')` }} />)}
```
zenweasel: This seems like a step backwards in terms of readability imho.

lib/api/products.js
```diff
@@ -560,7 +544,7 @@ ReactionProduct.cloneProduct = function (productOrArray) {
  */
 ReactionProduct.archiveProduct = function (productOrArray) {
   const products = !_.isArray(productOrArray) ? [productOrArray] : productOrArray;
-  const productIds = _.map(products, product => typeof product === "string" ? product : product._id);
+  const productIds = _.map(products, (product) => typeof product === "string" ? product : product._id);
```
zenweasel: Should we use the native `map` here?

server/api/core/core.js
```diff
@@ -904,15 +904,11 @@ export default {
...
-            return config.name === packageSetting.name;
-          });
+          settingsFromFixture = _.find(registryFixtureData[0], (packageSetting) => config.name === packageSetting.name);
         }
 
         // Setting already imported into the packages collection
-        const settingsFromDB = _.find(packages, (ps) => {
-          return (config.name === ps.name && shopId === ps.shopId);
-        });
+        const settingsFromDB = _.find(packages, (ps) => (config.name === ps.name && shopId === ps.shopId));
```
zenweasel: Can we use the native `find`?

server/methods/core/orders.js
```diff
@@ -42,10 +40,8 @@ export function orderCreditMethod(order) {
  * @return {Pbject} returns entire payment method
```
zenweasel: Typo

imports/plugins/included/product-variant/components/productGridItems.js
kieckhafer: I agree. There are a lot of places where readability seems to go backwards with all these new linter rules.

What do we do here though? Leave it and pass the linter?

imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
```diff
@@ -489,18 +491,21 @@ function orderToSalesInvoice(order) {
   const currencyCode = company.currency;
   const orderShipping = order.getShippingTotal();
   const orderDate = moment(order.createdAt).format();
-  const lineItems = order.items.map((item) => {
+  const lineItems = order.items.reduce((items, item) => {
```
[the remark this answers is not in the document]
kieckhafer: @spencern and I paired on this for a bit, and decided to go with `reduce` because we can then filter out any `undefined` results that were occurring with the map, creating a dense array with pure values only, not a sparse array with values and `undefined`'s.

Prior to this update, if the condition `items.variants.taxable` was not being met, nothing was returning. This new linter rule requires a `return`, which would mean we'd be returning `undefined` for all items where the if statement condition wasn't met.

By reducing it instead, we are eliminating those `undefined` items.

kieckhafer:
@zenweasel all issues updated or addressed, good for another look.

zenweasel:
@kieckhafer Looks like we have conflicts now?

imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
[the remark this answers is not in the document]
zenweasel: Makes sense to me

kieckhafer:
@zenweasel merge conflicts fixed. MERGE MERGE MERGE before more come along!!!

Issue #3568, opened 2018-01-24
Enable and fix eslint rules (3)


aldeed:
Uncomment the following rules in .eslintrc, run `npm run lint`, and fix all errors, either manually or using `eslint --fix`

import/export
import/first
import/newline-after-import
import/no-duplicates
import/no-mutable-exports
import/no-named-default
new-parens
newline-per-chained-call

PR #3580, opened 2018-01-25
Fix: import/export eslint rule issues

spencern:
Resolves part 1/8 of #3568 

Biggest change here is having to rename 3 exports from `imports/plugins/core/ui/client/containers/index.js`

Alerts => AlertContainer
  This was a breaking change and I had to change productDetailSimple as well

App => AppContainer
ReactionAvatar => ReactionAvatarContainer

I also had to rename the Currency schema exported from `lib/collections/schemas/payments.js` but it does not appear that schema was used outside of that file.

Breaking this PR out from the rest of the eslint updates in my issue because these should be reviewed a little more intentionally.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -271,7 +271,16 @@
     // "dot-location": ["error", "property"],
     // "function-paren-newline": ["error", "multiline"],
     "import/export": "error",
-    // "import/first": ["error", "absolute-first"],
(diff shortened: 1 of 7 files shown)


spencern:
Closed by #3585 which contains this PR as well.

Issue #2040, opened 2017-03-27
React + SSR conversion of remaining components


mikemurray:
# React + SSR

This epic is meant to track the conversion of all customer / admin facing components, layouts, templates to React. This means no more Blaze wrappers around any react component, and no Blaze components at any point in the React layout tree. All react all the time.

**Keep in mind, for SSR to work, no react component may access `document` or `window` while being rendered server-side. If you need them, you'll need to check if they exists before trying to access.**

**Below List in ordered by approximate priority**

## Router

- [x] #2049 Must attempt render a react layout first, fallback to blaze until the conversion is complete. (PR #2123) @mikemurray
- [x] #2049 `ReactionLayout` should support react components / layouts if one is registered (PR #2123) @mikemurray
- [ ] **BLOCKED** Enable SSR support after everything is react @mikemurray
- [x] FlowRouter -> React Router (PR #2123) @mikemurray

## Layout

- [x] Separate `coreAdminLayout` from `coreLayout` (PR #2123) @mikemurray
- [x] #2049 Layout should use the `React` version of ReactionLayout, used in the PDP page @mikemurray

## Navigation bar

- [x] #2041 Navigation bar itself 
- [x] #2042 Cart drawer & checkout popup when adding to cart @andela-aatanda
- [x] #2043 Tag navigation (already has a react version, for the tag component / `imports/plugins/core/ui/core/client/compnents/tags/ & imports/plugins/core/ui/core/client/containers/tagListContainer) @kieckhafer
- [x] #2044 Alerts, Language and other dropdowns

## Product Grid

- [x] #2050 Grid layout & items

## Search

- [x] #2051 Search modal and all subviews (`ui-search` package)
- [x] #2050 Grid layout & items for products (needs to use the exact same component as the grid)

## Checkout

- [ ] #2052 Checkout layout
- [ ] #2052 Checkout progress
- [ ] #2055 Login (shared with login dropdown and 403 page)
- [ ] #2053 & #2052 Address book, (is also used in user profile)
- [ ] #2052 Shipping options
- [ ] #2052 or #2042 Summary
- [ ] #2052 Payment options

## Product Detail

- [x] #2032 PDP Variant admin view (PR #2086) @kieha 

## Orders

- [x] #2045 Orders summary card (PR #2105) @joykare 

## Admin

- [x] Action view should render a react component if one exists. (DONE, if you register a react component with the same name you use for `template` while opening an ActionView, it will use the react component before falling back to a blaze template.)
[...]

Issue #3105, opened 2017-10-13
Accounts Admin - Implement accordion editing for group editing 


rymorgan:
### Expected behavior

- [ ] The groups should be editable in accordion fashion vs below the table. Where it is placed now loses context for the user because the correlation between the group and it’s permissions aren’t apparent.

Closed: https://zpl.io/bJq6ZDa
Open: https://zpl.io/ad7orpV

- [ ] Also, the group name should be editable via the penicel icon button. 
- [ ] The styles should match the design prototypes

**Important note** 
Before starting let's strategize on implementation. This should be componentized using a current component or built as a new component and added to the style guide. I have design plans to use this interaction in other places in the app.**

### Actual behavior
The UX is sloppy. The permissions toggles appear below the table where context is easily lost.

Issue #405, opened 2015-07-10
Accessability review


aaronjudd:
A thorough review of accessibility implementation (which is currently partial at best). 
A good start might be:
http://a11yproject.com/checklist.html

vinh3928:
I might be interested in doing this as I am exploring the code base. Can you elaborate more on what you are looking for with this issue?

aaronjudd:
What had I had in mind was just general run through of the important nav, forms, sectional elements and ensure they have role and aria properties.  See: http://w3c.github.io/aria-in-html/

I was trying to find a good intro doc, this might be helpful http://learn.shayhowe.com/advanced-html-css/semantics-accessibility/

We're in the process of making quite a few UI updates, which is one reason this is in a backlog right now, but this does needs some focus as the implementation has been a little haphazard.

minimart:
I would be very interested in continuing work on this issue! 😃 

There is a very nice Chrome plugin called [aXe](https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd?hl=en-US) that we have been using for a while now that seems to do a pretty thorough job of finding violations.

 Would listing out each violation as a separate issue and PR be the best way to go about starting to make these changes? Or would keeping them all within this issue be better for the project?

aaronjudd:
@mikemurray can you review as part of your Aria review, close this when done #3574 #3582, and create new issues for remaining tasks.

mikemurray:
@aaronjudd will do

mikemurray:
Closed by https://github.com/reactioncommerce/reaction/pull/3582 and additional tickets https://github.com/reactioncommerce/reaction/issues/3623 https://github.com/reactioncommerce/reaction/issues/2040

PR #3582, opened 2018-01-25
(refactor): 3574 eslint-9 and Aria

### LINKED ISSUE #3574: Enable and fix eslint rules (9)
Uncomment the following lines in .eslintrc, run `npm run lint`, and fix all errors, either manually or using `eslint --fix`

```
// "extends": [
  //   "plugin:jsx-a11y/recommended"
  // ],
```

mikemurray:
Resolves #3574 

## Summary

Refactor components to fix eslint issues.

## Changes
- (refactor): `tabIndex` should always be defined `tabIndex={0}` never use positive integer values, only 0.
- (refactor): Made several components that used `href="#` into `ComponentsButton` and removed href.
- (refactor): Some `span` and `div` elements converted to simulated buttons with role=`button` and `onKeyUp` event handlers.
- [breaking] [low-impact] Removed onClick from badge. Badges are not supposed to be clickable. Use a `Button` instead.

## Testing
- Run `npm run lint` observe no issues
- Run app, and everything should work, and more things should be tabbable/accessible due to better Aria compliance

### General

- [x] header role="banner" for global page header.
- [x] nav role="navigation" for navigation (taken care of by `nav` element for main nav)
- [x] main role="main" Wraps the focal content of document. Use only once. `same as <main>` in coreLayout.js
- [ ] ~~article role="article" PDP page?~~ All content in main, there are no real "articles"
- [x] aside role="complementary". (actionView)
- [ ] ~~footer role="contentinfo" main page footer, even though it's blank at the moment~~ No footer
- [x] form role="search" search field

### Language
- [x] html lang="en"

### Document Outline
- [x] Use semantic headings and structure

### Links
- [x] Ensure links have :focus state. 
- [x] Ensure links are recognizable (underlined). Most links wither have a different color, or some other decoration to make them visible.
- [ ] ~~Provide a “Skip to main content” link.~~ N/A

### Images
- [x] Use appropriate alt text. Read article on using ALT text

### JavaScript
- [x] Unobtrusive JavaScript (never use in line scripting).
- [ ] ~~No-JS Alternatives `<noscript>`.~~ This is SSR, or some alternative, which has many tickets. https://github.com/reactioncommerce/reaction/issues/2040

### Forms
- [x] Logical layout for tabs
- [x] Associated label for all form controls (e.g. input, select etc.) 
- [x] Make sure placeholder attributes are not being used in place of label tags. WHATWG 
An exception to this rule would be smaller forms with one or two fields (eg. search or log in forms)
- [ ] ~~Group related form elements with `fieldset` and describe the group with `legend`~~ This is a matter of form design of form component.

### Media (Audio and Video) (Not currently applicable)
- [ ] ~~Provide text transcripts~~
- [ ] ~~Synchronized subtitles for videos~~

### Color and Contrast https://github.com/reactioncommerce/reaction/issues/3623
- [ ] ~~Test color contrast~~

### Test for different types of color blindness. https://github.com/reactioncommerce/reaction/issues/3623
- [ ] ~~Deuteranopia~~
- [ ] ~~Protanopia~~
- [ ] ~~Tritanopia~~

### Testing
- [ ] Test using a screen reader 
- [x] Test using keyboard only. (App is more navigatable with tabbing)


imports/plugins/core/ui/client/components/badge/badge.js
```diff
@@ -4,52 +4,26 @@ import classnames from "classnames/dedupe";
...
+  state = {
+    tooltipOpen: false
+  };
 
   get isTooltipOpen() {
     return this.state.tooltipOpen;
   }
 
-  handleButtonMouseOver() {
+  handleMouseOver = () => {
```
pmn4: I am admittedly new to ES# syntax, but this technique yields a method where `this` is set properly?
at first glance, it looks like `this` would refer to the class `Badge` and not an instance of `Badge`

imports/plugins/core/ui/client/components/checkbox/checkbox.js
```diff
@@ -20,13 +30,15 @@ class Checkbox extends Component {
 
   render() {
     return (
-      <label>
+      <label htmlFor={this.id}>
```
pmn4: I was under the impression that you do not need `for` if the label wraps the input 🤷‍♂️

imports/plugins/core/ui/client/components/badge/badge.js
mikemurray: it refers to the instance of the class

imports/plugins/core/ui/client/components/checkbox/checkbox.js
mikemurray: I was surprised by it too, but the rules state otherwise.

mikemurray:
Along with fixing the linting issues, I'm going to run through the checklist here https://a11yproject.com/checklist.html provided by ticket https://github.com/reactioncommerce/reaction/issues/405 and make sure the ones that apply to us are taken care of.

spencern:
@mikemurray is this ready for review now?

kieckhafer:
The app is definitely "more" tabbable, and easier navigated, when using the keyboard. A huge improvement overall.

Some notes:

**Ordering a product**
I was able to nearly go all the way from product grid to final checkout with just the keyboard, with two issues:
1) The cart icon was not "clickable" via keyboard. See TypeError in previous comment, this should fix that error.
2) If I was able to tab to the slideout cart fast enough, and get to checkout, I was able to create an account, add my address, but Shipping options were not tabbable. Once I clicked (with the mouse) on shipping options, i was able to again complete the rest with my keyboard only.

**Navigating the site**
- Cart icon isn't clickable (see other comment)
- Tabbing skips Login dropdown completely (but is tabbable once you are already logged in)
- The alerts dropdown is tabblable, but Currency, Language, and User profile dropdowns open the menu, but the menu items are not tabbable nor navigatable via arrow keys
- Same menu items above, should the `esc` key exit these menus? Currently, only pressing `return` again on the dropdown trigger will close the item, so you can have multiple items open at once
reaction

**As an admin**
- If i tab / select and of the dashboard shortcuts icons, i have to tab all the way through everything again (or tab backwards) to get back to the menu i just opened.
- I cannot toggle the old style switch using `return`
reaction
- New style switches ARE tabbable 👍 
reaction
- `react-table` isn't tabbable (I've spoken with the Devs there, they "want to get it done", but I think they really want someone to PR aria work for them 
- Checkboxes do not change when "clicked" via `return`
reaction

imports/plugins/core/checkout/client/components/cartItems.js
```diff
@@ -51,10 +51,10 @@ class CartItems extends Component {
         >
           {handleImage(item) ?
             <div className="center-cropped" style={{ backgroundImage: `url('${handleImage(item).url({ store: "small" })}')` }}>
-              <img src={handleImage(item).url({ store: "small" })} className="product-grid-item-images img-responsive" />
+              <img src={handleImage(item).url({ store: "small" })} className="product-grid-item-images img-responsive" alt="" />
             </div> :
             <div className="center-cropped" style={{ backgroundImage: "url('/resources/placeholder.gif')" }}>
-              <img src="/resources/placeholder.gif" className="product-grid-item-images img-responsive" />
+              <img src="/resources/placeholder.gif" className="product-grid-item-images img-responsive" alt="" />
```
kieckhafer: Should we put something int he `alt` tag, instead of leaving it blank?

imports/plugins/core/checkout/client/components/completedOrderItem.js
```diff
@@ -18,7 +18,7 @@ const CompletedOrderItem = ({ item, handleDisplayMedia }) => {
   }
   return (
     <div className="row order-details-line">
-      <div className="order-details-media"><img src={image} /></div>
+      <div className="order-details-media"><img src={image} alt="" /></div>
```
kieckhafer: Should we put something int he alt tag, instead of leaving it blank?

imports/plugins/core/orders/client/components/lineItems.js
```diff
@@ -61,11 +61,11 @@ class LineItems extends Component {
 
     if (displayMedia(uniqueItem)) {
       return (
-        <img src={displayMedia(uniqueItem).url()}/>
+        <img src={displayMedia(uniqueItem).url()} alt="" />
```
kieckhafer: Should we put something int he alt tag, instead of leaving it blank?

imports/plugins/core/ui/client/components/button/button.jsx
```diff
@@ -51,6 +51,12 @@ class Button extends Component {
     }
   }
 
+  handleKeyUp = (event) => {
+    if (event.keyCode === 32 || event.keyCode === 13) {
```
kieckhafer: Somewhere else in the app, you've commented what these event.keyCodes are equal to, we should add that here too for easy reference.

imports/plugins/core/ui/client/components/checkbox/checkbox.js
```diff
@@ -3,6 +3,16 @@ import PropTypes from "prop-types";
 import { Components, registerComponent } from "@reactioncommerce/reaction-components";
 
 class Checkbox extends Component {
+  // Create a semi-random ID
+  //
+  // ID's are required for form controls and their labels for ARIA compliance.
+  // This will be the default ID for this component unless once is passed in through props.
```
kieckhafer: Can you clarify, comment is a little confusing

imports/plugins/included/social/client/components/googleplus.js
```diff
@@ -2,7 +2,7 @@ import React, { Component } from "react";
 import PropTypes from "prop-types";
 import Helmet from "react-helmet";
 import classnames from "classnames";
-import { Translation } from "/imports/plugins/core/ui/client/components";
+import { Components } from "/imports/plugins/core/components/lib";
```
kieckhafer: Should we import instead from `import { Components } from "@reactioncommerce/reaction-components";`?

imports/plugins/included/social/client/components/pinterest.js
```diff
@@ -1,7 +1,7 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
 import classnames from "classnames";
-import { Translation } from "/imports/plugins/core/ui/client/components";
+import { Components } from "/imports/plugins/core/components/lib";
```
kieckhafer: Should we import instead from `import { Components } from "@reactioncommerce/reaction-components";`?

imports/plugins/included/social/client/components/twitter.js
```diff
@@ -2,7 +2,7 @@ import React, { Component } from "react";
 import PropTypes from "prop-types";
 import Helmet from "react-helmet";
 import classnames from "classnames";
-import { Translation } from "/imports/plugins/core/ui/client/components";
+import { Components } from "/imports/plugins/core/components/lib";
```
kieckhafer: Should we import instead from `import { Components } from "@reactioncommerce/reaction-components";`?

imports/plugins/included/social/client/components/facebook.js
```diff
@@ -4,7 +4,7 @@ import PropTypes from "prop-types";
 import Helmet from "react-helmet";
 import classnames from "classnames";
 import { $ } from "meteor/jquery";
-import { Translation } from "/imports/plugins/core/ui/client/components";
+import { Components } from "/imports/plugins/core/components/lib";
```
kieckhafer: Should we import instead from `import { Components } from "@reactioncommerce/reaction-components";`?

imports/plugins/core/checkout/client/components/completedOrderItem.js
mikemurray: No, as the image is for decoration. https://www.w3.org/WAI/tutorials/images/decorative/ namely these two bullet points:

- Illustrative of adjacent text but not contributing information (“eye-candy”);
- Identified and described by surrounding text.

imports/plugins/core/orders/client/components/lineItems.js
mikemurray: No, as the image is for decoration. https://www.w3.org/WAI/tutorials/images/decorative/ namely these two bullet points:

- Illustrative of adjacent text but not contributing information (“eye-candy”);
- Identified and described by surrounding text.

imports/plugins/core/checkout/client/components/cartItems.js
mikemurray: No, as the image is for decoration. https://www.w3.org/WAI/tutorials/images/decorative/ namely these two bullet points:

- Illustrative of adjacent text but not contributing information (“eye-candy”);
- Identified and described by surrounding text.

mikemurray:
These selects are from the old blaze components. When this panel is converted to react, then it should work like all the other react switches.

image

imports/plugins/core/ui/client/components/checkbox/checkbox.js
mikemurray: When you have a form control, like an input `<input />` and you have a label for that input `<label />`. The label must have a prop `htmlFor` `<label htmlFor="idOfInputField>My Label</label>`. This then requires your input to have an ID `<input id="idOfInputField" />`. 

Thus, an ID prop is required for all of our form components, as they include labels by default. And the getter in question ensures there's at least a default.

mikemurray:
@kieckhafer 

> I cannot change any user permissions for any group. When I click (mouse or return) the switch, I see this error, but no console / server log errors:

I see the same issue, but if I select the group first, then it works. I confirmed it also happens on `master` so this is an existing issue. Can you confirm on your end? This issue is probably best addressed by this ticket: https://github.com/reactioncommerce/reaction/issues/3105.

reaction

mikemurray:
@kieckhafer 

image

The checkbox you tested is still in Blaze and should follow default HTML practices. I believe it's the `spacebar` key that you use to select the checkbox. `enter/return` in this case is for submitting a form if there is one.

The `spacebar` key is mostly used for selection like on checkboxes; `enter/return` seems to be more for buttons. See the keyboard navigation section https://webaim.org/techniques/keyboard/.

mikemurray:
@kieckhafer made a ticket for those dropdown issues: https://github.com/reactioncommerce/reaction/issues/3633

mikemurray:
@kieckhafer Ready for another review

kieckhafer:
> I see the same issue, but if I select the group first, then it works. I confirmed it also happens on master so this is an existing issue. Can you confirm on your end? This issue is probably best addressed by this ticket: #3105.

Yep, this works if the group is clicked first, will refer to #3105 on that.

PR #3585, opened 2018-01-25
Style: 3568 enable eslint import rules 

spencern:
Resolves part 2 of ?? for #3568
Enables eslint rules for
```
import/order
import/newline-after-import
import/no-duplicates
```

Breaking this PR out as some of the deduplication should be reviewed extra carefully.
The last commit https://github.com/reactioncommerce/reaction/commit/d48102013cd1d0060a29de2b71bb6d3ca3e26d85 contains all of the complex de-duplication work.

**To Test**
Make sure that the deduplicated code resolves the same as the previous implementation with duplicated imports.


imports/plugins/core/accounts/client/containers/mainDropdown.js
```diff
@@ -1,14 +1,13 @@
 import { compose, withProps } from "recompose";
-import { registerComponent, composeWithTracker, withCurrentAccount } from "@reactioncommerce/reaction-components";
 import { Meteor } from "meteor/meteor";
 import { Accounts } from "meteor/accounts-base";
 import { Roles } from "meteor/alanning:roles";
 import { Session } from "meteor/session";
-import { Reaction, Logger } from "/client/api";
-import { i18nextDep, i18next } from  "/client/api";
+import { registerComponent, composeWithTracker, withCurrentAccount } from "@reactioncommerce/reaction-components";
```
zenweasel: We probably need to figure out whether we are treating this as a local import or as an npm package as far as import order. I would support importing it as an NPM package since hopefully that's where it will end up soon. We are doing both ways now. Clearly not a huge deal but since we are making everything consistent.

spencern: This is something I've been considering lately as well. I've ordered the imports that I've changed from `@reactioncommerce/reaction-components` right after any `meteor` imports, but I haven't found a way to use the `import/order` eslint plugin to enforce that yet.

@machikoyasuda we should write the import order up in the docs as well now that it's defined a little better.

PR #3589, opened 2018-01-25
(fix): eslint rule updates (1) - PR 1 of 2

kieckhafer:
Resolves 3/4 of #3566 

`array-bracket-spacing`
`arrow-body-style`
`arrow-parens`

**Notes**

Enabling these rules has resulted in many cases where a semi-colon is removed. While it's not explicitly required, in the past we have required them internally, so perhaps something to look into.

It has also pushed a few lines over our 160 character limit, which introduced new linting errors, so I've disabled the line limit on those lines, as that seems like the less important lint.

### DIFF
--- a/imports/plugins/core/accounts/client/containers/auth.js
+++ b/imports/plugins/core/accounts/client/containers/auth.js
@@ -108,13 +108,11 @@ class AuthContainer extends Component {
     return false;
   }
 
-  formMessages = () => {
-    return (
-      <Components.LoginFormMessages
-        messages={this.state.formMessages}
-      />
-    );
-  }
+  formMessages = () => (
+    <Components.LoginFormMessages
+      messages={this.state.formMessages}
+    />
+  )
 
   services = () => {
     const serviceHelper = new ServiceConfigHelper();
@@ -131,9 +129,7 @@ class AuthContainer extends Component {
     return !!Package["accounts-password"] && enabledServices.length > 0;
   }
 
-  capitalizeName = (str) => {
-    return LoginFormSharedHelpers.capitalize(str);
-  }
+  capitalizeName = (str) => LoginFormSharedHelpers.capitalize(str)
 
   handleSocialLogin = (value) => {
     let serviceName = value;
@@ -159,9 +155,7 @@ class AuthContainer extends Component {
     });
   }
 
-  hasPasswordService = () => {
-    return !!Package["accounts-password"];
-  }
+  hasPasswordService = () => !!Package["accounts-password"]
 
   renderAuthView() {
     if (this.props.currentView === "loginFormSignInView") {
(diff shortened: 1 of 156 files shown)


imports/plugins/included/product-variant/components/productGridItems.js
```diff
@@ -66,9 +66,7 @@ class ProductGridItems extends Component {
       if (this.props.isMediumWeight()) {
         return (
           <div className={`product-additional-images ${this.renderVisible()}`}>
-            {this.props.additionalMedia().map((media) => {
-              return <span key={media._id} className="product-image" style={{ backgroundImage: `url('${media.url({ store: "medium" })}')` }} />;
-            })}
+            {this.props.additionalMedia().map((media) => <span key={media._id} className="product-image" style={{ backgroundImage: `url('${media.url({ store: "medium" })}')` }} />)} // eslint-disable-line max-len
```
nnnnat: I'm still seeing a `max-len` error for this line when I run `eslint .`

imports/plugins/core/accounts/client/containers/auth.js
```diff
@@ -159,9 +155,7 @@ class AuthContainer extends Component {
     });
   }
 
-  hasPasswordService = () => {
-    return !!Package["accounts-password"];
-  }
+  hasPasswordService = () => !!Package["accounts-password"]
```
nnnnat: Wasn't sure where we landed on the missing semicolons, are we just rolling without them for these kind of arrow func?

imports/plugins/included/product-variant/components/productGridItems.js
kieckhafer: This is fixed in Part 2 of this PR: https://github.com/reactioncommerce/reaction/pull/3578/files#diff-a25b0ba02ae71bebe4fe3574e45e01f8R69

imports/plugins/core/accounts/client/containers/auth.js
kieckhafer: @spencern @aldeed Where did we land on those semi-colons? Is this OK since it passes the linter?

spencern: I'd prefer to have them, but not sure which rule needs to be enabled to turn them on. If we don't have a rule requiring/eliminating them, it seems we're likely to end up with inconsistencies from file to file.

spencern: It looks to me like tslint and prettier both enforce semicolons here (class functions), but that eslint doesn't. I can't seem to find an option to enable them for eslint either.
https://palantir.github.io/tslint/rules/semicolon/
https://github.com/prettier/prettier/issues/1444

aldeed: @spencern I agree. Maybe don't worry about it for this issue, but you could create a ticket to pull in Prettier with just that rule enabled as a trial?

kieckhafer:
I initially broke this up into two parts, expecting this part to be merged before I was able to get the second part complete.

However, the second part is now complete, and includes the same fixes that are in this PRse same fixes, so I am closing this PR in favor of #3578.

Issue #3573, opened 2018-01-24
Enable and fix eslint rules (8)


aldeed:
Uncomment the following rules in .eslintrc, run `npm run lint`, and fix all errors, either manually or using `eslint --fix`

prefer-destructuring
space-unary-ops
template-curly-spacing

impactmass:
Other PRs have been approved/merged. Last PR to close this issue is #3610

PR #3591, opened 2018-01-25
Fix warnings after turning on eslint prefer-destructuring

impactmass:

This is the first of a group of PRs to enable `prefer-destructuring` eslint rule. This can be merged as is (when approved).

Some features to check (based on the files touched):
- `reactionApps`: This is used in a number of places, e,g the main dropdown menu, test that nothing breaks around there
- Template helper hasPermission: Used in blaze, e.g [here](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/accounts/client/templates/dashboard/dashboard.html#L2) as a wrapper around accounts dashboard template. 
- Localisation panel in dashboard, specifically units of measures
- Orders panel & invoiceContainer in Orders dashboard side panel

### DIFF
--- a/client/modules/core/helpers/apps.js
+++ b/client/modules/core/helpers/apps.js
@@ -50,7 +50,6 @@ export function Apps(optionHash) {
   let key;
   const reactionApps = [];
   let options = {};
-  let shopType;
 
   // allow for object or option.hash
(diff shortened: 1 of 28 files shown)


mikemurray:
Kicking the build process

PR #3592, opened 2018-01-25
Flatten startup code and speed up translation import

aldeed:
This PR does a few things aimed at improving development experience.

1. I reorganized a bit the files in /server/startup, removing the subfolders and moving some things into /server/startup/index.js so that it is clearer what is all happening at startup and in what order. (The order of the beforeInit/afterInit hooks is still opaque, but that is something to resolve later.)
1. Otherwise the primary changes are in /server/startup/i18n.js.
    1. I switched all the `fs.*` calls to use a promisified version of the `fs` calls with `await`, rather than `*Sync` versions. The sync calls block the Node event loop, slowing down loads.
    1. I updated `loadTranslation` function to register bulk upserts rather than actually doing the inserts synchronously, and then do a single `execute` after all packages + core have called `loadTranslation`.

I also added a missing `Reaction.Import.flush();` after the `Assets.find().forEach`, which I don't think usually caused any problems because other code later called `flush`, but it should technically be there.

Testing with METEOR_PROFILE=1, I see ~4 seconds faster reloads on average.

@zenweasel @spencern


aldeed:
Looking into the test failure. I think probably just need to increase the timeout since the `i18n/flushTranslations` method is working correctly now.

aldeed:
Should merge https://github.com/reactioncommerce/reaction/pull/3577 first, and then I can rebase this on that.

server/startup/load-data.js
```diff
@@ -20,13 +20,12 @@ export default function () {
...
 
-    // make sure the default shop has been created before going further
-    while (!Reaction.getShopId()) {
-      Logger.debug("Loading default shop, waiting until it's ready before moving on...");
-      Meteor._sleepForMs(1000);
+      // make sure the default shop has been created before going further
+      while (!Reaction.getShopId()) {
+        Logger.debug("Loading default shop, waiting until it's ready before moving on...");
+        Meteor._sleepForMs(1000);
+      }
```
aldeed: Moved this `while` to be inside the `if` because there is no need to check `Reaction.getShopId()` again right after already checking it in the `if` condition

aldeed:
@spencern The calls to `loadTranslation` are in top-level code rather than `Meteor.startup`, so the savings actually show up during the "build app" phase, and not in "server startup". I timed that independently and saw ~4 seconds drop to 25ms with my change, but then the flush now happens during the "server startup" phase (after everything has finished calling `loadTranslation`), so that actually increases startup phase by ~1 second.

So I think there is actually some minimal savings, but you're right that the cumulative effect is not very noticeable.

spencern:
Still worth doing. Did you want to rebase this with the 1.7.0 release or are we good to merge?

aldeed:
@spencern Rebased and good to merge

pmn4:
💯

aldeed:
@spencern I rebased again. I changed it to a Promise.all in i18n.js to resolve the `await` in loop. At one point I think you mentioned concerns about too much memory consumption when using Promise.all. What do you think about this situation? Should I go back to the await in loop?

spencern:
@aldeed I think in this scenario (and probably most scenarios) the Promise.all is the right way to go. I was concerned with memory consumption in our data connector plugins (Shopify et al.) where we could be importing thousands of large documents. In most cases I think that's probably not a serious concern

PR #3593, opened 2018-01-25
(refactor): 3574 fix eslint and ARIA issues for notifications

mikemurray:
Part of #3574

The notifications dropdown required some substantial changes to bring it in line with Aria compliance due to updated eslint rules.

### To Test
- As any user, add products to cart and checkout
- After checkout is completed, return the home age and check the alerts in navigation bar
- The alert drop-down should look close to the same design as it used to but now be fully tabbable and (spacebar/enter) should work on most links. The same should be true for the `/notifications` page.

**Note** Circle CI test will fail as this is a small part of the `mm-3574-eslint-9` branch and more linting issues are being worked on there. The PR merges back into that branch when approved.

### DIFF
--- a/imports/plugins/core/ui/client/components/list/listItem.js
+++ b/imports/plugins/core/ui/client/components/list/listItem.js
@@ -28,6 +28,14 @@ class ListItem extends Component {
     }
   }
 
+  handleOnKeyUp = (event) => {
+    // keyCode 32 (spacebar)
+    // keyCode 13 (enter/return)
+    if (event.keyCode === 32 || event.keyCode === 13) {
+      this.handleClick(event);
+    }
+  }
+
   handleSwitchChange = (event, isChecked, name) => {
     event.preventDefault();
     event.stopPropagation();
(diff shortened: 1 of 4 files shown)


mikemurray:
@kieckhafer can you pair with @nnnnat and @willopez and go through how we test PR's

imports/plugins/core/ui/client/components/list/listItem.js
```diff
@@ -28,6 +28,14 @@ class ListItem extends Component {
     }
   }
 
+  handleOnKeyUp = (event) => {
+    // keyCode 32 (spacebar)
+    // keyCode 13 (enter/return)
```
kieckhafer: 👍

imports/plugins/included/notifications/client/components/notificationRoute.js
```diff
@@ -43,14 +36,23 @@ class NotificationRoute extends Component {
...
   }
 
+  handleMarkAllAsRead = () => {
+    const { notificationList, markAllAsRead } = this.props;
+    markAllAsRead(notificationList);
+  }
+
   renderDropdownHead() {
-    const { notificationList, unread, markAllAsRead } = this.props;
+    const { unread } = this.props;
```
kieckhafer: @mikemurray you mentioned not to import with `{ }` if it's a single const for speed, should we turn this into just `const unread = this.props.unread;`?

impactmass: we do have a rule `prefer-destructuring` coming in #3573  that is to enforce doing `{ } ` everywhere, even for single const.

mikemurray: @kieckhafer @impactmass Since the new rules dictate this new style, I'll make the necessary adjustments when #3573 hits master.

impactmass: sure, sounds good! I just thought to add 👍

PR #3594, opened 2018-01-25
refactor: Dynamic load SMS pkgs to make server startup faster

aldeed:
When profiling Meteor, 2-3 seconds of start/restart time is spent requiring the Twilio and Nexmo packages. I moved these into dynamic imports, which means startup is faster but the first method call for each provider will be slower. This seems like a good tradeoff.

### Testing

1. Run with `METEOR_PROFILE=1 meteor` command and notice in the `Top leaves:` list at the very end that `require("twilio")` and `require("nexmo")` are no longer listed because they don't happen during startup.
1. Log in as the default admin user, or any admin user.
1. Enter a real phone number (yours) for the user on the user profile form (add default address)
1. In admin dashboard > SMS settings, choose Twilio and enter Twilio credentials and a Twilio phone number that is owned by that account and has SMS enabled.
1. In the Chrome console, enter `Meteor.call("sms/send", "TEST", Meteor.userId(), ReactionCore._shopId.get(), console.log.bind(console))`. Verify that your phone number receives the text in a few seconds.
1. In admin dashboard > SMS settings, switch to Nexmo and test that the same way.

I tested Twilio with my own account but did not test Nexmo. I'm not sure if we have some testing credentials we usually use.

### DIFF
--- a/imports/plugins/included/sms/server/methods/sms.js
+++ b/imports/plugins/included/sms/server/methods/sms.js
@@ -1,9 +1,23 @@
-import Twilio from "twilio";
-import Nexmo from "nexmo";
 import { Meteor } from "meteor/meteor";
 import { check } from "meteor/check";
 import { Sms, Accounts } from "/lib/collections";
 import { Reaction, Logger } from "/server/api";
+
+// We lazy load these in order to shave a few seconds off the time
+// it takes Meteor to start/restart the app.
+let Twilio;
+async function lazyLoadTwilio() {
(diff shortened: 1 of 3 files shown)


aldeed:
@mikemurray I added testing instructions. There was not a ticket for this specifically.

mikemurray:
@aldeed 👍

spencern:
@mikemurray what were the results of your test here?

PR #3598, opened 2018-01-26
(refactor): Fix warnings after turning on eslint prefer-destructuring

impactmass:

Some things to test (based on files touched):
- publishControls: Try publishing a produt
- tagGroups: Create tags and subtags
- dropDownMenu: Confirm that dropdowns work (e.g main dropdown)
- Sortable table pagination (e.g Email table in admin dashboard)
- Some migrations: (might have to `reaction reset` to confirm this one). Migrations should still run successfully as before
- Shopify connector: Trying importing products with [test cred](https://github.com/reactioncommerce/rc-config-files/blob/master/shopify-test-shop-creds.md)


imports/plugins/core/revisions/client/components/publishControls.js
```diff
@@ -132,7 +132,7 @@ class PublishControls extends Component {
 
   get isVisible() {
     if (Array.isArray(this.props.revisions) && this.props.revisions.length && this.primaryRevision) {
-      const primaryRevision = this.primaryRevision;
+      const primaryRevision = this.primaryRevisionObj;
```
impactmass: `const primaryRevision = this.primaryRevision;`
eslint sees this as something that should use object destructuring, but it's not, since this.primaryRevision is pointing to that method on the class.
Hence changing the method name

aldeed: It would be more backward compatible to change the variable name rather than the getter method name.

imports/plugins/core/ui/client/components/button/buttonSelect.js
```diff
@@ -29,13 +29,13 @@ class ButtonSelect extends Component {
...
   handleDefaultState = () => {
-    const props = this.props;
+    const { props } = this;
     let defaultButton = props.buttons.filter(button => {
       if (button.active === true) {
         return button;
       }
     });
-    defaultButton = defaultButton[0];
+    [defaultButton] = defaultButton;
```
aldeed: This code should actually use `props.buttons.find` instead. Not sure if you want to fix that now.

imports/plugins/core/versions/server/migrations/8_update_registry_provides_to_array.js
```diff
@@ -39,7 +39,8 @@ Migrations.add({
         // of the array
         const updatedRegistry = pkg.registry.map((entry) => {
           if (Array.isArray(entry.provides)) {
-            entry.provides = entry.provides[0];
+            const [provides] = entry.provides;
+            entry.provides = provides;
```
aldeed: This is one example of it actually being more confusing to read, and more lines, with destructuring. Maybe we just want to disable the rule for this line, and any similar lines? Thoughts @spencern ?

spencern: I'm actually more concerned with the `entry.provides[0]` to start with.

grabbing just the first item from the entry.provides array makes it seem like we need a way to set a default on the `provides` item, or that we should be selecting the default instead of the 0th index if we already have a way to identify a default.

If the 0th index is intentional and okay for now, then I'm okay with disabling.

In general, our lint rules should help code be more readable and/or more performant. If it's not doing either, that's a good candidate for using disable.

impactmass: I think it's better to revert this to previous state (and avoid confusion in reading it). And disable the rule for this line.
@spencern your concern about the [0] is actually noted in the comments above this line. There is a description about why it was written this way

impactmass:
@spencern ready for review

imports/plugins/core/router/lib/router.js
```diff
@@ -313,15 +313,13 @@ Router.reload = () => {
...
 
   if (current.route) {
-    const path = current.route.path;
-
-    if (group && group.prefix) {
-      prefix = current.route.group.prefix;
-    }
+    const { path } = current.route;
+    const { prefixArg } = group || {};
+    const prefix = prefixArg || defaultPrefix;
```
spencern: Why define `defaultPrefix` at all if it's just used here? A comment that the empty string is the default would be fine.

imports/plugins/core/router/lib/router.js
```diff
@@ -313,15 +313,13 @@ Router.reload = () => {
...
+  const defaultPrefix = "";
 
   if (current.route) {
-    const path = current.route.path;
-
-    if (group && group.prefix) {
-      prefix = current.route.group.prefix;
-    }
+    const { path } = current.route;
+    const { prefixArg } = group || {};
```
spencern: Where does `prefixArg` come from? It looks like previously it was `current.route.group.prefix`

imports/plugins/core/router/lib/router.js
spencern: I think this could be `const { prefix } = group || { prefix: "" };` and could eliminate ln 321 and 317.

imports/plugins/core/ui/client/components/button/buttonSelect.js
```diff
@@ -29,14 +29,8 @@ class ButtonSelect extends Component {
...
-    const props = this.props;
-    let defaultButton = props.buttons.filter(button => {
-      if (button.active === true) {
-        return button;
-      }
-    });
-    defaultButton = defaultButton[0];
-
+    const { props } = this;
+    const defaultButton = props.buttons.find(button => button.active);
```
spencern: Good change from filter to find. 👍

imports/plugins/included/payments-paypal/server/methods/express.js
```diff
@@ -231,8 +231,7 @@ export const methods = {
 
     const options = PayPal.expressCheckoutAccountOptions();
     const previousTransaction = _.last(paymentMethod.transactions);
-    const transactionId = previousTransaction.transactionId;
-    const currencycode = previousTransaction.currencycode;
+    const { transactionId, currencycode } = previousTransaction.transactionId;
```
spencern: this should probably just be `{ transactionId, currencycode } = previousTransaction`

imports/plugins/core/router/lib/router.js
impactmass: Thanks for spotting. I think my attempt at the refactor was not so good. Updated it now.

impactmass:
Requested changes done. Can be reviewed again.

PR #3599, opened 2018-01-26
Fix 3569 nnnnat eslint rules 4

nnnnat:
This PR corrects code style issues brought about by the new eslint rules. This PR covers the following rules:
```
no-await-in-loop
no-bitwise
no-case-declarations
no-confusing-arrow
no-empty-pattern
no-lonely-if
no-mixed-operators
no-multi-assign
no-multi-spaces
no-plusplus
no-prototype-builtins
no-tabs
```
**Testing**
- I've been following these [testing scripts](https://docs.reactioncommerce.com/reaction-docs/master/acceptance-testing)
- Running `reaction test` and `eslint .`

**Note to Reviewer**
There was a kinda strange `while` loop on line 173 of [this file](https://github.com/reactioncommerce/reaction/commit/b48746af3d598304f63fd13e5730374ec41a3bae#diff-2962a85ec68e0bc6f880768cd8262929) that I'm a little concerned about it working the same as it did before.

### DIFF
--- a/imports/plugins/core/checkout/server/methods/workflow.js
+++ b/imports/plugins/core/checkout/server/methods/workflow.js
@@ -29,7 +29,7 @@ Meteor.methods({
    * @param {String} [cartId] - cart._id
    * @return {Array|Boolean|Number} return
    */
-  "workflow/pushCartWorkflow": function (workflow, newWorkflowStatus, cartId) {
+  "workflow/pushCartWorkflow"(workflow, newWorkflowStatus, cartId) {
     check(workflow, String);
     check(newWorkflowStatus, String);
     check(cartId, Match.Optional(String));
@@ -62,14 +62,14 @@ Meteor.methods({
     });
 
     // loop through packages and set the defaultPackageWorkflows
-    packages.forEach(function (reactionPackage) {
+    packages.forEach((reactionPackage) => {
       // todo fix this hack for not filtering nicely
       if (!reactionPackage.layout.layout) {
         const layouts = _.filter(reactionPackage.layout, {
-          workflow: workflow
+          workflow
         });
         // for every layout, process the associated workflows
-        _.each(layouts, function (layout) {
+        _.each(layouts, (layout) => {
           // audience is the layout permissions
           if (typeof layout.audience !== "object") {
             const defaultRoles = Groups.findOne({
@@ -110,7 +110,7 @@ Meteor.methods({
     // loop through all shop configured layouts, and their default workflows
     // to determine what the next workflow step should be
     // the cart workflow status while processing is neither true nor false (set to template)
-    _.each(defaultPackageWorkflows, function (workflow, currentStatusIndex) {
+    _.each(defaultPackageWorkflows, (workflow, currentStatusIndex) => {
       if (workflow.template === currentWorkflowStatus) {
         // don't go past the end of the workflow
         if (currentStatusIndex < maxSteps - 1) {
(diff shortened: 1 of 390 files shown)


spencern:
Paired with @nnnnat on the `no-await-in-loop` rule to resolve issues in shopify importer and exporter. Made the decision to disable the rule for the 4 instances in the importer because of concerns that parallelizing the importers could drastically decrease server performance when importing large datasets. 

For example, the product importer paginates API calls to get products from a Shopify shop and then operates on each of those sets of products independently. Right now we fetch up to 50 products at a time. If we were to resolve using `Promise.all()` it seems that we would have to store all imported products in memory before working to resolve/create Reaction products from the imported Shopify products. For larger shops, this could result in storing thousands of imported products in memory at once vs storing up to 50 at a time right now.

@aldeed I'd be interested in your perspective on this approach and if you have any suggestions for how to do this better.

server/api/core/import.js
```diff
@@ -439,9 +439,10 @@ Import.object = function (collection, key, object) {
       $setOnInsert: defaultValuesObject
     });
   }
-  if (this._count[this._name(collection)]++ >= this._limit) {
+  if (this._count[this._name(collection)] >= this._limit) {
     this.flush(collection);
   }
+  this._count[this._name(collection)] += 1;
```
aldeed: You've translated this correctly, but I don't think it was ever being done in the correct order. The logical way to do this is to increment the count and _then_ check whether we should flush. So I would move the `+=1` line above the `if`. (This shows the dangers of using ++ in this way. Harder to notice that the logic is wrong.)

server/methods/catalog.js
```diff
@@ -398,10 +399,10 @@ Meteor.methods({
             `${variant.compareAtPrice}`
         });
       } else {
-        const parentIndex = sortedVariant.ancestors.indexOf(variantId);
+        const parentIndex = sortedVariant.ancestors.includes(variantId);
```
aldeed: Since the actual `parentIndex` value is used for the splice below, this needs to stay `indexOf`. You can change the line below to

```js
if (parentIndex >= 0) ancestorsClone.splice(parentIndex, 1, variantNewId);
```

aldeed:
@spencern @nnnnat I agree with disabling the await rule. I left a couple other comments about some issues. Otherwise looks good.

imports/plugins/core/checkout/server/methods/workflow.js
```diff
@@ -237,7 +237,7 @@ Meteor.methods({
     // get index of `newWorkflowStatus`
     const resetToIndex = workflow.indexOf(newWorkflowStatus);
     // exit if no such step in workflow
-    if (!~resetToIndex) return false;
+    if (resetToIndex < 0) return false;
```
nnnnat: @aldeed I've updated this file with your suggestions. I've returned the `workflow.includes` to `workflow.indexOf` and updated the `if` on the next line

server/api/core/import.js
nnnnat: I've updated this in [this commit](https://github.com/reactioncommerce/reaction/pull/3599/commits/5685d3e6e084af7bd4b66dd7ed3ec7668cf87885)

server/methods/catalog.js
nnnnat: @aldeed I believe this [commit](https://github.com/reactioncommerce/reaction/pull/3599/commits/d7c26811445f9c6f6e40d21be19fd4c179d60958) is fixing the issue.

nnnnat:
I've updated this PR with all of @aldeed revisions, let me know if I missed anything or misunderstood the revisions.

impactmass:
@nnnnat, looks like the 1.7 merge just introduced some conflicts here

aldeed:
@nnnnat Changes look good. OK to merge after you rebase on master and resolve the conflicts.

nnnnat:
Conflict should be fixed in latest commit

Issue #3034, opened 2017-10-04
Deprecate aldeed:meteor-autoform


aaronjudd:
Deprecate https://github.com/aldeed/meteor-autoform, all validation should be done in React components.

### Expected behavior
No use of blaze templates or autoform in core code or UI components.

aaronjudd:
See related: 2917

spencern:
Looks like autoform also contains a dependency on the `momentjs:moment` and `mrt:moment-timezones` meteor packages.

![image](https://user-images.githubusercontent.com/1203639/35599984-c58f691e-05e8-11e8-9c1d-5d4e1f76f89d.png)

aldeed:
@spencern `mrt:moment-timezone` is a weak dep, which just means load it before the autoform pkg if the user has added it to their app. It shouldn't pull it in automatically. I could probably update autoform to get rid of the `momentjs:moment` strong dep, but I'd rather put my effort into getting rid of autoform entirely. :)

Although maybe I'll do both, if updating autoform looks to be a quick thing. It's been in the back of my mind to do that for awhile.

spencern:
@aldeed sounds good. 
I'm okay just focusing on the removal of autoform, but if it's quick to remove the moment dep, then 👍

willopez:
This looks like a good replacement: https://github.com/jaredpalmer/formik

aldeed:
@willopez Thanks for the link, I hadn't seen that one. I am planning to use and improve my [reacto-form](http://forms.dairystatedesigns.com/reacto-form/) package, which actually has similar goals to formik, but a completely different way of achieving. To me the ReactoForm way makes more sense, but I would have to pull down and play with formik to say for sure. I'm interested to know your opinion if you have one. I've also used ReduxForms, but I don't think tight coupling with redux at that level is the correct approach.

willopez:
@aldeed I have not used reacto-form myself, so I am not qualified to give an opinion on that package. I would consider the time and effort it would take prepare reacto-form to be used, plus the integration with reaction, against the time and effort necessary to integrate formik. I have played around with formik and to me it seems like a solid choice.

danilomiranda:
Hey @spencern ,
What lib will replace aldeed:meteor-autoform?

PR #3602, opened 2018-01-26
(refactor): Dynamically import moment.js

### LINKED ISSUE #2635: Dynamically load moment and moment-timezone modules
Reduce application startup load by dynamically importing `moment` and `moment-timezone` only where necessary.  It is mostly used in the various admin dashboards, but also on client order views.

Suggest reviewing usage, I'd suggest that there are some possible lighter weight replacements as well. 
`moment-timezone` is also used to populate some client side data, such as timezones, locales.

### Expected behavior

Dates are formatted.  ~400k saved from initial load bundle.

kieckhafer:
Resolves #2635 - the Reaction codebase portion of this ticket.

This ticket DOES NOT update any npm or meteor packages that use `moment` or `moment-timezone`.
- `autoform` is the only `meteor` package that uses `moment` ( #3034 ).
- multiple `npm` packages use `moment` ( #3621 )

This ticket updates all files to dynamically import `moment.js` or `moment-timezone.js` in Reaction code.

- React components use the new `withMoment` HOC to wrap the component
- Blaze / raw JS files import moment using `async`/`await` in each file

If a file or function that imported `moment.js` is no longer used (i.e. old Blaze templates and helpers), it's been removed entirely, which might cause seom BREAKING CHANGES:

- Removed `timezoneOptions` Blaze template helper from `client/modules/core/helpers/templates.js`, as it's no longer used in any core files.
- Removed `ordersList`, `orderPage/details` and `orderPage` Blaze templates, which were replaced by React templates
- Moved `dateFormat` Blaze template helper out of the global helpers, and into a specific template helper, since it's only used in one place


imports/plugins/core/accounts/client/components/groupsTableCell.js
```diff
@@ -2,12 +2,21 @@ import { Meteor } from "meteor/meteor";
...
 import _ from "lodash";
 import PropTypes from "prop-types";
-import moment from "moment";
-import { Components, registerComponent } from "@reactioncommerce/reaction-components";
+import { Components, registerComponent, withMoment } from "@reactioncommerce/reaction-components";
 import { Reaction } from "/client/api";
 import { getUserAvatar } from "/imports/plugins/core/accounts/client/helpers/helpers";
 
-const GroupsTableCell = ({ account, columnName, group, adminGroups, handleRemoveUserFromGroup, handleUserGroupChange, ...props }) => {
+const GroupsTableCell = (props) => {
```
pmn4: 💯

pmn4:
lgtm

spencern:
FYI @kieckhafer - it looks like https://github.com/reactioncommerce/reaction/issues/3034 will be needed to fully remove any moment globals.

zenweasel:
Doing some quick testing I am getting this warning in the console when I login (multiple instances, just pasted one here):
```
Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the lifecycle(Reaction(NotificationDropdown)) component.
```

zenweasel:
Getting this error when I go to the "Print an Invoice" page
```

meteor.js?hash=cbcc712d51de4298c275e8dcf25c66c29914f19a:992 Exception in template helper: ReferenceError: moment is not defined
    at Object.<anonymous> (http://localhost:3000/app/app.js?hash=be27a142ba5e6b2a8df3558d6a5b89c4a6ef8e7d:121512:3)
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3051:16
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:1715:16
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3103:66
    at Function.Template._withTemplateInstanceFunc (http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3744:12)
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3102:27
    at Spacebars.call (http://localhost:3000/packages/spacebars.js?hash=ebf9381e7fc625d41acb0df14995b7614360858a:172:18)
    at Spacebars.mustacheImpl (http://localhost:3000/packages/spacebars.js?hash=ebf9381e7fc625d41acb0df14995b7614360858a:106:25)
    at Object.Spacebars.mustache (http://localhost:3000/packages/spacebars.js?hash=ebf9381e7fc625d41acb0df14995b7614360858a:110:39)
    at Blaze.View._render (http://localhost:3000/app/app.js?hash=be27a142ba5e6b2a8df3558d6a5b89c4a6ef8e7d:62393:22)
```

zenweasel:
Get this warning when I am checking out anonymously and try to register. Weird that it's the `ProductDetailContainer` but....there it is
```
Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the ProductDetailContainer component.
```

zenweasel:
Get this error in the server console when I go to checkout as a registered/non-admin user
```
meteor.js?hash=cbcc712d51de4298c275e8dcf25c66c29914f19a:992 Exception in template helper: TypeError: Cannot read property 'items' of undefined
    at Object.cartCount (http://localhost:3000/app/app.js?hash=be27a142ba5e6b2a8df3558d6a5b89c4a6ef8e7d:51318:16)
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3051:16
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:1715:16
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3103:66
    at Function.Template._withTemplateInstanceFunc (http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3744:12)
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3102:27
    at Object.Spacebars.call (http://localhost:3000/packages/spacebars.js?hash=ebf9381e7fc625d41acb0df14995b7614360858a:172:18)
    at http://localhost:3000/app/app.js?hash=be27a142ba5e6b2a8df3558d6a5b89c4a6ef8e7d:51133:22
    at Blaze.View.<anonymous> (http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:2734:44)
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:1934:20
```

client/modules/i18n/main.js
```diff
@@ -107,7 +106,9 @@ Meteor.startup(() => {
         if (result) {
           const locale = result;
           locale.language = getBrowserLanguage();
-          moment.locale(locale.language);
+          import("moment").then(moment => {
+            moment.locale(locale.language);
+          });
```
aldeed: This is going to have the effect of loading moment right away during the page load, anyway. Won't be in the main bundle, but it still seems like we could get rid of this so that there is less chance of it ever having to load for some people.

All this is doing is setting the locale for moment to use. So that is something we can do right after we load moment at some later point in time.

Therefore, I would add a file somewhere that exports an `importMoment` function like:

```js
export default async function importMoment() {
  const moment = await import("moment");
  moment.locale(locale.language); // get the locale from somewhere where this "shop/getLocale" callback has saved it
  return moment;
}
```

And then change all other `import("moment")` to use our custom `importMoment()`

imports/plugins/core/components/lib/hoc.js
```diff
@@ -26,6 +32,56 @@ export function withCurrentUser(component) {
...
+    componentDidMount() {
+      try {
+        import("moment").then(moment => {
+          this.setState({
+            moment: moment.default
+          });
+        });
+      } catch (error) {
+        Logger.debug(error, "moment.js async import error");
+      }
```
aldeed: This `catch` will not catch errors because they will get swallowed by the promise. You should catch the promise instead:

```js
import("moment")
  .then(...)
  .catch((error) => { Logger.debug(error, "moment.js async import error"); });
```

Or you can `await` the import line and then `try/catch` will work.

imports/plugins/core/components/lib/hoc.js
```diff
@@ -26,6 +32,56 @@ export function withCurrentUser(component) {
...
+ * @memberof Components
+ */
+export function withMoment(component) {
+  return lifecycle({
+    componentDidMount() {
+      try {
+        import("moment").then(moment => {
+          this.setState({
+            moment: moment.default
+          });
```
aldeed: Whenever you call `setState` in a callback, you risk getting the errors that @zenweasel mentioned if the component has already been destroyed before the callback/then got called.

The way I've dealt with this in the past is to set, unset, and check an `isMounted` flag. Some people recommend using a cancelable promise and canceling it when the component dismounts. See answers here: https://stackoverflow.com/questions/39767482/is-there-a-way-to-check-if-the-react-component-is-unmounted

imports/plugins/core/i18n/client/containers/localizationSettings.js
```diff
@@ -135,12 +134,15 @@ function composer(props, onData) {
...
-    });
+  const moment = props.momentTimezone;
+  if (moment) {
+    const timezones = moment.names();
+    for (const timezone of timezones) {
+      timezoneOptions.push({
+        value: timezone,
+        label: timezone
+      });
+    }
```
aldeed: It's worth considering that rather than hit the server and bring down all of moment-timezone library, we could just make a method that returns this array and `Meteor.call` that?

client/modules/i18n/main.js
kieckhafer: I had a similar initial thought, placing a `getMoment()` in a central location, and ended up with a function inside of `Reaction.`, so we could call `Reaction.getMoment()`, but after chatting to a few people realized that would be called of the bat when `Reaction.` was built, and went this other direction.

Where's a good location for this `importMoment()` function to live? `/lib/api/helpers`

imports/plugins/core/components/lib/hoc.js
kieckhafer: Which is better practice / should we be using? Promieses or async / await?

client/modules/i18n/main.js
aldeed: I think it would be needed only in client code (`/lib` loads on server too), and also I'm trying to insist that all new code go within `/imports` so that it does not load anywhere unless we import it there. I'm a little unsure about what pattern we want to use, given the current idea of "plugin". @aaronjudd @spencern ? If we have functions we want to have available for import in other files, but not really a plugin or something that needs `registerPackage`, where should those go? Can we make `/imports/helpers`?

If we stick with the pattern of putting all the helpers on the `Reaction` object, then I think we'll end up loading more code than we need to in some cases. That's why I'm thinking we just stick them somewhere in `/imports` and import directly where needed.

imports/plugins/core/components/lib/hoc.js
aldeed: In many cases I'd say async / await, but something like this... I'm not sure that adding `async` to a React instance method would be a good idea. Maybe it would work, but I'd lean toward the safety of promise+then.

client/modules/i18n/main.js
spencern: I agree that moving all code to `/imports` is the direction we should be going. I'm okay with some `helpers` living in their own package. I can't see any reasons that getting the moment lib should be attached to the `Reaction` object right now.

From what I remember, using the Reaction object for helpers like this in the past was because the previous pattern (pre `import`) was that we had several globals, so moving them to the `Reaction` namespace was a way of cleaning that up.

imports/plugins/core/components/lib/hoc.js
kieckhafer: Won't this cause the data (`moment`) to be unavailable entirely, if the promise is cancelled when the component is unmounted, if the data isn't loaded yet?

kieckhafer:
@zenweasel Can you re-check this for your `errors` only, (PDF page not importing moment). We are aware the `setState` warnings are still here, but those might be resolved by @mikemurray's #3582, so we are going to wait until that's pulled in to deal with those.

zenweasel:
@kieckhafer I am not getting the missing moment error any longer on the print invoice page but now I am getting: 

```
Exception in template helper: TypeError: Cannot read property '0' of undefined
    at Object.billing (http://localhost:3000/app/app.js?hash=9243d2f487f2200898d6da7082550c853f7a9091:62981:29)
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3051:16
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:1715:16
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3103:66
    at Function.Template._withTemplateInstanceFunc (http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3744:12)
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:3102:27
    at Object.Spacebars.dot (http://localhost:3000/packages/spacebars.js?hash=ebf9381e7fc625d41acb0df14995b7614360858a:234:13)
    at http://localhost:3000/app/app.js?hash=9243d2f487f2200898d6da7082550c853f7a9091:62476:37
    at Blaze.View.<anonymous> (http://localhost:3000/packages/spacebars.js?hash=ebf9381e7fc625d41acb0df14995b7614360858a:264:18)
    at http://localhost:3000/packages/blaze.js?hash=f33d3dfed63a491d24e3aa07ad66c24b5fe8c761:1934:20
```

kieckhafer:
@zenweasel that should be fixed, please take a look again when you can.

kieckhafer:
@aldeed moved the `locale` set so it doesn't run on startup, I think that was the only blocker from you.

zenweasel:
Approved this with the caveat that I am still seeing the `setState` errors everywhere

aldeed:
@zenweasel setState should only be warnings, and it is a little bit bigger project to resolve those properly, so I said it would be fine to merge. @kieckhafer will make a new issue for resolving those warnings.

zenweasel:
@aldeed Yeah, that's why I went ahead and approved it.

kieckhafer:
New issue here: https://github.com/reactioncommerce/reaction/issues/3654

machikoyasuda:
@kieckhafer - Added a ticket for you in Docs to document this change https://github.com/reactioncommerce/reaction-docs/issues/494

spencern:
Disconnecting because this issue does not resolve #3757

PR #3606, opened 2018-01-27
(fix): Add permission checks to template method and publication

spencern:
Adds permission checks to the `template/email/update` server method.

I've tested and verified that as an admin you can still view and modify email templates.
I've also tested and verified that users without the `reaction-templates` role for the primary shop cannot view or subscribe to the templates, and cannot modify them.

We currently _do not_ use shop specific templates, the default marketplace install of Reaction disables template editing for shops and uses the primary shop's templates for all emails. I've added this code to be compatible with marketplace shops, but it's not enabled right now anyway, and there's not a great way to test that.

**To test:**
1. Verify that as the admin user or user with the `reaction-templates` role you can still view and modify templates.
2. As an unauthenticated user. (incognito window works)
3. Attempt to subscribe to the templates publication, verify that you are not sent any template documents.
4. Attempt to modify a template by calling the `template/email/update` method from the client console. Verify that you are not able to do so successfully.

image

### DIFF
--- a/imports/plugins/core/templates/server/methods.js
+++ b/imports/plugins/core/templates/server/methods.js
@@ -1,5 +1,6 @@
 import { Meteor } from "meteor/meteor";
 import { check } from "meteor/check";
+import { Reaction } from "/server/api";
 import { Templates } from "/lib/collections";
 
 /**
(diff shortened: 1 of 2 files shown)


spencern:
@aaronjudd good catch. Pretty sloppy of me not to remember that we use templates for the PDP as well. Fixed by permitting all users to subscribe to templates (previous functionality) but keeping the role check in the update method.

imports/plugins/core/templates/server/methods.js
```diff
@@ -23,13 +24,19 @@ export const methods = {
...
+    // Check that this user has permission to update templates for the active shop
+    if (!Reaction.hasPermission("reaction-templates", userId, shopId)) {
+      throw new Meteor.Error("access-denied", "Access Denied");
+    }
+
     return Templates.update({
       _id: templateId,
-      type: "email"
+      type: "email",
+      shopId: shopId // Ensure that the template we're attempting to update is owned by the active shop.
```
pmn4: 💯

pmn4:
lgtm

zenweasel:
As a shop admin I am able to execute the method for templates that are not mine but the method uses the active shop ID so I am prevented from actually doing anything harmful.

Issue #3608, opened 2018-01-29
ProductField is very difficult to target via CSS because of missing class name.


prinzdezibel:
### Expected behavior
It should be straightforward to overwrite the CSS styles of PDP fields like title, vendor or description through using class selectors. This is not possible in a straightforward manner. One could use CSS selectors like `.pdp.field:nth-child(n)`, but these would be very brittle.

### Actual behavior
No field specific class name is added to generic ProductFields in markup.

### Steps to reproduce the behavior
Go to page PDP. See HTML source of page.

![developer_tools_-_http___localhost_3000_product_reaction-mug_and_reaction_mug](https://user-images.githubusercontent.com/1733229/35507813-55b5fcfc-04ee-11e8-97fa-cba90820c79d.jpg)

zenweasel:
Closed via #3609

PR #3609, opened 2018-01-29
Add CSS class to generic product fields

prinzdezibel:
This PR resolves issue #3608 

It adds an additional class name "fieldname" to the existing ones (pdp, field).

To test:
- Navigate to PDP page
- Inspect the title field with Chrome inspector
- HTML source should contain the field name for fields like title, vendor, description, etc.

developer_tools_-_http___localhost_3000_product_reaction-mug_and_reaction_mug

### DIFF
--- a/imports/plugins/included/product-detail-simple/client/components/productField.js
+++ b/imports/plugins/included/product-detail-simple/client/components/productField.js
@@ -133,7 +133,7 @@ class ProductField extends Component {
     const classNames = classnames({
       pdp: true,
       field: true,
-      [this.fieldName]: this.fieldName
+      [this.fieldName]: !!this.fieldName
     });
 
     if (this.props.element) {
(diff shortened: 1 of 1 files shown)


imports/plugins/included/product-detail-simple/client/components/productField.js
```diff
@@ -130,16 +130,22 @@ class ProductField extends Component {
       return this.renderTextField();
     }
 
+    const classNames = classnames({
+      pdp: true,
+      field: true,
+      [this.fieldName]: this.fieldName
```
mikemurray: This should be a Boolean `true` or `false` however you decide to calculate that, or literal `true` or `false`.

`[this.fieldName]:  true|false|typeof this.fieldName === "string"|whatever`

prinzdezibel: you mean something like this?

```
[this.fieldName]: !!this.fieldName
```

prinzdezibel: @mikemurray  I did this, because even
```
classnames({a: true, undefined})
```
yields the correct class names. And I think we can't guarantee that this.fieldName is given (that comes from template definition, right)?!

PR #3610, opened 2018-01-29
(refactor): Enable eslint prefer-destructuring

impactmass:
To be merged after #3591 and #3598.

Some things to test (based on files touched):
- PDP (product should load properly, also images)
- Update/Add address info on profile page
- Orders approve/capture process


pmn4:
lgtm

imports/plugins/included/product-detail-simple/client/containers/social.js
```diff
@@ -27,11 +27,11 @@ class ProductSocialContainer extends Component {
 function composer(props, onData) {
   const product = ReactionProduct.selectedProduct() || {};
   const selectedVariant = ReactionProduct.selectedVariant() || {};
-  let title = product.title;
+  let { title } = product;
   let mediaUrl;
 
   if (selectedVariant) {
-    title = selectedVariant.title;
+    ({ title } = selectedVariant);
```
spencern: @impactmass 
While I understand why we're doing this, I think this pattern will need a bit of documentation.

imports/plugins/included/sms/client/components/smsSettings.js
```diff
@@ -48,8 +48,7 @@ class SmsSettings extends Component {
 
 
   render() {
-    const settings = this.state.settings;
-    const isSaving = this.state.isSaving;
+    const { settings, isSaving } = this.state.settings;
```
spencern: this should be just `const { settings, isSaving } = this.state`

imports/plugins/included/sms/server/methods/sms.js
```diff
@@ -68,7 +68,8 @@ Meteor.methods({
     // check for addressBook phone
     if (user && addressBook) {
       if (addressBook[0].phone) {
-        phone = addressBook[0].phone;
+        const [phoneData] = addressBook;
```
spencern: This borders on less readable.. I think I'll leave this one for now, but not sure I like where this is going.

imports/plugins/included/social/client/components/twitter.js
```diff
@@ -9,8 +9,7 @@ export function getTwitterMeta(props) {
   const title = props.title || document.title;
   const preferredUrl = props.url || location.origin + location.pathname;
   const url = encodeURIComponent(preferredUrl);
-  const username = props.settings.username;
-  const description = props.settings.description;
+  const { username, description } = props.settings.username;
```
spencern: Should be `const { username, description } = props.settings`

server/methods/accounts/accounts.js
```diff
@@ -156,13 +156,13 @@ function getValidator() {
   let geoCoder;
   // Just one?, use that one
   if (geoCoders.length === 1) {
-    geoCoder = geoCoders[0];
+    [geoCoder] = geoCoders;
   }
   // If there are two, we default to the one that is not the Reaction one
   if (geoCoders.length === 2) {
-    geoCoder = _.filter(geoCoders, (coder) => {
+    [geoCoder] = _.filter(geoCoders, (coder) => {
```
spencern: should be a find instead, should use Array.prototype.find and not lodash

server/methods/catalog.app-test.js
```diff
@@ -183,7 +183,7 @@ describe("core product methods", function () {
       variant["title"] = "Updated Title";
       variant["price"] = 7;
       Meteor.call("products/updateVariant", variant);
-      variant = Products.find({ ancestors: [product._id] }).fetch()[0];
+      [variant] = Products.find({ ancestors: [product._id] }).fetch();
```
spencern: should be a findOne instead of `find().fetch()[0]`

server/methods/core/methods.app-test.js
```diff
@@ -113,7 +113,7 @@ describe("Server/Core", function () {
       tag = Factory.create("tag");
       Meteor.call("shop/updateHeaderTags", "updated tag", tag._id);
       expect(Tags.find().count()).to.equal(1);
-      tag = Tags.find().fetch()[0];
+      [tag] = Tags.find().fetch();
```
spencern: should be `findOne`

server/methods/core/orders.js
```diff
@@ -306,7 +306,7 @@ export const methods = {
    */
   "orders/approvePayment"(order) {
     check(order, Object);
-    const invoice = orderCreditMethod(order).invoice;
+    const { invoice } = orderCreditMethod(order);
```
spencern: stuff like this feels like it might need a guard, but that's beyond this PR

server/methods/core/orders.js
```diff
@@ -320,8 +320,7 @@ export const methods = {
     // that the math all still adds up.
     const shopId = Reaction.getShopId();
     const subTotal = invoice.subtotal;
-    const shipping = invoice.shipping;
-    const taxes = invoice.taxes;
+    const { shipping, taxes } = invoice;
```
spencern: why not do subTotal and discounts here as well?

imports/plugins/included/sms/server/methods/sms.js
impactmass: Proposing an alternative here ... #3644

imports/plugins/included/product-detail-simple/client/containers/social.js
impactmass: here's the pattern:
```
let a = "" // some default value

if (ref && ref.something) { // some of the checks here makes it not easy to refactor the whole block
  a = ref.x.a
}
```

PR #3611, opened 2018-01-29
refactor: import Reaction from /client|server|lib/api when possible

spencern:
Simple refactor of `import { Reaction }` statements that were using a relative path instead of `/client/api` so that our Reaction imports are consistently styled.

**To Test**
1. Make sure the app starts
2. Layouts should load
3. Templates should load
4. `hasPermissions` works


zenweasel:
@spencern Thanks for opening this PR

The test instructions are a little _broad_ but a basic test of functionality seems to be working

Issue #2858, opened 2017-09-14
Docker build from tagged branches 


aaronjudd:
When creating a tagged release (commit to master, making a new release using GitHub) our CI workflow should automatically push the successful build to Docker Hub.  I really don't think there is any need to have this logic be special to `master` or any other branch. Just any time we release/tag a branch we should build, push and label the image based on the tag name.  This might not be great if we're pulling `latest` in production, but that's not something I'd suggest we/people rely on anyways..  ie using `latest` is akin to always living in alpha state.

aaronjudd:
Maybe this would also imply.. that we get rid of `prequel`?

aaronjudd:
Not sure if it's helpful, but a "tag" https://developer.github.com/v3/activity/events/types/#createevent is different from a "release". https://developer.github.com/v3/activity/events/types/#releaseevent

aaronjudd:
CircleCI in theory should support tags now (Aug 1).

https://circleci.com/changelog/#circleci-2-0-git-tags-support

`CircleCI will not run a job for a Git tag unless a tags filter is specified.

https://circleci.com/docs/2.0/configuration-reference/#tags

spencern:
@jshimko any update here?

jshimko:
The short answer is we either need to make sure to add the version tag to the merge commit once it's on the master branch or we just manually build/push the image after the fact like I did last time.  Either is an easy solution until I have time to refactor the CI config to use all of their new features.

The longer answer is I need to refactor the CI config to make use of Circle's new [workflows](https://circleci.com/docs/2.0/workflows/) feature (which supports parallel jobs, tag regex's, etc.).  I have that partially finished on a local branch, but I'm trying to get unburied with other more pressing blockers before I go down that rabbit hole for half a day.

spencern:
@aaronjudd is this a blocker for our `1.5.0` release still?

aaronjudd:
@impactmass Let's discuss, this might be a project you could take a look at...

aldeed:
I have set up / worked on a few apps using circle workflows. I am not yet up to speed on how the current reaction build is set up, but I might be of some minimal help if you get stuck on the workflows config

impactmass:
Started working on this with help from @aldeed. Approach is to:
- [x] Update our current circleci config.yml to use the workflow in v2 (done)
- [ ] Update the deploy job to use filters tag, to achieve this issue's goal
- [ ] Test & PR

[...]

PR #3618, opened 2018-01-31
(refactor): Update circleci config to circleci workflowsv2

impactmass:
Refactor circleci config.

Part of #2858. 

This PR updates:
1. the circleci config.yml to use workflows which enables:
    - faster builds running jobs within a workflow independently
    - faster re-runs by allowing re-runs on each job of a workflow
2. updates the test node image to a `-browsers` image that contains headless Chrome available for running client tests

This PR adds `test_publish_deploy` workflow with three jobs (which we already had before):
- `tests`: which runs the app tests 
- `publish` - builds docker image and publish
- `deployDocs` - runs jsdoc and deploys to aws


.circleci/config.yml
```diff
@@ -1,11 +1,12 @@
...
 jobs:
-  build:
-    working_directory: /home/reaction
+  # run automated tests
+  tests:
+    working_directory: ~/reaction
 
     docker:
-      - image: node:8
+      - image: circleci/node:8.9.3-browsers
```
impactmass: Changing this per comment made [here](https://github.com/reactioncommerce/reaction/issues/2858#issuecomment-360261622)

.circleci/config.yml
```diff
@@ -14,63 +15,134 @@ jobs:
...
+      # Install Meteor
+      - run: .circleci/install_meteor.sh
 
-      - save_cache:
-          name: Saving Meteor to cache
-          key: meteor
-          paths:
-            - ~/.meteor
+      # install OS dependencies
+      - run: sudo apt-get update
```
impactmass: Added `sudo` here (and a couple other places) because the new node image throws permission denied errors on those commands. If there's a better way to this please point me to it...

.circleci/config.yml
```diff
@@ -14,63 +15,134 @@ jobs:
...
       - run: meteor npm install
 
-      # run tests
       - restore_cache:
           name: Restoring Meteor dev_bundle cache
-          key: dev_bundle
+          key: dev_bundle-{{ .Branch }}
+
+      # install Reaction CLI
+      - run: sudo npm install -g reaction-cli
```
impactmass: This was done previously with `yarn`. I switched to `npm` here (for consistency), but I'll like to know if there's reason why yarn is preferred.

.circleci/config.yml
```diff
@@ -1,11 +1,12 @@
 version: 2
 
 jobs:
-  build:
-    working_directory: /home/reaction
+  # run automated tests
+  tests:
+    working_directory: ~/reaction
```
impactmass: Changed from `/home/reaction` to `~/reaction`. The `8.9.3-browsers` image throws permission denied error on root folder. Any better way?

.circleci/config.yml
```diff
@@ -14,63 +15,134 @@ jobs:
...
 
     steps:
-      - setup_remote_docker
       - checkout
 
-      # install OS dependencies
       - restore_cache:
           name: Restoring Meteor cache
-          key: meteor
+          key: meteor-{{ .Branch }}
```
impactmass: Added branch name to the keys for the cache primarily because the old cache that was generated on previous node:8 image is to be linked/restored to root locations on the new 8.9.3 (which also resulted in permission denied errors).

.circleci/config.yml
```diff
@@ -14,63 +15,134 @@ jobs:
...
+
+      - save_cache:
+          name: Saving Meteor to cache
+          key: meteor-{{ .Branch }}
+          paths:
+            - ~/.meteor
+
+
+  # build a Reaction image and publish to DockerHub
+  publish:
```
impactmass: I haven't tested the publish & deploy jobs yet. Will do after approach/discussion on the test job is approved.

.circleci/config.yml
```diff
@@ -14,63 +15,134 @@ jobs:
...
           name: Saving Meteor dev_bundle to cache
-          key: dev_bundle
+          key: dev_bundle-{{ .Branch }}
           paths:
-            - /home/reaction/.meteor/local
+            - ~/reaction/.meteor/local
+
+      - save_cache:
+          paths:
+            - ~/node_modules
```
jshimko: That's not where the app's `node_modules` are.  It's `~/reaction/node_modules`

.circleci/config.yml
ticean: Hi @impactmass. This is a minor suggestion because I'm sure that this will work as-is. `~` is expanded to home by the shell when it's not prefixed. But `~` is a typing convenience for the command line. In scripts it's generally preferred to use the `$HOME` var.

See:

* [SO: Difference between home and tilde](https://stackoverflow.com/questions/11587343/difference-between-home-and-tilde)
* [SO: Why use home over tilde in a shell script](https://stackoverflow.com/questions/5930671/why-use-home-over-tilde-in-a-shell-script)
* [Bash Tilde Expansion Documentation](https://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html).

---

That said, you did the right thing. The old path didn't work on that image because the user `reaction` doesn't exist in `circleci/node:8.9.3-browsers` and therefore the `/home/reaction` directory doesn't exist either.

impactmass: Thanks. I changed to $HOME. With that circleci fails on the checkout step. 
`Unable to create directory '$HOME/reaction`. Not sure why yet. 

Also, I did not get this part `The old path didn't work on that image because the user `reaction` doesn't exist ... and therefore the /home/reaction`. Were we creating a `reaction` user in the previous setup?

jshimko: There was no `reaction` user.  That was just the directory I picked to checkout the code within the image.  `~/` definitely was a problem previously though, so that's why I explicitly set it to `/home/`.

impactmass:
@jshimko please do a full review of the changes to config.yml and install.sh (the other commits are from a PR #3629 which is already in release-1.8.0). If this is all good, it can go with release-1.8.0, otherwise I will attend to requested changes when I'm back.

zenweasel:
@impactmass Do you still need @jshimko to look at this or are you working with @ticean on it?

impactmass:
closing this, will be opened as part of the linked issue above

Issue #3621, opened 2018-01-31
Audit and fix / replace any npm or meteor packages which use moment or moment-timezone


kieckhafer:
PR #3602 dynamically import moment and moment-timezone in all Reaction owned filed.

This ticket is to audit and discuss solutions for `npm` and `meteor` packages which use either `moment` or `moment-timezone`.

### Meteor
`aldeed:meteor-autoform`: Ticket #3034 is already removing this package

### npm
`bunyan` -> `moment`
`joi` -> `moment`
`node-loggly-bulk` -> `moment`
`react-dates` -> `react-moment-proptypes` -> `moment`
`twilio` -> `moment`

kieckhafer:
@spencern @aldeed 

I spoke with @mikemurray about `react-dates`, and it's used only in our calendar picker. We can use `react-loadable` to dynamically import the Calendar Picker component where it's needed.

Thoughts on the others?

spencern:
If we're importing the packages that include `moment` dynamically (e.g. Twilio), I would imagine that is enough to not load moment globally.

`joi` is not a top level dependency for us, and is included with `Nexmo` which I think we are now dynamically importing.

It looks like we use `bunyan` for our client logger, but perhaps we can rethink that choice or only import for certain `env`s 

`loggly` is used for our server logs, and we could try to find a logger that doesn't use moment, or we can just eat the cost on the server, for now. I'm definitely more concerned with moment on the client.

kieckhafer:
Yes, Twilio and Nexmo are both loaded dynamically now via #3594 

So `bunyan` is the main culprit, and then `loggly` if we want it off the server as well?

aaronjudd:
Loggly is  / should be an optional tool, perhaps you could avoid loading on client, use on server, and only if the ENV vars have been configured.  Bunyan should be available both client and server, but we could easily add a client logger that just sent the message to the server installation that feeds the client log data to a server-side Bunyan, thus removing the requirement for the client?   At this point though, bunyan remains a core requirement (but I could see us transferring responsibility to a service once that part of our next architecture is up and running)

aaronjudd:
You may also want to consider http://moment.github.io/luxon/ instead of the full moment on client.

kieckhafer:
@aaronjudd I've already got moment to load dynamically on everything we can control: #3602 , was just wondering about our npm and meteor packages here.

Do you think luxon is a better choice in the long run for our code?

aldeed:
[...]

PR #3622, opened 2018-01-31
(refactor): dynamically import DayPickerRangeController

kieckhafer:
Resolves portions of #3621 

As part of an effort to dynamically load `moment`, we should find a way to dynamically load npm packages that have moment in their own dependencies.

This PR adds the `react-loadable` npm package, which allows us to dynamically import React components.

It then uses `react-loadable` to dynamically import `react-dates` (which uses `moment`) and it's `DayPickerRangeController`, which we use for our calendar picker on our orders page.

**Testing**
- Create one or more orders
- As an admin, go to the Orders dashboard, and start to filter orders by date. See the loading component on the calendar dropdown while `react-dates` dynamically loads, and then see the calendar after it's loaded. The component will cache, so hard refresh if you need to test again.

**Notes**
We need to disable an eslint-rule, `react/no-multi-comp` on this file, to allow multiple React components to be declared.


kieckhafer:
@spencern @aldeed 

I've disabled an eslint-rule to make this PR work. Looking through the docs of this particular rule, if we are going to dynamically load React components frequently, it might benefit us to change the rule `react/no-multi-comp` to use the option `ignoreStateless`, which should solve the issue here and in the future.

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-multi-comp.md#ignorestateless

aldeed:
@kieckhafer Actually that is a rule that I personally like to stick to, and I think it's beneficial to be strict about it, even for this situation. It helps keep things understandable, and as we move components out into a separate repo and have better tests, snapshots, etc. for them, a one-to-one mapping will be helpful for that, too.

So in this particular case, I would not disable the rule, and would instead make a new folder/file in the `components` folder `DayPickerRangeController.js`, which should export the lazy-loader wrapper. I'm pretty sure the whole dynamic thing would still work with that pattern.

kieckhafer:
@aldeed So I tried that, and there are actually two React components being used inside this function: https://github.com/reactioncommerce/reaction/pull/3622/files#diff-a054f2138c1bcc6cc677e32da601cc61R10

So even moving it into it's own file will still cause that error inside the new file.

aldeed:
@kieckhafer oh because of the `loading` function component? Well, yeah that would be fine to disable in the loader file, then, but I'd still make that `DayPickerRangeController` be in its own file separate from `CalendarPicker`.

And no, I don't know any alternatives, and I've heard good things about `react-loadable` so that would have been my choice, too. Haven't used it before, though

willopez:
LGTM, ship it :shipit:

PR #3631, opened 2018-02-01
Refactor 3482 nnnnat replace imagemagick

### LINKED ISSUE #3482: Deprecate ImageMagick/GraphicsMagick dependency
The requirement for using ImageMagick / GraphicsMagick for resizing images requires an additional dependency outside of our application environment, and there has always been a host of related issues... at the time there didn't appear to be any good Node modules available. 

http://sharp.pixelplumbing.com/en/stable/

It appears that [sharp](https://github.com/lovell/sharp) should be a drop in replacement and faster as well...

nnnnat:
Fixes #3482 
## Summary
Replaced GraphicMagick/ImageMagick with [sharp](http://sharp.pixelplumbing.com/en/stable/).
- Added sharp to the project and dynamically loading in the package.
- Updated image transforms to to use the `sharp()` functions.
- Refactored the `"Media" FS.Collection` to map the image transforms through a `buildGFS()` function to create each  `FS.Store.GridFS` collection.
- Updated and added new code comments.

## Testing
- Login as Admin and go to the Product Details Page.
- Open your network panel in dev-tools and select the `images` filter
- Upload a new product image and copy the new image's request URL from the network panel
- Open a new browser tab and paste the image's URL into the address bar
- At the very end of the URL after the token, you can add `&store={size}` to view each image transform.
- Image transforms sizes are `large`, `medium`, `small` & `thumbnail`
- Image transforms should be the same as the old `gm` crops

**Note to reviewer**
Wasn't sure if I'm doing the lazy load of sharp in the right way. I'm calling `lazyLoadSharp()` from within the `buildGFS()` function which is being mapped over when I create the `stores` Array. I'm guessing that the `if (sharp) return;` inside of `lazyLoadSharp()` is preventing sharp from being loaded more then once.

Never used JSDoc before so let me know if I'm failing at it


lib/collections/collectionFS.js
```diff
@@ -13,80 +22,77 @@ FS.HTTP.setHeadersForGet([
...
+ * 2. `medium` - 600px by 600px - preserves aspect ratio
+ * 3. `small` - 235px by 235px - crops to square - creates png version
+ * 4. `thumbnail` - 100px by 100px - crops to square - creates png version
+ */
+const imgTransforms = [
+  { name: "image" }, // creates the first `image` GridFS store
+  { name: "large", transform: { size: 1000, helper: "max" } }, // max helper
+  { name: "medium", transform: { size: 600, helper: "max" } }, // max helper
+  { name: "small", transform: { size: 235, helper: "png" } }, // png helper
+  { name: "thumbnail", transform: { size: 100, helper: "png" } } // png helper
```
aldeed: For "small" and "thumbnail", the existing transforms have "^" appended, and gm docs say:

> Append a ^ to the geometry so that the image aspect ratio is maintained when the image is resized, but the resulting width or height are treated as minimum values rather than maximum values.

So I think helper should be `"min"` on those, and then "png". 

The "extent" should not be necessary for those I would think, since it will fill the square. I'm not sure why they have `.gravity("Center")` on them, but @aaronjudd might know whether we can safely omit that.

aldeed:
@nnnnat Overall looks great. A few comments to address.

I'm also wondering if we should put a "safety" transform on the primary "image" store. Just some largest-we-would-ever-want-to-store resizing to avoid giant files in mongo. And maybe force all the larger sizes to JPEG? Generally the recommendation I've heard is JPEG for larger images, as long as transparency isn't required, and PNG for smaller. Thoughts @aaronjudd?

nnnnat:
@aldeed I've made the changes you've suggested and made a few changes to the imgTransform objects. 

## Updates to imgTransform
- Renamed the helper property to be called mod
- Added a format property to each transform to be used with the [toFormat](http://sharp.pixelplumbing.com/en/stable/api-output/#toformat) function and removed the `png()` modifiers
- Added `transform` settings to the first `"image"` I've set the cap size to be 1600px on the largest side and format to jpg
- Added the `crop()` modifier for the two square crops `"small"` & `"medium"` 
 
**Notes**
I tried to use the `.min()` modifier on the `"small"` & `"thumbnail"` transforms but I wasn't getting a square image from the transform. I think the `min()` function resizes the image to make the shortest side equal the size that's passed to `resize` and the other side will be relative to that. I ended up trying the `crop()` modifier and seemed to be getting the same image transform as what on master.

*thumbnail transform from master w/ imgMagick*
master-thumbnail

*thumbnail transform w/ sharp & `min()` modifier*
min-thumbnail

*thumbnail transform w/ sharp & `crop()` modifier*
crop-thumbnail

aldeed:
@nnnnat Yes `crop` looks right. Thanks for the good research, posting comparison pics to help explain, etc.

machikoyasuda:
Adding a Doc ticket for this ticket https://github.com/reactioncommerce/reaction-docs/issues/493 to update Installation docs to remove ImageMagick step

PR #3639, opened 2018-02-02
(refactor): remove meteor-collection-hooks dependency for orders

willopez:
This PR removes the meteor-collection-hooks package dependency for new orders. `Hooks.Events` is now used to achieve the same functionality 

**Testing Procedure**
- Place an order by completing the checkout process
- Verify that a new AnalyticsEvent document with a field named value is set to the new order's `_id`


willopez:
@aldeed tests are now fixed.

PR #3653, opened 2018-02-06
update Docker base for multi-stage builds 

jshimko:
This is a companion PR to https://github.com/reactioncommerce/base/pull/10. 

Using [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) the new Dockerfile to export the built app into a new smaller image without any unnecessary build dependencies.  This means a much faster build and smaller image.  See Docker's docs on multi-stage builds for more detail.

This PR also: 
- Updates existing `docker-compose.yml` to serve as the config for running a local development environment.
- Adds  a new `docker-compose-demo.yml` for testing out production builds (this is the replacement for the previous `docker-compose.yml`).

**Note that https://github.com/reactioncommerce/base/pull/10 needs to be approved and merged _before_ this PR.**


jshimko:
FYI, to test this PR...

```sh
# get the new base image PR branch
git clone -b multi-stage-builds --single-branch https://github.com/reactioncommerce/base.git

cd base

# build the new base image that this PR depends on
docker build -t reactioncommerce/base:v4.0.0 .

# switch to Reaction repo and this PR branch, then...
cd ..
git clone -b docker-base-updates --single-branch https://github.com/reactioncommerce/reaction.git

cd reaction

# build Reaction using the new base image you created above
docker build -t reactioncommerce/reaction:latest .

# make sure it works
docker-compose up -d

# http://localhost:3000
```
The last build command will build this branch of Reaction using the `reactioncommerce/base:v4.0.0` from your local cache.

mattt416:
Hey @jshimko , I'm testing this now but `docker build -t reactioncommerce/reaction:latest .` is taking an absolute eternity.  Is that expected?  I had to pass in `TOOL_NODE_FLAGS="--max-old-space-size=6144"` otherwise the build would fail.

mattt416:
My bad -- my VM's space had filled up due to some previously failed builds.  Cleaned it up and getting further now.  :)

jshimko:
@spencern Sorry, I didn't include that flag in my example command since everyone has a different amount that works on their machine.  The following has worked for me nearly 100% of the time since we started using that flag.

```sh
docker build --build-arg TOOL_NODE_FLAGS="--max-old-space-size=2048"  -t reactioncommerce/reaction:latest .
```

Higher definitely isn't better.  If I set it too much higher than that, it just runs forever or eventually fails.

Also, your first try with the value prepended won't work like that...

```sh
TOOL_NODE_FLAGS="--max-old-space-size=4096" docker build -t ...
```

Environment variables in your current environment don't get passed into the container's build environment.  That's why we have to use the `--build-arg` flag.  That [sets the environment inside the build image](https://github.com/reactioncommerce/base/blob/b173c69f6f2abc051c696a3bf8ec645d23b4fc3e/Dockerfile#L30-L31).

Dockerfile
```diff
@@ -1,5 +1,50 @@
-FROM reactioncommerce/base:v3.0.0
+FROM reactioncommerce/base:v4.0.0 as builder
+
+MAINTAINER Reaction Commerce <architecture@reactioncommerce.com>
+
+# copy the app into the build container
+COPY . $APP_SOURCE_DIR
+
+# build the app with Meteor
+RUN ${BUILD_SCRIPTS_DIR}/build-meteor.sh
```
ticean: We can include the script actions directly in the Dockerfile.

Dockerfile
```diff
@@ -1,5 +1,50 @@
...
+ENV MONGO_MAJOR 3.4
+
+# PhantomJS
+ARG INSTALL_PHANTOMJS
+ENV INSTALL_PHANTOMJS $INSTALL_PHANTOMJS
+ENV PHANTOM_VERSION 2.1.1
+
+# install optional dependencies if an above --build-arg is supplied
+RUN /tmp/scripts/install-phantom.sh
+RUN /tmp/scripts/install-mongo.sh
```
ticean: We can include the script actions directly in the Dockerfile.

aaronjudd:
@impactmass if your recent should have @ticean review again, can you dismiss his review and move this forward?  If you guys are pairing and your updates aren't being capture here, can you close the PR or mark it as [WIP]?  Status is unclear...

impactmass:
@ticean, I've pushed the updates for the Dockerfile cleanup and for the dir changes in reactioncommerce/base#10

PR #3661, opened 2018-02-07
(refactor): remove TranslationProvider from lower level components

kieckhafer:
Resolves unreported issue
Impact: minor
Type: refactor

## Issue
Reaction has leftover `<TranslationProvider />` React wrappers in various parts of the application, from a time when `TranslationProvider` did not wrap the entire application at the second level (just below `BrowserRouter`).

## Solution
Remove `<TranslationProvider>` component wrappers at all lower level parts of the application, leaving only the top level wrapper, and an additional wrapper around the Admin toolbar.

## Breaking Changes
- None

## Testing
- Start the app
- See translations working throughout the app with no errors


kieckhafer:
@mikemurray updated the format

Issue #2963, opened 2017-09-27
Large action view not properly implemented on accounts/permissions admin


rymorgan:
### Expected behavior
The cards in the action panel shouldn't have extra space or double lines. And the detail view, in general, should be working properly. 

The large view accounts/permissions admin should work like the order large panel with the master and detail part of the page scrolling independently of each other. 

### Actual Behavior
They have extra space and double lines. And the panel lines don't extend all the way down the page. 
![accounts](https://user-images.githubusercontent.com/20409254/30719683-40d48b62-9ed9-11e7-8043-344eeb2e8b3d.png)

The panels aren't working properly and are simply floating out in the middle of the panel. 

### Steps to Reproduce the Behavior
Navigate to accounts. 
Open close and navigate through the panels and see that it janky. 

### Versions
Node: 4.5.0
NPM: 5.4.1
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.12.0
Reaction: 1.5.0
Reaction branch: marketplace

rymorgan:
@spencern I added this for more detail. We can pull in the PR for https://github.com/reactioncommerce/reaction/pull/2949 separately.

If this issue is a lower priority and doesn't make it in that's ok per Aaron's comments today. But hopefully, we can get this done right after release in that case.

spencern:
Any chance this is related to #2921 and could be fixed there? If so I think we can fit it in, otherwise I think we punt to `next`

rymorgan:
@spencern @mikemurray Wanted a separate issue for this fix. This issue is the actual underlying issue behind #2921

rymorgan:
@joykare While this an important issue for me, check with @spencern whether this is a priority over other crucial issues we need to push out for release.

joykare:
*Notes*
On pause till other more urgent marketplace bugs are resolved

spencern:
@joykare thanks for the update.

@rymorgan are you okay pushing this issue out until `release:next`?

rymorgan:
@spencern Short answer: yes. 

Long answer, yes, but I want to do it properly and do all of the requirements here: https://github.com/reactioncommerce/reaction/issues/2922

PR #3665, opened 2018-02-08
Fix error when creating/update groups and/or group permissions for non-admin user

### LINKED ISSUE #3649: Error thrown when creating/editing group as shop user with dashboard access
### Expected behavior
Creating / Editing a group should not throw `Error creating group`. It should create/edit group successfully. 

screenshot 2018-02-05 21 18 51

### Actual behavior
Creating / Editing a group should throws `Error creating group` when performed by a user other than the shop owner.

1. Add a user to shop manager group, and login as that user
2. Try creating / editing a group as that user

Some Notes: 
- This could be a case of user not having required permission. 
- Maybe show a better error message for user not with right permission to perform the action
- Or maybe should not show the edit/create module for users without right permission in first place

### Versions
Node: 8.9.1
NPM: 5.6.0
Meteor Node: 8.9.3
Meteor NPM: 5.5.1
Reaction CLI: 0.26.4
Reaction: master

prinzdezibel:
Resolves #3649

This PR does prevent an error being displayed when a user without the
role "shop-admin" or "owner" tries to create a new group or tries
to update permissions of an existing group.

It does so by not displaying editing buttons and having the permission
toggles not updating when switching on/off. I.e, they behave like
read-only widgets.

### DIFF
--- a/imports/plugins/core/accounts/client/components/editGroup.js
+++ b/imports/plugins/core/accounts/client/components/editGroup.js
@@ -21,6 +21,7 @@ import { groupPermissions } from "../helpers/accountsHelper";
 class EditGroup extends Component {
   static propTypes = {
     accounts: PropTypes.array,
+    canEdit: PropTypes.bool,
     groups: PropTypes.array,
     onChangeGroup: PropTypes.func,
     packages: PropTypes.array,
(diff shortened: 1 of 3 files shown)


prinzdezibel:
Agree to your thougts, @impactmass. I was thinking the same and I'm not entirely happy with my PR.

> Is there a read-only display for the permissions toggle list?
No, I've investigated. When I started, I thought there's a disabled / read-only visual representation of the toggles. But there isn't. Only active and non-active.

> On higher level, that section of the dashboard is "Edit Groups". I think it should probably be hidden entirely when user has no edit rights. I can't pick out a reason for a read-only version for non-admins yet. (this will need experience team input as well)

Yes, I guess that is one direction we could head towards, if it doesn't make sense to have the permissions displayed read-only. 

@rymorgan I'd appreciate some instructions on this ticket from your side. What's the right thing to do here?

impactmass:
I had a chat with @rymorgan and also agree with the approach mentioned above... that is:
- we should completely hide the "edit groups" module for users without permission to edit/perform any action in it.
- there is no use to use user seeing the list (read-only) since they cannot modify
So we can take that approach in this PR. 

PS: Taking this approach will also increase the priority of #2963 - which will handle the way the main section and the smaller side bar scrolls

prinzdezibel:
@impactmass: done

prinzdezibel:
@impactmass  thanks!

PR #3713, opened 2018-02-09
(i18n): Updated French translations

loanlaux:
This PR updates French translations for most `core` and `included` plugins, many of which were wrong or clumsy. Any feedback from another French native speaker would be greatly appreciated!


imports/plugins/core/checkout/server/i18n/fr.json
```diff
@@ -10,13 +10,13 @@
...
         "orderTotal": "Total de la commande",
-        "fulfilledBy": "Réalisé par",
+        "fulfilledBy": "Effectué par",
         "quantityTotal": "Quantité totale",
         "orderSubtotal": "Total partiel",
-        "orderShipping": "livraison",
+        "orderShipping": "Livraison",
         "orderTax": "Paramètres de taxe",
-        "trackYourDelivery": "Les mises à jour concernant la commande seront envoyées à",
-        "registerGuest": "salut! Ajouter un email et recevoir des mises à jour de commande"
```
prinzdezibel: @loanlaux shouldn't that start with capital `S`?

imports/plugins/core/payments/server/i18n/fr.json
```diff
@@ -12,21 +12,21 @@
         "dashboard": {
           "paymentsLabel": "Paiements",
           "paymentsTitle": "Paiements",
-          "paymentsDescription": "configuration des paiements"
```
prinzdezibel: Capital?

imports/plugins/core/checkout/server/i18n/fr.json
prinzdezibel: sorry. looking at the old stuff :)

imports/plugins/core/checkout/server/i18n/fr.json
```diff
@@ -10,13 +10,13 @@
...
+        "trackYourDelivery": "Les informations concernant la commande seront envoyées à",
+        "registerGuest": "Bonjour ! Renseignez votre adresse e-mail pour être informé de l'avancement de votre commande"
```
mattt416: Honest question here -- is it normal in French to put spacing before punctuation like `!`, `?`, etc.?  I see you've done this a fair bit but similar exists in places in the original translations.

loanlaux: That's one of the weirdest little-know quirks of the French language! We do put a space before what I'd call "double punctuation" (that is, any punctuation composed of two parts, such as "!", "?", ":" and ";"). However, we don't put spaces before periods or commas.

I'd have a hard time explaining why. I've seen Canadians often do the same though.

mattt416: I'll be honest, it does read slightly better and I do it subconsciously from time to time (Canadian here).  Thanks for the info!  :)

PR #3737, opened 2018-02-13
Prevent mobile views having elements that are being cut off.

### LINKED ISSUE #3396: Mobile views have elements that are being cut off.
### Expected behavior
Nothing should be cut off on mobile views. 

### Actual behavior
The header is too wide and is cutting off elements on mobile. 

Look at Reaction Commerce on mobile as an admin. 

screen shot 2017-12-12 at 3 15 08 pm
screen shot 2017-12-12 at 3 14 18 pm

prinzdezibel:
Closes issue #3396.

- Retained behaviour: Uses 400px width for the action view where viewport width >= 400 px
- Changed: Uses viewport width for action view when viewport width < 400px
- Removed unnecessary media query from inline style (@media will not work there anyway)
```
-    "@media only screen and (max-width: 949px)": {	
-        width: "100vw"
-    },
```
- Removed unnecessary  width: "50%" , because width is calculated through flex-basis (has precedence)

- Add new property viewportWidth to container, which serves as equivalent to a css media query . Moved it out to container, because it's environment dependent and `windows` is possibly not defined. Let me know if you think it makes sense to put it back to the component.
- Remove unnecessary width rule from .admin-controls-content

### DIFF
--- a/imports/plugins/core/dashboard/client/components/actionView.js
+++ b/imports/plugins/core/dashboard/client/components/actionView.js
@@ -17,14 +17,14 @@ import { getComponent } from "@reactioncommerce/reaction-components";
 
 
 const getStyles = (props) => {
-  let viewSize = 400;
+  const minWidth = Math.min(props.viewportWidth, 400);
+  let viewSize = minWidth;
   const actionView = props.actionView || {};
   const provides = actionView.provides || [];
   // legacy provides could be a string, is an array since 1.5.0, check for either.
   // prototype.includes has the fortunate side affect of checking string equality as well as array inclusion.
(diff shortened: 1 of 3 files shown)


spencern:
Gonna merge this since it has an approved PR. 
One thing I would like to see going forward is that PRs which make changes or fix UI issues include screenshots of before and after. This will help us know what it now looks like and permit @rymorgan and @cassytaylor to get involved with design reviews without having to boot up the whole app

zenweasel:
@spencern Can we add that additional requirement to the Pull Request template doc?

spencern:
@zenweasel good idea

PR #3745, opened 2018-02-14
Fixes #2736: added unique to slug

### LINKED ISSUE #2736: [Marketplace] Can create shop with same name as another shop
### Expected behavior
Shops should not be able to share a name / slug, each should have a unique name / slug to access it

### Actual Behavior
Adding two shops with the same name causes an issue where only 1 of the shops shows up. The second (or third or fourth...) shop is still inside the database, so the info is saved, we just pull the data from the most recently added.

1) Invite a new shop to marketplace
2) Sign in with your new shop
3) Name your shop "Test Reaction Shop" (or anything)
4) Invite a second new shop to marketplace
5) Sign in with your new shop
6) Name your shop the same name as you did the previous new shop ("Test Reaction Shop")
7) See that only the most recently added shop is available

### Possible solution
We do not save a shops slug on the shop object, we dynamically create it by slugifying the shop name. If we were to save a shop slug, we could save this as a unique field, which would allow two shops to have the same name.

Or... we just don't allow two shops to have the same name

### Versions
Node: 4.2.6
NPM: 5.3.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.11.0
Reaction: 1.5.0
Reaction branch: marketplace
Docker: 17.06.1-ce

Akarshit:
Resolves #2736 
Impact: major
Type: bugfix

## Issue
Shops with the same name were not accessible

## Solution
Slug is already saved in the Shop Schema, we only need to check for its uniqueness.

## Testing
1. Invite a new user to marketplace
1. Sign in as the new user
1. Name your shop "Reaction2" (or anything)
1. Invite a second new user to marketplace
1. Sign in with your new user
1. Name your shop the same name as you did the previous new shop ("Reaction2")
1. See that both of the shops are available for the `owner` user.
1. Also as the 2nd user try to change the `slug` to a existing slug(reaction2)
1. Note the form won't allow you to do that.


zenweasel:
What happens to the slug of the second shop?

zenweasel:
@Akarshit Looks like you have a few failing tests there

Akarshit:
The user would have to change the slug to something unique for the form to get saved(the user will get alerts "Error saving the shop settings, slug is not unique).

I am looking into why the test are failing.

Akarshit:
The tests were failing because the shop being used in the factory had the name as a value instead of a function returning the random value. 
Hence all the shops were getting created by the same name and slug. This was not a problem before but will be a problem from now.

spencern:
@Akarshit does this need a migration? What happens if there are already duplicate slugs when this update is merged?

Akarshit:
@spencern yes you are right. This will need a migration otherwise there will be problems when the `Shop` document is updated. 
But would it be a good idea to change the slugs of existing shop in the migration?

zenweasel:
@spencern I think we are waiting for your feedback on this?

Akarshit:
@jshimko As you said
> The schema validator only effects inserts/updates

What would happen in the following case
Suppose `Shops` contain 2 documents with Ids 111, 222 and same slug. Would this operation succeed?
`Shops.update({ _id: 111}, {$set: {someUnrelatedField: "something"}})`

zenweasel:
@akarshit I don't think the schema validator would throw on already existing data but this seems like something you could test yourself?

PR #3755, opened 2018-02-15
Fix 2246: Setting or changing a products perma-link causes hard refresh

### LINKED ISSUE #2246: Setting or changing a products perma-link causes hard refresh
### Expected behavior
When I publish a product, I'd expect the app to do so without refreshing

### Actual Behavior
http://g.recordit.co/DwTDf6K2O9.gif
When updating a product's permalink (even for the first time), the app refreshes upon publish.

1. Clone `development`
2. Login as admin
3. Create a new product.
4. Set variant label to something
5. Click the edit button next to the social icons
6. Open the product settings accordion
7. Set the product title (and the perma-link will be set as well)
8. Hit publish
9. Observe hard refresh.

### Versions
```
Node: 7.8.0
NPM: 4.5.0
Meteor Node: 4.8.2
Meteor NPM: 4.5.0
Reaction CLI: 0.8.1
Reaction: 1.1.1
Reaction branch: development
Docker: 17.03.1-ce
```

prinzdezibel:
Resolves #2246
Impact: minor
Type: bugfix

## Issue
Changing or creating the permalink of a product did trigger a page reload. Also this action was not reactive, meaning that other connected clients did not receive the new URL after publishing.

## Solution/Changes

- Don't trigger route change from callback value of meteor call `revisions/publish`, because
1)  it's changing server state that gets delivered to connected clients asynchronously. It's not possible to rely on the publications already being synced to the client side when the method returns. This led to race conditions when trying to change the route client side, therefore it was done the hard way by setting `window.location.href`. Tried to observe permalink changes client side, but that was unreliable as well, probably because tracker runs may re-instantiate observed cursors at any time.

2) it's not reactive (other connected clients won't get noticed about the route change), when the permalink has changed

->  INSTEAD we observe changes to handles server side and let the client know it has changed through a new property `changedHandleWas`, which will be published like any other property (gets persisted to db, but if this would be a problem we could eliminate that through usage of `Products.observe` instead of `Products.observeChanges`)

- Don't use window.location.href for route changes any more. Router.go() is used instead.

- Changes to the Product publication which seems wrong  regarding updating product.__revision    
  Don't publish the whole product object to the client. The code changed is only about the property __revisions on the product collection. Therefore it's wrong to push the whole object to the client. If done anyway, it will push product data to the client (e.g. handle, title) that is going to be updated soon after due to meteor method revision/publish, which results in **outdated** data on the client (which in turn results in template "not found" being displayed)

- Also I think it's wrong to do database queries to decide what should be pushed to the client, e.g.
```
  const revisions = Revisions.find({	
        "documentId": id,	
         "workflow.status": {	
            $nin: [	
                "revision/published"	
            ]	
          }	
        }).fetch();	
	
     fields.__revisions = revisions;	
     this.changed("Products", id, fields);
```
doesn't seem correct. To know what needs to be synced with the client, I think the source of truth needs to be the merge box (session collection view). Using db queries will not work reliably, as it seems. 

    
## Breaking changes
none expected

## Testing
1. As an admin create a new product
2. Change permalink of the product
3. Publish changes
4. See updates to content and URL without page refresh
5. URL for connected clients should update as well


prinzdezibel:
It still doesn't work every time. Need to test more.

prinzdezibel:
Update: After more investigation, this gets really wicked (I'm sure this does not come as surprise to all guys having been working here already) !

This is my take away so far:
1) I'm confident that this can be streamlined significantly, but it needs to be tested extensively
2) I believe that the original ticket (changing the handle of a product with subsequent client side refresh without hard-reload) can't be achieved without addressing the following fact in the old code:
  The **preSelector-cursor** might change when the user changes the permalink (`handle`): https://github.com/reactioncommerce/reaction/blob/ac07d02a971797863d0f593fc4608dd4b0a2611c/server/publications/collections/product.js#L93
   This will have impact on the result of this cursor, that gets observed: https://github.com/reactioncommerce/reaction/blob/ac07d02a971797863d0f593fc4608dd4b0a2611c/server/publications/collections/product.js#L119
3) I believe because server side publish functions are not reactive, the observer will not push changes to the client, if `Products.findOne(preSelector)` changes.

I've tried to get rid of that cascading database query and it seems working so far. But this is no easy change and I'm not so naive to proclaim we're all good now!

prinzdezibel:
### Below are my testing comments and open things
- When a new product is created, the default variant is not selected by default. This sometimes results in an picture being displayed that belongs to another product. I need to find out why the first variant is not selected automatically (like seen in master branch).

prinzdezibel:
I have overhauled the `Product` and `Products` publications significantly. I think they are very buggy in its current state. My observation is that they behave extremely unpredictable and erratic as there's a lot of async stuff going on behind the scene and the order of the events is not deterministic. I can't give a guarantee that what I've done is correct or even working in all circumstances. But I know for sure that the current implementation is buggy. I would bet $5 that it's not possible to reliably solve the task at hand (the client side redirect when the permalink changes) without overhauling the `Product` and `Products` publication. I tested as much as I can and it looked ok though..

zenweasel:
@prinzdezibel I feel like overhauling the two main product publications is way out-of-scope for this ticket, before you do any more work in that avenue let's talk.

prinzdezibel:
@zenweasel Not planning any more work on that, as it looks good to me. But would be great if somebody could have a look into it. Would be a sad is we need to ditch that. I think it's an improvement, really (assuming I have not overseen something important - which I wouldn't rule out of course).

zenweasel:
@prinzdezibel Looks like you have conflicts here

jshimko:
@prinzdezibel There's a lot going on here with a bunch of things I'm not very familiar with. Can you please resolve this merge conflict since you're more familiar with these proposed code changes? Once you've done that, I'll confirm this is all doing what it's supposed to.

prinzdezibel:
Yes, @jshimko, will do. I also wanted to check the marketplace feature in-depth, but I can't because of another bug. I'm unassigning you, until it's ready. Because right now - shame on me - it's not prepared adequately.

prinzdezibel:
Merged, but can't test because of #3851 and #3846.

prinzdezibel:
@jshimko  I think this is ready for a review from you. After it has passed this first stage, I'll also assign Spencer as second reviewer. I've seen some errors, but these are happening on master as well (listed in open issues above). Therefore I would propose to ignore them until this is approved, because maybe even more things come up during testing. I'm going to file all those issues later as new, separate issues.

For testing, you'd need to pull in two other PRs from Brent:
- https://github.com/reactioncommerce/reaction/pull/3856
- https://github.com/reactioncommerce/reaction/pull/3857

prinzdezibel:
@jshimko Tests are passing. Spencer was suggesting that you reviewing it should be enough. The reason I wanted him as second peer reviewer was because Aaron put him on the reviewer's list a couple of days ago. Anyway, we can always ask another person to look into it, if we think it helps..

aaronjudd:
@prinzdezibel can you review and fix up the conflicts one more time for 1.9.0? 👍 @spencern we've discussed but since the catalog updates are pending, I think this should be reasonable to merge - we can pair on a review if necessary for 1.9...

PR #3770, opened 2018-02-16
FIxes #3763: | replaced with ||

### LINKED ISSUE #3763: Artifact in order summary
### Expected behavior
Should just display date without additional pipe character

### Actual behavior

Using a single pipe rather than a double-pipe for OR
https://github.com/reactioncommerce/reaction/blob/4fae50ddac1bc69474b90e98b9564f68a83f785a/imports/plugins/core/orders/client/components/orderSummary.js#L101

1. Place an order
1. Go to the admin
1. Observe the extra character

### Versions
Node: 8.9.1
NPM: 5.5.1
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.27.0
Docker: 17.12.0-ce

Akarshit:
Resolves #3763  
Impact: minor  
Type: bugfix

## Testing
1. Place an order
1. Go to the admin
1. Observe the time to be correct without any extra character.

### DIFF
--- a/imports/plugins/core/orders/client/components/orderSummary.js
+++ b/imports/plugins/core/orders/client/components/orderSummary.js
@@ -98,7 +98,7 @@ class OrderSummary extends Component {
             <div className="order-summary-form-group">
               <strong data-i18n="order.created">Created</strong>
               <div className="invoice-details">
-                {moment && moment(order.createdAt).fromNow()} | {dateFormat(order.createdAt, "MM/D/YYYY")}
+                {(moment && moment(order.createdAt).fromNow()) || dateFormat(order.createdAt, "MM/D/YYYY")}
               </div>
             </div>
 


imports/plugins/core/orders/client/components/orderSummary.js
```diff
@@ -98,7 +98,7 @@ class OrderSummary extends Component {
             <div className="order-summary-form-group">
               <strong data-i18n="order.created">Created</strong>
               <div className="invoice-details">
-                {moment && moment(order.createdAt).fromNow()} | {dateFormat(order.createdAt, "MM/D/YYYY")}
+                {(moment && moment(order.createdAt).fromNow()) || dateFormat(order.createdAt, "MM/D/YYYY")}
```
mikemurray: @Akarshit I think its supposed to show both the `fromNow` date and the `actual date` the `|` probably needed to be like `{"|"}`.

Akarshit: @mikemurray If that is the case then the issue is basically moment not being present, which is tracked in #3778

mikemurray: @Akarshit That seems to be on the checkout page, whereas this is for the orders dashboard.

Akarshit: @mikemurray Sorry I meant #3773.
I get your point I will close this PR.

mikemurray: @Akarshit no, I think your PR is fine. #3773 is probably related to the list view. I think you only need to add the `{"|"}` because this is for display is not an OR. It's supposed to display both dates.

The `moment && moment()` is only to keep errors out of the console until `moment` has been loaded.

mikemurray:
Closing in favor of #3773

PR #3773, opened 2018-02-16
Fix error when switching table layout in order dashboard

### LINKED ISSUE #3760: Cannot switch orders dashboard between list and table view
### Expected behavior

Should be able to switch between table and list view of orders dashboard.

### Actual behavior

Receive a `moment is not a function` error, and the view never switches.

basic_reaction_product

screen shot 2018-02-15 at 2 41 55 pm

1. As any user, create an order by checking out with products
1. As an admin, open the order dashboard
1. In the upper right corner, there will be two icons that will allow you to toggle table and list view. (see screenshot)
1. Click the one that's not highlighted
1. In the console, observe generated errors

### Versions

Node: 7.2.0
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.27.0
Reaction: 1.8.0
Docker: 17.12.0-ce

prinzdezibel:
Resolves #3760
Impact: major
Type: bugfix

## Issue
As an admin it's not possible to switch the table layout in the order dashboard.

## Solution/Changes
- The withMoment higherOrderComponent would always return undefined,
    because there's no property default on the moment object

- Render getLocaleString() on the date object until moment gets
    available (async)

## Breaking changes
none expected

## Testing
1. create a order and checkout
2. As an admin, go to the order dashboard
3. Switch table layout
4. It should work without error

localhost_3000

### DIFF
--- a/imports/plugins/core/components/lib/hoc.js
+++ b/imports/plugins/core/components/lib/hoc.js
@@ -44,10 +44,10 @@ export function withMoment(component) {
   return lifecycle({
     componentDidMount() {
       import("moment")
-        .then(moment => {
+        .then((moment) => {
           moment.locale(Reaction.Locale.get().language);
           this.setState({
-            moment: moment.default
+            moment
           });
         })
         .catch((error) => {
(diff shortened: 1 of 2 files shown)


prinzdezibel:
@mikemurray Sorry, I filled that out already, but requested a pull into the wrong branch: https://github.com/reactioncommerce/reaction/pull/3773 

Have closed that already and will copy over the instructions. Sorry, my bad.

PR #2022, opened 2017-03-21
Cancel Order

michael-alade:
**Resolves**
#1238

**Testing**
- Login as admin
- Choose a payment provider
- In the payment provider settings, choose the methods supported by the payment provider.
**Authorize**: This happens when the payment processor receives the payment.
**De-authorize**: This method runs when you want to cancel a payment before a capture has been done.
**Capture**: This method runs when payment is sent to admin.
**Refund**: This allows the admin make either a full/partial refund to the customer.
- Make an order
- Navigate to the orders page
- Click on the order to start processing

- To cancel before approval, click on the multi-function button toggle then select cancel.
1. Click the cancel button on the multi-function button.
2. A modal would be displayed, telling you it's going to do a full refund of the payment made by the customer. You could select either to restock or not restock the product after the cancel.
3. The order would be moved to the completed tab.

- To cancel the order after payment capture.
1. Click on cancel order.
2. A modal would be displayed, telling you it's going to do a full refund of the payment made by the customer. You could select either to restock or not restock the product after the cancel.
3. The order would be moved to the completed tab.

PR #3776, opened 2018-02-16
(fix): Correct Inventory updates when canceling an order

### LINKED ISSUE #3678: Inventory updates work incorrectly when cancelling an order
### Expected behavior
- If I cancel an order, and choose to restock, my available inventory for the product should increase by the number of items I've "cancelled". For example, if I started with 20 products, ordered 5, captured the payment, and then cancelled the order with restocking, I should have 20 available products.
- If I cancel an order, and choose not to restock, my available inventory for the product should equal the original quantity, minus the quantity of this order. For example, if I started with 20 products, ordered 5, captured the payment, and then cancelled the order without restocking, I should have 15 available products.

### Actual behavior
- If I cancel an order, and choose to restock, my available inventory for the product stays at the depleted amount. For example, if I started with 20 products, ordered 5, captured the payment (the inventory drops to 15), and then cancelled the order **with** restocking, I still have 15 in my inventory.

https://youtu.be/PJKnsPruzMw

- If I cancel an order, and choose not to restock, my available inventory for the product is taken away twice, instead of just once. For example, if I started with 20 products, ordered 5, captured the payment (the inventory drops to 15), and then cancelled the order without restocking, the inventory drops an additional 5, and I now have 10 in my inventory (instead of 15).

https://youtu.be/izMjluy9fEQ

1. Write down the inventory of a product, and then create an order with X amount of the product
2. As an admin, begin to process the order, and capture the payement
3. Revisit the product and see that inventory has dropped by X
4. Cancel the order, choose "without restocking"
5 Revisit the product and see that inventory has dropped by an additional X

---

1. Write down the inventory of a product, and then create an order with X amount of the product
2. As an admin, begin to process the order, and capture the payement
3. Revisit the product and see that inventory has dropped by X
4. Cancel the order, choose "with restocking"
5 Revisit the product and see that inventory is still at the initial inventory - X, instead of being at the original inventory number

### Versions

```
Node: 9.5.0
NPM: 5.6.0
Meteor Node: 8.9.3
Meteor NPM: 5.5.1
Reaction CLI: 0.27.0
Reaction: 1.8.0
Docker: 18.01.0-ce
```

Browser: http://whatismybrowser.com/w/KE449ZE

### QA

QA by @kieckhafer + @cassytaylor

impactmass:
Resolves #3678   
Impact: major
Type: bugfix

## Issue
1. If I cancel an order, and choose to RESTOCK, my available inventory for the product stays at the depleted amount. 
2. If I cancel an order, and choose not to restock, my available inventory for the product is taken away twice, instead of just once. 

## Solution
Within the cancelOrder method, the restock logic is like this:
```
if (!returnToStock) {
  Products.update({
  ...
  }, {
    $inc: {
      inventoryQuantity: -item.quantity
    }
  });
}
```
which means that when when the admin decides NOT to restock, the product quantity still gets decreased. This is led to the problem of double decrease in quantity (first at order capture, then second at order cancellation). This is the cause of the bug (introduced in #2022).

The correct logic should be:
```
if (returnToStock) {
  Products.update({
  ...
  }, {
    $inc: {
      inventoryQuantity: +item.quantity
    }
  });
}
```
which will mean that ONLY when the user decides to return to stock, we go ahead and increment the current inventory with the canceled item quantity.

The remaining changes I made are to the tests.

## Testing
Flow A
1. Note the current inventory of a product, and then create an order with X amount of the product
2. As an admin, begin to process the order, approve and capture the payment
3. Revisit that product (on the PDP) and see that it's inventory has dropped by X
4. Cancel the order, choose "without restocking"
5. Revisit the product (on the PDP) and confirm that the inventory stays the same (i.e no restocking)

Flow B
1. Again, note the current inventory of a product, and then create an order with X amount of the product
2. As an admin, begin to process the order, approve and capture the payment
3. Revisit that product (on the PDP) and see that it's inventory has dropped by X
4. Cancel the order, choose "with restocking"
5. Revisit the product (on the PDP) and confirm that the inventory increased by X (i.e restocked)


impactmass:
Thanks. I'll work on this later today (this slipped due to other tasks).

kieckhafer:
@impactmass I'll take over this for you since you've moved onto other issues.

impactmass:
cool 👍

kieckhafer:
@willopez I took over this ticket from @impactmass, and made the changes to fix the issue I had commented above.

Since I made the changes, I'd like someone else to do another review. Please follow Seun's instructions from above, and also add this new test to assure my fix is working:

---

Flow A
1. Note the current inventory of a product, and then create an order with X amount of the product
2. As an admin, begin to process the order, but DO NOT capture or approve the order
3. Cancel the order by clicking the drop-down button
4. Cancel the order, choose "with restocking"
5. Revisit the product (on the PDP) and confirm that the inventory stays the same

Now that seems a bit confusing, that we choose "With Restocking", but the inventory does not change. My thought on this was to keep it how it is, so that there is always the same and buttons when cancelling an order, and let the code do the magic in the background. I'm open to removing that "With Restocking" button altogether, and just have a "Cancel" button there with no options, however I think that might cause more confusion since there would then be two different visual representations of cancelling a product from basically the same screen.

willopez:
@kieckhafer sure, I'll take a look.

willopez:
@kieckhafer I have tested all the three flows. I do agree that there is an inconsistency between the canceling flows. I would suggest keeping the cancel confirmation dialog the same for all three scenarios. I would decrement the inventory quantities when a new order is received. The user already expects to receive a product since the order was placed.

kieckhafer:
@willopez The inventory updates aren't based on the users actions, but the admins. I don't think changing the timing on that is the right thing to do.

willopez:
@kieckhafer OK, I was not aware of that. LGTM

kieckhafer:
Explanation of changes in the test:

Inside the changed test, the `order` object was being updated inside the first `Meteor` method, `orders/approvePayment`, but we never got that data returned from that method. So we were still using the original `order` object inside the next `Meteor` method, `orders/cancelOrder`, but since my change made it so the inventory update uses the data that is changed inside `orders/approvePayment`, it was failing.

So now, after the `orders/approvePayment` call, I'm re-finding the order from the database, so we have the most up to date order info.

Issue #3685, opened 2018-02-08
Email Status doesn’t update until you navigate away from the email mail provider setting 


cindyfirestone:
### Expected behavior
When I enter in my mail provider setting under the email icon of the dashboard, and the information is successfully entered, the `status` should change immediately from red to green. 

### Actual behavior
The `status` remains red until I navigate somewhere else on dashboard and then come back to the email settings or do a hard refresh and go back to the email settings - then it will display the `status` as green. 
![email status doesn t change](https://user-images.githubusercontent.com/22083616/36004698-d88bcfa6-0ce8-11e8-8171-72d5d078555c.gif)

### Steps to reproduce the behavior

1. Log in as marketplace owner
2. Navigate to email and enter edit mail provider settings 
3. Select Provider Name and enter in correct credentials for User and Password
4. Hit Save
5. See that Status icon remains red. 
6. Navigate to another item on the dashboard or do a hard refresh
7. Go back to email settings
8. See `Status` is now green. 

### Versions
```
Node: 9.5.0
NPM: 5.6.0
Meteor Node: 8.9.3
Meteor NPM: 5.5.1
Reaction CLI: 0.27.0
Reaction: 1.8.0
Reaction branch: release-1.8.0
Docker: 17.03.0-ce
```

spencern:
This seems to be an issue with the status not getting updated reactively. I think this may be the same issue reported by #3701

PR #3781, opened 2018-02-16
Fix 3701 nnnnat email status

### LINKED ISSUE #3701: Email status notification confirms and turns green even when connection fails
### Expected behavior
After entering incorrect email settings, messaging should indicate error and status should turn red.

### Actual behavior
alt

1. Click on Email.
2. Click on 'Mail Provider' settings icon. 
3. Enter incorrect connection info into 'Edit Settings'
4. Click 'Save.' 
5. Click either 'app.save' or 'app.cancel,' it takes you to the same message regardless.

### Versions
Node: 6.5.0
NPM: 3.10.3
Meteor Node: 8.9.3
Meteor NPM: 5.5.1
Reaction CLI: 0.19.0
Reaction: 1.8.0

nnnnat:
Resolves #3701 & #3685   
Impact: major  
Type: bugfix

## Issue
Mail provided status was not updating in real-time if the status had changed between `valid` or `error`. The [EmailConfig](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/email/client/containers/emailConfig.js) container was only checking the email settings status in the `componentWillMount` method and was unaware if the status changed after the component was mounted.

## Solution
I moved the logic to check the email setting status into it's own method and call it from both `componentWillMount` and `componentWillUpdate` methods. 

## Testing
1. As an Admin open the email admin panel to view the current Mail Provider status.
2. Select the edit button and update the email settings to have a incorrect username or password. 
3. Click save and the Mail Provider status should now display as red/error.
3. Edit the email settings again by entering the correct username or password.
4. Click save and the Mail Provider status should now display as green/valid.


nnnnat:
The error that's being logged to the terminal is in the `release-1.8.0` branch but it being repeatedly logged is only happening in this PR branch. 

From what I can tell the `componentWillUpdate` method inside the EmailConfig component is getting called 700+ times while idling and therefore trying to call the `Meteor.call("email/verifySettings", ...);` method every time.

I'm going to have to figure out a better way of checking the email status, ether through using a different life cycle method or changing how this component get it's data.

nnnnat:
@mikemurray I got the component updated to only check the email status if the email settings have changed. I've replaced the `componentWillUpdate` method with `componentWillReceiveProps` and now checking if the email settings have changed on the `nextProps` before calling the `checkEmailStatus` method.

Right now if you repeat the testing steps above you'll see the `ERROR Reaction: Invalid login: 535 Authentication Credentials Invalid` in the terminal twice.

The first time the error shows if from this confirmation modal calling the `Meteor.call("email/verifySettings", ...)` method before saving the updated settings.
screen shot 2018-02-21 at 2 09 48 pm

After you click the "save" button, the EmailConfig component will see that the email settings have changed and will call the `Meteor.call("email/verifySettings", ...)` to update it's own state. This is when the second terminal error occurs.

@spencern @aldeed  I know we talked about a few different ways to handle the error coming from the server, [here](https://github.com/reactioncommerce/reaction/blob/ac07d02a971797863d0f593fc4608dd4b0a2611c/server/methods/email.js#L62) is where that error is being created. How do you guys feel we should best handle this and should that be part of this PR?

imports/plugins/core/email/client/containers/emailConfig.js
```diff
@@ -33,15 +33,33 @@ const wrapComponent = (Comp) => (
...
       const { service, host, port, user, password } = settings;
 
       if (service && host && port && user && password) {
         Meteor.call("email/verifySettings", (error) => {
           if (error) {
             this.setState({ status: "error" });
+          } else {
+            this.setState({ status: "valid" });
           }
-          this.setState({ status: "valid" });
```
aldeed: Anytime calling `setState` in a callback, we should get in the habit of checking to make sure the component is still mounted, to avoid errors in the console from React. The simplest way is to add componentDidMount and componentWillUmount, setting `this._isMouted` to true/false in those. And then in this callback, `if (!this._isMounted) return;` at the top.

nnnnat:
@mikemurray this is ready for review again, I've made the changes suggested above by @aldeed and fixed a linting issue.

mikemurray:
After further testing for the 1.8.1 release, I've discovered that this PR doesn't fully resolve the issue. The display will only update if the credentials, good or bad, were previously saved from either a `reaction.json` or previously entered and saved from the settings form.

Only noticed this as I tested 1.8.1 from a copy of reaction with no `reaction.json`.

mikemurray:
Created a new issue for this here: https://github.com/reactioncommerce/reaction/issues/3832

PR #3782, opened 2018-02-16
Replace CFS

### LINKED ISSUE #1221: Replace or maintain CollectionFS

aldeed:
Resolves #1221   
Impact: breaking 
Type: feature

## Issue

The Meteor CollectionFS packages are not maintained. Although they work fine in most cases, we have limited ability to troubleshoot and fix bugs, and we can't easily implement additional stores and file types.

## Solution

The Meteor CollectionFS packages are gone and are replaced with new, similar but cleaner and more modern NPM packages. These live here: https://github.com/reactioncommerce/reaction-file-collections

- The main client and server files for file handling are now in /imports/plugins/core/files
- There is now a `MediaRecords` backing collection exported with the other collections, and a `Media` MeteorFileCollection exported from the files plugin. `MediaRecords` is the equivalent of what you previously got from doing `Media.files`, i.e., the direct collection reference. But going forward it should just be imported like any other collection.
- I've consolidated some of the duplicated code around getting product images (see `getPrimaryMediaForItem`). Also `Reaction.Email.getShopLogo` for emails. There may even be places where more reuse could happen, but sometimes the Media selector is slightly different and we'd have to think about whether those differences are truly necessary.
- I created a quick React component to replace the brand image uploader/selector in Shop Settings. This probably needs some design love, but it's functional. We may want to tackle better design for it as a separate ticket.
- I did NOT remove MediaRecords collection hooks. Will create a separate ticket for that.
- Added indexes on MediaRecords collection to make those queries faster.

## Breaking changes

If you've saved the file URLs anywhere, they're now different.

```
/assets/files/:collectionName/:fileId/:filename
```

becomes

```
/assets/files/:collectionName/:fileId/:primaryStoreName/:filename
```

and

```
/assets/files/:collectionName/:fileId/:filename?store=storeName
```

becomes

```
/assets/files/:collectionName/:fileId/:storeName/:filename
```

I've deleted some unused Blaze templates rather than update URL handling within them:

- shopBrandImageOption
- ordersListItems
- select
- upload
- productMetaField
- productMetaFieldForm
- metaComponent
- productDetailEdit
- productDetailField
- productImageGallery
- imageDetail
- imageUploader
- productSocial
- variantList
- variant

Media-related publishing is changed and improved:

- "CartItemImage" publication is removed
- "CartImages" now takes an ID
- Added "ProductGridMedia" to replace Media being included with the products publication for the grid
- Added "ProductMedia"
- Added "OrderImages", similar to CartImages, used for order now rather than reusing CartImages

## Testing

Test each of these
- logged in as admin
- logged in as non-admin
- not logged in

1. Grid - Product images show up and look correct for each different view option
1. PDP - Product images look correct
1. Emails
    - make sure the correct shop logo is used in all emails, or the correct fallback logo if there isn't one

And test these only as admin:

1. Grid - thumbnail image in the panel appears
1. Shop settings
    - Click to add brand images
    - Drop brand images
    - Click brand image to change
    - Delete brand images (and if you cancel, make sure not deleted)
    - Try to upload non-images (shouldn't be allowed)
1. PDP
    - Click + to add images
    - Drop images
    - Try to upload non-images (shouldn't be allowed)
    - Drag to reorder images
    - Zoom view should work when not an admin
    - Delete product images (and if you cancel, make sure not deleted)
1. Orders
    - Make sure correct photo appears next to each product in invoice
1. Cloning products with media
    - make sure it succeeds and all images are also copied, in all stores 
1. Shopify Import
    - Make sure product images come over
1. Anywhere else that you see was touched when reviewing the code diff


aldeed:
@spencern snyk is complaining about the `google-cloud` packages used by `tus-node-server`, but we don't even use those adapters so we're not affected. However, I don't have a button to ignore that one. Are you able to ignore it?

aldeed:
@spencern Description has been updated.

I did notice that there are sometimes max recursion errors in the console when clicking around on product grid. I think it's because I'm setting `primaryMedia` and `additionalMedia` as properties on each product in `imports/plugins/included/product-variant/containers/productsContainer.js`. At some point something in Meteor must be trying to parse/clone the product document and choking on the FileRecord instances because they aren't cloneable. The solution is probably to pass through those media items as a separate prop to the React components (there are a few layers), and then let the final component that uses them (the `renderMedia` and `renderAdditionalMedia` methods of `ProductGridItems`) grab them off some other prop object instead.

spencern:
Assigning myself to try to get this across the finish line

spencern:
In regards to Blocker 1
The issue is definitely related to the way images are passed through the props as @aldeed suggests. By commenting out these two lines https://github.com/reactioncommerce/reaction/blob/feat-1221-aldeed-npm-cfs/imports/plugins/included/product-variant/containers/productsContainer.js#L211-L212
We lose images on the grid, but the grid clicking starts working again and we no longer see the Maximum call stack size exceeded issue.

I'm going to experiment with passing the product images through a different prop.

spencern:
I've resolved the issue with the product grid exceeding the max call stack size by passing product media to the grid item via a separate prop from the product.

spencern:
A couple other issues that I've stumbled upon

1. "Micro image" in action view when viewing active PDP product is stuck to whatever the last PDP image  was viewed and does not seem to be connected to the active product in the grid.
image

Screen cap here: http://recordit.co/TtDsHZ5frL

2. I don't see the proper images coming through in an order confirmation email. The images that are coming through are not showing up - that part may be in part due to the way gmail handles images in emails. But if I view the links for the images that are being wrapped by gmail, I'm seeing the placeholder image and not the imported images. I'll need to do more testing here, as it may be because of variants or options not having images selected, but I think this needs more work as well.

aldeed:
@spencern Reordering the media gallery now works when revisions are disabled. I think something weird is still happening with revisions logic, which cancels out the reordering sometime. Maybe a publication issue. I might just go ahead and move the `MediaRecords` revisions out of collection-hooks to be able to understand better what is happening, since we need to do that anyway. But feel free to take another look and let me know if the problem with revisions is obvious to you.

spencern:
@aldeed I'll take a look

aldeed:
@mikemurray The search error should be fixed now, and images appearing in search. @spencern FYI To get images showing in search, I had to move the `productMediaById` composer logic from `productsContainer` to `productGridContainer` because search uses `productGridContainer` directly.

spencern:
@mikemurray does this change impact the work you and @nnnnat are doing with the product grid?

spencern:
@aldeed In regards to the issue with rearranging images not working, there seems to be an issue with the `MediaRecords.update` call here: https://github.com/reactioncommerce/reaction/blob/feat-1221-aldeed-npm-cfs/imports/plugins/core/revisions/server/methods.js#L25

It's returning false and failing to update the MediaRecords, even if the media record is available and has the correct ID, and the `Revisions.update` that happens a few lines down succeeds.

mikemurray:
@spencern, @nnnnat and I removed all logic from the `productsContainer` and split it into `productsContainerAdmin` and `productsContainerCustomer`. `productsContainer` is now just used to determine if you're an admin or customer and serves the correct product grid.

PR #3789, opened 2018-02-19
Fix: Email verification link for admin user

### LINKED ISSUE #3714: Email verification link for admin user does not work.
### Expected behavior
When defining the environment variable REACTION_SECURE_DEFAULT_ADMIN=true and start a stock Reaction with empty database, a new admin user should be created. Depending on wether a mail  host is defined or not the console should show a verification link (or sent it to the mailhub respectively), which can be copied into the browser's address bar. After loading that link, the user's email should be verified (in database)

### Actual behavior
When pasting the verification link into the browser's address bar, the index route loads. There's no visual feedback or anything that signals the user if the verification process was successful or not.
In fact, the verfication does not happen at all, because the email address is still unverified in database.

1) Clone reaction
2) export REACTION_SECURE_DEFAULT_ADMIN=true 
3) reaction reset -n && reaction
4) Watch out in console window for something along the lines
```
  ***************************************************
            IMPORTANT! EMAIL VERIFICATION LINK

             Email sending is not configured.

    Go to the following URL to verify email: hpwgdbe3@localhost

    http://tsp.dyn.projekt-und-partner.de:3000/#/verify-email/OLzrJJbwW_Bb7KcIz1hufcdH9GxdDX9C7j7zwl80Smw
    ***************************************************
```

5) copy the link adress and paste it in the browser window
6) See nothing happen – apart from the index page being loaded
6) Browse mongodb. Both Accounts and Users collection will still show the unverified email addresses. Correct would be: `Accounts.emails.$.verified : true` and `Users.emails.$.verified: true`

prinzdezibel:
This PR resoves issue #3714

Resolves #3714  
Impact: minor  
Type: bugfix

## Issue
When pasting the verification link into the browser's address bar, the index route loads. There's no visual feedback or anything that signals the user if the verification process was successful or not.
In fact, the verfication does not happen at all, because the email address is still unverified in database.

## Solution
The previously used hash URL did not seem to work at all. It just redirects to the
index route and leaves the verified fields in database in false state.

The fix is to use the verification URL that's used for other users as well. This will provide visual feedback and set the verified fields in db to true.

## Breaking changes
none expected

## Testing
1) Clone reaction
2) export REACTION_SECURE_DEFAULT_ADMIN=true 
3) reaction reset -n && reaction
4) Watch out in console window for something along the lines
```
  ***************************************************
            IMPORTANT! EMAIL VERIFICATION LINK

             Email sending is not configured.

    Go to the following URL to verify email: hpwgdbe3@localhost

    http://localhost:3000/account/profile/verify?email=testing%40reactioncommerce.com
    ***************************************************
```

5) copy the link adress and paste it in the browser window
6) A verification success notification should be displayed.
6) Browse mongodb. Both Accounts and Users collection shoud show the verified email addresses. You should see the following values: `Accounts.emails.$.verified : true` and `Users.emails.$.verified: true`

### DIFF
--- a/server/api/core/accounts/password.js
+++ b/server/api/core/accounts/password.js
@@ -172,7 +172,9 @@ export function sendVerificationEmail(userId, email) {
   });
 
   const shopName = Reaction.getShopName();
-  const url = Accounts.urls.verifyEmail(token);
+
+  const encodedEmailAddress = encodeURIComponent(address);
+  const url = `${Meteor.absoluteUrl()}account/profile/verify?email=${encodedEmailAddress}`;
 
   const dataForEmail = {
     // Reaction Information
@@ -289,7 +291,9 @@ export function sendUpdatedVerificationEmail(userId, email) {
   });
 
   const shopName = Reaction.getShopName();
-  const url = Accounts.urls.verifyEmail(token);
+
+  const encodedEmailAddress = encodeURIComponent(address);
+  const url = `${Meteor.absoluteUrl()}account/profile/verify?email=${encodedEmailAddress}`;
 
   const dataForEmail = {
     // Reaction Information


prinzdezibel:
@jshimko Agree. I was asking the same, but then thought it must be an ok thing, because the all profiles are verified that way.

jshimko:
All of the guts of the Meteor accounts/email stuff was pulled into the app (in the file you edited here) back when I rewrote all the `Email.send()` stuff to use Nodemailer, so I don't see any reason why we can't just fix what's there and takes the safer route with tokens.  Most of the work is already done (because it definitely used to work), so it'd be a matter of fixing whatever broke and then making the UI experience a little better when someone visits the link.  Probably not a lot to work out there.

So essentially, you could replace Meteor's [`Accounts.urls.verifyEmail(token)`](https://github.com/reactioncommerce/reaction/blob/2f128a1f7253f09be87f12ca1430bb4a43bd9484/server/api/core/accounts/password.js#L175) with whatever custom route you want (with the token included, of course), then set up the client side route and manually process that token with Meteor's [`Accounts.verifyEmail(token, callback)`](https://docs.meteor.com/api/passwords.html#Accounts-verifyEmail).  That also has the useful side effect of automatically logging the user in (which is what used to happen with the admin verification before it was broken).  In that callback, you could do whatever notification junk you want in the UI.

...or whatever.  Something like that.

prinzdezibel:
@jshimko right, let me tackle this tomorrow. I guess, it would make sense for the all verification links, right?

jshimko:
Yep.  Seems reasonable to me.

spencern:
@prinzdezibel is this ready for another review?

prinzdezibel:
@spencern Yes, Brent was proposing to review this as is and file another issue that addresses the general security concerns, that @jshimko brought up.

zenweasel:
@jshimko, what do you think? Should we fix all of these in this ticket or open a separate ticket for the security issues in the other email verification links?

jshimko:
I suppose we could make this PR only do the original intention, but it should definitely be a high priority to then completely undo this work and get the security hole patched.  It might be easier to just close this and fix it the right way.  Or change the title of this PR if that somehow makes more sense.

prinzdezibel:
Agree with @jshimko . Should be done the proper way for admin and non-admin. Closing this PR.

PR #3794, opened 2018-02-20
Fixes #2810: /shop added to URL

### LINKED ISSUE #2810: Link for "Your order was accepted" is broken
`getShopPrefix()` in https://github.com/reactioncommerce/reaction/blob/master/server/api/core/core.js#L201
 
always returns the shop name as slug.

Plugins like the notifcation view use this method to determine the route.
For single shops this results in a **not found** page, as the notifaction route is registered as **/notifcations**,
and the Link routes to **/shopname/notifications**.

I see two options here:

1.  Modify `getShopPrefix()` to return "" for single shops. In /lib/api/helpers.js we had a similar issue with the getShopPrefix method and Paypal: https://github.com/reactioncommerce/reaction/pull/2563/files#diff-7143087245674b35fd0f58f5c1e8a70cR41

2. Change the way getShopPrefix is used. Maybe leave it untouched, but replace the parts where it is used with an additional method, that first checks shop count and invokes it if necessary. Could possibly prevent additional things from breaking.

### Steps to Reproduce
1. Configure shop to take order (configure shipping/payment)
1. Place an order
1. Return to home page
1. Click on the "bell" icon in the top nav
1. Click on the "Your Order was accepted" link
1. Observe that you get a 404 error

Akarshit:
Resolves #2810  
Impact: minor  
Type: bugfix

## Issue
The route returned was not prefixed with `/shop` hence was resulting in a 404 page

## Solution
Added `/shop` to the route being returned. This is done by added a param `/shopPrefix` to the marketplace settings.

## Testing
1. Configure shop to take order (configure shipping/payment)
1. Place an order
1. Return to home page
1. Click on the "bell" icon in the top nav
1. Click on the "Your Order was accepted" link
1. Observe that you get to the notifications page.

I am not sure the notifications page is where you are supposed to land but from the code it looks like that was the intended behaviour.


Akarshit:
@aaronjudd I tried using the way you suggested, but getting stuck using Router on server side. 
Will fix this on Monday.

Akarshit:
Just had a conversation with Mike, he suggested that it would be best to move the hardcoded `/shop` into the MarketPlace package settings. 
Working on it.

client/modules/core/main.js
```diff
@@ -474,7 +474,7 @@ export default {
   getShopPrefix() {
     const shopName = this.getShopName();
     if (shopName) {
-      return `/${this.getSlug(shopName.toLowerCase())}`;
+      return `/shop/${this.getSlug(shopName.toLowerCase())}`;
```
pmn4: I know that this is not the point of the PR, but a Shop _has_ a slug property. Computing one from the name feels like a bug.
(less important, but `slugify`, the underlying method for `getSlug` lowercases by default)

Akarshit: @pmn4 Thanks for pointing that out. I was not sure if I should change that in this PR. 
But maybe I should.

Akarshit:
@jshimko Just for sanity check - This would need a migration where `shopPrefix: "/shop"` is added to the `Package` collection in the DB?

jshimko:
@Akarshit I think @spencern might be more qualified to weigh in on that detail.  Added him as a reviewer.

Akarshit:
Thanks :)

Akarshit:
I think migration would be necessary, so I have added them in the PR. If someone is going to have a look at this, it would make sense to look at it with the migration.

Akarshit:
Tested with marketplace, works fine.

zenweasel:
I've tested and verified that this works as expected. 👍

PR #3801, opened 2018-02-20
Update React to 16.2.0

### LINKED ISSUE #3495: Upgrade to React 16
Got some queries from the community about when we are going to do this as people want to utilize some of the new features. This is a placeholder for any issue we encounter when trying to do that.

Theoretically this should be a painless upgrade but it depends on whether all of the dependent libraries are compatible.

mikemurray:
Resolves #3495  
Impact: breaking
Type: chore

## Issue

The current version of React in the `master` branch is `15.6.2`. This PR is to update to the `16.2.0` release of React and any incompatible modules.

## Solution

- Update `react` and `react-dom` to `16.2.0`
- Use `immutability-helper` instead of `react/lib/update` package

## Breaking changes

**Breaking change 1**

`react/lib/update` has been removed from React 16.2.0. Updated all imports to use https://github.com/kolodny/immutability-helper as a drop-in replacement.

`import update from "react/lib/update";`  -> `import update from "immutability-helper";`

**Potential breaking change 2**

With `compseWithTracker` in `imports/plugins/core/components/lib/composer/compose.js`; removed the setting of `ref` on `<Child />` component.

In React 16 it throws a warning that setting `ref` on stateless functional components will fail. With some research, it seems this was the case since React 15, only that no warning was displayed. 

This potentially never work 100% of the time, so it has been removed. We very rarely use `refs` in Reaction, and never on a container component.

## Testing

1. Run through the entire application clicking on everything everywhere
1. Add products to cart and checkout
1. As an admin fulfill orders, sort the orders table view
1. As an admin, add products to the grid, drag and drop re-order the grid items
1. Run through acceptance tests https://docs.reactioncommerce.com/reaction-docs/master/acceptance-testing


mikemurray:
Added release 1.8.0 as a blocker. This PR will need to be rebased once it's been merged into `master`.

zenweasel:
Saw this issue: https://github.com/kadirahq/subs-manager/issues/93 
Possibly something to double-check for

mikemurray:
@kieckhafer are these issues present in `master`?

kieckhafer:
@mikemurray No, these are new issues I'm seeing.

`master`, showing product images and gravatar working:

blank

mikemurray:
@kieckhafer Is there anything special with your gravatar config? I'm not seeing the issues you pointed out in my own gravatar in the navbar or orders.

kieckhafer:
No, nothing special, just a straight email address.

Just tested again, everything was OK.

Then a third time, and same issue.

So I can't reproduce every time, might just be an existing issue I haven't seen before.

mikemurray:
@kieckhafer I'm unable to reproduce the product image issue you're seeing. I duplicated the basic product on the grid, removed and added a new image, once to the top level variant, and one to a child variant. In both cases, the image on the grid was updated properly.

The `ref` change was only for the container components themselves and only if you try to import and use a container component, and then try to put a ref on the container component itself. 

The `ref` in the `MediaGallery` is on a `div` not a container component.

PR #3802, opened 2018-02-21
(fix): Errors when updating default shipping and billing addresses

### LINKED ISSUE #3672: Choosing between several Shipping/Billing addresses is blocked by UI lag
### Screenshot gifs

| Expected: 1.7   	| Actual: 1.8     	|
|----------	|:------------:	|
| out2| out|
|In 1.7, choosing between Shipping/Billing addresses should be fast and responsive. After I click on a different address, I see the blue icon to indicate that that address has been chosen.|In 1.8, after I click on a different address, the blue icon sometimes does not move. The UI is blocked by an operation on the back-end: tax calculation.|

1. Go through the steps of checking out as an anonymous or logged in user
2. Add a second address during Step 2
3. Toggle between the two addresses

### Versions

```
Node: 8.2.1
NPM: 5.6.0
Meteor Node: 8.9.3
Meteor NPM: 5.5.1
Reaction CLI: 0.27.0
Reaction: 1.8.0
Docker: 17.09.0-ce
```

Browser: http://whatismybrowser.com/w/KE449ZE

### QA

QA by @willopez + @machikoyasuda

kieckhafer:
Resolves #3672
Impact: major
Type: bugfix

## Issue
When changing default shipping and billing addresses, if addresses were updated multiple times in succession, the UI would eventually stop working, causing the user to be stuck with whatever address they were currently locked on.

## Solution
The bulk of the issue seems to stem from our `afterAccountsUpdate` hook, which is used to update the Accounts search record. This hook was running three times every time the `addressBookUpdate` function was run: once for the shipping address, once for the billing address, and once at the end of the function.

The following error is returned when changing the address too much in too short of a span.

```
Exception while invoking method 'accounts/addressBookUpdate' WriteError({"code":11000,"index":0,"errmsg":"E11000 duplicate key error collection: meteor.AccountSearch index: _id_ dup key: { : \"oW372kXYR8Yi3gE2M\" }","op":{"_id":"oW372kXYR8Yi3gE2M","shopId":"J8Bhq3uTtdgwZx3rz","emails":[--------------address info redacted--------------)
    at Function.MongoError.create (/Users/ek/.meteor/packages/npm-mongo/.2.2.34.6do6sz.tll37++os+web.browser+web.cordova/npm/node_modules/mongodb-core/lib/error.js:31:11)
    at toError (/Users/ek/.meteor/packages/npm-mongo/.2.2.34.6do6sz.tll37++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/utils.js:139:22)
    at /Users/ek/.meteor/packages/npm-mongo/.2.2.34.6do6sz.tll37++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/collection.js:668:23
    at handleCallback (/Users/ek/.meteor/packages/npm-mongo/.2.2.34.6do6sz.tll37++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/utils.js:120:56)
    at resultHandler (/Users/ek/.meteor/packages/npm-mongo/.2.2.34.6do6sz.tll37++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/bulk/ordered.js:421:14)
    at /Users/ek/.meteor/packages/npm-mongo/.2.2.34.6do6sz.tll37++os+web.browser+web.cordova/npm/node_modules/mongodb-core/lib/connection/pool.js:469:18
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickDomainCallback (internal/process/next_tick.js:218:9)
```

That error seems to stem from [here](https://github.com/reactioncommerce/reaction/blob/release-1.8.0/imports/plugins/included/search-mongo/server/methods/searchcollections.js#L287-L293). The `_id` is a required field, so if it's being inserted twice - which may happen if the function is called twice (or three times, as stated above) - it will throw the error.

Because of this error, the remainder of the function is never run, which would cause instances where the new address has updated to set `isShippingDefault` and `isBillingDefault` to equal `true`, and never unset the older addresses to be false, so both sets of addresses were set to the defaults.

The fix here seems to be to only run the `afterAccountsUpdate` hook once, at the end of the `addressBookUpdate` function. This will ensure all data is updated as it should be before the hook is run, and not run `afterAccountsUpdate` unnecessarily on every change, instead running it after all the changes are complete.

In addition, I've reworked the `afterAccountsUpdate` hook to force the second parameter to be an object, which contains the user `_id` (which was being passed as a solo string before), as well as an `Array` of fields of the users profile that were updated prior to this hook running.

Because our search only searches `emails`, `firstName`, `lastName`, and `phone`, it does not make sense to re-index the search if a field other than these is updated.

The updated fields `Array` is run against the `Array` of fields used in search, and if a match is found, then the indexing is run.

In addition, I've added `forceIndex` as an option on the updated fields array, so that can be passed at anytime, and is also passed in various instances, such as when a new address is created or an address is deleted.

## Breaking changes
The `afterAccountsUpdate` hook now requires the second parameter to be an object, instead of a string of just the user id. The object still contains `accountId`, and also contains `updatedFields`.

## Testing
1. Log in to your Reaction account
2. Navigate to your Profile page
3. Add multiple addresses to your account
4. Click on the addresses in random orders, making different addresses the default
5. See that you can update your address
6. Repeat above steps in address book of the checkout


imports/plugins/included/connectors-shopify/server/methods/import/customers.js
```diff
@@ -170,7 +170,10 @@ export const methods = {
             ids.push(reactionCustomerId);
 
             Accounts.update({ _id: reactionCustomerId }, { publish: true });
-            Hooks.Events.run("afterAccountsUpdate", Meteor.userId(), reactionCustomerId);
+            Hooks.Events.run("afterAccountsUpdate", Meteor.userId(), {
+              accountId: reactionCustomerId,
+              updatedFields: ["forceUpdate"]
```
aldeed: You mean `forceIndex`?

imports/plugins/included/search-mongo/server/methods/searchcollections.js
```diff
@@ -261,13 +261,48 @@ export function buildOrderSearch(cb) {
...
+  const account = Accounts.findOne(accountId);
+  // let's ignore anonymous accounts
+  if (account && account.emails && account.emails.length) {
+    const accountSearch = {};
+
+    // Not all required fields are used in search
+    // We need to filter through fields that are used,
+    // and only update the search index if one of those fields were updated
+    // forceIndex is included to forceIndexing on startup, or when manually added
+    const searchableFields = ["forceIndex", "emails", "firstName", "lastName", "phone"];
```
aldeed: `requiredFields.accounts` also includes "shopId" so that should be listed here?

imports/plugins/included/search-mongo/server/methods/searchcollections.js
```diff
@@ -261,13 +261,48 @@ export function buildOrderSearch(cb) {
...
+      for (const field of requiredFields.accounts) {
+        if (transformations.accounts[field]) {
+          accountSearch[field] = transformations.accounts[field](account[field]);
+        } else {
+          accountSearch[field] = account[field];
+        }
+      }
+      AccountSearch.insert(accountSearch);
+      const rawAccountSearchCollection = AccountSearch.rawCollection();
+      rawAccountSearchCollection.createIndex({ shopId: 1, emails: 1 }).catch(handleIndexUpdateFailures);
```
aldeed: I don't see why we're creating this index every time we insert a record. Only needs to happen once on server startup.

server/methods/accounts/accounts.js
```diff
@@ -517,13 +530,30 @@ export function addressBookUpdate(address, accountUserId, type) {
...
   }
 
+  // Update the Meteor.users collection with new address info
   Meteor.users.update(Meteor.userId(), userUpdateQuery);
 
+  // Update the Reaction Accounts collection with new address info
   const updatedAccount = Accounts.update({
     userId,
     "profile.addressBook._id": address._id
   }, accountsUpdateQuery);
```
aldeed: This is OK, but it wouldn't be that hard to eliminate the earlier two `Accounts.update` calls entirely. It already starts with a `findOne` setting `account`. So in place of the earlier `Accounts.update` calls, you can loop through all of `account.profile.addressBook` array and set every `true` isBillingDefault or `isShippingDefault` to `false`. Then down here, replace one of the `account.profile.addressBook` items with the updated `address` based on its ID.

Then you only need one final actual DB update, doing `{ $set: { "profile.addressBook": account.profile.addressBook } }`

server/methods/accounts/accounts.js
```diff
@@ -1058,7 +1092,10 @@ export function setProfileCurrency(currencyName) {
   check(currencyName, String);
   if (this.userId) {
     Accounts.update(this.userId, { $set: { "profile.currency": currencyName } });
-    Hooks.Events.run("afterAccountsUpdate", this.userId, this.userId);
+    Hooks.Events.run("afterAccountsUpdate", this.userId, {
+      accountId: this.userId,
+      updatedFields: ["profile.currency"]
```
aldeed: Here you are passing in "profile.currency" as the field, but in `buildAccountSearchRecord` you are looking for fields like "firstName" rather than "profile.firstName", and for the address update, you're passing in the address fields directly without `profile.addressBook` prefix. So there needs to be consistency about whether you are passing in the full object path or just the last piece.

To simplify, you could just pass in the top-level field that changed, i.e. "profile". The downside is that you may then run the AccountSearch stuff more than you need to, but it would be easier to properly build the `updatedFields` array. I'm fine with whatever as long as it works and is consistent.

imports/plugins/included/connectors-shopify/server/methods/import/customers.js
kieckhafer: Yes.

imports/plugins/included/search-mongo/server/methods/searchcollections.js
kieckhafer: I don't think shopId should be here.

Based on the [transformations file](https://github.com/reactioncommerce/reaction/blob/fix-3672-kieckhafer-billingShippingAddressChangeBugs/imports/plugins/included/search-mongo/server/methods/transformations.js#L39), it seems that what I've listed here are the only `searchable` fields.

The `required` fields are different - not all are searchable. I'm trying to only update the index when one of the searchable fields is update, not when a required field is updated.

imports/plugins/included/search-mongo/server/methods/searchcollections.js
kieckhafer: I think you're right, the code on line 309 should take care of this, it runs via a build job.

server/methods/accounts/accounts.js
kieckhafer: I thought about passing the `profile` down, but as you mentioned, that will cause the it to run more frequently. All the addressBook is inside the `profile`, so it would actually run on every update, and would essentially be no different than what we are currently doing.

I'm going to go with just the last piece.

imports/plugins/included/search-mongo/server/methods/searchcollections.js
aldeed: OK makes sense

aldeed: Except... if the `shopId` changes in the account, doesn't it also need to change in the `AccountSearch` for security reasons? I'm guessing this is a rare thing, but best to be safe.

kieckhafer:
@aldeed I've moved the removal into the `buildAccountSearchRecord`, but I've been experiencing issues with using `upserts` the entire time doing this issue.

`Exception in setTimeout callback: Error: Invalid modifier. Modifier must be an object.`

I get that on startup, and when trying to update. Everything works fine when doing a removal and insert.

server/methods/accounts/accounts.js
```diff
@@ -449,57 +461,49 @@ export function addressBookUpdate(address, accountUserId, type) {
...
   }
 
+  // Update all addresses to set new default address to true, and the rest to false
+  account.profile.addressBook.forEach((addr) => {
+    if (addr._id === address._id) {
+      Object.assign(addr, { [type]: true });
+    } else {
+      Object.assign(addr, { [type]: false });
+    }
+  });
```
aldeed: This changes the behavior a bit from before, in that you are no longer updating the whole address object where `addr._id === address._id`. Since you've already done `Object.assign(address, { [type]: true });` above, I think you can do this:

```js
account.profile.addressBook.forEach((addr) => {
  if (addr._id === address._id) {
    Object.assign(addr, address);
  } else if (typeof type === "string") {
    Object.assign(addr, { [type]: false });
  }
});
```

Note that I also added a check for whether `type` is a string to the `else`. That's because it seems like this is intended to work without passing a type also, if you just want to update the address.

aldeed:
I am approving the code only. @spencern can you test the change in the app?

spencern:
@kieckhafer you've got a failing test, looks directly related to this PR.

```
329 passing (2m)
1 failing

1) Account Meteor method  addressBookUpdate should disable isShipping/BillingDefault properties inside sibling address if we enable them while editing:
   AssertionError: expected true to be false
    at Assertion.<anonymous> (packages/practicalmeteor_chai.js:1252:10)
    at Assertion.get (packages/practicalmeteor_chai.js:4095:29)
    at Test.<anonymous> (server/methods/accounts/accounts.app-test.js:391:9)
    at run (packages/practicalmeteor:mocha-core/server.js:56:16)
    at Context.wrappedFunction (packages/practicalmeteor:mocha-core/server.js:63:33)
    at run (packages/practicalmeteor:mocha-core/server.js:52:15)
```

kieckhafer:
@spencern Test is fixed.

Issue #3617, opened 2018-01-31
Marketplace - allow sellers to signup - error on "Become a seller" click


dancastellon:
### Expected behavior
When Marketplace is enabled, and the checkbox is checked to allow users to register themselves as sellers, users should be able to create a shop.

### Actual behavior
When "Become a seller" is clicked in a user's profile, the following error appears in console and the entire site is blank from there on out.

TypeError: Cannot read property 'name' of undefined
    at TagNav.renderShopSelect (tagNav.js:67)
    at TagNav.render (tagNav.js:108)

### Steps to reproduce the behavior

1. npm init, cd reaction, reaction
2. Login as admin, enable Marketplace, check checkbox to allow users to register as sellers
3. Logout and register a new acccount
4. Visit profile, click "Become a seller"

### Versions
(run `reaction -v` from your reaction directory)
Node: 6.2.2
NPM: 5.0.3
Meteor Node: 8.9.3
Meteor NPM: 5.5.1
Reaction CLI: 0.26.4
Reaction: 1.7.0
Reaction branch: master
Docker: 17.09.1-ce

zenweasel:
This is probably the same bug as #3652

JoeyLyman:
I get this too

PR #3809, opened 2018-02-21
(fix: 3619) Fix for: Shop switcher opens off-screen

### LINKED ISSUE #3619: Shop Switcher Opens Off-Screen
It seems that once you get over `n` Shops, the Shop Switcher becomes too long, and is positioned above the label. (I'm not positive what _`n`_ needs to be to make this happen, but I have 24 currently)

image

### Expected behavior
The drop-down to appear on the screen (and be scrollable, I guess?)

### Actual behavior
It appears above the selector, but since that is located in the navbar, it is rendered off screen, above the top of the page (see screenshot).

Create 24 Shops on a 13" MPB (that may not be the exact number, but that is what I have now and notice the issue)

### Versions
`master`

prinzdezibel:
Resolves #3619  
Impact: major
Type: bugfix

## Issue
In a marketplace setup with many shops the dropdown popover can go offscreen completely if the window height is small enough.

## Solution
This PR prevents the shop switch dropdown to go off screen. The tether
library has an option for this : "pin". That makes sense in this case,
because the nav bar is also fixed and won't go off-screen. Not sure
if it makes sense to put this option to all dropdowns in general.

## Breaking changes
none expected

## NOTICE

This PR requires  [#3617](https://github.com/reactioncommerce/reaction/pull/3725/files) to be applied.

## Testing
1. As an admin, set up marketplace and add approx. 10 shops
2. Activate them
3. Navigate to index route
4. Reduce window height
5. Click on shop selector
6. Dropdown list should not go offscreen
7. Dropdown should be scrollable


prinzdezibel:
@jshimko No matter what I try, I can't reproduce this scrollbar artifact you're seeing. But I find it interesting that your font looks very blurry and aliased. Maybe there's some number rounding going on behind the scenes that makes this scrollbar appearing.

jshimko:
@prinzdezibel  Hmm.  I can't _not_ reproduce it.  I get it in every browser.

reaction_and_reaction_and_reaction_and_reaction

jshimko:
And the font rendering issue was on my end because that particular browser tab was zoomed in to 110%.  Zooming back to 100% made the font crisp again.

prinzdezibel:
@zenweasel I'm running out of ideas. Can you test that in your environment?

prinzdezibel:
Going to pair with Jeremy on this one ASAP.

prinzdezibel:
Gotcha! It worked on my machine because of this system setting:

general_and_developer_tools_-_http___localhost_3000__and_basic_reaction_product

Changed from 'scroll' to 'auto'. Please try again, @jshimko!

jshimko:
Resolved.  Thanks!

PR #3824, opened 2018-02-22
(fix: 3783) Discount codes limits are not honored

### LINKED ISSUE #3783: Discount codes account limit and total limit not honored
### Expected behavior

Discount `Account Limit` and `Total Limit` should be honored.

### Actual behavior

Discount codes can be used past their limits.

cart_completed

1. Add a discount code in payment settings: see screenshot
2. Check out and apply the code
3. Code should be applied successfully, complete checkout
4. Check out again and apply the code once more
5. The code is applied again and should not be. (Not only is the account limit 1, to total uses is 1 as well)
6. Complete checkout, and view orders dashboard
7. There should be two orders both with the discount applied.

This happens on both `master` and `release-1.8.0` branch.

### Versions
Node: 7.2.0
NPM: 5.6.0
Meteor Node: 8.9.3
Meteor NPM: 5.5.1
Reaction CLI: 0.27.0
Reaction: 1.7.0
Reaction branch: master
Docker: 17.12.0-ce

prinzdezibel:
Resolves #3783  
Impact: major  
Type: bugfix

## Issue
Discount codes could be applied without limits, i.e. they don't expire.

## Solution
  In method cart/submitPayment any existing billing.paymentmethod (e.g. paymentmethod for discount codes) was overwritten.

I guess this hasn't been working since introducing marketplace (just a wild guess). The solution was to retain any already existing billing.paymentmethods in the array.

NOTICE: Because at various places (copyCartToOrder) the assumption is that all relevant billing data (address, shopId, currency, etc.) is on billing[0] , this does only work if it's appended at the end of the
array. I think it's not good to have this order predefined, but hesitated to dig deeper here without a mandate.

Also I'm not sure if the discout mechanism in general plays well with the marketplace feature.

## Breaking changes
none expected

## Testing
1. As an admin, create a discount code with limit 1
2. As a user, apply the code and checkout
3. As a user, order a second time and checkout again.
4. Try to re-apply the code.
5. A message should appear "Expired code"

### DIFF
--- a/imports/plugins/core/discounts/client/components/list.js
+++ b/imports/plugins/core/discounts/client/components/list.js
@@ -74,17 +74,13 @@ function composer(props, onData) {
   });
 
   const listItems = [];
-  for (const billing of currentCart.billing) {
-    if (billing.paymentMethod && billing.paymentMethod.processor === "code") {
-      listItems.push({
-        id: billing._id,
-        code: billing.paymentMethod.code,
-        discount: billing.paymentMethod.amount
-      });
-      // Found discount payment method. break here for good measure, just in case there's another "code" payment method.
-      // Which I think can happen only if "cart/submitPayment" is executed multiple times, e.g. if DDP connection broke.
-      break;
-    }
+  const listItem = currentCart.billing.find((element) => element.paymentMethod && element.paymentMethod.processor === "code");
+  if (listItem) {
+    listItems.push({
+      id: listItem._id,
+      code: listItem.paymentMethod.code,
+      discount: listItem.paymentMethod.amount
+    });
   }
 
   onData(null, {
--- a/imports/plugins/core/discounts/client/components/list.js
+++ b/imports/plugins/core/discounts/client/components/list.js
@@ -81,6 +81,9 @@ function composer(props, onData) {
         code: billing.paymentMethod.code,
         discount: billing.paymentMethod.amount
       });
+      // Found discount payment method. break here for good measure, just in case there's another "code" payment method.
+      // Which I think can happen only if "cart/submitPayment" is executed multiple times, e.g. if DDP connection broke.
+      break;
     }
   }
 
--- a/server/methods/core/cart.js
+++ b/server/methods/core/cart.js
@@ -1102,14 +1102,14 @@ Meteor.methods({
     // we'll just update the workflow and billing data where
     // method-hooks can process the workflow update.
 
-    const payments = [];
-    let paymentAddress;
-
     // Find the payment address associated that the user input during the
     // checkout process
+    let paymentAddress;
     if (Array.isArray(cart.billing) && cart.billing[0]) {
       paymentAddress = cart.billing[0].address;
     }
+
+    const payments = [];
 
     // Payment plugins which have been updated for marketplace are passing an array as paymentMethods
     if (Array.isArray(paymentMethods)) {
@@ -1149,6 +1149,9 @@ Meteor.methods({
         shopId: Reaction.getPrimaryShopId()
       });
     }
+
+    // e.g. discount records would be already present on the billing array. Add to the end of the array.
+    Array.prototype.push.apply(payments, cart.billing.filter((billingInfo) => billingInfo.paymentMethod));
 
     const selector = {
       _id: cartId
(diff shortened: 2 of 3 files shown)


zenweasel:
@jshimko I'm not sure we support Codes as the only payment method, I think they are supposed to be combined with another payment method

prinzdezibel:
@jshimko  I also don't think that the Discount payment provider was ever intended to be used stand-alone. It doesn't even feature a submit button that triggers "cart/paymentSubmit". But was fixing the errors you observed. The error happened, whenever a code was revoked again, because the code billing record had the same _id as the address billing record. So a mongo pull purged both of them. This can be seen in your screenshot, because the address billing record is not selected (checked) with the green icon any longer.

jshimko:
So are we saying there isn't a bug here?  Even if it's because another payment method isn't turned on, shouldn't we have some kind of feedback indicating that?

aaronjudd:
True that discount codes are not meant to trigger the checkout process, but there should be multiple of them allowed, until it hits one of the usage limits..

prinzdezibel:
@aaronjudd  You mean something similar to what is shown in the below screenshot?
A way to add arbitrarily many codes during checkout? 

discount

If so, I'd propose to create a new task for that and fix the issue at hand independently of that.

prinzdezibel:
@jshimko Would it help to have a hint in the payment settings for discount codes that this method is intended to be used in combination?

aaronjudd:
@prinzdezibel It was intentional that we'd only allow one code entered in the UI, I chose not to implement the UI in your screenshot, not that I don't see it being useful for the cart (I do) -> but because it was more work than I had time for at the time to think about how the order fulfillment flow would handle multiple discounts -> so I choose to leave it for now at a single entry - so if we do have a need to add that UX you're right it should be a new issue.   

Structurally though, yes, multiple codes should be able to applied (albeit, not the same one multiple times).

@jshimko you're probably right, there should be a warning that no payment methods are enabled.. although I don't think that's a new bug, but rather an enhancement that was never implemented, so also perhaps a new issue?

prinzdezibel:
If I got that right, this is ready for another review, @jshimko  and I would create a new ticket for a warning that Discounts are intended to be used in combination with another payment provider, ok?

prinzdezibel:
Ticket created. Can you please review this, @jshimko ?

server/methods/core/cart.js
```diff
@@ -1150,6 +1150,9 @@ Meteor.methods({
       });
     }
 
+    // e.g. discount records would be already present on the billing array. Add to the end of the array.
+    Array.prototype.push.apply(payments, cart.billing.filter((billingInfo) => billingInfo.paymentMethod));
```
spencern: This is clever but also a little difficult to read.
What are we filtering here, billingInfo that has paymentMethod?

Would this be identical in function?
```js
const paymentMethods = cart.billing.filter((billingInfo) => billingInfo.paymentMethod));
payments.push(...paymentMethods)
```

Looking at it further, it seems that payments is always initialized to an empty array. What is the difference between `payments` and the result of the `cart.billing.filter` function?

e.g.
```
const payments = cart.billing.filter((billingInfo) => billingInfo.paymentMethod));
```

prinzdezibel: Agree, the spread operator looks much nicer. Changed.

What is the difference between payments and the result of the cart.billing.filter function?

Difference is, that between line 1115 and 1150 additional transaction records are pushed to the payments array.

imports/plugins/core/discounts/client/components/list.js
```diff
@@ -81,6 +81,9 @@ function composer(props, onData) {
         code: billing.paymentMethod.code,
         discount: billing.paymentMethod.amount
       });
+      // Found discount payment method. break here for good measure, just in case there's another "code" payment method.
+      // Which I think can happen only if "cart/submitPayment" is executed multiple times, e.g. if DDP connection broke.
+      break;
```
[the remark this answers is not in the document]
prinzdezibel: Changed.

> Is listItems always an array of one item in this instance?
Yes, although it looks like the array was used to accommodate potentially more than one code, in reality the component just supports one discount code at max.

prinzdezibel:
@spencern  Please follow up on this if you have time.

PR #3826, opened 2018-02-22
Fix 3702 order refunding number input

### LINKED ISSUE #3702: JS NaN error when refunding individual items in order processing
### Expected behavior
When processing an order, and trying to refund individual items, I should be able to use both the up / down quantity arrows, as well as type in a number to the input field to change the quantity.

### Actual behavior
When processing an order, and trying to refund individual items, I can use the up / down quantity arrows to change quantity, but when I try to type a number inside the input box, I get a `NaN` error.

- Create an order with multiple quantities of the same variant / option
- Begin to process the order
- Open the individual item processing flow by clicking the product image inside the processing dashboard
- Attempt to change the quantity by typing a number in the input box

unknown-1

### Versions

```
Node: 9.5.0
NPM: 5.6.0
Meteor Node: 8.9.3
Meteor NPM: 5.5.1
Reaction CLI: 0.27.0
Reaction: 1.8.0
Docker: 18.01.0-ce
```

### QA

QA by @kieckhafer + @cassytaylor

nnnnat:
Resolves #3702  
Impact: **major**  
Type: **bugfix**

## Issue
The NumberTypeInput wasn't aware of input value changes that didn't come from the up/down arrow buttons.

The inputs were also not displaying properly in Firefox
screen shot 2018-02-21 at 5 14 50 pm

## Solution
* I fixed the Firefox issue by updating it's `flex` property to be `0 0 auto` so the input won't try to fill the available space.
* Updated the `handleChange` method to check if the param value is undefined. If it is set the value to be `event.target.value` and then update the components state. Now the component will be aware of any value changes from the keyboard.
* Removed the handler binding in the `constructor` since all the handler methods are using arrow function I figured the binding wasn't needed.
* Added a disabled property to the down arrow button that disables the button if the input's `value` and `minValue` are equal.

## Testing
* Create an order with multiple quantities of the same variant / option
* As an Admin begin to process the order
* Open the individual item processing flow by clicking the product image inside the processing dashboard
* Attempt to change the quantity by typing a number in the input box

**Notes**
RC test passed
Was able to complete acceptance test

### DIFF
--- a/imports/plugins/core/orders/client/containers/invoiceContainer.js
+++ b/imports/plugins/core/orders/client/containers/invoiceContainer.js
@@ -182,7 +182,7 @@ class InvoiceContainer extends Component {
       if (refundedQuantity !== 0) {
         editedItems.push(isEdited);
       }
-    } else {
+    } else if (refundedQuantity !== 0) {
       editedItems.push({
         id: lineItem._id,
         title: lineItem.title,
--- a/imports/plugins/core/ui/client/components/numericInput/numberTypeInput.js
+++ b/imports/plugins/core/ui/client/components/numericInput/numberTypeInput.js
@@ -40,28 +40,30 @@ class NumberTypeInput extends Component {
   }
 
   handleChange = (event, value) => {
-    const { maxValue, minValue } = this.state;
-    console.log("handle change", event, value)
-
     let newValue = parseInt(value, 10);
-    console.log("new val", newValue)
+    const { maxValue, minValue } = this.state;
+    // prevent the new value from being
+    // greater or less than the min and max values
     if (newValue > maxValue) newValue = maxValue;
     if (newValue < minValue) newValue = minValue;
 
 
-    // TODO: comment on this handler
+    // setting the value state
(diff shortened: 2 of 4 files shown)


spencern:
@nnnnat @Akarshit I'd like to make sure that this currency input is using the same component that was added/fixed in https://github.com/reactioncommerce/reaction/pull/3808

nnnnat:
@kieckhafer I was able to resolve the issue, the component won't send any NaNs back to the form now and it will correct invalid values (empty strings, string, negative numbers, etc).

@spencern  I was able to replace the NumberTypeInput component's HTML input with the updated TextField component, I had to update the TextField slightly to deal with returning the number 0 and accepting a min/max input attribute.

I also updated the invoceConatiner to make sure the `refundQuantity` wasn't 0 before pushing the new total because of a weird false positive that would display a refunded amount of $0.00.

Everything seems to be working still, was able to do all acceptance test and all RC test pass.

kieckhafer:
Changes look good, everything seems to work now 👍

PR #3835, opened 2018-02-26
Fix for "Changing user currency does not update prices"

### LINKED ISSUE #3833: Changing user currency does not update prices
## Prerequisites

## Issue Description
Changing user currency is not updating prices.

This is not an issue with reactivity, the prices stay in USD when refreshing as well.
It does not appear to be a permissions issue, as being logged in as an admin does not resolve this issue.

## Steps to Reproduce
1. Start the app
2. Change user currency to something other than USD

### Versions
```
Node: 8.9.0
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.28.0
Reaction: 1.8.1
Reaction branch: master
Docker: 17.12.0-ce
```

prinzdezibel:
Resolves #3833  
Impact: **critical**  
Type: **bugfix**

## Issue

## Solution
Changes:
Seems not possible to have a job that's scheduled via later.js to have
started immediately. Tried with delay() and after(), but no avail.

The solution was to just schedule a one-off task whenever the server
starts.

Also I did refactor the import of later.js , because it was not
consistently. Sometimes it was required directly, sometimes via
`Jobs.later`

## Breaking changes
none expected

## Testing
1. Reset the database
2. Start server
3. The currency exchange rates should be fetched immediately (or very short after spinning up the server)
4. Try to change the currency in the nav bar
5. Prices should change

### DIFF
--- a/imports/plugins/included/jobcontrol/server/jobs/exchangerates.js
+++ b/imports/plugins/included/jobcontrol/server/jobs/exchangerates.js
@@ -1,8 +1,8 @@
-import later from "later";
 import { Meteor } from "meteor/meteor";
 import { Job } from "/imports/plugins/core/job-collection/lib";
 import { Jobs } from "/lib/collections";
 import { Hooks, Logger, Reaction } from "/server/api";
+
 
 export function setupFetchFlushCurrencyHooks() {
   // While we don't necessarily need to wait for anything to add a job
@@ -22,13 +22,23 @@ export function setupFetchFlushCurrencyHooks() {
           backoff: "exponential" // delay by twice as long for each subsequent retry
         })
         .repeat({
-          schedule: later.parse.text(refreshPeriod)
+          schedule: Jobs.later.parse.text(refreshPeriod)
         })
         .save({
           // Cancel any jobs of the same type,
           // but only if this job repeats forever.
           cancelRepeats: true
         });
+
+      // Run the first time immediately after server start. The repeat({schedule}) option via
+      // later.js scheduling won't run before the first scheduled time, even with delay() or after()
+      new Job(Jobs, "shop/fetchCurrencyRates", {})
+        .priority("normal")
+        .retry({
+          retries: 5,
+          wait: 10000
+        })
+        .save();
     } else {
       Logger.warn("OpenExchangeRates API not configured. Not adding fetchRates job");
     }
@@ -50,7 +60,7 @@ export function setupFetchFlushCurrencyHooks() {
           backoff: "exponential"
         })
         .repeat({
-          schedule: later.parse.text(refreshPeriod)
+          schedule: Jobs.later.parse.text(refreshPeriod)
         })
         .save({
           cancelRepeats: true
--- a/imports/plugins/included/jobcontrol/server/jobs/exchangerates.js
+++ b/imports/plugins/included/jobcontrol/server/jobs/exchangerates.js
@@ -15,7 +15,6 @@ export function setupFetchFlushCurrencyHooks() {
       const refreshPeriod = exchangeConfig.refreshPeriod || "every 4 hours";
       Logger.debug(`Adding shop/fetchCurrencyRates to JobControl. Refresh ${refreshPeriod}`);
       new Job(Jobs, "shop/fetchCurrencyRates", {})
-        .priority("normal")
         .retry({
           retries: 5,
           wait: 60000,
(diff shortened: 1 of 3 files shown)


spencern:
@jshimko as this is an `impact-critical` issue, would you prioritize it above other non critical reviews?

jshimko:
@spencern Yep.  Looking at it now.

imports/plugins/included/jobcontrol/server/jobs/exchangerates.js
```diff
@@ -22,13 +22,23 @@ export function setupFetchFlushCurrencyHooks() {
...
         .save({
           // Cancel any jobs of the same type,
           // but only if this job repeats forever.
           cancelRepeats: true
         });
+
+      // Run the first time immediately after server start. The repeat({schedule}) option via
+      // later.js scheduling won't run before the first scheduled time, even with delay() or after()
+      new Job(Jobs, "shop/fetchCurrencyRates", {})
+        .priority("normal")
```
jshimko: Not a big deal, but `.priority("normal")` is the default, so this is redundant/unnecessary to have here.

jshimko: If we're going to standardize on the version of `later` included with the job queue package, we should probably remove the one from Reaction's package.json.  Since `@reactioncommerce/job-queue` is now on npm and [already has `later` defined as a dependency](https://github.com/reactioncommerce/reaction-job-queue/blob/master/package.json#L27), we shouldn't need to require `later` again and inject that dependency [here](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/job-collection/lib/index.js#L6-L8) (right @mikemurray?).  But that's outside the scope of this issue and probably needs an issue to be filed over at https://github.com/reactioncommerce/reaction-job-queue.  Either way, I deleted `later` from Reaction's `package.json` and confirmed everything still works.  It was only used for the jobs and the job queue package installs it for you already, so I think you can safely remove it in this PR.

Otherwise, I can confirm that the fix works, so I think we're good to go.  @mikemurray can you just confirm how that `later` stuff is being handled in the npm package?  Did you do that dependency injection to solve some kind of issue?

jshimko:
If we're going to standardize on the version of `later` included with the job queue package, we should probably remove the one from Reaction's package.json.  Since `@reactioncommerce/job-queue` is now on npm and [already has `later` defined as a dependency](https://github.com/reactioncommerce/reaction-job-queue/blob/master/package.json#L27), we shouldn't need to require `later` again and inject that dependency [here](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/job-collection/lib/index.js#L6-L8) (right @mikemurray?).  But that's outside the scope of this issue and probably needs an issue to be filed over at https://github.com/reactioncommerce/reaction-job-queue.  Either way, I deleted `later` from Reaction's `package.json` and confirmed everything still works.  It was only used for the jobs and the job queue package installs it for you already, so I think you can safely remove it in this PR.

Otherwise, I can confirm that the fix works, so I think we're good to go.  @mikemurray can you just confirm how that `later` stuff is being handled in the npm package?  Did you do that dependency injection to solve some kind of issue?

mikemurray:
@jshimko Later.js is not currently used in the npm package, though it is in the `package.json` to ensure it gets included. It was dependency injected into the package from Reaction. An update would need to be made in the `@reactioncommerce/job-queue` package use the npm version of later, rather than depending on dependency injection.

prinzdezibel:
@jshimko removed the `priority("normal")`

imports/plugins/included/jobcontrol/server/jobs/exchangerates.js
prinzdezibel: done

prinzdezibel:
@mikemurray That means `Jobs.later` would be the right way to refer to it, correct?

jshimko:
The job queue package makes sure it's in the `node_modules` dir, so the `require("later")` works whether Reaction declares it as a dependency or not.  We'll have to create a new issue in the other repo to clean it up, but it's not an emergency.  It's only installed once anyway since they both refer to the same version.

mikemurray:
@prinzdezibel you could use `Jobs.later`, it's just a reference to the imported `later` so you could always just import and use later if needed.

PR #3838, opened 2018-02-26
Improve unnecessary translation loading

### LINKED ISSUE #3284: Improve unnecessary translation loading
Something I noticed while working on SimpleSchema changes...

`loadTranslations` runs for each plugin with translations, every time the server restarts, doing a ton of upserts that are essentially no-ops. This *might* not be necessary on every startup in production, but it's *definitely* not necessary on every startup while developing.

The solution would probably involve some combination of caching/comparing a checksum and/or testing whether we're in production mode. Might also help to have a version in each json file.

Or the whole thing could be triggered manually by an admin instead of on startup.

I haven't given it enough thought to know what's best, but I wanted to mention it before I forget.

Potentially dynamic imports solves this issue.

Consider adding a configuration file for enabled languages. 

Import languages once a shop has toggled on i18n.

Hook `i18n/flushTranslations` up to a button in reaction devtools to make resetting translations easily.

mikemurray:
Resolves #3284 
Impact: **major**  
Type: **performance**

## Issue

Improve app rebuild performance by not re-importing already imported translations on app start and reload.

## Solution

Tested to see if there will be any improvement by reducing the re-import of translations. Those results can be found here https://github.com/reactioncommerce/reaction/issues/3284#issuecomment-367862061

Then proceeded to solve the issue by checking the `Assets` collection for assets of type `i18n` and if any are found, skip the import. The translations are imported into `Assets`, then into `Translations` on startup. Translations will still be prepared to be imported into assets, but the bulk operation is discarded if the import is not required.

Added a button in the **Localization and i18n** control panel to reload translations for the current or all shops. Current shop `click`, all shops `alt + click` if you're the marketplace admin

basic_reaction_product

## Documentation

https://github.com/reactioncommerce/reaction-docs/pull/540

## Breaking 

Translations are no longer reloaded automatically on restart/refresh if they've already been loaded. To reload you can call `Meteor.call("i18n/flushTranslations")` for the current shop or `Meteor.call("i18n/flushAllTranslations")` for all shops. You can also reload the translations by going to the Internationalization control panel and using the reload translations button.

## Testing

**From a fresh reset**

1. Start the app with `REACTION_LOG_LEVEL=DEBUG reaction`
2. In the terminal, you should see some messages like `Translation assets bulk update prepared for  [some-package-name]`
3. Eventually followed by `Importing en translation for "[some-package-name]"`
4. Stop and start the app with `REACTION_LOG_LEVEL=DEBUG reaction` again
5. you should only see `Translation assets bulk update prepared for  [some-package-name]`
6. you should **NOT** see `Importing en translation for "[some-package-name]"`
7. From the Internationalization settings, use the reload button. you should see in your terminal `Importing en translation for "[some-package-name]"`


nnnnat:
We're going to ignore the Codacy issues and approve

PR #3845, opened 2018-02-27
(Fix 3836) Fix invite shop owner

### LINKED ISSUE #3: Connectors Documentation

### LINKED ISSUE #3836: New Shop Unable To Change Address In Shop Panel
## Prerequisites
* [y] Are you running the latest version? **1.8.1**
* [y] Are you able to consistently reproduce the issue?
* [y] Did you search the issue queue for existing issue? [Search issues](https://github.com/reactioncommerce/reaction/issues) - tried to, but couldn't find anything. Also posted in gitter to check.

## Issue Description
Upon signing up as a new seller in marketplace, (by clicking on 'become seller', or via invitation), going to the 'Shop' tab on the RHS and trying to input an address results in the following toast error showing, and no submission: "Shop general settings update failed. Error: Access denied [403]"

## Steps to Reproduce
1. Fresh init of reaction commerce
2. Turn on marketplace, and allow seller signup
3. Sign up a new account.
4. Go to 'Shop' tab on RHS and try to input the address.
5. Error shows.

## Possible Solution
Did not exist for me in 1.8.1. Looking in the console suggests that the new seller does not have permission to change their address in the 'Shop' panel.

### Versions
Node: 8.9.0
NPM: 5.5.1
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.28.0
Reaction: 1.8.1
Reaction branch: master

zenweasel:
Resolves #3836 
Impact: **critical**  
Type: **bugfix**

## Issue
Two issues:
1. the "shopSettings" permissions was not being assigned to shop owners when invited
1. `getShopId` was not looking in the right place for the `activeShopId` so it was never finding the role even when it exists

## Solution
1. Add the role when creating a shop
1. create a new method called `setUserPreferences` that sets preferences for a user, and then call that when creating a shop

## Breaking changes
If other things are expecting `acitveShopId` to exist elsewhere it might break something

## Testing
1. Invite a shop owner
1. Login in incognito as that user and set password
1. Try to change the address
1. Observe that you receive no error

### DIFF
--- a/server/api/core/core.js
+++ b/server/api/core/core.js
@@ -371,7 +371,7 @@ export default {
    * @summary Get shop ID
    * @todo This should intelligently find the correct default shop Probably whatever the main shop is or marketplace
    * @param  {String} userId User ID String
-   * @return {StringId}        active shop ID
+   * @return {String} active shop ID
    */
   getShopId(userId) {
     check(userId, Match.Maybe(String));
@@ -595,6 +595,24 @@ export default {
     return defaultValue || undefined;
   },
 
+  /**
+   * @name setUserPreferences
+   * @method
+   * @summary save user preferences in the Accounts collection
+   * @param {String} packageName
+   * @param {String} preference
+   * @param {String} value
+   * @param {String} userId
+   * @return {Number} setPreferenceResult
+   */
+  setUserPreferences(packageName, preference, value, userId) {
+    const setPreferenceResult = AccountsCollection.update(userId, {
+      $set: {
+        [`profile.preferences.${packageName}.${preference}`]: value
+      }
+    });
+    return setPreferenceResult;
+  },
   /**
    *  @name insertPackagesForShop
    *  @method
--- a/server/api/core/groups.js
+++ b/server/api/core/groups.js
@@ -76,6 +76,7 @@ function getDefaultGroupRoles() {
 
   // we're making a Shop Manager default group that have all roles except the owner role
   const shopManagerRoles = ownerRoles.filter((role) => role !== "owner" && role !== "admin");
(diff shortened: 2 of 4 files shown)


server/api/core/core.js
```diff
@@ -595,6 +595,24 @@ export default {
...
+   * @name setUserPreferences
+   * @method
+   * @summary save user preferences in the Accounts collection
+   * @param {String} packageName
+   * @param {String} preference
+   * @param {String} value
+   * @param {String} userId
+   * @return {Number} setPreferenceResult
+   */
+  setUserPreferences(packageName, preference, value, userId) {
```
pmn4: does it make sense to set a default `userId = Meteor.userId()`?

server/api/core/groups.js
```diff
@@ -76,6 +76,7 @@ function getDefaultGroupRoles() {
 
   // we're making a Shop Manager default group that have all roles except the owner role
   const shopManagerRoles = ownerRoles.filter((role) => role !== "owner" && role !== "admin");
+  shopManagerRoles.push("shopSettings");
```
pmn4: this line feels odd. I can't quite put my finger on why, but I guess I don't understand why `"shopSettings"` is such a unique role that it is called out explicitly here where no other roles are.

zenweasel: I sort of agree with you, but we don't really have a place (or alternately I did not find the place) where we are setting these defaults. The existing set up is just "everything that owner gets minus `owner` and `admin`" which also seems a little, weird?

Plus this is basically the only place where a shop owner is defined, so it would seem to make sense to assign any roles that are explicitly "shop owner only" here?

PR #3847, opened 2018-02-27
[WIP] Fix 3821: currency component not converting currency

### LINKED ISSUE #3821: Prices should not be converted in admin view
### Expected behavior
The shop base currency should be honored for all prices shown in admin view.

### Actual behavior
If the shop's base currency is set to EUR and the admin does add a new product, the product's price get converted immediately from USD->EUR, because shop assumes that the entered value is USD.

EDIT: Strangely, the above stated does not happen reproducible

EDIT 2: Is happens reliably, if the currency selector in the nav bar doesn't match the shop base currency.

But the following seems to happen always:
Additionally it's not possible to enter the in the shop base locale, e.g. with a comma as separator for EUR currency.

- As an admin, set the shop base currency to EUR. And for good measure, also set the currency selector in the nav bar to EUR.
- Add a product
- Input a value
- The value gets converted as soon as the field blurs.

### Versions
Node: 8.2.1
NPM: 5.3.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.26.4
Reaction: 1.8.0
Docker: 17.03.1-ce

### Analysis
It could have to do with this PR: https://github.com/reactioncommerce/reaction/commit/a9937023c911334d7cae5d6a84252666bd802b99

implicit_conversion

Akarshit:
Resolves #3821 
Impact: **major**  
Type: **bugfix**

## Issue
The issue was basically that the component was using the user selected currency for conversion.

## Solution
The conversion function now uses the shop's base currency
Also fixed some issues like
1. Only number can be entered in the currecny
1. There is a separate reusable CurrencyInput component now.

## Testing
1. As an admin, set the shop base currency to EUR.
1. Edit a variant of the `Example Product`
1. Input a price
1. The value should not get converted to any other value as soon as the field blurs. Only currency formatting should be added.
1. Try entering a non-digit character in the input box. 
1. Notice you are unable to do so.


imports/plugins/core/ui/client/components/currencyInput/currencyInput.js
```diff
@@ -0,0 +1,72 @@
+import { unformat } from "accounting-js";
+import { registerComponent } from "@reactioncommerce/reaction-components";
+import TextField from "../textfield/textfield";
```
jshimko: For the sake of consistency with the components API, this should probably be...

```js
import { getRawComponent } from "@reactioncommerce/reaction-components";

const TextField = getRawComponent("TextField");

class CurrencyInput extends TextField {
...
``` 

That way if someone replaces the default `TextField` component with `replaceComponent`, you will inherit from the correct source when extending the class.  Probably a bit of a corner case that isn't likely to happen, but it would at least leave the option open.

jshimko: See the section called "Extending Components" in the Components API docs.

https://docs.reactioncommerce.com/reaction-docs/master/components-api

Akarshit:
@jshimko Thanks. Updated the PR.

Akarshit:
Codacy says not to do 
```js
this.setState({
        value: undefined
      });
```
I don't know how to avoid it. ESlint doesn't complain about this.

jshimko:
@Akarshit can you just set the state value to `null` or an empty string instead?

Akarshit:
OK doing that.

aaronjudd:
@Akarshit I've disabled Codacy integrations with Reaction -> we can still review it manually, but after research / conversations with Codacy -> they're not respecting our .eslint configurations so we've decided to instead rely on our own lint tooling with CI, and to stop using external services for this.

Akarshit:
😢 Now who will praise my PRs saying "Good work! A positive pull request."

aaronjudd:
@Akarshit Good work! A positive pull request.

Akarshit:
@jshimko I am not updating this, because we no longer care about codacy and ESLint is good with the PR.

prinzdezibel:
It does not work for me. Maybe we should pair on this?
basic_reaction_product

Akarshit:
Checking on this.

spencern:
@Akarshit @prinzdezibel holding off on merging this for now and adding a WIP tag, will you please remove WIP tag once it's resolved?

@prinzdezibel probably worth creating a real review for this next time and "requesting changes" so that it doesn't get merged accidentally if it's not ready.

Akarshit:
@prinzdezibel I checked and yes this was happening because of a bug.
> When I click inside this, it gets converted to USD

This is a bug, I am fixing it.

> Also it's not localised

When you are editing the currency, it's just like a number(so if you have to enter 1.112,50€, you will have to enter 1112.50).

prinzdezibel:
@Akarshit  yes, but suppose the shop base currency is German and you're going to add a 12 EUR product to the shop, you'd expect to input "12,00" into the text field, not "12.00". This does not work in the current state, I'm afraid.

aaronjudd:
@Akarshit @prinzdezibel please remove the approved review then (otherwise this is going to get merged) -> and if it's going back into surgery - add WIP to the title.

imports/plugins/core/ui/client/components/currencyInput/currencyInput.js
```diff
@@ -0,0 +1,88 @@
...
+    return this.props.value || "";
+  }
+
+  /**
+   * onValueChange
+   * @summary set the state when the value of the input is changed
+   * @param  {Event} event Event object
+   * @return {void}
+   */
+  onChange = (event) => {
```
machikoyasuda:
```
/**
 * @method onChange
```

Akarshit:
This would need more work to keep the currency formatting even while editing. Working to implement that.

prinzdezibel:
@Akarshit As discussed in our daily sync: Maybe you can reuse the NumericInput for this task which would make sense to avoid redundant currency controls, IMHO. Please see this PR: https://github.com/reactioncommerce/reaction/pull/3893

Akarshit:
Closing this as the new fix would be using a different component entirely.

PR #3848, opened 2018-02-27
Call OrdersList as a Component

willmoss1000:
Resolves 3849
Impact: **minor**  
Type: **bugfix**

## Issue
OrdersList is currently called via direct import. It is therefore not possible to replace this component. 

## Solution
Change from Direct Import to "Components.OrdersList"

## Breaking changes
N/A

## Testing
1. Make amendment
2. Testing - go to profile and check for no errors.

1. Anywhere in code, paste this snippet:
```
const MyOrdersList = () => {
  return <div>custom orders</div>
};

replaceComponent("OrdersList", MyOrdersList);
```
2. As an user or admin navigate to the "Your orders" section on the profile page: http://localhost:3000/account/profile

3. The custom component should be rendered


prinzdezibel:
Tested and verified it works.

pmn4:
lgtm. nice one @willmoss1000!

Issue #3654, opened 2018-02-06
React throws setState warnings for NotificationDropdown component


kieckhafer:
### Expected behavior
There should be no setState in the console when using Reaction. This particular issue relates to the `setState` warnings seen in the `NotificationDropdown` component.

### Actual behavior
When loading Reaction, as a user who has notifications enabled, a `setState` warning is given as `moment` loading causes the component to unmount before the state is set.

### Steps to reproduce the behavior
- Login as a user who has notifications enabled.
- Open your console
- See errors when Reaction is loaded

spencern:
Specific things to check with this issue:
Make sure that we're not loading moment while the dropdown is hidden.

mikemurray:
This component doesn't exist in the new Operator experience.

PR #3872, opened 2018-03-01
(performance): delayed response in localization settings

### LINKED ISSUE #3497: Perf - Localisation panel is slow to respond.
### Expected behavior
The panel should be quick to reflect the changes.
### Actual behavior
After toggling any language, there is a lag after which the toggle happens.
local

1. Run `reaction`
1. Open `Localisation` panel from Action Menu
1. Toggle any language and observe the lag.

kieckhafer:
Resolves #3497 
Impact: **minor**  
Type: **performance**

## Issue
1) When turning on / off languages and currencies in your Shop settings, there is a slight delay in response after the toggle is clicked.
2) The app was re-rendering `x+1` times, with `x` being the number of times you've toggled a switch.

## Solution
There are various issues addressed in this PR, stemming from the original ticket:
1) The "issue" here seems to be that since we are changing settings on the `Shops` collection, everything that depends on the `shop` settings, and has `Trackers` on the `shop`, needs to update - which means the whole app updates. Add this in with the slight RT delay when making a `Meteor.call` to the server, and we are seeing this delay of about `1500ms`. The solution here is to add an optimistic UI update with `state`, and then do the `Meteor.call`. Because the timing of the `Meteor.call` is just delayed enough to return a result about `50%` of the way through the `state` animation, we add a `200ms` timeout to make sure the call doesn't happen in the middle of the animation, and make the animation look broken.
2. Add a tracker `.stop()` to the top level app tracker

## Breaking changes
None

## Testing
1. Login as an admin
2. Navigate to the i18n settings in the dashboard
3. Toggle on / off various languages and currencies
4. See no delay in toggle UI updates, and see that the currencies and languages are enabled / disabled as they should be

### DIFF
--- a/imports/plugins/core/i18n/client/components/localizationSettings.js
+++ b/imports/plugins/core/i18n/client/components/localizationSettings.js
@@ -27,27 +27,32 @@ class LocalizationSettings extends Component {
     super(props);
 
     this.state = {
+      currencies: props.currencies,
       languages: props.languages
     };
   }
 
   componentWillReceiveProps(nextProps) {
     this.setState({
+      currencies: nextProps.currencies,
       languages: nextProps.languages
     });
   }
 
-  renderCurrencies() {
-    return this.props.currencies.map((currency, key) => (
-      <Components.ListItem
-        actionType={"switch"}
-        key={key}
-        label={currency.label}
-        switchOn={currency.enabled}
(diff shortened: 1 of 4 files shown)


nnnnat:
Talked to @kieckhafer  and we're going to address the setState warning when we tackle issue #3654

PR #3876, opened 2018-03-01
[WIP] (refactor): 3662 customer product grid publishing

### LINKED ISSUE #3662: Prevent publishing all Products for non-admin user, it breaks the app for large product sets.
### Expected behavior
Navigation should be snappy and fast, regardless of the number of products that are available in shop. Browser should not freeze at any time.

### Actual behavior
When having 1000 or more products, the app becomes unusable. Navigating to the index route (or tag route) will take minutes before it gets displayed. On many occasions it freezes and the browser tab becomes unresponsive.

1. Clone reaction-devtools into /imports/plugins/custom
```
$ cd import/plugins/custom
$ git clone https://github.com/reactioncommerce/reaction-devtools.git
```
2. Reset database and start
```
$ reaction reset -n
$ reaction
```

3. Login as admin and open developer tools
Load medium dataset. see screenshot. 
reaction

4. Logout

5. As non-admin user, navigate to http://localhost:3000

6. Wait for a long time to see the products to appear. Sometimes the page doesn't load at all.

Finding a solution for this issue would involve reviewing this code: https://github.com/reactioncommerce/reaction/blob/f40ff536c139d70da02ca10ae12655247452d658/server/publications/collections/products.js#L539-L542

nnnnat:
Resolves #3662  
Impact: **breaking**  
Type: **performance|refactor**

## Issue
The customer view of the product grid was taking an incredibly long time to load or in many cases just crashing the browser tab. 

Several issues with the productGridItem component were causing slow down:
* grid items had deep and confusing nested props.
* grid items were using inherited prop callback to get values already in the component's props.
* grid items had a lot of conditional logic only needed for admin views.
* the product grid was destructuring all of its props onto each grid item, this was causing each grid item to have not only it's own product data but all the entire array of products.

Issues with the "Products" publication
* 

## Solution
@mikemurray and I have been approaching this issue by separating the customer view from the admin view. 

* Created a new customer's "Products" publication that doesn't worry about any admin needed data, attaches product variants to the top-level product and fetches the appropriate product media.
* Created a few React components that only deal with customer interaction.

## Breaking changes
The old `imports/plugins/included/product-variants/containers/productsContainer.js` has been renamed to `productsContainerAdmin.js` and a new component names `productsContainer.js` now handles which products container to load based on the user's permissions.

*You'll need the [reaction-devtool](https://github.com/reactioncommerce/reaction-devtools) installed to test this PR.*
## Testing
1. Start up a fresh RC shop and log in as an Admin.
2. Open devtools panel and load the medium data set. Once that finishes loading, load the images as well.
3. In another browser view the RC shop as a customer to see product grid loading the large data set.
4. Follow [acceptance tests](https://docs.reactioncommerce.com/reaction-docs/master/acceptance-testing) 1A + 1B to verify customer views still working.

### DIFF
--- a/imports/plugins/included/product-variant/components/customer/productGridItem.js
+++ b/imports/plugins/included/product-variant/components/customer/productGridItem.js
@@ -1,7 +1,6 @@
 import React, { Component } from "react";
 import PropTypes from "prop-types";
-import { Components } from "@reactioncommerce/reaction-components";
-import { formatPriceString } from "/client/api";
+import { formatPriceString, Reaction } from "/client/api";
 
 class ProductGridItems extends Component {
   static propTypes = {
@@ -21,6 +20,7 @@ class ProductGridItems extends Component {
     weightClass: PropTypes.func
   }
 
+  // action event handlers
   handleDoubleClick = (event) => {
     if (this.props.onDoubleClick) {
       this.props.onDoubleClick(event);
@@ -33,12 +33,18 @@ class ProductGridItems extends Component {
     }
   }
 
+  // modifiers
   renderPinned() {
     return this.props.positions().pinned ? "pinned" : "";
   }
 
   renderVisible() {
     return this.props.product.isVisible ? "" : "not-visible";
+  }
+
+   // render class names
+  renderHoverClassName() {
+    return this.props.isSearch ? "item-content" : "";
   }
 
   renderOverlay() {
@@ -49,13 +55,49 @@ class ProductGridItems extends Component {
     }
   }
 
+  // notice
+  renderNotices() {
+    return null
+    // return (
+    //   <div className="grid-alerts">
+    //     <GridItemNotice isSoldOut={() => this.props.product.isSoldOut} />
+
+    //   </div>
+
+    // );
+  }
+
+  // getters
+
+  // get product detail page URL
+  get productURL() {
+    const { product: { handle } } = this.props;
+    return Reaction.Router.pathFor("product", {
+      hash: {
+        handle
+      }
+    });
+  }
+
+  // render product image
   renderMedia() {
-    const { product, productMedia } = this.props;
+    // const { product, productMedia } = this.props;
+
+    // return (
+    //   <Components.ProductImage displayMedia={() => productMedia.primaryMedia} item={product} size="large" mode="span" />
+    // );
+    const { product } = this.props;
+    const MEDIA_PLACEHOLDER = "/resources/placeholder.gif";
+    const { url } = (Array.isArray(product.media) && product.media[0]) || { url: MEDIA_PLACEHOLDER };
 
     return (
-      <Components.ProductImage displayMedia={() => productMedia.primaryMedia} item={product} size="large" mode="span" />
+      <span
+        className="product-image"
+        style={{ backgroundImage: `url(${ url })` }}
+      />
     );
   }
+
 
   renderAdditionalMedia() {
     const { isMediumWeight, productMedia } = this.props;
@@ -78,17 +120,6 @@ class ProductGridItems extends Component {
     );
   }
 
-  renderNotices() {
-    return null
-    // return (
-    //   <div className="grid-alerts">
-    //     <GridItemNotice isSoldOut={() => this.props.product.isSoldOut} />
-
-    //   </div>
-
-    // );
-  }
-
   renderGridContent() {
     const { product } = this.props;
 
@@ -96,7 +127,7 @@ class ProductGridItems extends Component {
       <div className="grid-content">
         <a
           // href={this.props.pdpPath()}
-          href={`product/${product._id}`}
+          href={this.productURL}
           data-event-category="grid"
           data-event-action="product-click"
           data-event-label="grid product click"
@@ -116,25 +147,13 @@ class ProductGridItems extends Component {
     );
   }
 
-  renderHoverClassName() {
-    return this.props.isSearch ? "item-content" : "";
-  }
-
   render() {
     const { product } = this.props;
-    const MEDIA_PLACEHOLDER = "/resources/placeholder.gif";
-
-    // TODO: use this to get product url
-    // Reaction.Router.pathFor("product", {
-    //   hash: {
-    //     handle
-    //   }
-    // });
 
     // TODO: isSelected is not needed. Others may not need to be functions
     // ${this.renderPinned()} ${this.props.weightClass()} ${this.props.isSelected()}
 
-    const { url } = (Array.isArray(product.media) && product.media[0]) || { url: MEDIA_PLACEHOLDER };
+
     return (
       <li
         className={`product-grid-item `}
@@ -145,7 +164,7 @@ class ProductGridItems extends Component {
           <span className="product-grid-item-alerts" />
 
           <a className="product-grid-item-images"
-            href={`product/${product._id}`}
+            href={this.productURL}
             data-event-category="grid"
             data-event-label="grid product click"
             data-event-value={this.props.product._id}
@@ -153,12 +172,7 @@ class ProductGridItems extends Component {
             onClick={this.handleClick}
           >
             <div className={`product-primary-images`}>
-              <span
-                className="product-image"
-                style={{ backgroundImage: `url('${url}')` }}
-              />
-              {/* {this.renderMedia()} */}
-              {/* {this.renderOverlay()} */}
+              {this.renderMedia()}
             </div>
 
             {/* {this.renderAdditionalMedia()} */}
--- a/imports/plugins/included/product-variant/components/customer/productGridItem.js
+++ b/imports/plugins/included/product-variant/components/customer/productGridItem.js
@@ -2,15 +2,13 @@ import React, { Component } from "react";
 import PropTypes from "prop-types";
 import classnames from "classnames";
 import { formatPriceString, Router } from "/client/api";
+import { Components } from "@reactioncommerce/reaction-components";
 
 class ProductGridItem extends Component {
   static propTypes = {
-    displayPrice: PropTypes.func,
-    isMediumWeight: PropTypes.func,
     isSearch: PropTypes.bool,
     position: PropTypes.object,
-    product: PropTypes.object,
-    weightClass: PropTypes.func
+    product: PropTypes.object
   }
 
   // getters
@@ -48,14 +46,6 @@ class ProductGridItem extends Component {
     });
   }
 
-  renderOverlay() {
-    if (this.props.product.isVisible === false) {
-      return (
-        <div className="product-grid-overlay" />
-      );
-    }
-  }
-
   handleClick = (event) => {
     event.preventDefault();
     Router.go(this.productURL);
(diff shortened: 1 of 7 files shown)


mikemurray:
@spencern 

**Have working**

- Have the products being published to the `Catalog` collection when clicking the publish button.
- Publish button is unlocked. It can be clicked any time regardless of if there's a revision to publish to the catalog.
- Denormalized `isSoldOut`, `isBackorder`, and `isLowQuantity`
- Denormalized 4 product images, one primary and four additional for the medium size grid item
- Denormalized a flat array of variants

---

**Need to get working**

- Need to fix console errors when trying to publish when there's no revision
- Need to ensure that publishing to the catalog works from the product grid
- Need to be able to unpublish from the catalog. (Should making a product not visible remove it from the catalog, or just make it not visible)
- Need to attach an actual schema to the catalog collection
- Devtools needs to be updated to include a method to mass publish to catalog
- Need to re-implement filtering for "Products/grid" publication
- Need to filter fields in "Products/grid" publication. We don't need the variants on the client really
- Verify that search still works
- **Need to verify that all of this works with marketplace.**

mikemurray:
Some more thoughts from @aaronjudd 

- Think about publishing inventory to the grid
- Remove price and inventory fields from catalog documents
- No schema for now
- Set product `type: "simple"` to `type: "simple-product"` in the catalog

spencern:
When testing this:
- [ ] Test with filters that are not currently available 
- [ ] Test that search works and includes media

spencern:
Here's an initial set of testing steps for this PR. Please add to this as you think of things that should be tested:
To test:
- [ ] Test with filters that are not currently available 
- [ ] Test that search works and includes media 
- [ ] Test that when inventory is updated (up or down) that the product grid item badges show correctly
- [ ] Test that when an order is canceled and it changes the inventory status of an item, that the grid items update properly
- [ ] Test that when an image is deleted from an item that the grid images update properly
- [ ] Test that the different grid item layouts (medium specifically) show the correct set of images

imports/plugins/core/catalog/server/methods/catalog.js
```diff
@@ -0,0 +1,216 @@
...
+import { Media } from "/imports/plugins/core/files/server";
+import { ProductRevision as Catalog } from "/imports/plugins/core/revisions/server/hooks";
+
+/**
+ * isSoldOut
+ * @summary We are stop accepting new orders if product marked as `isSoldOut`.
+ * @param {Array} variants - Array with top-level variants
+ * @return {Boolean} true if summary product quantity is zero.
+ */
+export function isSoldOut(variants) {
```
machikoyasuda: ```@method isSoldOut```

mikemurray:
@machikoyasuda I've made changes to the docs, could you re-review?

machikoyasuda:
@mikemurray @nnnnat JSDoc updates are 💯 👍

spencern:
@mikemurray @nnnnat is this almost ready for code review and QA?
I see you've still got some tests failing and a WIP label.

nnnnat:
@spencern  almost ready, I'm looking into these failing test this morning and @mikemurray has been working on adding some test for the new publication. We should be ready for review once we've got these test worked out.

mikemurray:
@spencern I don't think this will be ready to fully review even once the tests are passing. This PR is branched from https://github.com/reactioncommerce/reaction/pull/3782 which has yet to be reviewed and contains some merge conflicts, as well as issue https://github.com/reactioncommerce/reaction/pull/3782#pullrequestreview-102009112 .

imports/plugins/core/catalog/server/methods/catalog.js
```diff
@@ -0,0 +1,241 @@
...
+    thumbnail: `${media.url({ store: "thumbnail" })}`,
+    small: `${media.url({ store: "small" })}`,
+    medium: `${media.url({ store: "medium" })}`,
+    large: `${media.url({ store: "large" })}`,
+    image: `${media.url({ store: "image" })}`
+  }));
+
+  // Denormalize product fields
+  product.media = productMedia;
+  product.type = "product-simple";
```
kieckhafer: Is this a new `type`? Why don't we use `simple` or `variant` like in other places?

imports/plugins/core/catalog/server/methods/catalog.js
```diff
@@ -0,0 +1,241 @@
...
+      ids = [productIds];
+    }
+
+    // Find all products
+    const productsToPublish = Products.find({
+      _id: { $in: ids }
+    }).fetch();
+
+    if (Array.isArray(productsToPublish)) {
+      const canUpdatePrimaryShopProducts = Reaction.hasPermission("createProduct", this.userId, Reaction.getPrimaryShopId());
```
kieckhafer: Why not do this check at the start of the method with the other permission check?

kieckhafer: Is there a way to combine the three separate permissions checks in this method?

imports/plugins/core/catalog/server/methods/catalog.js
mikemurray: Yes. For the catalog, products of type `simple` will be `product-simple` as it's more descriptive and because other types of data may live in the Catalog, not just products.

imports/plugins/core/revisions/client/containers/publishContainer.js
```diff
@@ -12,8 +12,22 @@ import { Reaction, i18next } from "/client/api";
  * PublishContainer is a container component connected to Meteor data source.
  */
 class PublishContainer extends Component {
+  publishToCatalog(collection, documentIds) {
+    Meteor.call(`catalog/publish/${collection}`, documentIds, (error, result) => {
+      if (result) {
+        Alerts.toast(i18next.t("catalog.productPublishSuccess", { defaultValue: "Product published to catalog" }), "success");
```
kieckhafer: Needs translation in `en.json`

imports/plugins/core/catalog/server/methods/catalog.js
mikemurray: One call is to check for the admin of the marketplace, the other to check to see if the merchant shop user has permissions.

Marketplace admins can publish/unpublish any product.
Merchant shop owners can only publish their own products.

imports/plugins/included/product-variant/components/customer/productGridItem.js
```diff
@@ -0,0 +1,191 @@
...
+        </div>
+      </div>
+    );
+  }
+
+  // render product image
+  renderMedia() {
+    const { product } = this.props;
+    const MEDIA_PLACEHOLDER = "/resources/placeholder.gif";
+    const { large } = (Array.isArray(product.media) && product.media[0]) || { large: MEDIA_PLACEHOLDER };
```
kieckhafer: `large` seems like a strange const name. Not a big deal at all, just seemed odd reading through it. WHy not just `media` or `img`?

imports/plugins/included/product-variant/components/customer/productGridItem.js
```diff
@@ -0,0 +1,191 @@
...
+
+  render() {
+    const { product, isSearch } = this.props;
+    return (
+      <li
+        className={this.productClassNames}
+        data-id={product._id}
+        id={product._id}
+      >
+        <div className={(isSearch) ? "item-content" : ""}>
```
kieckhafer: Why do we allow ternary's in React components, but nowhere else?

More of a discussion overall, since the code here is OK based on our docs.
https://docs.reactioncommerce.com/reaction-docs/master/styleguide#methods

imports/plugins/included/product-variant/containers/productsContainer.js
```diff
@@ -1,217 +1,33 @@
...
-
-  reactiveProductIds.set(productIds);
-
-  canLoadMoreProducts = productCursor.count() >= Session.get("productScrollLimit");
-
-  const isActionViewOpen = Reaction.isActionViewOpen();
-  if (isActionViewOpen === false) {
-    Session.set("productGrid/selectedProducts", []);
-  }
+  const isAdmin = Reaction.hasPermission("createProduct", Meteor.userId());
```
kieckhafer: Can we do this outside of a composer, and then get rid of the composer since it's the only thing inside of it? Or does the admin check need to be run inside a composer?

imports/plugins/included/product-variant/components/customer/productGridItem.js
mikemurray: `large` is one of the sizes of the media object. `small`, `medium`, 'large`, 'image', `thumbnail`

imports/plugins/included/product-variant/containers/productsContainerAdmin.js
```diff
@@ -0,0 +1,248 @@
...
+      this.ready = this.ready.bind(this);
+      this.loadMoreProducts = this.loadMoreProducts.bind(this);
+    }
+
+    ready = () => {
+      if (this.props.showNotFound === true) {
+        return false;
+      }
+
+      const isInitialLoad = this.state.initialLoad === true;
```
kieckhafer: Can this just be `const isInitialLoad = this.state.initialLoad;`?

imports/plugins/included/product-variant/containers/productsContainer.js
mikemurray: It's a reactive data source, so if your permissions change, the grid can reflect that change with the admin/customer product grid.

imports/plugins/included/product-variant/containers/productsContainerAdmin.js
```diff
@@ -0,0 +1,248 @@
...
+
+    productMediaById[product._id] = {
+      additionalMedia,
+      primaryMedia
+    };
+
+    return {
+      ...applyProductRevision(product)
+      // additionalMedia,
+      // primaryMedia
```
kieckhafer: Can we remove these commented values?

imports/plugins/included/product-variant/containers/productsContainerAdmin.js
mikemurray: The `productsContainerAdmin.js` was moved with `git mv` from `productsContainer.js` so it's a verbatim copy. Perhaps the original author intended to be triply sure that `initialLoad` was true/false and not just undefined or a string.

lib/collections/collections.js
```diff
@@ -113,6 +113,14 @@ export const Packages = new Mongo.Collection("Packages");
 
 Packages.attachSchema(Schemas.PackageConfig);
 
+/**
+ * Catalog Collection
+ * @ignore
+ */
+export const Catalog = new Mongo.Collection("Catalog");
+
+// Catalog.attachSchema(Schemas.Catalog);
```
kieckhafer: Can we remove this commented code?

imports/plugins/included/product-variant/containers/productsContainerAdmin.js
mikemurray: will do

lib/collections/collections.js
mikemurray: will remove, and add a TODO to attach a schema.

server/publications/collections/products.js
```diff
@@ -496,3 +503,35 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
...
+ * @name Products/grid
+ * @method
+ * @memberof Core
+ * @summary Publication method for a customer facing product grid
+ * @param {number} productScrollLimit - product find limit
+ * @param {object} productFilters - filters to be applied to the product find
+ * @param {object} sort - sorting to be applied to the product find
+ * @return {MongoCursor} Mongo cursor object of found products
+ */
+Meteor.publish("Products/grid", function (productScrollLimit = 24, productFilters, sort = {}) {
```
kieckhafer: Why do we use a capital `P` here as opposed to lowercase?

mikemurray: I think it was to match the name of the `Products` publication and collection name.

kieckhafer:
@willopez and I did a pair review on this. 

There are a few issues related to the `Sold Out`, `Backorder` and `Low Inventory` badges, which is the 2nd test step, but we were seeing the same issues on other branches so I wouldn't say it's a blocker for this PR.

I will make a new ticket for those issues.

Aside from that, this looks like a good PR.

willopez:
I Agree with @kieckhafer, the issues we are seeing are related to badges are existing issues. :shipit:

nnnnat:
@spencern, @mikemurray and I were able to resolve the failing test, I believe everything else has been approved and we're ready to merge.

PR #3885, opened 2018-03-02
Fix #3754: check permission before publish

### LINKED ISSUE #3754: Security: Shop ids sent from client are published
### Expected behavior
Client should not be able to control what gets published
### Actual behavior
The publisher publishes the shop whose id id sent from the client.

### Cause
[Here](https://github.com/reactioncommerce/reaction/pull/3725/files#diff-b0d0d70a497afa4201d210de916e2c6fR14), the shopIds are being sent from client to the publisher to publish.

There is a Shops publication that should publish the shops that user has "admin" permission for.
For this we use something like
```Meteor.publish("MerchantShop", () => {
   // This part is not run again as server side publish function are not reactive.
   const query = { _id: { $in: Reaction.getShopsOfUser(this.userId) } };
   return Shop.find(query);
})
```
When a new shop is added the code inside the publisher is not run again and hence the new shop is not published to the user.
Now one solution to this would be to observe the whole Shops collection and then check when a new shop is added and then refresh this publication.
Something like
```Shops.find().observe({
  added() {
    // refresh MerchantShops publication
  }
})
```
But the problem with this is the way we add shops, we first add the shop to the `Shops` collection and then later add permission to the user.
So what will happen is when we run the publish code again `Reaction.getShopsOfUser(this.userId)` will not contain the new shop(as user still hasn't receive permission for it) and therefore the new shop will not be published.
A simple solution is to change the publisher to
```Meteor.publish("MerchantShop", (shopsOfUser) => {
   // This part is not run again as server side publish function are not reactive.
   const query = { _id: { $in: shopsOfUser} };
   return Shop.find(query);
})
```
and pass `Reaction.getShopsOfUser(this.userId)`  as shopsOfUser from the client, but this introduces the security bug that I mentioned.
A quick solution for this would be to again check inside the publisher if `Reaction.getShopsOfUser(this.userId) === shopsOfUser`. This should fix all the security issues.
But do you think that is OK to do.
Also all this would run only when a shop is added for a user(basically only when the user click the Become a seller button on profile) or when the user changes.

### Versions
`release-1.8.0`

Akarshit:
Resolves #3754  
Impact: **minor**  
Type: **bugfix**

## Issue
The shops to publish that were being sent from the client were not being checked for access.

## Solution
Added a check to check if the user has permission to see the shops that are being asked to publish.

## Testing
2. Login as admin, enable Marketplace, check checkbox to allow users to register as sellers
3. Logout and register a new account
4. Visit profile, click "Become a seller"
5. Observe no error and everything working.


zenweasel:
@Akarshit Make sure you request a reviewer when you submit a PR, otherwise Jeremy is not going to know he needs to look at it

Akarshit:
Sorry, missed it somehow.

server/publications/collections/shops.js
```diff
@@ -14,6 +15,11 @@ Meteor.publish("PrimaryShop", () => Shops.find({
 Meteor.publish("MerchantShops", function (shopsOfUser = Reaction.getShopsForUser(["admin"], this.userId)) {
   check(shopsOfUser, Array);
 
+  // Check if user has permissions for those shops
+  if (!isEqual(Reaction.getShopsForUser(["admin"], this.userId), shopsOfUser)) {
```
spencern: Can we not use the `hasPermission` method here?

spencern: @Akarshit

Akarshit: @spencern AFAIK To use `hasPermission` here, I would have to do something like
```
const allow = shopsOfUser.reduce((allow, shop) => allow && Reaction.hasPermission(...), true);
```
The `getShopsForUser` kind of contains the above code, so I used it.

Though I can change it if you want.

spencern: I think we should abstract this functionality into a common Reaction API then. We can use either method but should be well documented. Even with your comment on line 18, I'm not entirely clear on how that method is working which is why I thought using `hasPermissions` might work first.

Akarshit: >  abstract this functionality into a common Reaction API

I didn't get what do you mean by this.

spencern: @Akarshit I mean that if we have a common need to check the permissions of a user for all shops, we should probably have a Reaction method such as `Reaction.hasPermissionForAnyShop` or whatever the method is. That method could wrap the calls to getShops for users and be a well-documented way to accomplish that specific task.

Akarshit: OK doing that.

Akarshit:
@spencern Updated.

server/publications/collections/shops.js
```diff
@@ -14,6 +14,11 @@ Meteor.publish("PrimaryShop", () => Shops.find({
 Meteor.publish("MerchantShops", function (shopsOfUser = Reaction.getShopsForUser(["admin"], this.userId)) {
   check(shopsOfUser, Array);
 
+  // Check if user has permissions for those shops
+  if (Reaction.hasPermissionForAll(["admin"], this.userId, shopsOfUser)) {
```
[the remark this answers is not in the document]
Akarshit: @spencern only active shops and shops for which the user has "admin" permission are published.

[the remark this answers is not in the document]
spencern: I think I'm still not fully understanding exactly what this publication is used for. If it's exclusively for Shop admins, do we have a different publication for shop operators?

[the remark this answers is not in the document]
spencern: Maybe it's the "active" vs "disabled" shop thing that I'm getting stuck by.

PR #3892, opened 2018-03-05
Fix #3875: card validator

### LINKED ISSUE #3875: Incorrect Card Validation in payments-authnet
Your documentation mentions that "by default" it checks to make sure the number is between 12 and 19 digits. See here: https://docs.reactioncommerce.com/reaction-docs/master/payment-packages-authnet

However, this is not the case. In the schema it checks for 12-19 digits.

https://github.com/reactioncommerce/reaction/blob/766c97ebfacec67afecedb021cb3bac79658d1f1/imports/plugins/included/payments-authnet/lib/collections/schemas/authnet.js#L48-L53

However, in the method it checks for 14-16 digits.

https://github.com/reactioncommerce/reaction/blob/766c97ebfacec67afecedb021cb3bac79658d1f1/imports/plugins/included/payments-authnet/server/methods/authnet.js#L46

Akarshit:
Resolves #3875  
Impact: **major**  
Type: **bugfix**

## Issue
Card validator regex was checking for number in the range of 14-16 though 12-19 is what it should be checking for.

## Solution
Fixed the regex and moved the validator regex to `helper` for reuse.

## Testing
1. Enable Authorize.net payment.
2. Enter a card with just 12 numbers(424242424242)
3. Observe that you don't get a `Match failed` error.
4. Try entering a number with less than those digit. Observe you get the error that the card should have atleast 12 digits.


lib/api/helpers.js
```diff
@@ -367,3 +367,29 @@ export function convertWeight(from, to, weight) {
   // if we made it here, something has gone wrong
   throw new Meteor.Error("invalid-parameter", "Invalid from or to value specified");
 }
+
+/**
+ * @name luhnValid
```
zenweasel: I wonder if we shouldn't namespace some of these helpers? What do you think @jshimko?

jshimko: Sure.  @spencern, care what that namespace is?

aaronjudd: @jshimko what would you suggest?

spencern: I haven't thought about this much, open to suggestions.

spencern: @zenweasel is this a blocker for this PR? If so can you "request changes"

spencern:
@Akarshit some conflicts here with 1.9

Akarshit:
@spencern updated.

PR #3814, opened 2018-02-21
(refactor): 2301 nnnnat removed analytics plugin

nnnnat:
Resolves #2301   
Impact: major  
Type: refactor|chore

## Issue
The current implementation of analytics has been broken for a little while now and needed removed from the `plugins/included` directory.

## Solution
Deleted analytic plugin.

## Testing
1. Startup RC and log in as an Admin.
2. Click the RC logo at the top of the admin sidebar nav to open dashboard panel.
3. Dashboard panel should not have any options for analytics in either the actions or settings sub-panels.

**Notes**
All RC test passed.
Was able to complete acceptance test.

PR #3894, opened 2018-03-06
(Fix) Add back missing browser policy

zenweasel:
Resolves N/A  
Impact: **major**  
Type: **bugfix**

## Issue
The policy file that allows us to use remove web fonts and facebook login were removed in this PR: https://github.com/reactioncommerce/reaction/pull/3814 but we need to keep the policy files to allow us to use webfonts. 

## Solution
Restore the policy files minus the remote tracking scripts. I guessed at the best directory for this but I could be convinced to use any other

## Breaking changes
None

## Testing
1. View the site in a deployed environment (I guess. I don't think you will see them in a dev environment).
1. Observe that you don't receive content warnings

(To see what the warnings look like you can view https://swag.getreaction.io)

### DIFF
--- a/imports/plugins/core/ui/server/index.js
+++ b/imports/plugins/core/ui/server/index.js
@@ -1,1 +1,2 @@
 import "./i18n";
+import "./policy";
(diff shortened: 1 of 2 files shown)


zenweasel:
I was seeing this error on `swag.getreaction.io` but now am no longer seeing it.

PR #3940, opened 2018-03-07
Fix nnnnat remove products on grid items

### LINKED ISSUE #3938: Product grid is passing data for all products to each product item component
## Prerequisites
* [X] Are you running the latest version?
* [X] Are you able to consistently reproduce the issue?
* [X] Did you search the issue queue for existing issue? [Search issues](https://github.com/reactioncommerce/reaction/issues)

## Issue Description
The [ProductGrid](https://github.com/reactioncommerce/reaction/blob/766c97ebfacec67afecedb021cb3bac79658d1f1/imports/plugins/included/product-variant/components/productGrid.js#L14) component is applying all of its props via a spread operator to each [ProductGridItems](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/included/product-variant/components/productGridItems.js) component. This is causing each ProductGridItems component to not only have it's own product data but also have the `products` & `productIds` array attached to each grid item's props. This is probably causing a lot of unnecessary rendering of components in the product grid.

*ProductGridItems Component props in React devtool*
screen shot 2018-03-07 at 3 23 12 pm

## Steps to Reproduce
*You'll need [React devtools](https://github.com/facebook/react-devtools) installed to reproduce*
1. Spin up a fresh RC shop.
2. Open React devtools and inspect the ProductGridItems component. (The node will be called `Reaction(ProductGridItems)`)
3. Verify that the `products`, `productIds` and `productsByKey` are all in the component props.

## Possible Solution
Removing the spread operator and passing down the explicit props to the ProductGridItems component.

### Versions
```sh
Node: 9.4.0
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.28.0
Reaction: 1.8.2
Reaction branch: master
Docker: 18.02.0-ce
```

nnnnat:
Resolves #3938
Impact: **major**  
Type: **performance**

## Issue
The ProductGrid component was using a spread operator to apply all of it's own props to each ProductGridItems, this was causing the products array and several other unneeded sets of data to be added to each component's props.

## Solution
* Replaced the spread operator with explicit props for the ProductGridItems component
* Updated the key to use the `product._id` instead of the `.map` index. [ref](https://reactjs.org/docs/lists-and-keys.html#keys)
* Refactored the `renderProductGridItems` to not be an arrow function and to not accept arguments since the products are available on the component's props.
* Updated the ProductGrid Proptypes.
* Fixed a small eslint issue in an unrelated file.

## Testing
*You'll need [React devtools](https://github.com/facebook/react-devtools) installed to reproduce*
 1. Spin up a fresh RC shop.
 2. Open React devtools and inspect the ProductGridItems component. (The node will be called Reaction(ProductGridItems))
 3. Verify that the products, productIds, and productsByKey are no longer in the component props.


spencern:
@mikemurray @nnnnat 
This seems like a PR that's worth finishing at some point for performance reasons. Any ideas as to what it's going to need to finish?

mikemurray:
@spencern I'm not so sure it will be beneficial. I think the props that were removed were probably required by a component further down the tree. A more worthwhile endeavor may be simplifying the back and forth between HOC and presentational components on the product grid.

There are currently about 8 components that make up the admin product grid. 4 containers and 4 presentational. It jumps back and forth multiple times as it goes between the container -> presentational components. Somewhere in there the `productIds` is probably used for drag-and-drop handling.

Also, a breaking change notice would need to be added as removing props would break any replaced components that depended on those props that were removed.

spencern:
@mikemurray in that case, it might be better to punt on this until we start building this out with our GraphQL connection. Should we just close this PR?

nnnnat:
@spencern @mikemurray  Yeah after some testing it looks like you do need to pass the `products`, `productsByKeys`, `productIds` for the drag n drop to work, might be best to just close this one since it would mean some heavy refactoring to keep all the features working.

mikemurray:
Closing in favor of rebuilding the product grid components with a GraphQL connection.

PR #3951, opened 2018-03-08
chore: remove ability to load Meteor.settings from settings.json

### LINKED ISSUE #3934: Deprecate METEOR_SETTINGS file handling
## Issue Description
Remove Reaction handling, checks, and handling for `METEOR_SETTINGS`.

Meteor will use [`METEOR_SETTINGS`](https://docs.meteor.com/environment-variables.html) that are exported to ENV variables by default.

A replacement, and updating the pattern for docs would be:
```
export METEOR_SETTINGS=$(cat settings.json)
```

```
METEOR_SETTINGS='{ "server_only_setting": "foo", "public": { "client_and_server_setting": "bar" } }'
```
Related to #3906 

The goal of this issue to remove a start up step that only adds time and is mostly a deprecated pattern, because almost all settings for both reaction and meteor come from the ENV now, and only old Atmosphere era meteor packages would be relying on the actual settings file.

## Recommended solution
* `git rm settings/*` and our internal checks/messaging/handling of said file. 
* updated docs

kieckhafer:
Resolves #3934   
Impact: **breaking**  
Type: **chore**

## Issue
Deprecate `METEOR_SETTINGS` file handling

## Solution
Remove code that handles getting data from `/settings/settings.json` to provide to `METEOR_SETTINGS`.

## Breaking changes
Settings inside your `/settings/settings.json` file will no longer be loaded in the app. These settings will need to be passes as `ENV` variables now.

## Testing
1. Reset the app, and see that any settings in `/settings/settings.json` are not applied to your instance
2. Reset the app and pass settings in as `ENV` variables, see that they are applied (https://docs.reactioncommerce.com/reaction-docs/master/configuration)

Docs PR: https://github.com/reactioncommerce/reaction-docs/pull/553
CLI PR: https://github.com/reactioncommerce/reaction-cli/pull/63


jshimko:
So right now all this does is stop you from being able to load your default admin user with a Meteor settings file. Even if you set the `METEOR_SETTINGS` env var manually, it would no longer work for the admin user config. That's fine, but I was under the impression all @aaronjudd was talking about in #3934 was removing the automatic loading of a `/settings/settings.json` when it's present (which [happens in the CLI](https://github.com/reactioncommerce/reaction-cli/blob/master/src/commands/run.js#L28-L40)). We can obviously remove that automatic loading and still support the option of using Meteor settings if you prefer it for some reason.  Then if people still wanted to use a `settings.json` file for the default admin, they could do either of these...

```sh
reaction --settings settings.json

export METEOR_SETTINGS=$(cat settings.json)
```

(both are functionally identical)

I can remove the automatic loading of the file from the CLI, but do we want to remove the ability to set up the default admin with `METEOR_SETTINGS`?  Seems like there's no harm in leaving that as an option there. I've personally never used it, but some users might find it easier to manage than setting individual env vars for each item.  I don't really care either way.  But if we want to stop looking for and automatically loading a `settings.json` that still needs to happen to resolve the original referenced issue. 

What say you @aaronjudd?

PR #3952, opened 2018-03-08
chore: upgrade react-dates to 16.3.6

### LINKED ISSUE #3443: Update react-dates version
Bithound is reporting that `react-dates` is falling behind. We should update this package provided it's got a clear upgrade path.

It looks like both a `15.5.3` and a `16.0.0` versions exist. I'd expect that the 16 version tracks with the React updates as well, but not certain.

image

mikemurray:
Resolves #3443   
Impact: **breaking**  
Type: **chore**

## Issue
Update [react-dates](https://github.com/airbnb/react-dates) package to version `16.3.6`

## Solution

1. `meteor npm install react-dates@16.3.6`
2. Change prop `renderDay` to `renderDayContents`
3. Fix styling issues with start and end dates in date ranges

## Breaking changes

- change prop `renderDay` to `renderDayContents`

## Testing

1. Install the reaction devtools package https://github.com/reactioncommerce/reaction-devtools
2. As an admin, load the orders from the small dataset from devtools panel
3. Open Orders panel and filter dates

### DIFF
--- a/imports/plugins/core/ui/client/components/calendarPicker/calendarPicker.js
+++ b/imports/plugins/core/ui/client/components/calendarPicker/calendarPicker.js
@@ -80,7 +80,7 @@ CalendarPicker.defaultProps = {
   initialEndDate: null,
 
   // day presentation and interaction related props
-  renderDay: null,
+  renderDayContents: null,
   minimumNights: 1,
   isDayBlocked: () => false,
   isDayHighlighted: () => false,
@@ -126,7 +126,7 @@ CalendarPicker.propTypes = {
   onOutsideClick: PropTypes.func,
   onPrevMonthClick: PropTypes.func,
   renderCalendarInfo: PropTypes.func,
-  renderDay: PropTypes.func,
+  renderDayContents: PropTypes.func,
   withPortal: PropTypes.bool
 };
 
--- a/imports/plugins/core/ui/client/components/calendarPicker/calendarPicker.js
+++ b/imports/plugins/core/ui/client/components/calendarPicker/calendarPicker.js
@@ -52,14 +52,21 @@ class CalendarPicker extends Component {
     });
   }
 
+  renderDayContents(day) {
+    return (
+      <span className="CalendarDay__contents">{day.format("D")}</span>
+    );
+  }
+
   render() {
-    const { focusedInput, startDate, endDate } = this.state;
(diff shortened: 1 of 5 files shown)


mikemurray:
This was branched off of `release-1.9.0` and currently targets `master`.

nnnnat:
I'm seeing a few errors from the client and server when I complete the checkout process, I'll verify these issues are on the `release-1.9.0` branch cause they don't seem related to any of these changes.

screen shot 2018-03-09 at 9 51 03 am

```sh
15:50:23.872Z  INFO Reaction: Transitioned cart 9BLcKzCfxBm4JxRnb to order eqk9aANvnzu7PQ9Le
15:50:24.653Z  WARN Reaction:
  Error: error occured while sending sms { Error: Error: Incorrect format for phone number. [invalid-parameter] [invalid-parameter]
      at formatPhoneNumber (lib/api/helpers.js:227:11)
      ...
    isClientSafe: true,
    error: 'invalid-parameter',
```

prinzdezibel:
@nnnnat Error looks the same like the one that was resolved with this PR: https://github.com/reactioncommerce/reaction/pull/3948

nnnnat:
@prinzdezibel Yeah I think you're right, it looks like the browser console error is only happening in this PR branch and not in `release-1.9.0` we probably just need to rebase. 

I am still seeing the `Incorrect format for phone number.` error on the `release-1.9.0` branch.

Akarshit:
@nnnnat I am responsible for that error. That error comes when you enter a phone number of a different country or a non-serviceable phone number.

Is there some other case when you are seeing that error? Also this get's checked even if the SMS service is not enabled which is a issue(for which I am opening a "issue" right now).

nnnnat:
@Akarshit I'm able to reproduce it but it's a pretty specific use case, I'll add steps to reproduce in your new issue.

Akarshit:
@nnnnat I think it's OK if you don't give the steps. I am working to fix it anyway.

Akarshit:
@nnnnat Sorry I might have said/done things too soon. I am able to find the reproduction steps to see `Incorrect format for phone number.`) 
So if you are getting that it might be a good idea if you can open a issue with the steps or the full stack.

nnnnat:
@Akarshit here is the new [issue](https://github.com/reactioncommerce/reaction/issues/3965) for that error

PR #3964, opened 2018-03-09
Fix for: Cloning products is not reactive

### LINKED ISSUE #3957: Cloning products is not reactive
## Prerequisites

## Issue Description
Cloning a product via clone button does apparently have no effect. Only refreshing the page shows that the product indeed has been cloned. 

## Steps to Reproduce
1. As a admin, create a product
2. Navigate back to product grid
3. Clone the product in the `Grid settings` via clone button
4. See nothing happen
5. Refreshing the page shows up the cloned product, though

### Versions
Node: 8.2.1
NPM: 5.3.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.26.4
Reaction: 1.9.0
Docker: 17.03.1-ce

prinzdezibel:
Resolves #3957
Impact: **major**  
Type: **bugfix**

## Issue
Cloning a product via clone button does apparently have no effect. Only refreshing the page shows that the product indeed has been cloned.

## Solution/Changes
Whenever a product is cloned an (unobserved) server side cursor in publications/collections/products.js changes its results. But that does
not result in the changes being pushed to the client, because server
publications are not reactive. Pullling in one of several  meteor packages
available was not an option, therefore I've decided to trick the client
to resubscribe whenever that happens.

## Breaking changes
none expected

## Testing
1. As a admin, create a product
2. Navigate back to product grid
3. Clone the product in the Grid settings via clone button
4. See the cloned product appearing immediately in product grid.


jshimko:
The testing steps do work as expected, but as @zenweasel mentioned, this problem doesn't happen on `master`, so do we know what changed to require this PR?

spencern:
@jshimko I think that's the right question to be asking. This did use to work and I'm not sure who changed the publication last or what else would have been done to break this.

prinzdezibel:
The bug creeped in as side effect of https://github.com/reactioncommerce/reaction/pull/3755.
Formerly, whenever a revision was created, the whole product document was pushed to the client, which in turn has caused its own problems with outdated fields being pushed to the client due to async nature of database queries.

fix_2246__setting_or_changing_a_products_perma-link_causes_hard_refresh_by_prinzdezibel_ _pull_request__3755_ _reactioncommerce_reaction

spencern:
@prinzdezibel did you test this with large numbers of products? Does it impact performance in any significant way?

prinzdezibel:
I will test this with many products tomorrow. Because this observer does only affect non-published revisisions I don't expect too much data going over the wire due to this. But I'll verify, to be sure.

spencern:
Specifically worth testing with the Shopify importer (or any method of importing products) as they tend to overload observers like this

prinzdezibel:
I've done some general performance tests regarding the "Product" & "Products" subscriptions. Can't say it was specifically scientific, but I couldn't feel a difference in the perceived performance between master and release-1.9.0.

I tested the Product Grid and the PDP page. Specifically I imported 1000 products from Brent's development plugin and also imported via Shopify connector in a second run. What I did then to get some load on the "Revisions" subscription was to reorder the products in the "Products grid". This results in new revisions for all products that are loaded in the grid, because their order changed.

But I couldn't recognize any difference in performance, nor did I when navigating to PDP.

What I found is that there's a problem with the badges in product grid that signal that a product is "dirty", though. I.e. the revision workflow is `revision/update`. I'm going to file a separate ticket for that.

prinzdezibel:
@jshimko If you find time the today or tomorrow, it would be great if you could review this one. thanks.

spencern:
I tested this in depth with @mikemurray and later with @zenweasel and found that this did not have both reactivity on product cloning as well as accuracy of the revision indicators.

We may want to consider closing this issue in favor of having someone rewrite the core revisions publication and subscriptions.

prinzdezibel:
Yes, after merging 1.9.0 I've ran into an infinite resubscription loop on the client, which originated from repeatedly mounting/unmounting the container (which is an ok thing, but that seemed not to happen before). 

Now I've taken another approach which seems to work reliably. It does not involve the server to decide if the subscription should be refreshed, but rather the clone action itself requests the resubscription. 

We may test if this solves our problems and if not,  I'm fine with closing this PR in favor of something else.

zenweasel:
@prinzdezibel the other issue we saw was that the "dirty" badge was not persisting after a refresh

prinzdezibel:
Pushed the alternative approach and changed merge base. Will check the dirty badge in 30 minutes or so..

zenweasel:
@spencern MJ and I talked about this yesterday and he seemed to have this as a solution but feel free to just close this PR if it's not working or you want to look at another approach.

lib/api/products.js
```diff
@@ -3,6 +3,7 @@ import orderBy from "lodash/orderBy";
 import _ from "lodash";
 import { Meteor } from "meteor/meteor";
 import { ReactiveDict } from "meteor/reactive-dict";
+import { Session } from "meteor/session";
```
spencern: I'd prefer to avoid using Meteor session, we removed that dependency a while ago. If you need something reactive, our preference has been to use reaction var

prinzdezibel: done.

imports/plugins/included/product-variant/containers/productsContainerAdmin.js
```diff
@@ -159,7 +157,12 @@ function composer(props, onData) {
   }
 
   const queryParams = Object.assign({}, tags, Reaction.Router.current().query, shopIds);
-  const productsSubscription = Meteor.subscribe("Products", scrollLimit, queryParams, sort, editMode);
+  let productsSubscription = Meteor.subscribe("Products", scrollLimit, queryParams, sort, editMode);
+  if (resubscribeAfterCloning.get()) {
+    resubscribeAfterCloning.set(false);
+    productsSubscription.stop();
+    productsSubscription = Meteor.subscribe("Products", scrollLimit, queryParams, sort, editMode);
+  }
```
aldeed: Would this be enough?

```js
const productsSubscription = Meteor.subscribe("Products", scrollLimit, queryParams, sort, editMode);
if (resubscribeAfterCloning.get()) {
  resubscribeAfterCloning.set(false);
}
```

Doing the `Meteor.subscribe` call twice seems unnecessary. The important thing is triggering this whole function again through the reactive var. If you need to explicitly stop it first (before the first call) you could, but I think it stops itself due to being tracked anyway.

PR #3825, opened 2018-02-22
Remove collection hooks for Products collection

willopez:
Resolves #3634
Impact: Breaking
Type: Refactor

## Issue
Remove dependency on `matb33/meteor-collection-hooks` for the Products collection. This PR includes hooks related to Revisions. Further, there are additional related hooks for: Orders, Cart, see testing instructions for more details. 

## Solution
Refactor collection hooks to use Reactions hook system, and call explicitly were necessary. 

## Testing Instructions

**Revisions Before|After Insert Hook** - inserts a product revision into the `Revisions` collection before or after inserting a product into the `Products` collection.

Hooks Location: `imports/plugins/core/revisions/server/hooks.js`

Hook Names: `beforeInsertCatalogProductInsertRevision | afterInsertCatalogProductInsertRevision ` - perform a project search to find call sites

**Testing Procedures:**

User story 1: Create a new product, uses `afterInsertCatalogProductInsertRevision` hook

1. Create a new product
2. Verify a new revisions were inserted, one for the parent product `type=simple` and another one for `type=variant`
3. Verify new revision data was not published to `Products` collection
 

User story 2: Create a new varian on an existing product, uses `beforeInsertCatalogProductInsertRevision`

1. On an existing product create a new variant
2. Verify a new revisions was inserted for the new variant.

User story 3: Clone an existing product, uses `beforeInsertCatalogProductInsertRevision`

1. Clone a product
2. Verify corresponding revision were inserted.

User story 4: Clone an existing variant, uses `beforeInsertCatalogProductInsertRevision`

1. Clone a variant
2. Verify corresponding variant was inserted

___

**Revisions Before Update Hook** - updates a product revision in the `Revisions` collection before updating a product into the `Products` collection. The hook's return value determines whether the upderlying product should be updated.

Hooks Location: `imports/plugins/core/revisions/server/hooks.js`

Hook Names: `beforeUpdateCatalogProduct` - perform a project search to find call sites

User story 1: Update variant's position
1. On an existing product, have at least two variants created.
2. Rearrange variants positions
3. Verify position field: `index` was updated in the `Revisions` collection.
4. Publish product
5. Verify position field: `index` was updated in the `Products` collection

User story 2: Update a single product or variant field.
1. Update a product or variant field either: Title, price, visivility, vendor, description etc.
2. Verify field is updated in the corresponding revision.
3. Publish product
4. Verify field was updated in `Products` collection

User story 3: Update a product's tags
1. Update an existing product tag
[...]

PR #3984, opened 2018-03-12
Fix for "dirty badge in product grid does not work"

### LINKED ISSUE #398: Deeper security review of the Cart publication
Make sure we aren't leaving an open security hole w/ the Cart pub.

Why take the user id here? (Cart)

```
dig into 'getCurrentCart' that's the heart of session/user.  You can be a registered account (userId) and share your cart across all logged in sessions, or not be a user/ logged out and use just sessions.
```

### LINKED ISSUE #3982: "Dirty badge" in product grid does not work
## Prerequisites
* [x]  Did you search the issue queue for existing issue? [Search issues]

## Issue Description
The product grid should display unpublished changes via red dot (a badge). This tells the admin that products shown in the grid hold un-commited changes.

## Steps to Reproduce
1. As an admin , create at least two products
2. Navigate to the product grid
3. Swap the order of the products via drag'n'drop
4. See the red dot on each product that tells the admin about the un-commited (unpublished) changes
5. Reload the page
6. See the dots are gone.

## Possible Solution
Could be similar in scope than this PR: https://github.com/reactioncommerce/reaction/pull/3970 , which is somewhat analogous, but focuses on the `product` subscription

### Versions

Node: 8.2.1
NPM: 5.3.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.26.4
Reaction: 1.9.0
Docker: 17.03.1-ce

prinzdezibel:
Resolves #3982 
Impact: **major**  
Type: **bugfix**

## Issue
The product grid should display unpublished changes via red dot (a badge). This tells the admin that products shown in the grid hold un-commited changes.

## Solution 

This PR fixes the issue described above AND a problem on master where revisions are over-published (see attached gif).

This change is very similar to what have been done in this companion PR: https://github.com/reactioncommerce/reaction/pull/3970 

##  Changes
Product revisions need to be pushed to client, even if the product document (that contain all other fields besides __revisions) is not published yet.

Just adding
  `this.added("Products", revision.documentId, { __revisions: [revision] });`
in line 400 is not sufficient, because then revisions would be over-published. (Which is actually the case in master branch right now.)

To prevent over-publishing, the Revision observer needs to be more specific and select only those revisions, that are not published AND are shown in the product grid. This constraint resulted in the new function "extendSelectorWithVariants".

## Breaking changes
none expected

## Testing
1. As an admin , create at least two products
2. Navigate to the product grid
3. Swap the order of the products via drag'n'drop
4. See the red dot on each product that tells the admin about the un-commited (unpublished) changes
5. Reload the page
6. All dots should still be there until "Publish all" button is pressed (ensure that no product is selected, otherwise the button will have the text "Publish" and does only affect the currently selected product.

### DIFF
--- a/imports/plugins/included/product-variant/containers/productsContainer.js
+++ b/imports/plugins/included/product-variant/containers/productsContainer.js
@@ -178,17 +178,15 @@ function composer(props, onData) {
     shopId: { $in: activeShopsIds }
   });
 
-  const sortedProducts = ReactionProduct.sortProducts(productCursor.fetch(), currentTag);
-  Session.set("productGrid/products", sortedProducts);
-
   const productIds = [];
-  const stateProducts = sortedProducts.map((product) => {
+  const products = productCursor.map((product) => {
     productIds.push(product._id);
 
-    return {
-      ...applyProductRevision(product)
-    };
+    return applyProductRevision(product);
   });
+
+  const sortedProducts = ReactionProduct.sortProducts(products, currentTag);
+  Session.set("productGrid/products", sortedProducts);
 
   reactiveProductIds.set(productIds);
 
@@ -201,7 +199,7 @@ function composer(props, onData) {
 
   onData(null, {
     canLoadMoreProducts,
-    products: stateProducts,
+    products: sortedProducts,
     productsSubscription
   });
 }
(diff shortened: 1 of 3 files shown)


prinzdezibel:
TODO: Placeholder for gif that shows over-publication of Revision collection on master

prinzdezibel:
@mikemurray Yes, that's what I'm thinking as well. My original intention was to put that into a own ticket to keep the changes more managable. It does also affect the product subscription in the same way ( I assume this is what you mean regarding PR #3825, right?)

mikemurray:
Approved. Fix an issue with sorting on the grid with revisions. Waiting on tests to pass / a double-check review from @willopez

willopez:
Badges are now updating reactively and repositioning products works. 👍

Issue #3518, opened 2018-01-19
Avalara does not catch address validation when API throws an error because of invalid address


zenweasel:
### Expected behavior
When the API throws an error that is about address validation that error should be bubbled up to the user and dealt with

### Actual behavior
Error is thrown in the server console, but not addressed in the client and checkout continues w/o calculating tax.

### Steps to reproduce the behavior
1. Login as admin
1. Enable Avalara
1. Add an item to cart and begin to checkout
1. For address enter this address
US
Fake Name
123 Main St.
LA, 90036
1. Observe that this error is thrown: 
```
Reaction: failed [400] {"error":{"code":"GetTaxError","message":"The address number is out of range","target
````
1. Observe that taxes are not calculated.

### Versions
```
Node: 8.9.1
NPM: 5.5.1
Meteor Node: 8.9.3
Meteor NPM: 5.5.1
Reaction CLI: 0.24.2
Reaction: 1.6.4
Reaction branch: bug-3474-akarshit-tax-logic
Docker: 17.12.0-ce
```

zenweasel:
Also similar error if you enter a large mismatch between postal code and city/state. (just change the state to a random one)

Akarshit:
@zenweasel 
Whenever we update the shipping address, because of [this](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/taxes/server/hooks/taxes.js#L21) hook, the tax is recalculated. This is the place where we know if Avalara would be able to process tax for this or not, but this is done in async to the saving of address workflow so its a little difficult notify the user here.

One idea could be to do this in sync(before saving the address) and not let the user proceed until Avalara is able to process tax on the address.
Any other ideas?

spencern:
@zenweasel does this still happen on the release-1.8.0 branch?

zenweasel:
So digging more into this flow and it looks like the address validation may have been broken since we changed the `register.js` "provides" parameter from a string to an array. So validation errors were only being throw when we actually try to calculate tax which is why this error appeared. This is fixed now but there's more errors related to validation feedback that I want to take care of.

@rymorgan 
So I think we need to clarify this flow, there are two different scenarios we encounter with address validation. This is not really specific to Avalara as most address validators I looked worked similarly.

1. **API error level validation.**
This is the "I have no idea what to do with this address" error. For this sort of flow I think we need to 
 - Display the error to the user (although the Avalara ones are not always super meaningful but it's all we have)
[...]

PR #3990, opened 2018-03-13
(refactor): 2053 nnnnat address book to react

### LINKED ISSUE #2053: Convert address book to react
Part of #2040 and takes care of part of `Checkout`

---

Address book to be converted to react.

Things it must do:

- Work as well as it does now
- Work on both the user profile page and the cart checkout

nnnnat:
Resolves #2053   
Impact: **breaking**  
Type: **refactor**

## Issue
The AddressBook components were still using Blaze templates.

## Solution
Refactored the blaze AddressBook components to React
 * Created a new AddressBookContainer that binds data and reducers to the component.
 * Created a new AddressBook component that creates the base layout and handles data changes
 * Created a new AddressBookForm component that handles adding/editing an address.
 * Created a new AddressBookGrid component that displays the shipping/billing address.
 * Updated the Profile and Checkout blaze templates to use the new react AddressBook

## Breaking changes
The new React components should be working the same the old ones, but they are no longer blaze templates. 

All of the new components are available from the `@reactioncommerce/reaction-components` package.

## Testing
The AddressBook needs to be tested from both the Profile and Checkout views.

### Setup
**Profile**
1. Log in as an Admin or create a new account.
2. From the drop-down click the profile link.
3. Follow the testing steps below.

**Checkout**
1. As an anonymous customer, add a few items to the Cart and go to check out.
2. Enter an email address.
3. Follow the testing steps below.

### Address Book Test
**Adding a new address**
1. Enter a new shipping/billing address.
2. Click the "save and continue" button.
3. The new address should now display in the AddressBookGrid and checkout step 3 should be available.

**Adding a second address**
1. Now click the "Add Address" button at the top of the AddressBook
2. Enter the second address.
3. Both addresses should display in the AddressBookGrid.

**Changing Shipping/Billing address**
1. Click on the new address to make it the default billing address
2. The blue checkmark should show up on the new address and the shipping should recalculate.
3. Try switching the default billing/shipping address using the keyboard.

**Editing an address**
1. Click the edit button of an address in the AddressBookGrid.
2. Change the some of the fields and click "save and continue".
3. The updated address should now show in AddressBookGrid.

**Removing an address**
1. Click the delete/trash button of an address in the AddressBookGrid.
2. Address should now be removed from the grid.

### Meteor methods
I created some helper/reducer methods for the AddressBook inside the AddressBookContainer. I'm just wrapping the Meteor methods that do the CRUD operations on the account's addresBook. I've copy most of that code out of the template helpers and I'm not confident I'm handling the callbacks correctly.

### Validation
**General Validation**
Currently, this version of the AddressBook only shows a toast alert if there is an address validation issue, I know we have a few issues around form/address validation being worked on and I wanted to get this reviewed so others can weigh in on how to best handle displaying field errors.

**Avalara Validation and the AddressReview view**
While testing with Avalara enabled I was unable to get the AddressReview template to show even if Avalara threw an error on the server, the invalid address would be added to the address grid anyway.
Could be related to issue #3518 could be user error.

I left the blaze addressBook template code in the project, for now, in case it's needed as a reference but once this PR is approved I'll remove those files.


imports/plugins/core/accounts/client/containers/addressBookContainer.js
```diff
@@ -0,0 +1,131 @@
...
+    static propTypes = {
+      addressBook: PropTypes.array, // array of address objects
+      heading: PropTypes.object // heading content
+    }
+
+    addAddress = (address) => addAddress(address);
+
+    updateAddress = (address, property) => updateAddress(address, property);
+
+    removeAddress = (_id) => removeAddress(_id);
```
aldeed: These three lines aren't needed. Just remove `this.` below and pass the three functions directly.

imports/plugins/core/accounts/client/containers/addressBookContainer.js
```diff
@@ -0,0 +1,131 @@
...
+ * @param {Object} address - address to be updated.
+ * @param{String} property - property to be updated.
+ */
+function updateAddress(address, property) {
+  if (property) {
+    Meteor.call("accounts/addressBookUpdate", address, null, property);
+  } else {
+    Meteor.call("accounts/addressBookUpdate", address);
+  }
+}
```
aldeed: You should deal with any potential errors by having `Meteor.call` callbacks like the other two functions have. But also, I recommend making all of these return promises, which reject in case of error. Then, rather than calling `Alerts.toast` directly in these action functions, you should supply an `onError` function to all of the components.

So for example removeAddress becomes:

```js
function removeAddress(_id) {
  return new Promise((resolve, reject) => {
    Meteor.call("accounts/addressBookRemove", _id, (error, result) => {
      if (error || !result) {
        reject(i18next.t("addressBookGrid.cantRemoveThisAddress", { err: error && error.message }));
      } else {
        resolve(result);
      }
    });
  });
}
```

And then in your component you do

```js
onRemove = (_id) => {
  const { onError, removeAddress } = this.props;
  removeAddress(_id).catch(onError);
}
```

passing an `onError` similar to:

```js
function onError(errorMessage) {
  Alerts.toast(errorMessage, "error");
}
```

This keeps better separation which will make testing and debugging easier.

imports/plugins/core/checkout/client/templates/checkout/addressBook/addressBook.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+Template.checkoutAddressBook.helpers({
+  /**
+   * @method AddressBook
+   * @summary returns a component for updating a user's address.
+   * @since 2.0.0
+   * @return {Object} - contains the component for updating a user's address.
+   */
+  AddressBook() {
+    const { status, position } = Template.instance().data;
```
aldeed: `template.data` is the nonreactive one. Use `Template.currentData()` instead to ensure it will update when `status` or `position` change after initial render.

```js
const { status, position } = Template.currentData();
```

aldeed:
@nnnnat Overall this looks great, and was clearly a lot of work. Nice job! I suggested some changes to try to better separate certain things, deal with edge cases, and improve prop types.

nnnnat:
@aldeed I updated this PR with the requested changes and it's ready for review.

imports/plugins/core/accounts/client/containers/addressBookContainer.js
```diff
@@ -0,0 +1,214 @@
...
+      Object.keys(states).forEach((i) => {
+        regions.push({
+          label: states[i].name,
+          value: i
+        });
+      });
+    }
+    country[key] = regions;
+    regionsByCountry = { ...regionsByCountry, ...country };
+  });
```
aldeed: This will create a new map every loop, which is inefficient. Since you're just building the `regionsByCountry` map right now, you don't have to worry about mutating it causing problems. So you can do:

```js
const regionsByCountry = {};
```

And then skip the `country` variable and instead directly do `regionsByCountry[key] = regions;`

imports/plugins/core/accounts/client/containers/addressBookContainer.js
```diff
@@ -0,0 +1,214 @@
...
+  const countries = Countries.find().fetch();
+  const template = Template.instance();
+  const { data } = template || { data: undefined };
+  const { heading: templateHeading } = data || { heading: undefined };
+  const shop = Collections.Shops.findOne();
+  const shopCountries = shop.locales.countries;
+
+  let regionsByCountry;
+  Object.keys(shopCountries).forEach((key) => {
+    const { states } = shopCountries[key] || undefined;
```
aldeed: The end of this should be `|| {}` since you can't destructure `undefined`.

imports/plugins/core/accounts/client/components/addressBook.js
```diff
@@ -0,0 +1,397 @@
...
+    onError: PropTypes.func,
+    /**
+     * regions by county
+     */
+    regionsByCountry: PropTypes.shape({
+      countryCode: PropTypes.arrayOf(PropTypes.shape({
+        label: PropTypes.String,
+        value: PropTypes.String
+      }))
+    }),
```
aldeed: This prop type is wrong. There isn't a key named "countryCode" but rather one arbitrary key for each country code. So since this is a map, it's OK to do `regionsByCountry: PropTypes.object` instead of shape.

imports/plugins/core/accounts/client/components/addressBookForm.js
```diff
@@ -0,0 +1,336 @@
...
+    hasAddress: PropTypes.bool,
+    /**
+     * regions by county
+     */
+    regionsByCountry: PropTypes.shape({
+      countryCode: PropTypes.arrayOf(PropTypes.shape({
+        label: PropTypes.String,
+        value: PropTypes.String
+      }))
+    })
```
aldeed: See other comment. `regionsByCountry: PropTypes.object`

imports/plugins/core/accounts/client/components/addressBookGrid.js
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  /**
+   * @method defaultShippingAddressId
+   * @summary getter that returns ether the default shipping address _id or and empty string.
+   * @since 2.0.0
+   * @return {String} - default shipping address _id or empty string.
+   */
+  get defaultShippingAddressId() {
+    const { addressBook } = this.props;
```
aldeed: Since you're calling `.find` blindly on this, I recommend the same `this.addressBook` getter pattern you did in the other component.

imports/plugins/core/accounts/client/components/addressBookGrid.js
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  /**
+   * @method defaultBillingAddressId
+   * @summary getter that returns ether the default billing address _id or and empty string.
+   * @since 2.0.0
+   * @return {String} - default billing address _id or empty string.
+   */
+  get defaultBillingAddressId() {
+    const { addressBook } = this.props;
```
aldeed: Since you're calling `.find` blindly on this, I recommend the same `this.addressBook` getter pattern you did in the other component.

imports/plugins/core/accounts/client/components/addressBookGrid.js
```diff
@@ -0,0 +1,224 @@
...
+    edit: PropTypes.func,
+    /**
+     * removes the selected address
+     */
+    remove: PropTypes.func,
+    /**
+     * selects a default shipping or billing address
+     */
+    select: PropTypes.func
+  }
```
aldeed: Add default no-ops for all functions since you call them without checking whether they're set.

```js
static defaultProps = {
  edit() {},
  remove() {},
  select() {}
};
```

imports/plugins/core/accounts/client/containers/addressBookContainer.js
```diff
@@ -0,0 +1,214 @@
...
+ * @summary helper function that adds an address in the account's addressBook via a meteor method.
+ * @since 2.0.0
+ * @param {Object} address - address to be added.
+ * @return {Promise}
+ */
+function addAddress(address) {
+  return new Promise((resolve, reject) => {
+    Meteor.call("accounts/validateAddress", address, (error, result) => {
+      if (error || !result) reject(i18next.t("addressBookAdd.failedToAddAddress", { err: error.message }));
+      if (result && result.validated) {
```
aldeed: This second `if` leaves the possibility of a swallowed error, when `result.validated` is falsy. I think it'd be better to remove this second `if` and change above to `if (error || !result || !result.validated)`

imports/plugins/core/accounts/client/containers/addressBookContainer.js
```diff
@@ -0,0 +1,214 @@
...
+ * @param {Object} address - address to be updated.
+ * @param{String} property - property to be updated.
+ * @return {Promise}
+ */
+function updateAddress(address, property) {
+  return new Promise((resolve, reject) => {
+    if (property) {
+      Meteor.call("accounts/addressBookUpdate", address, null, property, (error, result) => {
+        if (error || !result) reject(i18next.t("addressBookGrid.somethingWentWrong", { err: error.message }));
+        if (result) resolve(result);
```
aldeed: If for some reason `error` and `result` are ever both truthy, this would end up calling `reject` and `resolve`, which results in an error. It's best to use `else` for all of these situations to ensure it will never call both.

```js
if (error || !result) {
  reject(i18next.t("addressBookGrid.somethingWentWrong", { err: error.message }));
} else {
  resolve(result);
}
```

With promises we always want to make sure exactly one of resolve or reject will be called eventually, down every possible path.

imports/plugins/core/accounts/client/containers/addressBookContainer.js
```diff
@@ -0,0 +1,214 @@
...
+      });
+    } else {
```
aldeed: I'm not entirely sure that the `if (property)` check and this `else` are necessary. If `property` is falsy, then it will just send method arguments `address, null, null`, which will be treated the same way, looking at the server method code.

imports/plugins/core/accounts/client/containers/addressBookContainer.js
```diff
@@ -0,0 +1,212 @@
...
+      /**
+       * regions by county
+       */
+      regionsByCountry: PropTypes.shape({
+        countryCode: PropTypes.arrayOf(PropTypes.shape({
+          label: PropTypes.String,
+          value: PropTypes.String
+        }))
+      })
+    }
```
aldeed: This `regionsByCountry` prop type should be fixed, too. However, because this is a very lightweight wrapper, I'd suggest not setting the explicit propTypes here but rather grabbing them off Comp.

```js
static propTypes = Comp.propTypes
```

imports/plugins/core/accounts/client/containers/addressBookContainer.js
```diff
@@ -0,0 +1,212 @@
...
+  return new Promise((resolve, reject) => {
+    Meteor.call("accounts/validateAddress", address, (error, result) => {
+      if (error || !result || !result.validated) reject(i18next.t("addressBookAdd.failedToAddAddress", { err: error.message }));
+      Meteor.call("accounts/addressBookAdd", address, (err, res) => {
+        if (err || !res) {
+          reject(i18next.t("addressBookAdd.failedToAddAddress", { err: err.message }));
+        } else {
+          resolve(res);
+        }
+      });
```
aldeed: `reject` does not return out of the function, so for errors in the outer call it will still call `accounts/addressBookAdd` with this logic. You can just add an explicit return:

```js
if (error || !result || !result.validated) {
  reject(i18next.t("addressBookAdd.failedToAddAddress", { err: error.message }));
  return;
}
```

nnnnat:
@aldeed I've made the requested changes and fixed a few issues I caused when making changes earlier today. Everything should be working again and I believe I've got all the requested changes in. I was having one issue with the updateAddress reducer, I refactored it like you suggested removing the `if (property)` check to something like this

```js
function updateAddress(address, property) {
  return new Promise((resolve, reject) => {
      Meteor.call("accounts/addressBookUpdate", address, null, property, (error, result) => {
        if (error || !result) {
          reject(i18next.t("addressBookGrid.somethingWentWrong", { err: error.message }));
        } else {
          resolve(result);
        }
      });
  });
}
```
updating the default shipping/billing worked but updating the address fields would give me this Match error from the server
```sh
Exception while invoking method 'accounts/addressBookUpdate' Error: Match error: Failed Match.OneOf, Match.Maybe or Match.Optional validation
    at check (packages/check/match.js:35:15)
    at MethodInvocation.addressBookUpdate (server/methods/accounts/accounts.js:419:3)
    at packages/check/match.js:108:16
    at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1186:15)
    at Object._failIfArgumentsAreNotAllChecked (packages/check/match.js:107:41)
    at DDP._CurrentMethodInvocation.withValue (packages/ddp-server/livedata_server.js:719:19)
    at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1186:15)
    at DDPServer._CurrentWriteFence.withValue (packages/ddp-server/livedata_server.js:717:46)
    at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1186:15)
    at Promise (packages/ddp-server/livedata_server.js:715:46)
    at new Promise (<anonymous>)
    at Session.method (packages/ddp-server/livedata_server.js:689:23)
Sanitized and reported to the client as: Match failed [400]
```
I tried updating the params to be `function updateAddress(address, property = null)` and still got the match error. I reverted it to use the `if (propery)` and call two different update meteor methods since that seems to be working.

nnnnat:
@aldeed  I've updated the `accounts/addressBookUpdate` Meteor method to do `check(type, Match.Maybe(String)` and I've refactored the `updateAddress` reducer to have only one Meteor method call. The AddressBook seems to be working in both views and updating the address from both the grid and form is working.

spencern:
Closing in favor of #4054 which will include the work done here as well as the address validation work.

PR #3991, opened 2018-03-13
(feat): Create GraphQL account query

### LINKED ISSUE #3912: Create GraphQL account query
**Summary**
Create an `account` query in GraphQL.

**Implementation**
The `account` query should have the following signature
```
account(id: ID!): Account
```
It has one argument, `id` which is required returns an `Account` associated with that ID if the user requesting the account has permissions to access that account.

This query should always return the account with an `_id` that matches the `id` argument.

## Acceptance Criteria
Using GraphQL, I can use the `account(id: ID!)` query to get an `Account` by id.

kieckhafer:
Resolves #3912 
Impact: **minor**  
Type: **feature**

## Issue
Create an `account` query in GraphQL.

## Solution
Create a query with one argument, `id`, which is required, and return an Account associated with this ID, if the user calling the query has the correct permissions.

## Breaking changes
None

## Testing
1. Start up Reaction
2. Open Mongo (Robo 3T), and copy the `id` of any user
3. Use `Terminal` to Base64 encode the ID into base64 along with an `account` prefix - `echo account:{id} | base64` - and copy the encoded value
4. Navigate to `http://localhost:3000/graphiql`
5. Query to find an account from this encoded ID
```
{
   account(id: {encodedId})
}
```
6. Change the query to an incorrect ID, and see you get a "No account found" Error message

Docs: https://github.com/reactioncommerce/reaction-docs/pull/559


imports/plugins/core/accounts/server/methods/userAccountQuery.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * @name userAccountQuery
+ * @method
+ * @summary query the Accounts collection and return user account data
+ * @param {Object} context - an object containing the per-request state
+ * @param {String} id - id of user to query
+ * @return {Object} user account object
+ */
+export function userAccountQuery(context, id) {
+  const userId = context.user._id;
```
aldeed: I'd suggest passing only `userId` since this is unlikely to ever need anything else that's on `context`.

Also, `context.user` will be `null` for an anonymous request, so you'll have to guard against that. But the `user` prop will always be there. I suggest:

```js
const { _id: userId } = context.user || {};
```

imports/plugins/core/accounts/server/methods/userAccountQuery.js
```diff
@@ -0,0 +1,29 @@
...
+
+  // Check to make sure current user has permissions to view queried user
+  if (!Reaction.hasPermission("reaction-accounts", userId)) throw new Meteor.Error("User does not have permission");
```
aldeed: For the errors, follow https://docs.reactioncommerce.com/reaction-docs/master/error-handling-guide

Specifically, make sure there's a machine-readable designator as the first argument, since we'll eventually need to evaluate errors we receive in the browser code, and decide what kind of UI message to show.

example for this one: `throw new Meteor.Error("access-denied", "User does not have permission")`

imports/plugins/core/graphql/server/resolvers/account/query/account.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+export default function account(_, { id }, context) {
+  // Trasform ID from base64
+  // Returns an object. Use `.id` to get ID
+  const idFromBase64 = transformIdFromBase64(id);
+
+  // Pass decoded id (idFromBase64.id) and context into userAccountQuery function
+  const userAccount = userAccountQuery(context, idFromBase64.id);
+
+  // Return result of userAccountQuery()
```
aldeed: > // Return result of userAccountQuery()

Is pretty obvious from just the code. Comment seems unnecessary.

imports/plugins/core/graphql/server/resolvers/account/query/account.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {Object} context - an object containing the per-request state
+ * @return {Object} user account object
+ */
+export default function account(_, { id }, context) {
+  // Trasform ID from base64
+  // Returns an object. Use `.id` to get ID
+  const idFromBase64 = transformIdFromBase64(id);
+
+  // Pass decoded id (idFromBase64.id) and context into userAccountQuery function
+  const userAccount = userAccountQuery(context, idFromBase64.id);
```
aldeed: On this return object, you'll need to change all of the IDs to the base64 IDs using `transformIdToBase64`. Add `_id` to your graphql query any you'll notice that it doesn't match the one you requested (it's the Mongo one). That's because you didn't transform the result before returning it.

imports/plugins/core/accounts/server/methods/userAccountQuery.js
kieckhafer: I see what you're saying, it makes sense to only pass the one `id` in. My thought here was trying to standardize passing the parameters in, always having `context` as the first parameter.

In this case, we only do need the `id`, but in other instances we will need need more from the `context`.

aldeed: Yes that could be true. Whatever you want to do is fine. It will become clear eventually what the best pattern is.

aldeed:
@kieckhafer Now that the other GraphQL stuff is merged, I recommend `git rebase release-1.10.0` and force push this branch, and then change the base on this PR to `release-1.10.0`

kieckhafer:
Bulk of the work is done, just finishing up writing tests and it should be good to go.

spencern:
@kieckhafer you've got failing lint tests here.

PR #3993, opened 2018-03-13
Fix for "PDP Image gallery does not handle portrait sized images well"

### LINKED ISSUE #3883: PDP Image gallery does not handle portrait sized images well
## Prerequisites
*  [x] Are you running the latest version?

## Issue Description
The PDP image gallery does not handle images that are taller than they are wide appropriately. As a non-admin, the images can block off other elements of the page.
Admin:

Non-admin:

## Steps to Reproduce
1. As an admin, visit the PDP for the default product
2. Upload an image that is taller than it is wide by a significant margin (4:3 or greater) 
3. Upload a few other images of various size.

### Versions
```
Node: 8.9.0
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.28.0
Reaction: 1.8.2
Reaction branch: master
Docker: 17.12.0-ce
```

prinzdezibel:
Resolves #3883   
Impact: **major**  
Type: **bugfix**

## Issue
The PDP image gallery does not handle images that are taller than they are wide appropriately. As a non-admin, the images can block off other elements of the page.

## Solution/Changes
- Don't allow to get the image container bigger than it's outer
container.
- Set the images width and height to 100%, but prevent distortion via
object-fit rule.
- the value is set to `contain` which reflects the spec from Ryan.
Alternatively the value `cover` would also be an option, which ensures
that the whole available image area is covered.

## Breaking changes
none

## Testing
1) As an admin, visit the PDP for the default product
2) Upload an image that is taller than it is wide by a significant margin (4:3 or greater)
3) The image should not overflow its parent container. It should not cover other content.

### DIFF
--- a/imports/plugins/core/ui/client/components/media/mediaGallery.js
+++ b/imports/plugins/core/ui/client/components/media/mediaGallery.js
@@ -168,7 +168,10 @@ class MediaGallery extends Component {
 
   renderMediaGalleryUploader() {
     const { mediaGalleryWidth: containerWidth, uploadProgress } = this.props;
-    const classes = { "admin-featuredImage": Reaction.hasAdminAccess() };
+    const classes = {
+      "admin-featuredImage": Reaction.hasAdminAccess(),
+      "featuredImage": true
+    };
 
     return (
       <div className="rui media-gallery">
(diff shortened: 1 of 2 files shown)


zenweasel:
@rymorgan If you could take a look at this

rymorgan:
Looks good to me. Tested with various aspect ratios all together and it works as expected.

PR #3995, opened 2018-03-13
GraphQL Jest testing pattern

### LINKED ISSUE #3936: Create tests pattern for new GraphQL functions
Using one initial mutation and one resolver, create a branch with a proposed pattern for unit tests, preferably outside of Meteor as much as possible.

aldeed:
Resolves #3936 
Impact: **minor**  
Type: **test**

## Description
Testing pattern, Jest dependencies, and some tests for new GraphQL plugin

## Breaking changes
None

## Testing
1. Clone and `meteor npm install`
2. Verify that `npm run test:unit` and `npm run test:unit:watch` work and the tests pass.


imports/plugins/core/graphql/server/resolvers/account/query/getUserId.js
```diff
@@ -1,5 +1,7 @@
+import { Meteor } from "meteor/meteor";
+
 export default function getUserId(_, __, context) {
   const userId = context && context.user && context.user._id;
-  if (!userId) throw new Error("access-denied");
+  if (!userId) throw new Meteor.Error("access-denied");
```
kieckhafer: Do we need the longer version of the Error here too? Or is the short version sufficient for something like this?

aldeed: That's a good point, but I'm not going to fix it because I'm going to suggest we remove this query anyway. It duplicates the viewer query. Will get this merged and then remove it in a future PR soon.

PR #3997, opened 2018-03-13
Bug Fix: Restore Global Hooks

pmn4:
_(Resolves n/a)_ 
Impact: minor
Type: bugfix

## Issue
A bug was introduced in v1.8.0 which prevents "GLOBAL" hooks from running

## Solution
Simple bug fix to resolve it.

## Testing
1. I noticed this when the body classname corresponding the the current route went missing. If you put a breakpoint in `/plugins/core/router/lib/hooks#onEnter` you will see that the callback, when Global, is set as the key in _hooks dictionary, with an undefined value (as opposed to key = "GLOBAL", value = the callback).

I wrote tests to ensure the behavior is correct (running them was a challenge!)

for reference:
https://github.com/reactioncommerce/reaction/commit/b1e7c262da0180fb766e489d506b049255a0cd7b#diff-7a0053142e3429cdf8d5779fd11692c9

### DIFF
--- a/imports/plugins/core/router/lib/hooks.js
+++ b/imports/plugins/core/router/lib/hooks.js
@@ -35,24 +35,24 @@ const Hooks = {
     }
   },
 
-  onEnter(routeName, callback, ...args) {
+  onEnter(routeNameOrGlobalCallback, callback) {
     // global onEnter callback
-    if (arguments.length === 1 && typeof args[0] === "function") {
-      const cb = routeName;
+    if (arguments.length === 1 && typeof routeNameOrGlobalCallback === "function") {
+      const cb = routeNameOrGlobalCallback;
       return this._addHook("onEnter", "GLOBAL", cb);
     }
     // route-specific onEnter callback
-    return this._addHook("onEnter", routeName, callback);
+    return this._addHook("onEnter", routeNameOrGlobalCallback, callback);
   },
 
-  onExit(routeName, callback, ...args) {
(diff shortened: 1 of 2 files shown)


imports/plugins/core/router/lib/hooks.js
```diff
@@ -35,24 +35,24 @@ const Hooks = {
     }
   },
 
-  onEnter(routeName, callback, ...args) {
+  onEnter(routeNameOrGlobalCallback, callback) {
     // global onEnter callback
-    if (arguments.length === 1 && typeof args[0] === "function") {
```
pmn4: `args[0]` is the bug. in this case, since there are two named parameters, `args[0]` === `arguments[2]` which is always undefined. Since we name the parameters, there is no need to use the `args` array

zenweasel:
Pat, I think I already put in a fix for this. Also, please don't request specific reviewers. We have a whole process for who reviews what.

zenweasel:
Pat, I think this is already fixed in `release-1.9.0`. Can you double-check? Nice work on the tests though. Tempted to bring this in just to get those.

imports/plugins/core/router/lib/hooks.js
zenweasel: I think this came about as trying to fix `prefer rest over arguments` linting.

zenweasel:
I'm going close as this was already fixed and it has conflict even though I would love to have the tests

PR #3998, opened 2018-03-13
Remove unused account exemption feature

aldeed:
Impact: **minor**  
Type: **chore**

Removes unused form and related code, as suggested by @aaronjudd in Account schemas design meeting.

## Breaking changes
I'm told this was never implemented fully so I don't expect this will break anything. But technically, `account.taxSettings` object will no longer be used, which is a breaking change if someone is setting that outside of Reaction.

## Testing
Removed Avalara plugin code that was seemingly not needed. Verify that everything related to Avalara plugin still works.


aldeed:
@zenweasel @mikemurray Let me know if this seems ok to you.

zenweasel:
@aldeed I am not sure this is truly not needed. This allows an account to be set as tax exempt and worked at one time.

aldeed:
@zenweasel The blaze form to set it up ("taxSettingsPanel") doesn't seem to be rendered anywhere, so how are people able to use it? @aaronjudd said he didn't think info like this should be on an account record in the first place.

Should I remove just the blaze template but not the logic that uses the `taxSettings` object? My main goal here is removing the autoform (or converting it to React if it's truly needed).

zenweasel:
I think when this got converted to React this was lost because it's not visible unless you have Avalara enabled. But the option to mark a customer as tax exempt needs to available somewhere. Again, I know this existed and worked at some point and if we remove it we would need to reimplement it. If you want to remove the unused blaze template I guess that's fine

zenweasel:
This probably needs to used for any other tax method. If we don't store it on account I don't know where else we would store it?

aldeed:
I think the feedback was to store it with other Avalara settings. Like a `taxExemptAccounts` array. I will close this PR and convert the autoform to React and make it visible again instead. Thanks for the explanation.

zenweasel:
@aldeed thanks

PR #3999, opened 2018-03-14
Fix #3821: use numericInput for pricing

### LINKED ISSUE #3821: Prices should not be converted in admin view

Akarshit:
Resolves #3821  
Impact: **major**  
Type: **bugfix**

## Issue
The component's value was changing because stale value was being stored in state.

## Solution
Used the `NumericInput` component modified by MJ for pricing.

## Testing
- As an admin, set the shop base currency to EUR.
- Add a product
- Input a value for price.
- The value **will not** get converted as soon as the field blurs.
- Publish the product, and observe that the price value was retained as is.


Akarshit:
Looking into why the tests are failing.

Akarshit:
Passing the test now :)

imports/plugins/included/default-theme/client/styles/numericInput.less
```diff
@@ -0,0 +1,12 @@
...
+        > input {
+            width: 100%;
+            padding: 6px 12px;
+            border: 1px solid #cccccc;
+            border-radius: 2px;
+            background-color: #fafafa;
+            color: #505558;
+        }
+    }
+}
```
jshimko: Minor linting... this should all be 2 space indents.

Akarshit:
@jshimko The error was 
1. You set your shop's currency as USD.
2. Then the currency rates are downloaded.
3. You set your shop'd currency as GBP. (but still rate of GBP is 0.71)
4. When the currency was being displayed in PDP, the rate was getting multiplied with the original value.
5. This was causing the values to be displayed incorrectly.

Fix ->
1. I have stopped currency conversion for admin view.

Also fixed the indenting error.

jshimko:
After a fresh shop startup, I set my default currency as EUR and then tried editing the price of an existing product.  That's what ended up with the results I reported.  Is that expected behavior though?  At the very least, that's different than the current master branch behavior.

Akarshit:
@jshimko Yes, that should be fixed now.

jshimko:
I guess I don't see how my comment above isn't a bug.  

- fresh clone of this branch
- log in as admin
- set shop currency to EUR
- set current user currency to EUR
- create a new product
- set the price
- observe a percentage conversion being made on that price (presumably based on US dollars)

This doesn't happen on master and I don't see a way to set a price in EUR and _not_ have it converted.  Am I missing something?

Akarshit:
@jshimko I tried the above steps but couldn't reproduce the bug after the last commit. Maybe I missed something checking again.

Akarshit:
@jshimko Would it be best to pair on this, then maybe I would have better idea what the issue it.

jshimko:
Yeah, sorry.  I think maybe I hadn't pulled the last commit.  I just did all of the testing again on the latest and I think we're good to go.  Thanks!

PR #3959, opened 2018-03-09
Update circleci config to circleci workflows 2

impactmass:
Related to #2858, cont of #3618  
Type: **feature|chore**

# Changes
Rewrites CircleCI config.yml to use Workflows v2. It contains the following Workflow jobs:
  - `docker-build`: builds the production bundle image
  - `docker-push`: pushes the production bundle image to Docker Hub. Requires `docker-build`.
  - `deploy-docs`: builds JSDoc files and deploy them to S3. Filtered to master branch only
  - `test`: Runs `reaction test`
  - `eslint`: Runs `npm run lint`

In a later PR, we'll add Snyk job to run Snyk tests. I didn't include that in this PR since the project doesn't have Snyk setup yet.

PR #4002, opened 2018-03-14
Add Snyk check to CI config; Bump base version

impactmass:
Relates to #3959.

## Adds Snyk test as part of CI workflow.
This job runs a Snyk vulnerability check (`snyk test`) on the project's dependencies.

This _should_ replace the current Synk Github integration.

Note:
- Snyk currently complains about https://snyk.io/vuln/npm:hoek:20180212 from Hoek package which is a dependency of a number of our dependencies (e.g bunyan). We can't immediately update most of them to the latest version of Hoek (which fixed the vuln) without incurring breaking changes, so this PR updates the `.snyk` file to ignore this particular package for Snyk's default 30 days. We'll revisit the packages then.

## Tasks to be done after merging:
- [ ] Add `synk-security` as a required check before PR gets merge in GitHub settings
- [ ] Remove existing Snyk Github WebHook check in favor of this job


package.json
```diff
@@ -142,7 +142,7 @@
     "@babel/cli": "7.0.0-beta.38",
     "@babel/core": "7.0.0-beta.38",
     "@babel/node": "^7.0.0-beta.38",
-    "@babel/preset-env": "^7.0.0-beta.38",
+    "@babel/preset-env": "7.0.0-beta.38",
```
impactmass: @aldeed the error below gets thrown in CI test. I found that removing the `^` to ensure preset-env is exactly 7.0.0-beta.38 resolves the issue. Is this a good way to resolve this?
```
[BABEL] /home/circleci/reaction-app/imports/plugins/core/graphql/server/resolvers/account/Query/administrators.test.js:
Requires Babel "^7.0.0-beta.41", but was loaded with "7.0.0-beta.38". 
You'll need to update your @babel/core version. (While processing: "/home/circleci/reaction-app/node_modules/@babel/preset-env/lib/index.js")
```

impactmass:
@ticean, 
reverting the package-lock and re-installing Snyk (as you suggested) fixed the error earlier reported. This is ready now.

PR #4006, opened 2018-03-14
(Fix) Add catalog migrations

zenweasel:
Resolves NA  
Impact: **minor**  
Type: **chore**

## Issue
Users who migrate from `1.8.2` to `1.9.0` would end up with an empty catalog

## Solution
Publish all existing visible products to the catalog

## Testing
1. Start the app on `master` (1.8.2)
1. Stop the app
1. checkout this branch
1. `rm -rf node_modules`
1. `meteor npm install`
1. Start the app
1. Observe that the migration runs and the default product is visible to anonymous users

### DIFF
--- a/imports/plugins/core/versions/server/migrations/24_publish_all_existing_visible_products.js
+++ b/imports/plugins/core/versions/server/migrations/24_publish_all_existing_visible_products.js
@@ -5,7 +5,10 @@ import { publishProductsToCatalog } from "/imports/plugins/core/catalog/server/m
 Migrations.add({
   version: 24,
   up() {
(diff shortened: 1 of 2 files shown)


zenweasel:
@mikemurray I am not seeing that issue adding a product

mikemurray:
@zenweasel I'm seeing the basic example product with no handle. And it seems when the handle is a blank string it causes an issue with `Router.pathFor`. Not sure how the default product ended up with a blank handle.

zenweasel:
@mikemurray Can we tell if this is related to the migration or is an issue on `1.9.0`?

mikemurray:
Tested again and cannot reproduce the issue I commented about.

PR #4032, opened 2018-03-18
Fix #4009: inventory badge component added

### LINKED ISSUE #4009: Backorder flag applied incorrectly on PDP
## Prerequisites

## Issue Description
As an admin, when creating a new product, the backorder badge is applied incorrectly. As seen in this image, the backorder badge is visible on the variant while the quantity is > 0. It seems that the backorder badge is being applied when the `low inventory` badge should be applied as an admin.
image

The backorder badge is also applied incorrectly on the consumer view of the PDP:
image

This seems to happen when viewing a product where the `allow backorder` toggle is enabled and the `Quantity` is <= the number provided in the `Warn at` input on the PDP.

The Admin Product Grid displays the correct badge
image

The new catalog based product grid displays most badges correctly, but when `Allow backorder` is enabled, the `Low inventory` badge never displays. This is incorrect. The Low inventory badge should always show when the quantity is greater than 0, but less than or equal to the `warn at` number.

When the quantity is equal to 0, the badge shown should be `Backorder` if `Allow backorder` is enabled, or `Sold out` if not.

## Steps to Reproduce
1. Create a product
2. Set the `Quantity` to 5
3. Set the `warn at` number to 5
4. Ensure that `Allow backorder` is enabled

## Possible Solution
My best guess is that we're not using the same logic/helper to determine which badge is shown in all spots. We should probably have a single component that calculates which badge is shown so that it's consistent and correct across the app.

### Versions
```
Node: 8.9.0
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.9.0
Reaction branch: master
Docker: 17.12.0-ce
```

Akarshit:
Resolves #4009  
Impact: **major**  
Type: **bugfix**

## Issue
The logic to decide which badge to display was not correct.

## Solution
A new badge component is made that uses the `Product` to decide the badge to idspa

## Testing
1. Create a product
1. Set the Quantity to 5
1. Set the warn at number to 5
1. Ensure that Allow backorder is enabled
1. `Limited Supply` label is shown.
1. Change Quantity to 0
1. `Backorder` label is displayed.
1. Disable Allow backorder.
1. `Sold out` label should be displayed.
1. Set quantity to 10.
1. No label is displayed.

### DIFF
--- a/imports/plugins/core/ui/client/containers/inventoryBadge.js
+++ b/imports/plugins/core/ui/client/containers/inventoryBadge.js
@@ -1,0 +1,32 @@
+import { registerComponent, composeWithTracker } from "@reactioncommerce/reaction-components";
+import { InventoryBadge } from "../components/badge";
+
+const composer = (props, onData) => {
+  const { variant, soldOut } = props;
+  const { inventoryManagement, inventoryPolicy, lowInventoryWarningThreshold, inventoryQuantity } = variant;
+  let label = null;
+  let i18nKeyLabel = null;
+  let status = null;
+  // TODO: update this to use Catalog API.
+  if (inventoryManagement && !inventoryPolicy && Number(inventoryQuantity) === 0) {
+    status = "info";
+    label = "Backorder";
+    i18nKeyLabel = "productDetail.backOrder";
+  } else if (soldOut) {
+    status = "danger";
+    label = "Sold Out!";
+    i18nKeyLabel = "productDetail.soldOut";
+  } else if (inventoryManagement && inventoryPolicy) {
+    if (Number(lowInventoryWarningThreshold) >= Number(inventoryQuantity)) {
+      status = "warning";
+      label = "Limited Supply";
+      i18nKeyLabel = "productDetail.limitedSupply";
+    }
+  }
+
+  onData(null, { ...props, label, i18nKeyLabel, status });
+};
+
+registerComponent("InventoryBadge", InventoryBadge, composeWithTracker(composer));
+
+export default composeWithTracker(composer)(InventoryBadge);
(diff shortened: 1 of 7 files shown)


Akarshit:
Working to see if this can be replaced by the `Catalog` API.

Akarshit:
@zenweasel Ready for review.

imports/plugins/core/ui/client/containers/inventoryBadge.js
```diff
@@ -0,0 +1,32 @@
...
+import { InventoryBadge } from "../components/badge";
+
+const composer = (props, onData) => {
+  const { variant, soldOut } = props;
+  const { inventoryManagement, inventoryPolicy, lowInventoryWarningThreshold, inventoryQuantity } = variant;
+  let label = null;
+  let i18nKeyLabel = null;
+  let status = null;
+  // TODO: update this to use Catalog API.
+  if (inventoryManagement && !inventoryPolicy && Number(inventoryQuantity) === 0) {
```
zenweasel: Why do we have to keep converting these to `Number`? Shouldn't they just be a number? What happens if we can an actual string there?

zenweasel:
If I follow the testing steps but set inventory on the option this does not work

Akarshit:
All the quantities are getting stored as `string` in the revisions, this is all the checks to malfunction.

Solution is to use NumericInput instead of TextField for the quantities. This would be easy to do once #3999 is merged.

imports/plugins/core/ui/client/containers/inventoryBadge.js
Akarshit: Yes, you are correct these should be numbers. The problem is with the components used for getting the quantities. Working to replace them.

zenweasel:
@Akarshit now that #3999 is merged, you are unblocked here?

Akarshit:
Yes.

zenweasel:
I think something doesn't look right

I have Quantity and Warn At set to 5 but I am still seeing "Backorder" instead of "Limited Quantity"

Akarshit:
Updated.

PR #4033, opened 2018-03-19
Reduce Risk of Test False Positives

pmn4:
_(Resolves n/a)_
Impact: **minor**  
Type: **test**

## Issue
I noticed a few things in the test suite that could easily result in false positives.
1) ensure that the `.to.not.throw` expectation runs by adding the `()` (I imagine that `throw` being a reserved word is a problem here... but not sure).
2) remove the parameters from `.to.not.throw`, which is even highlighted in the chai documentation:
> However, it’s dangerous to negate .throw when providing any arguments. The problem is that it creates uncertain expectations by asserting that the target either doesn’t throw an error, or that it throws an error but of a different type than the given type, or that it throws an error of the given type but with a message that doesn’t include the given string. It’s often best to identify the exact output that’s expected, and then write an assertion that only accepts that exact output.

this broke a few tests, so the second commit resolves those failures

## Testing
1. I think it is a bug in the testing framework, but the following test does not fail:
``` JavaScript
it("handles .to.not.throw", function () {
  expect(() => { throw "error"; }).to.not.throw;
});
```
whereas this does fail.
``` JavaScript
it("handles .to.not.throw()", function () {
  expect(() => { throw "error"; }).to.not.throw();
});
```

I wound up re-writing the test groups for `inviteShopMember` and `inviteShopOwner` which were failing after these changes.

### DIFF
--- a/imports/plugins/core/job-collection/lib/jobCollection.app-test.js
+++ b/imports/plugins/core/job-collection/lib/jobCollection.app-test.js
@@ -41,6 +41,8 @@ describe("JobCollection", function () {
   let testColl;
 
   before(function () {
+    this.timeout(10000);
+
     clientTestColl = new JobCollection("ClientTest", { idGeneration: "MONGO" });
     serverTestColl = new JobCollection("ServerTest", { idGeneration: "STRING" });
 
--- a/server/methods/accounts/accounts.app-test.js
+++ b/server/methods/accounts/accounts.app-test.js
@@ -5,23 +5,21 @@ import Random from "@reactioncommerce/random";
 import { Meteor } from "meteor/meteor";
 import { Factory } from "meteor/dburles:factory";
 import { check, Match } from "meteor/check";
-import { Accounts as MeteorAccount } from "meteor/accounts-base";
+import { Accounts as MeteorAccounts } from "meteor/accounts-base";
 import { expect } from "meteor/practicalmeteor:chai";
 import { sinon } from "meteor/practicalmeteor:sinon";
-import { Accounts, Packages, Orders, Products, Shops, Cart } from "/lib/collections";
-import { Reaction } from "/server/api";
+import { SSR } from "meteor/meteorhacks:ssr";
+import { Accounts, Groups, Packages, Orders, Products, Shops, Cart } from "/lib/collections";
+import { Logger, Reaction } from "/server/api";
 import { getShop, getAddress } from "/server/imports/fixtures/shops";
 import Fixtures from "/server/imports/fixtures";
 
 Fixtures();
 
-before(function () {
-  this.timeout(10000);
-});
-
 describe("Account Meteor method ", function () {
   let shopId;
-  const fakeUser = Factory.create("account");
+  let fakeUser;
+  let fakeAccount;
   const originals = {};
   let sandbox;
 
@@ -46,6 +44,16 @@ describe("Account Meteor method ", function () {
   beforeEach(function () {
     shopId = getShop()._id;
     sandbox = sinon.sandbox.create();
+
+    fakeUser = Factory.create("user");
+    const userId = fakeUser._id;
+    // set the _id... some code requires that Account#_id === Account#userId
+    fakeAccount = Factory.create("account", { _id: userId, userId, shopId });
+    sandbox.stub(Meteor, "userId", () => userId);
+    sandbox.stub(Meteor, "user", () => fakeUser);
+    sandbox.stub(Reaction, "getShopId", () => shopId);
+
+    Object.keys(originals).forEach((method) => spyOnMethod(method, userId));
   });
 
   afterEach(function () {
@@ -55,82 +63,77 @@ describe("Account Meteor method ", function () {
   function spyOnMethod(method, id) {
     return sandbox.stub(Meteor.server.method_handlers, `cart/${method}`, function (...args) {
       check(args, [Match.Any]); // to prevent audit_arguments from complaining
-      this.userId = id;
+      this.userId = id; // having to do this makes me think that we should be using Meteor.userId() instead of this.userId in our Meteor methods
       return originals[method].apply(this, args);
     });
   }
 
   describe("addressBookAdd", function () {
     beforeEach(function () {
-      Cart.remove({});
-      Accounts.remove({});
+      const sessionId = Random.id(); // Required for creating a cart
+      // editing your address book also udpates your cart, so be sure there
+      // is a cart present
+      Meteor.call("cart/createCart", fakeAccount.userId, sessionId);
     });
 
-    it("should allow user to add new addresses", function (done) {
-      let account = Factory.create("account");
-      sandbox.stub(Meteor, "userId", function () {
-        return account.userId;
-      });
+    it("should allow user to add new addresses", function () {
       const address = getAddress();
       // we already have one address by default
-      expect(account.profile.addressBook.length).to.equal(1);
+      expect(fakeAccount.profile.addressBook.length).to.equal(1);
+
       Meteor.call("accounts/addressBookAdd", address);
-      account = Accounts.findOne(account._id);
+
+      const account = Accounts.findOne({ _id: fakeAccount._id });
       expect(account.profile.addressBook.length).to.equal(2);
-      return done();
     });
 
-    it("should allow Admin to add new addresses to other users", function (done) {
+    it("should allow Admin to add new addresses to other users", function () {
       sandbox.stub(Reaction, "hasPermission", () => true);
-      let account = Factory.create("account");
       const address = getAddress();
-      expect(account.profile.addressBook.length).to.equal(1);
-      Meteor.call("accounts/addressBookAdd", address, account.userId);
+      expect(fakeAccount.profile.addressBook.length).to.equal(1);
 
-      account = Accounts.findOne(account._id);
+      Meteor.call("accounts/addressBookAdd", address, fakeAccount.userId);
+
+      const account = Accounts.findOne({ _id: fakeAccount._id });
       expect(account.profile.addressBook.length).to.equal(2);
-      return done();
     });
 
-    it("should insert exactly the same address as expected", function (done) {
-      let account = Factory.create("account");
-      sandbox.stub(Meteor, "userId", function () {
-        return account.userId;
-      });
+    it("should insert exactly the same address as expected", function () {
       const address = getAddress();
+
       Meteor.call("accounts/addressBookAdd", address);
-      account = Accounts.findOne(account._id);
+
+      const account = Accounts.findOne({ _id: fakeAccount._id });
       expect(account.profile.addressBook.length).to.equal(2);
       const newAddress = account.profile.addressBook[
         account.profile.addressBook.length - 1];
       delete newAddress._id;
       delete newAddress.failedValidation;
       expect(_.isEqual(address, newAddress)).to.be.true;
-      return done();
     });
 
-    it("should throw error if wrong arguments were passed", function (done) {
+    it("should throw error if wrong arguments were passed", function () {
       const accountSpy = sandbox.spy(Accounts, "update");
 
       expect(function () {
         return Meteor.call("accounts/addressBookAdd", 123456);
-      }).to.throw;
-
-      expect(function () {
-        return Meteor.call("accounts/addressBookAdd", {});
-      }).to.throw;
+      }).to.throw(Error, /must be an object/);
 
       expect(function () {
         return Meteor.call("accounts/addressBookAdd", null);
-      }).to.throw;
+      }).to.throw(Error, /must be an object/);
 
       expect(function () {
         return Meteor.call("accounts/addressBookAdd");
-      }).to.throw;
+      }).to.throw(Error, /must be an object/);
 
       expect(function () {
         return Meteor.call("accounts/addressBookAdd", "asdad", 123);
-      }).to.throw;
+      }).to.throw(Error, /must be an object/);
+
+      expect(function () {
+        return Meteor.call("accounts/addressBookAdd", {});
+      }).to.throw(Error, /Full name is required/);
 
       // https://github.com/aldeed/meteor-simple-schema/issues/522
       expect(function () {
@@ -138,17 +141,12 @@ describe("Account Meteor method ", function () {
           "accounts/addressBookAdd",
           () => { expect(true).to.be.true; }
         );
-      }).to.not.throw;
+      }).to.not.throw();
 
       expect(accountSpy).to.not.have.been.called;
-
-      return done();
     });
 
-    it("should not let non-Admin add address to another user", function (done) {
-      sandbox.stub(Meteor, "userId", function () {
-        return fakeUser._id;
-      });
+    it("should not let non-Admin add address to another user", function () {
       const account2 = Factory.create("account");
       const updateAccountSpy = sandbox.spy(Accounts, "update");
       const upsertAccountSpy = sandbox.spy(Accounts, "upsert");
@@ -157,29 +155,15 @@ describe("Account Meteor method ", function () {
           "accounts/addressBookAdd", getAddress(),
           account2._id
         );
-      }).to.throw();
+      }).to.throw(Meteor.Error, /Access denied/);
       expect(updateAccountSpy).to.not.have.been.called;
       expect(upsertAccountSpy).to.not.have.been.called;
-
-      return done();
     });
 
     it(
       "should disable isShipping/BillingDefault properties inside sibling" +
       " address if we enable them while adding",
-      function (done) {
-        const account = Factory.create("account");
-        sandbox.stub(Meteor, "userId", function () {
-          return account.userId;
-        });
-        sandbox.stub(Reaction, "getShopId", function () {
-          return shopId;
-        });
-        const sessionId = Random.id(); // Required for creating a cart
-        spyOnMethod("setShipmentAddress", account.userId);
-        spyOnMethod("setPaymentAddress", account.userId);
-
-        Meteor.call("cart/createCart", account.userId, sessionId);
+      function () {
         // cart was created without any default addresses, we need to add one
         const address = Object.assign({}, getAddress(), {
           isShippingDefault: true,
@@ -196,18 +180,14 @@ describe("Account Meteor method ", function () {
         Meteor.call("accounts/addressBookAdd", newAddress);
 
         // now we need to get address ids from cart and compare their
-        const cart = Cart.findOne({ userId: account.userId });
+        const cart = Cart.findOne({ userId: fakeAccount.userId });
         expect(cart.shipping[0].address._id).to.equal(newAddress._id);
         expect(cart.billing[0].address._id).to.equal(newAddress._id);
-
-        return done();
       }
     );
   });
 
   describe("addressBookUpdate", function () {
-    // Required for creating a cart
-    const sessionId = Random.id();
     let removeInventoryStub;
 
     before(function () {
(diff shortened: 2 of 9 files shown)


server/methods/accounts/accounts.app-test.js
```diff
@@ -5,11 +5,12 @@ import { Meteor } from "meteor/meteor";
 import { Factory } from "meteor/dburles:factory";
 import { check, Match } from "meteor/check";
 import { Random } from "meteor/random";
-import { Accounts as MeteorAccount } from "meteor/accounts-base";
+import { Accounts as MeteorAccounts } from "meteor/accounts-base";
```
pmn4: to match accounts.js

server/methods/accounts/accounts.app-test.js
```diff
@@ -5,11 +5,12 @@ import { Meteor } from "meteor/meteor";
...
 import { expect } from "meteor/practicalmeteor:chai";
 import { sinon } from "meteor/practicalmeteor:sinon";
-import { Accounts, Packages, Orders, Products, Shops, Cart } from "/lib/collections";
-import { Reaction } from "/server/api";
+import { SSR } from "meteor/meteorhacks:ssr";
```
pmn4: wanted to stub some methods to add additional expectations

server/methods/accounts/accounts.js
```diff
@@ -625,11 +630,13 @@ export function inviteShopOwner(options) {
 
   const emailLogo = Reaction.Email.getShopLogo(primaryShop);
   const token = Random.id();
-  const currentUser = Meteor.users.findOne(this.userId);
+  const currentUser = Meteor.user();
```
pmn4: I replaced this line here and below, which I _think_ is equivalent. I made this change to make the test easier, which I know is not a great reason to modify code, but on the plus side, it saves a db query.

server/methods/accounts/accounts.js
```diff
@@ -695,7 +709,7 @@ export function inviteShopMember(options) {
     throw new Meteor.Error("bad-request", "Cannot directly invite owner");
   }
 
-  const currentUser = Meteor.users.findOne(this.userId);
+  const currentUser = Meteor.user();
```
pmn4: see PR comment above

server/methods/accounts/accounts.app-test.js
```diff
@@ -504,76 +505,179 @@ describe("Account Meteor method ", function () {
...
-      sandbox.stub(Reaction, "hasPermission", () => true);
-      // TODO: Need to udpate this test to properly check the account created
-      // there's currently an error with Media branding asset when trying to do that
-      expect(() =>
-        Meteor.call("accounts/inviteShopMember", {
-          shopId,
-          groupId: Random.id(),
-          email: fakeUser.emails[0].address,
-          name: fakeUser.profile.addressBook[0].fullName
-        })).to.not.throw(Meteor.Error, /Access denied/);
```
pmn4: several of the expectations did not exactly match the test descriptions, so I re-wrote the whole describe block for `inviteShopMember` and `inviteShopOwner`.

server/methods/accounts/accounts.app-test.js
```diff
@@ -22,7 +23,8 @@ before(function () {
 
 describe("Account Meteor method ", function () {
   let shopId;
-  const fakeUser = Factory.create("account");
```
pmn4: this spec had been using User and Account interchangeably. Further, it was creating new accounts here and there, which wind up using the same User (i.e., the userId of all of them was the same). I am fairly certain this resulted in further false-positives

server/methods/accounts/accounts.app-test.js
```diff
@@ -47,6 +49,16 @@ describe("Account Meteor method ", function () {
...
+
+    fakeUser = Factory.create("user");
+    const userId = fakeUser._id;
+    // set the _id... some code requires that Account#_id === Account#userId
+    fakeAccount = Factory.create("account", { _id: userId, userId, shopId });
+    sandbox.stub(Meteor, "userId", () => userId);
+    sandbox.stub(Meteor, "user", () => fakeUser);
+    sandbox.stub(Reaction, "getShopId", () => shopId);
+
+    Object.keys(originals).forEach((method) => spyOnMethod(method, userId));
```
pmn4: no need to do this in each spec. doing it here in the beforeEach works, since we know the current userId

server/methods/accounts/accounts.app-test.js
```diff
@@ -56,100 +68,90 @@ describe("Account Meteor method ", function () {
...
   }
 
   describe("addressBookAdd", function () {
     beforeEach(function () {
-      Cart.remove({});
-      Accounts.remove({});
+      const sessionId = Random.id(); // Required for creating a cart
+      // editing your address book also udpates your cart, so be sure there
+      // is a cart present
+      Meteor.call("cart/createCart", fakeAccount.userId, sessionId);
```
pmn4: many individual specs required a cart, so I moved it up to the before block.

server/methods/accounts/accounts.app-test.js
```diff
@@ -56,100 +68,90 @@ describe("Account Meteor method ", function () {
...
     });
 
-    it("should allow user to add new addresses", function (done) {
-      let account = Factory.create("account");
```
pmn4: previously, we had created the user once at the top of the file (i.e., not in a beforeEach). In doing so, creating an account here made sense. the beforeEach now creates a user and an account to cut down on some of the boiler plate test code

server/methods/accounts/accounts.app-test.js
```diff
@@ -56,100 +68,90 @@ describe("Account Meteor method ", function () {
...
-      expect(account.profile.addressBook.length).to.equal(1);
-      Meteor.call("accounts/addressBookAdd", address, account.userId);
+      expect(fakeAccount.profile.addressBook.length).to.equal(1);
+
+      Meteor.call("accounts/addressBookAdd", address, fakeAccount.userId);
 
-      account = Accounts.findOne(account._id);
+      const account = Accounts.findOne(fakeAccount._id);
       expect(account.profile.addressBook.length).to.equal(2);
-      return done();
```
pmn4: done is only for async tests.

server/methods/accounts/accounts.app-test.js
```diff
@@ -56,100 +68,90 @@ describe("Account Meteor method ", function () {
...
+    it("should allow user to add new addresses", function () {
       const address = getAddress();
       // we already have one address by default
-      expect(account.profile.addressBook.length).to.equal(1);
+      expect(fakeAccount.profile.addressBook.length).to.equal(1);
+
       Meteor.call("accounts/addressBookAdd", address);
-      account = Accounts.findOne(account._id);
+
+      const account = Accounts.findOne(fakeAccount._id);
```
zenweasel: Nitpick, but we are trying to move away from the shorthand `findOne(id)` to the more explicit `findOne({ _id: fakeAccount._id });`

server/methods/accounts/accounts.app-test.js
```diff
@@ -222,133 +206,95 @@ describe("Account Meteor method ", function () {
...
     });
 
     it("should throw error if wrong arguments were passed", function () {
       const updateAccountSpy = sandbox.spy(Accounts, "update");
-      expect(() => Meteor.call("accounts/addressBookUpdate", 123456)).to.throw;
-      expect(() => Meteor.call("accounts/addressBookUpdate", {})).to.throw;
-      expect(() => Meteor.call("accounts/addressBookUpdate", null)).to.throw;
-      expect(() => Meteor.call("accounts/addressBookUpdate")).to.throw;
-      expect(() => Meteor.call("accounts/addressBookUpdate", "asdad", 123)).to.throw;
+      expect(() => Meteor.call("accounts/addressBookUpdate", 123456)).to.throw();
```
zenweasel: @pmn4 Should we make these expectations more explicit while we are here? Doesn't this mask any sort of other error is thrown? I think we have done a better job lately of making these error messages more consistent. I realize that's how the existing code works but since we are changing this, maybe we can make it better?

server/methods/accounts/accounts.app-test.js
```diff
@@ -431,149 +362,250 @@ describe("Account Meteor method ", function () {
...
+      sendEmailSpy = sandbox.stub(Reaction.Email, "send"); // stub instead of spy so we don't actually try to send
+
+      groupId = Random.id();
+      group = Factory.create("group");
+      sandbox.stub(Groups, "findOne", () => group).withArgs({ _id: groupId });
+    });
+
+    it("requires reaction-accounts permission", function () {
+      stubPermissioning({ hasPermission: false });
+      sandbox.stub(Logger, "error") // since we expect this, let's keep the output clean
```
zenweasel: This is clever and we should be doing more of this

server/methods/accounts/accounts.app-test.js
```diff
@@ -431,149 +362,250 @@ describe("Account Meteor method ", function () {
...
 
-    it("should confirm if email already exists before creating", function (done) {
-      sandbox.stub(Reaction, "hasPermission", () => true);
-      expect(() =>
-        Meteor.call("accounts/inviteShopOwner", {
-          email: fakeUser.emails[0].address,
-          name: fakeUser.profile.addressBook[0].fullName
-        })).to.not.throw(Meteor.Error, /Access denied/);
+    it("invites existing users", function () {
+      this.timeout(1000);
```
zenweasel: Isn't this less than the default?

server/methods/accounts/accounts.js
```diff
@@ -356,6 +356,7 @@ export function addressBookAdd(address, accountUserId) {
           "profile.addressBook.$.isShippingDefault": false
         }
       });
+      // should the userId passed to this hook match the userId of the updated Account?
```
zenweasel: These seem like unrelated comments

server/methods/catalog.app-test.js
```diff
@@ -873,7 +873,7 @@ describe("core product methods", function () {
       expect(() => Meteor.call(
         "products/updateProductPosition",
         product._id, position, tag.slug
-      )).to.not.throw(Meteor.Error, /Access Denied/);
+      )).to.not.throw();
```
zenweasel: I'm not sure this is an improvement. Also it seems like the expectation description is incorrect?

server/methods/catalog.app-test.js
```diff
@@ -1081,10 +1081,10 @@ describe("core product methods", function () {
       sandbox.stub(Reaction, "hasPermission", () => true);
       let product = addProduct();
       const { isVisible } = product;
-      expect(() => Meteor.call("products/publishProduct", product._id)).to.not.throw(Meteor.Error, /Access Denied/);
+      expect(() => Meteor.call("products/publishProduct", product._id)).to.not.throw();
```
zenweasel: I'm not sure if I agree that this is an improvement

server/methods/accounts/accounts.app-test.js
pmn4: oops, didn't mean to commit this. the test was failing on Circle CI, but not on local, so I shortened the timeout to mimic

server/methods/catalog.app-test.js
pmn4: the problem with passing arguments to `.not.throw` is that it can mask some evils. For example, if you have `.not.throw(Meteor.Error, /Access Denied/)` but the code has a null pointer error, this test passes.

server/methods/accounts/accounts.js
pmn4: it is definitely unrelated. I noticed it when trying to stub some data. I guess it would have been better as a PR comment than a code comment

server/methods/catalog.app-test.js
zenweasel: Yes, understood, but is that what we are testing? If we want to `expect` that it doesn't throw an error at all maybe that's a different test or at least a separate expectation. Thinking of how to make the test results as useful as possible.

zenweasel: If we get a null pointer error, we have some much bigger problems, like using Java

pmn4: haha, I was brought up on C++ and Java... some terms die hard. plus, doesn't NPE have a nicer ring to it than `"Uncaught TypeError: Cannot read property 'method' of undefined"`?

pmn4: ok, if I rewrite the test to be more of a "happy path" test, will that satisfy the concern?
To me, testing that a particular error doesn't get raised is probably fine at the time the test is written, but any future modifications to the code make the test unreliable as a marker for system stability.
As a hypothetical, I could add as the first line of the method `const a = 1/0;` and this test would pass without the code having even reached the line where the error would have been thrown.

zenweasel: Yeah, I don't want to be too pedantic about it but I don't want to get into this trap where every test tests everything. Hopefully we have a test that tests whether the function does what it is supposed to, and then a test that permissions work as they supposed to. I am fine with the permission test not failing if the method doesn't work because that's going to be give me a clearer picture of what's happening than if I get 10 tests failing with descriptions like "only allows an admin to execute this function". If I don't have a test that just tests that the function does what it's supposed to then that's a separate problem, but the solution would be to write another test that does more than just test that the method doesn't throw an error.

imports/plugins/core/job-collection/lib/jobCollection.app-test.js
```diff
@@ -41,6 +41,8 @@ describe("JobCollection", function () {
   let testColl;
 
   before(function () {
+    this.timeout(10000);
```
pmn4: in `accounts.app-test.js`, there was an unscoped before block which set this timeout. accounts.app-test.js does not need a timeout that long, so I moved it here (where some timeouts were observed).

server/methods/accounts/accounts.app-test.js
```diff
@@ -5,24 +5,21 @@ import { Meteor } from "meteor/meteor";
...
-import { Reaction } from "/server/api";
+import { SSR } from "meteor/meteorhacks:ssr";
+import { Accounts, Groups, Packages, Orders, Products, Shops, Cart } from "/lib/collections";
+import { Logger, Reaction } from "/server/api";
 import { getShop, getAddress } from "/server/imports/fixtures/shops";
 import Fixtures from "/server/imports/fixtures";
 
 Fixtures();
 
-before(function () {
```
pmn4: I believe using the `before` block here sets the timeout globally. that does not seem necessary for most test files.

server/methods/accounts/accounts.app-test.js
```diff
@@ -5,24 +5,21 @@ import { Meteor } from "meteor/meteor";
...
-  this.timeout(10000);
-  Meteor._sleepForMs(7000);
-});
-
 describe("Account Meteor method ", function () {
   let shopId;
-  const fakeUser = Factory.create("account");
+  let fakeUser;
+  let fakeAccount;
```
pmn4: instead of creating a single user for all tests, create a new one for each test. this reduces the chances of one test impacting another

server/methods/accounts/accounts.app-test.js
```diff
@@ -56,100 +63,90 @@ describe("Account Meteor method ", function () {
...
       const accountSpy = sandbox.spy(Accounts, "update");
 
       expect(function () {
         return Meteor.call("accounts/addressBookAdd", 123456);
-      }).to.throw;
-
-      expect(function () {
-        return Meteor.call("accounts/addressBookAdd", {});
-      }).to.throw;
+      }).to.throw(Error, /must be an object/);
```
pmn4: added the check for specific errors. h/t @zenweasel

server/methods/accounts/accounts.app-test.js
```diff
@@ -158,29 +155,15 @@ describe("Account Meteor method ", function () {
           "accounts/addressBookAdd", getAddress(),
           account2._id
         );
-      }).to.throw();
+      }).to.throw(Meteor.Error, /Access denied/);
```
pmn4: added the check for specific errors. h/t @zenweasel

server/methods/accounts/accounts.app-test.js
```diff
@@ -222,133 +201,118 @@ describe("Account Meteor method ", function () {
...
       const updateAccountSpy = sandbox.spy(Accounts, "update");
-      expect(() => Meteor.call("accounts/addressBookUpdate", 123456)).to.throw;
-      expect(() => Meteor.call("accounts/addressBookUpdate", {})).to.throw;
-      expect(() => Meteor.call("accounts/addressBookUpdate", null)).to.throw;
-      expect(() => Meteor.call("accounts/addressBookUpdate")).to.throw;
-      expect(() => Meteor.call("accounts/addressBookUpdate", "asdad", 123)).to.throw;
+
+      expect(() =>
+        Meteor.call("accounts/addressBookUpdate", 123456)
+      ).to.throw(Error, /must be an object/);
```
pmn4: added the check for specific errors. h/t @zenweasel

server/methods/accounts/accounts.app-test.js
```diff
@@ -222,133 +201,118 @@ describe("Account Meteor method ", function () {
...
     it("should not let non-Admin to edit address of another user", function () {
-      const account = Factory.create("account");
       const account2 = Factory.create("account");
-      sandbox.stub(Meteor, "userId", () => account.userId);
       const accountUpdateSpy = sandbox.spy(Accounts, "update");
-      expect(() => Meteor.call("accounts/addressBookUpdate", getAddress(), account2._id)).to.throw;
+
+      expect(() =>
+        Meteor.call("accounts/addressBookUpdate", getAddress(), account2._id)
+      ).to.throw(Meteor.Error, /Access denied/);
```
pmn4: added the check for specific errors. h/t @zenweasel

server/methods/accounts/accounts.app-test.js
```diff
@@ -431,149 +380,264 @@ describe("Account Meteor method ", function () {
   });
 
   describe("addressBookRemove", function () {
+    beforeEach(function () {
+      const sessionId = Random.id(); // Required for creating a cart
+      // editing your address book also udpates your cart, so be sure there
+      // is a cart present
+      Meteor.call("cart/createCart", fakeAccount.userId, sessionId);
```
pmn4: most tests below require a cart, so I moved it here to DRY up the code

server/methods/accounts/accounts.app-test.js
```diff
@@ -431,149 +380,264 @@ describe("Account Meteor method ", function () {
...
       const updateAccountSpy = sandbox.spy(Accounts, "update");
-      expect(() => Meteor.call("accounts/addressBookRemove", 123456)).to.throw;
-      expect(() => Meteor.call("accounts/addressBookRemove", {})).to.throw;
-      expect(() => Meteor.call("accounts/addressBookRemove", null)).to.throw;
-      expect(() => Meteor.call("accounts/addressBookRemove")).to.throw;
-      expect(() => Meteor.call("accounts/addressBookRemove", "asdad", 123)).to.throw;
+
+      expect(() =>
+        Meteor.call("accounts/addressBookRemove", 123456)
+      ).to.throw(Match.Error, /Expected string, got number/);
```
pmn4: added the check for specific errors. h/t @zenweasel

server/methods/catalog.app-test.js
```diff
@@ -827,26 +812,33 @@ describe("core product methods", function () {
...
   describe("updateProductPosition", function () {
+    let product;
+    let tag;
+
     beforeEach(function () {
-      return Tags.remove({});
+      Tags.remove({});
+
+      product = addProduct();
+      tag = Factory.create("tag");
```
pmn4: DRY'ing up the tests below

server/methods/catalog.js
```diff
@@ -1493,7 +1493,7 @@ Meteor.methods({
     }).fetch();
     let variantValidator = true;
 
-    if (typeof product === "object" && product.title.length > 1) {
+    if (typeof product === "object" && product.title && product.title.length > 1) {
```
pmn4: previously, when the test which runs this code ran, we ensured that an Access Denied error was not being thrown. It wasn't (and was therefore passing) because `product.title.length` was throwing a `"Cannot read property 'length' of undefined"` error instead.

server/methods/catalog.app-test.js
```diff
@@ -1114,7 +1095,11 @@ describe("core product methods", function () {
         selector: { type: "simple" },
         validate: false
       });
-      expect(() => Meteor.call("products/publishProduct", product._id)).to.not.throw(Meteor.Error, /Access Denied/);
+
+      expect(() =>
+        Meteor.call("products/publishProduct", product._id)
+      ).to.throw(Meteor.Error, /Bad Request/);
```
pmn4: see comment below in `catalog.js`. further, the description suggests this test is more about validation than permission

server/methods/accounts/accounts.js
pmn4: but is is a bug, right?
this method allows for a userId to be passed in, but several of the operations take place on the _current_ user.

pmn4:
@zenweasel I think I have addressed all of your comments. let me know if there is anything further you'd like

zenweasel:
@pmn4 Hey Pat, I've been on vacation for the last week and a half. Will take a look at this again soon.

zenweasel:
@pmn4 This looks good. Thanks for the changes, these are all nice improvements. Just have one conflict left.

zenweasel:
@pmn4 Looks like we have failing tests and some conflicts to resolve

PR #4035, opened 2018-03-19
3994 nnnnat remove media hooks

### LINKED ISSUE #3992: Product media deletion does not respect revision control
## Prerequisites

## Issue Description
With revisions enabled, deleting a product image deletes it permanently and skips the revision control.

## Steps to Reproduce

As an admin

1. Enable revisions
1. Edit a product and add some images.
1. Publish
1. Delete one image but don't publish.

When you view the product as another user, you won't see the deleted image, even though the deletion hasn't yet been published.

## Possible Solution

This is a long standing issue, but did not have an issue already. I noticed this when converting away from CollectionFS, but it was not deemed necessary to fix as part of that. The file-collections package that we now use automatically deletes the backing data when you call `Media.remove`, so we'll need to add some way to skip that, and then some way to manually do it later, upon publication.

### Versions
Node: 8.9.1
NPM: 5.7.1
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.28.0
Reaction: 1.9.0
Docker: 17.12.0-ce

### LINKED ISSUE #3994: Remove Media hooks
Search for `MediaRecords.before.` in `/imports/plugins/core/revisions/server/hooks.js` and remove those hooks, but keep the revision control that they implement.

## For inserts

The only `Media.insert` that needs revision control is in /imports/plugins/core/ui/client/containers/mediaGallery.js

Move that insert to a Meteor method, and move the `MediaRecords.before.insert` logic into that method.

## For updates

Make an "updateMediaMetadata" method to replace `MediaRecords.before.update`.

Needs to be called in one place: /imports/plugins/core/files/server/methods.js in the `sortedMediaIDs.forEach`

## For removes

Make a "removeMedia" method to replace `MediaRecords.before.remove`.

Call "removeMedia" method instead of `Media.remove` directly in /imports/plugins/core/ui/client/containers/mediaGallery.js

Should be able to fix #3992 while doing this.

nnnnat:
Resolves #3994  
Impact: **breaking**  
Type: **refactor**

## Issue
This is part of removing the collection-hooks package. 

## Solution
* Created two new Meteor methods for inserting and removing media and moved/refactored the old `media.before.insert` and `media.before.remove` code to be part of these new methods.
* Created a `updateMediaMetadata` helper function with most of the logic from the old `media.before.update` hook that is called from the Meteor `media/updateProperties` method.
* Removed the old revision media hooks
* Fixed the ProductRevisions publication by updating its Revision.find to look for the correct document props, imported this publication into the server/index.js file.
* Fixed the revisions `handleImageRevisions` helper function to set the correct workflow status to the image's metadata.
* Updated the PublishContainer to use the correct ProductRevisions publication.
* Updated the MediaGallery component to use the new Meteor methods.

## Breaking changes
`media.before.insert` `media.before.update` `media.before.remove` hooks have all been removed.

## Testing
*Test this with two browser open, one as an admin another as a customer to verify published changes make it to the customer catelog*
*Need a Mongo DB viewer to check data values*

**Uploading Product Images**
1. Spin up a fresh RC shop and login as an admin.
2. Select the Basic Reaction Product and upload a product image.
3. The Image should upload and have a green unpublished indicator in the top right corner. The customer view should see no image.
4. As an admin click the "publish" button, the green indicator should disapear and the image should now display for the customer.

**Reordering Product Images**
1. As an admin upload a few more images then click "publish"
2. As an admin select one of the newly uploaded images and drag to reorder them.
3. Green unpublished indicators should appear.
4. Click "publish" and changes should also happen for the customer view.

**Deleting Product Images**
1. As an Admin upload a new image to the product but do not publish.
2. Click the X on the newly uploaded image and follow the promps to remove the image.
3. The image should be removed from the product.
4. Now click the X for one of the older images and follow the promps to remove.
5. The image should now have the red "archived" indicator in the top right corner.
6. Now click the "publish" button the "archived" image should be removed form both the admin and customer views.

## Notes
This PR should also fix #3992. If you remove an unpublished images from a product the image will be deleted, If you remove a published image from a product the image will be archived.

Uploading Brand asset from the shop admin panel skips revisions.


willopez:
I also see the upload error shown in the image below. The uploads work correctly, in both chrome and firefox and I only see that error in chrome.

screen shot 2018-03-20 at 6 29 06 pm

nnnnat:
I've been having trouble reproducing the image not deleting for the customer. We've created an issue to track that problem because it might be an edge case and the issue might be coming from the file uploading lib and not any of the media hook replacement.
#4051

PR #4047, opened 2018-03-20
Fix for "Browser console shows error when replacing Tag schema"

### LINKED ISSUE #4046: Browser console shows error when replacing Tag schema
## Prerequisites

## Issue Description
It's not possible to replace the `Tag` schema without browser console errors shown up in console.

## Steps to Reproduce
1. Put this code snippet somewhere in a plugin:
```
import { Tag } from "/lib/collections/schemas";
const TagSchema = Tag.clone().extend({
  newProp: {
    type: String,
    optional: true
  }
});
registerSchema("Tag", TagSchema);
```
2. modify at least one `Tag` document in the mongo db to have a property called `newProp`
3. Navigate to landing page with console open
4. Observe the following error message:
```
Failed prop type: Objects in array must be of type: Schemas.Tag
    in Reaction(TagList) (created by TagListContainer)
    in TagListContainer (created by Tracker(TagListContainer))
```

### Versions
Node: 8.2.1
NPM: 5.3.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.26.4
Reaction: 1.9.0
Reaction branch: master
Docker: 17.03.1-ce

prinzdezibel:
Resolves #4046  
Impact: **minor**  
Type: **bugfix**

## Issue
It's not possible to replace the Tag schema without browser console errors shown up in console.

## Solution / Changes
- Instead of including the Schema directly, the code refers to the Tags schema through the schema registry via getSchemas()
- Don't create validation context instance once during application bootstrap. Instead the context instantiation is defered, until the schema is validated. This ensures that the latest, registered schema is used for validation.

## Breaking changes
None

## Testing
1. Put this code snippet somewhere in a plugin:
```
import { Tag } from "/lib/collections/schemas";
const TagSchema = Tag.clone().extend({
  newProp: {
    type: String,
    optional: true
  }
});
registerSchema("Tag", TagSchema);
```
2. modify at least one `Tag` document in the mongo db to have a property called `newProp`
3. Navigate to landing page with console open. There should be no error message.

### DIFF
--- a/lib/api/prop-types.js
+++ b/lib/api/prop-types.js
@@ -1,15 +1,14 @@
 import _ from "lodash";
 import { check } from "meteor/check";
-import * as Schemas from "/lib/collections/schemas";
+import { getSchemas } from "@reactioncommerce/reaction-collections";
+
+const Schemas = getSchemas();
 
 /**
  * @file **PropTypes** - React Component PropTypes methods for validating Tags
  *
  * @namespace PropTypes
  */
-
-const TagSchema = Schemas.Tag.newContext();
-
 export const PropTypes = {};
 
 /**
@@ -24,7 +23,7 @@ export const PropTypes = {};
 PropTypes.Tag = (props, propName) => {
   check(props, Object);
   check(propName, String);
-
+  const TagSchema = Schemas.Tag.newContext();
   if (_.isEmpty(props[propName]) === false) {
     if (TagSchema.validate(props[propName]) === false) {
       return new Error("Tag must be of type: Schemas.Tag");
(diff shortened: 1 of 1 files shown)


prinzdezibel:
@aldeed I'd like to get a review from you because not sure if the context creation for simple schema is at the right place (could also be in Meteor.startup callback or similar).

aldeed:
@prinzdezibel We updated the docs to say that you should extend schemas directly, basically what you are doing but without `.clone()` first, and without needing to register.

```
import { Tag } from "/lib/collections/schemas";
const TagSchema = Tag.extend({
  newProp: {
    type: String,
    optional: true
  }
});
```

Is that what you were doing that caused the errors in console? And what do the errors say?

prinzdezibel:
@aldeed Thanks Eric, that's actually a much nicer way of doing it. It also felt a bit awkward to call the same function called `registerSchema` twice, instead of using something more explicit like `replaceSchema` But extending it in-place is even better. Going to close.

PR #4048, opened 2018-03-20
GraphQL connections and devserver app

### LINKED ISSUE #3903: Implement MongoDB Cursors for use in GraphQL Query Connections
**Summary**
Based on @ticean's exploratory work on our GraphQL schemas, we'll need to build an implementation to inject MongoDB cursors into our Query Connections. For this ticket, we're specifically focused on the `AccountConnection` and the `GroupConnection`. As seen in the Github v4 API, we'll need to implement these within the [Edges](https://developer.github.com/v4/object/commitcommentedge/) and in [PageInfo](https://developer.github.com/v4/object/pageinfo/)

This ticket involves determining how to do this with Mongo's cursor and building an abstraction that will permit us to do it reliably for a variety of Collections and Queries.

**Resources**
- [GraphQL pagination docs](http://graphql.org/learn/pagination/)
- [github GraphQL user connection](https://developer.github.com/v4/object/userconnection/) #

### LINKED ISSUE #3905: Determine if we need a special, root-level item for paginated results
Depending on what gets decided in our issue for choosing a state management system (#3904), we should determine if we need a special, root-level item for paginated results.

There's a good discussion of this issue on the facebook relay repo:  https://github.com/facebook/relay/issues/112

## Acceptance Criteria
The solution will identify if we need such an item, and make a case for why or why-not. If the decision is that we _do_ need an item such as this, this solution will involve creating a new issue with all relevant information needed to start work on the root-level item, or, if the solution is simple, creating a PR to resolve the issue.

### LINKED ISSUE #3909: Design pagination parameters for GraphQL API queries
**Summary**
Work with @ticean to design pagination parameters for GraphQL queries that can be paginated.

## Acceptance Criteria
We should have a documented pattern for implementing pagination in our GraphQL queries.

### LINKED ISSUE #3913: Create administrators GraphQL query
**Summary**
Create an `administrators` query in GraphQL.

**Implementation**
The `administrators` query should have the following signature
```
administrators(shopId: ID!): AccountConnection
```
It has one argument, `shopId` which is required, and returns an `AccountConnection`. If a `shopId` is provided, the query should return all administrator accounts associated with that `shopId`.

## Acceptance Criteria
Using GraphQL, I can use the `administrators(shopId: ID!)` query to get an `AccountConnection` by shopId.

aldeed:
Resolves #3903
Resolves #3905
Resolves #3909
Resolves #3913   
Impact: **minor**  
Type: **feature**

## Changes
- There is a now a pure Node development server that can be used to run our GraphQL and GraphiQL endpoints without involving Meteor. It uses mocks whenever it would otherwise interact with Meteor code, so all development should be tested within the full Reaction Meteor app eventually. But the dev server provides fast startup and restart for the initial phase of developing GraphQL queries and mutations. Usage: `npm run devserver`
- All core development for connections, pagination, and sorting is done. Refer to `administrators` query as a reference implementation.
- Added two `administrators` resolvers. It's unclear right now whether it's best to have `administrators` be a root query or live within a `Shop`. For now, it's in both places, and this is fine because most code is shared.

## Breaking changes
None. GraphQL endpoint is officially not yet released.

## Testing Dev Server

```
meteor npm install
npm run devserver
```

- Verify that the server starts up, and that you can access GraphiQL at the URL it shows and make queries, which will receive mocked data. There is no real authorization on this mock server.
- While the devserver is running, change some file in the graphql plugin folder, and verify that the server sees this change and rebuilds + restarts.

## Testing administrators query

Do these steps with both root `administrators` and an `administrators` within a `shop` query response.

1. Add a bunch of admin users for a single shop in the database
2. Test basic `administrators(shopId)`, asking for all possible data back.
    - You should get `edges` and `nodes` back, with correct account data in them. The only difference is that `edges` have an extra object wrapping them, which tells you the `cursor` for each node.
    - You should get `pageInfo` back, with correct paging info.
    - It should be limited by default to 50 nodes. 
3. Test with combinations of `first`, `last`, `before`, `after` and verify that the results are what you expect. Refer to [the spec](https://facebook.github.io/relay/graphql/connections.htm)
4. Test `orderBy` and `sortOrder` params. They are both enums, so test every possible value. Verify that the defaults are "createdBy" and "asc".


aldeed:
@spencern This should be good to go now.

PR #4049, opened 2018-03-21
(fix) Wait for slugify to show up before returning slug

### LINKED ISSUE #4011: Imported product sometimes has blank handle
## Issue Description
Sometimes when the Basic Reaction Product is imported on startup it gets imported without a handle which causes the grid to crash with this error:
```
TypeError: Expected "handle" to match "[^\/]+?", but got ""
```

## Steps to Reproduce
1. Start the app with `reaction-devtools` installed
1. Do "reset data" and load small data-set (everything will be fine)
1. Stop the app and start it again so that "Basic Reaction Product" gets imported
1. Refresh
1. Observe that the grid does not load
1. If you look in Mongo observe that the handle for the "Basic Reaction Product" is blank

### Versions
```
Node: 8.9.4
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Docker: 17.12.0-ce
```

zenweasel:
Resolves #4011 
Impact: **major**  
Type: **bugfix**

## Issue
We weren't waiting for slugify to be resolve and if it wasn't resolved we just returned an empty string, which completely crashes the app

## Solution
`await` rather than `resolve`. I also removed the check for `slugify` that just bailed and returned an empty string because I can't see why we would want that.

## Testing
1. Fully load the app with built in data and `devtools` loaded
1. Choose "Reset Data"
1. Stop the app
1. Restart it
1. Observe that the data loads properly and the app doesn't crash

### DIFF
--- a/lib/api/helpers.js
+++ b/lib/api/helpers.js
@@ -199,8 +199,8 @@ export function getCurrentTag() {
  */
 export function getSlug(slugString) {
   let slug;
-  Promise.resolve(lazyLoadSlugify());
-  if (slugString && slugify) {
+  Promise.await(lazyLoadSlugify());
+  if (slugString) {
     slug = slugify(slugString.toLowerCase());
   } else {
     slug = "";


jshimko:
This definitely appears to be causing failing tests on both CircleCI and local.

I'm also noticing that `getSlug` is duplicated in 3 places (client, lib, and server).  Do we know if that's necessary?  Seems like it shouldn't be since the code is essentially identical in all 3 places.  

**client**
https://github.com/reactioncommerce/reaction/blob/a3b5cb5b7581a1659cb446b564f7bd5341ab5cae/client/modules/core/helpers/utils.js

**lib**
<https://github.com/reactioncommerce/reaction/blob/a3b5cb5b7581a1659cb446b564f7bd5341ab5cae/lib/api/helpers.js#L14-L34>
<https://github.com/reactioncommerce/reaction/blob/a3b5cb5b7581a1659cb446b564f7bd5341ab5cae/lib/api/helpers.js#L191-L209>

**server**
https://github.com/reactioncommerce/reaction/blob/a3b5cb5b7581a1659cb446b564f7bd5341ab5cae/server/api/core/utils.js

zenweasel:
I don't know about the other places, but the place it's getting called here is from a schema so it seems like it needs to be in `lib` for that purpose since this code might get called from either client or server. Additionally, maybe I don't understand something but is there a purpose for lazy-loading a library on the server-side. Could we be wrapping these in `isServer` blocks and just load it at start time?

zenweasel:
Tests fail if you remove the code that just swallows the error so...

zenweasel:
@jshimko Take a look at this and tell me if you think this makes sense

PR #4054, opened 2018-03-22
Fix #3518: AddressBook to react

### LINKED ISSUE #3518: Avalara does not catch address validation when API throws an error because of invalid address

Akarshit:
Resolves #3518   
Impact: **major**  
Type: **feature|performance**

## Issue
The Address review workflow had to be redesigned.

## Solution
Whenever the client enters address that can't be validated by Avalara, show client a warning and give a chance to correct the address.
Also did performance optimizations for preventing hitting the Avalara API many times and prevented re-rendering of the AddressBook, because of fake updates to the Cart.

## Test without tax
1. Login as admin
1. But an item and go to checkout screen
1. Try adding multiple addresses, changing default shipping and billing address
1. Try editing already added address
1. Everything should work.

## Test addressBook in profile
1. Login and go to profile section.
1. Try repeating above steps.

## Test with Avalara
1. Enable Avalara
1. Enable address validation
1. Enable tax calculation.
1. Add an item to cart and begin to checkout
1. For address enter RC office address
1. Observe it takes you to the review screen.
1. Click on `Edit entered address`, it should take you back to the edit address screen.
1. Again save the address with same mistakes.
1. It should take you to the review screen again.
1. Select the Suggested Address(should be selected by default)
1. Click on `Continue with selected address`.
1. It should save that address, and tax should be calulated.
1. Again repeat the above process, but on the review screen select the `Entered Address`
1. It should save the address but tax won't be calculated.
1. After placing the order, login as admin(if you weren't).
1. Go to the Orders from Action Menu, click on the above order, you should see it labelled `Tax not collected`.


aldeed:
Changed the base on this for now because Nat's PR isn't merged into 1.10.0 yet

aldeed:
@Akarshit There are a few new commits in `refactor-2053-nnnnat-address-book-react` branch you'll have to pull in.

Akarshit:
@aldeed Yes did that.

Akarshit:
@rymorgan When tax is not calculated for an order, we show a badge saying "Tax not calculated" in the order summary for the admin to see.
The text is too long, so I thought maybe you would have a better design.
Also you once suggested to add `High Risk` tag for this, but "tax calculation failed" doesn't make a order high risk, so it might be better to show admin something more appropriate than `High Risk`.

rymorgan:
@Akarshit - Just to be clear I never suggested showing "high risk," I suggested that we use the same badge component/style to show that Tax had not been calculated. I was showing that high risk as an example of how we should handle it visually. Also, see below, I think this will fit fine. Unless there's something about this that I missing. Please note, there's also a dot notification on the order in the sortable table. 

orders - high risk flag copy

rymorgan:
In generally the UX here is a big improvement but we're missing some styling. See Cassy's style here for details on how it should look here: 
https://github.com/reactioncommerce/reaction/issues/3518#issuecomment-374366833

The text on the warning inline alert should be: @rui-warning-text and may need be corrected in the actual inline alert component as I've seen this multiple times. 

See attached for missing/wrong styles: 
avalara validation

Akarshit:
@rymorgan Oh, then I would have misunderstood you. I will implement the correct styling.

Akarshit:
@rymorgan I was trying to find the styles for these elements but was unable to do so.
I found `@rui-warning-text` and `@rui-warning-bg` but am not sure how to style the "inline red alert". Is there some ready-made class that I am supposed to use?

kieckhafer:
@Akarshit to get the red alerts, replace `warning` with `danger` in all instances.

rymorgan:
for the entered address validation alert it should be these styles: 
danger-alert

imports/plugins/core/accounts/client/containers/addressBookContainer.js
```diff
@@ -0,0 +1,223 @@
...
+  onData(null, {
+    addressBook,
+    countries,
+    heading: this.heading,
+    initMode,
+    regionsByCountry
+  });
+}
+
+registerComponent("AddressBookContainer", AddressBook, [
```
jshimko: This should be named `AddressBook` (without the `Container` in the name).  That decision was originally made because it's potentially a little confusing when you do something like...

```js
replaceComponent("AddressBookContainer", CustomAddressBook);
```
That isn't actually replacing the container.  It's replacing the component and inheriting the same container.  So anything that gets registered and added to `Components` with `registerComponent` should have the name of the UI component.

imports/plugins/core/accounts/client/templates/profile/profile.js
```diff
@@ -114,6 +114,18 @@ Template.accountProfile.helpers({
...
 
+  /**
+   * @method AddressBook
+   * @summary returns a component for updating a user's address.
+   * @since 2.0.0
+   * @return {Object} - contains the component for updating a user's address.
+   */
+  AddressBook() {
+    return {
+      component: Components.AddressBookContainer
```
jshimko: See other comment about component name not having `Container` in it.

imports/plugins/core/accounts/client/containers/addressBookContainer.js
```diff
@@ -0,0 +1,223 @@
...
+  const cart = Collections.Cart.findOne();
+  // If we have passed the address step, show the grid
+  if (cart.workflow.status !== "checkoutAddressBook") {
+    initMode = "grid";
+  }
+  // AddressBook heading will be different in different views
+  // If the view template that's using the AddressBook has a
+  // heading object, set it as the AddressBook heading
+  if (!this.heading) {
+    const template = Template.instance();
```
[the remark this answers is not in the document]
Akarshit: We pass down the data from the Blaze template. That's all I know at-least.

imports/plugins/core/accounts/client/templates/addressBook/add/add.js
```diff
@@ -1,126 +0,0 @@
...
-import { ReactiveVar } from "meteor/reactive-var";
-import { AutoForm } from "meteor/aldeed:autoform";
-import { Template } from "meteor/templating";
-import { Reaction, i18next } from "/client/api";
-import * as Collections from "/lib/collections";
-
-Template.addressBookAdd.onCreated(function () {
-  this.currentCountry = new ReactiveVar(null);
-  // hit Reaction's GeoIP server and try to determine the user's country
-  HTTP.get("https://geo.getreaction.io/json/", (err, res) => {
```
jshimko: This HTTP call was hitting our GeoIP server to get the country the user is currently in and automatically set their country in the select field.  It looks like this got ditched completely.  Do we have a suitable alternative to maintain that UX?

Akarshit:
Working to incorporate all these changes.

imports/plugins/core/accounts/client/templates/addressBook/add/add.js
Akarshit: @jshimko No there wasn't. Now I have added it.

Akarshit:
@jshimko updated the PR with the fixes.

PR #4063, opened 2018-03-26
Fix for "TaxCloud service not calculating tax"

### LINKED ISSUE #3973: TaxCloud service not calculating tax
## Issue Description
When enabled, the TaxCloud service is not calculating tax in the cart. There are no errors on the client or server console and I am using the current API key

## Steps to Reproduce
1. Enable TaxCloud tax service
1. Add an item to cart and enter address
1. Observe that tax is not calculated

### Versions
```
Node: 8.9.4
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.28.0
Reaction: 1.9.0
Docker: 17.12.0-ce
```

prinzdezibel:
Resolves #3973  
Impact: **major**  
Type: **bugfix**

## Issue
When enabled, the TaxCloud service is not calculating tax in the cart. There are no errors on the client or server console and I am using the current API key

## Solution/Changes
- Use Reaction.getPackageSetting to retrieve the package settings
- Don't check the state of the pkgSettings.enabled which formerly
  indicated if a plugin is "installed" (e.g. enabled is a misnomer).
  If so, it would have been displayed in the dashboard. This seems no
  longer be true for any of the plugins, i.e. they appear in dashboard
  regardlesss of the state of that boolean. Also it will be obsolete
  anyway as soon as we're transitioning to plugins as NPM packages.
  Because then the plugin is installed not through toggling a checkbox,
  but rather through a dedicated `npm install`. I'm going to create a separate
  issue for this TODO.

## Notes

The diff that is being displayed is pretty weird, essentially only the branch condition was changed, the other changes results through indentation only.

## Breaking changes
None expected

## Testing
1. As an admin, enable TaxCloud tax service
2. As an user, add an item to cart and enter address during checkout
3. Observe that tax is calculated

### DIFF
--- a/imports/plugins/included/taxes-taxcloud/server/hooks/hooks.js
+++ b/imports/plugins/included/taxes-taxcloud/server/hooks/hooks.js
@@ -90,11 +90,15 @@ MethodHooks.after("taxes/calculate", (options) => {
           };
 
           try {
-            let response = HTTP.post(url, request);
+            const response = HTTP.post(url, request);
             let taxRate = 0;
             // ResponseType 3 is a successful call.
             if (response.data.ResponseType !== 3) {
-              throw new Error("Error calling taxcloud API", JSON.stringify(response.data));
+              let errMsg = "Unable to access service. Check credentials.";
+              if (response && response.data.Messages[0].Message) {
+                errMsg = response.data.Messages[0].Message;
+              }
+              throw new Error("Error calling taxcloud API", errMsg);
             }
 
             let totalTax = 0;
@@ -111,11 +115,7 @@ MethodHooks.after("taxes/calculate", (options) => {
             const taxes = []; // only populated for Avalara
             Meteor.call("taxes/setRate", cartId, taxRate, taxes);
           } catch (error) {
-            let errMsg = "Unable to access service. Check credentials.";
-            if (response && response.data.Messages[0].Message) {
-              errMsg = response.data.Messages[0].Message;
-            }
-            Logger.warn("Error fetching tax rate from TaxCloud:", errMsg);
+            Logger.warn("Error fetching tax rate from TaxCloud:", error.message);
           }
         }
       }
--- a/imports/plugins/included/taxes-taxcloud/server/hooks/hooks.js
+++ b/imports/plugins/included/taxes-taxcloud/server/hooks/hooks.js
@@ -89,30 +89,34 @@ MethodHooks.after("taxes/calculate", (options) => {
             }
           };
 
-          HTTP.post(url, request, (error, response) => {
+          try {
+            let response = HTTP.post(url, request);
             let taxRate = 0;
             // ResponseType 3 is a successful call.
-            if (!error && response.data.ResponseType === 3) {
-              let totalTax = 0;
-              for (const item of response.data.CartItemsResponse) {
-                totalTax += item.TaxAmount;
-              }
-              // don't run this calculation if there isn't tax.
-              if (totalTax > 0) {
-                taxRate = (totalTax / cartToCalc.getSubTotal());
-              }
-              // we should consider if we want percentage and dollar
-              // as this is assuming that subTotal actually contains everything
-              // taxable
-              Meteor.call("taxes/setRate", cartId, taxRate, response.CartItemsResponse);
-            } else {
-              let errMsg = "Unable to access service. Check credentials.";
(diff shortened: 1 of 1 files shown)


prinzdezibel:
@zenweasel Done. The taxcloud API call was async. Converting to sync did the trick.

imports/plugins/included/taxes-taxcloud/server/hooks/hooks.js
```diff
@@ -42,84 +37,85 @@ MethodHooks.after("taxes/calculate", (options) => {
...
-        if (!apiKey || !apiLoginId) {
-          Logger.warn("TaxCloud API Key is required.");
-        }
-        if (Array.isArray(cartToCalc.shipping) && cartToCalc.shipping.length > 0 && cartToCalc.items) {
-          const shippingAddress = cartToCalc.shipping[0].address;
+      if (!apiKey || !apiLoginId) {
+        Logger.warn("TaxCloud API Key is required.");
+      }
+      if (Array.isArray(cartToCalc.shipping) && cartToCalc.shipping.length > 0 && cartToCalc.items) {
+        const shippingAddress = cartToCalc.shipping[0].address;
```
zenweasel: We need to not do this, but this should probably be addressed in #4089

imports/plugins/included/taxes-taxcloud/server/hooks/hooks.js
```diff
@@ -42,84 +37,85 @@ MethodHooks.after("taxes/calculate", (options) => {
...
-              }
+          // format cart items to TaxCloud structure
+          let index = 0;
+          for (const items of cartToCalc.items) {
+            // only processs taxable products
+            if (items.variants.taxable === true) {
+              const item = {
+                Index: index,
+                ItemID: items.variants._id,
+                TIC: "00000",
```
zenweasel: If this is the tax code, should we be pulling this from the product rather than hard-coding it here? I understand this is existing code but it seems like a simple fix.

imports/plugins/included/taxes-taxcloud/server/hooks/hooks.js
```diff
@@ -42,84 +37,85 @@ MethodHooks.after("taxes/calculate", (options) => {
...
-                // taxable
-                Meteor.call("taxes/setRate", cartId, taxRate, response.CartItemsResponse);
-              } else {
-                let errMsg = "Unable to access service. Check credentials.";
-                if (response && response.data.Messages[0].Message) {
-                  errMsg = response.data.Messages[0].Message;
-                }
-                Logger.warn("Error fetching tax rate from TaxCloud:", errMsg);
+          try {
+            const response = HTTP.post(url, request);
```
zenweasel: Is there another solution rather than making this sync? I feel like I can see a noticeable sluggishness. Since we already have it calculated on the cart I don't quite understand why we need block to calculate this again to get it to show up on the completed order.

prinzdezibel: As you correctly stated, the tax has already been calculated before (through running this very method). The thing to understand here is that this method is always running in conjunction with the method "taxes/calculate" in /imports/plugins/core/taxes/server/methods/methods.js.

So the excecution order is always the following:

The core method "taxes/calculate" sets the tax to 0.00 in Line 245.  After that this method is executed, which overwrites the previously calculated value (0.00) to the value calculated from taxcloud.

If this is method is not synchronous, the 0.00 value from core will be in place when the copyCartToOrder method runs, because the taxcloud value has not arrived yet.

Therefore it's made synchronous (as it is in avalara as well)

imports/plugins/included/taxes-taxcloud/server/hooks/hooks.js
prinzdezibel: Any actions to take here?

zenweasel: No, nothing to do for now I suppose

prinzdezibel:
Closing this ticket in favor of #4089. You may have a look into /imports/plugins/core/taxes/server/methods/methods.js, Line 245, @Akarshit. Perhaps it's possible to avoid to have that `Meteor.call("taxes/setRate", cartToCalc._id, cartTaxRate);` call there, which in turn would allow to avoid having that taxcloud API call to be synchronous.

PR #4073, opened 2018-03-27
Fix 3978 Product visibility when revisions are turned off

### LINKED ISSUE #3978: Can't make product visible when revisions are turned off
## Issue Description
This might be user error, but if there is a way to make the product visible when Revisions are turned off I can't find it.

## Steps to Reproduce
1. Login as admin
1. Go to Catalog dashboard panel and disable Revisions
1. Add a product
1. Observe the visibility button is not available in the header
1. Observe that `isVisible` is set to `false` in the db so the product is not shown in the product grid

### Versions
```
Node: 8.9.4
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.28.0
Reaction: 1.9.0
Docker: 17.12.0-ce
```

prinzdezibel:
Resolves #3978   
Impact: **major**  
Type: **bugfix**

## Issue
If Product Revisioning is turned off, one can't publish products to the catalog.sal.

## Solution
Publishing & Archiving products should also work when Products Revisions are    disabled.

## Changes:
- Show visibility and publish button, even if Catalog Revisions are turned off.
- Remove unused function `renderViewControls`
- If revision control is disabled, the archive action does set property isDeleted directly on the selected variants (instead of writing to revisions as intermediary)
- Because product variants with `isDeleted === true` are not available any longer in admin view, the Publish button for that product will no longer be available (the product vanishes immediately). Therefore the code does publish the changes to the catalog immediately through calling `publishProductsToCatalog`

## Breaking changes
None expected

## Testing

Test 1)
1. As an admin, turn-off Revision control
2. Navigate to the product grid and change visibility of example product
3. Hit publish button
4. Product should be invisible/visible for non-admin users.

Test 2)
1. As an admin, turn-off Revision control.
2. Navigate to the PDP of the example product and change values there
3. You should be able to publish changes through the "Publish" button
4. HOWEVER, because the PDP does not retrieve it's data from the Catalog, all changes are displayed immediately for non-admin users (changes are not held in revision collection, but saved directly in `Products` collection). This is a bug which will be addressed once the PDP is also retrieving it's data from catalog. I'm creating a new ticket for that:  PLACEHOLDER NEW TICKET

### DIFF
--- a/server/methods/catalog.js
+++ b/server/methods/catalog.js
@@ -979,7 +979,8 @@ Meteor.methods({
 
     if (!RevisionApi.isRevisionControlEnabled()) {
       // Publish products that have been marked as deleted immediately.
-      publishProductsToCatalog(ids);
+      const products = productsWithVariants.filter((doc) => doc.type === "simple").map((doc) => doc._id);
+      publishProductsToCatalog(products);
     }
 
     Logger.debug(`${numFlaggedAsDeleted} products have been flagged as deleted`);
@@ -1042,7 +1043,6 @@ Meteor.methods({
       update = EJSON.parse(`{"${field}":${stringValue}}`);
     }
 
-
     // we need to use sync mode here, to return correct error and result to UI
     let result;
 
@@ -1067,6 +1067,10 @@ Meteor.methods({
     // meaning the update went past revision control,
     // denormalize and attach results to top-level product
     if (result === 1) {
+      if (type === "simple" && !RevisionApi.isRevisionControlEnabled()) {
+        // Publish product changes immediately.
+        publishProductsToCatalog(_id);
+      }
       if (type === "variant" && toDenormalize.indexOf(field) >= 0) {
         denormalize(doc.ancestors[0], field);
       }
--- a/server/methods/catalog.js
+++ b/server/methods/catalog.js
@@ -940,7 +940,7 @@ Meteor.methods({
         "documentData.isDeleted": true
       }).count();
     } else {
-      //eslint-disable-next-line no-shadow
+      // eslint-disable-next-line no-shadow
       productsWithVariants.forEach((product) => {
         Products.update({
           _id: product._id
--- a/server/methods/catalog.js
+++ b/server/methods/catalog.js
@@ -5,6 +5,8 @@ import { EJSON } from "meteor/ejson";
 import { Meteor } from "meteor/meteor";
 import { ReactionProduct } from "/lib/api";
 import { ProductRevision as Catalog } from "/imports/plugins/core/revisions/server/hooks";
+import { publishProductsToCatalog } from "/imports/plugins/core/catalog/server/methods/catalog"
+import { RevisionApi } from "/imports/plugins/core/revisions/lib/api/revisions";
 import { Hooks, Logger, Reaction } from "/server/api";
 import { MediaRecords, Products, Revisions, Tags } from "/lib/collections";
 import { Media } from "/imports/plugins/core/files/server";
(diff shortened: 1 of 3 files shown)


zenweasel:
This isn't a problem with this PR per-se but now you can get yourself in a state where you set the product to private but not published, so you get a 404 when you click on the product. Do we care about that at all? The solution is to just publish but I wonder when revisions are turned off should we auto-publish since your changes take effect immediately on the PDP anyway so you could end up with different data on the PDP and on the grid.

Some clarity on how this should work would be helpful.

@mikemurray

prinzdezibel:
> The solution is to just publish but I wonder when revisions are turned off should we auto-publish since your changes take effect immediately on the PDP anyway so you could end up with different data on the PDP and on the grid.

Yes, this is addressed in this ticket: https://github.com/reactioncommerce/reaction/issues/4074

prinzdezibel:
@mikemurray, @rymorgan  Any insights on this? Perhaps it could make sense to have changes autopublished as Brent suggested if ProductRevisions are disabled.

zenweasel:
If the product was previously public and I make it private it will still show up in the grid, but produce a 404 when you go to the PDP

zenweasel:
Any change is going to essentially "autopublish" to the PDP in this scenario because it does not use the catalog

rymorgan:
Yea, that's not good then, that renders the public/private toggle completely useless it seems. My understanding of public/private is to have a published product that I can turn off/on by toggling it. But if it still shows up on a grid and goes to a 404 that's basically a broken scenario.

zenweasel:
Yeah, I don't think that the "revisions turned off" scenario was completely considered. But if things were automatically published (only in this revisions-off mode obviously) then things would always be synced. And if you want to hide something you can just hide it.

rymorgan:
So changes are always 'live'? But if you wanted to edit a draft, it would need to be a set to private? I suppose that makes sense.

zenweasel:
If you are in revisions-off mode, I would think that would be the expected behavior right?

rymorgan:
Let’s define it as that, yes.

zenweasel:
@prinzdezibel At the risk of upscoping this issue, could we add that to this PR? I would like to keep it in one place for now to be able to test and understand the consequences.

prinzdezibel:
@zenweasel : Applied the following changes that should be sufficient to get this over the finish line:

- Hide Publish button completely if revision control is disabled
- Autopublish field changes
- Check for product type before publishing productIds to prevent the
error: 12:43:05.756Z ERROR Reaction: After applying the update to the document {_id: "t9oHdLoeuNzHCq9cd", ...}, the (immutable) field '_id' was found to have been altered to _id: "oE8bhrE8sNMBt3eTS"

zenweasel:
Something still not working correctly. When I created a product and made is visible the price in the grid is $0. On the PDP it's correct.

prinzdezibel:
Done.

prinzdezibel:
@zenweasel  Could you have a look into this again?

zenweasel:
@prinzdezibel When I have Revisions turned off I no longer see the Publish button

prinzdezibel:
@zenweasel. this is intentional. Changes are auto-published everywhere, so no need for it any longer.

zenweasel:
Ok, just following the test instructions

server/methods/catalog.js
```diff
@@ -675,17 +677,39 @@ Meteor.methods({
...
-    });
+    if (RevisionApi.isRevisionControlEnabled()) {
+      // Flag the variant and all its children as deleted in Revisions collection.
+      // eslint-disable-next-line no-shadow
+      toDelete.forEach((variant) => {
+        Hooks.Events.run("beforeRemoveCatalogProduct", variant, { userId: this.userId });
+        Hooks.Events.run("afterRemoveCatalogProduct", this.userId, variant);
+      });
+    } else {
+      // eslint-disable-next-line no-shadow
```
zenweasel: I wonder if there shouldn't be some hook here too in case someone wants to track deletions (e.g. for syncing with an external service)?

prinzdezibel: Ok, right now the hooks are doing revisions related stuff, AFAIK. Probably it would be better to run the hooks unconditionally and put the switch in there.

loanlaux:
Any status on this? I was looking to see if I could file the issue but was happy to see you had noticed already. How far are we from having the fix released?

aldeed:
@zenweasel This seems like stuff we should fix, but if we aren't going to hotfix this into 1.12.0, then this will need to be rewritten for merge into `1.13.0`, which no longer has revisions support.

mikemurray:
@aldeed @zenweasel This PR may no longer be necessary as I these issues should have been resolved in 1.13.0, in much the same way they are here.

spencern:
closed in favor of the solution to #4319

PR #4069, opened 2018-03-26
(feat): add GraphQL roles query

kieckhafer:
Resolves #3916   
Impact: **minor**  
Type: **feature**

## Issue
Add GraphQL roles query to be able to query the roles available for a shop

## Solution

- New method that fetches the `roles` data, which lives in the `accounts` package
- New resolvers
- Update shop and role schemas

## Breaking changes
- none

## Testing
1. Start the app and sign in
1. open `localhost:3000/graphiql`
1. Run the following queries to get results

```
roles(shopId: "cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg0K") {
  edges {
    node {
      name
    }
  }
}
```

```
{
  shop(id: "cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg0K") {
    roles(first:2) {
      nodes {
        name
    	}
      pageInfo {
        endCursor
        startCursor
      }
      totalCount
    }
  }
}
```

PR #4076, opened 2018-03-27
(feat): add GraphQL roles query

### LINKED ISSUE #3916: Create roles GraphQL query
**Summary**
Create a `roles` query in GraphQL.

**Implementation**
The `roles` query should have the following signature
```
roles(shopId: ID!): [String]
```
It has one argument, `shopId` which is required, and returns an array of Strings which are roles associated with that shopId if the user requesting the groups has permissions to access those roles. For this ticket, it may be necessary to create a new function on the server to fetch the list of roles for this Query. These roles should be a filtered list of all roles available for this shop, where only roles for packages that are enabled are returned.

## Acceptance Criteria
Using GraphQL, I can use the `roles(shopId: ID!)` query to get an array of roles as `String`s by shopId.

kieckhafer:
Resolves #3916   
Impact: **minor**  
Type: **feature**

Closed #4069 in favor of this PR, as there were some issues with capitalization of folders on the last PR.

## Issue
Add GraphQL roles query to be able to query the roles available for a shop

## Solution

- New method that fetches the `roles` data, which lives in the `accounts` package
- New resolvers
- Update shop and role schemas

## Breaking changes
- none

## Testing
1. Start the app and sign in
1. open `localhost:3000/graphiql`
1. Run the following queries to get results

```
roles(shopId: "cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg0K") {
  edges {
    node {
      name
    }
  }
}
```

```
{
  shop(id: "cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg0K") {
    roles(first:2) {
      nodes {
        name
    	}
      pageInfo {
        endCursor
        startCursor
      }
      totalCount
    }
  }
}
```


imports/plugins/core/graphql/server/resolvers/account/Query/roles.js
```diff
@@ -0,0 +1,25 @@
...
+ * @name roles
+ * @method
+ * @summary find and return the roles for a shop
+ * @param {Object} _ - unused
+ * @param {Object} args - an object of all arguments that were sent by the client
+ * @param {String} args.shopId - id of shop to query
+ * @param {String} args.after - Connection argument
+ * @param {String} args.before - Connection argument
+ * @param {Number} args.first - Connection argument
+ * @param {Number} args.last - Connection argument
```
aldeed: Mike made a custom type you can reference in the jsdocs here instead of every pagination arg which are the same everywhere

imports/plugins/core/graphql/server/resolvers/account/Query/roles.test.js
```diff
@@ -0,0 +1,57 @@
...
+  const rolesQuery = jest.fn().mockName("rolesQuery").mockReturnValueOnce(getFakeMongoCursor("roles", rolesData));
+
+  const result = await roles(null, { shopId: shopBase64ID }, {
+    queries: { roles: rolesQuery },
+    userId: "123"
+  });
+
+  expect(result.nodes[0].name).toEqual(role1);
+  expect(result.nodes[1].name).toEqual(role2);
+  expect(result.nodes[2].name).toEqual(role3);
```
aldeed: Faster to use `toBe` rather than `toEqual` when comparing things that are `===`

imports/plugins/core/graphql/server/schemas/role.js
```diff
@@ -1,14 +1,16 @@
...
-  # This may be required if the role name must match a specific pattern.
-  # Could also be made more generic depending on pattern.
-  # Remove if not used.
-  scalar RoleName
+  # The fields by which you are allowed to sort any query that returns an \`RoleConnection\`
+  enum RoleSortByField {
+    _id
+    name
+    createdAt
+    updatedAt
```
aldeed: Does that collection have `createdAt` and `updatedAt`? If not, remove them from here and make `name` the default sort.

kieckhafer: No it does not, good catch.

kieckhafer:
@aldeed updated based on your comments

imports/plugins/core/graphql/server/resolvers/account/Query/roles.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param {Object} context - an object containing the per-request state
+ * @return {Object[]} Promise that resolves with array of user Roles objects
+ */
+export default async function roles(_, { shopId, ...connectionArgs }, context) {
+  // Transform ID from base64
+  const dbShopId = decodeShopOpaqueId(shopId);
+
+  const query = await context.queries.roles(context, dbShopId);
+
+  return await getPaginatedRolesResponse(query, connectionArgs);
```
aldeed: `await` isn't necessary after `return` since `async` functions will always await return values automatically. It doesn't hurt anything as far as I know, but just letting you know.

imports/plugins/core/graphql/server/schemas/role.js
```diff
@@ -26,6 +26,6 @@ export const typeDefs = `
   }
 
   extend type Query {
-    roles(shopId: ID!): RoleConnection
+    roles(shopId: ID!, after: ConnectionCursor, before: ConnectionCursor, first: ConnectionLimitInt, last: ConnectionLimitInt, sortOrder: SortOrder = asc, sortBy: RoleSortByField = createdAt): RoleConnection
```
aldeed: At the end of this, change `sortBy: RoleSortByField = createdAt` to `sortBy: RoleSortByField = name`. That is where the default sortBy field is defined.

imports/plugins/core/graphql/server/schemas/shop.js
```diff
@@ -11,6 +11,9 @@ export const typeDefs = `
 
     # Returns a list of groups for the shop with ID \`shopId\`, as a Relay-compatible connection.
     groups(after: ConnectionCursor, before: ConnectionCursor, first: ConnectionLimitInt, last: ConnectionLimitInt, sortOrder: SortOrder = asc, sortBy: GroupSortByField = createdAt): GroupConnection
+
+    # Returns a list of roles for the shop with ID \`shopId\`, as a Relay-compatible connection.
+    roles(after: ConnectionCursor, before: ConnectionCursor, first: ConnectionLimitInt, last: ConnectionLimitInt, sortOrder: SortOrder = asc, sortBy: RoleSortByField = createdAt): RoleConnection
```
aldeed: At the end of this, change `sortBy: RoleSortByField = createdAt` to `sortBy: RoleSortByField = name`. That is where the default sortBy field is defined.

imports/plugins/core/graphql/server/schemas/shop.js
```diff
@@ -11,6 +11,9 @@ export const typeDefs = `
 
     # Returns a list of groups for the shop with ID \`shopId\`, as a Relay-compatible connection.
     groups(after: ConnectionCursor, before: ConnectionCursor, first: ConnectionLimitInt, last: ConnectionLimitInt, sortOrder: SortOrder = asc, sortBy: GroupSortByField = createdAt): GroupConnection
+
+    # Returns a list of roles for the shop with ID \`shopId\`, as a Relay-compatible connection.
```
aldeed: I know you just copied the comment above, but that one is a bit incorrect. Change both this comment and the one above to say `for this shop` (matching the administrators comment) rather than `for the shop with ID shopId`.

kieckhafer:
@aldeed 👍 updated based on your comments

PR #4082, opened 2018-03-28
(fix) Make shipping discount code case-insensitive

### LINKED ISSUE #4080: Shipping discount is not applied to cart/checkout/order
## Issue Description
The "shipping" discount type is supposed to discount shipping but actually does nothing

## Steps to Reproduce
1. Login as admin
1. Go the Payments/Codes dashboard
1. Add a new discount code with the discount method "shipping" and the code to be "FREESHIPPING" and set the amount to 2.99.
1. Add to cart and apply this discount when checking out
1. Observe that the discount is not applied

### Versions
````
Node: 8.9.4
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.10.0
Reaction branch: master
Docker: 18.03.0-ce
```

### LINKED ISSUE #4150: Shipping discount doesn't work if code does not match shipping name exactly
## Issue Description
Shipping discounts are based on a case-sensitive match between the "discount" parameter and the name of the shipping method. This is more error-prone than it needs to be.

## Steps to Reproduce
1. Login as admin
1. Enable the Standard Flat Rate shipping type
1. Create a discount with the Calculation method being "shipping" and the discount set to "StAnDaRd" and the code as "FreeShipping"
1. Add an item to cart and go to checkout
1. Select Standard Shipping
1. In the payment section apply the "FreeShipping" payment code.
1. Observe that the discount is not taken

### Versions
```
Node: 8.9.4
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.10.0
Reaction branch: fix-3974-prinzdezibel-error-enabling-shopify-hooks
Docker: 18.03.0-ce
```

zenweasel:
Resolves #4150 
Impact: **minor**  
Type: **bugfix**

## Issue
When creating a shipping discount you need to put the name of the method you want to discount and currently the comparison is case-sensitive. When I looked into why the code wasn't working I saw what I going on so I fixed it. I also saw a case where it was trying to calculate shipping when `cart.shipping` was empty so I added a check around the calculation.

## Solution
Make the comparison case-insensitive

## Testing
1. Login as admin
1. Enable the Standard Flat Rate shipping type
1. Create a discount with the Calculation method being "shipping" and the discount set to "StAnDaRd" and the code as "FreeShipping"
1. Add an item to cart and go to checkout
1. Select Standard Shipping
1. In the payment section apply the "FreeShipping" payment code.
1. Observe that it works even though the case of the discount doesn't match

### DIFF
--- a/imports/plugins/included/discount-codes/server/methods/methods.js
+++ b/imports/plugins/included/discount-codes/server/methods/methods.js
@@ -92,13 +92,13 @@ export const methods = {
     let discount = 0;
     const discountMethod = Discounts.findOne(discountId);
     const cart = Cart.findOne(cartId);
-
-    for (const shipping of cart.shipping) {
-      if (shipping.shipmentMethod && shipping.shipmentMethod.name === discountMethod.discount) {
-        discount += Math.max(0, shipping.shipmentMethod.rate);
+    if (cart.shipping && cart.shipping.length) {
+      for (const shipping of cart.shipping) {
+        if (shipping.shipmentMethod && shipping.shipmentMethod.name.toUpperCase() === discountMethod.discount.toUpperCase()) {
+          discount += Math.max(0, shipping.shipmentMethod.rate);
+        }
       }
     }
-
     return discount;
   },
 


zenweasel:
@prinzdezibel Can you make sure you have alerts set up for when reviews are requested and you are checking them?

prinzdezibel:
I've set up alerts, but I guess I've dumped it accidentally.

zenweasel:
Fortunately this is the least important PR ever so....we're all good

PR #4083, opened 2018-03-28
Default settings for parcel size

### LINKED ISSUE #3019: Add settings panel to define dimensions for default parcel to shipping dashboard

Akarshit:
Resolves #3019   
Impact: **minor**  
Type: **chore**

## Issue
When there was no `parcel` present in the `cart` the shipping APIs would not be able to calculate the tax and would err. This case would occur when the Admin hasn't added the dimensions of the product.

## Solution
1. Create a form where the `Admin` can enter the default dimensions that will can be used as a fallback when product dimensions are missing.
1. Create a default for the "default parcel size".

## Breaking changes
None

## Testing
1. Do a `reaction reset -y`
1. Enable shippo.
1. Create and publish a product, but don't add it's dimensions.
1. Buy the product with shippo. Check in the database that `Order.item[0].parcel` = default dimension.
1. Go to shipping -> Parcel Size and change the values
1. Repeat above steps and check the same results.
1. Create a new marketplace shop, enable it.
1. Invite a user to it and log in as that user.
1. Repeat above steps and check the same results. (the parcel settings should get copied over from primary shop to new shop.)

Behavior for old users ->
1. Until the old shop's Admin update the `Parcel Size` settings, the behavior would be same as that was before for them, i.e  the shipping APIs would not be able to calculate the tax and would err if product dimensions are not there.


Akarshit:
Will have to work to fix tests. Marking this WIP.

Akarshit:
@jshimko I have these tests passing in the local with the exact same command. Would have any idea what could be causing this to fail on circle CI.
Basically [this](https://github.com/reactioncommerce/reaction/pull/4083/files#diff-9893c5f8322946b9a0737364d1993d8dR463) doesn't return a shop and I am not sure what could be causing this.

Akarshit:
NVM found a way to reproduce it.

Akarshit:
In the test, `product` was using the `shopId` that it had the time it was created. But during the tests, shops are created and deleted. This was leading to the product having a `shopId` which didn't exist.
I have changed the `shopId` to return a function so that it is fetched from the current shops when a product is created.

imports/plugins/core/shipping/client/templates/settings/shipping.js
```diff
@@ -1,5 +1,6 @@
 import { Template } from "meteor/templating";
 import { Meteor } from "meteor/meteor";
+import { Components } from "@reactioncommerce/reaction-components";
 /*
  * Template shippinges Helpers
```
zenweasel: Not your code but let's fix this typo here

imports/plugins/core/shipping/server/methods/methods.js
```diff
@@ -62,6 +62,40 @@ export const methods = {
...
+  "shipping/updateParcelSize"(parcel) {
+    check(parcel, {
+      weight: Match.Optional(Number),
+      height: Match.Optional(Number),
+      length: Match.Optional(Number),
+      width: Match.Optional(Number)
+    });
+
+    if (!Reaction.hasPermission(shippingRoles)) {
+      throw new Meteor.Error(403, "Access Denied");
```
zenweasel: Please follow the documentation for how to throw errors https://docs.reactioncommerce.com/reaction-docs/master/error-handling-guide

lib/api/helpers.js
```diff
@@ -403,6 +403,59 @@ export function convertWeight(from, to, weight) {
...
+ * @method
+ * @memberOf Helpers
+ * @summary Convert length from/to different Units of Measure
+ * @param {String} from The UOL to convert from
+ * @param {String} to The UOL to convert to
+ * @param {Number} length The value to convert
+ * @returns {Number} The converted value
+ */
+export function convertLength(from, to, length) {
+  check(from, String); // we need to check for specific values
```
zenweasel: Can we check for specific values here?

server/methods/core/cart.js
```diff
@@ -467,6 +473,17 @@ Meteor.methods({
...
       parcel = { weight: variant.weight, height: variant.height, width: variant.width, length: variant.length };
     }
+    // Merge parcel with defaultParcel
+    // so that all fields are filled.
+    if (!parcel) {
+      parcel = defaultParcelSize;
+    } else {
+      Object.keys(parcel).forEach((key) => {
+        parcel[key] = parcel[key] || defaultParcelSize[key];
+      });
```
zenweasel: How does that work? You are assigning a value to itself?

zenweasel:
@Akarshit do we need to write a migration to add those defaults to users with existing shops?

zenweasel:
@Akarshit Should parcel settings be at the top rather than the bottom?

zenweasel:
If I change UOM at the shop level, shouldn't the default values in this settings form change?

zenweasel:
Except for the above questions this seems to work well

server/methods/core/cart.js
zenweasel: No, I was just reading it incorrectly. 👍

Akarshit:
To answer @zenweasel's questions
> do we need to write a migration to add those defaults to users with existing shops?

We do not need a migration for existing shops, those would work the same way they were working previously. Since this is new feature, the shops admins would have to go to the shipping toolbar to enable this. They don't have to change the settings, just going to that menu will add the `defaultParcelSize` values to the DB.

>  Should parcel settings be at the top rather than the bottom?

These are setting that would be set once in a real use case. So it doesn't make to put them at the top IMO.

> If I change UOM at the shop level, shouldn't the default values in this settings form change?

Implementing this.

Akarshit:
Updated

zenweasel:
If I change UOM it still doesn't change the default values

Akarshit:
Hmm, it was working for me. Looking again.

Akarshit:
@zenweasel I tried but couldn't reproduce 
> If I change UOM it still doesn't change the default values

Can you please check again. Also if it happens, can you also pass me the reproduction steps.

zenweasel:
I'm following the testing instructions and when I went to check out with my created product I get this error:


```
06:53:39.331Z ERROR Reaction: The data you sent was not accepted as valid
06:53:40.422Z ERROR Reaction: The data you sent was not accepted as valid
06:53:40.424Z  WARN Reaction: Failed to get shipping methods from these packages: [ { packageName: 'shippo', fileName: 'shippo.js' } ]
```

Akarshit:
@zenweasel The above error should occur when you have a item in your cart before merging this PR. Can you try this with an empty cart. I am not able to reproduce this with a fresh cart.

zenweasel:
@Akarshit There seems to be a problem in the testing instructions. When I try to checkout out with a product with no dimensions I get the "No shipping methods are configured" message even though I have checkout enabled.

imports/plugins/core/shipping/client/containers/parcelSizeSettings.js
```diff
@@ -0,0 +1,115 @@
...
+import { Shops } from "/lib/collections";
+import { ShippingParcel } from "/lib/collections/schemas";
+import { convertWeight, convertLength } from "/lib/api";
+import ParcelSizeSettings from "../components/parcelSizeSettings";
+
+/**
+ * @method saveDefaultSize
+ * @summary call "shipping/updateParcelSize" method
+ * @param {Object} size - size object to be saved
+ * @since 2.0.0
```
zenweasel: since 2.0?

imports/plugins/core/shipping/client/containers/parcelSizeSettings.js
```diff
@@ -0,0 +1,115 @@
...
+ * @return {Function} callback
+*/
+const saveDefaultSize = (size) => {
+  const parcel = ShippingParcel.clean(size);
+  Object.keys(parcel).forEach((key) => {
+    parcel[key] = Number(parcel[key]);
+  });
+  try {
+    ShippingParcel.validate(parcel);
```
zenweasel: We are just catching this and throwing it away?

imports/plugins/core/shipping/client/containers/parcelSizeSettings.js
Akarshit: Fixed

imports/plugins/core/shipping/client/containers/parcelSizeSettings.js
Akarshit: Fixed

zenweasel:
@Akarshit We just have some merge conflicts and then this is ready to go

lib/collections/schemas/shops.js
```diff
@@ -88,6 +89,52 @@ export const Currency = new SimpleSchema({
 
 registerSchema("Currency", Currency);
 
+/**
+ * @name Currency
```
aldeed: Fix the name

lib/collections/schemas/shops.js
```diff
@@ -88,6 +89,52 @@ export const Currency = new SimpleSchema({
...
+ * @property {Number} length default value: 11.25in
+ * @property {Number} width default value: 8.75in
+ */
+export const ParcelSize = new SimpleSchema({
+  weight: {
+    type: Number,
+    defaultValue() {
+      const baseUOM = this.siblingField("baseUOM");
+      return convertWeight("lb", baseUOM, 8);
+    },
```
aldeed: `defaultValue` as a function doesn't work. You could switch to `autoValue`, but the logic will have to be more complex than this. I think it's easier to just add these defaults wherever you're inserting/updating and remove these defaults from the schema.

server/methods/core/cart.js
```diff
@@ -467,6 +473,19 @@ Meteor.methods({
...
+    // so that all fields are filled.
+    if (defaultParcelSize) {
+      if (!parcel) {
+        parcel = defaultParcelSize;
+      } else {
+        Object.keys(parcel).forEach((key) => {
+          parcel[key] = parcel[key] || defaultParcelSize[key];
+        });
+      }
+    }
```
aldeed: There is a simpler way to do this:

```js
parcel = {
  ...(defaultParcelSize || {}),
  ...(parcel || {})
};
```

imports/plugins/core/versions/server/migrations/25_deafult_parcel_size.js
```diff
@@ -0,0 +1,25 @@
...
+  up() {
+    const shops = Shops.find().fetch();
+    shops.forEach((shop) => {
+      const defaultParcelSize = {
+        weight: convertWeight("lb", shop.baseUOM, 8),
+        height: convertLength("in", shop.baseUOL, 6),
+        length: convertLength("in", shop.baseUOL, 11.25),
+        width: convertLength("in", shop.baseUOL, 8.75)
+      };
+      Shops.update({ _id: shop._id }, { $set: { defaultParcelSize } });
```
aldeed: Add a third argument: `{ bypassCollection2: true }`

This prevents the migration from failing as the schema changes in the future. For example, if `defaultParcelSize` is someday removed from the shop schema, we still want this to be able to set it without throwing an error because a later migration would remove it.

imports/plugins/core/versions/server/migrations/25_deafult_parcel_size.js
```diff
@@ -0,0 +1,25 @@
...
+    });
+  },
+  down() {
+    const shops = Shops.find().fetch();
+    shops.forEach((shop) => {
+      Shops.update({ _id: shop._id }, { $unset: { defaultParcelSize: 1 } });
```
aldeed: Add a third argument: `{ bypassCollection2: true }`

Also, this one can be done as a multi update (remove the `find` and `forEach`).

```js
Shops.update({}, { $unset: { defaultParcelSize: 1 } }, { bypassCollection2: true, multi: true });
```

imports/plugins/core/shipping/client/containers/parcelSizeSettings.js
```diff
@@ -0,0 +1,117 @@
...
+ * @summary call "shipping/updateParcelSize" method
+ * @param {Object} size - size object to be saved
+ * @since 1.1.12
+ * @return {Function} callback
+*/
+const saveDefaultSize = (size) => {
+  const parcel = ShippingParcel.clean(size);
+  Object.keys(parcel).forEach((key) => {
+    parcel[key] = Number(parcel[key]);
+  });
```
aldeed: Casting parcel values to `Number` should not be necessary here because `.clean` should be doing it.

imports/plugins/core/shipping/client/containers/parcelSizeSettings.js
```diff
@@ -0,0 +1,117 @@
...
+  if (!doc) {
+    doc = {
+      weight: convertWeight("lb", baseUOM, 8),
+      height: convertLength("in", baseUOL, 6),
+      length: convertLength("in", baseUOL, 11.25),
+      width: convertLength("in", baseUOL, 8.75)
+    };
+    // Save default settings for the first time user.
+    saveDefaultSize(doc);
+  }
```
aldeed: If we need this, I'd rather see it in server code. But why do we need it? You have the migration and fixtures updated. `shop/createShop` method actually just clones an existing one. If they import a shop without it, that's not our problem.

Akarshit:
Made all the changes requested.

aldeed:
@Akarshit Looking pretty good now. Please merge `release-1.14.0` into this, and when updating the migration filename, also fix the spelling of `default` in the filename. Then make sure all tests and lint are passing.

Akarshit:
@aldeed I have updated the PR with your changes.

spencern:
@aldeed This PR looks like it's probably good to go now.

PR #4086, opened 2018-03-28
Shippo address validation

### LINKED ISSUE #3823: "No shipping methods are configured" when the address is not deliverable to
### Expected behavior
If products can't be delivered to the address, show some relevant error like "Address not deliverable"
### Actual behavior
Getting "No shipping methods are configured" in the Shipping step.
1. Enable only Shippo
1. But a product and enter some fake/incorrect address.
1. Observe the shipping step say "No shipping methods are configured"

### Versions
`release-1.8.0`

jshimko:
Resolves #3823 
Impact: **critical**  
Type: **bugfix**

## Issue
When a user saves an invalid address in the address book form while checking out, Shippo is unable to find any rates and ambiguously fails because they require the address to be valid to get rates. The error returned is "No shipping methods are configured".

## Solution
Aside from obviously making that error more accurate, Shippo also needs to have the address validated _before_ getting rates is attempted. They provide a separate API endpoint for this purpose. This PR adds server side methods that use this API to either confirm validity or return a better suggested address.
So basically Shippo works the same as Avalara.

## Breaking changes
None expected.

## Testing
1. Add a product to cart.
1. Go to the Checkout Screen.
1. Try entering a partially valid address. 
1. You should get a toast, regarding the error in the address.
1. Fix it, enter a almost complete address.
1. You should be taken to the address review screen. This screen has the same functionality as Avalara address review screen.
1. Try `Edit entered address`, and submit with same address. (Basically testing all the workflows work)
1. Try `Use selected address`, the address should be added and tax should be calculated.

Repeat the above strep with Shippo disabled and Avalara enabled(to check nothing is broken for Avalara).

## Help
```js
// simply provide the cartId and the method will grab the shipping address
Meteor.call("shippo/validateCartAddress", cartId, (err, result) => {
  if (!err) {
    console.log(result);
  }
});

// if valid
{    
  "object_state": "VALID",
  "object_purpose": "QUOTE",
  "object_source": "VALIDATOR",
  "object_created": "2018-03-28T13:21:19.090Z",
  "object_updated": "2018-03-28T13:21:19.370Z",
  "object_id": "3d9599e77b474de0b188e834b6e93209",
  "object_owner": "admin@reactioncommerce.com",
  "name": "Jeremy Shimko",
  "company": "Reaction Commerce",
  "street_no": "",
  "street1": "123 Main St",
  "street2": "",
  "street3": "",
  "city": "Burlington",
  "state": "CT",
  "zip": "06013-1513",
  "country": "US",
  "longitude": -72.93827,
  "latitude": 41.7792,
  "phone": "",
  "email": "me@example.com",
  "is_residential": true,
  "ip": "",
  "messages": [],
  "metadata": "",
  "test": true
}

// if invalid
{
  "object_state": "INVALID",
  "object_purpose": "QUOTE",
  "object_source": "VALIDATOR",
  "object_created": "2018-03-28T13:26:14.105Z",
  "object_updated": "2018-03-28T13:26:14.636Z",
  "object_id": "1ad6b9337a96493c8e39b6e9c9383512",
  "object_owner": "admin@reactioncommerce.com",
  "name": "Jeremy Shimko",
  "company": "Reaction Commerce",
  "street_no": "",
  "street1": "whatever st",
  "street2": "",
  "street3": "",
  "city": "STUFF",
  "state": "AK",
  "zip": "12345",
  "country": "US",
  "longitude": null,
  "latitude": null,
  "phone": "",
  "email": "me@example.com",
  "is_residential": null,
  "ip": "",
  "messages": [
    {
      "source": "Shippo Address Validator",
      "code": "Invalid City",
      "type": "address_error",
      "text": "The City in the address submitted is invalid. Remember, city names cannot begin with numbers."
    }
  ],
  "metadata": "",
  "test": true
}
```


imports/plugins/included/shipping-shippo/server/methods/shippo.js
```diff
@@ -350,6 +350,65 @@ export const methods = {
...
+      Logger.error(msg);
+      throw new Meteor.Error("address-validation-error", msg);
+    }
+
+    const apiKey = getApiKey(cart.shopId);
+
+    if (!apiKey) {
+      const msg = "No Shippo API key was found.";
+      Logger.error(msg);
+      throw new Meteor.Error("shippo-api-error", msg);
```
zenweasel: I think this is an example of where we would be using `Logger.fatal`

imports/plugins/included/shipping-shippo/server/methods/shippo.js
```diff
@@ -350,6 +350,65 @@ export const methods = {
...
+    }
+
+    // get the buyer's email
+    const buyer = Accounts.findOne({ _id: cart.userId }, { fields: { emails: 1 } });
+
+    // check that there is address available in cart
+    if (!cart.shipping || !cart.shipping[0] || !cart.shipping[0].address) {
```
zenweasel: This check won't work in a marketplace context. Generally we need to never use `shipping[0]` and eliminate wherever we see it

Akarshit:
@jshimko @zenweasel I have pushed my changes to the branch(but it's not ready for review).

Basically I have made `shippo/addressValidation` to return the same object as `avalara/addressValidation` so that shippo's address validator can be used as a drop-in replacement for the avalara's address validator.

This is a working version that does address validation. Though some corner cases are still missing and will be fixed as I get answers to the doubts I have. Also this uses the old UI.

The plan is to get this final as soon as #4054 is merged.

Akarshit:
@jshimko I am not able to add you as a reviewer to this(maybe because you opened this PR). But still it would be better if you can have a look on this.

imports/plugins/included/shipping-shippo/register.js
```diff
@@ -31,6 +35,12 @@ Reaction.registerPackage({
     provides: ["shippingSettings"],
     container: "connection",
     template: "shippoSettings"
+  }, {
+    label: "Shippo",
+    icon: "fa fa-plane",
+    name: "shipping/addressValidation/shippo",
+    provides: ["addressValidation"],
+    template: "shippoSettings"
```
Akarshit: @jshimko Will this need a migration?

Akarshit: @spencern Since everyone is out can you help here?

spencern: @Akarshit what do you need help with?

Akarshit: @spencern I am not familiar with how change in registering a template works. So I have added a couple of fields above. Will they be automatically read or will they need a migration?

Akarshit: @spencern could you have a look at this.

spencern: I don't think that updates to the registry require a migration, but it might be required to update the template being requested

zenweasel:
@Akarshit Can you resolve conflicts here so I can review?

imports/plugins/included/shipping-shippo/register.js
```diff
@@ -31,6 +35,12 @@ Reaction.registerPackage({
     provides: ["shippingSettings"],
     container: "connection",
     template: "shippoSettings"
+  }, {
+    label: "Shippo",
```
zenweasel: I don't think you need a separate entry here. You should just be adding this to the `provides` array

imports/plugins/included/shipping-shippo/server/methods/shippo.js
```diff
@@ -30,6 +30,29 @@ function createShippoAddress(reactionAddress, email, purpose) {
   return shippoAddress;
 }
 
+function createValidatedAddress(apiResult) {
+  const formErrors = [];
+  if (apiResult.messages) {
+    apiResult.messages.forEach((message) => {
```
zenweasel: Could you just do this in a `map`?

imports/plugins/included/shipping-shippo/server/methods/shippo.js
```diff
@@ -30,6 +30,29 @@ function createShippoAddress(reactionAddress, email, purpose) {
...
+      formErrors.push({
+        summary: message.code,
+        details: message.text
+      });
+    });
+  }
+  return {
+    validatedAddress: {
+      address1: apiResult.street1,
+      address2: apiResult.street2,
```
zenweasel: Does this always return a result for `street2` or do we need to protect against this being `undefined`?

imports/plugins/included/shipping-shippo/server/methods/shippo.js
```diff
@@ -30,6 +30,29 @@ function createShippoAddress(reactionAddress, email, purpose) {
   return shippoAddress;
 }
 
+function createValidatedAddress(apiResult) {
```
zenweasel: JSDoc?

imports/plugins/included/shipping-shippo/server/methods/shippo.js
```diff
@@ -349,6 +372,59 @@ export const methods = {
...
+      const msg = "No Shippo API key was found.";
+      Logger.fatal(msg);
+      throw new Meteor.Error("shippo-api-error", msg);
+    }
+
+    // get the buyer's email
+    const buyer = Accounts.findOne({ _id: this.userId }, { fields: { emails: 1 } });
+
+    if (!address) {
+      const msg = "The 'shipping' property of this cart is either missing or incomplete.";
```
zenweasel: So this message feels slightly incorrect/misleading. If we checked that address is missing can we just say that?

imports/plugins/included/shipping-shippo/server/methods/shippo.js
```diff
@@ -349,6 +372,59 @@ export const methods = {
...
+      Logger.error(msg);
+      throw new Meteor.Error("address-validation-error", msg);
+    }
+
+    const shop = Shops.findOne({ _id: Reaction.getShopId() }, { fields: { emails: 1 } });
+
+    // TODO take a more elegant approach to guest checkout -> no email address
+    let email = shop.emails[0].address || "noreply@localhost";
```
zenweasel: I feel like there is a way to check for the `default` email rather just hardcoding the `0` index

imports/plugins/included/shipping-shippo/server/methods/shippo.js
```diff
@@ -349,6 +372,59 @@ export const methods = {
...
+    // make sure we mark this as an address validation for the Shippo API
+    shippoAddress.validate = true;
+
+    try {
+      const apiResult = ShippoApi.methods.validateAddress.call({ address: shippoAddress, apiKey });
+      if (apiResult.errors) {
+        return apiResult;
+      }
+      const validateAddress = createValidatedAddress(apiResult);
+      return validateAddress;
```
zenweasel: This should probably be `validatedAddress`. `validateAddress` sounds like it should be a function

imports/plugins/included/shipping-shippo/server/methods/shippo.js
```diff
@@ -349,6 +372,59 @@ export const methods = {
...
+    } catch (error) {
+      Logger.error(error);
+      throw new Meteor.Error(error, "Shippo API error");
```
zenweasel: The form here is "short description" for first parameter and then the longer description for the second

lib/collections/schemas/address.js
```diff
@@ -115,3 +115,32 @@ export const Address = new SimpleSchema({
...
+  },
+  "errorDetails": {
+    type: Array,
+    optional: true
+  },
+  "errorDetails.$": {
+    type: ErrorDetails,
+    optional: true
+  }
+});
```
zenweasel: Missing newline

zenweasel:
@Akarshit We have outstanding changes here and conflicts to resolve. Not sure if this makes sense to close this and reopen under your own name?

Is this still going to be relevant once we remove Shippo from core? I see a few changes to non-Shippo code? We should still fix this in the Shippo plugin if it doesn't need to go into core.

imports/plugins/included/shipping-shippo/register.js
Akarshit: @zenweasel There is some code that extracts "shippo" from the registry's `name`. This is how we get "avalara" from Avalara's registry. So this entry is needed.

imports/plugins/included/shipping-shippo/server/methods/shippo.js
Akarshit: @zenweasel No this won't always return a result but even it doesn't the `address2` will be undefined, which is not a problem.

Akarshit:
@zenweasel updated.

imports/plugins/included/shipping-shippo/server/methods/shippo.js
Akarshit: @zenweasel Just found `Reaction.getShopEmail()`, which does the exact same things, but will be better to use here.

spencern:
@Akarshit we're super close with this I think. Let's merge the updated branch of 1.14 which should resolve the snyk issues and update conflicts and get this reviewed again.

Akarshit:
@zenweasel this is ready for another review now.

zenweasel:
@Akarshit If I don't enable Shippo and I enter a valid address I am getting this error:

```
Exception while invoking method 'accounts/validateAddress' TypeError: Cannot read property 'settings' of undefined
    at getValidator (imports/plugins/core/accounts/server/methods/validateAddress.js:56:31)
    at MethodInvocation.validateAddress (imports/plugins/core/accounts/server/methods/validateAddress.js:163:21)
    at currentArgumentChecker.withValue (packages/check/match.js:118:15)
    at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1189:12)
    at Object._failIfArgumentsAreNotAllChecked (packages/check/match.js:116:43)
    at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1764:18)
    at DDP._CurrentMethodInvocation.withValue (packages/ddp-server/livedata_server.js:719:19)
    at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1189:12)
    at DDPServer._CurrentWriteFence.withValue (packages/ddp-server/livedata_server.js:717:46)
    at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1189:12)
    at Promise (packages/ddp-server/livedata_server.js:715:46)
    at new Promise (<anonymous>)
    at Session.method (packages/ddp-server/livedata_server.js:689:23)
```

zenweasel:
This looks and works 1000% better than it used to but it seems like we should still try to get these horizontally aligned

zenweasel:
When i do "Use Selected Address" from the above screen I get two the same address twice:

zenweasel:
If i enter an international address I get a "Address Validation service not available" message and I can't continue. Is that the desired behavior?

zenweasel:
@Akarshit I think the only critical issue is that we need to properly handle when Shippo is installed but not enabled properly (although, after 14 this will be a rare case). It's super close.

Akarshit:
@zenweasel Thanks for pointing these errors.
I have updated the code to fix the error on saving address, when Shippo was disabled.

The style error on address review is because `address2` isn't present on the suggested address which creates a empty div, which gets a height of 0 by default. It would be best to open a separate issue for this.

I was not able to reproduce the "When using use selected address, the address is saved twice" issue.

zenweasel:
@spencern If you have a moment, could you merge so we don't need to resolve conflicts, etc. again? And we can start on removing Shippo?

PR #4094, opened 2018-03-29
(feat): add setAccountProfileCurrency mutation

### LINKED ISSUE #3922: Create a setAccountProfileCurrency mutation in GraphQL
**Summary**
Create an `setAccountProfileCurrency` mutation in GraphQL.

**Implementation**

Signature:

```
setAccountProfileCurrency(input: SetAccountProfileCurrencyInput!): SetAccountProfileCurrencyPayload
```

Input argument:

```
  input SetAccountProfileCurrencyInput {
    # The account ID, which defaults to the viewer account
    accountId: ID

    # An optional string identifying the mutation call, which will be returned in the response payload
    clientMutationId: String

    # The currency code
    currencyCode: String!
  }
```

It returns the `Account` that was successfully updated. This mutation should always act upon the account of the viewer.

## Acceptance Criteria
Using GraphQL, I can use the mutation to set the currency stored on the viewer Account profile.

kieckhafer:
Resolves #3922  
Impact: **minor**  
Type: **feature**

## Issue
Add a `setAccountProfileCurrency` mutation to allow GraphQL to update and return currency data for an account

## Solution
- Create mutation file for `setAccountProfileCurrency`
- Update `accounts/setProfileCurrency` meteor method to take in a userId, and to return an account
- Update existing calls to `accounts/setProfileCurrency` to add userId

## Breaking changes
None. If a `userId` is not provided, the code will default to the current user, which is what happens at this time.

## Testing
1. Start up the app
1. Navigate to localhost:3000/graphiql
1. Run the following, with `accountId` being the base64 encoding of "reaction/account:{accountId}"
```
mutation {
  setAccountProfileCurrency(input: {
    accountId: "reaction/account:{accountId}"
    currencyCode: "USD"
    clientMutationId: "clientMutationId13579"
  }) {
    clientMutationId
    account {
      createdAt
      currency {
        code
        _id
        symbol
        rate
        format
      }
    }
  }
}
```


server/methods/accounts/accounts.js
```diff
@@ -1057,16 +1057,27 @@ export function createFallbackLoginToken() {
  * @name accounts/setProfileCurrency
  * @memberof Methods/Accounts
  * @method
+ * @param {String} accountId - accountId of user to set currency of
+ * @param {String} currencyName - currency symbol to add to user profile
  * @summary Sets users profile currency
  */
-export function setProfileCurrency(currencyName) {
+export function setProfileCurrency(accountId, currencyName) {
```
aldeed: I'd rather see `accountId` added as second arg, which is then backwards compatible and you don't have to modify the places that call this.

This also needs to check to make sure the current user has permission to update accountId

imports/plugins/core/graphql/server/resolvers/xforms/currency.js
```diff
@@ -0,0 +1,25 @@
...
+  assoc("_id", k)
+)(v);
+
+// map over all provided currencies, provided in the format stored in our Shops collection,
+// and conver them to the format that GraphQL needs
+const xformLegacyCurrencies = compose(map(xformCurrencyEntry), toPairs);
+
+// retrive all currencies from the Shops collection and xform them
+function getXformedCurrenciesByShop() {
+  const shopCurrencies = Reaction.getShopCurrencies();
```
aldeed: I think we'll want to merge my changes that give access to Mongo outside of Meteor first, and then change this to use the collections from `context`. We need to avoid importing `Reaction`

aldeed:
@willopez Do you mind reviewing this one? Ignore the snyk failures because they're fixed in my other PR

willopez:
@aldeed I'll take look.

willopez:
NOTE: There is a known issue with an account's rate that will be addressed in a future PR.

PR #4097, opened 2018-03-30
Update GraphQL queries to not need Meteor

aldeed:
Resolves 
Impact: **minor**  
Type: **refactor**

## Issue
We would rather have the GraphQL queries use Node Mongo directly and have no dependencies on Meteor.

## Solution
- The GraphqL resolver context now includes a `collections` object, which has all the raw mongo collections.
- All queries are updated to use `collections` object and pure mongo API, and do not import Meteor collections. To support this, they now return Promises.
- Some Meteor accounts functions are ported over, allowing the devserver to now have real authentication based on `meteor-login-token` header rather than a mock user context.
- A version of `Reaction.hasPermission` with no Meteor dependencies is available in the `userHasPermission` property of `context` and is used by the queries rather than importing the Meteor one.
- A new `docker-compose-devserver.yml` file enables running the devserver linked to the same mongo as the Meteor Reaction container.

## Breaking changes
None

#### First Use - Obtain a token from Meteor
For first time use, start Reaction AND devserver in the background:

    docker-compose up -d

Go to http://localhost:3000 and log in. Get the `Meteor.loginToken` value out of local storage.
Now you can keep Reaction running if you want or stop it with

    docker-compose stop reaction

Open GraphiQL standalone app and enter endpoint http://localhost:3030/graphql-alpha
In GraphiQL, edit headers and set meteor-login-token header to the value copied from local storage
You should now be able to make authenticated queries in GraphiQL

#### Run devserver only
After you have a token you can run the devserver alone, which will consume fewer resources.

    docker-compose up -d devserver

If you `npm install` or do anything that modifies `node_modules`, you'll have to rerun the `docker-compose up` commands with `--build` option.

## Testing
1. See "Running" steps above
2. With it running, all GraphQL queries should work and should return real data using real authentication from the token header.
3. All query responses should be identical to what the server running within Meteor returns (`localhost:3000/graphql-alpha` vs `localhost:3030/graphql-alpha`)


ticean:
@aldeed This is great. I pushed a change that combines the devserver into `docker-compose.yml`. Now developers won't have to specify a compose file for every command. I also updated instructions in the PR description.

PR #4142, opened 2018-04-03
Fix for Send messages sent to "Logger.fatal" to email"

### LINKED ISSUE #818: Send messages sent to "Logger.fatal" to email or other more visible means
We want to notify sysadmin if we hit a `fatal` error so that they can get their site back up. Probably just need to create a new stream. Possibly pump this to a collection?

Would be great to have it go to console for development and then to email/monitoring in production.

prinzdezibel:
Resolves #818   
Type: **feature**

This PR closes #818

Although this PR is working as intended, I'll put it in WIP state
because we should/could streamline it with the new Reaction Logger
package: https://github.com/reactioncommerce/logger

## Issue
We want to notify sysadmin if we hit a fatal error so that they can get their site back up.

## Solution / Changes

This PR implements the minimal requirements listed in #818: Sending email alerts to a admin whenever a Logger.fatal() is executed. There are many possible ways to enhance this through more sophisticated handling (e.g. persistency, ensure that no duplicates are send, etc, etc). This is just intended as starting point.

- Use a mutuable export to allow a email stream to be attached to the
 logger instance after the application has been intialized
 (afterCoreInit). This is necessary to have the environment in place to
 specify nodemailer's SMTP options

- Create a new email bunyan stream for sending fatal errors as email

- Introduce a new environment variable REACTION_ALERTS_NOTFICATION_EMAIL
to specify the recipient of the email alerts

## Open points
- Should the functionality of this PR be migrated to the newly Reaction Logger NPM package. If yes, should we pass the dependencies (host, server, password, user) into the ReactionLogger or should we go for a callback based approach (the plugin could offer an API to register logging streams)

- How do we deal with email alerts when installed client-side? e.g. SMTP password? 

## Breaking changes
None expected

## Testing 
**Important:** This test instructions were originally targeting Avalara with rc-configuration. Don't do this as it will lock up the account). Try with taxcloud instead, where there's also a Logger.fatal() call.

1. `meteor npm install`
2. use rc-config for settings
3.  export REACTION_ALERTS_NOTFICATION_EMAIL="yourname@reactioncommerce.com"
4. as an admin, enable avalara
5. put a wrong API password there and save
6. restart server
7. put an item to cart and checkout
8. see an error in the console
```
Error: failed [401] {"error":{"code":"AuthenticationException","message":"Authentication failed.","target":"HttpRequestHeaders","details":[{"code":"AuthenticationException","number":30,"message":"Authentication failed.","description":"Unable to authenticate the user or the account.","faultCode":"Client","helpLink":"http://developer.avalara.com/avatax/errors/AuthenticationExce
ption","severity":"Exception"}]}}
```
9. Check your email inbox for this error.

### DIFF
--- a/imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
+++ b/imports/plugins/included/taxes-avalara/server/methods/taxCalc.js
@@ -59,7 +59,7 @@ function checkConfiguration(packageData = taxCalc.getPackageData()) {
     }
   }
   if (!isValid) {
-    Logger.error("The Avalara package is not configured correctly. Cannot continue");
+    Logger.fatal("The Avalara package is not configured correctly. Cannot continue");
   }
   return isValid;
 }
@@ -108,6 +108,7 @@ function parseError(error) {
   }
 
   if (error.response && error.response.statusCode === 401) {
+    Logger.fatal("Avalara authentication failed", error.message);
     // authentification error
     errorData = {
       errorCode: 401,
--- a/imports/plugins/included/taxes-taxcloud/server/hooks/hooks.js
+++ b/imports/plugins/included/taxes-taxcloud/server/hooks/hooks.js
@@ -45,7 +45,7 @@ MethodHooks.after("taxes/calculate", (options) => {
       // check if plugin is enabled and this calculation method is enabled
       if (pkg && pkg.enabled === true && pkg.settings.taxcloud.enabled === true) {
         if (!apiKey || !apiLoginId) {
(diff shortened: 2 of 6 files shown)


zenweasel:
@prinzdezibel I like the idea of that variable to override the email but I would like it to at least default to sending those emails to the shop admin.

prinzdezibel:
@jshimko Would be great if you could have a look into that and if you're agreeing with the general direction, I would value some tips on how I should deal with this functionality in respect to the new Reaction Logger NPM package.

Issue #4071, opened 2018-03-27
Create NPM package for Hooks


zenweasel:
Move current server-side Hooks API to an NPM package that works with the generator and works on both Client and Server

jshimko:
This has been completed in #4147

PR #4147, opened 2018-04-04
migrate Hooks API to an npm package

jshimko:
Issue: #4071 
Impact: **minor**  
Type: **refactor**

## Issue
We're currently working on migrating as many plugins/API's to npm as possible.  `Hooks` has no Meteor/Reaction dependencies, so it was an easy win.

## Solution
Move the Hooks API to [an npm package](https://www.npmjs.com/package/@reactioncommerce/hooks) and update imports to use it. 

## Breaking changes
None expected.  `Hooks` is still exported at `/server/api`, so backward compatibility has been maintained.

## Testing
Start Reaction and note that all event hooks still work.

Example usage...

```js
import Hooks from "@reactioncommerce/hooks";

// extend the user object with a custom field when creating a new user

Hooks.Events.add("onCreateUser", (user) => {
  const newUserObj = Object.assign({}, user);

  newUserObj.customField = "someValue";
  
  return newUserObj;
});
```

Also note that now that this package is on npm, we can also use this API on the client to run client-only hooks (was previously server-only).  And there's nothing Reaction-specific about the package, so you could literally use it in any Node/javascript app.


package.json
```diff
@@ -24,6 +24,7 @@
     "@reactioncommerce/file-collections-sa-gridfs": "0.0.1",
     "@reactioncommerce/job-queue": "^1.0.4",
     "@reactioncommerce/nodemailer": "^5.0.5",
+    "@reactioncommerce/hooks": "^1.0.1",
```
zenweasel: I think for these core dependencies we should "hard pin" them to a specific version, so that a PR to core is still required before an upgrade happens.

aldeed: Discussed with Jeremy. I will do a separate PR to pin all of these

PR #4149, opened 2018-04-04
migrate schemas API to an npm package

jshimko:
Impact: **minor**  
Type: **refactor**

## Issue
We're currently working on migrating as many plugins/API's to npm as possible.  

## Solution
Move entire `Schemas` API to [an npm package](https://www.npmjs.com/package/@reactioncommerce/schemas) and update imports to use it. 

## Breaking changes
None expected. 

## Testing
Start Reaction and note that everything still works.

Available imports...

```js
import Schemas { Validation } from "@reactioncommerce/schemas";

// formerly...
// import { getSchemas, Validation } from "@reactioncommerce/reaction-collections";
```

Also note that there's nothing Reaction-specific about the new package, so you could literally use it in any Node/javascript app.


jshimko:
@aldeed Requested changes have been made.  Let me know if you see anything else.  Thanks!

PR #4158, opened 2018-04-05
Fix for "Receive error in Orders dashboard when using dollar discount codes

### LINKED ISSUE #4079: Receive error in Orders dashboard when using dollar discount codes
## Issue Description
The Order dashboard is trying to treat Discounts like other payment methods (which technically it is) and call `refund/list` on it, but of course that doesn't exist. Get this error on the client side. Related to #4078 but a different part of the code.

```

{message: "Method 'code/refund/list' not found [404]", name: "Error", stack: "Error: Method 'code/refund/list' not found [404]↵ …7063aa833005aff7419b3f8c222e877582f1a3e0:2044:15)", code: undefined, signal: undefined}
hostname
```

## Steps to Reproduce
1. Login as admin
1. Go the Payments/Codes dashboard
1. Add a new discount code with the discount method "credit" and the code to be "DOLLAROFF" and the amount to 1.
1. Add to cart and apply this discount when checking out
1. Go to the order dashboard and select that order
1. Observe the errors in the console

### Versions
```
Node: 8.9.4
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.10.0
Reaction branch: master
Docker: 18.03.0-ce
```

prinzdezibel:
Resolves #4079  
Impact: **major**  
Type: **bugfix**

## Issue
The Order dashboard is trying to treat Discounts like other payment methods (which technically it is) and call `refund/list` on it, but of course that doesn't exist. Get this error on the client side console. 

```
{message: "Method 'code/refund/list' not found [404]", name: "Error", stack: "Error: Method 'code/refund/list' not found [404]↵ …7063aa833005aff7419b3f8c222e877582f1a3e0:2044:15)", code: undefined, signal: undefined}
hostname
```

## Solution / Changes

- Ensure that shopId is specified for the billing record when payment
method is "discount"
- Implement refund/list for payment provider, which essentially returns
an empty list (no refunds possible)

## Breaking changes
none expected

## Testing
1. Login as admin
2. Go the Payments/Codes dashboard
3. Add a new discount code with the discount method "credit" and the code to be "DOLLAROFF" and the amount to 1.
4. Add to cart and apply this discount when checking out
5. Go to the order dashboard and select that order
6. No error should be displayed in console. 
7. Capture payment
8. No `$NaN.undefined` should be displayed in invoice field `Adjusted Total`


zenweasel:
Rather than doing this I wonder if we couldn't utilize the "supports" array and just say this payment doesn't support refunds?

prinzdezibel:
Good idea, @zenweasel. Done as you proposed.

imports/plugins/included/payments-authnet/server/methods/authnet.js
```diff
@@ -179,11 +179,6 @@ Meteor.methods({
     };
 
     return result;
-  },
```
Akarshit: @prinzdezibel Do we need this change to fix this issue?

Akarshit:
I keep getting [these](https://gist.github.com/Akarshit/7aca9c1f3ca0a3b70610d4a42cb7fb33) error while doing checkout. Could these be related to this PR?

server/methods/core/payments.js
```diff
@@ -32,16 +32,20 @@ export const methods = {
...
     const billing = cart.billing[0];
     const billingId = Random.id();
+    const shopId = Reaction.getShopId();
     const result = Collection.update({
       _id: id
     }, {
       $addToSet: {
         billing: {
           ...billing,
+          shopId,
```
zenweasel: This is not correct. It is causing this to create two billing records even when not in a marketplace scenario. So this second record is incomplete which is why there is no subtotal.

zenweasel:
Closing this as it continues down the wrong road of creating additional billing records for each payment method which just creates more problems. The real solution lies in: https://github.com/reactioncommerce/reaction/issues/4229

PR #4164, opened 2018-04-09
Fix for "react-addons-create-fragment should be removed"

### LINKED ISSUE #4163: react-addons-create-fragment should be removed
## Prerequisites

## Issue Description

The package react-addons-create-fragment should be removed for 2 reasons:

- It's a legacy React addon and it's not longer maintained. New code should not rely on it
- It causes problems when combined with ReactDOM.hydrate() / SSR, because it does not use the special React property `key` for order determination.

### Versions
Node: 8.2.1
NPM: 5.3.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.10.0
Reaction branch: master
Docker: 17.03.1-ce

prinzdezibel:
This PR closes issue #4163

Changes:
  Remove package react-addons-create-fragment.
  The migration path is to use arrays with explicit keys on individual elements.

Resolves #4163   
Impact: **minor**  
Type: **chore**

## Issue
The package react-addons-create-fragment should be removed for 2 reasons:

It's a legacy React addon and it's not longer maintained. New code should not rely on it
It causes problems when combined with ReactDOM.hydrate() / SSR, because it does not use the special React property key for order determination.

## Breaking changes
none

### DIFF
--- a/imports/plugins/core/layout/client/templates/layout/alerts/reactionAlerts.js
+++ b/imports/plugins/core/layout/client/templates/layout/alerts/reactionAlerts.js
@@ -16,8 +16,9 @@ import { getRootNode } from "/imports/plugins/core/router/client/browserRouter.j
  */
 
 const getAlertWrapper = () => {
-  const rootNode = getRootNode();
-  
+  getRootNode();
+  const rootNode = document.getElementById("react-root");
+
   rootNode.insertAdjacentHTML("beforebegin", "<div id='s-alert-wrapper'></div>");
   return document.getElementById("s-alert-wrapper");
 };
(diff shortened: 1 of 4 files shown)


prinzdezibel:
@mschipperheyn Thanks for the Tip! I've modified this PR to use React.Fragment.

New insight: This PR was specifically intended to solve issue no 121 of a customer's project (ICM+). But as it turned out the React addon was not the real cause of the problem. It's not clear what actually caused it. The reason this PR worked was because it introduced an additional wrapper div that somehow masked the error.

**TLDR** This PR only resolves the first point of the issue:
> It's a legacy React addon and it's not longer maintained.

It does not address point 2:

> It causes problems when combined with ReactDOM.hydrate() / SSR, because it does not use the special React property key for order determination.

zenweasel:
I don't feel qualified to review this. @aldeed Could you or @kieckhafer take a look at this?

PR #4172, opened 2018-04-12
Fix for "document.getElementBuId("react-root") may fail"

### LINKED ISSUE #4171: document.getElementBuId("react-root") may fail
## Prerequisites

## Issue Description
/core/layout/client/templates/layout/alerts/reactionAlerts.js calls `getRootNode()` which always ensures a DOM node that serves as React mount point to be created.

But there's a subsequent `document.getElementBuId("react-root")` which will only work as intended if getRootNode really creates a node called "react-root". This is not always the case (for a customer implementation which uses two-pass rendering there's a temporary React mount point that is called differently.

## Possible Solution
Use the return value of `getRootNode()` which always contains the correct node, no matter how it is named.

prinzdezibel:
Resolves #4171  
Impact: **minor**  
Type: **bugfix**

## Issue

## Solution
Use the return value of getRootNode() which always contains the correct node, no matter how it is named.

## Breaking changes
none

### DIFF
--- a/imports/plugins/core/layout/client/templates/layout/alerts/reactionAlerts.js
+++ b/imports/plugins/core/layout/client/templates/layout/alerts/reactionAlerts.js
@@ -16,8 +16,7 @@ import { getRootNode } from "/imports/plugins/core/router/client/browserRouter.j
  */
 
 const getAlertWrapper = () => {
-  getRootNode();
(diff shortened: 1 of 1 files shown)


spencern:
@Akarshit This may be a good fix, but it needs more detail in the description before I can merge it.

PR #4182, opened 2018-04-16
Fix #4141: Options now have parent's taxCode

### LINKED ISSUE #4141: Shopify Connector Error: Input buffer contains unsupported image format
## Issue Description
Ran shopify connector for my existing shop. The following error showed in the terminal `Error: Input buffer contains unsupported image format
=> Exited with code: 1`

I have not investigated further at this stage.

## Steps to Reproduce
Not yet defined.
To try:

- setup Shopify store with png images
- run connector

### Versions

Node: 8.9.0
NPM: 5.7.1
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.10.0
Docker: 18.03.0-ce

### LINKED ISSUE #4181: TaxCloud: Taxcodes attached to a variant not copied to option.
## Issue Description
The taxcode attached to a product is not copied over to it's options. Similarly when updating a variant with a different taxcode, the options don't get updated.

## Steps to Reproduce
1. Create a product
2. Add a taxcode to it
3. Create it's options, observe that the `taxCode` of options is still `"0000"` in the DB.

## Possible Solution
When changing taxCode all the option's `taxCode` field should be updated.

Akarshit:
Resolves #4181   
Impact: **minor**  
Type: **bugfix**

## Issue
The `taxCode` field was only available on variant level, not on option level. This was fine for the UI, but even in the database, options were not having the `taxCode` field updated/created. Hence all the options had the default value of `0000` in the `taxCode` field.

## Solution
Whenever a product is published, the `taxCode` field from its variants is copied over to its options. Also whenever a option is created, it gets a initial value of `taxCode` same as of its parent.

## Breaking changes
None

## Testing
1. Create a product with variants and options.
2. Set a taxCode on the product and publish.
3. Look in the database for its options. They should have the updated `taxCode` field.
4. Create a new option for this product.
5. Look in the database for this option. It should have the `taxCode` same as that of parent.


imports/plugins/core/taxes/server/methods/methods.js
```diff
@@ -124,6 +125,36 @@ export const methods = {
     });
   },
 
+  /**
+   * @name "taxes/updateTaxCode"
+   * @method
+   * @memberof Methods/Taxes
+   * @summary updates the taxcode on options of a product.
+   * @param  {String} products array of products to be updated.
+   * @return {undefined} returns nothing
```
zenweasel: I think it's best practice to always have a function return a value rather than undefined. Maybe a count of updated products or whatever

imports/plugins/core/taxes/server/methods/methods.js
```diff
@@ -124,6 +125,36 @@ export const methods = {
...
+   */
+  "taxes/updateTaxCode"(products) {
```
zenweasel: Was there a reason to make this a Meteor method rather than just directly importing the method? If we are going to make it a Meteor method, shouldn't we add some sort of permission check?

Akarshit: @zenweasel This method is being called from the client side code. Shouldn't methods that get called from the client side be "Meteor" methods?
I am adding the permission check.

zenweasel:
@Akarshit Let's fix the conflicts on this and try to get it into either 13 or 14

spencern:
@zenweasel @Akarshit looks like this may need the .snyk file updated from `master` and another review

Akarshit:
@spencern I have merged this with 1.14, let me know if I have to do something else.

spencern:
Just need a 👍 review and for the app tests to pass

zenweasel:
@Akarshit So this is good to go except the tests look like they are failing and they are failing for me locally.

Akarshit:
@zenweasel I didn't change anything and the tests are passing now, for me they are also passing locally.

PR #4184, opened 2018-04-16
Bug Fix: Admins Acting on Behalf of Other Users

pmn4:
Resolves _n/a_
Impact: **minor**  
Type: **bugfix**

## Issue
I noticed a few places where we kick off events using `Meteor.userId()`, however there are cases where an Admin may be acting on behalf of another user, so I _think_ we should be passing the target user's id instead.

## Testing
1. I added an additional expectation to the applicable tests.

### DIFF
--- a/server/methods/accounts/accounts.js
+++ b/server/methods/accounts/accounts.js
@@ -357,7 +357,7 @@ export function addressBookAdd(address, accountUserId) {
           "profile.addressBook.$.isShippingDefault": false
         }
       });
-      Hooks.Events.run("afterAccountsUpdate", Meteor.userId(), {
+      Hooks.Events.run("afterAccountsUpdate", userId, {
         accountId: account._id,
         updatedFields: ["isShippingDefault"]
       });
@@ -371,7 +371,7 @@ export function addressBookAdd(address, accountUserId) {
           "profile.addressBook.$.isBillingDefault": false
         }
       });
-      Hooks.Events.run("afterAccountsUpdate", Meteor.userId(), {
+      Hooks.Events.run("afterAccountsUpdate", userId, {
         accountId: account._id,
         updatedFields: ["isBillingDefault"]
       });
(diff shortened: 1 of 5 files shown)


server/methods/accounts/accounts.js
```diff
@@ -396,7 +396,7 @@ export function addressBookAdd(address, accountUserId) {
     accountsUpdateQuery.$set.name = address.fullName;
   }
 
-  Meteor.users.update(Meteor.userId(), userUpdateQuery);
+  Meteor.users.update(userId, userUpdateQuery);
```
pmn4: I'm not entirely sure why we have this information _both_ in `Accounts` and in `users`, so if this is incorrect, please let me know!

server/methods/accounts/accounts.js
```diff
@@ -356,7 +356,7 @@ export function addressBookAdd(address, accountUserId) {
           "profile.addressBook.$.isShippingDefault": false
         }
       });
-      Hooks.Events.run("afterAccountsUpdate", Meteor.userId(), {
+      Hooks.Events.run("afterAccountsUpdate", userId, {
```
zenweasel: I feel like if we are going to do this, we need to change these for every instance where this hook is run. For instance this same hook is called here:

https://github.com/reactioncommerce/reaction/blob/c30d70a962d42285be5367c66b847331f592191c/imports/plugins/included/connectors-shopify/server/methods/import/customers.js#L173

So we need to go through and change not just these but all of them otherwise you will get different info depending on where it is being called.

server/methods/core/groups.js
```diff
@@ -178,7 +178,7 @@ Meteor.methods({
     try {
       setUserPermissions({ _id: userId }, permissions, shopId);
       Accounts.update({ _id: userId }, { $set: { groups: newGroups } });
-      Hooks.Events.run("afterAccountsUpdate", loggedInUserId, {
+      Hooks.Events.run("afterAccountsUpdate", userId, {
```
pmn4: ^ one more @zenweasel

spencern:
Going to close this - we'll address the root of this problem in 2.0 - namely that it can be confusing which ID is the _acting_ user and which ID is the user that's being acted upon.

PR #4190, opened 2018-04-17
Fix 4173 Fix Simple Schema runtime errors

### LINKED ISSUE #4173: npm simple-schema errors in production v1.10
## Issue Description
The following errors were encountered when running in production:
`Error: "settings.cart.cleanupDurationDays" is in the schema but "settings.cart" is not `
`Error: "settings.public.segmentio.enabled" is in the schema but "settings.public.segmentio" is not`
Other similar errors show once these are fixed.

## Steps to Reproduce
Not sure how to reproduce locally. I was running a custom production image which was dockerized. However, the issue and cause are clear. These lines of code are only accessed in production, and so weren't found during the transition to npm simple-scheme.

## Possible Solution

Add 
```
  "settings.cart": {
    type: Object,
    optional: true,
    defaultValue: {}
  }
```
or similar to the relevant schemes.

## Note

I'm finding and fixes all instances and will create a pull request upon completion.

### Versions
Node: 8.9.0
NPM: 5.7.1
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.10.0
Reaction branch: shipping-by-weight
Docker: 18.03.0-ce

zenweasel:
Resolves #4173 
Impact: **critical**  
Type: **bugfix**

## Issue
The updated version of Simple Schema throws errors where it used to sliently let them pass. Primarily these are about there not being a parent schema entry for objects.

## Solution
This resolves those errors and hard-pins the version at 1.5.0

## Breaking changes
Should be None

## Testing
1. Start the app and observe that there are no errors


zenweasel:
I thought the currency error was related but it appears to not be, so removing WIP

PR #4198, opened 2018-04-18
Remove "Themes" collection

zenweasel:
Resolves #NA
Impact: **minor**  
Type: **chore**

## Issue
While fixing schema errors I noticed that we still had this unused `Themes` collection here that's empty. So I figured I might as well remove it. If we don't want to do this now for some reason just close this PR.

## Solution
Remove it.

## Breaking changes
I don't believe this was ever fully implemented and the Theme Editor was removed a while ago but if you depended on this Collection for anything it will no longer exist.

## Testing
1. Since this was unused you should just verify that this doesn't break anything


spencern:
Looks good to me, but has failing Snyk tests. I'll see what I can do to update those packages or update our .snyk

PR #4203, opened 2018-04-20
feat(GraphQL): uniqueId schema directives

aldeed:
Impact: **minor**  
Type: **feature**

## Issue
We've been implementing very similar logic throughout our resolvers to transform between globally unique GraphQL IDs and our internal, collection-specific Mongo `_id`s. This manual effort slows us down and is not necesary.

## Solution
This PR introduces two new [GraphQL schema directives](https://www.apollographql.com/docs/graphql-tools/schema-directives.html).

Usage for arguments that are of `ID` type:

```
account(id: ID!): Account @uniqueIdArg(namespace: "account", argument: "id")
```

(Include multiple directives if there are multiple arguments that need transformation.

Usage for input and type fields that have an `ID` type value:

```
primaryShopId: ID @uniqueId(namespace: "shop")
```

This is implemented in just two placed in the schema so far. We can switch everything to use this after this is reviewed and merged.

## Breaking changes
None

## Testing
Verify the two IDs that are now using this in the schema are still translating IDs properly.

### DIFF
--- a/imports/plugins/core/graphql/server/resolvers/account/Query/account.js
+++ b/imports/plugins/core/graphql/server/resolvers/account/Query/account.js
@@ -1,5 +1,3 @@
-import { decodeAccountOpaqueId } from "@reactioncommerce/reaction-graphql-xforms/account";
-
 /**
  * @name account
  * @method
@@ -11,6 +9,5 @@ import { decodeAccountOpaqueId } from "@reactioncommerce/reaction-graphql-xforms
  * @return {Promise<Object>} user account object
  */
 export default function account(_, { id }, context) {
-  const dbAccountId = decodeAccountOpaqueId(id);
-  return context.queries.userAccount(context, dbAccountId);
+  return context.queries.userAccount(context, id);
 }
--- a/imports/plugins/core/graphql/server/resolvers/account/Query/account.test.js
+++ b/imports/plugins/core/graphql/server/resolvers/account/Query/account.test.js
@@ -1,7 +1,5 @@
 import account from "./account";
 
-const base64ID = "cmVhY3Rpb24vYWNjb3VudDoxMjM="; // reaction/account:123
-
 const mockAccount = {
   _id: "123",
   name: "Reaction"
@@ -10,7 +8,7 @@ const mockAccount = {
 test("calls queries.userAccount and returns the requested user", async () => {
   const userAccount = jest.fn().mockName("queries.userAccount").mockReturnValueOnce(Promise.resolve(mockAccount));
 
-  const user = await account(null, { id: base64ID }, {
+  const user = await account(null, { id: mockAccount._id }, {
     queries: { userAccount },
     userId: "999"
   });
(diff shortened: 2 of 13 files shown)


ticean:
Hey @aldeed, this is an interesting experiment and it's cool to see how directives are implemented. 

Do you think this really provides more value than a plain function in the resolver for the purpose of a data transformation? 

Functions are chainable, easily testable, and easy to reason about. Some cases will require transformation on more attributes. Is it worth having two mechanisms?

It trades a line of resolver code for a block of schema code. Doesn't seem like a net change aside from the added code to implement the directive. (The tests have hardcoded opaque ID's but since those are integration tests you could use a function to calculate the expected value, if that's slowing things down.)

Not sure if you implemented this to see how it works of it you think we should definitely go in this direction. Directives could be good for some things. Auth, maybe? But is it worth having two mechanisms for a single concern - data transformation?

aldeed:
@ticean I do think this is much easier and it covers all bases, so I will be able to completely remove the existing ID transformation functions from all resolvers. There won't be two ways, just one. I'm just waiting for a next PR to do that so that we can have this debate first.

Regarding resolver vs. schema, I like the idea of making this part of the schema definition rather than resolver implementation, because it is then automatic. We have only a few people writing schemas but potentially many more people involved in implementing them. So this way we don't have to trust the implementer to add ID xforms.

In the end, the main thing I like is that moving all IDs to this approach will result in deleting a bunch of code, which makes it easier for everyone to understand and maintain. I am someone who always takes configuration over code, though, so I want to make sure everyone is on board.

aldeed:
@spencern My understanding is that directives are a draft part of the GraphQL specification. They are in the language and will not confuse anything reading the files, but Apollo has an actual implementation of them whereas some server packages might not. Clients such as graphiql will just ignore them, if they're even part of the published schema.

So long story short, they impact the server only, and if we switched to a different server implementation rather than Apollo, it would not complain about them being in the schema, but we would have to reimplement them with different code to make them work.

ticean:
Hi @aldeed, These PR examples only transform`_id`. But GQL API is not modeled 1 to 1 on the database schema, there will be numerous cases where more transformation is required. A directive will transform the `_id`. But the other attributes will be transformed by xform functions as it is today? Won't we have more code in those cases (the directive *and* the xform)? 

I'm not sold that this is an improvement but you and Spencer seem to like it. Going all-in on this is going to add a lot of directives to the schema.

---

* How are directives composed? It may be possible but I can't find an example that demonstrates it.
* Looking forward to a GraphQL gateway, what are the effects on schema management, stitching, etc? [Here's an issue from graphql-tools](https://github.com/apollographql/graphql-tools/issues/603) explaining that directives aren't merged in a schema merge?

spencern:
@aldeed @ticean This issue has begun to stagnate a bit. Is this still something we'd like to accomplish?

aldeed:
@spencern I think I'd still like it, but I need to get a few of my large pattern-altering PRs merged and then re-evaluate. For now let's keep it open.

aldeed:
Closing this. I will make an NPM package with the directives. We will eventually probably use directives only on inputs because that's where we get the most benefit and won't have any issues with the timing of when the transformation happens.

PR #4206, opened 2018-04-23
Limit Tags Publication to Those for the Current Shop

pmn4:
Resolves: n/a
Impact: **minor/major**  
Type: **bugfix**

## Issue
The `"Tags"` publication exposes all tags from all shops.

## Solution
Scope the Tags query to those of the current shop.

## Breaking changes
If it was intended that the Primary Shop see all tags from all shops, this is a breaking change and can be easily changed to restore that behavior.

## Testing
1. Create a new merchant shop
2. Notice that the Primary Shop's "Shop" tag appears in the Navigation Bar.
3. Merchant Shop Admins are given the impression they can edit, or remove it, but those requests fail without notification (that is perhaps a different issue, but reiterates the need for this change).

### DIFF
--- a/imports/plugins/core/ui/client/containers/tagListContainer.js
+++ b/imports/plugins/core/ui/client/containers/tagListContainer.js
@@ -241,7 +241,7 @@ function composer(props, onData) {
 
   if (props.product) {
     if (_.isArray(props.product.hashtags)) {
-      tags = _.map(props.product.hashtags, (id) => Tags.findOne(id));
+      tags = Tags.find({ _id: { $in: props.product.hashtags } }).fetch();
     }
   }
 
--- a/client/modules/core/main.js
+++ b/client/modules/core/main.js
@@ -607,6 +607,15 @@ export default {
   },
 
   /**
+   * @name isShopPrimary
+   * @summary Whether the current shop is the Primary Shop (vs a Merchant Shop)
+   * @return {Boolean}
+   */
+  isShopPrimary() {
+    return this.getShopId() === this.getPrimaryShopId();
+  },
+
+  /**
    * @name getShopName
    * @method
    * @memberof Core/Client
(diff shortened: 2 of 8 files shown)


imports/plugins/core/ui/client/containers/tagListContainer.js
```diff
@@ -241,7 +241,7 @@ function composer(props, onData) {
 
   if (props.product) {
     if (_.isArray(props.product.hashtags)) {
-      tags = _.map(props.product.hashtags, (id) => Tags.findOne(id));
+      tags = Tags.find({ _id: { $in: props.product.hashtags } }).fetch();
```
pmn4: I found myself doing a `compact` on tags but then realized there are N queries, which could return `null`. Changing to a single query eliminates the `null`s for us.
(this pattern was used in a few places. I replaced them all)

pmn4:
first, I rebased from master and am now seeing the `snyk-security` test fail. any suggestions?
second, @spencern, can you respond to @zenweasel's question above? It will be pretty straightforward to add a check for the primary shop but will wait to hear from you before doing so.

spencern:
@pmn4 and @zenweasel I'm not sure that tags have been fully fleshed out for a marketplace yet and I'm open to suggestions. I thought you would have only seen the tags associated with the active shop, but if that's not the case, I can see where @zenweasel's line of thinking could have ended up with publishing all tags to the primary shop.

Re: snyk issues, @nnnnat looked at those recently, so they may have been resolved since you last merged master / release branch.

pmn4:
re-triggering tests

stunjiturner:
I am still getting tags form shop 2 in the main shop 1 in release 1.14.1

aldeed:
@stunjiturner Can you please submit a new issue with reproduction steps and all the template filled out?

pmn4:
this behavior is expected. I didn't want to introduce a breaking change, and since all products show up in the Primary Shop, so do all tags.

PR #4208, opened 2018-04-24
chore: Use NPM random package

aldeed:
Impact: **minor**  
Type: **chore**

## Issue
In order to demeteorize some code for GraphQL development, we need a non-Meteor version of Meteor's "random" package.

## Solution
Created, published, and switched all imports to an NPM random package with no Meteor dependencies.

## Breaking changes
None

## Testing
Pick a couple places where `Random` is imported and used, and verify that they are still working the same as before.


jshimko:
I'm getting a blank page and this error...

localhost_3000

jshimko:
And that's from the `Random.id()` right here...

https://github.com/reactioncommerce/reaction/blob/chore-aldeed-npm-random/lib/collections/schemas/address.js#L39

aldeed:
@jshimko Sorry, not sure how I didn't see that. Maybe tested on wrong branch. In any case, I fixed the package and bumped the version and it seems to be working for me now.

PR #4228, opened 2018-05-01
Fix GraphQL pagination

### LINKED ISSUE #4227: GraphQL pagination bug
## Prerequisites

## Issue Description
GraphQL queries that take `before` and `last` params do not always properly paginate results.

## Steps to Reproduce
1. Query `catalogItems` for the primary shop ID
2. Query `catalogItems` again using `last: 10` and `before` param set to the `endCursor` from the previous query's `pageInfo`. You will get fewer results than expected.

### Versions
1.10.0

aldeed:
Resolves #4227   
Impact: **minor**  
Type: **bugfix**

## Issue
Incorrect GraphQL pagination

## Solution
- Fixed `getPaginatedResponse` to use the count after before/after is added to the filter, when providing count to the skip/limit function.
- Also switched to using object format for MongoDB `sort` rather than array because `nedb` is expecting object format. This way our integration tests are actually accurate when it comes to testing anything sort-related.

## Breaking changes
None

## Testing
Page forward and backward through `catalogItems` and `tags` GraphQL queries and verify that the results are what you expect.

### DIFF
--- a/imports/plugins/core/graphql/server/resolvers/util/applyBeforeAfterToFilter.js
+++ b/imports/plugins/core/graphql/server/resolvers/util/applyBeforeAfterToFilter.js
@@ -1,5 +1,3 @@
-import { decodeOpaqueId } from "@reactioncommerce/reaction-graphql-xforms/id";
-
 /**
  * @name applyBeforeAfterToFilter
  * @method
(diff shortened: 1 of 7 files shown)


mikemurray:
Re-running tests before merging.

mikemurray:
Tests run and pass on CI and via ssh manually. That state isn't reflected here for some reason. merging...

PR #4235, opened 2018-05-08
Supported methods is optional

### LINKED ISSUE #4234: Paypal update settings throwing error
## Issue Description
When trying to update the PayPal Express setting with PayFlow disables, Reaction throws the following error 

## Steps to Reproduce
1. Login as admin
2. Enable PayPal Express
3. Fill the form, try to save it.
4. Observe the error.

## Possible Cause

This error is happening because Reaction uses Autoform for rendering the form. But only a part of the schema is used to render the form(the other part is used for payflow.)
So when trying to save, Autoform detects that the document it got from the form is missing a "required" field(the supported field of PayFlow) and hence is throwing a error.

Akarshit:
Resolves #4234   
Impact: **critical**  
Type: **bugfix**

## Issue
This error is happening because Reaction uses Autoform for rendering the form. But only a part of the schema is used to render the form(the other part is used for payflow.)
So when trying to save, Autoform detects that the document it got from the form is missing a "required" field(the supported field of PayFlow) and hence is throwing a error.

## Solution
I have made the `supported` array optional. So now when the form is saved for Express and the document without `supported` array for PayFlow is saved, there is no validation error.

## Breaking changes
None

## Testing
1. Login as admin
1. Enable PayPal Express
1. Fill the form, try to save it.
1. Observe no error.

### DIFF
--- a/imports/plugins/included/payments-paypal/lib/collections/schemas/paypal.js
+++ b/imports/plugins/included/payments-paypal/lib/collections/schemas/paypal.js
@@ -23,7 +23,8 @@ export const PaypalPackageConfig = PackageConfig.clone().extend({
   },
   "settings.express.support": {
     type: Array,
-    label: "Payment provider supported methods"
+    label: "Payment provider supported methods",
+    optional: true
   },
   "settings.express.support.$": {
     type: String,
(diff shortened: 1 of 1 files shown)


zenweasel:
@Akarshit I don't know that this is the correct approach, as this array is in fact required for the app to function correctly. Maybe we can just provide a default value instead.

zenweasel:
@Akarshit If you made this change, you didn't push it

zenweasel:
@Akarshit So if I remove the API key and save it doesn't do any validation

zenweasel:
@Akarshit Since this plugin is going to removed in 1.14, let's close this and PR this to the new repo

PR #4238, opened 2018-05-10
Performance/Refactor remove revision control

### LINKED ISSUE #4237: Remove revisions tracking system in favor of published catalog products
1. Remove revisions tracking system from existing meteor app
1. Remove Revisions collection and all code that updates or interacts with data in the Revisions collection.
1. Make sure that product publication still publishes to the catalog

mikemurray:
Resolves #4237   
Impact: **breaking**  
Type: **performance/refactor**

## Issue
Remove the `Revision` control system in favor of the` Catalog` as having both is redundant.

## Solution

Remove all code, hooks, collections, and packages related to revision control.

## Breaking changes

- The revision control system is removed, any custom packages that depend on it will no longer work.
- removed `applyProductRevision` from `lib/api/products.js`
- Cannot undo changes when editing a product
- PDP page will have an instant reflection of changes. Product grid will only see changes upon publishing.

## Testing
1. As an admin
2. Add products, see that the changes show up immediately on PDP for customer and admin
3. Check the product grid has no changes for the customer, but immediate change for admin
4. Publish product, see product grid change for customers
5. Add products to cart and checkout
6. Fulfill orders and verify that inventory updates on the product via the PDP admin tools


spencern:
@aldeed I think we should probably merge this into Release 2.0 and not into 1.12.0

spencern:
Issue that was created to flag a product which has unpublished changes: https://github.com/reactioncommerce/reaction/issues/4264

PR #4243, opened 2018-05-14
Fix #4090: Scrolling to bottom adds more products to the view

### LINKED ISSUE #4090: Products more then 24 are not getting loaded in the productContainerCustomer page
## Issue Description

If we have products more them 24. the first 24 are getting loaded( in the productContainerCustomer page ) and when we scroll down to the bottom of the page we don't find the LoadMoreProducts button and a client console error occurs saying loadProducts is not a function. actually, it is getting undefined and the next 24 products are not being loaded. below is the screenshot of the client console error.

image

## Steps to Reproduce

1. start the reaction

2. log in as administrator and add products more than 24 say( around 30) and publish them.

3. log out and check the productContainerCustomer page. for the first time after scrolling down beyond the 24 products, you can see the rest of all the 30 products without any error.

4.) now if you revisit the page or restart the reaction and visit the productContainerCustomer page and scroll down to view the products beyond 24 you will get the error as shown in the above screenshot and the rest of the products cannot be loaded.

## Possible reason for the Bug
the this.props.loadProducts is getting undefined.

### Versions

Reaction: 1.10.0

Akarshit:
Resolves #4090  
Impact: **major**  
Type: **bugfix**

## Issue
The current logic was checking for equality in the scroll position, the scroll value and the screen height. This usually works but is not true in some cases where the values are in decimal. There is a offset of `1`

## Solution
Instead of equality now the code checks for `+-1`.

## Testing
1. View store as anonymous.
2. Zoom out the browser.
3. Try to scroll, new products should get added.


imports/plugins/included/product-variant/components/customer/productGrid.js
```diff
@@ -24,7 +24,7 @@ class ProductGrid extends Component {
     const { canLoadMoreProducts, loadProducts } = this.props;
     const { scrollY, innerHeight } = window;
     const { body: { scrollHeight } } = document;
-    const atBottom = (innerHeight + scrollY === scrollHeight);
+    const atBottom = (innerHeight + scrollY >= (scrollHeight - 1)) && (innerHeight + scrollY <= (scrollHeight + 1));
```
aldeed: This might be less confusing:

```js
const atBottom = (Math.floor(innerHeight + scrollY) === Math.floor(scrollHeight));
```

imports/plugins/included/product-variant/components/customer/productGrid.js
```diff
@@ -24,7 +24,7 @@ class ProductGrid extends Component {
     const { canLoadMoreProducts, loadProducts } = this.props;
     const { scrollY, innerHeight } = window;
     const { body: { scrollHeight } } = document;
-    const atBottom = (innerHeight + scrollY === scrollHeight);
+    const atBottom = Math.abs(innerHeight + scrollY - scrollHeight) <= 1;
```
Akarshit: @aldeed Would this do?

PR #4250, opened 2018-05-16
[WIP] (refactor) Remove unsupported plugins

zenweasel:
Resolves NA  
Impact: **minor**  
Type: **performance**

## Issue
Reaction is a huge app that takes forever to build. Also we need to reduce the "surface area" of the app that we officially support.

## Solution
This is not the complete solution, but remove code that is not required. These can be moved to separate repos and added as community plugins (which is where many of these originated).

This PR is intended as a starting point for discussion about what should be included and what shouldn't.

## Testing
1. Ensure that the core app functions normally


zenweasel:
Closing in favor of more specific PR's

Issue #4092, opened 2018-03-29
Products created in one shop showing up in all shops


kathiresh-i2i:
## Issue Description
This issue only applies to a marketplace setup.

As a merchant admin, the product grid shows all products available. That is, you can see products from other merchants as well as the marketplace owner's products. You can't edit them though, so there's no direct security issue arising through this bug.

As a user, only those products that belong to the appropriate shop are displayed, which is the expected behaviour, IMHO.
E.g.
- <http://localhost:3000> shows all products from all merchants
- <http://localhost:3000/shop/merchant1> shows all products from merchant 1
- <http://localhost:3000/shop/merchant2> shows all products from merchant 2

## Steps to Reproduce
1. As marketplace owner, enable marketplace and create a product
2. Invite two merchants: merchant1 and merchant2
3. As merchant1 admin, create product
4. Observe that the logged-in merchant 2 can see products from marketplace owner and merchant 1

### Versions
Node: 8.2.1
NPM: 5.3.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.10.0
Reaction branch: master
Docker: 17.03.1-ce

willmoss1000:
I've notice this too. Seems to be that this section of code is not functioning: https://github.com/reactioncommerce/reaction/blob/138b9e84253c3d991327bd98dfbe9dffa89cf7d3/server/publications/collections/products.js#L141-L150

I put a logger either side, and they give the following outputs when viewing the customer grid on the reaction shop page.

Top logger gives: ` [ 'reaction' ] 'shopIdsOrSlugs'`
Bottom logger gives `[] 'shopIds'`

```
 Logger.info(shopIdsOrSlugs, "shopIdsOrSlugs");

  if (shopIdsOrSlugs) {
    // Get all shopIds associated with the slug or Id
    const shopIds = Shops.find({
      "workflow.status": "active",
      "$or": [{
        _id: { $in: shopIdsOrSlugs }
      }, {
        slug: { $in: shopIdsOrSlugs }
      }]
    }).map((shop) => shop._id);

    Logger.info(shopIds, "shopIds");
```

kathiresh-i2i:
# **Issue Description and resolution**

### Products created in one shop showing up in all shops.

1) Logged in to 'Testshop1' as 'Testuser1' 

2) Created a product called 'Created on 'test shop 1' in 'Testshop 1'


[...]

PR #4256, opened 2018-05-18
[4254] Limit Products when in Edit Mode

### LINKED ISSUE #4254: Edit Access to Products for All Shops
## Issue Description
In the Products Publication, we test whether an Admin has the `"createProducts"` role, however we only go as far to say, "does the admin have this role for _any_ shop?".

``` JavaScript
if (editMode && userAdminShopIds && Array.isArray(userAdminShopIds) && userAdminShopIds.length > 0) {
```

This leads to confusion for Merchant Shop owners to whom it appears they can edit products which they do not own. (There _are_ checks in place to prevent them from actually updating those Products).

Part of this problem is that Merchant Shop Owners can currently see products from other shops on their grid, but there is an [Issue](https://github.com/reactioncommerce/reaction/issues/4092) to address this already. Once that is resolved, the issue will still remain for Primary Shop Managers, who can both 1) see all products, and 2) be in edit mode for the Primary Shop.

The suggestion is to show only editable Products in a Grid when the Edit Mode toggle is turned on.

## Steps to Reproduce
1. Add products to the Primary Shop and to a Merchant Shop.
2. Log in as Merchant Shop Owner
3. Visit the Product Grid
4. Turn on Edit Mode
5. Clicks on a product which belongs to the Primary Shop

## Possible Solution
I believe the issue is quickly resolved [here](https://github.com/reactioncommerce/reaction/blob/master/server/publications/collections/products.js#L345-L347):
``` JavaScript
  if (editMode && userAdminShopIds && Array.isArray(userAdminShopIds) && userAdminShopIds.length > 0) {
    selector.isVisible = {
    };
// >>>>>> Here
// >>>>>> instead of filtering by all activeShopIds, we should filter on
//        the intersection of activeShopIds and userAdminShopIds,
//        or perhaps just userAdminShopIds (ignoring status)
    selector.shopId = {
      $in: activeShopsIds
    };

  // ...the method continues...
```

### Versions
``` bash
Node: 8.9.1
NPM: 5.8.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.11.0
Reaction branch: master
Docker: 17.09.0-ce
```

pmn4:
Resolves #4254  
Impact: **major/minor**  
Type: **bugfix**

## Issue
When in Edit Mode, a grid will show all products from all shops**, editable or not.

_** - whether it is correct that all products from all shops should be displayed is a different issue, which I will be addressing shortly._

## Solution
We fetch a list of shops for which you have the `"createProduct"` role, but then return the products from all active shops. This PR changes that to only show the products from the shops where you have the role when `editMode` is `true`.

## Breaking changes
none.

## Testing
1. Add products to the Primary Shop and to a Merchant Shop.
2. Log in as Merchant Shop Owner
3. Visit the Product Grid
4. Turn on Edit Mode
5. Click on a product which belongs to the Primary Shop

### DIFF
--- a/server/publications/collections/product-publications.app-test.js
+++ b/server/publications/collections/product-publications.app-test.js
@@ -15,11 +15,15 @@ Fixtures();
 
 describe("Publication", function () {
   const shopId = Random.id();
+  const primaryShopId = Random.id();
   let sandbox;
 
   beforeEach(function () {
     Collections.Shops.remove({});
-    createActiveShop({ _id: shopId });
+
+    createActiveShop({ _id: shopId, shopType: "merchant" });
+    createActiveShop({ _id: primaryShopId, shopType: "primary" });
+
     sandbox = sinon.sandbox.create();
     sandbox.stub(RevisionApi, "isRevisionControlEnabled", () => true);
   });
@@ -29,6 +33,8 @@ describe("Publication", function () {
   });
 
   describe("with products", function () {
+    let collector;
+
     const priceRangeA = {
       range: "1.00 - 12.99",
       min: 1.00,
@@ -80,6 +86,20 @@ describe("Publication", function () {
         isSoldOut: false,
         isBackorder: false
       });
+
+      Collections.Products.insert({
+        ancestors: [],
+        title: "Garbage Pail Kids",
+        shopId: primaryShopId,
+        type: "simple",
+        price: priceRangeA,
+        isVisible: true,
+        isLowQuantity: false,
+        isSoldOut: false,
+        isBackorder: false
+      });
+
+      collector = new PublicationCollector({ userId: Random.id() });
     });
 
     describe("Products", function () {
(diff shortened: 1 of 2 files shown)


server/publications/collections/product-publications.app-test.js
```diff
@@ -80,6 +86,20 @@ describe("Publication", function () {
...
+        shopId: primaryShopId,
+        type: "simple",
+        price: priceRangeA,
+        isVisible: true,
+        isLowQuantity: false,
+        isSoldOut: false,
+        isBackorder: false
+      });
+
+      collector = new PublicationCollector({ userId: Random.id() });
```
pmn4: all tests did this, so I moved it to a before block

server/publications/collections/product-publications.app-test.js
```diff
@@ -88,14 +108,35 @@ describe("Publication", function () {
...
+          expect(products.length).to.equal(4);
+
+          if (!isDone) {
+            isDone = true;
+            done();
+          }
+        });
+      });
+
+      it("returns products from only the shops for which an admin has createProduct Role", function (done) {
```
pmn4: the diff is a little strange here, but this is a new test.
the user does not have the createProduct role in the "current" shop but does for other.
It is a bit of a contrived example, but I will make it more realistic in the next PR addressing the #4092

server/publications/collections/products.js
```diff
@@ -320,14 +320,6 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
     return this.ready();
   }
 
-  // Get active shop id's to use for filtering
```
pmn4: with the fix, I was able to move this code down, saving a query when in editMode

server/publications/collections/products.js
```diff
@@ -343,13 +335,17 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
...
+      $in: userAdminShopIds
     };
 
     // Get _ids of top-level products
     const productIds = Products.find(selector, {
       sort,
       limit: productScrollLimit
+    }, {
+      fields: {
+        _id: 1
```
pmn4: not relevant to this PR, but since we only care about the `_id` I figured I'd save some data transfer and request only the `_id` field

server/publications/collections/products.js
```diff
@@ -416,14 +412,16 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
...
     sort,
     limit: productScrollLimit
+  }, {
+    fields: {
+      _id: 1
+    }
   }).map((product) => product._id);
 
-  let newSelector = { ...selector };
+  let newSelector = _.omit(...selector, ["hashtags", "ancestors"]);
```
pmn4: both branches of the conditional below perform the `omit` operation, so I moved it here and combined it with the initial spread/clone

server/publications/collections/products.js
```diff
@@ -486,6 +482,18 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
...
   }
 
+  // Get active shop id's to use for filtering
+  const activeShopsIds = Shops.find({
+    $or: [
+      { "workflow.status": "active" },
+      { _id: primaryShopId }
+    ]
+  }, {
+    fields: {
```
pmn4: this is a copy/paste from above with the `fields` added

server/publications/collections/products.js
```diff
@@ -416,14 +412,16 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
...
     sort,
     limit: productScrollLimit
+  }, {
+    fields: {
+      _id: 1
+    }
   }).map((product) => product._id);
 
-  let newSelector = { ...selector };
+  let newSelector = _.omit(selector, ["hashtags", "ancestors"]);
```
pmn4: both branches of the conditional below perform the `omit` operation

PR #4259, opened 2018-05-18
[4092] Bug Fix: Product Visibility for Marketplace Shops

### LINKED ISSUE #4092: Products created in one shop showing up in all shops

pmn4:
Resolves #4092  
Impact: **major**  
Type: **bugfix**

## Issue
Marketplace Shops see Products from all shops.

## Solution
Limit the products that a Marketplace Shop sees to those which are created specifically for it.
One catch is that the Primary Shop should see all products from all shops.

*please note*: this PR is against a branch other than `master`, rather a [PR](#4256) which laid some groundwork for this one.

In addition to this fix, I believe I found a bug when filtering Products for a set of Shop Ids. The test for this behavior was (again, IMO) yielding a false positive. Please see my notes inline.

## Testing
1. Please see issue #4092 for repro steps.

### DIFF
--- a/imports/plugins/included/product-variant/containers/productsContainerCustomer.js
+++ b/imports/plugins/included/product-variant/containers/productsContainerCustomer.js
@@ -84,16 +84,8 @@ function composer(props, onData) {
     window.prerenderReady = true;
   }
 
-  const activeShopsIds = Shops.find({
-    $or: [
-      { "workflow.status": "active" },
-      { _id: Reaction.getPrimaryShopId() }
-    ]
-  }).map((activeShop) => activeShop._id);
-
   const catalogCursor = Catalog.find({
-    "product.type": "product-simple",
-    "shopId": { $in: activeShopsIds }
+    "product.type": "product-simple"
   }, {
     $sort: {
       [`product.positions.${tagIdForPosition}.position`]: 1,
--- a/server/publications/collections/orders-publications.app-test.js
+++ b/server/publications/collections/orders-publications.app-test.js
@@ -47,8 +47,7 @@ describe("Order Publication", function () {
         expect(collections.Orders.length).to.equal(1);
         const shopOrder = collections.Orders[0];
         expect(shopOrder.shopId).to.equal(order.shopId);
-        done();
-      });
+      }).then(() => done(/* empty */)).catch(done);
     });
 
     it("should not return shop orders for a non-admin", function (done) {
@@ -65,8 +64,7 @@ describe("Order Publication", function () {
       const collector = new PublicationCollector({ userId: Random.id() });
       collector.collect("Orders", (collections) => {
         expect(collections.Orders.length).to.equal(0);
-        done();
-      });
+      }).then(() => done(/* empty */)).catch(done);
     });
   });
 });


server/publications/collections/product-publications.app-test.js
```diff
@@ -14,18 +14,37 @@ import { RevisionApi } from "/imports/plugins/core/revisions/lib/api/revisions";
...
+  let primaryShopProductIds;
+  let allProductIds;
+  let activeMerchantProductIds;
+
+  const productScrollLimit = 24;
+
   beforeEach(function () {
+    shopId = Random.id();
+    merchantShopId = Random.id();
+    primaryShopId = Random.id();
```
pmn4: I was occasionally getting a duplicate shop id error, so I moved these to a before block.

server/publications/collections/product-publications.app-test.js
```diff
@@ -14,18 +14,37 @@ import { RevisionApi } from "/imports/plugins/core/revisions/lib/api/revisions";
...
+  let primaryShopId;
   let sandbox;
 
+  let merchantShop1ProductIds;
+  let merchantShop2ProductIds;
+  let primaryShopProductIds;
+  let allProductIds;
+  let activeMerchantProductIds;
+
+  const productScrollLimit = 24;
```
pmn4: DRY'ing a few tests

server/publications/collections/product-publications.app-test.js
```diff
@@ -99,50 +130,58 @@ describe("Publication", function () {
...
 
         collector.collect("Products", 24, undefined, {}, (collections) => {
-          const products = collections.Products;
-
-          expect(products.length).to.equal(4);
+          const productIds = collections.Products.map(p => p._id);
 
-          if (!isDone) {
-            isDone = true;
-            done();
```
pmn4: I had been noticing that on expectation failures the errors in the terminal output cited a timeout, instead of information about the failure. 
this is because `expect` throws an error on failure, skipping the `done()` call.
`PublicationCollector#collect` returns a Promise, so we can use `Promise#then` to yield better error messages by _always_ calling `done`

server/publications/collections/product-publications.app-test.js
```diff
@@ -99,50 +130,58 @@ describe("Publication", function () {
...
-          }
-        });
+          expect(productIds).to.eql(allProductIds);
```
pmn4: compare product ids instead of counts for a stronger test

server/publications/collections/product-publications.app-test.js
```diff
@@ -99,50 +130,58 @@ describe("Publication", function () {
...
+        sandbox.stub(Reaction, "getShopId", () => primaryShopId);
+        sandbox.stub(Roles, "userIsInRole", () => true);
+        sandbox.stub(Reaction, "hasPermission", () => true);
+        sandbox.stub(Reaction, "getShopsWithRoles", () => [shopId]);
+
+        collector.collect("Products", 24, undefined, {}, (collections) => {
+          const productIds = collections.Products.map(p => p._id);
+
+          expect(productIds).to.eql(merchantShop1ProductIds);
+        }).then((_collections) => done(), done);
```
pmn4: not terribly important note:
the lambda is required to be sure that `done` is called with no parameters, indicating success
in addition, the underscore in `_collections` is a Ruby convention to say "this param is ignored". I included it because I didn't want a future refactorer to think, "`() => done()` is the same as `done`, let's change it". making it `.then(done, done)` would pass a value to `done` and result in a failed test

server/publications/collections/products.js
```diff
@@ -147,6 +147,10 @@ function filterProducts(productFilters) {
       }, {
         slug: { $in: shopIdsOrSlugs }
       }]
+    }, {
+      fields: {
+        "_id": 1
+      }
```
pmn4: not required for this bug fix, but since we only care about the _id, we should save some data transfer by only requesting _id from the db

server/publications/collections/products.js
```diff
@@ -347,7 +356,7 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
       $in: [true, false, null, undefined]
     };
     selector.shopId = {
-      $in: userAdminShopIds
+      $in: _.intersection(userAdminShopIds, activeShopsIds)
```
pmn4: for the Primary Shop, this will have little impact, however, for Merchant Shops, owners will only see their products.
Again, probably not a huge impact because Merchant Shop Owners are unlikely to have createProduct roles in other shops.

server/publications/collections/products.js
```diff
@@ -320,17 +324,22 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
...
-      { _id: primaryShopId }
-    ]
-  }, {
-    fields: {
-      _id: 1
-    }
-  }).fetch().map((activeShop) => activeShop._id);
+  // if the current shop is the primary shop, get products from all shops
+  // otherwise, only list products from _this_ shop.
+  if (shopId === primaryShopId) {
```
pmn4: this is the primary bug fix.

server/publications/collections/products.js
```diff
@@ -495,10 +503,17 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
...
+  if (newSelector.shopId && newSelector.shopId.$in) {
+    requestedShopIds = _.intersection(newSelector.shopId.$in, activeShopsIds);
+  } else {
+    requestedShopIds = activeShopsIds;
+  }
+
   newSelector = {
     ...newSelector,
     shopId: {
-      $in: activeShopsIds
```
pmn4: this is the primary bug fix applied to the query

server/publications/collections/products.js
```diff
@@ -494,11 +502,21 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
     });
   }
 
+  // if shops were requested in the product filter, limit the set to those
+  // which are active
```
pmn4: product filters allow for a set of shops. While some work had been done to validate the requested Shops, the filter is ultimately discarded by the `{ $in: activeShopIds }` selector.
This takes the intersection of requested shops and the active shops and returns those products.
the related test was yielding a false positive, which became apparent when I added more shops and products to the tests.

loanlaux:
I confirm that this is a problem on marketplace projects. Thanks a lot for the fix @pmn4!

server/publications/collections/products.js
```diff
@@ -320,24 +324,34 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
...
 
   const selector = filterProducts(productFilters);
 
   if (selector === false) {
     return this.ready();
   }
 
+  // if shops were requested in the product filter
+  if (selector.shopId && Array.isArray(selector.shopId.$in)) {
+    activeShopsIds = _.intersection(selector.shopId.$in, activeShopsIds);
```
pmn4: product filters allow for a set of shops.
While some work had been done to validate the requested Shops, we wound up filtering with `{ $in: activeShopIds }`, discarding the requested shops. this restores that behavior

imports/plugins/core/catalog/server/methods/catalog.app-test.js
```diff
@@ -3,13 +3,11 @@
 import { Random } from "meteor/random";
 import { expect } from "meteor/practicalmeteor:chai";
 import { sinon } from "meteor/practicalmeteor:sinon";
-import { Roles } from "meteor/alanning:roles";
```
pmn4: a little bit of code cleanup in this file, but no changes to the tests within

imports/plugins/included/product-variant/containers/productsContainerCustomer.js
```diff
@@ -89,17 +89,9 @@ function composer(props, onData) {
     window.prerenderReady = true;
   }
 
-  const activeShopsIds = Shops.find({
```
pmn4: this is all handled server-side now

server/publications/collections/orders-publications.app-test.js
```diff
@@ -65,8 +64,7 @@ describe("Order Publication", function () {
       const collector = new PublicationCollector({ userId: Random.id() });
       collector.collect("Orders", (collections) => {
         expect(collections.Orders.length).to.equal(0);
-        done();
-      });
+      }).then(() => done(/* empty */), done);
```
pmn4: putting `done()` after an `expect` statement yields a timeout error when the expectation failed. this isn't super useful, so use the Promise provided by `PublicationCollector#collect` to show better error messages.

server/publications/collections/product-publications.app-test.js
```diff
@@ -98,51 +132,71 @@ describe("Publication", function () {
...
+      merchantShop1VisibleProductIds = [productId2, productId3];
+      activeShopProductIds = [productId1, productId2, productId3, productId4, productId5];
+      activeShopVisibleProductIds = [productId2, productId3, productId4, productId5];
+      activeMerchantProductIds = [productId2, productId3, productId4];
 
       collector = new PublicationCollector({ userId: Random.id() });
     });
 
     describe("Products", function () {
-      it("should return all products to admins", function (done) {
```
pmn4: the diff is working a little too hard here. I replaced the "should return all products to admins" test with a few tests to get a little more granular

server/publications/collections/product-publications.app-test.js
```diff
@@ -152,166 +206,195 @@ describe("Publication", function () {
...
         sandbox.stub(Roles, "userIsInRole", () => false);
 
-        let isDone = false;
-
         collector.collect("Products", 24, undefined, {}, (collections) => {
           const products = collections.Products;
           const data = products[0];
           const expectedTitles = ["Fresh Tomatoes", "Shopkins - Peachy"];
 
-          // the correct expectation should be 2, but there is an issue where
```
pmn4: ignore this comment.
this PR is not against master, but a branch I created to ensure that admins see only the shops/products for which they have edit rights.

server/publications/collections/products.js
```diff
@@ -308,47 +341,24 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
...
-  // Get a list of shopIds that this user has "createProduct" permissions for (owner permission is checked by default)
-  const userAdminShopIds = Reaction.getShopsWithRoles(["createProduct"], this.userId);
-
-  // Don't publish if we're missing an active or primary shopId
-  if (!shopId || !primaryShopId) {
-    return this.ready();
-  }
-
-  // Get active shop id's to use for filtering
-  const activeShopsIds = Shops.find({
```
pmn4: moved this work up to the `filterProducts` methods so that it can be shared by both the `"Products"` and `"Products/grid"` publications.

server/publications/collections/products.js
```diff
@@ -308,47 +341,24 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
...
-  }).fetch().map((activeShop) => activeShop._id);
-
   const selector = filterProducts(productFilters);
 
   if (selector === false) {
     return this.ready();
   }
 
+  // Get a list of shopIds that this user has "createProduct" permissions for (owner permission is checked by default)
+  const userAdminShopIds = Reaction.getShopsWithRoles(["createProduct"], this.userId);
```
pmn4: this was simply moved down a few lines to prevent a query if we return early above.

server/publications/collections/products.js
```diff
@@ -308,47 +341,24 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
...
   // userAdminShopIds is a list of shopIds that the user has createProduct or owner access for
-  if (editMode && userAdminShopIds && Array.isArray(userAdminShopIds) && userAdminShopIds.length > 0) {
+  if (editMode && Array.isArray(userAdminShopIds) && userAdminShopIds.length > 0) {
     selector.isVisible = {
       $in: [true, false, null, undefined]
     };
-    selector.shopId = {
-      $in: userAdminShopIds
-    };
+    selector.shopId.$in = _.intersection(selector.shopId.$in, userAdminShopIds);
```
pmn4: the intersection of "accessible shops", "requested shops", and "shops for which I am an admin"

server/publications/collections/products.js
```diff
@@ -494,14 +503,6 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
     });
   }
 
-  // Adjust the selector to include only active shops
-  newSelector = {
```
pmn4: this is taken into consideration in `filterProducts` above

server/publications/collections/products.js
aldeed: Are you sure we want to see products that can't be edited when `editMode` is on? Might that be confusing? Or we just publish them all and let the client UI indicate which are editable? @spencern or @rymorgan

server/publications/collections/products.js
```diff
@@ -308,47 +341,24 @@ Meteor.publish("Products", function (productScrollLimit = 24, productFilters, so
...
+
   // We publish an admin version of this publication to admins of products who are in "Edit Mode"
   // Authorized content curators for shops get special publication of the product
   // with all relevant revisions all is one package
   // userAdminShopIds is a list of shopIds that the user has createProduct or owner access for
-  if (editMode && userAdminShopIds && Array.isArray(userAdminShopIds) && userAdminShopIds.length > 0) {
+  if (editMode && Array.isArray(userAdminShopIds) && userAdminShopIds.length > 0) {
     selector.isVisible = {
       $in: [true, false, null, undefined]
     };
```
aldeed: @pmn4 In my PR I change this to `delete selector.isVisible;`, which results in a better performing Mongo query.

But if you keep the intersection of `selector.shopId.$in` line below (see my comment there), then I would think we actually want to publish hidden products ONLY for `userAdminShopIds`, but still only publish visible products for all other shops.

aldeed:
I generally agree with all the publication code changes, except for the couple comments I made. I will copy the revised tests into my branch for #4260 to make sure this still works in 2.0 track. But I will let @spencern or @zenweasel test and sign off since I'm not that familiar with how marketplace is supposed to work.

@pmn4 Which branch was this originally from / which branch should we put this in? I'm thinking we include it with release-1.12.0

pmn4:
this PR compares two of my branches which were both created from master a few days ago. they should be fairly up-to-date

server/publications/collections/products.js
pmn4: the current behavior on master is to show them all, which I agree, is confusing.
this line will omit shops for which you do not have edit rights.
I think we are saying the same thing.

further, I think your point about deleting `isVisible` is correct here too

nnnnat:
@pmn4  a `reaction reset` will usually resolve the issues with updated GQL schemas.

server/publications/collections/products.js
```diff
@@ -165,33 +185,46 @@ function filterProducts(productFilters) {
...
+  return selector;
+}
+
+function filterProducts(productFilters) {
+  // if there are filter/params that don't match the schema
+  // validate, catch except but return no results
+  try {
+    if (productFilters) filters.validate(productFilters);
+  } catch (e) {
+    Logger.debug(e, "Invalid Product Filters");
```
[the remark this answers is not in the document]
pmn4: thanks for the review @zenweasel.
the diff here is a little strange. code was inserted above, but I did not edit this method. Let me know whether you'd like that `Logger.debug` to be a `Logger.error/warn` and I will change it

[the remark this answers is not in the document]
zenweasel: I guess just a warning?

server/publications/collections/products.js
```diff
@@ -144,17 +144,37 @@ function extendSelectorWithVariants(collectionName, selector, productFilters, pr
   return newSelector;
 }
 
-function filterProducts(productFilters) {
```
pmn4: this file's diff looks more complex than it should. the summary of changes is that I pulled out the shop selection logic from both the "Products" and "Products/grid" methods into a common place. Doing so allowed for some code cleanup in each of the methods, which accounts for the reset of the diff

aldeed:
@pmn4 Since this didn't make it into 1.12.0, I think we may be able to close it. I made your same changes in https://github.com/reactioncommerce/reaction/pull/4260, which will be merged into 1.13.0. I copied your new/changed tests into that PR, too, so everything should be correct. The only difference is that all revisions support is removed from mine because 1.13.0 will no longer support revisions tracking.

If https://github.com/reactioncommerce/reaction/pull/4260 looks like it resolves this issue for you, you can close this.

pmn4:
sounds good to me @aldeed. thanks for the support on this

pmn4:
the diff is very small now, and despite most of it already included in 1.13.0, I think this should still be merged.

aldeed:
@pmn4 I agree we can merge this, but you will need to change those test promises to use `.catch(done)` to make our new eslint rules happy

spencern:
@aldeed it looks like this PR has been updated to make the eslint rules happy. Passing all tests. Are we good to merge into 1.14.0?

aldeed:
@spencern Yep, merged. Thanks @pmn4 !

PR #4265, opened 2018-05-23
feat: update to Meteor 1.7

### LINKED ISSUE #4239: Update to Meteor 1.7
## Work
Meteor 1.7 is in RC now. Update to it.

- Do `meteor update`
- Update the docker-compose mongo database to the same version as used by Meteor 1.7.
- Update the `mongodb` npm package to the same pinned version that Meteor 1.7 includes.

This should also fix #3861 

## Testing
1. Make sure there are no errors on startup when using `reaction` command to run.
1. Make sure there are no errors on startup when running with docker-compose.
1. Test all of the main test scripts. Make sure no errors appear on client or server.

aldeed:
Resolves #4239   
Impact: **minor**  
Type: **feature**

Updated to Meteor 1.7

## Breaking changes
Probably none, but if you've added extra Babel packages, be sure to update them to the same version as all the included Babel packages.

## Testing
1. Make sure there are no errors on startup when using `reaction` command to run.
1. Make sure there are no errors on startup when running with docker-compose.
1. Run `npm test`. Make sure no errors appear on client or server and all tests pass.


spencern:
Tested with a freshly reset app.
Initial startup seemed faster, but didn't benchmark as @aldeed warned me about running with the profiler on.

Server restarts took ~45s
Client restarts took ~30s

I'm seeing this warning on every restart, for some reason it's getting shown twice. Happens on both client and server changes.
```
app/client/plugins.less.css: warn: There are some @import rules those are not taking effect as they are required to be in the beginning of the file.
app/client/plugins.less.css: warn: There are some @import rules those are not taking effect as they are required to be in the beginning of the file.
```

spencern:
Here's a gist profiling the startup for a previously initialized Reaction application:
`METEOR_PROFILE=1 reaction --inspect`
https://gist.github.com/spencern/9728f36c8f4dc4b8e83a80d5e0c75416

spencern:
Filtered profile from an initial startup:
https://gist.github.com/spencern/6cc9999fed6fea58b3d6eec6dcbd32ad

Filtered profile from a pre-initialized startup: 
https://gist.github.com/spencern/586e5dbf9c49c6a30c2626b435a1415c

aldeed:
This is now blocked by https://github.com/meteor/meteor/issues/9919

aldeed:
I switched to a fork of inline-import pkg, and that has resolved the blocking error. I removed the symlinks that are no longer necessary.

This is working for me now, and I'm seeing restarts consistently in the 70-90 seconds range. Initial startup avgs around 4.5 minutes.

spencern:
@aldeed this may be the issue we're having with the test failure here. I noticed this line getting called out in the test failure stack trace: https://github.com/reactioncommerce/reaction/blob/release-1.13.0/server/publications/collections/reactiveAggregate.js#L8

aldeed:
Yep, definitely. I'll try out a fix

aldeed:
@spencern There is a published version of inline-import plugin with my fix, and switching to that resolved the problems I was having with starting devserver in Docker. I also bumped the Node version in the Dockerfiles to match what Meteor 1.7 uses.

If tests pass and everything looks good to you, this may be ready to merge. I didn't try a prod build.

spencern:
Thanks @aldeed, I'll do a production build and then merge if it all goes well.

spencern:
@aldeed Seeing this issue when running our docker-compose demo to test production: https://github.com/meteor/meteor/issues/9928 but able to resolve it with a more specific ROOT_URL setting. I'll push up a change and then I think we're good.

PR #4266, opened 2018-05-23
(refactor): 4263 nnnnat non meteor schemas

### LINKED ISSUE #4263: Make SimpleSchema definitions work without Meteor
## Work

Move/change the lib/collections/schemas files so that we can get access to schemas in devserver/Jest tests without a Meteor context.

nnnnat:
Resolves #4263 
Impact: **breaking**  
Type: **refactor**

## Issue
Move/change the lib/collections/schemas files so that we can get access to schemas in devserver/Jest tests without a Meteor context.

## Solution
 1. Move schemas out of `lib/collections/schemas` and into `imports/collections/schemas`
 2. Removed `meteor/check` and `meteor/Tracker` from each schema contructor.
 3. Removed `shopIdAutoValue`, `shopIdAutoValueForCart` and `shopDefaultCountry` autoValue functions from schemas in favor of explicitly setting those values on collection `insert`.
 4. Updated `lib/collections/schemas/index.js` to import the non-meteor schemas and add the `meteor/check` and `meteor/Tracker` to the schema's constructor for use inside the meteor application.
 5. Removed the use of `shopIdAutoValue` from plugin schemas and updated usage to provide a `shopId` on collection `insert`
      1. Discounts
      2. Taxes
      3. EmailTemplates
      4. Shipping-shipo
  6. Refactor `server/api/core/utils.js getSlug` method to just import transliteration by default and removed the `autoValue` functions being used in schema constructors in favor of setting the `slug/handle` on collection inserts.
  7. Updated schema import paths for several plugins to just import from the `/lib/collections/schema/index.js` and not the schema files. 

## Breaking changes
The schemas have moved, we're still importing them inside the `lib/collections/schemas/index.js` to apply the `meteor/check` and `meteor/Tracker` to the schemas when being used with meteor. I don't believe this is a breaking change necessarily but I wanted to outline it.

Aside from `createdAtAutoValue` and `updatedAtAutoValue` we're moving away from using autoValues in the schemas in favor of providing required properties. `shopIdAutoValue`, `shopIdAutoValueForCart` and `shopDefaultCountry` autoValue helper functions will no longer be available.

If you have custom plugins that are inserting into any collection that has a `shopId` property, and you rely on `shopId` being automatically set, you will now have to explicitly set it yourself in the inserted object.

## Testing
*You may need to rebuild your docker image for this branch to work*
1. Spin up a fresh RC shop, app should work as it always has.
2. Run the app test make sure everything is passing


imports/collections/schemas/index.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * Reaction uses {@link https://github.com/aldeed/simple-schema-js SimpleSchema} to apply basic content and structure validation to Collections.
+ * See {@link https://docs.reactioncommerce.com/reaction-docs/master/simple-schema full documentation}.
+ * @typedef {SimpleSchema} SimpleSchema
+ * @summary SimpleSchema for Collections
+ * @namespace Schemas
+ */
+
+/**
+ * Reaction uses {@link https://github.com/aldeed/simple-schema-js SimpleSchema} to apply basic content and structure validation to Collections.
```
kieckhafer: Do we need both of these jsdoc blocks?

server/imports/fixtures/products.js
```diff
@@ -212,10 +213,13 @@ export default function () {
     max: 12.99
   };
 
+  const productTitle = faker.commerce.productName();
+
   const product = {
-    title: faker.commerce.productName(),
+    title: productTitle,
```
kieckhafer: All the other fields are being populated on the fly here, why move the productTitle into it's own `const`?

imports/collections/schemas/index.js
nnnnat: I think so, I believe @aldeed  did this when he cleaned up the jsdoc a few months back

server/imports/fixtures/products.js
kieckhafer: Nevermind... didn't see it was being used in the `handle`

PR #4270, opened 2018-05-24
feat: add Media to Tag Schema and GraphQL query

kieckhafer:
Resolves https://github.com/reactioncommerce/reaction-next-starterkit/issues/79
Impact: **minor**  
Type: **feature**

## Issue
We need to update the `Tag` schema to allow for a `media` object, which will be used to provide a hero style image to a tag product grid in the [Reaction Next Starterkit](https://github.com/reactioncommerce/reaction-next-starterkit)

## Solution
- Add an optional `media` blackbox object to the `Tag` schema
- Provide `media` data in the tag GraphQL endpoint

## Breaking changes
n/a

## Testing
- Add a media object to any existing tag, in the `Tag` collection
    - There is no UI to add an image, so the easiest way to do this is to upload an image to a product, publish that product, find the product in the `Catalog` collection, then copy the `product.media` object over to the tag in `Robo3T`. Only copy a single object, not an array of objects.
- Use localhost:3000/graphqli to query the data. Query both a `tag-slug` and a tag `id`, both should work.
```
query {
  tag(slugOrId: {tag-slug OR tagId}) {
    _id
    media {
          URLs {
            large
          }
        }
  }
}
```


ticean:
+1 to @aldeed'recommendation of appending `Url` to the media attribute. I'd suggest some further refinement to designate that it's a `heroMediaUrl`. I don't think it's safe to assume that hero will be the only media attachment or that it should get top billing as the main media.

ticean:
Of course, I mentioned the URL attribute but I recommend the same `hero` modification the the underlying data model, not just the grapqhl API. 😁

kieckhafer:
@aldeed @ticean updated, ready for another look.

PR #4276, opened 2018-05-25
(feat): 4246 nnnnat new mocks factory

### LINKED ISSUE #4246: Create a non-Meteor factory/fixtures solution
## Goal
Remove the large similar objects we've been copy-pasting in all of the integration tests (and some Jest unit tests). Instead use a factory pattern to get fake data.

## Proposal
Here is my proposal for API, but feel free to alter this as necessary to get everything we need to keep our tests simple.

Make it an NPM package (maybe `@reactioncommerce/data-factory`?)
Depend on https://www.npmjs.com/package/simpl-schema-mockdoc

In the reaction files where schemas are defined, do something like:

```js
import { createFactoryForSchema } from "@reactioncommerce/data-factory";
createFactoryForSchema("Shop", ShopSchema);
```

This will add `Shop` property to a `Factory` object that can be imported in tests, in devtools, anywhere.

```js
import { Factory } from "@reactioncommerce/data-factory";

const shop = Factory.Shop.makeOne({ prop: val });
const shops = Factory.Shop.makeMany(5, { prop: val });
```

`makeOne` returns a single object to be inserted/used, while `makeMany` returns multiple in an array (array length = first arg value).

The `{ prop: val }` object is where you specify override properties you specifically want. For any properties not there, or if you omit that object, you'll get back the object created by the `simpl-schema-mockdoc` package.

If `simpl-schema-mockdoc` does not meet our needs or is buggy, we can copy the relevant code into this data-factory package and update it to work for us.

## Work
1. Create the package
1. Add the package to Reaction, pass in all the schemas.
1. Change all of the integration tests in the `/tests` folder to use this to build their mock data objects.

nnnnat:
Resolves #4246   
Impact: **minor**  
Type: **feature|refactor**

## Issue
Writing GQL integration test has become difficult because we've been creating mock data ad-hoc. 
More info [here](https://github.com/reactioncommerce/reaction/issues/4246)

## Solution
 1. Create a new `@reactioncommerce` npm package that can take the de-meteorized `SimpleSchema` and create a mocks factory for each schema. See [data-factory](https://github.com/reactioncommerce/data-factory)
 2. Create a new `test-utils/helper` that will build a mocks factory using the [data-factory](https://github.com/reactioncommerce/data-factory) package and export a `Factory` object to be used in the integration test.
3. Removing some leftover meteor deps from the schemas that we missed in PR [4266](https://github.com/reactioncommerce/reaction/pull/4266).
4. Replacing Tags integration test mock data with mocks from the test utils `Factory`.

## Breaking changes
1. Possible breaking change: RC's product variant schema still had a meteor dep inside one of the properties. This is being removed/refactored to not use the `ReactionProduct` API from `lib/api`

## Testing
1. Spin up a fresh RC shop confirm it still works.
2. Create, edit and archive a new Product
2. Run integration test `docker-compose run devserver npm run test:integration`


nnnnat:
@aldeed  I've made the requested changes and it's ready for review.

PR #4285, opened 2018-05-31
chore: run snyk when package.json has changed or base is master

spencern:
**Summary**
Snyk flags packages that have security vulnerabilities. Our process for resolving these dependencies involves a number of potential resolution steps: updating dependency versions, creating PRs to resolve security issues within dependencies, creating issues to notify package owners of potential security issues, and sometimes ignoring the issue if nothing can be done to resolve the issue at the time. 

When we ignore an issue, we set a expiration date on the ignore that will require us to address the issue again when the ignore has expired. 

Once an ignore expires, every PR in that repository will fail until the snyk issue has been resolved. This is undesirable because we would rather deal with snyk failures where they are introduced in the case of a PR adding a new dependency with a vulnerability, or in a separate PR to a release branch when the snyk failure is caused by the expiration of an ignore.

This PR adjusts the CircleCI config file to run snyk tests conditionally. 
- When a commit does not have a connected PR, snyk should run.
- When a commit is connected to a PR to the `master` branch, snyk should run.
- When a commit is connected to a PR that includes a change to package.json from the base branch into which the PR is requesting to be merged, snyk should run.
- When a commit is connected to a PR and does not include a change to package.json, snyk should not run.

**Docker**
This PR installs a new binary [jq](https://github.com/stedolan/jq/) into the `meteor-dev` stage of our Dockerfile. `jq` makes it simple to work with JSON formatted data within a Unix shell environment. I'm using it to determine the base branch of a PR so that we can accurately determine whether a given PR has made changes to`package.json`

**Testing**
I've made a few commits in this PR that should demonstrate that this circle config is correctly determining the base/target branch of a PR and that the bash script to conditionally run snyk is working correctly. You could also copy this code into your own branch and push it to github to trigger a CI process and see if it's running snyk in the proper circumstances. I've also `ssh`ed into a running CI instance to ensure that `jq` is correctly installed and set up and that the script correctly identifies the PR number and base/target branch for a PR.

Other testing is welcomed, though I'm unsure how else to best test a PR like this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,9 @@ ENV PATH $PATH:/home/node/.meteor
 COPY --chown=node package-lock.json $APP_SOURCE_DIR/
 COPY --chown=node package.json $APP_SOURCE_DIR/
 
+RUN curl -L "https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64" -o /home/node/jq
+RUN chmod +x /home/node/jq # Make jq executable so that we can run it against the
+
 # Because Docker Compose uses a named volume for node_modules and named volumes are owned
 # by root by default, we have to initially create node_modules here with correct owner.
 # Without this NPM cannot write packages into node_modules later, when running in a container.
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,12 +10,6 @@ ENV PATH $PATH:/home/node/.meteor
 COPY --chown=node package-lock.json $APP_SOURCE_DIR/
 COPY --chown=node package.json $APP_SOURCE_DIR/
 
-
-# # Install jq, a lightweight and flexible command-line JSON processor.
-# We use it to determine the base branch of a PR so we can run diffs to conditionally run circle workflows
-RUN curl -L "https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64" -o /home/node/jq
-RUN chmod +x /home/node/jq
-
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -10,6 +10,12 @@ ENV PATH $PATH:/home/node/.meteor
 COPY --chown=node package-lock.json $APP_SOURCE_DIR/
 COPY --chown=node package.json $APP_SOURCE_DIR/
 
+
+# # Install jq, a lightweight and flexible command-line JSON processor.
+# We use it to determine the base branch of a PR so we can run diffs to conditionally run circle workflows
+RUN curl -L "https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64" -o /home/node/jq
+RUN chmod +x /home/node/jq
+
```
ticean: What do you think about removing the `jq` addition in project Dockerfile? Doesn't look like it's required. The CI script is running in the CircleCI build environment - there's no `docker run ...` in the command.

Unless you found otherwise, the script will run using the `circleci/node:8-stretch` image, as inherited from the default config. It already contains `jq`. You can verify that with 

```sh
docker pull circleci/node:8-stretch
docker run --rm circleci/node:8-stretch jq -h
```

spencern: I did not check to see if jq was in `node:8-stretch` if that's the case, then we shouldn't need to install `jq` here. I'll remove it.

Issue #4202, opened 2018-04-20
Unable to update shop name, Promise.await is not a function


nnnnat:
## Prerequisites
* [x] Are you running the latest version?
* [x] Are you able to consistently reproduce the issue?
* [x] Did you search the issue queue for existing issue? [Search issues](https://github.com/reactioncommerce/reaction/issues)

## Issue Description
`TypeError: Promise.await is not a function` in browser console when saving updated shop name. I believe this same issue is effecting the product handle and possibly other fields in the UI.

## Steps to Reproduce
1. Start a fresh RC shop and log in to RC as a shop admin
2. Open the shop panel, update the shop name and click the save button
3. See that the button disables
4. Close the shop panel and open it again
5. See the shop name has been reset to "Reaction"

## Possible Solution
I think this might be a regression from [PR#4049](https://github.com/reactioncommerce/reaction/pull/4049). 
Trying to lazy load the slugify/transliteration has been really tricky mainly because there are 3 versions of the same `getSlug` helper method. One that only runs on the server, one that only runs on the client and another that tries to run in both. It also seems that different parts of the UI use the different `getSLug` methods (tag nav is using `getSlug` from the `client` dir while shop name use the `getSlug` method inside of the `lib` dir).

I feel like there are two main issues to tackle
1. Creating a `getSlug` method that:
    *  takes a string as a param.
    *  can resolve/await the lazy loading of slugify/transliteration.
    *  returns the slugified string.
2. Consolidating `getSlug` usage:
    *  to a single `getSlug` method inside of `lib` that can be used by both server and client.
    *  to a server and client versions of the method and change every import of the old `lib` version.

### Versions
```sh
Node: 9.8.0
NPM: 5.8.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.11.0
Reaction branch: master
Docker: 18.03.0-ce
```

tizzle:
I'm experiencing the same problem when trying to change e.g. timezone in localization/i18n settings.

mikepuglisi:
Same Problem. Fresh install on linux mint using the reaction-cli and I can't update the store name without receiving the `Promise.await is not a function` javascript error.

JacobGadawski:
Hi, I have same problem. I create new reaction project on fresh linux mint and I can't update store and localisation settings.

loanlaux:
[...]

PR #4200, opened 2018-04-18
feat: 4106 nnnnat catalogItem GQL query

nnnnat:
Resolves #4106   
Impact: **minor**  
Type: **feature**

## Issue
**Summary**
Create a query to retrieve a single catalog product by the provided slug or ID.

**Implementation**
The catalogProduct query should have the following signature

```
catalogProduct(slugOrId: String): CatalogItemProduct
```

See https://github.com/reactioncommerce/reaction/issues/4105 for schema detail.

**Acceptance Criteria**
Using GraphQL, I can use the catalogProduct(slugOrId: String): CatalogItemProduct query to get a single CatalogProduct by it's slug.

## Solution
* Created a Catalog Collection Query that returns the product from the catalog collection
* Created a catalogItemProduct GQL query that returns a CatalogItemProduct object
* Created unit test for both of the new query functions
* Created integration test for the CatalogItemProduct GQL query. 

## Breaking changes
N/A

## Testing
1. Spin up a fresh RC shop, login as an admin and create a few products and publish them to the catalog.
    *  **Note:** if you use the reaction devtools to speed products you'll need to update your RC shop name to be "reaction swag shop" or you'll get an error from reaction. You may need to change the shop name from the DB because of this (#4202) issue.
2. From graphiql or localhost:3030/graphiql query a product by slug with `catalogItemProduct(slugOrId: "example-product") { _id }` and copy the returned ID
3. Try the query again this time with the ID instead of the slug `catalogItemProduct(slugOrId: "opaqueID") { _id product { slug } }`
4. The same product should return.

PR #4287, opened 2018-05-31
Release 1.12.0

### LINKED ISSUE #4090: Products more then 24 are not getting loaded in the productContainerCustomer page

### LINKED ISSUE #4106: Create a catalogProduct query in graphQL
**Summary**
Create a query to retrieve a single catalog item for the provided handle (or ID?)

**Implementation**
The catalogProduct query should have the following signature

```
catalogProduct(handle: String!): CatalogProduct
```

The `CatalogProduct` type:
```
_id: ID!
contentType: String!
media: MediaConnection

/** From Product **/
ancestors: [ID] // Ordered array of ancestors  - should this be a ProductConnection? 
changedHandleWas ???? - added a little while back, I dont think we need this for queries
createdAt: DateTime
description: String
facebookMsg: String
googleplusMsg: String
handle: String
hashtags: TagsConnection
isBackorder: Bool (should this be transformed? inventoryStatus: String!)
isDeleted: Bool
isLowQuantity: Bool
isSoldOut: Bool
isVisible: Bool (Transform?)
metaDescription: String (for SEO purposes)
metafields: [Metafield]
originCountry: String/Enum?
pageTitle: String [effectively subtitle] (Page/SEO/Meta Title could be useful too, though used differently)
parcel: Parcel
pinterestMsg: String
positions: Position // was ProductPosition
price: DenormalizedPrice!
productType: String! [Currently appears unused, but would be a better fit than `type`]
publishedAt: DateTime Doesn't appear to be used currently
publishedScope: ?? Doesn't appear to be used? @mikemurray?
requiresShipping: Bool (Transform to "fullfillmentProcess: String!"?)
shop: Shop! // Was shopId
template: String!
title: String!
twitterMsg: String
type: ?? Diff between this and productType? Currently used for `simple`, `variant`, etc
updatedAt: DateTime!
vendor: String
workflow: Workflow

/** From Catalog **/
variants: ProductVariant
```

The `ProductVariant` type
```
_id: ID!
ancestors: [ID]
barcode: String
compareAtPrice: Number
createdAt: DateTime!
eventLog: [Event]  // Should synchronize on a unified way to log events for products, orders, carts, etc
fulfillmentService: String
height: Number
index: Int // Position in variant list
inventoryManagement: Bool // True means we track inventory, Should consider string here to permit multiple types of inventory mgmt
inventoryPolicy: Bool // True means, disallow backorders, should consider a string here for multi option policy
inventoryQuantity: Number // How should we connect this with inventoryService? 
isDeleted: Bool
isLowQuantity: Bool
isSoldOut: Bool
isVisible: Bool (Transform?)
length: Number
lowInventoryWarningThreshold: Number // Warn of low inventory at this number
metafields: [Metafield]
minOrderQuantity: Number // optional, not sure if it's baked into the UI right now or not
optionTitle: String // Super confusing, I'd love to rearchitect this - Option internal name, default value: `"Untitled option"`
originCountry: String/Enum?
price: Number!
shop: Shop! // was shopId
sku: String
taxable: Boolean // Could be string?
taxCode: String // default value: `"0000"`
taxDescription: String
title: String // Label for customers, default value: `""`
type: // default value: `"variant"` - really is product type in some ways. Is tied into Product Schema currently
updatedAt: DateTime
weight: Number
width: Number
workflow: Workflow
```

The `DenormalizedPrice` type
```
min: Number!
max: Number!
range: String!
```

The `Position` type
```
tag: String
position: Number
pinned: Bool
weight: Number
updatedAt: DateTime!
```

**Acceptance Criteria**
Using GraphQL, I can use the CatalogProduct(handle: String!) query to get a single CatalogProduct by it's handle

### LINKED ISSUE #4186: [GraphQL] Fix default `first` value for pagination
## Issue
It seems that the default of 50 for `first` param for paginated connections got lost at some point. All items are returned if you don't include the `first` or `last` param.

## Solution
Write integration tests for this on one of the connection queries, and then fix the `getPaginatedResponse` function.

## Testing
1. When I query for a connection that accepts a `first` argument, and I don't include `first` or `last`, I get only the first 50 results and `hasNextPage` is correctly set.
1. When I do include `first` or `last`, they are respected but if I ask for more than 50, I only get 50.

### LINKED ISSUE #4245: Add price sorting to catalogItems GraphQL query
## Goal
Ability for a GraphQL client to request `catalogItems` sorted by `minPrice` either ascending or descending.

## Work
1. In the GraphQL schema, add `minPrice` to `CatalogItemSortByField` enum.
1. In the GraphQL schema, add `sortByPriceCurrencyCode` argument to the `Query.catalogItems` field.
1. In `/imports/plugins/core/graphql/server/resolvers/catalog/Query/catalogItems.js`, before `getPaginatedResponse` is called, adjust the actual sort-by field name to be `product.pricing.<sortByPriceCurrencyCode argument value>.minPrice`

## Add Integration Tests
1. When I query `catalogItems` passing a the `sortByPriceCurrencyCode` as the shop's default currency code and any positive integer for `minPrice`, the results are sorted by the product minPrice for that currency, from high to low.
1. Same as above, but pass `sortOrder: desc` explicitly, and results should be the same.
1. Same as above, but pass `sortOrder: asc` explicitly, and results should be sorted from low to high.
1. Passing an invalid `sortByPriceCurrencyCode` should have an error response (throw error in the resolver).

### LINKED ISSUE #4254: Edit Access to Products for All Shops

spencern:
# v1.12.0

## Breaking changes
We've made some significant changes to the structure of the data that we publish to the Catalog in #4218. We have created an automated migration for these changes in #4272. If you have made other changes to the catalog in your app or in a plugin, this migration may not work out of the box for you. We'd recommend migrating a test or staging instance of your application with similar data before applying this update to your production application.

From the PR notes to #4218, here's a list of the changes

### Method argument change
"products/updateProductPosition" method now takes a tag ID rather than tag name as its third argument

### Media change

The `media` property on catalog items is still an array of objects, but the structure of those objects has changed.

BEFORE:
```js
{
  metadata: {}, // the full metadata object from the Media doc
  thumbnail: "", // a URL
  small: "", // a URL
  medium: "", // a URL
  large: "", // a URL
  image: "", // a URL
}
```

AFTER:
```js
{
  priority, // copied from metadata
  toGrid, // copied from metadata
  productId, // copied from metadata
  variantId, // copied from metadata
  URLs: {
    thumbnail: "", // a URL
    small: "", // a URL
    medium: "", // a URL
    large: "", // a URL
    original: "", // a URL
  }
}
```

Also, the media array is now sorted in ascending `priority` order.

### Deleted and hidden variants change

When published to the catalog, the `variants` array on the catalog item now includes only those variants where `isDeleted` IS NOT true and `isVisible` IS true.

### Product props separated from catalog item props

Most product properties that were formerly on the catalog item itself are now moved to a `product` object property.

### Explicit property copying

Previously all product and variant props were blindly copied into the catalog item. Now, we explicitly copy only the props we want, sometimes changing their names.

### Property name changes

- `variant.taxable -> variant.isTaxable`
- `variant.variantId` added, currently the same as variant._id but might not always be. A reference back to the variant in the Products collection.
- `product.taxable -> product.isTaxable`
- `product.productId` added, currently the same as product._id but might not always be. A reference back to the product in the Products collection.
- `product.handle -> product.slug`
- `product.hashtags -> product.tagIds`

`product.twitterMsg` and `product.facebookMsg` and `product.googleplusMsg` and `product.pinterestMsg` are converted to a `product.socialMetadata` array like this:

```js
socialMetadata: [
  { service: "twitter", message: product.twitterMsg },
  { service: "facebook", message: product.facebookMsg },
  { service: "googleplus", message: product.googleplusMsg },
  { service: "pinterest", message: product.pinterestMsg }
]
```

### New Pricing Object

We've deprecated the `price` property on catalog products, variants, and options. There is a new property called `pricing`, which is a map keyed by currency code in MongoDB, and in GraphQL is converted to an array.

The GraphQL schema:

```graphql
"The product price or price range for a specific currency"
type ProductPricingInfo {
  """
  A comparison price value, usually MSRP. If `price` is null, this will also be null. That is,
  only purchasable variants will have a `compareAtPrice`.
  """
  compareAtPrice: Float

  "The code for the currency these pricing details applies to"
  currency: Currency!

  """
  UI should display this price. If a product has multiple potential prices depending on selected
  variants and options, then this is a price range string such as "$3.95 - $6.99". It includes the currency
  symbols.
  """
  displayPrice: String!

  "The price of the most expensive possible variant+option combination"
  maxPrice: Float!

  "The price of the least expensive possible variant+option combination"
  minPrice: Float!

  """
  For variants with no options and for options, this will always be set to a price. For variants
  with options and products, this will be `null`. There must be a price for a variant to be
  added to a cart or purchased. Otherwise you would instead add one of its child options to a cart.
  """
  price: Float
}
```

These are currently mapped as follows, but in the future you will be able to have different prices per currency that your shop supports.

```js
pricing: {
      [shopCurrencyCode]: {
        compareAtPrice: variant.compareAtPrice || null,
        displayPrice: variantPriceInfo.range,
        maxPrice: variantPriceInfo.max,
        minPrice: variantPriceInfo.min,
        price: typeof variant.price === "number" ? variant.price : null
      }
    },
```

### Catalog Item props

All product props other than `positions` are now on the `product` object, so there are a few new properties on the catalog item itself:
- _id (this is no longer the same as the product ID)
- shopId (stays here AND on `product` obj)
- createdAt
- updatedAt

### ReactionProduct.getTag

The `ReactionProduct.getTag` helper function is renamed to `ReactionProduct.getTagIdForPosition` and returns the tag ID rather than name. If there's no tag in the route path, it returns "_default" rather than the lowercased shop name.

### Features and Improvements
 - feat: Catalog schema changes (#4218)
 - feat: Catalog schema migration (#4272)

### Bugfixes
 - fix: Limit Products when in Edit Mode (#4256) .. Resolves #4254
 - fix: added event param to onBlur method in numericInput component (#4251)
 - fix: Load more products working now (#4233) .. Resolves #4090
 - fix: show `.00` cents for whole-dollar amounts in price range (#4222)
 - fix: get react root node correctly (#4172)

### Tests
 - Reduce Risk of Test False Positives (#4033)

### Chores
 - chore: Update to our latest eslint config (#4282)
 - chore: resolve snyk issues (#4257)
 - chore: Use NPM random package (#4208)

### Features and Improvements
 - feat: catalog price sort (#4255) .. Resolves #4245
 - feat: devserver improvements (#4220)
 - feat: add description and name to shop query (#4209)
 - feat: catalogItem GQL query (#4200) .. Resolves #4106
 - feat: create connection/edge resolver helper (#4225)

### Bugfixes
 - fix: hasNextPage being false, even with a known next page (#4249)
 - fix: improve hasNext/hasPrevious (#4231)
 - fix: pagination (#4228)
 - fix: first value pagination (#4204) .. Resolves #4186

### Chores
 - chore: access media through devserver app (#4216)

## Docs
 - docs: remove community calls from readme.md (#4279)
 - docs: clean up jsdoc comments, namespace organization (#4213)

## Contributors
Thanks to @pmn4 for contributing to this release! 🎉


nnnnat:
@spencern did some acceptance testing this afternoon. 


**Tested:**
 * Creating, editing, archiving Products, Variants, and Options
 * Publishing Products from the grid
 * Order fulfillment, refunding and canceling
 * Anonymous checkout
 * Registering as a user during checkout

This [issue](https://github.com/reactioncommerce/reaction/issues/4202) is still happening in 1.12.0 but we may have inadvertently fixed it in 1.13.0 by removing the `autoValue` functions from the Shops schema.

spencern:
@nnnnat if it turns out that removing the `autoValue` from the Shops schema does resolve #4202, we can backport that into v1.12.1 as necessary.

PR #4288, opened 2018-06-01
(fix) add getAutoValues: false to discounts/codes/remove

### LINKED ISSUE #4081: Can't remove discount code in checkout once applied
## Issue Description
Once you apply a discount code to your cart, you can't remove it even though there is an X next to it which seems to imply you can

## Steps to Reproduce
1. Create a discount code
1. While checking out apply the discount code
1. Click on the X on the discount code to remove
1. Observe that a spinner spins for a second and cart refreshes but nothing happens

### Versions
```
Node: 8.9.4
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.10.0
Reaction branch: master
Docker: 18.03.0-ce
```

ajporlante:
Resolves #4081 
Impact: **major**  
Type: **bugfix**

## Issue
When there is an applied discount to a cart, clicking the button to remove the discount does not work.

The `Payment` schema has two autovalues: `_id` and `currency` which defaults to `{}`. Because of this, when cart is being updated through `$pull` of the discount payment, SimpleSchema modifies the selector `{ _id: codeId }` to `{ _id: codeId, currency: {} }` which does not match the discount payment that needs to be removed.

## Solution
Add `getAutoValues:false` to the options when updating the cart to avoid the selector for the $pull to be modified.

## Breaking changes
None

## Testing
1. Set up a discount code.
2. Create an order and apply the discount code on checkout.
3. Click the "X' button to remove the discount code.

### DIFF
--- a/imports/plugins/included/discount-codes/server/methods/methods.js
+++ b/imports/plugins/included/discount-codes/server/methods/methods.js
@@ -181,7 +181,7 @@ export const methods = {
     const result = Collection.update(
       { _id: id },
       { $set: { discount: currentDiscount }, $pull: { billing: { _id: codeId } } },
-      { multi: true }
+      { multi: true, getAutoValues: false } // See https://github.com/aldeed/meteor-collection2/issues/245 
     );
 
     // calculate discounts
--- a/imports/plugins/included/discount-codes/server/methods/methods.js
+++ b/imports/plugins/included/discount-codes/server/methods/methods.js
@@ -178,6 +178,9 @@ export const methods = {
       Collection.update(selector, update);
     }
     // TODO: update a history record of transaction
+    // The Payment schema's currency defaultValue is adding {} to the $pull condition.
+    // If this issue is eventually fixed, autoValues can be re-enabled here
+    // See https://github.com/aldeed/simple-schema-js/issues/272
     const result = Collection.update(
       { _id: id },
       { $set: { discount: currentDiscount }, $pull: { billing: { _id: codeId } } },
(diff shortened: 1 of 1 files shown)


aldeed:
`billing` is an array of `Payment`, which have an autoValue on `_id` prop here: https://github.com/reactioncommerce/reaction/blob/master/lib/collections/schemas/payments.js#L245

But that autoValue function is here: https://github.com/reactioncommerce/reaction/blob/master/lib/collections/schemas/helpers.js#L90-L98

And it already seems to properly handle set values and $pull in server code. So I would put some logging into that autoValue and try to figure out what’s happening there. Setting `getAutoValues` to `false` is fine if there’s no other solution, but it could have unintended side effects and there should be a less blunt way to fix.

aldeed:
@ajporlante ah, ok. This could be a bug in the SimpleSchema package. It resolves `defaultValue` to a built-in `autoValue` function defined here: https://github.com/aldeed/simple-schema-js/blob/master/package/lib/SimpleSchema.js#L893-L907

There might need to be some `$pull` handling added there. Would you mind creating an issue in https://github.com/aldeed/simple-schema-js/issues?

Also can you add a comment above the changed line in this PR? Something like "The Payment schema's currency defaultValue is adding `{}` to the $pull condition. If this issue is eventually fixed, autoValues can be re-enabled here." And then add a URL in the comment to this issue you create.

Thanks!

ajporlante:
Done @aldeed.

aldeed:
@spencern This is good to merge, but not sure if you want it in 1.13 or wait until next.

spencern:
@aldeed I'm trying to keep 1.13 light, so I added a release-1.14.0 branch, but since this is a bugfix, let's merge it into 1.13.0 for now

PR #4312, opened 2018-06-08
(fix): fix permissions of shop social settings

### LINKED ISSUE #4205: Primary Shop Social Settings Visible To all Marketplace Shops
## Prerequisites

## Issue Description
The social settings for all merchants in a marketplace are currently the same. This is a security issue, and not the expected behaviour. Each merchant should be able to set their own social settings.

## Steps to Reproduce
1. Log-in as admin in a marketplace
2. Enter some social settings
3. Open your social settings in a marketplace shop - the settings from the primary shop will show.

## Possible Solution
Doesn't look like a simple fix. I think this component hasn't yet been modified for use in marketplace?

### Versions
Node: 8.9.0
NPM: 5.7.1
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.10.0
Reaction branch: master
Docker: 18.03.0-ce

ajporlante:
Resolves #4205 
Impact: major
Type: bugfix

## Issue
`socialPackage` returned in `socialSettingsContainer` is always the package from the primary shop. This causes the social settings to be displayed to the dashboard of other shop owners.

In displaying the social buttons in product page, the settings of the primary shop owners are always used. Example, if product is from primary shop where all social settings are enabled, then all social buttons should be displayed. If product is from another shop where only Twitter is enabled, then product page should only show the Twitter button, but this is currently not the case.

## Solution
Add `shopId` filter to `socialSettingsContainer` and `createSocialSettings`. 

## Breaking changes
None

## Testing
1. Login as admin. Set the social settings in primary shop (all enabled). Publish a product.
2. Enable the marketplace. Invite a shop owner and activate the shop.
3. Login as shop owner. Set the shop's social settings (some possibly disabled). You should be able to see and save only the shop's credentials.


pmn4:
totally my opinion here, but it feels like the work should be done server-side. Having to pass the shopId from the client feels like a security issue

zenweasel:
@ajporlante What's the status on this?

server/publications/collections/packages.js
```diff
@@ -102,7 +102,14 @@ Meteor.publish("Packages", function (shopId) {
       }
       // observe and transform Package registry adds i18n and other meta data
       const observer = Packages.find({
-        shopId: myShopId
+        $or: [
+          { shopId: myShopId },
+          {
+            name: {
+              $in: ["reaction-social"]
```
zenweasel: This definitely seems like the wrong approach here. We shouldn't need to make a change to this global query for just this one package.

ajporlante:
I modified code to solve only the issue raised in https://github.com/reactioncommerce/reaction/issues/4205. While working on this, I discovered another bug and have already created a ticket for it. https://github.com/reactioncommerce/reaction/issues/4352 

@pmn4 There's a security rule protection in the server side so I think it's not much of an issue to need to pass `shopId` from the client side when updating social credentials.

spencern:
@ajporlante you have failing CI tests here as well

ajporlante:
@spencern All good now. Just had to rerun the build.

zenweasel:
This is working for me. To address @pmn4's note. This really does nothing beyond what is already being published in the `Packages` collection, and that data is available publicly on the primary site. If that is private information we need to be gating it in the publication. I double checked that we aren't leaking other things like api keys for Stripe, etc. and we don't.

PR #4317, opened 2018-06-11
Fix NPM build issue in Docker build

impactmass:
Impact: **major**  
Type: **bugfix**

## Issue
There is to an async race issue when installing packages with the NPM version (v5.10.0) in Meteor 1.7, the app intermittently throws a `npm ERR! write after end` error during the NPM install step. 

More on the issue: https://github.com/meteor/meteor/issues/9940

## Solution
We are switching to using the NPM version installed in the base image (v5.6.0).This prevents the "write after end" errors seen with this command. This will be reverted when Meteor updates to an NPM version without this issue.

## Breaking changes
None

## Testing
* Clean up old stuffs
  * `docker-compose down -v --rmi local --remove-orphans`
* Rebuild your reaction images
  * `docker-compose up --build --force-recreate --renew-anon-volumes`.
  * If you only want to run reaction `docker-compose up --build --force-recreate --renew-anon-volumes reaction`.
* Confirm there's no `npm ERR! write after end` during the build process; and confirm that the app starts successfully and available in browser

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ##############################################################################
 # meteor-dev stage - builds image for dev and used with docker-compose.yml
 ##############################################################################
-FROM reactioncommerce/base:v4.0.2 as meteor-dev
+FROM reactioncommerce/base:meteor-1.7.0.1 as meteor-dev
 
 LABEL maintainer="Reaction Commerce <architecture@reactioncommerce.com>"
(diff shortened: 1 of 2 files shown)


aldeed:
Can be merged after a couple more people test on this branch

nnnnat:
worked 4 me

PR #4318, opened 2018-06-12
Use new Reaction component library components for the SMS settings form

aldeed:
Impact: **minor**  
Type: **feature**

## Changes
To start working toward better UI consistency, a simple form pattern, and removal of remaining autoforms, we're planning to use ReactoForm package with our own form input components for our new operator UI. This PR pulls in those NPM packages and tries out the approach with a single form, the SMS Settings form.

## Breaking changes
None

## Testing
1. Start the app and log in as admin
1. Open SMS admin dashboard
1. Play around with the form entering various valid and invalid values. Make sure you see proper errors for invalid values. Make sure valid values are saved.
1. When going to a different admin panel and then back, make sure current saved values are shown in the form. When refreshing the page, make sure current saved values are shown in the form.


aldeed:
@nnnnat Input icons should be fixed now. For now, I'm passing them in directly. Later we can replace this with the idea of pulling components from React context.

PR #4324, opened 2018-06-14
(fix): Use Catalog collection for PDP

### LINKED ISSUE #4319: [release-1.13.0]: PDP and Product Grid sometimes display different data to customers
## Issue Description
**Versions** This issue only occurs on `release-v1.13.0` and later

In #4237 we removed the existing Revision control system. Publishing a product publishes data to the Catalog. When a customer is browsing the Product Grid, the products are sourced from the Catalog. However, currently when the customer views a product on the Product Detail Page, that information is sourced from the Products collection. This means that customers could see "unpublished" changes to a Product and that the customer could see different information when navigating from the Product Grid to the PDP.

This is an `impact-critical` bug which is `release-blocker` for `v1.13.0`

## Steps to Reproduce
1. In one session (an incognito window or different browser), as an anonymous customer navigate to the example product's detail page.
2. In a second session, login as an admin, navigate to the detail page of the example product and change the title
3. In the anonymous customer session, observe that the title on the product detail page is updated without the operator having to publish the changes.
4. In the anonymous customer session navigate back to the grid and notice that the title is the Published title for the product from the Catalog.

## Possible Solution
We'll need to change the data for the Product Detail page for customers and operators without the `create-product` role to source data from the Catalog collection instead of the Product collection.

### Versions
```
Node: 9.11.1
NPM: 5.6.0
Meteor Node: 8.11.2
Meteor NPM: 5.10.0
Reaction CLI: 0.29.0
Reaction: 1.13.0
Docker: 18.03.1-ce
```

mikemurray:
Resolves #4319   
Impact: **major**  
Type: **feature/bugfix**

## Issue

In a post revision control world, the product detail page receives changes immediately and is out of sync with the product on the product grid.

## Solution

Using `Approach 4`

> Do the heavy lifting in the Product publication, using observables to deconstruct a catalog item product back into multiple documents for the product and variants.

## Breaking changes

PDP page will use the Catalog collection for customers only.

## Testing
1. Edit a product as an admin
2. In another browser or incognito mode, while not signed in, see that no changes have appeared on page
3. Publish changes
4. In the other browser, see changes have been made.

---

1. As a customer add a product to the cart
1. Checkout as normal
1. As an admin fulfil that order
1. Ensure that inventory has be decremented properly

### DIFF
--- a/imports/plugins/core/ui/client/containers/inventoryBadge.js
+++ b/imports/plugins/core/ui/client/containers/inventoryBadge.js
@@ -1,25 +1,52 @@
 import { registerComponent, composeWithTracker } from "@reactioncommerce/reaction-components";
 import { InventoryBadge } from "../components/badge";
+import { Reaction } from "/client/api";
 import { ReactionProduct } from "/lib/api";
 
 const composer = (props, onData) => {
   const { variant, soldOut } = props;
-  const { inventoryManagement, inventoryPolicy, lowInventoryWarningThreshold } = variant;
-  const inventoryQuantity = ReactionProduct.getVariantQuantity(variant);
+  const {
+    inventoryManagement,
+    inventoryPolicy,
+    lowInventoryWarningThreshold,
+    isSoldOut,
+    isBackorder,
+    isLowQuantity
+  } = variant;
   let label = null;
   let i18nKeyLabel = null;
   let status = null;
-  // TODO: update this to use Catalog API.
-  if (inventoryManagement && !inventoryPolicy && inventoryQuantity === 0) {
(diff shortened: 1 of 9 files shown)


mikemurray:
@nnnnat @spencern 

**Outstanding items**

- [ ] Fix broken tests
- [ ] Ensure that inventory is updated properly
- [ ] Ensure cart and order summary in profile display the correct product info from the catalog
- [ ] Ensure the order in the admin shows the correct data from the catalog item

spencern:
Thanks @mikemurray - we'll pick this up from here.

PR #4325, opened 2018-06-14
(feat) Use token to create Stripe charges

### LINKED ISSUE #4323: Create Stripe plugin that uses client-side tokensization
@zenweasel commented on [Fri Apr 13 2018](https://github.com/reactioncommerce/solutions-engineering-tasks/issues/11)

The current Stripe implementation submits card info directly to our server which we then submit to Stripe. The "preferred" method is to collect the data client-side and submit it to Stripe and get back a token which then can be used for payment. This has the advantage of our server never having the credit card information, drastically diminishing the security vector, but also has the challenge of needing to integrate Stripes client-side code in a way that leads to a good user experience and fits within the design guidelines of the site.

This plug should work like the existing plugin in most ways except it should use this method of collecting user information.

I am suggesting making this a separate plugin, but if we can integrate the client code well enough it could probably replace the existing server-side only code.

ajporlante:
Resolves #4323 
Impact: major
Type: feature

## Issue
Sensitive card data touch our server prior to submitting to Stripe which is not the ideal security practice.

## Solution
Tokenize credit card data on the client using [Stripe Elements](https://stripe.com/elements) which are pre-built Stripe UI components. This way, only a token will reach the backend that can be used to create charges. The source code for the React components is [here](https://github.com/stripe/react-stripe-elements).

## Breaking changes
1. The checkout screen will change.
- Name is no longer required.
- The card number input creates a space every 4 characters.
- Expiry month and year are in a single input, no dropdown for month selection.
- Postal code is required.
- There is field validation but errors are not linked to a field, so all error messages are displayed just above the submit button and are determined only after user submits the form.
- Spacing and font are not yet exactly the same as our other payment method implementations but I believe these are relatively easier to fix as compared to the other mentioned changes.

screen shot 2561-06-14 at 17 29 53

2. Added a public field `settings.public.publishable_key` in setting up the primary shop's Stripe credentials. It is used the public key to tokenize credit card data on the client.

screen shot 2561-06-14 at 17 16 05

3. Added `react-stripe-elements` as a dependency.
4. Revised the test for `stripe/payment/createCharges`. This PR modifies the function mentioned to accept a `token` object instead of `cardData`. Hence, there is a necessity to revise the test to adjust the parameters needed for it to pass. The expected results from the old test was not changed, so there should not be a huge impact.

Aside from this test, I added 2 additional tests for `stripe/payment/createCharges` for better coverage namely:
`should throw an error with an invalid api key` and
`should be able to create multiple charges`

## Testing
1. Set up Stripe credentials in the Payment Dashboard (Secret API Key, Publishable Key, Public Client ID and percentage fee.)
2. Create an order and pay.
3. As an admin, you should be able to capture and refund the payment.

To test marketplace payments:
1. Turn on the marketplace feature. Invite a shop owner and activate the shop.
2. Log in as a shop owner and create and publish a product.
3. In Shop Settings Dashboard, update the shop address and click Start Accepting Payment under Setup Stripe Account.
4. As a guest, buy items from the primary shop and the new shop.
5. After payment, the admin and shop owner should be able to capture and refund payment of customer.


ajporlante:
I updated the testing instructions @zenweasel .

rymorgan:
Ok, I poked around and found a publishing key and tested this for the card entry side of things. I'm good with this implementation except for some small formatting issues. The long form fields should all match up with the length of the button. 

cart_checkout

rymorgan:
cart_checkout-error

ajporlante:
@rymorgan Some of these things you suggest are not the current implementations with the other payment methods. If you look at the Braintree or the example payment method:
1. Border color and text color of field in error both remain the same. Instead, only the label becomes red.
2. Alignment of button and fields in 2 columns is off in the other payment methods, but I guess these are not intended.

This is the Braintree implementation:
screen shot 2561-06-18 at 15 30 10

From my end, I will fix the alignment and the font size in the Stripe implementation, which Im already able to do. But this means, that the other implementations have to be fixed too.

Need your confirmation with how the field validation should change the form.

zenweasel:
@ajporlante All payment methods except for Stripe and the Example payment method are going to be removed in 1.14. These will be the new "canonical" implementations

ajporlante:
Managed to display field validations, correct the font and alignment. This is how it looks so far:

screen shot 2561-06-18 at 18 15 49

Let me know what you think.

rymorgan:
@ajporlante I hear what you are saying with the other packages and, yes, unfortunately, the app isn't always consistent in implementation nor is it always using best practices. So what we've done in the past isn't always the greatest marker for what we are going to be doing moving forward. Yes, also to echo what @zenweasel said this payment package is meant to be our supported example so therefore it should include all our best practices. We're fighting these UI/UX issues on two fronts, one as we go in the context of priority GitHub issues and two, as we begin to work on and implement the component library but that's going to be a longer slog.

rymorgan:
The alignment looks good now, thanks for fixing that and the validation of the field is looking good too. This will help users not miss data and know what they are missing when they do skip something.  

Two small final items -

- The labels should also be red when in error. 
- The left to right padding in the payment card doesn't have enough padding or match the padding on the other cards. W/out checking the css it looks like it needs to be bumped up to 20px to match. 

cart_checkout

ajporlante:
Fixed the label styling, the side paddings with my latest commit.

rymorgan:
LGTM

imports/plugins/included/payments-stripe/client/checkout/containers/injectedCardForm.js
```diff
@@ -0,0 +1,231 @@
+import { Meteor } from "meteor/meteor";
```
zenweasel: Import order:  https://docs.reactioncommerce.com/reaction-docs/master/styleguide

imports/plugins/included/payments-stripe/client/checkout/containers/injectedCardForm.js
```diff
@@ -0,0 +1,231 @@
...
+    if (this.state.errorMessage) {
+      return (<div className="alert alert-danger">{this.state.errorMessage}</div>);
+    }
+    return null;
+  }
+
+  render() {
+    const style = {
+      base: {
+        "fontSize": "14px",
```
zenweasel: I don't think we want styles hardcoded in here. I should be able to change styling without modifying core code.

imports/plugins/included/payments-stripe/client/checkout/containers/stripePaymentFormContainer.js
```diff
@@ -0,0 +1,53 @@
+import { Meteor } from "meteor/meteor";
```
zenweasel: Import order

imports/plugins/included/payments-stripe/client/checkout/containers/injectedCardForm.js
ajporlante: This is unavoidable as Elements API only allow font styles to be passed as props. I have put other non-font-related css, like borders and paddings, into a separate less file.

imports/plugins/included/payments-stripe/client/checkout/containers/stripePaymentFormContainer.js
ajporlante: Fixed and will keep this in mind moving forward!

stunjiturner:
From the stripe documentation, the use of the secret key in a implementation is for when you are using your own servers. There are also two types of API keys: publishable and secret.
Publishable API keys are meant solely to identify your account with Stripe, they aren’t secret. In other words, they can safely be published in places like your Stripe.js JavaScript code, or in an Android or iPhone app. Publishable keys only have the power to create tokens.

Secret API keys should be kept confidential and only stored on your own servers. Your account’s secret API key can perform any API request to Stripe without restriction.

spencern:
@stunjiturner Does this PR publish or otherwise compromise the secret API key?

ajporlante:
We still use the secret key in creating the charge and capturing payments.

stunjiturner:
as long as the charges and captures are are on the server side. can you point me to the server side code for capturing so I can take a look.

ajporlante:
It's here https://github.com/reactioncommerce/reaction/blob/ddab1cbdd871faa9b5da6d0709ef9700015bec8d/imports/plugins/included/payments-stripe/server/methods/stripe.js#L238-L239

and yes, it's server side.

zenweasel:
Seeing this error in the browser console:
```
Exception while simulating the effect of invoking 'cart/submitPayment' errorClass {message: "Match error: Expected plain object", path: "", sanitizedError: errorClass, errorType: "Match.Error", stack: "Error: Match error: Expected plain object↵    at c…7fae7e488e45468ea4fda5047b1baf99d39dfd73:1962:12)"} Error: Match error: Expected plain object
    at check (http://localhost:3000/packages/check.js?hash=03574af9b5218423b2e7b6d9bae433f2573596f1:77:17)
    at MethodInvocation.cart/submitPayment (http://localhost:3000/app/app.js?
```

ajporlante:
Hi @zenweasel . That is really a nice catch. Fixed the issue in my most recent commits.

spencern:
@ajporlante do we need another review here? Let's get this merged!

spencern:
Thanks @zenweasel I've added @willopez to review the React code

imports/plugins/included/payments-stripe/client/checkout/components/injectedCardForm.js
```diff
@@ -0,0 +1,273 @@
...
+          submitting: false
+        });
+        return;
+      });
+    }
+    return;
+  };
+
+  handleCardNumberChange = (change) => {
+    if (change && change.error) {
```
willopez: this `if` block can be extracted to a function and params passed in. And then use on all the change handlers to reduce code duplication.

imports/plugins/included/payments-stripe/client/checkout/components/injectedCardForm.js
```diff
@@ -0,0 +1,273 @@
...
+      expDateErrorMessage: "",
+      postal: props.postal,
+      postalErrorMessage: "",
+      submitMessage: "Complete your order",
+      submitting: false
+    };
+
+    /* eslint-disable camelcase */
+    this.errorCodes = {
+      card_declined: i18next.t("checkout.errorMessages.cardDeclined", "Your card has been declined. Please use another card."),
```
willopez: Can this error message be retrieved from the package translation file? this will remove code duplication.

ajporlante:
Hi @willopez , thanks for the review! Updated the code per your recommendations.

imports/plugins/included/payments-stripe/client/checkout/components/injectedCardForm.js
```diff
@@ -0,0 +1,278 @@
...
+        return;
+      });
+    }
+    return;
+  };
+
+  changeHasError = (change) => (change && change.error);
+
+  handleCardNumberChange = (change) => {
+    if (this.changeHasError(change)) {
```
willopez: Can this `if el` block be refactored into a function? and used in the other event handlers?

ajporlante: What happens within the `if else` are different for each function and `handlePostalChange` has an extra needed step. So, I don't think I can do what you suggest,

willopez: I see, then this PR looks good to ship.

willopez:
@ajporlante just noticed the snyk issues, please fix and why is the docker-build failing?

willopez:
@spencern @ajporlante @zenweasel This PR will be ready to merge after this PR https://github.com/reactioncommerce/reaction/pull/4444 is merged.

ajporlante:
@spencern @willopez I made a wrong commit to `package-lock.json` earlier today causing the unit tests to fail, Fixed it, so it's all good now. All required tests passed.

willopez:
@spencern Looks good to me.

PR #4331, opened 2018-06-15
(fix): Update prerender.js

### LINKED ISSUE #4330: Prerender does not work with prerender.io
## Issue Description
Reaction prerender feature does not work with prerender.io

## Steps to Reproduce
1. Get a token from [Prerender](https://prerender.io) by signing up/logging in to their service.
2. Using this token, set your environment variables as:
```
PRERENDER_TOKEN=token
PRERENDER_SERVICE_URL=http://service.prerender.io/
```
3. In your prerender.io dashboard, the message "We haven't seen a request with your Prerender token yet." is not removed despite correct set up in Step 2.

## Possible Solution
Update `server/startup/prerender.js` to require `PRERENDER_SERVICE_URL` instead of `PRERENDER_HOST`

See https://github.com/prerender/prerender-node

### Versions
```
Node: 8.11.2
NPM: 5.6.0
Meteor Node: 8.11.2
Meteor NPM: 5.10.0
Reaction CLI: 0.29.0
Reaction: 1.12.1
Reaction branch: master
```

ajporlante:
Resolves #4330 
Impact: minor
Type: bugfix

## Issue
Current prerender implementation does not have provision to set `PRERENDER_SERVICE_URL` as a parameter which is required if one intends to use [prenrender.io](https://prerender.io) as the prerender server.

## Solution
Update `server/startup/prerender.js` to set `PRERENDER_SERVICE_URL` in the prerender function when present as an environment variable.

## Breaking changes
None

## Testing
1. Sign up/log in to a prerender.io account. Get your token.
2. With that token, add below to your environment variables:
```
PRERENDER_TOKEN=token
PRERENDER_SERVICE_URL=http://service.prerender.io/
```
3. Access a page in your Reaction site and append `?_escaped_fragment_=` to the url.
4. After a couple of minutes, you will see that there is no more error message in your prerender.io dashboard saying "We haven't seen a request with your Prerender token yet."

### DIFF
--- a/server/startup/prerender.js
+++ b/server/startup/prerender.js
@@ -3,9 +3,12 @@ import { WebApp } from "meteor/webapp";
 import { Logger } from "/server/api";
 
 export default function preRender() {
-  if (process.env.PRERENDER_TOKEN && process.env.PRERENDER_HOST) {
+  if (process.env.PRERENDER_TOKEN && process.env.PRERENDER_SERVICE_URL) {
     prerender.set("prerenderToken", process.env.PRERENDER_TOKEN);
-    prerender.set("host", process.env.PRERENDER_HOST);
+    prerender.set(“prerenderServiceUrl”, process.env.PRERENDER_SERVICE_URL);
+    if (process.env.PRERENDER_HOST) {
+      prerender.set("host", process.env.PRERENDER_HOST);
+    }
     prerender.set("protocol", "https");
     WebApp.rawConnectHandlers.use(prerender);
     Logger.info("Prerender Initialization finished.");
--- a/server/startup/prerender.js
+++ b/server/startup/prerender.js
@@ -13,8 +13,8 @@ export default function preRender() {
     WebApp.rawConnectHandlers.use(prerender);
     Logger.info("Prerender initialization finished.");
   } else if (process.env.PRERENDER_TOKEN) {
-     Logger.error("Prerender initialization failed. Please set PRERENDER_SERVICE_URL in your environment variables.");
+    Logger.error("Prerender initialization failed. Please set PRERENDER_SERVICE_URL in your environment variables.");
   } else if (process.env.PRERENDER_SERVICE_URL) {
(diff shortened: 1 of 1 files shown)


zenweasel:
@ajporlante Remember to assign a reviewer or set your PR to WIP

server/startup/prerender.js
```diff
@@ -3,9 +3,12 @@ import { WebApp } from "meteor/webapp";
 import { Logger } from "/server/api";
 
 export default function preRender() {
-  if (process.env.PRERENDER_TOKEN && process.env.PRERENDER_HOST) {
+  if (process.env.PRERENDER_TOKEN && process.env.PRERENDER_SERVICE_URL) {
     prerender.set("prerenderToken", process.env.PRERENDER_TOKEN);
-    prerender.set("host", process.env.PRERENDER_HOST);
+    prerender.set("prerenderServiceUrl", process.env.PRERENDER_SERVICE_URL);
+    if (process.env.PRERENDER_HOST) {
```
zenweasel: You are looking for `process.env.PRERENDER_HOST` here, but you are already guarding for only having `process.env.PRERENDER_TOKEN && process.env.PRERENDER_SERVICE_URL` so this code would never get executed?

ajporlante: `PRERENDER_HOST` is just an optional parameter. Both `PRERENDER_TOKEN` AND `PRERENDER_SERVICE_URL` are required. I don't think it makes sense to add `PRERENDER_HOST` without these 2 required values.

zenweasel: Isn't `PRERENDER_HOST` the old version of `PRERENDER_SERVICE_URL`? Why would we allow both?

ajporlante: I dont think so. Looking at the history of https://github.com/prerender/prerender-node, I see it has always been `PRERENDER_SERVICE_URL` to specify the prerender server. `PRERENDER_HOST` is for another purpose which is to specify the site URL.

zenweasel: I am confused by why we grouped those two parameters together. They are both required? If so should we throw an error if one is provided and the other is not?

PR #4342, opened 2018-06-19
Bugfix: low quantity/sold out flags not saving correctly in db

### LINKED ISSUE #4328: Incorrect inventoryStatus on Product Variants and Options when published to Catalog
## Prerequisites

## Issue Description
When a Product is published to the Catalog collection the nested `variants` and `options` will always have inventory status properties of `false`.

## Steps to Reproduce
1. Spin up a fresh RC shop and log in as an admin.
2. Select the "Basic Reaction Product" and update an Option's quantity to be 0.then publish.
3. You should see a "Sold Out" badge on the Option.
4. Looking at the Catalog collection in mongo you'll see the Option's`isSoldOut` property is false.
5. Looking at the same Option in the Products collection you'll see the `inventoryQuantity` is `0`.

## Possible Solution
`xformVariant` is trying to [cast undefined properties to booleans](https://github.com/reactioncommerce/reaction/blob/0e54ea1ae46eb82805d83256c2cc2710633cf915/imports/plugins/core/catalog/server/no-meteor/utils/publishProductToCatalog.js#L26) when creating a Catalog entry from a Product variant. `isSoldOut` and `isLowQuantity` properties are optional on Product/Variant schema and don't get added.

Update/refactor `xformVariant` to use `isSoldOut` and `isLowQuantity` util functions when setting properties.

### Versions
```sh
Node: 8.9.4
NPM: 5.6.0
Meteor Node: 8.11.2
Meteor NPM: 5.10.0
Reaction CLI: 0.29.0
Reaction: 1.13.0
```

willopez:
Resolves #4328 
Impact: minor
Type: bugfix

## Issue
When a Product is published to the Catalog collection the nested variants and options will always have inventory status properties of false.

## Solution
Update/refactor `xformVariant` to use `isSoldOut` and `isLowQuantity` util functions when setting properties

## Testing
1. Set option/variant quantity to 0 and publish
2. Verify corresponding property is set correctly in `Catalog` collection
3. Repeat process for low quantity property.


aldeed:
@nnnnat @willopez Based on the way the operator UI is laid out, it seems like the intention is for "warn at" to apply only to the top-level variant when there are options. In other words, when the sum of the qty of all options < warn at, then the top-level variant `isLowQuantity`. So in that paradigm, I'd say that whenever the top-level variant's `isLowQuantity` is `true`, all child options should be set to `true`.

That would match the UI, but as to whether that is correct... I'm not sure. It seems more sensible to me to track a warn-at value per option. @rymorgan or @spencern?

willopez:
Looking at the top level variant's inventory level would make sense for now. Tracking inventory per variant option would be better, however, that would require creating UI to manage it.

spencern:
@willopez we have an inventory quantity per option, but not flags correct? We absolutely need to be tracking inventory quantity on an option level.

aldeed:
@willopez I'm ok with that.

In general I think we're doing these wrong anyway. They should ideally be calculated dynamically per request since qty is constantly changing. Hopefully we're dealing with that now by updating the catalog product every time quantity changes, but I'll likely make an issue suggesting that we switch to dynamic.

spencern:
Agree with @aldeed.

PR #4366, opened 2018-06-25
Move all /server files to plugins

### LINKED ISSUE #4363: Migrate /server to plugins
## Work

Move all files in `/server` to some plugin within `/imports/plugins`. Update imports everywhere as necessary.

aldeed:
Resolves #4363 
Impact: **breaking**  
Type: **refactor**

## Changes
Move all files from within /server to various plugins

## Breaking changes
For any custom code or community plugins, all imports with paths that begin with `/server` will need to be changed.

## Testing
General smoke testing to catch any refactor errors. No substantial code or logic changes were made.


aldeed:
@spencern I fixed the order completion issue. One of the method hooks wasn't firing, maybe because of timing of when it was added, but it made more sense to just move that code into the method it was hooking. I verified that the other method hooks are running.

I see the address slowness thing, too, but I don't see how my changes would have impacted that. It does work eventually, but it's slow. I feel like we debugged and fixed that issue awhile back. Not sure it is worth our time.

I also noticed errors because Twilio and Nexmo were not being imported correctly, and adjusted those imports. You may want to retest sms sending.

spencern:
Issues with checkout have been resolved.
Going to see if I can determine the reason that the address book is responding differently

zenweasel:
@spencern Not sure if it's the same thing, but seeing really bad performance from `account/addressBookAdd` when doing load testing, which we are doing on 1.13

PR #4381, opened 2018-06-28
getSlug does not leave characters that will be escaped in a URL

### LINKED ISSUE #4344: getSlug leaves characters that will be escaped in a URL
## Prerequisites
* [ X ] Are you running the latest version?
* [ X ] Are you able to consistently reproduce the issue?
* [ X ] Did you search the issue queue for existing issue? [Search issues](https://github.com/reactioncommerce/reaction/issues)

## Issue Description
It seems that the default replace options of either the transliteration https://github.com/reactioncommerce/transliteration or slugify https://github.com/simov/slugify packages Reaction is using, do currently allow characters that will be escaped within a URL to be included in shopSlugs and product handles. 

## Steps to Reproduce

When for example, a user creates a new shop with name "@Sarah's Corner#$%" the resulting shopSlug will be "@sarah's-cornerdollar" which is going to become "%40sarah's-cornerdollar" when inside a URL. 

This kind of default behavior can result in plenty of unnecessary confusion and can cost in terms of customer support resources. The same behavior can be seen in the making of product handles as the same function (getSlug) is responsible for this as well.

## Possible Solution

getSlug uses whatever definition of slugify is passed on to her from lazyLoadSlugify. In that sense since the getSlug slugify might be either transliteration or slugify, the solution I came up with was to replace the following lines of code from lazyLoadSlugify. 

`// slugify is exported to modules.default while transliteration is exported to modules.slugify
  slugify = mod.default || mod.slugify;`

with this:

`// slugify is exported to modules.default while transliteration is exported to modules.slugify
  if (mod.default) {
    slugify = mod.default;
    slugify.extend({
      " ": "-", "$": "-dlr-", "&": "-n-", "`": "-opn-qte-", ":": "-cln-", "<": "-l-an-br-", ">": "-r-an-br-",
      "[": "-l-sq-br-", "]": "-r-sq-br-", "{": "-l-cu-br-", "}": "-r-cu-br-", '"': "-qts-", "+": "-plus-", "#": "-otrp-",
      "%": "-pct-", "@": "-at-", "/": "-sls-", ";": "-scl-", "=": "-eql-", "?": "-qstn-", "\\": "-bsl-", "^": "-crt-",
      "|": "-or-", "~": "-tld-", ",": "-cma-"
    });
  } else {
    slugify = mod.slugify;
    slugify.config({
      replace: {
        " ": "-", "$": "-dlr-", "&": "-n-", "`": "-opn-qte-", ":": "-cln-", "<": "-l-an-br-", ">": "-r-an-br-",
        "[": "-l-sq-br-", "]": "-r-sq-br-", "{": "-l-cu-br-", "}": "-r-cu-br-", '"': "-qts-", "+": "-plus-", "#": "-otrp-",
        "%": "-pct-", "@": "-at-", "/": "-sls-", ";": "-scl-", "=": "-eql-", "?": "-qstn-", "\\": "-bsl-", "^": "-crt-",
        "|": "-or-", "~": "-tld-", ",": "-cma-"
      }
    });
  }
`

Like getSlug, lazyLoadSlugify is defined multiple times throughout the latest working version of Reaction. So in order to centralize all necessary replacements I have defined this once as a const just below the latinLangs const in lib/api/helpers:

`export const replacementOptions = {
  " ": "-",
  "$": "-dlr-",
  "&": "-n-",
  "`": "-opn-qte-",
  ":": "-cln-",
  "<": "-l-an-br-",
  ">": "-r-an-br-",
  "[": "-l-sq-br-",
  "]": "-r-sq-br-",
  "{": "-l-cu-br-",
  "}": "-r-cu-br-",
  '"': "-qts-",
  "+": "-plus-",
  "#": "-otrp-",
  "%": "-pct-",
  "@": "-at-",
  "/": "-sls-",
  ";": "-scl-",
  "=": "-eql-",
  "?": "-qstn-",
  "\\": "-bsl-",
  "^": "-crt-",
  "|": "-or-",
  "~": "-tld-",
  ",": "-cma-"
};
`

### Versions

While I was not able to test this in the latest Reaction version 1.13 because of this issue https://github.com/reactioncommerce/reaction/issues/4202 I have tested this on our team's 1.11 backed development site and it works as expected.

alexandros84:
Resolves #4344
Impact: **major**  
Type: **bugfix**

## Issue
It seems that the default replace options of either the transliteration https://github.com/reactioncommerce/transliteration or slugify https://github.com/simov/slugify packages Reaction is using, do currently allow characters that will be escaped within a URL to be included in shopSlugs and product handles. 

When for example, a user creates a new shop with name:

> "\@sarah's Corner#$%" 

the resulting shopSlug will be -> 

> "\@sarah's-cornerdollar" 

which is going to become: 

> "%40sarah's-cornerdollar" 

inside a URL.

This kind of default behavior can result in plenty of unnecessary confusion and can cost in terms of customer support resources. The same behavior can be seen in the making of product handles as the same function (getSlug) is responsible for this as well.

## Solution
Since the transliteration reaction commerce package and the slugify package use different syntax for their replace options I have chosen to override their respective default behavior. My first attempt in solving this issue was to implement a regex in every getSlug definition like so: 

> slugString.replace(/[@#$%^&*`"<>\[\]\\{}:+/?=|,~']/g, "");

Since Reaction however, is giving us the option to deal with it effectively through designated replace options that will provide also the opportunity to sellers to differentiate creatively their shop names (for example " #hot? ") I decided to offer a better integrated solution that will actually use the aforementioned designated replacement options.

This solution also seems to me easier to maintain than implementing a client side validation since replacement options are being defined in one place.

## Testing

Try creating a new shop or a custom made slug using any of the following characters:

https://www.werockyourweb.com/url-escape-characters/ 

The shopSlugs or resultant Tags should be produced according to the following object: 

// Object that is regulating the way how characters that will be escaped in a URL should be replaced.
// This behavior can affect the composition of both shopSlugs and product handles.

`export const replacementOptions = {`
 ` " ": "-",`
  `"$": "-dlr-",`
 ` "&": "-n-",`
  "`": "-opn-qte-",`
  `":": "-cln-",`
 ` "<": "-l-an-br-",`
  `">": "-r-an-br-",`
  `"[": "-l-sq-br-",`
 ` "]": "-r-sq-br-",`
  `"{": "-l-cu-br-",`
  `"}": "-r-cu-br-",`
  `'"': "-qts-",`
  `"+": "-plus-",`
 ` "#": "-otrp-",`
 ` "%": "-pct-",`
 ` "@": "-at-",`
  `"/": "-sls-",`
  `";": "-scl-",`
  `"=": "-eql-",`
 ` "?": "-qstn-",`
 ` "\\": "-bsl-",`
  `"^": "-crt-",`
  `"|": "-or-",`
  `"~": "-tld-",`
  `",": "-cma-"`
`};`

Credit @artlimes


client/modules/core/helpers/utils.js
```diff
@@ -40,6 +49,7 @@ export function getSlug(slugString) {
   Promise.resolve(lazyLoadSlugify()); // eslint-disable-line promise/catch-or-return
   if (slugString && slugify) {
     slug = slugify(slugString.toLowerCase());
+    console.log(slug);
```
zenweasel: Need to remove the console.log

lib/api/helpers.js
```diff
@@ -209,6 +248,7 @@ export function getSlug(slugString) {
   Promise.await(lazyLoadSlugify());
   if (slugString) {
     slug = slugify(slugString.toLowerCase());
+    console.log(slug);
```
zenweasel: Remove console.log

package.json
```diff
@@ -19,7 +19,7 @@
     "url": "https://github.com/reactioncommerce/reaction/issues"
   },
   "dependencies": {
-    "@babel/runtime": "7.0.0-beta.49",
+    "@babel/runtime": "^7.0.0-beta.49",
```
zenweasel: Please remove this change. We purposefully fully pin packages for security reasons

server/api/core/utils.js
```diff
@@ -11,5 +12,10 @@ import { slugify } from "transliteration";
  * @return {String} slugified string
  */
 export function getSlug(slugString) {
+  slugify.config({
+    replace:
+      replacementOptions
+  });
+  console.log(slugify(slugString));
```
zenweasel: Remove console.log

aldeed:
@alexandros84 Do you have time to merge/rebase on 1.17.0 and address @zenweasel's feedback?

spencern:
I think there are some good ideas in this PR, but not sure it's quite right as is for the core product.
A few comments:
1. Determining which characters should be replaced may be a core concern, perhaps determined by characters that are escaped during [URL encoding](https://en.wikipedia.org/wiki/Percent-encoding), but potentially augmented by other characters. I'd like to have a broader discussion on this topic at some point.

2. Deciding what string to replace each character with seems like a shop/marketplace decision. Some may want to transliterate characters into things like `-at-` but others may just want to replace universally with a hyphen `-` or some other character.

3. Thinking about this problem from the ground level, I believe the main issue can be solved by permitting slugs (shop, tag, product, etc) to be set manually rather than automatically. We're starting work on more robust tag management system that will take this issue into account for tags, and I think the concept applies more broadly to any system that creates a slug.

I'm going to close this issue for now, but we should continue to keep the concerns outlined here in mind.

Issue #4264, opened 2018-05-22
Add a hash of published product properties to all top-level products


mikemurray:
## Work
- As part of publishing a Product to the Catalog, create an object hash of all of the properties that should be considered "changed". This will be most properties, but not system-managed properties like `updatedAt`.
- Save that hash onto a new property on Products collection, on the top-level product only. Name can be something like `publishedProductHash`
- Ensure that that field is being published to client code when an admin is logged in ("Products" publication)
- Add a migration that calculates this field for all existing CatalogProducts and saves it onto their related Product document.

The remaining client work will be done in #4281 

## Testing
- Create a Product and publish it. When logged in as admin, ensure there is a published product hash property on the related Product in client code.
- Create Products in the base branch and publish them. Update to this branch and ensure that migrations run. Ensure that all Products have a published product hash property in client code.

aldeed:
Whoever does this, note the potentially-related changes in https://github.com/reactioncommerce/reaction/pull/4073

kieckhafer:
How is this going to work to compare hashes? It seems we are saving the hash on the original publication of the product to the catalog, but then seems like we might just be overwriting the hash the next time the product is published.

How are we able to compare and see that the hash has changed? Is that done prior to updating the hash? And if so, where do we save the info that the hash has been changed?

aldeed:
@kieckhafer When publishing, you'll hash all the properties we care about for the product and save it on the `Product`. Then, when an admin is editing that product, client code will dynamically re-hash those same properties and compare to the saved hash. So we only ever need the "last published hash" stored because the "current product hash" is generated as needed from the product doc. (That comparison stuff is all part of #4281, not this issue.)

Issue #4262, opened 2018-05-22
API.Core.loadPackages() runs takes too long with many shops and packages


zrisher:
## Prerequisites
* [X] Are you running the latest version? - No, but the relevant code is still present in master
* [X] Are you able to consistently reproduce the issue?
* [X] Did you search the issue queue for existing issue? [Search issues](https://github.com/reactioncommerce/reaction/issues)

## Issue Description
The `loadPackages()` function in [API.Core](https://github.com/reactioncommerce/reaction/blob/master/server/api/core/core.js#L953) takes approximately 83% of our 35 minute boot time in production. We have 71 shops and 62 packages, of which 13 packages are custom. 

After adding logging and watching the process, no particular shop/package combination takes a disproportionate amount of time. Rather it seems the number of times certain loops must be run that leads to the overall run length. 

The top level of the loop on lines 985-1038, for example, must run 4402 times for our app, because that is the number of entries in the `Packages` collection, since once is created for each Shop X Package combo.

## Steps to Reproduce
1. Start a new application
2. Add a large number of shops (at least 71) and packages (at least 13).
3. Run the application

## Possible Solution
Clearly one solution would be to use less Shops. We are definitely not currently using it as it was originally intended, and we do plan on moving to an implementation with less shops. However, this seems to be a structural issue with the code base if it provides an upper limit on the number of shops the application can use and still run well.

It's also possible that our version of the code is so outdated that this is no longer an issue in production - I've scanned through the differences of `git diff a70495c3c0fe3e3ca1a36d3c31ad075703d87d91 server/api/core/core.js` and it doesn't appear to me anything has changed that would significantly affect this, but I could be mistaken. Unfortunately it's not an easy task to update our whole project to the latest reaction, but if you believe this would make a major difference I can focus on upgrading this particular area of the code.

Another option would be to not run this function as often - it seems its only duty is to update the packages collection with the info from provided fixtures and register.js files. If we know for certain these things haven't changed (e.g. if you're starting any server after the first on a multi-server deploy), it shouldn't be necessary to run this function. Perhaps Reaction could store git hashes of previous loads and skip if already loaded (and git hash isn't dirty).

### Versions
```
Node: 4.8.4
NPM: 2.15.11
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.29.0
Reaction: 1.5.3
```

aldeed:
Resolved in 3.0

PR #4383, opened 2018-06-28
[WIP] feat: Indicator to notify of pending product changes not yet published to catalog

### LINKED ISSUE #4281: Add an indicator to display whether a product has been changed since last published
Using the hash created in #4264 add an indicator to the PDP to inform the user of unpublished changes to a product. The indicator should show if the product or any of it's variants or options has unpublished changes.

Edit By @mikemurray

Historically we've used "dots" with edit icons to indicate changes

basic_reaction_product__edited_

The home for these "indicators", whatever form they take, should be placed in the toolbar either next to the publish button or take the form of the publish button.

basic_reaction_product__edited_

kieckhafer:
Resolves #4281 
Impact: **minor**  
Type: **feature**

## Issue
With the removal of revisions, we no long have a way to know if product data has been published to the `Catalog`, or if the `Product` collection contains updated data that has not been published.

## Solution
On the PDP, we create a hash of the current data on the page, using the same functions created in #4264. We then compare this new hash to the hash saved on the product object (also created in #4262). If the hashes match, the data is consistent between the two collections. If the hashes do not match, it means there are changes waiting to be published, and we display an indicator on the Publish button.

ticket_to_anywhere_t-shirtss

## Breaking changes
None.

## Testing
1. Create a new product
1. Publish this product to the catalog
1. Update the title (or any field that is published to the catalog)
1. See the indicator appear on the `Publish` button
1. Press the publish button
1. See the indicator go away

### DIFF
--- a/imports/plugins/core/catalog/client/components/publishControls.js
+++ b/imports/plugins/core/catalog/client/components/publishControls.js
@@ -313,7 +313,7 @@ class PublishControls extends Component {
   }
 
   renderHashCalculation = () => {
-    const productDocument = this.props.documents && this.props.documents[0];
+    const productDocument = this.props && this.props.documents && this.props.documents[0];
 
     if (productDocument) {
       Meteor.call("products/getpublishedProductHash", productDocument._id, (err, result) => {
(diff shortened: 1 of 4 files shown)


spencern:
This is great! Nice work @kieckhafer 

@nnnnat I'm def stealing that gif for the release notes.

PR #4390, opened 2018-06-29
Cart checkout schemas

aldeed:
Impact: **minor**  
Type: **feature**

## Changes
Add Cart checkout related GraphQL schemas.

## Breaking changes
NONE

### DIFF
--- a/imports/plugins/core/graphql/server/no-meteor/schemas/payment.graphql
+++ b/imports/plugins/core/graphql/server/no-meteor/schemas/payment.graphql
@@ -32,8 +32,8 @@ extend type Query {
   active/inactive, IP/region, shop, etc. To get the full list, use the `paymentMethods`
   query with proper authorization.
   """
-  availablePaymentMethods(shopId: ID!): [PaymentMethod]
+  availablePaymentMethods(shopId: ID!): [PaymentMethod]!
 
   "Get a full list of all payment methods"
-  paymentMethods(shopId: ID!): [PaymentMethod]
+  paymentMethods(shopId: ID!): [PaymentMethod]!
 }
--- a/imports/plugins/core/graphql/server/no-meteor/schemas/payment.graphql
+++ b/imports/plugins/core/graphql/server/no-meteor/schemas/payment.graphql
@@ -1,19 +1,29 @@
-"Payment plugins should extend or override this"
-type PaymentMethodData {
-  """
-  Example: savedCardId. Other examples might be transaction IDs, wallet keys.
-  Do not save full card details entered by a shopper here.
-  """
-  savedCardId: String
+"Payment plugins will extend or override this"
+enum PaymentMethodName {
+  example
+  exampleSavedCard
 }
+
+"An example"
+type ExamplePaymentMethodData {
+  example: String!
+}
+
+"An example"
+type ExampleSavedCardPaymentMethodData {
+  cardId: String!
+  cardDisplayName: String!
+}
+
+union PaymentMethodData = ExamplePaymentMethodData | ExampleSavedCardPaymentMethodData
 
 "Describes a payment method"
 type PaymentMethod {
-  "The payment method type. Any valid type that has been registered by a payment plugin. e.g., saved_card"
-  type: String!
-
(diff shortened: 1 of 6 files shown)


spencern:
@ticean did you have comments on this schema PR before we merge?

imports/plugins/core/graphql/server/no-meteor/schemas/payment.graphql
```diff
@@ -0,0 +1,29 @@
+"Payment plugins should extend or override this"
+type PaymentMethodData {
+  """
+  Example: savedCardId. Other examples might be transaction IDs, wallet keys.
+  Do not save full card details entered by a shopper here.
+  """
+  savedCardId: String
+}
```
ticean: This is an example but in a hard type definition. Does it make sense to define a Union `PaymentMethodData` along with a `StripePaymentMethodData`? Would that the best way to define a valid schema for this now that can evolve as new payment methods are added? What do you think, @aldeed?

ticean: Yes, the stitching by the payment system sounds good to me. I hope we can avoid all new instances of `JSONObject`.

`UNION` is a good fit here because we can't define inheritable fields on them. So we don't need to define some base kind of payment provider data. We just need to define one that will work as an example for now.

To keep the example completely abstract we could include an `ExamplePaymentMethodData` and put it in the union. Include any attributes you want. `savedCardId` would be fine.

By doing that here we define *how* the schema works and *how* it can be extended. We can drop in a real implementation in the future.

aldeed: @ticean I added the "Example" prefix but I can't do it as a union unless we make two of them. I could union it with JSONObject, but that seems to defeat the purpose. I'm still leaning toward making it JSONObject in here because nobody will see it. As long as you have the payment service spun up and stitched in, which it will be by default, you'll see our default Stripe schema pasted over it.

aldeed: Good point about needing the `on`. Given that, I don't think this can be a union because our client code would need to use `on` if there are 2+ payment plugins or not use `on` if there's just one.

So now I think we should:
- Remove "Example" prefix. We'll have only a single type with a single example property
- Plugins will extend that _type_ to add additional properties they need (rather than altering the data type of `data` prop)

So the base is something like this:

```graphql
type PaymentData {
  transactionId: String
}
```

```graphql
input PaymentDataInput {
  transactionId: String
}
```

And the Stripe payment plugin schema that gets stitched in would have something like:

```graphql
extend type PaymentData {
  savedCardId: String
  # etc.
}
```

```graphql
extend input PaymentDataInput {
  savedCardId: String
  name: String
  # etc.
}
```

ticean: Hi @aldeed. If we merge additional properties into a single type then it won't be possible to require any of them because a required field for payment type A would case input for payment type B to fail as it wouldn't have A's required field. 

I still think a union is the way to go. The example payment (the default) and an example plugin would give us two payment methods to use. This means that the payments would be in the schema but doesn't require that they be enabled. The payments service can overwrite the union with any payment types and remove the example plugin.

When using a *merged type* without required properties we would rely on app code to validate the fields in the payload. When using a *union*, as I suggest, we would rely on app code to check the enabled status of the plugin, which I think would be easier but most importantly would provide more useful information in the schema to the consumer and edge validation against that schema.

---

We could also consider lifting the polymorphism from the `PaymentData` up to the `PaymentMethod`. Meaning a union on the `PaymentMethod`.

The reason is that even with strongly typed data there's no guarantee by the type system that it matches the `PaymentMethod.type`.

I've been working on experimental Avro types which have similar requirements. There, I have a type with an enum property that has a single value, and a typed data property. It's sure to enforce that the type as a whole is correct.

aldeed:
@spencern @ticean For reference, here is the GraphQL issue about adding support for input unions: https://github.com/graphql/graphql-js/issues/207

PR #4392, opened 2018-06-29
bugfix: Archived products not being removed from Catalog

### LINKED ISSUE #4347: Archived products don't get removed from the Catalog
## Prerequisites
* [ ] Are you able to consistently reproduce the issue?

## Issue Description
Products that have been archived may still appear in the catalog to customers.

## Steps to Reproduce
1. As an operator publish a product
2. As a customer load the product grid and view that product
3. As an operator archive and publish that same product
4. As a customer, check to see if the product is still visible in the grid

## Possible Solution
It seems that the product publication method may not consider archival status for products, variants, and or options.

### Versions
```
Node: 9.11.1
NPM: 5.6.0
Meteor Node: 8.11.2
Meteor NPM: 5.10.0
Reaction CLI: 0.29.0
Reaction: 1.13.0
Docker: 18.03.1-ce
```

kieckhafer:
Resolves #4347   
Impact: **major**  
Type: **bugfix**

## Issue
Products that have been archived sometimes still appear in the catalog that customers see.

When a product is archived, the first thing that happens is it is updated in the `Products` collection to be `isDeleted: true`.

If you are on the PDP, you are immediately re-routed to the Product grid after archiving, so you don't have a chance to "Publish" the product to the `Catalog` collection, so the product in the `Catalog` collection never gets that updated `isDeleted` value - it still has the state of the product from before the archive / the last publication.

Even when you "Publish All", from the Product grid, you're only publishing the products that are available inside the `Products` publication, since the Admin gets the `Products` publication, while a non-admin gets the `Catalog` publication. Since the `isDeleted: true` is updated in the `Products` collection correctly, it's no longer on the grid, and therefore isn't published to the `Catalog` when you `Publish All`.

If you archive a product from the Product Grid, you will most likely not notice this error if you press "publish" at any time before navigating off the grid. This is because the Product Grid is not reactive, so even after you archive a product, the product is still visible on the Grid until you navigate away from the grid and refresh the publication.

## Solution

Add a `Hook` to the `products/archiveProduct` method that always archives the `Catalog` version of the product along with the `Products` version when it is archived.

## Breaking changes
None.

## Testing
1. Open a few browsers sessions, one as an admin, one as a logged out user (incognito probably), and one that is using the Starterkit (to make sure the GraphQL data is being updated).
1. Create a few different products
1. See that all the products show on all your sessions
1. Archive a product from the Product grid
1. See that the product is archived over all your sessions (you may need to refresh on the starterkit)
1. Archive a product from the PDP
1. See that the product is archived over all your sessions (you may need to refresh on the starterkit)

### DIFF
--- a/imports/plugins/core/catalog/server/hooks/hooks.js
+++ b/imports/plugins/core/catalog/server/hooks/hooks.js
@@ -1,0 +1,25 @@
+import { Catalog } from "/lib/collections";
+import { Hooks } from "/server/api";
+
+/**
+ * @summary archives product in the Catalog collection when it is archived in Products collection
+ * @param {String} userId - userId of user making the call
+ * @param {Object} doc - product document
+ * @return {undefined}
+ * @private
+ */
+Hooks.Events.add("afterRemoveCatalogProduct", (userId, doc) => {
+  // If this is a parent product, update the `Catalog` collection
+  // to set `isDeleted` to true
+  if (doc.type === "simple") {
+    Catalog.update({
+      "product.productId": doc._id
+    }, {
+      $set: {
+        "product.isDeleted": true
+      }
+    });
+  }
+
+  return doc;
+});
--- a/imports/plugins/core/catalog/server/hooks/hooks.js
+++ b/imports/plugins/core/catalog/server/hooks/hooks.js
@@ -9,7 +9,7 @@ import { Hooks } from "/server/api";
  * @private
  */
 Hooks.Events.add("afterRemoveCatalogProduct", (userId, doc) => {
-  // If this is a parent product, update the `Catalog` collection
+  // If this is a top level product, update the `Catalog` collection
   // to set `isDeleted` to true
   if (doc.type === "simple") {
     Catalog.update({
--- a/imports/plugins/core/catalog/server/methods/catalog.js
+++ b/imports/plugins/core/catalog/server/methods/catalog.js
@@ -937,6 +937,9 @@ Meteor.methods({
       return numFlaggedAsDeleted;
     }
 
+    // We need this hook here, in addition to the similar hook inside the `productsWithVariants.forEach` loop.
+    // The previous hook updates the data of a `variant` product (i.e. `type: variant`) only.
+    // This hook here updates the data of a parent product (i.e. `type: simple`).
     Hooks.Events.run("afterRemoveCatalogProduct", this.userId, product);
 
     Logger.debug(`${numFlaggedAsDeleted} products have been flagged as deleted`);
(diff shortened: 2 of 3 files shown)


imports/plugins/core/catalog/server/methods/catalog.js
```diff
@@ -937,6 +937,8 @@ Meteor.methods({
       return numFlaggedAsDeleted;
     }
 
+    Hooks.Events.run("afterRemoveCatalogProduct", this.userId, product);
+
```
aldeed: It looks like this would have been already done during `productsWithVariants.forEach` above?

imports/plugins/core/catalog/server/hooks/hooks.js
```diff
@@ -0,0 +1,25 @@
...
+  // to set `isDeleted` to true
+  if (doc.type === "simple") {
+    Catalog.update({
+      "product.productId": doc._id
+    }, {
+      $set: {
+        "product.isDeleted": true
+      }
+    });
+  }
```
aldeed: Did you test toggling visibility/archive of variants and options, without deleting the parent product? Pretty sure I tested that back when we changed the catalog schema, but I just want to make sure that works, too.

willopez: @aldeed  @kieckhafer  I have verified that visibility and archiving of options works. 👍

imports/plugins/core/catalog/server/methods/catalog.js
willopez: @aldeed the loop flags the product as deleted in the `Products` collection, the hook does the same in the `Catalog` collection.

imports/plugins/core/catalog/server/methods/catalog.js
```diff
@@ -937,6 +937,8 @@ Meteor.methods({
       return numFlaggedAsDeleted;
     }
 
+    Hooks.Events.run("afterRemoveCatalogProduct", this.userId, product);
```
willopez: Lets add a comment on this hooks, to make clear that it is updating the product in the `Catalog` collection.

imports/plugins/core/catalog/server/methods/catalog.js
kieckhafer: @willopez already said it but yes, it does run in the variants, but it wasn't running for the product itself, and since we only push the parents to the `Catalog`, we need to run it for the overall product as well.

kieckhafer:
@willopez added a comment to explain why we need that hook

aldeed:
@kieckhafer `productsWithVariants` looks like it should include the top-level product in addition to the variants. The query does `$or` on the `_id`, which would return the top-level, so hook would now run twice for top-level? Maybe I'm missing something?

Can you also fix this?

```
isDeleted: {
  $in: [false, undefined]
}
```

to be:

```
isDeleted: {
  $ne: true
}
```

To make sure we're covering null and keeping the query fast.

PR #4411, opened 2018-07-09
(fix) #4247 Discount Codes React Table Not Rendering Correctly

### LINKED ISSUE #4247: Coupon list in Admin only shows first 10 coupon codes
## Issue Description
If you have more than 10 coupon codes you can't see them as the list only shows the first 10

## Steps to Reproduce
1. Create 11 coupon codes
1. Open the "Payment" panel
1. Observe that only the first 10 coupons are shown and there is no way to naviate to it

### Versions
reaction version: 1.11

```
Node: 8.9.4
NPM: 5.6.0
Meteor Node: 8.11.1
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
```

ajporlante:
Resolves #4247   
Impact: **major**  
Type: **bugfix**

## Issue
The `ReactTable` component for discount codes is not rendering correctly because it is subscribed to an incorrect count publication.

## Solution
Change the subscription being passed as string in the `customDiscountCodes`  template from `discounts-count` to `discount-codes-count`.

## Breaking changes
None

## Testing
1. In Payments, enable codes and create discount codes.
2. Verify that the filter and pagination works.

### DIFF
--- a/imports/plugins/included/discount-codes/client/templates/settings.js
+++ b/imports/plugins/included/discount-codes/client/templates/settings.js
@@ -3,10 +3,10 @@ import { Meteor } from "meteor/meteor";
 import { Template } from "meteor/templating";
 import { ReactiveDict } from "meteor/reactive-dict";
 import { AutoForm } from "meteor/aldeed:autoform";
-import { DiscountCodes } from "../collections/codes";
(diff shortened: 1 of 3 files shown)


dancastellon:
Confirmed that this works as expected

PR #4413, opened 2018-07-09
#4353 Plugin for auto-generated sitemaps

### LINKED ISSUE #4353: Generate dynamic sitemaps on recurring schedule.
## Summary
Create recurring job to generate sitemaps for product pages and tag/category product grid pages.
We need to be generating a `sitemap.xml` that site operators can submit to Google's webmaster tools, prerender.io, etc.

## Suggested approach
We should create a job to generate a dynamic sitemap.xml file that can be updated on a recurring basis. This initial sitemap should have the index `/` route, as well as all PDP routes and tag product grid routes.

The tag routes should be generated from all visible tags.

We should generate a PDP route for each visible, non-archived product in the Catalog.  We will not generate routes for variants or options at this time.

We should be able to look at this meteor plugin to get some ideas for how to do this quickly.
https://github.com/gadicc/meteor-sitemaps

Here's a next.js example from @nnnnat: https://gist.github.com/evantahler/ba75e277c756fce337b7370035b298e7

This initial approach can be focused on single-shop instances, but we would welcome a PR to add support for multi-shop marketplaces as well.

The solution here should work on the starter kit as well.

dancastellon:
Resolves #4353
Impact: **minor**  
Type: **feature**

## Issue
Reaction currently has no sitemap.xml - a file that allows search engines visibility into a website's pages. Without it, search engines must manually crawl and find pages through links.

## Solution
A sitemap generator plugin that creates and stores XML for a sitemap index, as well as sitemaps for tag pages, PDPs, and arbitrary URLs that can be added via an event hook.

There is a recurring job that runs every 24 hours (can be changed) that generates the sitemaps. There is also a button to manually trigger a refresh, at Dashboard -> Shop -> Options - along with a notification that appears when it's completed.

GraphQL TBD (separate issue)

## Breaking changes
None

## Testing
1. Reset the db (translations updated) and use `reaction-data-loader` to load the dev dataset
2. Create some products and tags - some archived/non-visible, some visible/published
3. Login as admin and visit Dashboard -> Shop Settings -> Options
4. Click "Refresh sitemap"
5. Confirm you get a notification (bell icon) as well as a Toast alert.
6. See Sitemaps collection and confirm correct XML by handle: sitemap.xml, sitemap-pages-1.xml, sitemap-tags-1.xml, sitemap-products-1.xml
7. Visit /sitemap.xml and confirm it loads and displays sitemap index XML
8. Visit each of the individual sitemap URLs and confirm they load/display valid XML
9. In the DB, confirm you see a Sitemaps collection with a document for each sitemap.
10. Confirm you see "BASE_URL" in place of any actual URLs in DB sitemaps
11. Without updating a tag or product, click the sitemap refresh button again and confirm that the sitemaps for tags and products did not regenerate, but that the sitemap index (sitemap.xml) and the one for pages, has regenerated.
12. Update a tag's title (I did it by editing the main tag nav), refresh sitemap, and confirm the tag sitemaps did regenerate, but not product sitemaps
13. Update a product field (such as the title), refresh sitemap, and confirm product sitemaps did regenerate
14. Run against retailer and miderprise databases - confirm job runs in background (doesn't stop app UI from functioning), and creates the correct number of sitemaps (1k URLs per sitemap max).
15. At Shop -> Options, change the "Sitemap refresh period" value to something really short, like "every 30 seconds"
16. Restart the server for the job schedule to update (this appears to be the same as other refresh period settings, like open exchange rates)
17. Confirm the job runs as scheduled.


zenweasel:
@dancastellon You've got some lint problems here. Make sure it passes `npm run lint`

imports/plugins/included/sitemap-generator/server/lib/generateSitemaps.js
```diff
@@ -0,0 +1,144 @@
+import { Meteor } from "meteor/meteor";
+import { check, Match } from "meteor/check";
+import Hooks from "@reactioncommerce/hooks";
+import Reaction from "/imports/plugins/core/core/server/Reaction";
```
zenweasel: This should be imported from `/server/api`

imports/plugins/included/sitemap-generator/server/lib/generateSitemaps.js
```diff
@@ -0,0 +1,144 @@
...
+  const sitemapPageUrls = urls.slice(startIndex, endIndex)
+
+  sitemapPageUrls.forEach((url) => {
+    xml += `
+      <url>
+        <loc>URL</loc>
+        <lastmod>LAST_MOD</lastmod>
+        <changefreq>daily</changefreq>
+        <priority>0.8</priority>
+      </url>`.replace("URL", url);
```
zenweasel: Are we replacing `LAST_MOD` somewhere?

imports/plugins/included/sitemap-generator/server/lib/generateSitemaps.js
dancastellon: @zenweasel Release 1.14 changes that. It looks like all references to `/server/api` have been moved to `/imports/plugins/core/core/server/Reaction` it looks like.

spencern: as @dancastellon mentioned, there is no more `/server/api` in release 1.14

imports/plugins/included/sitemap-generator/server/lib/generateSitemaps.js
dancastellon: @zenweasel Yes, see `getSitemapXML` below.

dancastellon:
@zenweasel Today I was able to test the sitemap generator with the miderprise dataset. It took roughly 3.5 minutes, but the app was useable while it was generating. I believe this one's ready for review.

zenweasel:
@dancastellon Following up on the notes from the call, if I click on "Refresh Sitemaps Now" I don't see any visual feedback that anything has started, much less completed in either the server console or the UI. I should *at least* some sort of "started regenerating sitemap" message or something.

zenweasel:
Shouldn't I be seeing these value substituted in the actual XML?

imports/collections/schemas/registry.js
```diff
@@ -282,7 +282,8 @@ export const CorePackageConfig = PackageConfig.clone().extend({
   },
   "settings.openexchangerates.appId": {
     type: String,
-    label: "Open Exchange Rates App Id"
+    label: "Open Exchange Rates App Id",
+    optional: true
```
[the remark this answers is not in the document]
dancastellon: @zenweasel Without making this field optional I'm unable to save the shop options form, unless I add an App ID for Open Exchange Rates. Shop options is where I added the sitemap refresh setting & regenerate button. I just noticed that on a fresh install, I also had to make `settings.cart` in the same schema optional. Does it make sense to have the sitemap settings be in a separate form? Or perhaps somewhere else, like the Shop -> General form?

imports/plugins/included/sitemap-generator/server/lib/getSitemapXml.app-test.js
```diff
@@ -0,0 +1,29 @@
+import Random from "@reactioncommerce/random";
```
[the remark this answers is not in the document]
dancastellon: Thanks @zenweasel - updated all filenames to be hyphenated

imports/plugins/included/sitemap-generator/server/methods/index.js
```diff
@@ -0,0 +1,23 @@
+import { Meteor } from "meteor/meteor";
```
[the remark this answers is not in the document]
dancastellon: @zenweasel I updated this to follow the core plugin's pattern for defining methods. Method itself in a separate file, that is simply imported in methods/index.js and exported.

dancastellon:
@zenweasel Thanks for the review. To follow up on your comments:

- Fixed linting issues: I had missed files I touched outside of my new plugin
- I always get an alert that reads "Sitemap refresh initiated" when I click "Refresh sitemaps now". The `Alerts.toast` call has always been in the `Meteor.call` callback in `generate-sitemaps-button.js`. I have seen in previous versions of RC alerts stop working altogether. Do you see an alert when you click the main "Save changes" button on the shop options form?
- Seeing BASE_URL & LAST_MOD in the XML saved to the DB - this is intended. It isn't until the XML is served up via `handle-sitemap-routes.js` that those variables are replaced. This is for 2 reasons: so that BASE_URL can easily be replaced when the XML is loaded via GraphQL in the storefront (since the Meteor app & nextJS app will be hosted on separate URLs), and so that LAST_MOD always reports today's date.

zenweasel:
@dancastellon Why would `LAST_MOD` be today's date? Shouldn't it always be when it was last modified? Aren't you going to needlessly force a full re-index?

zenweasel:
@dancastellon We had also talked getting feedback when the creation/update is completed. Vishal said he would like to see that implemented. Also we talked about having a direct link to the sitemap file.

dancastellon:
@zenweasel Yes, the bell in the header that usually shows order notifications.

rymorgan:
@dancastellon I get what you are trying to do here but the bell icon is for storefront customers not for admin folks so this would be a confusing UX for an operator. I'm currently having all kinds of problems getting your branch running so have not been able to preview the work you've done so far but my take would be -- Use a toast alert with a spinner that stays active until the sitemap is generated. The toast alert should notify the user that it could take up to X minutes for the sitemap to be generated.

zenweasel:
@rymorgan Actually the notification icon was created for admin notifications and that's how order notifications are currently sent so I don't think that would be a shocker for an admin. I think what Dan was going for was something where you didn't need to remain on the page to see when it wold be finished.

zenweasel:
@dancastellon I spoke with @rymorgan and he feels that this isn't the right place for this notification and is probably going to go away so let's just go with a global toast if you can, or a status indicator inside the panel (e.g. "Sitemap generation finished at 13:53" or something to that affect)

dancastellon:
@rymorgan @zenweasel Thanks for the feedback. I'll look into a global toast or status indicator inside the panel.

dancastellon:
@rymorgan @vwadher @zenweasel See these screenshots for the sitemap admin UI, let me know if I should change anything. Thanks!

shop-options-sitemap

sitemap-refresh-toast

dancastellon:
@zenweasel I believe this one to be ready for your review.

dancastellon:
Thanks @rymorgan, I've made the changes you suggested, see below:

updated-sitemap-buttons

updated-notification

dancastellon:
@rymorgan Also, the notification. Let me know if I should change the text:

notification

rymorgan:
@dancastellon I think the notification should say 'Sitemap refresh is complete' to match the language in the toast

Other than that it LGTM. The toast alert was coming up for me but could very well be something wrong with my local reaction instance so as long as @zenweasel can confirm that is working then I'm good with this.

dancastellon:
@rymorgan Done, thanks. I have seen times previously where the toast doesn't show up sometimes. @zenweasel Let me know if you still don't see the toast. Might be a core bug.

zenweasel:
If I put in an invalid interval I don't get any validation but it throws this error in the server console:
`No valid available later.js times in schedule after Fri Jul 27 2018 13:55:06 GMT+0800 (+08)`

zenweasel:
More Issues:
1. When I click on "Refresh sitemap" I get the translation key instead of the message.
1. The file referenced in the main file doesn't exist "http://localhost:3000/sitemap-pages-1.xml"

zenweasel:
This is a nitpick probably but: If I click on "View All" for notifications, each of the notifications is a link but that link just takes you to the home page

zenweasel:
@dancastellon Found a few little things to fix but it's getting pretty close

dancastellon:
@zenweasel I've made the changes you suggested. Tagging @rymorgan as I've switched out some form fields with their new Reaction Components counterparts

1. Used a select box to let the admin choose from 3 options - every 24 hours, 12 hours, 1 hour
2. Linked the notification to sitemap.xml
3. Added default values to translations so that Toast alert never shows the i18nkey

sitemap-settings-new-components

rymorgan:
@dancastellon Is this still under shops/options? I'm not seeing it anymore on your brand (might be user error). 

From your screenshot though I think this approach is good. And every place where we convert to components in our component library is a win. 👍

dancastellon:
@rymorgan Yes, it's still under Shop -> Options. I did rename the React component behind the sitemap settings, so maybe try a hard refresh?

rymorgan:
Ok, I got it working. From a UI/UX perspective, it looks good to me 👍

PR #4439, opened 2018-07-16
ActionView Component Typos

pmn4:
just something I noticed while debugging

Resolves #n/a
Impact: **minor**  
Type: **bugfix**

## Issue
Noticed two typos in this file. One is not a big deal, as it was a locally defined & used function, other was opaticy, er, opacity.

### DIFF
--- a/imports/plugins/core/dashboard/client/components/actionView.js
+++ b/imports/plugins/core/dashboard/client/components/actionView.js
@@ -356,19 +356,19 @@ class ActionView extends Component {
         display: "flex",
         position: "absolute",
         left: 20,
-        opaticy: 1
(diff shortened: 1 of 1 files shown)


zenweasel:
@pmn4 I know the function typo shouldn't matter, but should I see the change in `opacity` somewhere?

pmn4:
tbh, I'm not sure how to see it. It seems to be in a sub panel of the action view? not really sure.
this isn't a great answer, but the animation happens in 200ms, so I'm not sure you'd really even notice the opacity change (especially if there is an additional "reveal" animation). sorry I couldn't be more help!

Issue #4441, opened 2018-07-16
Eliminate dependency on Velocity


spencern:
## Issue Description
We currently have a dependency on `Velocity` for animating parts of our client application.
Velocity is a DOM animation library which makes doing some complex animations performant and simple.
We currently use velocity to do one of two things.
1. Animate the cart drawer opacity
2. Animate the background color of changed fields from green and then back to white.

Neither of these animations should need a library to do well or simply and don't justify the added overhead of [94.5kb](https://bundlephobia.com/result?p=velocity-react@1.4.1)

The animations using velocity are contained in one of 6 files
- containers/cartIconContainer.js
- components/metadata/metafield.js
- components/tagItem.js
- components/productAdmin.js
- components/productField.js
- components/variantForm.js

This issue is to eliminate the dependency on Velocity and replace any animations that are using Velocity with another method of animation.

dancastellon:
@zenweasel Does it make sense to try CSS animations or is jQuery fine?

zenweasel:
@dancastellon No jQuery ever.

spencern:
Animating with Javascript is fine though, I created this ticket because I don't think the type of animations we're performing warrant an animation library. Honestly this is not a high priority if it turns out to be a tough task.

PR #4442, opened 2018-07-16
chore: update version of velocity-react to 1.4.1

spencern:
Impact: **minor**  
Type: **chore**

## Issue
The version of `velocity-react` we've been using is out of date and depends on an old version of `lodash` which has a snyk reported medium severity vulnerability.

## Solution
This bumps the version of `velocity-react` to the latest version which includes updating the lodash dependency

## Breaking changes
none

## Notes
There are only 6 places in the app that we use Velocity and it's a heavier package. It would be ideal to eliminate the dependency on this package entirely and animate using CSS transitions where possible or eliminate complex animations requiring a package like this from core. That task has been created as a ticket here: https://github.com/reactioncommerce/reaction/issues/4441

## Testing
The animations using velocity are contained in one of 6 files
- containers/cartIconContainer.js
- components/metadata/metafield.js
- components/tagItem.js
- components/productAdmin.js
- components/productField.js
- components/variantForm.js

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3043,7 +3043,8 @@
     "buffer-crc32": {
       "version": "0.2.13",
       "resolved": "https://registry.npmjs.org/buffer-crc32/-/buffer-crc32-0.2.13.tgz",
-      "integrity": "sha1-DTM+PwDqxQqhRUq9MO+MKl2ackI="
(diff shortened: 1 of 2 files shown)


spencern:
This is replaced by #4444

PR #4446, opened 2018-07-17
make the snyk-security step a dependency for the docker-build step; s…

griggheo:
Type: **test**

## Issue
We would like to stop docker from building the reaction image if the snyk security step fails.

## Solution
- modified the CircleCi workflow and specified the snyk-security step as a dependency for the docker-build step.


griggheo:
Thanks for the review Spencer. The snyk tests are skipped because my branch was not a PR at the time I was testing. I added a new check for TARGET_BRANCH and if it's the string 'null' then it means it's not a PR. The snyk test step exits in this case with code 0 and prints "Not a PR. Skipping Snyk."

What is the next step here, do I need to do anything or can we merge?

Issue #4407, opened 2018-07-06
Indexes missing in stock reaction


Akarshit:
## Issue Description
From the results of load-testing, it seems that the following index need to be created for a production site to work.

1. `db.Catalog.createIndex({ createdAt: 1 })`
1. `db.Catalog.createIndex({ "product._id": 1 })`
1. `db.Catalog.createIndex({ "product.variants._id": 1 })`
1. `db.Catalog.createIndex({ "product.variants.options._id": 1 })`
1. `db.Products.createIndex({ "ancestors" : 1 })`
1. `db.Orders.createIndex({ cartId: 1, userId: 1 })`
1. `db.Jobs.createIndex({ status: 1, expiresAfter: 1 })`
1. `db.Assets.createIndex({ type: 1 })`
1. `db.users.createIndex({  roles: "hashed" })`
1. `db.users.createIndex({  "roles.__global_roles__": 1 })`
1. `db.users.createIndex({  "roles.J8Bhq3uTtdgwZx3rz": 1 })`
1. `db.Accounts.createIndex({ groups: 1 })`
1. `db.cfs.Media.filerecord.createIndex({ "metadata.type" : 1 })`
1. `db.cfs_gridfs.small.chunks.createIndex({ files_id: 1, n: 1 })`
1. `db.cfs_gridfs.image.chunks.createIndex({ files_id: 1, n: 1 })`
1. `db.cfs_gridfs.large.chunks.createIndex({ files_id: 1, n: 1 })`
1. `db.cfs_gridfs.thumbnail.chunks.createIndex({ files_id: 1, n: 1 })`
1. `db.cfs_gridfs.medium.chunks.createIndex({ files_id: 1, n: 1 })`

Will it make sense to include some of these(if not all) in the stock reaction code.
If this is not included in stock `reaction` can we at-least mention these indexes in the docs?

### Versions
v1.13

aldeed:
For Catalog, I think we more often look up by `"product.variants.variantId"` and `"product.variants.options.variantId"` rather than `"product.variants._id"` and `"product.variants.options._id"`. But a solution to that better than indexes on all of these would be saving a `productIds` array at the top level of the catalog item (indexed) and querying that. Then all of the queries should be faster in general, too.

After my current PR is merged, the orders one should be `{ cartId: 1, accountId: 1 }` rather than `{ cartId: 1, userId: 1 }` because user -> account.

spencern:
[...]

PR #4449, opened 2018-07-17
#4384 Improve app startup time when large number of Accounts/Users exists

### LINKED ISSUE #4384: Number of documents in Account collection affecting startup time
## Issue Description
As the number of documents(belonging to group "customer") are increased in the database, the startup time of reaction slows down.
Here are some stats
1. Number of Accounts - 100,000 - startup time - 30sec
1. Number of Accounts - 300,000 - startup time - 120sec
1. Number of Accounts - 500,000 - startup time - didn't start in 5 mins
1. Number of Accounts - 2,100,000 - startup time - got `Range Error` which is error when the you exceed the size limit of mongo.

## Steps to Reproduce
To reproduct these, you would have to connect the load-testing mongo server. Please contact me for that.

## Possible Solution
Maybe we are getting all Accounts somewhere during the startup or doing something similar.

dancastellon:
Resolves #4384  
Impact: **major**  
Type: **performance**

## Issue
When an RC install with a large number of Accounts/Users starts up, currently every user is loaded in a single query and updated in another query, to add roles to their accounts. The core function in question - `addRolesToGroups` - is called 3 times throughout the app and is intended to add new roles to groups and their associated accounts/users. Even if the roles passed to the function call are already in the groups passed, all Accounts are loaded and updated.

## Solution
The solution: 1) make it so Groups and Accounts/Users are only updated if the given roles aren't already present in the group, 2) batch the Accounts find query and Users update query to update a maximum of 100k Accounts/Users per call (to avoid a RangeError), and 3) have the heaviest queries run asynchronously in the background so that the app can finish starting up and start serving user requests.

There is one catch: With larger datasets (2,100,000 Accounts/Users tested), although the heaviest queries are run in the background, and the app is able to load, it is noticeably slower than usual while clicking through pages, due to the Mongo load. On the other hand, this large processing is now only done on the first startup. Subsequent startups no longer add roles to 2.1m accounts, because they already exist.

## Breaking changes
None

## Testing
1. Use reaction-data-loader to load the `ret` (retailer) dataset
2. Start the app with `export METEOR_DISABLE_OPTIMISTIC_CACHING=1 && REACTION_LOG_LEVEL="DEBUG" reaction`
3. Confirm in logs that "Reaction initialization finished" takes less less than 15s
4. Confirm you can visit app in browser and click around
5. Confirm you see debug messages about the group batch processing. Ex: `Processing user role update #2 of 21 for customer group`, but can still use the app
6. When finished processing, restart the app
7. Confirm you see messages such as `Skipping roles already assigned to...`, and that once the app starts up, there are no `Processing user role updates...` messages
8. Confirm roles are still added to correct groups and users


imports/plugins/core/core/server/Reaction/addRolesToGroups.js
```diff
@@ -77,17 +29,76 @@ export default function addRolesToGroups(options = { allShops: false, roles: [],
...
+ */
+function addRolesToGroupAndUsers({ _id, shopId, name }, roles) {
+  Groups.update({ _id }, { $addToSet: { permissions: { $each: roles } } });
+
+  const maxAccountsPerUpdate = 100000;
+  const accountSelector = { groups: _id };
+  const accountOptions = { fields: { _id: 1 } };
+  const numAccounts = Accounts.find(accountSelector, accountOptions).count();
+  if (numAccounts === 0) {
+    Logger.debug("No users need roles updated");
```
Akarshit: I think there should be `return` statement here.

imports/plugins/core/core/server/Reaction/addRolesToGroups.js
```diff
@@ -77,17 +29,76 @@ export default function addRolesToGroups(options = { allShops: false, roles: [],
...
+  }
+  Logger.debug(`Number of users to add roles to: ${numAccounts}`);
+  const numQueriesNeeded = Math.ceil(numAccounts / maxAccountsPerUpdate);
+  Logger.debug(`Number of updates needed: ${numQueriesNeeded}`);
+  for (let inc = 0; inc < numQueriesNeeded; inc += 1) {
+    Logger.debug(`Processing user role update #${inc + 1} of ${numQueriesNeeded} for ${name} group`);
+    accountOptions.skip = inc * maxAccountsPerUpdate;
```
Akarshit: Skip queries are not a good option in mongo. They are very slow and keep becoming slower as the number of documents to skip increases.
To quote mongoDB documentation
>The cursor.skip() method requires the server to scan from the beginning of the input results set before beginning to return results. As the offset increases, cursor.skip() will become slower.`
>Range queries can use indexes to avoid scanning unwanted documents, typically yielding better performance as the offset grows compared to using cursor.skip() for pagination.

imports/collections/schemas/accounts.js
```diff
@@ -187,7 +187,8 @@ export const Accounts = new SimpleSchema({
   "groups": {
     type: Array, // groupIds that user belongs to
     optional: true,
-    defaultValue: []
+    defaultValue: [],
+    index: 1
```
Akarshit: I don't know what our current status is on adding indexes: https://github.com/reactioncommerce/reaction/issues/4407
But if you need this index for the script to perform efficiently we should add it.

My guess is that there aren't that many groups that it would need a compulsory index.

zenweasel:
@dancastellon What's the status on this, it's gone 6 days with no update.

imports/collections/schemas/accounts.js
dancastellon: @Akarshit While there aren't a lot of groups, there are a lot of accounts, which is I added this index to Accounts.groups. If I don't have this index, wouldn't Mongo have to scan all 2.1 million accounts? I'm running the retailer dataset

imports/plugins/core/core/server/Reaction/addRolesToGroups.js
```diff
@@ -1,55 +1,7 @@
...
-          $each: roles
-        }
-      }
-    };
-
-    // Update users
-    Meteor.users.update(selector, operation, { multi: true });
-  });
-}
+import Logger from "@reactioncommerce/logger";
```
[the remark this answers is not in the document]
dancastellon: Done

imports/plugins/core/core/server/Reaction/addRolesToGroups.js
```diff
@@ -77,17 +29,76 @@ export default function addRolesToGroups(options = { allShops: false, roles: [],
...
+ * @param {Array} roles - Array of roles/permissions (strings)
+ * @returns {undefined} Nothing
+ */
+function addRolesToGroupAndUsers({ _id, shopId, name }, roles) {
+  Groups.update({ _id }, { $addToSet: { permissions: { $each: roles } } });
+
+  const maxAccountsPerUpdate = 100000;
+  const accountSelector = { groups: _id };
+  const accountOptions = { fields: { _id: 1 } };
+  const numAccounts = Accounts.find(accountSelector, accountOptions).count();
```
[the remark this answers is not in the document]
dancastellon: @Akarshit Good catch, removed

imports/plugins/core/core/server/Reaction/addRolesToGroups.js
dancastellon: @Akarshit Ah yep, you're right

dancastellon:
@zenweasel This one's ready for your review. I made the changes @Akarshit had mentioned.

imports/plugins/core/core/server/Reaction/addRolesToGroups.js
```diff
@@ -77,17 +29,92 @@ export default function addRolesToGroups(options = { allShops: false, roles: [],
...
+    if (lastUserIdUpdated) {
+      accountSelector._id = {
+        $gt: lastUserIdUpdated
+      };
+    }
+
+    const userIds = Accounts.find(accountSelector, accountOptions).map((account) => account._id);
+    lastUserIdUpdated = userIds[userIds.length - 1];
+
+    const selector = { _id: { $in: userIds } };
```
Akarshit: @dancastellon Since we already have the range of `userIds` we want to update, wouldn't it make more sense to send a `in` filter in the `selector`? Something like `{ _id: { $gte: userIds[0], $lte: lastUserIdUpdated]} }`
Since the range is on a indexed field it will be executed pretty quickly and we won't have to send 100k values.

dancastellon: @Akarshit done, great catch!

zenweasel:
@aldeed I would like to get this one merged as well if possible. Thanks!

PR #4462, opened 2018-07-19
[WIP] (chore): run migrate on startup

### LINKED ISSUE #4416: Move to a non-Meteor migrations solution
### Proposed Work

This would be a first stage. Eventually we will want to move this out of the app entirely to an NPM package / CLI.

1. Remove the Meteor percolate:migrations package and add this NPM package, which has the same API: https://github.com/emmanuelbuah/mgdb-migrator
2. Change the imports to come from the NPM package.
3. Minor changes to the `config`, primarily to add the db connection string, which we should be able to get from `process.env.MONGO_URL`
4. Rewrite all the migrations so that they do not import or use any Meteor code. Also make all of the `up` and `down` functions `async`.

This can also fix https://github.com/reactioncommerce/reaction/issues/3037 by setting the latest version as the current version in the DB as part of initial fixture import.

See also some discussion here: https://github.com/reactioncommerce/reaction/issues/4362

Akarshit:
Resolves #4416  
Impact: **major**  
Type: **chore**

## Issue
The migration were dependent on meteor and we want to move away from being meteor dependent.

## Solution
Migrate the migration to `tj/node-migrate`

### DIFF
--- a/imports/plugins/core/versions/server/migrations/1_rebuild_account_and_order_search_collections.js
+++ b/imports/plugins/core/versions/server/migrations/1_rebuild_account_and_order_search_collections.js
@@ -1,6 +1,5 @@
 import Logger from "@reactioncommerce/logger";
-import { Migrations } from "meteor/percolate:migrations";
-import { OrderSearch, AccountSearch } from "/lib/collections";
+import rawCollections from "/imports/collections/rawCollections";
 import Reaction from "/imports/plugins/core/core/server/Reaction";
 
 let buildOrderSearch;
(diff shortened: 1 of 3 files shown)


Akarshit:
Conversation I had with Eric D. regarding migrations not passing through babel:

**Akarshit**: Hey, I was trying to make migrate the migrations from "meteor's Migrations" to `tj/node-migrate`. 
But I am facing a problem, when I specify the migration directory(`/migrations`) that is getting resolved to `/Users/akarshit....../.meteor/local/bundle......./migrations` so basically relative to the path where the `meteor` bin is present. 
I think this means that the package is getting those files outside of a "meteor" environment.
The bigger problem is the ES6 code in those migrations files is giving an error. I think that is because it is not getting transpiled into ES5 code by babel, because that happens for the code that is in "meteor" environment.
**Eric**: I think this will work to fix the location but I don’t have time to test it right now: `stateStore: path.resolve("assets/migrate/.migrate"),` That will find them, but I don’t think they’ll go through Babel. If not, you’ll just need to use `require()` instead of `import` in them.

zenweasel:
@Akarshit If we aren't going to be working on this in the near term let's close this.

PR #3332, opened 2017-11-20
Consideration for Marketplace Shops on Different URLs

pmn4:
This PR is to address Merchant Shops which live on a different URL than the Primary Shop.
For example, my Primary Shop is hosted at https://reserve.brewline.io and a Merchant Shop at https://twolights.brewline.io. Visiting the latter shows the branding and products of the Primary Shop.
Opening developer tools on the Merchant Shop and running `ReactionCore.getPrimaryShopId() === ReactionCore.getShopId()` yields `true` (the merchant shop has its `domains` attribute set properly).

Digging into the code yielded that `Meteor.absoluteUrl` relies on `$ROOT_URL` which is a problem when using the domain as an identifier. To solve this, I capture the domain/hostname during `onConnection` (server-side) and via `window.location.hostname` (client-side).

I then wrapped `Meteor.absoluteUrl` in a function called `Reaction.absoluteUrl`, passing a `rootUrl` option, and doing a repo-wide find/replace.

Lastly, on `Meteor.startup` (client side), I subscribe to `"Shop"` (previously, `"PrimaryShop"`) to ensure the shop is ready before rendering.

---
**_disclaimer:_** I learned a ton doing this, but there are definitely still gaps in my understanding of Meteor and Reaction. Feel free to highlight anything that may look fishy!

**_further disclaimer:_** I don't know all of the Reaction features yet, so there is a lot of regression testing to be done before merging.

lastly, where should I write unit tests for these new methods?
thanks for the consideration!

PR #4471, opened 2018-07-22
fix: fix client errors related to domain lookup changes

aldeed:
Impact: **minor**  
Type: **bugfix**

## Issue
When https://github.com/reactioncommerce/reaction/pull/3332 was merged into 1.14.0, it caused some client errors.

## Solution
These are mainly due to the horrible circular dependencies of the Reaction object. Also fixed, `shopDomain` function to return `.host` rather than `.hostname` to match the behavior of `Meteor.absoluteUrl`

## Breaking changes
None

## Testing
Verify that the app starts and you can navigate a few of the main routes.

### DIFF
--- a/client/modules/core/domains.js
+++ b/client/modules/core/domains.js
@@ -3,7 +3,8 @@ import { composeUrl } from "/lib/core/url-common";
 
 export const DomainsMixin = {
   shopDomain() {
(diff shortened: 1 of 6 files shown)


zenweasel:
Did a quick test of this and not seeing any errors

PR #4481, opened 2018-07-24
WIP: #4480 Convert product grid to consume GraphQL data

### LINKED ISSUE #4480: Convert product grid to consume data via GraphQL
## Work
Setup Apollo Client within the Meteor app, and convert the product grid to consume data via GraphQL, rather than Meteor publications. The solution should re-use Storefront code where possible.

dancastellon:
Resolves #4480  
Impact: **major**  
Type: **performance**

## Changes
This PR adds Apollo Client integration in our Meteor app, and allows the product grid to consume data via GraphQL. This is in progress, with some additions needed to the GraphQL API in order for the grid to be regression free (Search for TODO in the code).

## Breaking changes
TBD - depending on how custom shops have extended the grid, it may/may not break things.

## Testing
1. Load up the homepage
2. Confirm products load, infinite scroll works, clicking through to PDP - basically any and all functionality the Meteor grid has.
3. Confirm caching works - visit homepage, visit tag page, go back to homepage & confirm it loads instantly without another graphql network request


imports/plugins/core/router/client/initApollo.js
```diff
@@ -0,0 +1,26 @@
...
+    },
+  }));
+
+  return forward(operation);
+});
+
+
+export default function initApollo() {
+  return new ApolloClient({
+    link: meteorAccountsLink.concat(new HttpLink({ uri: Meteor.absoluteUrl('graphql-alpha') })),
```
aldeed: If latest 1.14.0 is merged into this, `Meteor.absoluteUrl` can be changed to `Reaction.absoluteUrl`. (Everywhere else in app has been changed.)

imports/plugins/included/product-variant/containers/productsContainerCustomer.js
```diff
@@ -3,118 +3,131 @@ import PropTypes from "prop-types";
...
-
-  const catalogCursor = Catalog.find({
-    "product.type": "product-simple"
-  }, {
-    $sort: {
-      [`product.positions.${tagIdForPosition}.position`]: 1,
-      createdAt: -1
-    }
-  });
+  // TODO graphql custom tag sort - $sort: { [`product.positions.${tagIdForPosition}.position`]: 1 }
```
aldeed: We no longer support custom positioning so everything using `product.positions` can just be removed.

imports/plugins/included/product-variant/register.js
```diff
@@ -8,7 +8,7 @@ Reaction.registerPackage({
   registry: [{
     route: "/tag/:slug?",
     name: "tag",
-    template: "products",
+    template: "Products",
```
aldeed: I believe this needs a migration since this goes into the database. See migration 27 for an example

package.json
```diff
@@ -31,6 +31,10 @@
     "@reactioncommerce/random": "1.0.1",
     "@reactioncommerce/schemas": "1.1.0",
     "accounting-js": "^1.1.1",
+    "apollo-cache-inmemory": "^1.2.5",
```
aldeed: All added/changed packages should be pinned (no `^` prefix)

aldeed:
@dancastellon This is super exciting! I realize it's WIP but I left a few comments since I'm going to be out the rest of this week. It can be a separate followup PR, but I'd also like to merge the two ProductGrids (the CustomerProductGrid from here and the one in next-starterkit) into a single component that lives in the component library. It would be mostly a copy of the newer starter kit component, but replacing MaterialUI components with our own styled-components divs. cc @machikoyasuda

dancastellon:
@aldeed Thanks! I have made the updates from your comments.

dancastellon:
Created #4486 after reviewing with Nat - just need a way to load a shop by slug.

spencern:
@dancastellon this is such a great start. Super excited to get this work merged.
Do you think we should split porting the UI from the storefront into a separate ticket or do that as part of this ticket?

zenweasel:
@spencern I wonder if we shouldn't wait until the next release to merge this?

spencern:
@zenweasel for sure this isn't going in v1.14.0

dancastellon:
@spencern Maybe a separate ticket. I'm not familiar with the new component library yet, but how would the porting work? Would @machikoyasuda create the styled-component product grid and add it to the library, then I port the Meteor grid to it? Or do you want me to take a stab at doing all of it? Either way - Thanks!

imports/plugins/core/graphql/lib/hocs/withShopId.js
```diff
@@ -0,0 +1,40 @@
...
+          <Component {...this.props} />
+        );
+      }
+
+      const variables = { slug: shopSlug };
+
+      return (
+        <Query query={getShopId} variables={variables}>
+          {({ loading, data }) => {
+            if (loading) return null;
```
aldeed: In general (for all of the HOCs) we should ideally render the component always, passing something like `isLoadingShopId` prop so that the wrapped component can render its own loading UI.

imports/plugins/core/graphql/lib/hocs/withPrimaryShopId.test.js
```diff
@@ -0,0 +1,32 @@
...
+import { mount } from "enzyme";
+import { MockedProvider } from "react-apollo/test-utils";
+import getPrimaryShopId from "../queries/getPrimaryShopId";
+import withPrimaryShopId from "./withPrimaryShopId";
+
+const fakeOpaqueShopId = "cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg==";
+const MockComponent = () => <div>Mock</div>;
+const TestComponent = compose(withPrimaryShopId)(MockComponent);
+
+test("renders child component with correct shop id", async () => {
```
dancastellon: @aldeed Can you take a look at my first unit test for these HOC components? I want to make sure I have good patterns here before I write tests for the others.

imports/plugins/core/graphql/lib/hocs/withPrimaryShopId.test.js
```diff
@@ -0,0 +1,32 @@
...
+  }];
+
+  const wrapper = mount(
+    <MockedProvider mocks={mocks}>
+      <TestComponent />
+    </MockedProvider>
+  );
+
+  // Wait for next tick to get past loading state
+  await new Promise(resolve => setTimeout(resolve));
```
aldeed: If you have the HOC pass down `isLoadingPrimaryShopId` boolean prop, then you can wait for that to be `false` here, which will be a little more foolproof than a timeout.

imports/plugins/core/graphql/lib/hocs/withPrimaryShopId.test.js
```diff
@@ -0,0 +1,32 @@
+import React, { Component } from "react";
+import { compose } from "recompose";
+import { mount } from "enzyme";
+import { MockedProvider } from "react-apollo/test-utils";
+import getPrimaryShopId from "../queries/getPrimaryShopId";
+import withPrimaryShopId from "./withPrimaryShopId";
+
+const fakeOpaqueShopId = "cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg==";
+const MockComponent = () => <div>Mock</div>;
+const TestComponent = compose(withPrimaryShopId)(MockComponent);
```
aldeed: Is using `compose` necessary in this case when you could just do `withPrimaryShopId(MockComponent)`?

imports/plugins/core/graphql/lib/hocs/withPrimaryShopId.test.js
aldeed: @dancastellon See a couple of minor comments on the diff. Generally seems like a good pattern though

imports/test-utils/helpers/waitUntilLoaded.js
```diff
@@ -0,0 +1,22 @@
...
+ * @name waitUntilDataLoaded
+ * @summary Promise function that resolves when an enzyme-mounted React component's prop in question is truthy.
+ *  Useful for waiting until a GraphQL HOC has finished loading data (via Apollo's MockedProvider) in unit tests.
+ * @namespace TestHelpers
+ * @param {ReactWrapper} reactWrapper - result of calling enzyme's mount()
+ * @param {String} componentName - name of component to check
+ * @param {String} propName - name of prop on component to check for truthiness
+ * @returns {Promise} that resolves when prop is truthy
+ */
+export default function waitUntilDataLoaded(reactWrapper, componentName, propName) {
```
dancastellon: @aldeed I created this helper function for waiting until a GraphQL HOC has finished loading with the new `isLoading...` props I added. I still have to use a timeout, otherwise the function just keeps calling itself and `MockedProvider` never gets to update loading state to `false`.

imports/plugins/core/graphql/lib/hocs/withPrimaryShopId.test.js
dancastellon: Good catch, it isn't

imports/test-utils/helpers/waitUntilLoaded.js
aldeed: Thats cool. At least it won't fail if it's not done loading after the first tick.

dancastellon:
@zenweasel, @spencern, @aldeed - I believe this gql grid PR is ready to merge, after a final review. Next up: porting the storefront grid to component lib and the Meteor app. Thanks!

imports/plugins/core/versions/server/migrations/33_change_products_template_name.js
```diff
@@ -0,0 +1,40 @@
...
+
+    Packages.update({
+      "registry.template": "Products"
+    }, {
+      $set: {
+        "registry.$.template": "products"
+      }
+    }, { bypassCollection2: true });
+  }
+});
```
aldeed: These all need to be multi. Change `{ bypassCollection2: true }` to `{ bypassCollection2: true, multi: true }`

imports/plugins/included/product-variant/components/customer/productGridItem.js
```diff
@@ -14,7 +14,7 @@ class ProductGridItem extends Component {
       isLowQuantity: PropTypes.bool,
       isSoldOut: PropTypes.bool,
       media: PropTypes.arrayOf(PropTypes.object),
-      pricing: PropTypes.object,
+      pricing: PropTypes.array,
```
aldeed: Since you are relying on the shape of these, do `PropTypes.arrayOf(PropTypes.shape({ ... }))` with the correct props inside shape.

imports/plugins/core/graphql/lib/hocs/withCatalogItems.js
```diff
@@ -0,0 +1,51 @@
...
+          {({ loading, data, fetchMore }) => {
+            const props = {
+              ...this.props,
+              isLoadingCatalogItems: loading,
+              fetchMore
+            };
+
+            if (loading === false) {
+              const { catalogItems } = data;
+              props.catalogItems = catalogItems || {};
```
aldeed: I think it's best if we pass just the item nodes, i.e., do the map to `item.node` here before passing down. The wrapped component in almost all cases will not care about edges or cursors. If it ever needs them, we can add an option to provide them.

Also, fetchMore/loadMore should be passed down with `pageInfo` already bound to it. The wrapped component shouldn't need to know anything about `pageInfo`, cursors, or pagination. All it knows is that it calls `this.props.loadMore` when necessary, and you can pass down a `hasMoreItems` boolean so that it knows whether to display a load more button.

This is the pattern we've tried to use in the starter kit. It's a bit more complicated scenario, but you can see what I mean here: https://github.com/reactioncommerce/reaction-next-starterkit/blob/de613530d6aff10bdf5023d910b5ee07c06e127a/src/containers/cart/withCart.js#L205-L206

dancastellon:
Thanks @aldeed, just made the changes. I'm liking the pattern of `hasMoreCatalogItems` and `loadMoreCatalogItems`, `isLoadingCatalogItems`, etc - makes things really clean.

PR #4484, opened 2018-07-25
Do not publish empty variant to Catalog Customer

### LINKED ISSUE #4467: Product price range shows up undefined
## Issue Description
When you add a product and you publish it the product shows up as `$NaN.undefined` when you view the site as an anonymous user

## Steps to Reproduce
1. Create a new product
1. Make it visible and publish
1. Observe that the price is `$NaN.undefined`

## Possible Solution
It looks like making the top level product visible is no longer make the variant/options visible any longer.

### Versions

Release 1.14.0

hrath2015:
Resolves #4467 
Impact: **minor**  
Type: **bugfix**

## Issue
When a Product is made visible and published without any Variant,  Catalog customer gets price range as **$NaN.undefined - $-NaN.undefined**

## Solution
Included a condition in Products Publication to base  catalog selector to  not include products with empty variants.  Products without any Variant does not make any sense to end customer till it  is complete.  Let Admin work on the product in meantime and publish as many times as needed, customer will see only when product has at least 1 variant  in Catalog.

## Testing
From this branch start the app, login as admin, create a product by entering title, subtitle, vendor and description only( no variant detail). Make visible and publish.
Logout as admin and see product should not be displayed on homepage.

### DIFF
--- a/imports/plugins/core/core/server/publications/collections/products.js
+++ b/imports/plugins/core/core/server/publications/collections/products.js
@@ -367,7 +367,7 @@ function filterCatalogItems(catalogFilters) {
   // Init default selector - Everyone can see products that fit this selector
   const baseSelector = {
     "product.isDeleted": { $ne: true }, // by default, we don't publish deleted products
-    "product.isVisible": true // by default, only lookup visible products
+    "product.isVisible": true, // by default, only lookup visible products
     "product.variants": { $ne: [] } // if no variant, then do not publish
   };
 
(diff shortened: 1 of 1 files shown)


aldeed:
It would be better to not put these in the catalog in the first place. We don't support selling a product that has no variants afaik.

We could throw in `/imports/plugins/core/catalog/server/no-meteor/utils/publishProductToCatalog.js`, or just make sure there is a `min: 1` on the catalog schema, which is checked already. (Side note: It looks like we're catching and ignoring schema validation errors in that function, which we should not be doing.)

kieckhafer:
Closing in favor of #4518

PR #4490, opened 2018-07-26
WIP Ensure method hooks always run with correct timing (before and after method)

### LINKED ISSUE #4437: Method hooks do not always run after method is completed
## Prerequisites
* [x] Are you running the latest version? 1.14.0

## Issue Description
When using an "after" method hook for the core `catalog/publish/products` method, updates to the Catalog that happen in the method hook, are overwritten by the Catalog updates in the core method. It appears that my method hook is running before the completion of the method.

## Steps to Reproduce
1. Create a custom plugin that registers a method hook for catalog/publish/products, and attempts to add a field to catalog.product: `MethodHooks.after("catalog/publish/products", (options) => { Catalog.update({"product.productId": options.arguments[0][0])}, { $set: { "product.relatedTag": "test" } }) })`
2. Publish a product from the UI
3. Confirm update call in method hook is overwritten by `publishProductsMutation`'s `Catalog.update` call. At this stage, catalog.product.relatedTag will not exist, even though manual Catalog.update call added it

## Possible Solution
Changing line 10 of /imports/plugins/core/catalog/server/methods/publishProducts.js to read `const result = Promise.await(publishProductsMutation(context, productIds));` seems to fix the issue. May be happening on other methods that call GraphQL/pure async functions.

### Versions
Node: 8.11.3
NPM: 5.6.0
Meteor Node: 8.11.2
Meteor NPM: 5.10.0
Reaction CLI: 0.29.0
Reaction: 1.14.0
Reaction branch: release-1.14.0

dancastellon:
Resolves #4437  
Impact: **minor**  
Type: **bugfix**

## Issue
Due to current limitations with extending the Catalog schema, I created a `MethodHooks.after` callback to move a couple fields from Products to Catalog when a product is published. I registered the hook to the `catalog/publish/products` method. When it ran, I was seeing that my callback was firing before the completion of the method. 

## Solution
I identified that this was happening in Meteor methods that call GraphQL mutation functions, without `await`ing for them to complete before returning a result. There were 2 methods that did this - `catalog/publish/products` and `accounts/addressBookAdd`.

Meteor methods allow you to return a promise and will await it, so we should make method-hooks work in that scenario.

## Breaking changes
None

## Testing
1. Run Reaction in debug mode: `METEOR_DISABLE_OPTIMISTIC_CACHING=1 REACTION_LOG_LEVEL="DEBUG" reaction`
2. Create a new tag from the header, and confirm log reads "Created tag..." - this confirms a previously broken core "after" method hook now works
3. /imports/plugins/included/discount-codes/server/hooks/orders.js
As first line in MethodHooks.before callback, add logging: `console.log("before copyCartToOrder");`
4. /imports/plugins/included/notifications/server/hooks/notification.js
As first line in MethodHooks.after callback, add logging: `console.log("after copyCartToOrder");`
5. /imports/plugins/core/core/server/method-hooks.js
After line 105 (methodResult = ...) add logging: `console.log("method is done");`
6. Add a product to cart, complete the order, confirm server logs read:
before copyCartToOrder
cart/copyCartToOrder ...
Transitioned cart ... to order ...
sendOrderEmail status: new
method is done
after copyCartToOrder
- This confirms 2 core MethodHooks still work as expected
7. /imports/plugins/core/catalog/server/index.js
Add the following code to the end of the file:
import MethodHooks from "/imports/plugins/core/core/server/method-hooks";
MethodHooks.before("catalog/publish/products", (options) => {
  console.log("Before catalog/publish/products hook");
});
MethodHooks.after("catalog/publish/products", (options) => {
  console.log("After catalog/publish/products hook");
  return options.result;
});
8. Change a product's title (or any field), publish, confirm log reads:
Before catalog/publish/products hook
method is done
After catalog/publish/products hook
- This confirms the original reason for this PR - being able to register an after hook for catalog/publish/products so that the swag shop can push additional product fields to catalog after the method has completed.


dancastellon:
@zenweasel or @Akarshit can one of you review? I updated the testing instructions, should be much easier to confirm now. Thanks!

Akarshit:
@dancastellon 
When trying to add an address, as soon as I press the submit button, an empty address form is rendered instead of the grid displaying the address I just entered.
Also now the toast "Product published to Catalog" appears 5-6 secs later after clicking the publish button and there is no feedback to the user that something is happening. I don't know how big a problem this is but it might be worth discussing.

aldeed:
@dancastellon A better fix would be to add `Promise.await()` around `methodResult = MethodHooks._originalMethodHandlers[methodName].apply(this, args);` in `method-hooks.js`

Meteor methods allow you to return a promise and will await it, so we should make method-hooks work in that scenario.

It would not be the worst idea to also wrap the calling of all the hooks in `Promise.await` since those might also be async.

dancastellon:
Thanks @aldeed & @Akarshit. I've push the changes & updated the testing instructions. I'm now making changes to method-hooks.js instead of the 2 specific methods.

imports/plugins/core/core/server/method-hooks.js
```diff
@@ -74,21 +74,20 @@ MethodHooks._initializeHook = function (mapping, methodName, hookFunction) {
...
+    for (const beforeHook of beforeHooks) {
+      beforeResult = Promise.await(beforeHook.call(this, {
         result: undefined,
         error: undefined,
         arguments: args,
         hooksProcessed
-      });
+      }));
 
-      if (beforeResult === false) {
```
aldeed: Probably better to keep this early return here and remove the one after the loop, assuming that returning `false` is supposed to be a way to stop the remaining hooks from running. (Not sure if that's the case, or if any hooks actually use this.)

imports/plugins/core/core/server/method-hooks.js
```diff
@@ -99,27 +98,30 @@ MethodHooks._initializeHook = function (mapping, methodName, hookFunction) {
...
         result: methodResult,
         error: methodError,
         arguments: args,
         hooksProcessed
-      });
+      }));
       // If the after hook did not return a value and the methodResult is not undefined, warn and fix
-      if (_.isUndefined(hookResult) && !_.isUndefined(methodResult)) {
+      if (hookResult === undefined && methodResult !== undefined) {
         Meteor._debug("Expected the after hook to return a value.");
```
aldeed: We don't usually use `Meteor._debug`. Change this to `Logger.warn` while you're in here.

dancastellon:
Thanks @aldeed, made the changes you suggested. This one's ready. If we can get @Akarshit to confirm within the next day, can we merge this into 1.14? I ask because I have a PR to get the swag shop working with 1.14, but it requires this change.

Akarshit:
@dancastellon 
I am still seeing something weird when I try to add address, sometimes it gets added and sometimes it doesn't. Do you see that?
Also from the test instructions, if I add `return false` in the beforeHook of `MethodHooks.before("catalog/publish/products")` I am getting a error `Exception while invoking method 'catalog/publish/products' Error: Did not check() all arguments` is that expected?

dancastellon:
@Akarshit I'm not seeing that address issue. After yesterday's changes this PR is no longer touching that method. I do see the check error if I return false in the before MethodHook for copyCartToOrder, but that is likely because a few methods run after that one. I don't see a use case for people trying to prevent that method.. Try it with a simpler standalone method like shop/createTag.

Akarshit:
For me the address issue is happening in `release-1.14` branch too, so that is something not caused by this PR.

dancastellon:
@Akarshit & @aldeed I just confirmed that returning false in a before MethodHook always throws that `did not check() all arguments` error. That was happening even on 1.14 & 1.15, and there's nothing in the documentation mentioning the ability to return false in before hooks to prevent the main method from executing. I've decided to remove the code for that.

dancastellon:
The alternative would be removing or changing Meteor's `audit-arguments-checks` package, but removing it could potentially mean the introduction of security loopholes in the future (writing a method that doesn't check all args).

dancastellon:
@Akarshit & @aldeed Scratch that... I was able to make it work. Now when a before MethodHook returns false, it'll work as expected and not throw the `did not check() all args` error.

imports/plugins/core/core/server/method-hooks.js
```diff
@@ -74,52 +76,56 @@ MethodHooks._initializeHook = function (mapping, methodName, hookFunction) {
...
         error: undefined,
         arguments: args,
         hooksProcessed
-      });
+      }));
+
+      hooksProcessed += 1;
 
       if (beforeResult === false) {
+        check(args, [Match.Any]);
```
Akarshit: Let's move this to the top of the method. 
Since we are kind of skipping the check. It makes more sense to move this to the top.

dancastellon: If we do that, then methods that don't check all their args won't be able to take advantage of `audit-arguments-checks`. If we have it here, then we basically skip `audit-arguments-checks` only when it's not needed - i.e., exiting early, before the main method.

Akarshit: Yes makes sense.

dancastellon:
@spencern OK'd merging this in 1.14.1 once 1.14.0 is released.

zenweasel:
@dancastellon When testing this I do not see the log messages in the order you specified. Not sure how important that is but i see:
```
before copyCartToOrder
method is done
00:44:36.968Z DEBUG Reaction: cart/copyCartToOrder 2JxFMPkGWepEgocmu
00:44:37.099Z  INFO Reaction: Transitioned cart 2JxFMPkGWepEgocmu to order TmovHgRpeNgsErMy8
00:44:37.120Z DEBUG Reaction: sendOrderEmail status: new
after copyCartToOrder
```

dancastellon:
Closed this one and created a new PR for 1.14.1:
https://github.com/reactioncommerce/reaction/pull/4537

PR #4491, opened 2018-07-26
Fix migration error preventing app startup

### LINKED ISSUE #4362: Rolling back to previous Reaction version causes app to not start
## Issue Description
If you move to the current version and then decide you need to roll back for any reason, and that newer version contained a migration, the app will not start with the following error:

```
web_1  | 01:42:37.917Z  INFO Reaction: JobServer started
web_1  | 01:42:38.668Z  WARN Reaction: Skipped loading settings from reaction.json.
web_1  | 01:42:39.692Z  INFO Reaction: Reaction Version: 1.12.1
web_1  |
web_1  | /app/programs/server/node_modules/fibers/future.js:280
web_1  | 						throw(ex);
web_1  | 						^
web_1  | Error: [Can't find migration version 26]
web_1  |     at Object.Migrations._findIndexByVersion (packages/percolate_migrations.js:293:9)
web_1  |     at Object.Migrations._migrateTo (packages/percolate_migrations.js:209:23)
web_1  |     at Object.Migrations.migrateTo (packages/percolate_migrations.js:167:10)
web_1  |     at Hooks.Events.add (imports/plugins/core/versions/server/startup.js:18:14)
web_1  |     at /app/programs/server/npm/node_modules/@reactioncommerce/hooks/dist/main.js:71:14
web_1  |     at Array.reduce (<anonymous>)
web_1  |     at Object.Hooks.Events.run (/app/programs/server/npm/node_modules/@reactioncommerce/hooks/dist/main.js:70:22)
web_1  |     at Object.init (server/api/core/core.js:62:18)
web_1  |     at startup (server/startup/index.js:46:12)
web_1  |     at Function.time (/app/programs/server/profile.js:309:28)
web_1  |     at /app/programs/server/boot.js:427:13
web_1  |     at /app/programs/server/boot.js:472:5
web_1  |     at Function.run (/app/programs/server/profile.js:510:12)
web_1  |     at /app/programs/server/boot.js:470:11
```

## Steps to Reproduce
Using docker or a local installation roll forward to version 1.13.0, and then roll back to version 1.12.1.

## Possible Solution
The code that looks for migrations needs to be in a `try/catch` and throw an error but not stop the app from starting. Since the migration package we use is a Meteor package not sure if we want to either pull the package in or PR the change.

impactmass:
Resolves #4362   
Impact: **major**  
Type: **fix**

## Issue
If you move to the current version and then decide you need to roll back for any reason, and that newer version contained a migration, the app will not start, and you get an error similar to:
```
web_1  | /app/programs/server/node_modules/fibers/future.js:280
web_1  | 						throw(ex);
web_1  | 						^
web_1  | Error: [Can't find migration version 26]
```
## Solution
- Added a check to to compare the existing migration version with available code migration before running migrations. So, we only call `migrateTo("latest")` when it's actually possible to migrate higher. (as [suggested on issue ](https://github.com/reactioncommerce/reaction/issues/4362#issuecomment-403872554))
- Also added a log.warn with text: `You are running a Reaction version with migration version (${highestAvailableVersion}) below your current DB migration state (${currentMigrationVersion})`

## Breaking changes
None that I'm aware of.

## Testing
- Create a dummy migration after pulling this branch
- Start the app and ensure the app is migrated up to the dummy migration created
- Then, remove that dummy migration 

Alternative way to test:
- Start the app on this branch.
- Connect to your DB, and increment the version number of the only Migration collection document. 
- After incrementing, restart the app
- Confirm that there are no migration errors, and there's a Log warn message about migration version

### DIFF
--- a/imports/plugins/core/versions/server/startup.js
+++ b/imports/plugins/core/versions/server/startup.js
@@ -21,7 +21,7 @@ Hooks.Events.add("afterCoreInit", () => {
   const highestAvailableVersion = Migrations._list[Migrations._list.length - 1].version;
 
   if (currentMigrationVersion > highestAvailableVersion) {
-    Logger.warn(`You are running a Reaction version with migration version (${highestAvailableVersion}) below your currrent DB migration state (${currentMigrationVersion})`);
+    Logger.warn(`You are running a Reaction install with migration version (${highestAvailableVersion}) below your current DB migration state (${currentMigrationVersion})`);
     Migrations._setControl({ locked: false, version: highestAvailableVersion });
   } else if (!Meteor.isTest) {
     Migrations.migrateTo("latest");
--- a/imports/plugins/core/versions/server/startup.js
+++ b/imports/plugins/core/versions/server/startup.js
@@ -23,7 +23,7 @@ Hooks.Events.add("afterCoreInit", () => {
   if (currentMigrationVersion > highestAvailableVersion) {
     Logger.warn(`You are running a Reaction version with migration version (${highestAvailableVersion}) below your currrent DB migration state (${currentMigrationVersion})`);
     Migrations._setControl({ locked: false, version: highestAvailableVersion });
-  } else {
+  } else if (!Meteor.isTest) {
     Migrations.migrateTo("latest");
   }
 });
--- a/imports/plugins/core/versions/server/startup.js
+++ b/imports/plugins/core/versions/server/startup.js
@@ -1,3 +1,4 @@
+import { Meteor } from "meteor/meteor";
 import Hooks from "@reactioncommerce/hooks";
 import Logger from "@reactioncommerce/logger";
 import { Migrations } from "meteor/percolate:migrations";
@@ -16,5 +17,13 @@ Migrations.config({
 });
 
 Hooks.Events.add("afterCoreInit", () => {
-  Migrations.migrateTo("latest");
+  const currentMigrationVersion = Migrations._getControl().version;
+  const highestAvailableVersion = Migrations._list[Migrations._list.length - 1].version;
+
+  if (currentMigrationVersion > highestAvailableVersion) {
+    Logger.warn(`You are running a Reaction version with migration version (${highestAvailableVersion}) below your currrent DB migration state (${currentMigrationVersion})`);
+    Migrations._setControl({ locked: false, version: highestAvailableVersion });
+  } else {
+    Migrations.migrateTo("latest");
+  }
 });
--- a/imports/plugins/core/versions/server/startup.js
+++ b/imports/plugins/core/versions/server/startup.js
@@ -20,9 +20,13 @@ Hooks.Events.add("afterCoreInit", () => {
   const currentMigrationVersion = Migrations._getControl().version;
   const highestAvailableVersion = Migrations._list[Migrations._list.length - 1].version;
 
+  // Checks to ensure the app is running against a DB at the right migration state. Running the app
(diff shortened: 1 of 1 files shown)


zenweasel:
@impactmass Could I test this by just adding a dummy migration, starting the app, and then removing it?

I think @aldeed had also mentioned that stopping the app from running might be the expected behavior rather than letting the user run the app with the database in an unknown state. I see good arguments for each, but I would def bump the Log level to `error` or `fatal` because there is not-small chance your app will completely not work.

impactmass:
@zenweasel 

1. yes, that is actually an easier test step. 

2. Thinking (again) about the argument for & against this decision, I think I now lean towards the behaviour of preventing the app from starting (if the DB is in a different state).

Since Migrations are still built-in to the app right now, it makes sense that app startup is prevented. When Migrations becomes a separate entity, then the decision can change.

zenweasel:
Maybe it should not start but also not lock the db?

impactmass:
Well... the locking is as a result of the failed attempt to run the Migration (which is also what prevents the app from starting). Except if we write code to check the versions before calling `MigrateToLatest`. And if we find that the DB version is not compatible, we then make the app exit (and add an error description about why).

That way it doesn't lock too.

What do you think?

aldeed:
I agree with the proposal here.

- Don't do the `Migrations._setControl` line
- Print the message, but add to it some helpful suggestion like "Start a higher version of the app and run Migrations.migrateTo(x) in the Meteor shell before trying to run this version" (replace "x" with the value of `highestAvailableVersion` variable)
- Then `process.exit(0)`. I could see a case for a non-zero exit code, but it's not really a crash. Whatever seems best.

impactmass:
@zenweasel updates pushed to this branch based on conversations here and the issue (#4362).

Now, when the DB version is higher:
- It logs a message about the incompatibility
- It doesn't call Migrations._setControl
- It logs a message about how to fix the issue
- It exits without trying to migrate

It doesn't enable a flag to override and force-start the app, because that will still lead to the app malfunctioning since it'll be running against a wrong DB state. Is there more reason for 'optioning' a force-start?

aldeed:
I don't see a need to force it to start. You could manually adjust the migration record in the db if you really need to force it to start.

aldeed:
@zenweasel I was assuming this would only be happening in dev, but I guess you're correct that it could happen for a production rollback. The correct way to fix it is to connect with the app version first, run the "down" migrations back down to the version expected by the Reaction version that you are rolling back to, then connect with the older app version.

This process would be easier if migrations ran outside of the app, which is why we're working toward moving them out.

zenweasel:
@aldeed Yeah, there's a bunch of cases we should be handling here but we can't until we move migrations out of Meteor

aldeed:
I still think this PR is an improvement over the confusing crash we have now, so I'd like to merge it. Will defer to @zenweasel on that decision, though.

zenweasel:
I agree with @aldeed, it's an improvement that's worth doing. Sorry I didn't get a chance to test/approve today.

zenweasel:
So when starting with a missing migration I get the message (which is way better) but it doesn't exit, it gets stuck at `Your application is crashing. Waiting for file change`. Is that the expected behavior?

aldeed:
@zenweasel I did some searching but I don't see any way to circumvent the retries that Meteor does when in development mode. In prod, it should just exit.

zenweasel:
@aldeed Great. That's what I was hoping

PR #4508, opened 2018-08-06
GraphQL: Fix CartItem.currentQuantity

aldeed:
Impact: **minor**  
Type: **bugfix**

## Issue
In GraphQL response, `CartItem.currentQuantity` is always null.

## Solution
It now uses the correct lookup and `CartItem.currentQuantity` is the correct number.

## Breaking changes
None

## Testing
Do a GraphQL cart query or createCart mutation asking for `currentQuantity` of the `items`, and verify that it is present and correct.


mikemurray:
Waiting for tests to pass before merging, but everything looks good so far.

PR #4521, opened 2018-08-08
Added production bundle size check to CircleCI

Akarshit:
Impact: **minor**  
Type: **feature**

## Issue
Currently there is no way to limit the size of the production bundle that is pushed to the client. It is very easy for a developer to push code that increases the bundle by a significant amount.

## Solution
This PR adds a check to CircleCI, to check the bundle size. The `build` will fail if the bundle size is increased.

## Testing
1. Create a PR, click on the Circle CI `build` logs. You can see the logs of `Check .js file size` and `Check .css file size`.
2. Add a client library to reaction and push to the PR. Notice that it will fail and in the logs of `build` the developer can see the actual file size and expected file size.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,6 +41,31 @@ jobs:
       - run:
           name: Meteor NPM Install
           command: meteor npm install
+      - run:
+          name: Build dist bundle
+          command: meteor build --directory ../build/
+      - run:
+          name: Check .js size
+          command: |
+            maxSize=1468006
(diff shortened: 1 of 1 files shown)


zenweasel:
I think it would be fine if the limit was hardcoded but any PR that increases the limit could do that with a code change. That way it's at least explicit and obvious to a code reviewer.

aldeed:
@spencern @zenweasel So are we good to merge this then?

spencern:
@aldeed I'm okay with that.

zenweasel:
@aldeed @spencern Since you can just increase `maxsize` manually by editing the CircleCI config I think this is fine and protects you from accidentally increasing bundle size

PR #4547, opened 2018-08-14
Import `getSlug` instead of using `this.getSlug`

pmn4:
Impact: **minor**  
Type: **bugfix**

## Issue
`this.getSlug` is used in the Reaction core (server-side) file, however, `Reaction.getSlug` is defined outside of the file (in index.js). This is ticky-tack because the code is working everywhere it used, but I feel this is a more correct implementation.
This came up when writing a unit test for `getPrimaryShopPrefix`.

## Solution
import `getSlug` instead of relying on `this.getSlug` which could potentially be undefined.

## Testing
You will not notice a difference in practice because `this.getSlug` is defined by the time it is used, but running the test from this PR should highlight the issue.


imports/plugins/core/core/server/Reaction/core.app-test.js
```diff
@@ -155,6 +155,16 @@ describe("Server/API/Core", () => {
     });
   });
 
+  describe.only("#getPrimaryShopPrefix", () => {
```
aldeed: @pmn4 Remove `.only`

aldeed:
Looks good other than the .only

pmn4:
ah! I purged my local repo recently and my pre-push hooks went with it. I'll take care of it tonight

aldeed:
@pmn4 Reminder

PR #4556, opened 2018-08-16
Fulfillment pseudo-service

### LINKED ISSUE #4498: Create a Fulfillment pseudo-service
## Goal
To establish some of the service/plugin patterns for server code.

## Work
1. Move the `shipping/server` core plugin and flat rate shipping plugin into an `/imports/services/fulfillment` folder with no Meteor dependencies.
1. Move fulfillment GraphQL schemas into that service. Update GraphQL plugin to stitch it in.
1. Move related cart mutations such as `setShippingAddressOnCart` into the service.
1. Export methods, pubs, and schemas, and have the core `init` code wire them up.
1. Update any direct method calling to operate based on hooks instead (simulate listening to event stream).

aldeed:
Resolves #4498  
Impact: **breaking**  
Type: **feature**

## Changes
This introduces a new "services" pattern, by fully moving the "shipping" plugin server code out of Meteor and over to the new Node app.
- The `.reaction/devserver` app is moved to `/imports/node-app/index.js` and shall henceforth be referred to as Reaction Node (clever, eh?). It's also rearranged to more of a class instantiation model (see `ReactionNodeApp.js`). Note that this is a separate, fully autonomous Node app even though it lives within the Meteor app for now. The Meteor app in some cases still imports files from the Node app, but not vice versa. Run Reaction Node with `npm run reaction-node` or in Docker with `docker-compose up -d devserver`
- The core "shipping" plugin server code becomes a "service" in the `/imports/node-app/services` folder (what I'm calling a pseudo-service because it is service-like but does not have its own GraphQL app and is not networked over http like we will have in the future). It is renamed to "fulfillment". (See `ReactionService.js`)
- The "shipping-rates" plugin server code becomes a plugin of the "fulfillment" service, which has its own separate "plugins" folder in `/imports/node-app/services/fulfillment/plugins` now.
- Client code for both of these plugins is still where it has always been.
- Many of the non-Meteor files from `/imports/plugins/core/graphql` are moved to `/imports/node-app`
- Some of the file-collections instantiation for Reaction Node is moved to a `files` service in the `/imports/node-app/services` folder, which consists of only a startup function. (In the Meteor app, the file-collections startup code remains the same.)
- `shipping/rates/add` Meteor method is removed and replaced with `createFlatRateFulfillmentMethod` GraphQL mutation
- `shipping/rates/update` Meteor method is removed and replaced with `updateFlatRateFulfillmentMethod` GraphQL mutation
- `shipping/rates/delete` Meteor method is removed and replaced with `deleteFlatRateFulfillmentMethod` GraphQL mutation

## Breaking changes
This is breaking only for non-core plugins that rely on the existing folder structure of shipping/fulfillment plugins, or on the removed Meteor methods.

The following Meteor methods are changed:
- `shipping/updateShipmentQuotes` is removed
- `shipping/updateParcelSize` is renamed to `shop/updateDefaultParcelSize` and moved to core/shop service
- `shipping/provider/toggle` is removed
- `shipping/rates/add` is removed
- `shipping/rates/update` is removed
- `shipping/rates/delete` is removed

## Testing
- Test normal checkout flow, logged in and not
- Test creating, editing, and deleting flat rate shipping methods as an admin
- Test updating the default parcel size as an admin


imports/node-app/core/ReactionNodeApp.js
```diff
@@ -0,0 +1,100 @@
...
+      getRegisteredFunctionsForType,
+      registerFunction
+    };
+
+    this.mongodb = mongodb;
+
+    this.expressApp = createApolloServer({
+      addCallMeteorMethod(context) {
+        context.callMeteorMethod = (name) => {
+          console.warn(`The "${name}" Meteor method was called. The method has not yet been converted to a mutation that` + // eslint-disable-line no-console
```
spencern: This is a good idea.

imports/node-app/services/fulfillment/plugins/flat-rate/mutations/createFlatRateFulfillmentMethod.js
```diff
@@ -0,0 +1,59 @@
...
+      provider: {
+        name: "flatRates",
+        label: "Flat Rate"
+      }
+    });
+  }
+
+  method._id = Random.id();
+
+  // MongoDB schema still uses `enabled` rather than `isEnabled`
```
spencern: Good doc comment. I wouldn't have figured this out on my own.

imports/plugins/core/graphql/server/no-meteor/resolvers/cart/Mutation/index.js
```diff
@@ -5,7 +5,6 @@ import removeCartItems from "./removeCartItems";
 import setEmailOnAnonymousCart from "./setEmailOnAnonymousCart";
 import setShippingAddressOnCart from "./setShippingAddressOnCart";
 import updateCartItemsQuantity from "./updateCartItemsQuantity";
-import updateFulfillmentOptionsForGroup from "./updateFulfillmentOptionsForGroup";
```
spencern: Did this get moved or deleted?

imports/plugins/core/shipping/client/templates/settings/shipping.js
```diff
@@ -48,8 +48,6 @@ Template.shippingSettings.events({
     if (packageId) {
       // update package registry
       Meteor.call("registry/update", packageId, settingsKey, fields);
-      // also update shipping provider status
-      Meteor.call("shipping/provider/toggle", packageId, settingsKey);
```
spencern: Are we performing this functionality in some other way now?

package.json
```diff
@@ -192,7 +192,8 @@
     }
   },
   "scripts": {
-    "devserver": "NODE_ENV=devserver nodemon .reaction/devserver/index.js",
+    "devserver": "npm run reaction-node",
+    "reaction-node": "NODE_ENV=reaction-node nodemon ./imports/node-app/index.js",
```
spencern: Did we decide we want to keep this as `devserver` here?

imports/plugins/core/graphql/server/no-meteor/resolvers/cart/Mutation/index.js
aldeed: It was moved to being directly imported. In general I want to move away from calling all these mutations from context and rather import them directly where they are called, but some will still be put on context so that other plugins can call them. I'm changing this yet again and straightening out all the inconsistencies as part of getting back to the registerPlugin structure as we discussed.

imports/plugins/core/shipping/client/templates/settings/shipping.js
aldeed: I didn't see where this was being used. I believe it was only going by the package being enabled, so the first Meteor.call is the only one actually having an effect.

package.json
aldeed: Yes, I'll adjust in the other branch.

imports/node-app/index.js
```diff
@@ -0,0 +1,35 @@
...
+import filesService from "./services/files";
+import fulfillmentService from "./services/fulfillment";
+
+const { MONGO_URL, ROOT_URL } = process.env;
+if (!MONGO_URL) throw new Error("You must set MONGO_URL");
+if (!ROOT_URL) throw new Error("You must set ROOT_URL");
+
+const PORT = 3030;
+
+const app = new ReactionNodeApp({
```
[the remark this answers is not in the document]
aldeed: I will definitely do that in the orders branch

spencern:
Closed in favor of #4622

aldeed:
Some changes from here were moved to #4622

PR #4560, opened 2018-08-17
update the display message when adding  member to group

### LINKED ISSUE #4530: Untranslated text shows up in Account manager
## Issue Description
When making changes to an account, when successful the message shows up as
`admin.groups.addUserSuccess` rather than the user text string

## Steps to Reproduce
1. Login as admin
1. Add an account
1. Observe the toast displays `admin.groups.addUserSuccess`

### Versions
reaction 1.14.0

nadaa:
Resolves #4530
Impact: **minor**  
Type: **bugfix**

## Issue
When an admin adds a user to any group, a fixed message appears "admin.groups.addUserSuccess". Here request I updated the message to be [ username "is added successfully"].

## Solution
Changed the toast display message in the file: C:\Users\nada\Desktop\Reaction_OOS\reaction\imports\plugins\core\accounts\client\containers/accountsDashboardContainer.js

### DIFF
--- a/imports/plugins/core/accounts/client/containers/accountsDashboardContainer.js
+++ b/imports/plugins/core/accounts/client/containers/accountsDashboardContainer.js
@@ -33,7 +33,7 @@ const handlers = {
           Alerts.toast(i18next.t("admin.groups.addUserError", { err: err.message }), "error");
         }
         if (!err) {
-          Alerts.toast(i18next.t("admin.groups.addUserSuccess"), "success");
+          Alerts.toast(i18next.t(account.name+" is added Successfully"), "success");
         }
         if (onMethodDone) { onMethodDone(); }
       });


imports/plugins/core/accounts/client/containers/accountsDashboardContainer.js
```diff
@@ -33,7 +33,7 @@ const handlers = {
           Alerts.toast(i18next.t("admin.groups.addUserError", { err: err.message }), "error");
         }
         if (!err) {
-          Alerts.toast(i18next.t("admin.groups.addUserSuccess"), "success");
```
zenweasel: So we don't directly display text here. Those message are just keys that are used to find translations. So the thing to do would be to try and find if this translation is missing from a [here](https://github.com/reactioncommerce/reaction-demo/blob/0e1421e6758feebc7366ca7a0034c74de9b4f909/private/data/i18n/en.json#L29) and add that key and a default English version

zenweasel:
@nadaa You should always include testing instructions when submitting a PR. Also see the code comments as this is not the correct approach to fixing this problem.

imports/plugins/core/accounts/client/containers/accountsDashboardContainer.js
nadaa: Thanks @zenweasel , I will fix this.

nadaa: @zenweasel , I added the key and it works.  The code passed the liniting, but not the test, there are many failures !. What do you suggest me to do?

zenweasel: @nadaa You still have the wrong approach. We can't hard-code an English-only message like that. You need to put back the key `admin.groups.addUserSuccess` and make it work using that. Otherwise you will break it for every language except English.

You also need to run the linter (`npm run lint`) and fix any linting errors (you can ignore warnings). That's what is causing the tests to fail.

zenweasel:
@aldeed #4562 adds the translations but this PR tries to fix the actual issue with the key, which is the root of the original problem.

@nadaa Still looking for you to revisit this PR. Thanks!

nadaa:
Hi @zenweasel , shall I close this pull request?
Thanks

zenweasel:
Last time I looked he had just hard-coded the English message in, but will look again

zenweasel:
@aldeed Ah, you're right. I missed that that was what was going on.

nadaa:
I fixed the issue last time @zenweasel. As for translations, it should be added automatically by i18n, right?, another issue, when I run the test locally, it fails, why?

PR #4565, opened 2018-08-21
feat: return absolute media URLs from GraphQL

### LINKED ISSUE #4543: GraphQL: Update all media URLs to be absolute
## Issue
Currently all media URLs are returned by GraphQL resolvers as relative. We prefix the image server on the client. This won't work with URLs for an S3 adapter (for example) where we want the client to retrieve directly from a CDN. It's also more work for the client.

## Solution
- Update the URL resolvers to add the prefix to make all of the URLs absolute.
- Update client code in the Meteor app, and the starter kit and/or component library components to expect absolute URLs.

dancastellon:
Resolves #4543 
Impact: **minor**  
Type: **feature**

## Issue
Currently GraphQL returns relative media URLs. Because of this, storefront needs to prepend the returned paths with the Meteor server's base URL.

## Solution
Implement the currently unused `xformProductMedia` function in resolvers to prepend media URLs with the server's base URL. This will allow us in the future to take account for CDNs.

## Breaking changes
None

## Testing
1. Catalog Image URLs
    * Create a product and upload some images. Click on a variant option and add images to that option. Publish product.
    * Query GraphQL
```
{
  catalogItems(shopIds: ["cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg=="]) {
    edges {
      node {
        createdAt
        updatedAt
        ... on CatalogItemProduct {
          product {
            media {
              URLs {
                large
                medium
                original
                small
                thumbnail
              }
            }
            primaryImage {
              URLs {
                large
                medium
                original
                small
                thumbnail
              }
            }
            variants {
              media {
                URLs {
                  large
                  medium
                  original
                  small
                  thumbnail
                }
              }
              primaryImage {
                URLs {
                  large
                  medium
                  original
                  small
                  thumbnail
                }
              }
              options {
                media {
                  URLs {
                    large
                    medium
                    original
                    small
                    thumbnail
                  }
                }
                primaryImage {
                  URLs {
                    large
                    medium
                    original
                    small
                    thumbnail
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

* Confirm image URLs are absolute (and correct)
2. Cart Image URLs
    * Add an item to your cart that has an image
    * Query GraphQL
```
accountCartByAccountId(
    accountId: “...",
    shopId: “..."
  ) {
    items {
      edges {
        node {
          imageURLs {
            large
            medium
            original
            small
            thumbnail
          }
        }
      }
    }
  }
```
* Confirm image URLs are absolute (and correct)
3. Imports/plugins/core/cart/server/methods/addToCart.js
* After context on line 38, add logging:
```
  console.log("context user:", context.user);
  console.log("context rootUrl:", context.rootUrl);
  console.log("context getAbsoluteUrl:", context.getAbsoluteUrl("test"));
```
* Add something to your cart and confirm server log shows correct context user, root url, and absolute url

### DIFF
--- a/imports/plugins/core/core/server/util/getAbsoluteUrl.js
+++ b/imports/plugins/core/core/server/util/getAbsoluteUrl.js
@@ -1,5 +1,3 @@
-import ReactionError from "@reactioncommerce/reaction-error";
-
 /**
  * @name getAbsoluteUrl
  * @summary Returns the absolute/base URL, returning process.env.ROOT_URL if set,
--- a/imports/plugins/core/core/server/util/getAbsoluteUrl.js
+++ b/imports/plugins/core/core/server/util/getAbsoluteUrl.js
@@ -1,18 +1,14 @@
 /**
  * @name getAbsoluteUrl
- * @summary Returns the absolute/base URL, returning process.env.ROOT_URL if set,
- *  otherwise using the request's protocol & hostname
- * @param {Object} request Express request object
- * @param {Object} request.hostname Hostname derived from Host or X-Forwarded-Host heaer
- * @param {String} request.protocol Either http or https
- * @returns {String} URL
+ * @summary Combines and returns the given root URL and path
+ * @param {String} rootUrl URL ending with /
+ * @param {String} path Path that may or may not start with /
+ * @returns {String}
  */
-export default function getAbsoluteUrl(request) {
-  const { ROOT_URL } = process.env;
-  if (ROOT_URL) {
-    return `${ROOT_URL}/`;
+export default function getAbsoluteUrl(rootUrl, path = "") {
+  if (path.startsWith("/")) {
+    path = path.slice(1);
   }
 
-  const { hostname, protocol } = request;
-  return `${protocol}://${hostname}/`;
+  return `${rootUrl}${path}`;
 }
--- a/imports/plugins/core/core/server/util/getAbsoluteUrl.js
+++ b/imports/plugins/core/core/server/util/getAbsoluteUrl.js
@@ -3,12 +3,13 @@
  * @summary Combines and returns the given root URL and path
  * @param {String} rootUrl URL ending with /
  * @param {String} path Path that may or may not start with /
- * @returns {String}
+ * @returns {String} Full URL
  */
 export default function getAbsoluteUrl(rootUrl, path = "") {
+  let pathNoSlash = path;
   if (path.startsWith("/")) {
-    path = path.slice(1);
+    pathNoSlash = path.slice(1);
   }
 
-  return `${rootUrl}${path}`;
+  return `${rootUrl}${pathNoSlash}`;
 }
--- a/imports/plugins/core/core/server/util/getAbsoluteUrl.js
+++ b/imports/plugins/core/core/server/util/getAbsoluteUrl.js
@@ -1,0 +1,26 @@
+import url from "url";
+import ReactionError from "@reactioncommerce/reaction-error";
+
+/**
+ * @name getAbsoluteUrl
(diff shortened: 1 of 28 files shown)


imports/plugins/core/graphql/server/no-meteor/resolvers/xforms/catalogProduct.js
```diff
@@ -17,19 +18,20 @@ export const encodeCatalogProductOpaqueId = encodeOpaqueId(namespaces.CatalogPro
 export function xformProductMedia(mediaItem) {
   if (!mediaItem) return null;
 
-  const { metadata, large, medium, image, small, thumbnail } = mediaItem;
-
+  const { priority, toGrid, productId, variantId, URLs: { large, medium, original, small, thumbnail } } = mediaItem;
+  const absoluteUrl = Reaction.absoluteUrl().slice(0, -1);
```
dancastellon: @aldeed Is this all that's needed? How can I take into account when a CDN is being used?

imports/plugins/core/graphql/server/no-meteor/resolvers/catalog/CatalogProduct/index.js
```diff
@@ -11,5 +12,12 @@ export default {
   shop: resolveShopFromShopId,
   pricing,
   tagIds,
-  tags
+  tags,
+  media: (node, args, context) => node.media.map((mediaItem) => xformProductMedia(mediaItem, context)),
+  primaryImage: (node, args, context) => xformProductMedia(node.primaryImage, context),
+  variants: (node, args, context) => node.variants.map((variant) => {
+    variant.media = variant.media.map((mediaItem) => xformProductMedia(mediaItem, context));
+    variant.primaryImage = xformProductMedia(variant.primaryImage, context);
```
aldeed: Variants can have options, which have the same schema as variants. `if (variant.options)` then do the same map on the options.

imports/plugins/core/core/server/util/getAbsoluteUrl.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * @name getAbsoluteUrl
+ * @summary Returns the absolute/base URL, returning process.env.ROOT_URL if set,
+ *  otherwise using the request's protocol & hostname
+ * @param {Object} request Express request object
+ * @param {Object} request.hostname Hostname derived from Host or X-Forwarded-Host heaer
+ * @param {String} request.protocol Either http or https
+ * @returns {String} URL
+ */
+export default function getAbsoluteUrl(request) {
```
aldeed: Similar to how the current `Reaction.absoluteUrl` function works, this should take an optional `path` arg and then do the work of properly appending that path. It should check if it starts with `/` and remove it so that we don't end up with two `/`.

In your xform you will then do something like `large: context.getAbsoluteUrl(large)`.

Actually, performance wise, it's best if you have two different functions. This function you have now would be called only one time in `buildContext`. You could actually set the result of that on context, too, as `rootUrl`. Then `context.getAbsoluteUrl` would only do the work to combine root url with path. Something like:

```js
function buildContext() {
  context.rootUrl = getRootUrl(request); // this is the function you currently have, renamed
  context.getAbsoluteUrl = (path) => getAbsoluteUrl(context.rootUrl, path); // this just combines the strings intelligently
}
```

dancastellon:
@aldeed Thanks, changes made!

aldeed:
Not merging yet because we should merge this around the same time as the starter kit PR: https://github.com/reactioncommerce/reaction-next-starterkit/pull/243

PR #4566, opened 2018-08-22
[chore] Remove unused schemas

zenweasel:
Resolves NA
Impact: **minor**  
Type: **chore**

## Issue
This should have been removed when I removed the SMS and Logs package but I missed it.

## Solution
Removed

## Breaking changes
The SMS schema will be moved to the SMS package in contrib. Logs will be moved to the Logs package.

## Testing
1. Place an order and ensure you get no errors


zenweasel:
@aldeed I think this is ready to go. Everything's been moved into stuff in contrib

NuruddinBadawi:
I check the doc here https://docs.reactioncommerce.com/docs/sms-notifications, but the feature not exist, so why the sms notifications is not available ? 
I think it's worth to have it.

PR #4584, opened 2018-08-27
feat: use component library's CatalogGrid

### LINKED ISSUE #4583: Use Component Library's CatalogGrid for product grid
## Work
Now that there's a new `<CatalogGrid />` component in our component lib, we should use it in the Meteor app.

dancastellon:
Resolves #4583 and #4639 
Impact: **minor**  
Type: **feature**

## Solution
This PR replaces the current customer grid UI with the new `CatalogGrid` component.

## Breaking changes
Only breaking for custom sites as the customer-facing `productGridItem` component has been removed.

## Testing
1. Use reaction-data-loader on its 1.14 branch to load some products: `npx babel-node data/generator.js -m mongodb://localhost:3001/ -l dev -n meteor --max_old_space_size=8192 --presets es2015,stage-2`
2. Confirm grid with new styles works as expected - progressively loads images, responsive from mobile up, loading more as you scroll to the bottom
3. Currency - Change the currency via the dropdown in the header. Confirm prices are converted and displayed correctly, including prices where only 1 currency symbol should be displayed with a price range (RUB)
4. Search - Confirm improved search modal that uses the new grid, displays images (even when logged out), is perfectly responsive at any size, shows correct currency, and brings you to the PDP on click. NOTE: If using the data loader, the correct primary image may not be displayed due to a bug: https://github.com/reactioncommerce/reaction-data-loader/issues/8

### Badge label testing:
Note: Might be easier to create a new product and copy the PDP URL. If you create a product and try to find it in the admin UI (after loading the dev dataset), it'll take a while since the customer grid is sorted by newest first, but the admin grid is not.

1. Create a product, add title to variant, set quantity to 1, warn at 1. Publish & logout
2. Confirm “Limited supply” badge
3. Switch to other languages and confirm badge text updates reactively
4. Login and edit product, set variant quantity to 0, turn allow backorder off. Publish & logout
5. Confirm “Sold out” badge. Confirm translations
6. Login and turn allow backorder on for product. Publish & logout
7. Confirm “Backorder” label & translations

### Price range & search record update testing
Two core bugs were discovered while work on this PR: `product.price` was not always being updated, and search records were being updated before product changes were published. To confirm the fixes:
1. Create a product, add a title to the variant, set quantity, don't publish
2. Confirm the product doesn't appear in search yet
3. Publish the product
4. Confirm the product now does appear in search + the price range is correct, both in the main customer grid and search results
5. Edit the price for the variant and/or add a new visible variant with a price
6. Confirm in DB product.price range, min, and max are correct, but that the changes haven't propagated to search results yet
7. Publish the changes and confirm the search result's price range has updated
8. Confirm 6 & 7 with variants that have options.


ajporlante:
Im probably missing something here, but I don't see where `Link` is being used.

dancastellon:
@ajporlante See `appComponents.js`. That file sets up the components context. Components in the library can then just define a components prop type with the components they need - instead of importing them directly. This lets us easily replace components that are used in others, if that makes sense. In this case, we override the default `Link` component with a custom one that routes via React Router.

See https://stoic-hodgkin-c0179e.netlify.com/#!/CatalogGrid

zenweasel:
@dancastellon This feature before this change used to do an "infinite scroll". Now it seems it has regressed back to "load more products". Was that intentional?

zenweasel:
Additionally clicking on "load more products" reloads the entire grid again rather than just adding more products to the bottom of the grid.

zenweasel:
If I click "Load More Products" more than a couple of times, rendering the grid takes a **really** long time. Up to 10-15 seconds.

zenweasel:
Looking at this in incognito mode, this looks like the styling is all off

zenweasel:
If I change currencies the prices do not change. They do change if I am logged in as admin

zenweasel:
Looks like the Search Grid is broken (no images, misaligned)

imports/plugins/core/ui/client/components/link/link.js
```diff
@@ -0,0 +1,23 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+
+export default class Link extends Component {
```
zenweasel: If this is something we should be using elsewhere do we need to add some docs on it, or at least a JSDoc?

package.json
```diff
@@ -102,6 +102,7 @@
     "react-autosuggest": "^9.3.3",
     "react-avatar": "^2.5.1",
     "react-color": "^2.13.8",
+    "react-container-query": "^0.11.0",
```
zenweasel: I don't see this imported anywhere?

dancastellon: @zenweasel This is used in the CatalogGrid component. According to the "Verify Peer Dependencies" section [here](https://stoic-hodgkin-c0179e.netlify.com/), you need to install the the component lib's peer dependencies in your app. If I remove line 105 I get an error on startup:

```
Unable to resolve some modules:

  "react-container-query" in                  
/Users/dan/reaction/reaction/node_modules/@reactioncommerce/components/CatalogGrid/v1/CatalogGrid.js
(web.browser.legacy)
```

dancastellon:
@zenweasel In regards to the prices not including a currency symbol, it appears to be a data-loader issue. I've filed an [issue here](https://github.com/reactioncommerce/reaction-data-loader/issues/5).

imports/plugins/core/ui/client/components/link/link.js
dancastellon: @zenweasel I just added a JSDoc and created a docs page PR: https://github.com/reactioncommerce/reaction-docs/pull/653

dancastellon:
@zenweasel Re: currency switcher not working. I went as far back as 1.13, and the currency switcher's not working there either. @aldeed Said:

> Any currency is supported, but only one. You set the currency on the shop and all prices are assumed to be that. I don’t think we test other currencies, though, so there may be some bugs (hard coded “$” in a few places, etc.)

dancastellon:
@zenweasel Same with the search not displaying images. I created an issue [here](https://github.com/reactioncommerce/reaction-data-loader/issues/6).

Also:
- Do we want the grid in Storefront to match the grid in the Meteor app? That was my impression when I ported Storefront grid to the component library. The styling in the Meteor grid is now different than it used to be, but it's at least consistent with Storefront.
- It might take a good chunk of effort to get the search modal grid to use the grid from the component library, since the structure of the data is different. Perhaps I should just style the current search component to look the same as the new grid?

aldeed:
@dancastellon Use newer styles everywhere in the Meteor app whenever there is a question of styling.

aldeed:
@dancastellon I'm not sure if it will help make things easier for this, but here's one way you could quickly add currency support to pricing in GraphQL.

1. Update `CatalogProduct.pricing` resolver to take two optional args, `currencyCode` and `exchangeRate`. Since they're optional, existing queries should still work.
1. Add `currencyExchangePricing: CurrencyExchangeProductPricingInfo` to `ProductPricingInfo` GraphQL type, where `CurrencyExchangeProductPricingInfo` is a new type that is identical to `ProductPricingInfo` but without `currencyExchangePricing` field.
1. If the new args are provided to `CatalogProduct.pricing`, run all the prices through the currency exchange function and add the `currencyExchangePricing` field to the output.
1. In client grid code, use `pricing.currencyExchangePricing.displayPrice` instead of `pricing.displayPrice`. Would also need to rerun the query when the user-selected currency changes.

ajporlante:
The currency conversion does not work with a newly created product.

dancastellon:
@ajporlante Thanks - it looks like there were a couple core bugs going on here. Just pushed the fixes for these:

1. When a product with a single variant was created, the price object for the product was not always being set/updated.
2. Product search records were being updated on every change to the product, even before those changes were published.

imports/plugins/core/dashboard/client/components/actionView.js
```diff
@@ -177,12 +178,17 @@ class ActionView extends Component {
...
+    if (actionView) {
+      this.setState({ actionView });
+    }
   }
 
   componentDidUpdate(prevProps) {
     const { actionView } = this.props;
 
-    if (actionView.template && actionView.template !== prevProps.actionView.template) {
+    if (EJSON.equals(actionView, prevProps.actionView) === false) {
```
aldeed: Can you use `_.isEqual` here to avoid using a Meteor package?

imports/plugins/core/ui/client/components/link/link.js
```diff
@@ -0,0 +1,33 @@
...
+  static propTypes = {
+    href: PropTypes.string.isRequired
+  };
+
+  handleClick = (event) => {
+    event.preventDefault();
+    this.props.onClick(event);
+    ReactionRouter.go(this.props.href); // eslint-disable-line no-undef
+    window.scrollTo(0, 0);
```
aldeed: In order to get smooth scrolling in newer browsers and guard against errors in tests or older browsers, maybe update this to:

```js
if (typeof window !== "undefined" && typeof window.scrollTo === "function") {
  window.scrollTo({ left: 0, top: 0, behavior: "smooth" });
}
```

zenweasel:
Selecting a product and clicking on "Clone" does nothing

zenweasel:
Click on "Publish" from the main catalog grid gives this error:

```
publishContainer.js:17 Uncaught TypeError: Cannot read property 'filter' of undefined
    at Object.PublishContainer.handlePublishClick [as onPublishClick] (publishContainer.js:17)
    at PublishControls.handlePublishClick (publishControls.js:106)
    at Button.handleClick.event (button.jsx:25)
    at HTMLUnknownElement.callCallback (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:176208)
    at Object.invokeGuardedCallbackDev (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:176246)
    at Object.invokeGuardedCallback (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:176295)
    at Object.invokeGuardedCallbackAndCatchFirstError (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:176309)
    at executeDispatch (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:176569)
    at executeDispatchesInOrder (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:176591)
    at executeDispatchesAndRelease (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:176689)
    at executeDispatchesAndReleaseTopLevel (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:176700)
    at forEachAccumulated (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:176670)
    at runEventsInBatch (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:176831)
    at runExtractedEventsInBatch (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:176840)
    at handleTopLevel (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:180585)
    at batchedUpdates$1 (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:192768)
    at batchedUpdates (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:178239)
    at dispatchEvent (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:180664)
    at interactiveUpdates$1 (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:192823)
    at interactiveUpdates (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:178258)
    at dispatchInteractiveEvent (modules.js?hash=0a771b9aa9d32f184fc0b3a078faea3b0c887a30:180641)
```

dancastellon:
@zenweasel Changes made. After pulling in the latest from 1.17 I'm not seeing any console errors when clicking "Publish" from the main grid. Let me know if you're able to reproduce it. Thanks!

zenweasel:
In Firefox I am seeing a strange behavior when switching from one grid to the next where I see one huge product image for a moment, and then a spinner and then finally the grid.

firefox_grid

zenweasel:
Same issue in Safari although the flash is faster

zenweasel:
There's an issue with Firefox/Safari but it's a minor visual glitch so I think we can just open an issue for that separately (I expect it happens in the 2.0 branch as well) and deal with it there to not hold up this PR any more.

dancastellon:
From @aldeed - we'll be merging the 2.0 version of this PR. Closing this one.

spencern:
@dancastellon do we need to merge a version of this to resolve issues with Search #4639? Is that issue present in v1.16.0?

dancastellon:
@spencern yep, we merged the 2.0 version of this PR, which fixed #4639.

Issue #4525, opened 2018-08-08
Identity and Access Management with Keycloak Epic


impactmass:
## Goal
Replace Reaction’s usage of Meteor auth with an identity and authorization system based on OAuth2 and OpenID Connect.

## Stages
- Stage 1: Abstract existing identity & authorization code from Meteor (and other Meteor-related auth lib) into wrapper functions e.g `Meteor.userId()` `Roles.userIsInRole`. The wrapper functions will contain logic to determine the right auth system based on a set ENV.
- Stage 2: Implement Keycloak-based authentication and authorization (while keeping existing Meteor auth and `Roles` package. Selecting an auth system to be done with an ENV var)
- Stage 3: Remove Meteor accounts system and `Roles` package

### Stage 1

**Client code:** 
- Setup a wrapper function that returns the correct logged in user ID depending on the enabled authentication system (i.e either Meteor-Accounts or Keycloak). Replace all `Meteor.userId()` in client code with the new function.
- Replace direct usage of `Roles.userIsInRole` to `hasPermission` wrapper function to get us closer to be able to remove dependence on [Roles](https://github.com/alanning/meteor-roles) package.
- Setup a wrapper function to be used for authorization checks in the client code. This function will delegate to the correct authorization lib depending on the enabled authentication system. Replace usage of `hasPermission()` with the new function. 
- Remove current Anonymous login from the client code. It was mostly used for Cart setup which changed in master to use a cart token.
- Update _all_ Meteor.user() references to use the Accounts collection

**Server code:**
- Remove Meteor.userId() calls (to remove dependence on Meteor Accounts). Switch to this.userId. Confirm that this works across existing methods & publications.
- Unify hasPermission checks between the GQL server and Meteor server. Remove the old version.
- Setup a wrapper function to be used for authorization checks in the server code. This function will delegate to the correct authorization lib depending on the enabled authentication system. Replace usage of `hasPermission()` with the new function.  

**Both server & client:**
- Change usage of existing hasPermission extensions (e.g hasOwnerAccess, hasAdminAccess) to use just hasPermission function. Add a deprecation warning to those (hasAdminAccess etc) functions.

All work done at this stage should cause no breaking change.

### Stage 2

- Setup initialization of Keycloak-js lib and OAuth dance in the client code. Add Login / Logout button in the account dropdown. Completion of successful login should store the returned token in local storage. Also on successful login, set the userId context for all method calls
[...]

PR #4586, opened 2018-08-29
WIP - Fix 2895 impactmass has permisssion owner fix

### LINKED ISSUE #2895: Update Reaction.hasPermission's owner pass and replace instances of Roles.userIsInRole

impactmass:
Resolves #2895  
Type: **fix|?feat**

## Task
Fix `Reaction.hasPermission` to work properly with “anonymous” role check.

Reason: This change is needed so we can be able to replace ALL current use of `Roles.userIsInRole` in the code to `hasPermission` (thus paving the way for removing dependency on Roles package later). See issues #2895 (old but yet-to-be-fixed issue) and #4525 for more details.

## Overview
**Brief summary chosen approach:** Add an extra `option` param to hasPermission. This will allow us to call hasPermission in a no “appendOwner” mode. This mode will not append the “owner” role as we currently have it. Existing checks that do not pass a “appendOwner” flag will go ahead with appending the “owner” role.

## Other Approaches
**Approach A:** Remove the addition/appending of the “owner” role happening within the hasPermission method. This will require explicitly adding “owner” to the list of roles in the arguments passed to ALL existing hasPermission calls.

**Why not chosen:** This approach will be very extensive (with a large surface area of code to change). It will also be a “avoidable” breaking change for plugins.

## Negatives
None that I can think of. Going the route of adding an extra field doesn't affect current code setup.

## Structure
Where will this code live: The changes are done in the current Reaction.hasPermission (in the frontend code and server code)

## User Interface Changes
None

## Breaking Changes
None

## Future Effects
What effects on future code (either positive or negative) will this have?

**Positive:** The addition of an options object introduces a way to later extend the behavior of this function. More options can be added when needed.

## Security Concerns
None

### DIFF
--- a/client/modules/core/helpers/apps.js
+++ b/client/modules/core/helpers/apps.js
@@ -1,6 +1,5 @@
 import _ from "lodash";
 import { Template } from "meteor/templating";
-import { Roles } from "meteor/alanning:roles";
 import { Reaction } from "/client/api";
 import { Packages, Shops } from "/lib/collections";
 
@@ -94,7 +93,7 @@ export function Apps(optionHash) {
           // This checks that the registry item contains a permissions matches with the user's permission for the shop
           const hasPermissionToRegistryItem = item.permissions.indexOf(permission) > -1;
           // This checks that the user's permission set have the right value that is on the registry item
-          const hasRoleAccessForShop = Roles.userIsInRole(getUserId(), permission, Reaction.getShopId());
+          const hasRoleAccessForShop = Reaction.hasPermission(permission, getUserId(), Reaction.getShopId());
 
           // both checks must pass for access to be granted
           if (hasPermissionToRegistryItem && hasRoleAccessForShop) {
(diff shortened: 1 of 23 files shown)


spencern:
@impactmass This ticket has been in the icebox for a bit, what's the status of this PR?

impactmass:
@spencern I was working on this as part of the Keycloak Auth Epic, which was paused in favor of setting up Hydra Auth for 2.0-release. Now, we are starting to revisit the next steps of our Identity and Access Management setup, and that will determine our approach to this PR.

spencern:
@impactmass ping on this issue.

impactmass:
At this point, the conversation around the roles and authorization piece of our IAM is steering in the direction of having a 'separate' roles/policy evaluation that is not tightly-coupled with one service (i.e the Reaction API service as we have it right now). 

It's still fairly open right now as no full decision has been made. 

The work in progress on this PR (if completed) is useful if our strategy on roles and auth follows the current path.

I will say I close this PR, but keep the branch up in case(?).

PR #4607, opened 2018-09-06
fix: cart item attributes

### LINKED ISSUE #4606: Title attribute on cart items uses wrong value
## Issue Description

Item attributes when items are added to the cart are displayed incorrectly in the starter-kit and in the cart items component from the component library. Namely having the title attribute uses the `optionTitle` instead of `title`.

## Possible Solution

Here: https://github.com/reactioncommerce/reaction/blob/release-1.16.0/imports/plugins/core/cart/server/no-meteor/util/addCartItems.js#L99

1. Use `optionTitle` from the parent variant if available, otherwise...
1. Use `title` from the chosen variant
1. Add `label: null` for the future to be clear that there will be a label.

### Versions

Node: 10.7.0
NPM: 6.2.0
Meteor Node: 8.11.3
Meteor NPM: 5.10.0
Reaction CLI: 0.29.0
Reaction: 1.16.0
Reaction branch: release-1.16.0
Docker: 18.06.0-ce

mikemurray:
Resolves #4606  
Impact: **minor**  
Type: **bugfix**

## Issue

Item attributes for cart items use incorrect values. The `optionTitle` is used in place of `title` which sometimes leads to incorrect values being saved into the database.

## Solution

- Use only the title for the attributes

## Testing
1. Using reaction with [sample data](https://github.com/reactioncommerce/reaction-catalog-sample-data)
1. Add items to the cart
2. Open database in Robo3T to port `27017`, if running in docker
3. In the `Cart` collection, expand what should be the only cart
4. Check the items attributes They should look something like for following depending on products added to the cart. (there should be no `Unitited Option` attributes) 

robo_3t_-_1_2

### DIFF
--- a/imports/plugins/core/cart/server/no-meteor/util/addCartItems.js
+++ b/imports/plugins/core/cart/server/no-meteor/util/addCartItems.js
@@ -95,10 +95,12 @@ export default async function addCartItems(collections, currentItems, inputItems
     const attributes = [];
     if (parentVariant) {
       attributes.push({
+        label: null, // Set label to null for now. We expect to use it in the future.
         value: parentVariant.title
       });
     }
     attributes.push({
+      label: null, // Set label to null for now. We expect to use it in the future.
       value: chosenVariant.title
     });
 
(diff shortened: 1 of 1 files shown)


mikemurray:
@aldeed I used `title` for both because the issue with the wrong attribute was with the first attribute conditional

```
    if (parentVariant) {
      attributes.push({ value: parentVariant.optionTitle || parentVariant.title });
      attributes.push({
        label: null, // Set label to null for now. We expect to use it in the future.
        value: parentVariant.title // <---------------------- HERE
      });
    }
```

PR #4609, opened 2018-09-07
fix: Add missing Shops.layout migration

### LINKED ISSUE #4608: Missed Shops migration for Products template
## Issue
Migration 33 was added in v1.15.0 to change the "products" template name to "Products". It neglected to do this in the Shops collection, which also has layouts stored there. Depending on whether you are upgrading, whether your Packages collection is regenerated on startup, and what version your database is already on, this sometimes leads to the app starting as just a white screen with no errors in the console.

This is actually pretty likely to happen in an upgrade situation.

## Solution
Add another migration to convert the Shops collection.

aldeed:
Resolves #4608 (hopefully also #4585)
Impact: **minor**  
Type: **bugfix**

## Issue
Migration 33 was added in v1.15.0 to change the "products" template name to "Products". It neglected to do this in the Shops collection, which also has layouts stored there. Depending on whether you are upgrading, whether your Packages collection is regenerated on startup, and what version your database is already on, this sometimes leads to the app starting as just a white screen with no errors in the console.

This is actually pretty likely to happen in an upgrade situation.

## Solution
Adds another migration, covering what the first migration missed.

## Other changes
I also made a couple minor fixes to migration 13. This was something I encountered while debugging. If any carts or orders do not have a `shipping` field set, then setting `shipping.0.shopId` in the migration actually creates a `shipping` object rather than an array. This is a known quirk of MongoDB. So I added a filter to run the migration only for documents that already have an array there.

Since this has only caused problems in a weird artificial situation I created, I opted to only fix the existing migration rather than write a new one.

## Breaking changes
None

## Testing
This is easiest to test if you already have a database that's causing this issue. Otherwise, you can fake it by manually creating the problem in your database and then verify the migration fixes it:

1. In your primary shop, look at the `layout` array and find the item with `workflow = "coreWorkflow"`. Change `structure.template` of that item to "products" (lowercase first letter).
1. When you start the Meteor app (on release-1.16.0 branch, without this fix), it should now be a blank screen.
1. Stop the app, check out this branch, start the app and verify that migration 38 runs. If you now refresh the browser, the app should appear normally.

### DIFF
--- a/imports/plugins/core/versions/server/migrations/13_add_shopId_on_shipping.js
+++ b/imports/plugins/core/versions/server/migrations/13_add_shopId_on_shipping.js
@@ -2,6 +2,8 @@ import { Migrations } from "meteor/percolate:migrations";
 import { Cart, Orders } from "/lib/collections";
 import Reaction from "/imports/plugins/core/core/server/Reaction";
 
+const shippingFieldExistsAndHasItems = { shipping: { $exists: true, $ne: [] } };
+
 Migrations.add({
   version: 13,
   up() {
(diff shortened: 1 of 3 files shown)


zenweasel:
Tested via:

1. Start app on `v1.14.0`
1. Move to `master` and start app
1. Observe blank screen
1. Move to this branch
1. Observe that page renders normally
👍 🎉

PR #4635, opened 2018-09-18
Wait for mongodb replica set ready at reaction startup

### LINKED ISSUE #4629: Ensure MongoDB replica set command is finished before starting Reaction
## Background
In docker-compose, `reaction` depends on `mongo-init-replica` in order to initialize a replica set because Meteor reactivity needs oplog support.

## Problem
`mongo-init-replica` exits after running the mongo command, but it takes a bit of time for the replica set to fully initialize. This can cause Reaction to start up with errors.

## Solution
There are different possible solutions.
- Change the startup script for `reaction` to wait until there is a confirmed replica set before running `reaction` command
- Change `mongo-init-replica` to run a script that not only runs the command but won't exit until the replica set init is done.

focusaurus:
Resolves #4629   
Impact: **minor**  
Type: **bugfix**

## Issue
There's a docker startup race condition which can cause the application to not start if the mongodb replica set is not ready in time. See #4629 for details.

## Solution

Add a small node program to reaction startup that polls for replica set ready before starting reaction.

## Testing

  - Test against pristine install, should wait for replica then start properly without much delay
  - Test against existing install with healthy replica, should start properly without much/any delay
  - Test against pristine install where replica never becomes ready
   - You can force this by hacking the `docker-compose.yml` file to have reaction depend on `mongo` instead of `mongo-init-replica`. It should retry for 30s then fail with a clear error message and not starting reaction.
   - Same setup can test retry if you manually run `docker-compose up -d mongo-init-replica` while the reaction container is wating (you have a 30s window)

There are also some automated unit tests for the core retry logic code.

## Lint

Both files pass eslint for me locally but the `.reaction` directory is ignored in our eslint config it seems. It is likely my jsdoc does not meet all of our conventions but I put something reasonable in there.


focusaurus:
Gah wrong target branch. Closing.

PR #4636, opened 2018-09-18
Ensure MongoDB replica set is ready before start

### LINKED ISSUE #4629: Ensure MongoDB replica set command is finished before starting Reaction

focusaurus:
Resolves #4629   
Impact: **minor**  
Type: **bugfix**

## Issue
There's a docker startup race condition which can cause the application to not start if the mongodb replica set is not ready in time. See #4629 for details.

## Solution

Add a small node program to reaction startup that polls for replica set ready before starting reaction.

## Testing

  - Test against pristine install, should wait for replica then start properly without much delay
  - Test against existing install with healthy replica, should start properly without much/any delay
  - Test against pristine install where replica never becomes ready
   - You can force this by hacking the `docker-compose.yml` file to have reaction depend on `mongo` instead of `mongo-init-replica`. It should retry for 30s then fail with a clear error message and not starting reaction.
   - Same setup can test retry if you manually run `docker-compose up -d mongo-init-replica` while the reaction container is wating (you have a 30s window)

There are also some automated unit tests for the core retry logic code.

## Lint

Both files pass eslint for me locally but the `.reaction` directory is ignored in our eslint config it seems. It is likely my jsdoc does not meet all of our conventions but I put something reasonable in there.

### DIFF
--- a/.reaction/waitForReplica.js
+++ b/.reaction/waitForReplica.js
@@ -1,0 +1,144 @@
+const { MongoClient } = require("mongodb");
+
+/**
+ * Print a message to the console (no trailing newline)
+ * @param {string} message User progress message to print
+ * @returns {undefined}
+ */
+function defaultOut(message) {
+  process.stdout.write(message);
+}
+
+/**
+ * Print a message to the console (no trailing newline)
+ * @param {Object} options - Named options object
+ * @param {function()} options.out Function to show progress
+ * @param {number} options.max Number of retries attempted before full failure
+ * @param {function():Promise} options.check async/Promise function that implements
+ *   the main check operation. Throw an exception to cause a retry. Resolve the
+ *   promise to indicate success.
+ * @param {number} options.waitMs milliseconds to wait between checks
+ * @param {string} options.timeoutMessage for final timeout
+ * @returns {Promise} a promise indicating success/failure of the check
+ */
+async function checkWaitRetry({
+  out = defaultOut,
+  max = 30,
+  check = null,
+  waitMs = 1000,
+  timeoutMessage = "Timed out waiting for a prerequisite"
+} = {}) {
+  const messages = new Set();
+  /**
+   * Show a progress/info message, but not over and over again
+   *
+   * @param {string} message to be printed
+   * @param {number} count retry number for progress dots
+  * @returns {undefined}
+  */
+  function showOnce(message, count) {
+    if (!messages.has(message)) {
+      messages.add(message);
+      out(message);
+    }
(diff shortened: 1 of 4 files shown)


.reaction/waitForReplica.js
```diff
@@ -0,0 +1,161 @@
...
+  34:22  error    Expected parentheses around arrow function argument  arrow-parens
+  37:1   warning  Missing JSDoc comment                                require-jsdoc
+  39:3   warning  Missing JSDoc comment                                require-jsdoc
+  65:5   error    Unary operator '++' used                             no-plusplus
+  67:16  error    Unexpected `await` inside a loop                     no-await-in-loop
+  70:7   error    Unexpected `await` inside a loop                     no-await-in-loop
+  78:5   error    Unexpected `await` inside a loop                     no-await-in-loop
+
+✖ 11 problems (5 errors, 6 warnings)
+*/
```
aldeed: Did you intend to leave this comment in here?

tests/waitForReplica.test.js
```diff
@@ -0,0 +1,75 @@
+import { checkWaitRetry } from "../.reaction/waitForReplica";
+
+describe("waitForReplica.checkWaitRetry", () => {
+  it("should work in immediate success case", async () => {
```
aldeed: We use `test` instead of `it`, and we don't wrap in a `describe` block if there's just one in the file. The rest looks fine. More details here: https://docs.reactioncommerce.com/docs/writing-jest-unit-tests

.reaction/waitForReplica.js
focusaurus: Nope, just notes to self. Good catch. My eslint editor integration isn't set up yet for this environment. Fixed with an amend and force push.

tests/waitForReplica.test.js
focusaurus: Oh that's really nice. Hadn't found that guide yet, was basically assuming a mocha-compatible API layer. Fixed.

Issue #4639, opened 2018-09-18
Product search result has blank image


vchong88:
## Prerequisites
* [X] Are you running the latest version?
* [X] Are you able to consistently reproduce the issue?
* [X] Did you search the issue queue for existing issue? 

## Issue Description
When accessing RC as non-login user, the Product Search results does not contain product image
<img width="600" alt="screen shot 2018-09-18 at 12 02 53 pm" src="https://user-images.githubusercontent.com/22272050/45663956-bf322800-bb3b-11e8-9707-9def9eb336a4.png">

## Steps to Reproduce
Please provide starting context, i.e. logged in as a user, configure a particular payment method.
1. Open browser as incognito mode
2. Search product based on product tag
3. Search result does not contain product image

Note: If you have previously login as user
1. Sign-out user
2. Refresh browser
3. Search product based on product tag
4. Search result does not contain product image

## Possible Solution
Allow product search images to be displayed to non-login user

### Versions
Node: 6.2.1
NPM: 6.1.0
Meteor Node: 8.11.3
Meteor NPM: 5.10.0
Reaction CLI: 0.29.0
Reaction: 1.16.0
Reaction branch: master
Docker: 18.03.1-ce

dancastellon:
@vchong88 I noticed this while implementing our new `CatalogGrid` component in the Meteor app. Fix should be merged soon!

dancastellon:
Closing this - merging the 2.0 version of the catalog grid PR fixed this.

PR #4649, opened 2018-09-20
feat: use component library's CatalogGrid - 2.0

### LINKED ISSUE #4583: Use Component Library's CatalogGrid for product grid

dancastellon:
Note: This is the 2.0 version of this PR. The 1.17 version can be found here: https://github.com/reactioncommerce/reaction/pull/4584

Resolves #4583 and #4639 
Impact: **minor**  
Type: **feature**

## Solution
This PR replaces the current customer grid UI with the new `CatalogGrid` component.

## Breaking changes
Only breaking for custom sites as the customer-facing `productGridItem` component has been removed.

## Testing
1. Use reaction-data-loader on its 1.14 branch to load some products: `npx babel-node data/generator.js -m mongodb://localhost:3001/ -l dev -n meteor --max_old_space_size=8192 --presets es2015,stage-2`
2. Confirm grid with new styles works as expected - progressively loads images, responsive from mobile up, loading more as you scroll to the bottom
3. Currency - Change the currency via the dropdown in the header. Confirm prices are converted and displayed correctly, including prices where only 1 currency symbol should be displayed with a price range (RUB)
4. Search - Confirm improved search modal that uses the new grid, displays images (even when logged out), is perfectly responsive at any size, shows correct currency, and brings you to the PDP on click. NOTE: If using the data loader, the correct primary image may not be displayed due to a bug: https://github.com/reactioncommerce/reaction-data-loader/issues/8

### Badge label testing:
Note: Might be easier to create a new product and copy the PDP URL. If you create a product and try to find it in the admin UI (after loading the dev dataset), it'll take a while since the customer grid is sorted by newest first, but the admin grid is not.

1. Create a product, add title to variant, set quantity to 1, warn at 1. Publish & logout
2. Confirm “Limited supply” badge
3. Switch to other languages and confirm badge text updates reactively
4. Login and edit product, set variant quantity to 0, turn allow backorder off. Publish & logout
5. Confirm “Sold out” badge. Confirm translations
6. Login and turn allow backorder on for product. Publish & logout
7. Confirm “Backorder” label & translations

### Price range & search record update testing
Two core bugs were discovered while work on this PR: `product.price` was not always being updated, and search records were being updated before product changes were published. To confirm the fixes:
1. Create a product, add a title to the variant, set quantity, don't publish
2. Confirm the product doesn't appear in search yet
3. Publish the product
4. Confirm the product now does appear in search + the price range is correct, both in the main customer grid and search results
5. Edit the price for the variant and/or add a new visible variant with a price
6. Confirm in DB product.price range, min, and max are correct, but that the changes haven't propagated to search results yet
7. Publish the changes and confirm the search result's price range has updated
8. Confirm 6 & 7 with variants that have options.


dancastellon:
@zenweasel I applied the clone fix to this PR as well. Thanks!

zenweasel:
Might be a bit of a nitpick, but a longer title seems to break the price. Seems like the title should break on it's own plane.

zenweasel:
This may have been an existing error but is there any way we can fix this error I see in the console:

```
Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in SearchModal (created by SearchSubscription)
    in SearchSubscription (created by Reaction(SearchSubscription))
    in Reaction(SearchSubscription) (created by SearchModalContainer)
    in div (created by SearchModalContainer)
```

zenweasel:
Changed to `master` for now so it doesn't get accidentally merged in 2.0 but should probably go into 2.1

dancastellon:
@aldeed Pulled from 2.0-RC5 and confirmed this one's good to merge.

PR #4651, opened 2018-09-20
feat: Show correct form state during create account oauth flow

impactmass:
Part of reactioncommerce/reaction/issues/4640
Type: **feature**

## Issue
When a Starterkit user clicks "Register", we redirect to the Reaction Hydra login page, but it shows the "Sign In" view rather than "Create Account" view.

## Solution
Use extra query info passed from the client app (Starterkit) in the auth URL in Hydra (reactioncommerce/reaction-next-starterkit/pull/320). 

## Changes
- Get the auth URL connected to the Login request from Hydra. Get the `loginAction` field and pass it in the UI route query param (?action={loginAction}) to serve as state info to the Login component.
- Update the Auth components to show appropriate form fields based on route query param (e.g ?action=signin)
- Update the SignIn and SignUp core components to have a new `hasSwitchLinks` prop. This will determine if link to toggle between SignIn and SignUp views should be displayed. It defaults to true (so previous behaviour is kept - no breaking change)
- Update Auth container for OAuth IDP flow with hasSwitchLinks=false. This makes the state of the form depend on ONLY the route query
- Fix the Hydra session feature. Now, when a user who already signed in (and is sill within the set HYDRA_SESSION_LIFESPAN), tries to login again, we won't show the login form again.
- Add /logout endpoint for Consumer apps (like Starterkit) to call to delete user sessions from Hydra. The delete session endpoint in Hydra lives on the Administrative port (4445), so we are not exposing it to Consumer apps to consume directly.

## Breaking changes
N/A

## Testing
NB: This is to be tested along with reactioncommerce/reaction-next-starterkit/pull/320

1. From Starterkit, click "Sign In"
2. Confirm that you see a "Sign In" form on the Meteor Login page
3. From Starterkit, click "Create Account"
4. Confirm that you see a "Create Account" form on the Meteor Login page

Regression Testing
This PR touched SignIn and SignUp core components
1. In the Meteor app, open the login dropdown
2. Confirm that the SignIn and SignUp components behave as before

### DIFF
--- a/imports/plugins/core/accounts/client/components/login.js
+++ b/imports/plugins/core/accounts/client/components/login.js
@@ -63,10 +63,7 @@ class Login extends Component {
           <Components.OAuthFormContainer
             credentials={this.props.credentials}
             uniqueId={this.props.uniqueId}
-            currentView={this.state.currentView}
             onForgotPasswordClick={this.showForgotPasswordView}
-            onSignUpClick={this.showSignUpView}
-            onSignInClick={this.showSignInView}
           />
         );
       }
--- a/imports/plugins/core/accounts/client/components/signIn.js
+++ b/imports/plugins/core/accounts/client/components/signIn.js
@@ -8,6 +8,7 @@ import { Components, registerComponent } from "@reactioncommerce/reaction-compon
  * @memberof Accounts
  * @extends {Component}
  * @property {Object} credentials
+ * @property {Boolean} hasSwitchLinks
  * @property {Boolean} isLoading
  * @property {Function} loginFormMessages
  * @property {Object} messages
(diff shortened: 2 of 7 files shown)


mikemurray:
@impactmass I'm getting an infinite spinner when trying to sign in on the reaction side after initiating the login flow from the starter kit. Also, I'm not seeing any errors in of the logs for reaction, starter kit, or hydra.

impactmass:
thanks @mikemurray. Last prop change caused that one (my bad!). I have that fixed now, but seeing another issue from Passport throwing a `Forbidden` error. I'm working on resolving that.

impactmass:
Ready for another review

imports/node-app/core/util/getUserFromAuthToken.js
```diff
@@ -30,7 +31,7 @@ async function getUserFromAuthToken(loginToken, context) {
     throw new ReactionError("access-denied", "Token is not an access token");
   }
 
-  const _id = tokenObj.sub;
+  const _id = decodeAccountOpaqueId(tokenObj.sub);
```
impactmass: In a discussion with @ticean about encoding the `id` in this situation, it seems preferred to keep this as part of only GraphQL API.

PR #4670, opened 2018-09-24
(fix) Properly return 0 when no tax items are present

### LINKED ISSUE #4669: Cart totals show `NaN` when taxes are not defined
## Issue Description
Cart totals would sometimes show `NaN` when there were no cart items because `getTaxTotals` did not account for that possibility. This was caused by my earlier fix to custom taxes not taking that case into account

## Steps to Reproduce
1. Reset database
1. Attempt to checkout

## Possible Solution
PR forthcoming

### Versions
Reaction: 2.0.0-rc1

brent-hoover:
Resolves #4669 
Impact: **major*  
Type: **bugfix**

## Issue
If `item.tax` did not exist the `getTaxTotal` was returning `NaN` which would show up in cart totals. This was a result of my previous PR fixing custom taxes.

## Solution
Properly return 0 when these don't exist

## Testing
1. Start system with fresh db and don't enable taxes
1. Attempt to checkout
1. Observe that that the cart total is always calculated correctly


aldeed:
Approved, but we don't yet have a 2.0.0-rc.2 branch for merging this into.

PR #4673, opened 2018-09-24
Fix: console error when typing a tag name on PDP

### LINKED ISSUE #4438: Console error when typing tag names to add on PDP, admin sidebar
## Prerequisites
* [x] Are you running the latest version? 1.14.0

## Issue Description
On a fresh 1.14.0 install, when typing a name as an admin adding tags on the PDP, I get the following browser console error:
`Uncaught (in promise) TypeError: Cannot read property 'getShopId' of undefined
    at getShopLang (helpers.js:118)
    at lazyLoadSlugify (utils.js:8)
    at Object.getSlug (utils.js:40)
    at updateSuggestions (tagListContainer.js:17)
    at Object.TagListContainer.handleGetSuggestions.suggestionUpdateRequest [as onGetSuggestions] (tagListContainer.js:120)
    at TagItem.handleSuggestionsUpdateRequested.suggestion (tagItem.js:66)
    at onChange (modules.js?hash=432c9364a4be7e28f945d7307cb1b698d47a6ea0:268611)
    at HTMLUnknownElement.callCallback (modules.js?hash=432c9364a4be7e28f945d7307cb1b698d47a6ea0:250346)
    at Object.invokeGuardedCallbackDev (modules.js?hash=432c9364a4be7e28f945d7307cb1b698d47a6ea0:250385)
    at Object.invokeGuardedCallback (modules.js?hash=432c9364a4be7e28f945d7307cb1b698d47a6ea0:250242)`

## Steps to Reproduce
1. Clone and startup release-1.14.0
2. Edit a product and start typing in the new tag auto-complete field
3. Confirm browser console error

## Possible Solution
Updating line 57 in /lib/api/helpers.js appears to fix the issue:
`import("/client/api").then(module => {
    Reaction = module.Reaction;
  });`

### Versions
Node: 8.11.3
NPM: 5.6.0
Meteor Node: 8.11.3
Meteor NPM: 5.10.0
Reaction CLI: 0.29.0
Reaction: 1.14.0
Reaction branch: release-1.14.0

dancastellon:
Resolves #4438 
Impact: **minor**  
Type: **bugfix**

## Issue
On the PDP, when typing a tag name in the admin sidebar, a console error would appear: `Uncaught (in promise) TypeError: Cannot read property 'getShopId' of undefined at getShopLang (helpers.js:118)`. This was because `Reaction` was undefined when `getShopLang()` would run.

## Solution
The solution was changing a client-side `require` statement to a simple dynamic import.

## Breaking changes
None

## Testing
1. Visit PDP as admin
2. Click edit icon next to tags
3. Start typing a new tag name, confirm no console errors
4. Edit tag nav, start typing an existing tag name, confirm suggestions still work
5. Admin -> Accounts -> New group - Create a new group and confirm in the DB it has a correctly-formatted-slug


aldeed:
Also, given the timing nature of this, I believe the only true fix would be to make `Reaction` a reactive var so that all of the helpers rerun after it is set (in addition to the async change you already made).

aldeed:
Actually, @dancastellon, I looked at this file a bit closer. I know we've had repeated problems with the Reaction import here due to the circular nature of this import, so I'd prefer a final solution rather than a bandaid. I think we can completely get rid of the `Reaction` import in this helpers file:

- `getShopSettings` is unused. Delete it
- `getPrimaryShopLang` is unused. Delete it.
- `getAbsoluteUrl` is unused. Delete it
- Now `getShopPrefix` is unused. Delete it
- Now `getShopName` is unused. Delete it.
- As far as I can tell, the `getSlug` in helpers is only imported in `/imports/plugins/core/accounts/server/methods/group/createGroup.js` and is the same as `Reaction.getSlug`. If you can update that one place to use `Reaction.getSlug`, then you can also remove `getSlug` and `lazyLoadSlugify` from the helpers file.

Pretty sure that's all that's necessary to get rid of circular dependencies that are causing this.

dancastellon:
@aldeed Changes made!

aldeed:
@dancastellon It looks like the getSlug change is causing an app test to fail

dancastellon:
@aldeed App test fixed

client/modules/core/main.js
```diff
@@ -661,6 +669,44 @@ export default {
...
+      } else {
+        mod = await import("transliteration");
+      }
+
+      // Slugify is exported to modules.default while transliteration is exported to modules.slugify
+      slugify = mod.default || mod.slugify;
+    };
+
+    let slug;
+    Promise.resolve(lazyLoadSlugify()); // eslint-disable-line promise/catch-or-return
```
aldeed: `Promise.resolve` doesn't do anything here. It won't await, so the first time this is run, the return will always be "".

You can see this issue by opening the console and entering `ReactionCore.getSlug("abc")` twice. It only works the second time.

I suspect this was probably an existing issue that we just hadn't noticed.

I think the only solution is to make this whole `getSlug` function async, and await `lazyLoadSlugify()` here.

The places I see that will then need to be changed:
- `getPrimaryShopPrefix` in the same file would need to be made async, but it looks to be unused. If so, just delete it.
- `getShopPrefix` in the same file would need to be made async, but it looks to be unused. If so, just delete it. (`Reaction.getShopPrefix` is called a few times, but only from server code as far as I can see.)
- Call to `Reaction.getSlug` in `/imports/plugins/core/ui-tagnav/client/helpers/tags.js` needs to be made async
- Call to `Reaction.getSlug` in `/imports/plugins/core/ui/client/containers/tagListContainer.js` needs to be made async (btw, this `updateSuggestions` function is identical to the previous one, so maybe move that to a util file and share it?)

dancastellon:
@aldeed I made a new `getTagSuggestions` util function and made `getSlug` async.

PR #4674, opened 2018-09-25
Fix discounts and profile orders in 2.0

### LINKED ISSUE #4668: Unable to apply discounts
## Issue Description
I can create discounts in the admin but I never get prompted to enter a promo code so I am unable to apply it

## Steps to Reproduce
1. Create a promo code under the "Codes" section of the dashboard.
1. Attempt to check out
1. Observe that the prompt to enter a promo code never appears

### Versions
Reaction: 2.0.0 (`bfc65958f8b6257a64cf8e27824e12b3666b204c`) the released version

aldeed:
Resolves #4668   
Impact: **minor**  
Type: **bugfix**

## Changes
1. As reported in #4668, discounts in 2.0.0-rc.1 are not working. There is a client bug that prevents the checkout form from appearing in some cases. Even if it does appear, there is a calculation bug that causes the `cart.discount` value to be a string instead of a number, leading to the discounts not being applied properly. All of this is now fixed in this PR.
2. As noticed while doing this PR, we never updated the Orders list on the `/account/profile` page to work with the 2.0 order schema changes. This caused it to appear blank even when the account has orders. It's now updated to show the orders and get rid of all errors in the browser console.

## Breaking changes
None

## Testing
1. On the Payment admin form, enable "Codes" and create one of each type (calculation method) of discount.
1. Add some items to your cart and go through checkout. Before entering payment, verify that you see a link to add a promo code.
1. Try adding each code you created. Verify that you can add it, and that the calculated discount in the checkout summary is correct.
    - "shipping" code should reduce by the entire cost of shipping, if you selected the shipping method with the name that matches what was entered for the "Discount" value of the code. e.g., enter "Priority" in Discount field when creating the code and then choose Priority shipping at checkout.
    - "discount" should reduce by whatever percentage you entered in the Discount field. e.g., enter 50 and order should be discounted by 50%
    - "sale" should reduce the price of each item to whatever you entered in the Discount field (this one is currently pretty useless for real world)
    - "credit" should reduce the order total by the amount entered in Discount field. e.g., enter 5 and discount should be $5.00.
1. Verify that you can place an order with each type of discount. After placing the order, verify that the discount is shown correctly everywhere (PDF, emails, account profile orders list, admin orders list)

### DIFF
--- a/imports/plugins/core/accounts/client/components/ordersList.js
+++ b/imports/plugins/core/accounts/client/components/ordersList.js
@@ -7,37 +7,36 @@ import CompletedOrder from "../../../checkout/client/components/completedOrder";
  * @memberof Accounts
  * @extends {Component}
  * @property {Array} allOrdersInfo - array of orders
- * @property {Function} handeleDisplayMedia - function to display order image
  * @property {Boolean} isProfilePage - Profile or non-profile page
  */
 class OrdersList extends Component {
   static propTypes = {
-    allOrdersInfo: PropTypes.array,
+    allOrdersInfo: PropTypes.arrayOf(PropTypes.shape({
+      orderId: PropTypes.string.isRequired,
+      order: PropTypes.object,
+      paymentMethods: PropTypes.array
+    })),
     isProfilePage: PropTypes.bool
   }
 
   render() {
-    const { allOrdersInfo } = this.props;
+    const { allOrdersInfo, isProfilePage } = this.props;
 
     if (allOrdersInfo) {
       return (
         <div>
-          {allOrdersInfo.map((order) => {
-            const orderKey = order.orderId;
-            return (
-              <CompletedOrder
-                key={orderKey}
-                shops={order.shops}
-                order={order.order}
-                orderSummary={order.orderSummary}
-                paymentMethods={order.paymentMethods}
-                isProfilePage={this.props.isProfilePage}
-              />
-            );
-          })}
+          {allOrdersInfo.map((order) => (
+            <CompletedOrder
+              key={order.orderId}
+              order={order.order}
+              paymentMethods={order.paymentMethods}
+              isProfilePage={isProfilePage}
+            />
+          ))}
         </div>
       );
     }
+
     return (
       <div className="alert alert-info">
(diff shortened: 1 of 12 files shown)


zenweasel:
When I do the "SALE" method and attempt to checkout I get the message:

```
Client provided total price 70.15639999999999 for group with index 0, but actual total price is 40.186400000000006
```

zenweasel:
The "SALE" error was a bit hard to replicate. Here is one example where I applied the discount and went back and added more items to my cart. I was able to checkout several times with the SALE discount applied however so it's not an always thing.

aldeed:
@zenweasel I haven't been able to reproduce, but I'm thinking it may be a rounding difference. I did notice that the "Client provided total price" number in the error message wasn't the actual number being compared. So I fixed that. If you pull the latest and can reproduce the error again, I'd be interested to see if it's only off by a fraction of a cent. If so, then I can figure out how to ensure that client and server do the rounding the same way.

If it is a rounding issue, then it would make sense that only specific items combined with specific sale prices would result in the error.

zenweasel:
So I've been breaking this down and I think the issue is related to Tax not Discounts. So I am approving this PR and will open a separate issue.

aldeed:
@zenweasel I was able to get this to happen with tax enabled, but not reliably. I then adjusted the client math to happen in the same order as the server math, and I haven't seen the error since then. That doesn't mean for sure that it's fixed, but it seems to be a symptom of the inherent unreliability of floating point math in JS.

Pushed that change on this PR.

There are ways we could make this less prone to error, such as sending and comparing a rounded string representation of the amount, or using `bignumber.js` or `decimal.js` libraries. But those are larger discussions to be had.

zenweasel:
We are storing it everywhere as 1.999 but displaying it everywhere as $2.00. This is probably going to confuse somebody down the road. We probably shouldn't be storing currency values with a greater than cents accuracy.

zenweasel:
I've created a [separate issue](https://github.com/reactioncommerce/reaction/issues/4684) for the tax problem

spencern:
@zenweasel @aldeed do we need to apply this patch to the release-1.17.0 branch as well?

PR #4676, opened 2018-09-25
updating get in touch link

wnstn:
Resolves reactioncommerce/reaction-static#131
Impact: **minor**  
Type: **chore**

## Issue
Point Get a Demo button to the new Act On button.

## Solution
Updates href attribute to the correct url.

PR #4672, opened 2018-09-24
fix: broken link to schema docs

focusaurus:
Fixes #4671

Resolves #4671  
Impact: **minor**  
Type: **docs**

PR #4664, opened 2018-09-21
(fix) Fix custom taxes

zenweasel:
Resolves #4655 
Impact: **critical**  
Type: **bugfix**

## Issue
Sometimes it's `taxable` sometimes it's `isTaxable`. Also calculate taxes from cart.items rather than the global tax item.

## Solution
Make it `isTaxable`. Calculate taxes from `items.tax`

## Breaking changes
None

## Testing
1. Add a custom tax rate for CA
1. Check out with an address in CA
1. Observe that taxes are properly applied

PR #4689, opened 2018-09-27
Release 2.0.0-rc.2

spencern:
# v2.0.0-rc.2
This is our second **release candidate** for v2.0.0 of Reaction. Please check it out and let us know what works and what doesn't for you.

### OAuth Flow
- Get the auth URL connected to the Login request from Hydra. Get the loginAction field and pass it in the UI route query param (?action={loginAction}) to serve as state info to the Login component.(#4651)
- Update the Auth components to show appropriate form fields based on route query param (e.g ?action=signin)(#4651)
- Update the SignIn and SignUp core components to have a new hasSwitchLinks prop. This will determine if link to toggle between SignIn and SignUp views should be displayed. It defaults to true (so previous behaviour is kept - no breaking change)(#4651)
- Update Auth container for OAuth IDP flow with hasSwitchLinks=false. This makes the state of the form depend on ONLY the route query (#4651)
- Fix the Hydra session feature. Now, when a user who already signed in (and is sill within the set HYDRA_SESSION_LIFESPAN), tries to login again, we won't show the login form again.(#4651)
- Add /logout endpoint for Consumer apps (like Starterkit) to call to delete user sessions from Hydra. The delete session endpoint in Hydra lives on the Administrative port (4445), so we are not exposing it to Consumer apps to consume directly.(#4651)

## Taxes
If item.tax did not exist the getTaxTotal was returning NaN which would show up in cart totals. This was introduced in #4664 and is resolved in #4670

## Operator UI for editing product information
On the PDP, when typing a tag name in the admin sidebar, a console error would appear: `Uncaught (in promise) TypeError: Cannot read property 'getShopId' of undefined at getShopLang (helpers.js:118)`. This was because `Reaction` was undefined when `getShopLang()` would run. This is resolved in #4673

## Fixes
 - (fix) Properly return 0 when no tax items are present (#4670)
 - fix: Show correct form state during create account oauth flow (#4651)
 - Fix: console error when typing a tag name on PDP (#4673)
 - fix: The orders list on the /account/profile route is no longer blank (#4674)
 - fix: All discount methods are now calculated properly, and the link to enter a discount code properly appears during checkout (#4674)
 - fix: orders dashboard layout issue (#4688)
    Updated a few panel styles to fit within the viewport better and not end up behind the "modal overlay"

 ## Chores
 - chore: updating get in touch link (#4676)
 - chore: fix broken link in README to schema docs (#4672)


CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
-# v2.0.0
+# v2.0.0-rc.2
+WIP
+
+# v2.0.0-rc.1
 This is our first **release candidate** for v2.0.0 of Reaction. Please check it out and let us know what works and what doesn't for you.
```
zenweasel: *second?

spencern: It's the WIP changelog here, but yeah, the second.

PR #4709, opened 2018-10-04
Feat 4574 payment methods

### LINKED ISSUE #4574: GraphQL: Implement availablePaymentMethods query
Implement availablePaymentMethods query

rosshadden:
Resolves #4574   
Impact: **major**  
Type: **feature|refactor**

## Issue
Payment methods architecture needs to be refactored, as well as moved into GraphQL.

## Solution
This is a WIP refactor of the payment methods setup. It is either done or close to it, but I want @aldeed to look it over first and see if I'm missing anything.

The admin panel should also be modified to benefit from these changes, but that will be done in a separate PR.

# Breaking changes

The only "breaking" change I can think of is that `PaymentMethod@name` is now a `String` ( https://github.com/reactioncommerce/reaction/pull/4709/files#diff-1b440c506b0e1b83c8fcd1efb5b7aa8bL78 ). This should technically be a superset of the `PaymentMethodName` it was before, and thus not affect anything.

Otherwise, this PR is all additive, and we will not have breaking changes until some of the upcoming payment methods changes.

# Testing

Run the following query in GraphQL:

```graphql
query {
  paymentMethods(shopId: "<shopOpaqueId>") {
    name
    displayName
    pluginName
    isEnabled
  }
  availablePaymentMethods(shopId: "<shopOpaqueId>") {
    name
    displayName
    pluginName
    isEnabled
  }
}
```


spencern:
@rosshadden this is in review, but has failing tests and linting in CI

imports/plugins/core/payments/server/no-meteor/schemas/schema.graphql
```diff
@@ -71,9 +70,13 @@ union PaymentMethodData = ExamplePaymentMethodData
 
 "Describes a payment method"
 type PaymentMethod {
```
aldeed: Goal is for fields to always be sorted alpha. (Maybe there is some kind of graphql linter we can use?)

imports/plugins/core/core/server/Reaction/core.js
```diff
@@ -96,6 +96,11 @@ export default {
         functionsByType[type].push(...packageInfo.functionsByType[type]);
       });
     }
+    if (packageInfo.paymentMethods) {
+      for (const paymentMethod of packageInfo.paymentMethods) {
+        paymentMethods.add(paymentMethod);
```
aldeed: For some of the future PRs related to this, it might be necessary to have the name of the plugin that added the method, and just helpful in general. You can do this here:

```js
paymentMethods.add({ ...paymentMethod, pluginName: packageInfo.name  });
```

imports/plugins/core/payments/server/no-meteor/resolvers/Query/availablePaymentMethods.js
```diff
@@ -0,0 +1,10 @@
+import { decodeShopOpaqueId } from "@reactioncommerce/reaction-graphql-xforms/shop";
+import { paymentMethods } from "/imports/plugins/core/core/server/no-meteor/pluginRegistration";
+
+export default async function (_, { shopId }, context) {
+  const allPaymentMethods = Array.from(paymentMethods);
+  const dbShopId = decodeShopOpaqueId(shopId);
+  const shop = await context.queries.shopById(context, dbShopId);
+  return shop.availablePaymentMethods
+    .map((availableName) => allPaymentMethods.find(({ name }) => name === availableName));
```
aldeed: All of these lines except the `decodeShopOpaqueId` line should be moved to a new `availablePaymentMethods` query function in this plugin. We want to keep all real logic out of the resolvers. This function should only be:

```js
const dbShopId = decodeShopOpaqueId(shopId);
return context.queries.availablePaymentMethods(context, dbShopId);
```

See https://docs.reactioncommerce.com/docs/next/graphql-create-query

imports/plugins/core/payments/server/no-meteor/resolvers/Query/paymentMethods.js
```diff
@@ -0,0 +1,5 @@
+import { paymentMethods } from "/imports/plugins/core/core/server/no-meteor/pluginRegistration";
+
+export default async function () {
+  return Array.from(paymentMethods);
```
aldeed: Same as my comment on `availablePaymentMethods`, this should be moved to a separate query function. Also, you are missing JSdoc and the new query function will need a permissions check because getting the full list is intended only for admins. The permissions check will look like this:

```js
const { userHasPermission } = context;
if (!userHasPermission(["owner", "admin"], shopId)) {
  throw new ReactionError("access-denied", "Access denied");
}
```

aldeed:
@rosshadden One more thing I realized is the `paymentMethods` query will need to take `shopId` and add `isEnabled` prop to each method in the array. So you'll also need to add `isEnabled: Boolean!` to `type PaymentMethod` in the GraphQL schema

imports/plugins/core/core/server/no-meteor/pluginRegistration.js
```diff
@@ -3,3 +3,4 @@ export const mutations = {};
 export const queries = {};
 export const resolvers = {};
 export const schemas = [];
+export const paymentMethods = new Set();
```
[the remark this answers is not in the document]
rosshadden: No, good catch. As objects, the duplicates wouldn't be caught anyway. I'll make it an object keyed by the name. This does mean we'll have the key duplicated but it still makes sense to work with.

imports/plugins/core/payments/server/no-meteor/queries/availablePaymentMethods.js
```diff
@@ -0,0 +1,18 @@
...
+ * @summary TODO: query the Groups collection and return a MongoDB cursor
+ * @param {Object} context - an object containing the per-request state
+ * @param {String} shopId - shop ID for which to get payment methods
+ * @return {Object[]} TODO: Groups collection cursor
+ */
+export default async function (context, shopId) {
+  const shop = await context.queries.shopById(context, shopId);
+  const paymentMethodNames = Object.keys(paymentMethods);
+
+  return shop.availablePaymentMethods
```
aldeed: Change to `return (shop.availablePaymentMethods || [])`. Otherwise I get errors running the query until we make the migration to set this array.

imports/plugins/core/payments/server/no-meteor/queries/paymentMethods.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param {String} shopId - shop ID for which to get payment methods
+ * @return {Object[]} TODO: Groups collection cursor
+ */
+export default async function (context, shopId) {
+  const shop = await context.queries.shopById(context, shopId);
+
+  return Object.keys(paymentMethods)
+    .map((name) => ({
+      ...paymentMethods[name],
+      isEnabled: shop.availablePaymentMethods.includes(name)
```
aldeed: Change to `isEnabled: (shop.availablePaymentMethods || []).includes(name)`. Otherwise I get errors running the query until we make the migration to set this array.

imports/plugins/core/payments/server/no-meteor/queries/paymentMethods.js
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * @name paymentMethods
+ * @method
+ * @memberof Payments/NoMeteorQueries
+ * @summary TODO: query the Groups collection and return a MongoDB cursor
+ * @param {Object} context - an object containing the per-request state
+ * @param {String} shopId - shop ID for which to get payment methods
+ * @return {Object[]} TODO: Groups collection cursor
+ */
+export default async function (context, shopId) {
```
aldeed: All functions should be named, with the same name as the file name.

imports/plugins/core/payments/server/no-meteor/queries/paymentMethods.js
```diff
@@ -0,0 +1,20 @@
+import { paymentMethods } from "/imports/plugins/core/core/server/no-meteor/pluginRegistration";
+
+/**
+ * @name paymentMethods
+ * @method
+ * @memberof Payments/NoMeteorQueries
+ * @summary TODO: query the Groups collection and return a MongoDB cursor
+ * @param {Object} context - an object containing the per-request state
+ * @param {String} shopId - shop ID for which to get payment methods
+ * @return {Object[]} TODO: Groups collection cursor
```
aldeed: TODOs

imports/plugins/core/payments/server/no-meteor/queries/availablePaymentMethods.js
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * @name availablePaymentMethods
+ * @method
+ * @memberof Payments/NoMeteorQueries
+ * @summary TODO: query the Groups collection and return a MongoDB cursor
+ * @param {Object} context - an object containing the per-request state
+ * @param {String} shopId - shop ID for which to get payment methods
+ * @return {Object[]} TODO: Groups collection cursor
+ */
+export default async function (context, shopId) {
```
aldeed: All functions should be named, with the same name as the file name.

imports/plugins/core/payments/server/no-meteor/queries/availablePaymentMethods.js
```diff
@@ -0,0 +1,18 @@
+import { paymentMethods } from "/imports/plugins/core/core/server/no-meteor/pluginRegistration";
+
+/**
+ * @name availablePaymentMethods
+ * @method
+ * @memberof Payments/NoMeteorQueries
+ * @summary TODO: query the Groups collection and return a MongoDB cursor
+ * @param {Object} context - an object containing the per-request state
+ * @param {String} shopId - shop ID for which to get payment methods
+ * @return {Object[]} TODO: Groups collection cursor
```
aldeed: TODOs

imports/plugins/core/payments/server/no-meteor/queries/availablePaymentMethods.js
```diff
@@ -0,0 +1,18 @@
...
+ */
+export default async function (context, shopId) {
+  const shop = await context.queries.shopById(context, shopId);
+  const paymentMethodNames = Object.keys(paymentMethods);
+
+  return shop.availablePaymentMethods
+    .map((name) => paymentMethodNames.find((paymentMethodName) => paymentMethodName === name));
```
aldeed: When I set `shop.availablePaymentMethods` manually in the DB, this does not actually work properly. It should be returning the full methods. By changing to this it seems to work:

```js
return (shop.availablePaymentMethods || []).reduce((list, name) => {
    const paymentMethod = paymentMethods[name];
    if (paymentMethod) {
      list.push({
        ...paymentMethod,
        isEnabled: true
      });
    }
    return list;
  }, []);
```

rosshadden:
Great idea, I missed testing that particular branch of logic. Done! Any more you can think of? I just looked at it for a bit and think we're good to go.

PR #4724, opened 2018-10-10
docs: Link readers to Reaction Platform install instructions

machikoyasuda:
Resolves https://github.com/reactioncommerce/reaction-docs/issues/687
Impact: **minor**  
Type: **docs|chore**

## Issue
- We want all Reaction Platform install docs to live in one, official, canonical spot.
- That spot is here: https://docs.reactioncommerce.com/docs/installation-reaction-platform

## Solution
- I created a 2.0 section updated the README to link to https://docs.reactioncommerce.com/docs/installation-reaction-platform
- I kept a pre-2.0 section with links to the old instructions for OS X, Linux and Windows

## Breaking changes
None

## Testing
1. Test that the link works and docs look good: https://github.com/reactioncommerce/reaction-next-starterkit/blob/6418fa744dd3d6e111a8fc1b4dfb4cf50121c552/README.md
2. Test that the docs don't leave out anything necessary for starting starterkit

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,10 @@ We love your pull requests! Check our our [`Good First Issue`](https://github.co
 
 Pull requests should:
 
-- Pass all pull request Circle CI checks: Run `npm run lint` and `reaction test` to make sure you're following the [Reaction Code Style Guide](https://docs.reactioncommerce.com/reaction-docs/master/styleguide) and passing [acceptance tests and unit tests](https://docs.reactioncommerce.com/reaction-docs/master/testing-reaction).
+- Pass all Circle CI checks: 
+    - Run `docker-compose run --rm reaction npm run lint` to make sure your code follows [Reaction's ESLint rules](https://github.com/reactioncommerce/reaction-eslint-config).
+    - Run `docker-compose run --rm reaction reaction test` to run [acceptance tests and unit tests](https://docs.reactioncommerce.com/reaction-docs/master/testing-reaction).
+    - Make sure you're following the [Reaction Code Style Guide](https://docs.reactioncommerce.com/reaction-docs/master/styleguide) and 
 - Follow the pull request template.
 
 Get more details in our [Contributing Guide](https://docs.reactioncommerce.com/reaction-docs/master/contributing-to-reaction).
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ Follow the documentation for installing pre-2.0 Reaction on [OS X](https://docs.
 ## Read documentation & tutorials
 
 -   [Reaction Commerce: Developer documentation](https://docs.reactioncommerce.com)
+-   [Reaction Design System](http://designsystem.reactioncommerce.com/)
 -   [Reaction Commerce: API documentation](http://api.docs.reactioncommerce.com)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,52 +25,20 @@ Since anything in our codebase can be extended, overwritten, or installed as a p
...
+## Reaction 2.0
 
-Open `localhost:3000`
+Follow the documentation for [Reaction Platform](https://docs.reactioncommerce.com/docs/installation-reaction-platform) for all operating systems.
 
-This will use the `docker-compose.yml` file.
+## Pre-2.0 Reaction
 
-To learn more on how to develop on Docker, read our documentation on [developing Reaction on Docker](https://docs.reactioncommerce.com/docs/next/installation-docker-development) and [troubleshooting Docker](https://docs.reactioncommerce.com/docs/next/troubleshooting-development#docker-issues).
+Follow the documentation for installing pre-2.0 Reaction on [OS X](https://docs.reactioncommerce.com/docs/next/installation-osx), [Windows](https://docs.reactioncommerce.com/docs/next/installation-windows) or [Linux](https://docs.reactioncommerce.com/docs/next/installation-linux).
```
machikoyasuda: @ticean @spencern I wasn't sure whether to entirely remove the pre-2.0 instructions or not... so I made a section like this.

Should I remove this section?

spencern: I'm not opposed to keeping it somewhere - would it make sense to link to a pre-2.0 version of the docs as well then instead of `next`?

spencern: Maybe something like "Looking for docs to install pre-2.0 Reaction?"

README.md
```diff
@@ -26,52 +25,20 @@ Since anything in our codebase can be extended, overwritten, or installed as a p
...
-
-### Install and run with Docker
-
-You can also run the app locally using [`docker-compose`](https://docs.docker.com/compose/) by running:
-
-```sh
-docker network create api.reaction.localhost
-docker-compose up
-```
+## Reaction 2.0
```
spencern: Thinking now that we should eliminate the `2.0 and `pre-2.0` distinctions and just direct to the platform. We can also say something like "Looking for installation docs for an older version? https://docs.reactioncommerce.com/versions" or similar?

README.md
```diff
@@ -92,7 +59,7 @@ We love your pull requests! Check our our [`Good First Issue`](https://github.co
 
 Pull requests should:
 
-- Pass all pull request Cirlce CI checks: Run `npm run lint` and `reaction test` to make sure you're following the [Reaction Code Style Guide](https://docs.reactioncommerce.com/reaction-docs/master/styleguide) and passing [acceptance tests and unit tests](https://docs.reactioncommerce.com/reaction-docs/master/testing-reaction).
+- Pass all pull request Circle CI checks: Run `npm run lint` and `reaction test` to make sure you're following the [Reaction Code Style Guide](https://docs.reactioncommerce.com/reaction-docs/master/styleguide) and passing [acceptance tests and unit tests](https://docs.reactioncommerce.com/reaction-docs/master/testing-reaction).
```
spencern: Are these commands still accurate within the dockerized development environment?

PR #4739, opened 2018-10-15
Feat 4718 enable payment method for shop

### LINKED ISSUE #4718: Create enablePaymentMethodForShop mutation
## Work
Create a mutation that allows an admin to enable/disable a payment method for a shop. Create it in the "payments" core plugin.

### GraphQL schema
```graphql
"Input for the `enablePaymentMethodForShop` mutation"
input EnablePaymentMethodForShopInput {
  "An optional string identifying the mutation call, which will be returned in the response payload"
  clientMutationId: String

  "True to enable it or false to disable it"
  isEnabled: Boolean!

  "The name of the payment method to enable or disable"
  paymentMethodName: String!

  "The ID of the shop for which this payment method should be enabled or disabled"
  shopId: ID!
}

type EnablePaymentMethodForShopPayload {
  "The same string you sent with the mutation params, for matching mutation calls with their responses"
  clientMutationId: String

  "The full list of payment methods for the shop"
  paymentMethods: [PaymentMethod]!
}

extend type Mutation {
  enablePaymentMethodForShop(input: EnablePaymentMethodForShopInput!): EnablePaymentMethodForShopPayload!
}
```

rosshadden:
Resolves #4718   
Impact: **minor**  
Type: **feature**

## Description
Created a mutation that allows an admin to enable/disable a payment method for a shop.

> NOTE: I don't like that the mutation itself is testing for validity of payment methods when modifying them. I now think we should make `paymentMethods` a data structure with add, delete, list, get, and some other methods, instead of the basic `{}` it currently is. But I would like to hear thoughts on this.

## Breaking changes
None.

## Testing
Try queries similar to the following:

```graphql
mutation {
  enablePaymentMethodForShop(
    input: {
      clientMutationId: "foo"
      shopId: "<shopOpaqueId>"
      paymentMethodName: "<paymentMethodName>"
      isEnabled: true # or false
    }
  ) {
    clientMutationId
    paymentMethods {
      name
      displayName
      pluginName
      isEnabled
    }
  }
}
```

1. Try a payment method that exists (`iou_example`)
2. Try a payment method that does not exist (`foo`), and make sure it throws an error
3. Try with and without a `clientMutationId`

### DIFF
--- a/imports/plugins/core/payments/server/no-meteor/mutations/enablePaymentMethodForShop.js
+++ b/imports/plugins/core/payments/server/no-meteor/mutations/enablePaymentMethodForShop.js
@@ -4,21 +4,21 @@ import { Shops } from "/lib/collections";
 import { paymentMethods as allPaymentMethods } from "/imports/plugins/core/core/server/no-meteor/pluginRegistration";
 
 const paramsSchema = new SimpleSchema({
-  shopId: String,
+  isEnabled: Boolean,
   paymentMethodName: String,
-  isEnabled: Boolean
+  shopId: String
 });
 
 export default async function enablePaymentMethodForShop(context, input = {}) {
   paramsSchema.validate(input);
-  const { shopId, paymentMethodName, isEnabled } = input;
+  const { isEnabled, paymentMethodName, shopId } = input;
 
   if (!allPaymentMethods[paymentMethodName]) {
     throw new ReactionError("not-found", "Requested payment method is invalid");
   }
 
   const shop = await context.queries.shopById(context, shopId);
-  const methods = new Set(shop.availablePaymentMethods)
+  const methods = new Set(shop.availablePaymentMethods);
 
   if (isEnabled) {
     methods.add(paymentMethodName);
@@ -32,8 +32,5 @@ export default async function enablePaymentMethodForShop(context, input = {}) {
     { bypassCollection2: true }
   );
 
-  const paymentMethods = await context.queries.paymentMethods(context, shopId);
-  return {
-    paymentMethods
-  };
+  return context.queries.paymentMethods(context, shopId);
 }
--- a/imports/plugins/core/payments/server/no-meteor/mutations/enablePaymentMethodForShop.js
+++ b/imports/plugins/core/payments/server/no-meteor/mutations/enablePaymentMethodForShop.js
@@ -25,6 +25,10 @@ export default async function enablePaymentMethodForShop(context, input = {}) {
   const { Shops } = context.collections;
   const { isEnabled, paymentMethodName, shopId } = input;
 
+  if (!context.userHasPermission(["owner", "admin"], shopId)) {
+    throw new ReactionError("access-denied", "Access denied");
+  }
+
   if (!allPaymentMethods[paymentMethodName]) {
     throw new ReactionError("not-found", "Requested payment method is invalid");
   }
--- a/imports/plugins/core/payments/server/no-meteor/mutations/enablePaymentMethodForShop.js
+++ b/imports/plugins/core/payments/server/no-meteor/mutations/enablePaymentMethodForShop.js
@@ -1,6 +1,5 @@
 import ReactionError from "@reactioncommerce/reaction-error";
 import SimpleSchema from "simpl-schema";
-import { Shops } from "/lib/collections";
 import { paymentMethods as allPaymentMethods } from "/imports/plugins/core/core/server/no-meteor/pluginRegistration";
 
 const paramsSchema = new SimpleSchema({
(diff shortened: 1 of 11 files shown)


rosshadden:
This is WIP because I need to add tests and documentation. But I made the PR in case anyone wanted to look at it now.

imports/plugins/core/payments/server/no-meteor/mutations/enablePaymentMethodForShop.js
```diff
@@ -0,0 +1,36 @@
+import ReactionError from "@reactioncommerce/reaction-error";
+import SimpleSchema from "simpl-schema";
+import { Shops } from "/lib/collections";
```
aldeed: In no-meteor code we use the collections off `context` rather than importing them. The ones in /lib/collections are Meteor collections while the ones on context are straight from the `mongodb` NPM package.

```js
const { Shops } = context.collections;
```

docs ref: https://docs.reactioncommerce.com/docs/graphql-developing#using-mongodb-in-graphql

imports/plugins/core/payments/server/no-meteor/mutations/enablePaymentMethodForShop.js
```diff
@@ -0,0 +1,36 @@
...
+    methods.add(paymentMethodName);
+  } else {
+    methods.delete(paymentMethodName);
+  }
+
+  await Shops.update(
+    { _id: shop._id },
+    { $set: { availablePaymentMethods: Array.from(methods) } },
+    { bypassCollection2: true }
+  );
```
aldeed: `bypassCollection2` is only needed for Meteor collections, and generally only should be used in migration code. That bypasses checking the modifier against the schema that is attached to the collection.

Also, in the version of mongo driver we use, `update` is deprecated. So use `updateOne` instead.

imports/plugins/core/payments/server/no-meteor/mutations/enablePaymentMethodForShop.js
```diff
@@ -0,0 +1,36 @@
...
+
+export default async function enablePaymentMethodForShop(context, input = {}) {
+  paramsSchema.validate(input);
+  const { isEnabled, paymentMethodName, shopId } = input;
+
+  if (!allPaymentMethods[paymentMethodName]) {
+    throw new ReactionError("not-found", "Requested payment method is invalid");
+  }
+
+  const shop = await context.queries.shopById(context, shopId);
```
aldeed: `shop` might be undefined. If so, need to throw a not-found for that same as above.

imports/plugins/core/payments/server/no-meteor/resolvers/Mutation/enablePaymentMethodForShop.js
```diff
@@ -0,0 +1,14 @@
+import { decodeShopOpaqueId } from "@reactioncommerce/reaction-graphql-xforms/shop";
+
+export default async function enablePaymentMethodForShop(parentResult, { input }, context) {
+  const { clientMutationId } = input;
+  input.shopId = decodeShopOpaqueId(input.shopId);
+  delete input.clientMutationId;
```
aldeed: I generally try to avoid mutating an argument unless it's a function designed to mutate. This could have unintended consequences. Can you make a new input object instead?

imports/plugins/core/payments/server/no-meteor/mutations/enablePaymentMethodForShop.js
```diff
@@ -0,0 +1,49 @@
+import ReactionError from "@reactioncommerce/reaction-error";
+import SimpleSchema from "simpl-schema";
+import { paymentMethods as allPaymentMethods } from "/imports/plugins/core/core/server/no-meteor/pluginRegistration";
+
+const paramsSchema = new SimpleSchema({
+  isEnabled: Boolean,
+  paymentMethodName: String,
+  shopId: String,
+  clientMutationId: { type: String, optional: true }
+});
```
aldeed: It's mixing concerns to have `clientMutationId` (which only GraphQL cares about) in here. You could either explicitly build the whole object argument without this property, or you can tell validation to ignore extra properties:

```js
paramsSchema.validate(input, { ignore: [SimpleSchema.ErrorTypes.KEY_NOT_IN_SCHEMA] });
```

rosshadden: I went back and forth on this. The reason I landed where I did with having it passed is because `clientMutationId` is in the GQL schema itself. So I think your second suggestion is best: ignoring the field rather than defining it in the validation schema.

PR #4740, opened 2018-10-15
feat: decouple SMTP email sending logic from core to allow plugins to override

### LINKED ISSUE #4730: Use appEvents for email sending, so that plugins can integrate with non SMTP email providers
## Prerequisites
* [x] Are you running the latest version? 2.0 RC5

## Issue Description
Currently, Reaction only supports sending email through an SMTP account. This logic is hardcoded in `imports/plugins/core/core/server/Reaction/processEmailJobs.js`. There is no way to plug in to the email job handling and use a third-party email provider that doesn't use SMTP.

## Possible Solution
Use the new `appEvents` to allow other plugins to process email jobs and send with any email service. Example: `appEvents.emit("processEmailJobs")` from `processEmailJobs.js`, and `appEvents.on("processEmailJobs", (jobs, callback) => CUSTOM_LOGIC_HERE)`.

dancastellon:
Resolves #4730 
Impact: **minor**  
Type: **feature**

## Issue
Currently, SMTP email sending logic is hard-coded in core. This makes it impossible to integrate with a third-party email provider that does not support SMTP, without modifying core code.

## Solution
This PR decouples SMTP from core by creating a new `reaction-email-smtp` included plugin. Core now emits a `sendEmail` event for each email job, which any plugin can respond to. The included SMTP plugin responds to it, checks if an SMTP email provider is configured, and sends the email if so.

The email provider config form found at Dashboard -> Emails -> Mail Provider is now also able to be overridden. Plugins can use register.js to provide a React component to use here.

## Breaking changes
None

## Testing
1. Login as operator -> Dashboard -> Email -> Mail Provider. Confirm the existing config component works as expected (shows status icon & existing settings, gear icon shows config form, you can successfully configure an SMTP provider, etc)
2. After configuring an SMTP provider, complete an order and confirm the email was successfully delivered
3. Confirm in database that Emails.status is set to "completed"
4. Set SMTP settings to something invalid & save
5. Complete another order, confirm email error in logs, and that Emails.status is set to "failed" in database
6. Create an `email-provider-test` plugin in `imports/plugins/custom` with 3 files:
- register.js
```

Reaction.registerPackage({
  label: "Email Provider Test",
  name: "reaction-email-provider-test",
  icon: "fa fa-envelope-o",
  autoEnable: true,
  registry: [
    {
      provides: ["emailProviderConfig"],
      template: "TestEmailConfig"
    }
  ]
});
```
- client/TestEmailConfig.js:
```

class TestEmailConfig extends Component {
  render() {
    return (
      <div>Test email provider config here</div>
    );
  }
}

registerComponent("TestEmailConfig", TestEmailConfig);
```
- client/index.js:
```
import "./TestEmailConfig";
```
7. Restart Reaction and confirm you see "Test email provider config here" at Dashboard -> Email -> Mail Provider


imports/plugins/included/email-smtp/server/startup.js
```diff
@@ -0,0 +1,4 @@
+import appEvents from "/imports/node-app/core/util/appEvents";
+import sendSMTPEmail from "./util/sendSMTPEmail";
+
+appEvents.on("sendEmail", sendSMTPEmail);
```
aldeed: @dancastellon Check out `/imports/plugins/core/cart/server/no-meteor/startup.js`. This is the preferred way to do a startup function and call `.on` in there. Then you do

```js
functionsByType: {
    startup: [startup]
  },
```

As seen in `/imports/plugins/core/cart/register.js`

The way you're doing it here does work, but by registering it as a plugin startup function, you can be sure that it doesn't happen until AFTER all of the core startup code is loaded.

dancastellon: @aldeed All set here

PR #4742, opened 2018-10-16
fix: 4741 catalog variant inventory flags always false

### LINKED ISSUE #4741: Catalog variant `isSoldOut` and `isLowQuantity` are always false
## Issue Description

The denormalized inventory flags for `isSoldOut` and `isLowInventory` are always `false` for variants in the catalog.

## Steps to Reproduce
1. Go to a product
2. Set variant to the quantity to 0
3. Turn off backorders
4. Publish
5. Look in mongo for that Catalog Variant
6. See that `isSoldOut` is `false` even though it should be true

## Possible Solution

In `createCatalogProduct.js` function `xformVariant(variant, variantPriceInfo, shopCurrencyCode, variantMedia)` always receives `undefined` for variants.

`xformVariant` will have to do some additional processing to determine the correct inventory flags and not assume they come from the params.

### Versions

Node: 10.7.0
NPM: 6.2.0
Meteor Node: 8.11.4
Meteor NPM: 6.4.1
Reaction CLI: 0.29.0
Reaction: 2.0.0-rc.5
Reaction branch: fix-4722-mikemurray-compare-at-price
Docker: 18.06.1-ce

mikemurray:
Resolves #4741  
Impact: **major**  
Type: **bugfix**

## Issue

`isSoldOut` and `isLowQuantity` for variants and options in the catalog were always `false`, and not reflective of their actual status.

## Solution

Fix the transform that creates the catalog variants by processing the correct inventory flags just like it's done for the top-level product

## Breaking changes

Added a new, final param to `xformVariant` with the processed inventory flags

## Testing

Try to set the quantities and warn at thresholds to various variants and options and check the DB or GQL for the expected results.

For example:
- If you set a variant quantity to 0, it should be sold out if it has no options
- If you set quantities of the options of a  variant  to 0, each option should be sold out as well as the top-level variant

**Play around with different combinations of sold out, low inventory and back ordered**

**Below are some variations**

---

1. With a product that only has top-level variants
2. Set the quantity to 0 and publish
3. Query in GQL or Mongo for that catalog item variant
4. See that `isSoldOut` is `true`

---

1. With a product that only has top level variants
2. Set the low quantity threshold to 1000 and publish
3. Query in GQL or Mongo for that catalog item variant
4. See that `isLowQuantity` is `true`

---

1. With a product that has variant options
2. Set each of the options quantity to 0 and publish
3. Query in GQL or Mongo for that catalog item variant that contains those options
4. See that `isSoldOut` is `true`
5. See that `isSoldOut` is `true` for the option itself

---

1. With a product that has variant options
2. Set the low quantity threshold to 1000 on the top-level variant and publish
3. Query in GQL or Mongo for that catalog item variant that contains those options
4. See that `isLowQuantity` is `true` for the option itself

---

1. Checkout with a product that has a quantity of 1
1. Approving or canceling the order in the order panel should update inventory status accordingly

---

1. Checkout with a product that has a quantity of 11 and a `warn at` at 10
1. Approving or canceling the order in the order panel should update inventory status accordingly

---

**Test the Migrations**

1. Start reaction with this branch
2. `mongorestore --drop` the `reaction-2.0.0-rc.1` mongo dump
3. Run a graph QL query for a product that has legacy variant inventory flags, or check the catalog collection.

```
{
  catalogItemProduct(slugOrId: "converse-all-stars-black") {
    product {
      variants {
        title
        isSoldOut
        isLowQuantity
        inventoryPolicy
        inventoryManagement
        options {
          title
          isSoldOut
          isLowQuantity
        }
      }
    }
  }
}
```
4. See the variant and variant options `isSoldOut` are `false`. This is incorrect. 
5. Stop and start reaction again
6. Wait for migration 42 to run
7. Re-run the query and see the variant and variant options `isSoldOut` are now `true`


aldeed:
@mikemurray A bigger picture question. Why store these in the catalog at all given how dynamic they are and the fact that the source-of-truth inventory values are stored on Product and are not published? Can we just look them up in the `CatalogProduct.variants` resolver? (`/imports/plugins/core/catalog/server/no-meteor/resolvers/CatalogProduct/index.js`)

mikemurray:
@aldeed I agree it feels weird for the inventory values to be published to the catalog, but the only time they are/should be published is when there's a change. For example, a product is only sold out when it reaches 0 on all variants/options.

Making a resolver is a good idea, but we'll have to make a request to the products collection on every request for every product, which defeats the purpose of the catalog collection being a sort of cache for product data.

mikemurray:
The function `updateCatalogProductInventoryStatus` in `/imports/plugins/core/catalog/server/no-meteor/utils/updateCatalogProductInventoryStatus.js` needs to be updated to also include the variant and variant options when updating the inventory status.

mikemurray:
@nnnnat Pushed some changes `updateCatalogProductInventoryStatus` so it now also updates the variants and options of the catalog item.

Additional testing would include:
- Approving or canceling an order in the order panel should update inventory status properly

Should be ready for your review.

nnnnat:
@mikemurray I'm running into an issue with the migration and have been unable to get Reaction running to test.

```sh
reaction_1  | => Exited with code: 0
reaction_1  | 22:39:46.295Z  INFO Reaction: Load default data from /private/data/
reaction_1  | 22:39:46.304Z  INFO Reaction: JobServer started
reaction_1  | 22:39:47.328Z  WARN Reaction: Skipped loading settings from reaction.json.
reaction_1  | 22:39:49.199Z  INFO Reaction: Reaction Version: 2.0.0-rc.6
reaction_1  | 22:39:49.289Z FATAL Reaction: You are running a Reaction install with migration version (44) below your current DB migration state (45)
reaction_1  | 22:39:49.289Z FATAL Reaction: Upgrade to a version of Reaction containing migration 45 or higher.
reaction_1  | 22:39:49.289Z FATAL Reaction: If you really want to downgrade to this version, you should restore your DB to a previous state from your backup.

```

aldeed:
> I believe one thing that may need addressing is the customer PDP is still using Product collection and not Catalog

@nnnnat There is code in the server publication function that publishes `Catalog` collection to `Product` client collection, so it's actually catalog data even though it appears to be product data. This is not how it should be long term, but I guess was the quickest solution at the time. If we update  the grid to use GraphQL only, that will be the best proper solution.

This PR has some failing tests that will need to be fixed before it can be merged.

mikemurray:
@nnnnat All tests are passing now.

PR #4743, opened 2018-10-16
feat: Add a CORS-enabled endpoint for token refresh in Hydra plugin

impactmass:
Part of reactioncommerce/reaction-next-starterkit#350
Type: **feature**

## Issue
To help enable refresh token in Starterkit (reactioncommerce/reaction-next-starterkit#350)

## Changes
- Adds CORS-enabled endpoint to be called from Starterkit for token refresh

## Breaking changes
n/a

## Testing
1. Test steps in reactioncommerce/reaction-next-starterkit#394

### DIFF
--- a/imports/plugins/core/hydra-oauth/server/oauthEndpoints.js
+++ b/imports/plugins/core/hydra-oauth/server/oauthEndpoints.js
@@ -68,17 +68,26 @@ WebApp.connectHandlers.use("/consent", (req, res) => {
     .catch((errorMessage) => errorHandler(errorMessage, res));
 });
 
-WebApp.connectHandlers.use("/token/refresh", (req, res) => {
-  res.setHeader("Access-Control-Allow-Origin", process.env.OAUTH2_CLIENT_DOMAIN);
+WebApp.connectHandlers.use("/token/refresh", async (req, res) => {
+  const { origin } = req.headers;
+  const whitelist = process.env.OAUTH2_CLIENT_DOMAINS || "";
+  const whitelistArr = whitelist.split(",");
 
-  hydra
-    .refreshAuthToken(req.query)
-    .then((apiRes) => {
-      Logger.debug(`Refresh auth token call successful: ${apiRes.statusCode}`);
-      res.writeHead(200, { "Content-Type": "application/json" });
-      return res.end(JSON.stringify(apiRes));
-    })
-    .catch((errorMessage) => errorHandler(errorMessage, res));
+  if (whitelistArr.indexOf(origin) > -1) {
+    res.setHeader("Access-Control-Allow-Origin", origin);
+    res.setHeader("Vary", "Origin");
+  }
+
+  const apiRes = await hydra.refreshAuthToken(req.query);
+
+  if (apiRes.status_code < 200 || apiRes.status_code > 302) {
+    Logger.error("An error occurred while calling refresh API", apiRes.error_description);
+    return errorHandler(apiRes.error_description, res);
+  }
+
+  Logger.debug(`Refresh auth token call successful: ${apiRes.statusCode}`);
+  res.writeHead(200, { "Content-Type": "application/json" });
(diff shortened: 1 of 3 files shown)


imports/plugins/core/hydra-oauth/server/oauthEndpoints.js
```diff
@@ -68,6 +68,19 @@ WebApp.connectHandlers.use("/consent", (req, res) => {
     .catch((errorMessage) => errorHandler(errorMessage, res));
 });
 
+WebApp.connectHandlers.use("/token/refresh", (req, res) => {
+  res.setHeader("Access-Control-Allow-Origin", process.env.OAUTH2_CLIENT_DOMAIN);
```
aldeed: According to this, you need the `Vary: Origin` header, too: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#CORS_and_caching

There would likely be many different domains hitting this. I believe the way this is usually done is setting the value to `req.headers.origin` after verifying that that origin is allowed.

Maybe change ENV name to DOMAINS plural, split it to an array on commas, and then check whether the request origin header is in that array?

imports/plugins/core/hydra-oauth/server/util/hydra.js
```diff
@@ -78,12 +78,36 @@ function deleteUserSession(id) {
...
+  return fetch(`${HYDRA_TOKEN_URL}`, {
+    headers: { "Content-Type": "application/x-www-form-urlencoded" },
+    method: "POST",
+    body: `grant_type=refresh_token&refresh_token=${refreshToken}&response_type=token&client_id=${clientId}&client_secret=${clientSecret}`
+  })
+    .then(async (res) => {
+      if (res.status < 200 || res.status > 302) {
+        const json = await res.json();
+        Logger.error("An error occurred while calling refresh API", json.error_description);
+        return Promise.reject(new Error(json.error_description));
```
aldeed: This is an odd mix of normal promise and async/await. It seems like you could rewrite to be completely async/await:

```js
async function refreshAuthToken({ refreshToken, clientId, clientSecret }) {
  const res = await fetch(`${HYDRA_TOKEN_URL}`, {
    headers: { "Content-Type": "application/x-www-form-urlencoded" },
    method: "POST",
    body: `grant_type=refresh_token&refresh_token=${refreshToken}&response_type=token&client_id=${clientId}&client_secret=${clientSecret}`
  });

  const json = await res.json();

  if (res.status < 200 || res.status > 302) {
    Logger.error("An error occurred while calling refresh API", json.error_description);
    throw new Error(json.error_description);
  }

  return json;
}
```

imports/plugins/core/hydra-oauth/server/util/hydra.js
```diff
@@ -78,12 +78,36 @@ function deleteUserSession(id) {
...
 export default {
   getLoginRequest: (challenge) => get("login", challenge),
   acceptLoginRequest: (challenge, body) => put("login", "accept", challenge, body),
   rejectLoginRequest: (challenge) => put("login", "reject", challenge),
   getConsentRequest: (challenge) => get("consent", challenge),
   acceptConsentRequest: (challenge, body) => put("consent", "accept", challenge, body),
   rejectConsentRequest: (challenge, body) => put("consent", "reject", challenge, body),
-  deleteUserSession: (id) => deleteUserSession(id)
+  deleteUserSession: (id) => deleteUserSession(id),
+  refreshAuthToken: (options) => refreshAuthToken(options)
```
aldeed: For these last two, they are not adding any arguments and are named the same, so you can eliminate a function layer by instead doing:

```js
// ...
  deleteUserSession,
  refreshAuthToken
}
```

aldeed:
@impactmass I went ahead and merged this into RC6 to keep things moving. If changes are necessary based on starter kit decisions, let's do a new PR for those.

PR #4744, opened 2018-10-16
chore: add .nvmrc configuration file

focusaurus:
Useful for getting local node stuff like eslint running.

Impact: **minor**  
Type: **chore**

## Issue

Even though most work happens in docker, getting the right version of node available directly in the host OS is handy for things like eslint integration with your editor. I put an `.nvmrc` file in all my node projects **but** I'm totally fine if this gets closed as not useful (I can just do a local git exclude - no problem).

There is some maintenance burden keeping this file in sync with the `Dockerfile`.

Anyway, if others find this useful, merge away, but if not just close this PR.


aldeed:
I think it's a good idea for the reasons mentioned

PR #4747, opened 2018-10-16
feat: GraphQL query & resolver for loading the primary shop

### LINKED ISSUE #4727: Create Query.primaryShop GraphQL resolver
We have `Query.primaryShopId` and `Query.shopById`. A client currently needs two requests to those to get the primary shop data, which is the most common case and happens all the time on first load. We can simplify this for client development by creating `Query.primaryShop`, which combines those two lookups on the server (and reduces to a single DB query).

dancastellon:
Resolves #4727   
Impact: **minor**  
Type: **feature**

## Solution
This PR adds a `primaryShop` GraphQL query & resolver, eliminating the need to first query for the primary shop ID, followed by another query for shop by ID.

## Breaking changes
None

## Testing
1. Open Graphiql and run this query:
```
{
  primaryShop {
    _id
    name
  }
}
```

### DIFF
--- a/imports/plugins/core/core/server/no-meteor/resolvers/Query/primaryShop.js
+++ b/imports/plugins/core/core/server/no-meteor/resolvers/Query/primaryShop.js
@@ -1,0 +1,15 @@
+/**
+ * @name "Query.primaryShop"
+ * @method
+ * @memberof Shop/GraphQL
+ * @summary Gets the primary shop
+ * @param {Object} parentObject - unused
+ * @param {Object} args - unused
+ * @param {Object} context - an object containing the per-request state
+ * @return {Promise<o>} The shop, based on the domain in ROOT_URL
+ */
+export default async function primaryShopId(_, __, context) {
+  const shopId = await context.queries.primaryShopId(context.collections);
+  const shop = await context.queries.shopById(context, shopId);
+  return shop;
+}
--- a/imports/plugins/core/core/server/no-meteor/resolvers/Query/primaryShop.js
+++ b/imports/plugins/core/core/server/no-meteor/resolvers/Query/primaryShop.js
@@ -6,10 +6,9 @@
  * @param {Object} parentObject - unused
  * @param {Object} args - unused
  * @param {Object} context - an object containing the per-request state
- * @return {Promise<o>} The shop, based on the domain in ROOT_URL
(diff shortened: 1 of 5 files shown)


imports/plugins/core/core/server/no-meteor/resolvers/Query/primaryShop.js
```diff
@@ -0,0 +1,15 @@
...
+ * @summary Gets the primary shop
+ * @param {Object} parentObject - unused
+ * @param {Object} args - unused
+ * @param {Object} context - an object containing the per-request state
+ * @return {Promise<o>} The shop, based on the domain in ROOT_URL
+ */
+export default async function primaryShopId(_, __, context) {
+  const shopId = await context.queries.primaryShopId(context.collections);
+  const shop = await context.queries.shopById(context, shopId);
+  return shop;
```
aldeed: This is nice and simple, but it does two `Shops.findOne` calls when we often could do just one. You can make a separate `primaryShop` query function that does:

```js
const domain = url.parse(context.rootUrl).hostname;
let shop = await Shops.findOne({ domains: domain });
if (!shop) {
  shop = await Shops.findOne({ shopType: "primary" });
}
return shop;
```

And then call that in this resolver

imports/plugins/core/core/server/no-meteor/resolvers/Query/primaryShop.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @name "Query.primaryShop"
+ * @method
+ * @memberof Shop/GraphQL
+ * @summary Gets the primary shop
+ * @param {Object} parentObject - unused
+ * @param {Object} args - unused
+ * @param {Object} context - an object containing the per-request state
+ * @return {Promise<o>} The shop, based on the domain in ROOT_URL
```
aldeed: Should say `Promise<Object>`

imports/plugins/core/core/server/no-meteor/resolvers/Query/primaryShop.js
```diff
@@ -0,0 +1,15 @@
...
+ */
+export default async function primaryShopId(_, __, context) {
```
aldeed: Function name should be primaryShop

PR #4751, opened 2018-10-17
chore: Skip failing integration tests

focusaurus:
Impact: **minor**  
Type: **chore**

## Issue

Integration tests are failing with a mongo issue.

## Solution

Skip the tests until @aldeed has a chance to investigate. I'll file a separate issue to track the broken tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -202,7 +202,7 @@
   "scripts": {
     "devserver": "ROOT_URL=http://localhost:3030 MONGO_URL=mongodb://localhost:27017/reaction NODE_ENV=reaction-node BABEL_DISABLE_CACHE=1 nodemon ./imports/node-app/devserver/index.js",
     "lint": "eslint .",
-    "test": "npm run test:unit && npm run test:app",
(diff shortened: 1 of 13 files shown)


aldeed:
I already updated the `npm test` command in rc5 so you'll also have to put integration tests back there. At this point, if you rebase this on rc6 branch and fix conflicts, that should do it.

focusaurus:
OK I think this should be good to merge. Now we get through all the tests cleanly just the integration ones get skipped via jest.

spencern:
Looks good to me. Do we have a CI step we can re-enable now that they are skipped?

focusaurus:
I think it's the same`npm test` step, just now the integration tests are part of that, but skipped.

aldeed:
Integration tests have never run on CI because of a different issue. Both @jeffcorpuz and I tried getting them to work on CI but gave up. At some point we may have to re-evaluate the way the in-memory MongoDB is used.

Issue #4706, opened 2018-10-03
Use .env file


aldeed:
Most other projects in the Reaction ecosystem now use a `.env` file for ENV variables that need to be set. They include a committed `.env.example` file, which is copied to a gitignored `.env` before building.

## Work
1. Update the installation instructions in the public docs to document how environment variables work in the API project.
1. Create a `.env.example` file and `.env` file. Add all variables that are currently in the `environment` section of `docker-compose.yml`
1. Add `.env` to `.gitignore`
1. Remove the `environment` section from `docker-compose.yml` and replace with:
    ```
    env_file:
      - ./.env
    ```
1. Add pre-build project hook that creates/updates the `.env` file, the same as [the script in the other repos](https://github.com/reactioncommerce/reaction-next-starterkit/blob/develop/bin/setup). Note, however, that it cannot be in `bin/setup` because of Meteor's eager loading. It should be in `.reaction/scripts`.

PR #4758, opened 2018-10-18
Deploy release branches to staging ECS environment

griggheo:
Type: **feature**

## Issue
We want to be able to test code in a release branch in a staging environment.

## Solution
This PR uses the latest version of the propel CLI tool and a single propel.yaml manifest file to deploy the Docker image built in CircleCI to a staging environment in ECS. The data set used for the Reaction service is the 'dev' dataset hosted on a MongoDB server running on an EC2 instance in the staging environment.

## Testing
* perform functional testing on https://reaction-core.staging.reactioncommerce.com/


.circleci/bin/ecs-deploy.sh
```diff
@@ -1,64 +1,48 @@
...
+fi
 
-	mkdir -p ~/.aws
-	echo "[default]" > ~/.aws/credentials
-	echo "aws_access_key_id = ${AWS_ACCESS_KEY_ID}" >> ~/.aws/credentials
-	echo "aws_secret_access_key = ${AWS_SECRET_ACCESS_KEY}" >> ~/.aws/credentials
+mkdir -p ~/.aws
+echo "[default]" > ~/.aws/credentials
+echo "aws_access_key_id = ${AWS_ACCESS_KEY_ID}" >> ~/.aws/credentials
+echo "aws_secret_access_key = ${AWS_SECRET_ACCESS_KEY}" >> ~/.aws/credentials
```
aldeed: @griggheo Just curious, why write these to a file when AWS lib typically already looks for env variables with these names? Seems more secure to avoid any credentials in files.

propel.yaml
```diff
@@ -0,0 +1,33 @@
...
+    name: staging-core
+    containers:
+    - name: core
+      cpu: 500
+      memory: 1800
+      port_mappings:
+      - container_port: 3000
+        host_port: 3000
+      image: reactioncommerce/reaction
+      image_tag: release-2.0.0-rc.5
```
aldeed: So this will need to be updated every time we make a new release branch?

.gitignore
```diff
@@ -2,7 +2,7 @@
 .fileStorage/
 .vscode
 .idea
-.env
+.env*
```
aldeed: There's an open issue to add a `.env` file for this repo for local development, and that will involve creating a file named `.env.example` which will need to be committed (same pattern as in other repos). Can you change this ignore rule so that `.env.example` will be committed when we add that?

.circleci/config.yml
```diff
@@ -339,9 +339,9 @@ workflows:
       - deploy-to-ecs:
           requires:
             - docker-push
-#          filters:
-#            branches:
-#              only: /^release-2\.0.\d+$/
+          filters:
+            branches:
+              only: /^release*/
```
aldeed: Were we going to do `/^release-2*/` instead? There is currently a `release-1.17.0` branch and could at some point be `release-1.17.1`. Should branches like that overwrite staging deployment?

cc @spencern

propel.yaml
```diff
@@ -0,0 +1,33 @@
...
+      env_params:
+      - name: REACTION_AUTH
+      - name: REACTION_EMAIL
+      - name: REACTION_USER
+      - name: MONGO_URL
+      - name: ROOT_URL
+      - name: HYDRA_ADMIN_URL
+      - name: HYDRA_OAUTH2_INTROSPECT_URL
+      - name: SKIP_FIXTURES
```
aldeed: Does this list need to change when we add env variables? Where do the values come from? Maybe we can make this more straightforward when we do #4706

spencern:
Seems to work - I'm seeing an issue where prices on the grid are displaying without a currency symbol, but I doubt that is unique to this deployment.

image

propel.yaml
griggheo: The image_tag is used only if you deploy using propel from the command line. CircleCI deployments will automatically use the SHA1 of the newly built Docker image.

.gitignore
griggheo: Will do.

.circleci/config.yml
griggheo: OK I'll change it to /^release-2.*/ for now.

propel.yaml
griggheo: Those are there for informational purposes only. I can remove them once we have the .env.example file.

.circleci/bin/ecs-deploy.sh
griggheo: It didn't work for me when I specified the AWS keys without actually writing them to the credentials file. I'll try again. Agreed it's better not to save them to disk.

griggheo:
@aldeed I think I've addressed all your requests. Please take a look when you have a chance. Thanks!

aldeed:
This LGTM if you approve, @spencern. The app tests failed with a strange error, so I reran them. If they fail again, may need to investigate whether there is something in this PR that is causing it.

PR #4772, opened 2018-10-24
fix: keep toggles shown, width 100% in action view

### LINKED ISSUE #4713: Admin switches disappear at some widths
## Issue Description
Switches disappear at some widths

## Steps to Reproduce
1. Login as admin
1. Go to "Payment"
1. Drag browser width from mobile to larger and observe that the buttons disappear from about 920px to around 1300px

Impact minor as there is a simple workaround (make your browser wider)

### Versions
Reaction: 2.0.0-rc2

aldeed:
Resolves #4713   
Impact: **minor**  
Type: **bugfix**

## Issue
When screen width is > 949px and < 1420px, a style is applied to the action view that causes it to be wider than the width of its container. The right side of fields and the enable/disable toggles are hidden because of this.

## Solution
Removed the style that was causing this.

## Testing
See linked issue. The "Shop" settings is a good place to test.

### DIFF
--- a/imports/plugins/core/dashboard/client/components/actionView.js
+++ b/imports/plugins/core/dashboard/client/components/actionView.js
@@ -93,9 +93,6 @@ const getStyles = (props) => {
       "flexDirection": "column",
       "flex": "1 1 auto",
       "maxWidth": viewMaxSize,
-      "@media only screen and (max-width: 1420px)": {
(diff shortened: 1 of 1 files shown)


aldeed:
@nnnnat It looks like you added this media query about a month ago, so I don't know if you remember why. Seems it has an unintended consequence.

PR #4778, opened 2018-10-24
Deploy feature branches to ECS

griggheo:
Type: **feature**

## Issue
We would like to deploy feature branches to their own new services in ECS, so that changes can be tested by hitting a specific URL.

## Solution
A new service based on the current code is created and deployed to ECS. The service is called reaction-core-FEATURE-BRANCH-NAME.

A service for the storefront based on the reaction-next-starterkit master branch is also created and deployed to ECS. The service is called storefront-FEATURE-BRANCH-NAME.

## Testing
1. Perform functional testing at https://reaction-core-FEATURE-BRANCH-NAME.feat.reactioncommerce.com and https://storefront-FEATURE-BRANCH-NAME.feat.reactioncommerce.com


kieckhafer:
Closing in favor of #4833 (new branches for both features)

PR #4783, opened 2018-10-27
[WIP] feat: shipping method restrictions and surcharges

kieckhafer:
[WIP] [WIP] [WIP]

Impact: **major**  
Type: **feature**

## Issue
Shops should be able to restrict available shipping methods based on shipping location and attributes of items inside an order (price, brand, type of item, etc). They should also be able to apply surcharges to an order based on the same criteria.

## Solution
- Create both universal and per-method ways to allow or deny a shipping method based on the address attached to the an order, and attributes attached to items in an order. This is done by creating a `FlatRateFulfillmentMethodRestrictions` collection, and creating `allow` and `deny` restrictions for each method inside this collection.

- Create both universal and per-method ways to add surcharges to an order based on the address attached to the an order, and attributes attached to items in an order. This is done by creating a `Surcharges` collection, and applying the surcharges to an order if criteria matches.

## Breaking changes
None. These are add-on features. If no Allow or Deny restrictions are found on a method, the method will be allowed in the same way it currently is. If no surcharges are found, nothing is different.

## Testing [WIP]
- Use this mongo dump to load various shipping restrictions and surcharges into the database. ([WIP] link mongodump)
- Test adding products that match the criteria in the restrictions and surcharges to see if certain shipping methods are not allowed, or if surcharges are added to your order.

## Existing Test Updates
- Added new data to `getFulfillmentMethodsWithQuotes` test to allow our new code to work with it


kieckhafer:
Split code up into different branches and into PR's #4821 and #4829

PR #4785, opened 2018-10-31
Improve tax API, split out Custom Rates plugin

### LINKED ISSUE #4687: When multiple tax rates are applicable, only one is applied
## Issue Description
When you have multiple tax rates with different specificity all taxes rates that apply should be applied, however only one rate is being applied. (this is true in cases where the state applies a tax rate and the county and even the city applies additional taxes rates which is quite common in CA)

## Steps to Reproduce
1. Set up a tax rate for CA for 50%
1. Set up a second tax rate for 90405 for 1%
1. Go to checkout and enter an address for state of CA and postal of 90405
1. Observe that only the tax rate for 90405 is applied

## Possible Solution
If there is any de-duping by tax code I noticed that both tax rates get the same "Tax Code" of `RC_TAX`

### Versions
Reaction: 2.0.0-rc1

### LINKED ISSUE #4710: Not able to set tax on cart.
## Issue Description
There has been this change in 2.0 that `taxes/setRateByShopAndItem is replaced by non-Meteor getFulfillmentGroupItemsWithTaxAdded`.
But that function is not setting taxes in the cart. It basically gets taxes from `Taxes` collection and returns them merged with `cart.items`.
We need to have function that would set taxes in the cart, which can be used by tax-plugins.

I am blocked by this while writing a tax-plugin.

### LINKED ISSUE #4711: Updating cart causes multiple tax API calls.
## Issue Description
In the tax-plugins, the function to calculate the tax is run on `afterCartUpdate` event.
This causes the function to run everytime something is updated in the cart, which is quite a lot.
How do I implement something like: run this function only when the cart is being displayed or the user is in the checkout screen.

### LINKED ISSUE #4712: Should `onCartUpdate` be called after updating tax in a plugin
## Issue Description
The function to calculate the tax in tax-plugins are run on `afterCartUpdate` event. These functions get the tax an update the cart with the new tax. If `afterCartUpdate` is fire after this, it causes the function to run again.....
But shouldn't `afterCartUpdate` be called after updating the cart with tax rate?

aldeed:
Resolves #4710 
Resolves #4711
Resolves #4712 
Resolves #4687 
Impact: **breaking**  
Type: **feature**

Related docs updates: https://github.com/reactioncommerce/reaction-docs/pull/707

## Changes
- We've created a new `taxes-rates` plugin in the `included` folder, and all features related to custom rates have been moved there. This includes the "Custom Rates" panel in tax settings; the `Taxes` collection and its related schemas; the "taxes/addRate", "taxes/editRate", and "taxes/deleteRate" Meteor methods, and the "Taxes" Meteor publication.
- In the core `taxes` plugin, the general settings have been converted to React, in a `GeneralTaxSettings` component. This new component now allows you to set a single active tax service (or none). This replaces the previous way of enabling or disabling each service. All services are now enabled, but only one is active per shop.
- The core `taxes` plugin has a new API for registering tax services (such as the included "Custom Rates" service, or a custom Avalara service for example). They are registered by passing in a `taxServices` array to `registerPackage`:

```
import calculateOrderGroupTaxes from "./server/no-meteor/util/calculateOrderGroupTaxes";
import getTaxCodes from "./server/no-meteor/util/getTaxCodes";

Reaction.registerPackage({
  label: "Custom Rates",
  name: "reaction-taxes-rates",
  icon: "fa fa-university",
  autoEnable: true,
  taxServices: [
    {
      displayName: "Custom Rates",
      name: "custom-rates",
      functions: {
        calculateOrderGroupTaxes,
        getTaxCodes
      }
    }
  ]
  // ...
});
```

Every service must provide one function: `calculateOrderGroupTaxes`. It may also provide a `getTaxCodes` function to return a list of tax codes so that operators can select from a list of tax codes rather than having to enter as free text. Refer to updated public docs for more information about these functions.
- All tax-related fields on ProductVariant, CatalogProductVariant, CartItem, OrderItem, Cart, and OrderFulfillmentGroup are now removed from core code and instead extended onto the schemas within the core `taxes` plugin. This is true for both SimpleSchemas and GraphQL schemas.
- The core `taxes` plugin has two new GraphQL queries: `taxCodes` and `taxServices`
- The core `taxes` plugin also has two new React HOCs that can be used for getting the list of tax codes or registered tax services: `withTaxCodes` and `withTaxServices`
- The core `taxes` plugin adds a `getFulfillmentGroupTaxes` mutation (internal mutation, not GraphQL). This is called by the `order` and `cart` plugins to (re)calculate taxes for orders and carts. Most of the actual calculation work is done by the `calculateOrderGroupTaxes` function that was registered by the active tax service.
- Some tax-related fields on Cart, CartItem, Order, OrderFulfillmentGroup, and OrderItem have been moved, renamed, added, or removed. We've attempted to remove all unused fields, and group or rename other fields for clarity. One example is the `taxes` array, which now has a different schema and appears for individual items as well as the full cart or order fulfillment group.
- On `Products` documents, `taxable` is now `isTaxable`. This change had previously been made in the `Catalog` schema and now is made in `Products` to match.

## Breaking changes
These changes are breaking in a number of ways. There are breaking changes since 2.0.0-rc.5 release, but this also completes a broader set of breaking changes between 1.x and 2.x. The major version breaking changes are listed above.

In particular, for the Custom Rates plugin, be aware that the `taxCode` value is now used for filtering which products should be taxed at that rate. This requires a review of all your products to ensure that they have a tax code specified, in addition to being marked as taxable. If you'd rather not do this review, you can revert to the old behavior of ignoring tax codes by editing each of your Custom Rates entries, clearing the the "Tax Code" field, and saving.

*If you are upgrading from 1.x and use only Custom Rates for taxes, data migrations should provide a seamless transition. Most changes are breaking only for third-party non-included tax plugins. However, please verify after upgrading that the correct tax service is active.*


zenweasel:
Added Akarshit as a reviewer as he is testing out the API by actually writing a tax plugin

zenweasel:
When saving a default tax code the fields gets emptied out after save

zenweasel:
Testing this in the "classic UI" I don't get tax showing at all.

zenweasel:
Testing in Storefront on the `develop` branch with Custom Rates enabled also not seeing taxes

zenweasel:
Also get this error when checking out
```
=> App running at: http://localhost:3000/
02:08:15.737Z ERROR Reaction: createOrder: error creating payments Effective tax rate must be of type Number
02:08:15.745Z ERROR Reaction:  (errorId=cjo6ixoqp000037wbj7pkz7uw, userId=rsB7XKbyb3XBnnxoZ)
```

zenweasel:
So it looks like taxes are not being calculated because `isTaxable` is set to false. However I am not able to toggle  "Taxable" in the PDP. I am able to toggle this on the `release-2.0.0-rc.6` branch

zenweasel:
So even when I set `isTaxable` manually on the variant I am not able to get taxes to come up with "Custom Rates.

It also appears that something about the migration is not working as the default product is not getting `isTaxable` set correctly.

zenweasel:
Even when creating a new product I can't set "Taxable" in the PDP

imports/plugins/included/product-variant/containers/variantFormContainer.js
```diff
@@ -49,52 +49,8 @@ const wrapComponent = (Comp) => (
 
         return validationStatus;
       }
-    }
-
-    isProviderEnabled = () => {
-      const shopId = Reaction.getShopId();
 
-      const provider = Packages.findOne({
```
zenweasel: So this provided functionality to create a dropdown with available taxcodes that are valid for the tax provider. Are we not allowing that now?

aldeed: It still works, just in a different way. There's a new GraphQL query that is called to get the list of available tax codes, which then triggers showing the Select if a list is returned.

aldeed:
@zenweasel I fixed the UI issues. The settings form had a reactivity bug (although it was properly saving changes, didn't look like it). The variant saving was hitting a bug with the collection2 package, that was triggered by my moving around the tax schemas. I must not have retested that form after moving them. The bug is fixed in collection2 package and I updated Reaction to the latest version of that package.

Can you retest and let me know if you're still seeing any issues? A couple things to note:
- You should hard refresh the browser after changing the active tax service for the shop. There are non-reactive GraphQL queries involved now.
- Tax codes are respected by the custom rates service now, so to see taxes for a variant, make sure it `isTaxable` AND has `RC_TAX` tax code. (Although this is a fix, it might also be considered a breaking change. Maybe there's a way to make it clearer.)

Akarshit:
@aldeed 
I wrote a custom tax-plugin and was running into the same problem as Brent that Tax was displayed as `-`. Then I just pulled your latest commits ans still the Tax is displayed as `-`. 
Below is a the data returned by the tax-plugin's function:
```js
{ itemTaxes:
   [ { itemId: 'ZMgFEzXbusM7FqHRB',
       tax: 0.08,
       taxableAmount: 59.97,
       taxes:
        [ { _id: 'Ytuv7L9bvrBGQRSFC',
            jurisdictionId: '24354',
            sourcing: 'origin',
            tax: 0.8,
            taxableAmount: 19.99,
            taxName: 'SELLER_USE',
            taxRate: 0.04 },
          { _id: '5r6yPJBDmLDs297Ry',
            jurisdictionId: '25083',
            sourcing: 'origin',
            tax: 0.9,
            taxableAmount: 19.99,
            taxName: 'SELLER_USE',
            taxRate: 0.045 },
          { _id: 'ygjJAabt3r5MdDihD',
            jurisdictionId: '79774',
            sourcing: 'origin',
            tax: 0.08,
            taxableAmount: 19.99,
            taxName: 'SELLER_USE',
            taxRate: 0.00375 } ] } ],
  taxSummary:
   { calculatedAt: 2018-11-08T19:00:21.218Z,
     referenceId: '123',
     tax: 1.7800000000000002,
     taxableAmount: 59.97,
     taxes:
      [ { _id: 'cfG6Rpg6EXuNetEBe',
          jurisdictionId: 'RHfdBazBeq4uZdnEf',
          taxName: 'SELLER_USE',
          tax: 1.7800000000000002,
          taxableAmount: 59.97,
          sourcing: 'origin',
          taxRate: 0.02968150742037686 } ] } }
```

Akarshit:
Also I see that there is a checkbox to mark if tax should be calculated on shipping.
But I don't see that value being passed to the function `calculateOrderGroupTaxes`.
I don't think that a call should be made inside the function to fetch the core-taxes package and determine if tax should be calculated on shipping, right?

aldeed:
@Akarshit Sorry, I had not tested in the new storefront UI since the changes were all server-side. But it turns out that the new storefront never had tax summary implemented so it always shows `-` even when there are taxes in the data. I'll do that in a separate PR for the starter kit repo.

I verified just now that it does work for me in the classic UI checkout, though, so you can test your plugin using that for now.

aldeed:
@Akarshit Correction, there is one place in the new UI where tax is shown correctly: on the final review step of checkout. So if you enter payment info and get to that step, you should see the tax amount. I'll fix the other places.

aldeed:
@Akarshit Here is a starterkit branch with the UI fix: https://github.com/reactioncommerce/reaction-next-starterkit/pull/424

It should work with that branch of storefront and this branch of Reaction API

zenweasel:
@aldeed Also, do we need a migration to set the Tax Code for all existing products to RC_TAX if Custom Rates are in use? It seems like you were saying that this Tax Code needs to be set for it to calculate tax as well.

aldeed:
@zenweasel It is now working when purchasing options, and I fixed the Publish button indicator for the tax fields.

I'm nervous about trying to automatically figure out "if custom rates is enabled" since in the past you could enable multiple tax plugins at once. I wouldn't want to set all variants to `RC_TAX` if in reality they were already set properly for some other plugin.

Some possible ideas:
- Migration to set `taxCode` to `RC_TAX` only if `taxCode` is currently `0000`, which was previously the default.
- Just document the change in the release notes, with an example MongoDB command they can run manually to set all to `RC_TAX`
- Otherwise, since we don't yet have any UI to add additional tax codes, we could NOT require the tax code be set. We could add a plugin shop setting `filterByTaxCode` and have that be `false` by default. Then it would tax any product that is taxable without also needing `RC_TAX` tax code on it (like the behavior prior to this PR).
- Lastly, I could just plain remove the tax code check and go back to how it worked before this. I think it's sort of confusing that it shows `RC_TAX` in the custom rates table if it doesn't actually check that, but maybe not as confusing as fixing it.

@spencern Any thoughts on this?

zenweasel:
It seems to be calculating tax now but when setting "Charge tax on Shipping" to be true it's not changing the tax total, but it is reflected in the cart total so that those charges don't foot correctly and would cause customer confusion.

zenweasel:
Multi-jurisdiction pricing does seem to be working

aldeed:
`Charge tax on shipping` has never been implemented in custom rates plugin. I'm not sure if other tax plugins use it, but if not, we could potentially remove it. It seems like kind of a weird setting to have anyway since I would think that would usually be determined by the jurisdiction rules that the tax plugin uses.

spencern:
@aldeed @zenweasel I lean towards option 2:
> Just document the change in the release notes, with an example MongoDB command they can run manually to set all to RC_TAX

spencern:
We should have a way to charge tax on shipping for the custom rates - on a per-locale basis. If that's not been implemented to this point, it could be a separate bit of work.

zenweasel:
@aldeed Question, tax plugins can add additional settings still correct? I know Avalara has about another 10-15 settings that they require the user to set.

I think what's probably missing with "charge tax on shipping' would be a "shipping tax code". I know with Avalara that was how it determined how to charge or exclude shipping from the calculation.

imports/plugins/core/orders/server/no-meteor/mutations/createOrder.js
```diff
@@ -121,11 +120,14 @@ async function getCurrencyExchangeObject(collections, cartCurrencyCode, shopId,
...
 function getInvoiceForFulfillmentGroup(group, discountTotal) {
+  const { taxSummary } = group;
+
   // Items
   const itemTotal = group.items.reduce((sum, item) => (sum + item.subtotal), 0);
 
   // Taxes
-  const taxTotal = group.items.reduce((sum, item) => (sum + item.tax), 0);
+  const { tax: taxTotal, taxableAmount } = taxSummary;
+  const effectiveTaxRate = taxTotal / taxableAmount;
```
Akarshit:
```suggestion
  const effectiveTaxRate = taxableAmount > 0 ? taxTotal / taxableAmount : 0;
```

aldeed:
OK, so I've addressed the remaining things for now:
- I added `taxCode` to the custom rates add/edit form so now you can remove `RC_TAX` if you don't want the rate filtered based on the product tax code needing to match. I added a note about this in the breaking changes section of this PR description.
- Removed that "Tax shipping" check box since it doesn't appear that Avalara uses that either. Adding some way to tax shipping with the Custom Rates plugin is a separate project. I'll make an issue.
- Fixed it so that the "default tax code" actually does get set on new variants when you create them.

@Akarshit That checkout error appears to be an issue with storefront, that it isn't including the tax when trying to create the order. I'll fix that in the starterkit repo.

Pretty sure this branch should be good to merge now, but I'll let it hang out for another day or two while you're still testing the custom plugin against it.

aldeed:
@Akarshit Actually I was wrong. I believe that error would mean that your plugin's calculation function is returning `1.78` for `taxSummary.tax` when called for a cart, but is returning `0` for that when called for the order group. It's hard to say whether there is any core bug involved without seeing what your function is returning in each case.

aldeed:
I'm going to merge this since it's big and approved. Any more issues that are found can be submitted and fixed in a follow-up PR.

PR #4788, opened 2018-11-01
Support for Google Search Console Site Verification

### LINKED ISSUE #4784: Google Search Console Plugin for Reaction 2.0
## Issue Description
The current Google Search Console plugin is not compatible with the new decoupled architecture of Reaction's back-end operator app and the Front-end app.  The current plugin also has a spot for google analytics integration, which is no longer needed due to the segment integration that already exists.  Note that segment does not contain a google search console integration. 

## Possible Solution
Build a new plugin for only google search console integration that is built for Reaction 2.0 and above.  It could have a similar UI/UX for entering the url code generated by the google search console account.  This would need to be processed and published by the front end graphQL API in order for google bots to register and reconcile the ownership of the front end URL and the google search account.

ajporlante:
Resolves #4784
Impact: **minor**  
Type: **feature**

## Issue
This allows a shop owner to verify their ownership of their site for Google Search Console for better SEO performance.

## Solution
I created a new plugin `google-webtools` to set up the token that should be present in a header meta tag in the storefront.

## Breaking changes
None

## Testing
1. As an admin, go to the shop dashboard panel, and enter a string in Site Verification Token field and click save. The string is supposed to be the token provided in the Google Search Console dashboard.
2. The same string should be returned by the following query:
```
query siteVerificationTokenQuery {
  siteVerificationToken
}
```

### DIFF
--- a/imports/plugins/included/google-webtools/server/methods.js
+++ b/imports/plugins/included/google-webtools/server/methods.js
@@ -1,0 +1,30 @@
+import { Meteor } from "meteor/meteor";
+import { check } from "meteor/check";
+import { EJSON } from "meteor/ejson";
+import Reaction from "/imports/plugins/core/core/server/Reaction";
+import { Packages } from "/lib/collections";
+
+export const methods = {
+  /**
+   * @name googleWebTools/updateGoogleWebToolsSettings
+   * @summary Updates Google web tools settings
+   * @method
+   * @param  {Object} values - Object with field names as key and field values as value
+   * @return {Promise} - Promise
+   */
+  "googleWebTools/updateGoogleWebToolsSettings"(values) {
+    check(values, Object);
+    check(values.siteVerificationToken, String);
+
+    // must have core permissions
+    if (!Reaction.hasPermission("core")) {
+      throw new Meteor.Error("access-denied", "Access Denied");
+    }
+
+    const stringValue = EJSON.stringify(values);
+    const update = EJSON.parse(stringValue);
+    return Packages.update({ name: "google-webtools" }, { $set: { settings: { public: update } } });
+  }
+};
+
+Meteor.methods(methods);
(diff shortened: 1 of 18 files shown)


spencern:
This seems like a lot of code to add for some functionality that can be accomplished in several other ways. Why is this necessary? What benefit does this particular method have over simpler methods that don't add additional code to the core application?

image

ajporlante:
I think they wanted an easy way for a shop admin without much technical knowledge to do this. HTML file upload  requires admin to touch the server and domain verification is not easy when admin does not have access to their domain provider. I think @vwadher is the better guy to discuss this.

spencern:
Verifying the domain can be done really easily with a DNS record and is not something that should be baked into a commerce application.

vwadher:
The HTML file upload method is google's "recommended" method and is the easiest for a marketing role based operator.   Therefore this was the 1st choice keeping in mind the end user and google's suggestion on this being the recommended method.

image

That said, there is a valid point here so I will review/test out the other methods for ease of implementation and feasibility and will report back.

imports/plugins/included/google-webtools/server/methods.js
```diff
@@ -0,0 +1,30 @@
...
+
+export const methods = {
+  /**
+   * @name googleWebTools/updateGoogleWebToolsSettings
+   * @summary Updates Google web tools settings
+   * @method
+   * @param  {Object} values - Object with field names as key and field values as value
+   * @return {Promise} - Promise
+   */
+  "googleWebTools/updateGoogleWebToolsSettings"(values) {
```
spencern: We should not be adding any new Meteor Methods to the application.

spencern:
This PR does not upload an HTML file.

zenweasel:
Closing in favor of other approaches that don't require RC code modifications

PR #4798, opened 2018-11-09
Update graphql packages to support extend enum and extend union

### LINKED ISSUE #4759: Can't add a payment provider w/o modifying core
## Issue Description

Currently the available payment methods are hard-coded into the Reaction Meteor app. This means that plugins have no way of extending it to add or remove a payment method.

Comments in the code seem to indicate that `extend union` and `extend enum` were meant to be used here but these are currently unimplemented.

Current code in `imports/plugins/core/payments/server/no-meteor/schemas/schema.graphql` looks like this:
```
# These should be defined in their respective plugin schemas, but `extend union` isn't working yet
union PaymentData = ExampleIOUPaymentData | StripeCardPaymentData | MarketplaceStripeCardPaymentData | RadialCardPaymentData

# These should be defined in their respective plugin schemas, but `extend enum` isn't working yet
"The name of a payment method, which is how payment methods are keyed"
enum PaymentMethodName {
  iou_example
  stripe_card
}
```

My solution at the moment is to write custom code into this file which can be avoided if both `extend union` and `extend enum` are supported.

### Versions
```
Node: 10.10.0
NPM: 5.6.0
Meteor Node: 8.11.4
Meteor NPM: 6.4.1
Reaction CLI: 0.29.0
Reaction: 2.0.0
Reaction branch: master
Docker: 18.06.1-ce
```

ajporlante:
Resolves #4759
Impact: **minor**  
Type: **refactor**

## Issue
We are unable to add custom payment methods without modifying core code as some parts of the schema have to be extended in the plugins. Particularly, we have to use `extend union` and `extend enum` in custom plugins, but these graphql specs are not supported in the version of `graphql` and `graphql-tools` that Reaction currently uses.

## Solution
Upgrading `graphql` and`graphql-tools` to the respective latest releases enables supporting `extend union` and `extend enum`.

## Breaking changes
None

## Testing
In the graphql UI, search for `PaymentMethodName` and `PaymentData` in the schema docs. See that `PaymentMethodName` includes `stripe_card` as an option and `StripeCardPaymentData` as possible type.


package.json
```diff
@@ -68,12 +68,12 @@
     "fibers": "^2.0.0",
     "flatten-obj": "^3.1.1",
     "font-awesome": "^4.7.0",
-    "graphql": "0.13.1",
```
zenweasel: So we went from version 0 to version 14?

imports/plugins/core/payments/server/no-meteor/schemas/schema.graphql
```diff
@@ -55,14 +55,14 @@ type ExampleIOUPaymentData {
...
+# Use `extend union` to define a payment plugin's PaymentData type
+"Data identifying a payment for an order"
+union PaymentData = ExampleIOUPaymentData
 
-# These should be defined in their respective plugin schemas, but `extend enum` isn't working yet
+# Use `extend enum` to add a plugin's payment method name
 "The name of a payment method, which is how payment methods are keyed"
 enum PaymentMethodName {
   iou_example
-  stripe_card
```
zenweasel: Why do we remove stripe here without extending the enum elsewhere?

package.json
ajporlante: Yes, that's right. I was surprised too. `0.13.1` is nine months ago

imports/plugins/core/payments/server/no-meteor/schemas/schema.graphql
ajporlante: The extension has always been in the stripe plugin here:

https://github.com/reactioncommerce/reaction/blob/f9fd923a4a752e9a996853094693d7d48f2d05d7/imports/plugins/included/payments-stripe/server/no-meteor/schemas/schema.graphql#L43-L45

It just did not work before.

zenweasel:
For some reason I am not able to check out in the StarterKit although I can checkout in the Classic UI. I am getting no shipping methods even though they are configured

ajporlante:
Hi @zenweasel , I think the fulfillment methods are loaded to cart when cart is created and not updated when fulfillment methods are enabled/disabled. So a new cart needs to be created to load the available fulfillment methods into the cart and the UI.

I am also unable to make the `availablePaymentMethods` query work, but I think that's a bug that is outside the scope of this PR.

aldeed:
I pushed a commit to fix the failing unit tests

PR #4799, opened 2018-11-11
Fix CartCleanupJob

### LINKED ISSUE #4786: Cart cleanup job crashes on start
## Issue Description
When the cart cleanup job (which removes all anonymous carts over X days) starts, it crashes with this error:

```
Exception in setTimeout callback: TypeError: moment is not a function
    at JobQueue.Jobs.processJobs [as worker] (imports/plugins/included/jobcontrol/server/jobs/cleanup.js:36:23)
    at JobQueue._process (/Users/brenthoover/Projects/js/reaction/node_modules/@reactioncommerce/job-queue/dist/common/jobFactory.js:432:23)
    at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1304:12)
    at packages/meteor.js:620:25
    at runWithEnvironment (packages/meteor.js:1356:24)
```

## Steps to Reproduce
1. In the Dashboard in Shops/Options set the cleanup duration to a very short period.
1. Wait
1. Observe the error

## Possible Solution
With the way that carts are created now it may no longer be necessary to have this function but I am not sure.

This error probably has to do with someone trying to lazyload moment which is entirely not necessary on the server-side, which is where this code lives.

### Versions
Reaction: 2.0.0-rc5

willmoss1000:
Resolves #4786  
Impact: **minor**  
Type: **bugfix**

## Issue
Refactor of cartCleanupJob, which currently doesn't run.

## Solution

- Refactored for the new cart system. 
- Moment lazyload also wasn't working, so this is now imported directly.
- Using `job.progress` instead of `job.done` - this fixes "job not found" errors, and uses the Job package as intended.

## Breaking changes
N/A

## Testing
1. Set the cleanup schedule to a small time period
2. Observe old carts removed
3. Observe accounts without emails attached are removed

### DIFF
--- a/imports/plugins/included/jobcontrol/server/jobs/cart.js
+++ b/imports/plugins/included/jobcontrol/server/jobs/cart.js
@@ -4,12 +4,7 @@ import { Meteor } from "meteor/meteor";
 import { Accounts, Cart, Jobs } from "/lib/collections";
 import Reaction from "/imports/plugins/core/core/server/Reaction";
 import { Job } from "/imports/plugins/core/job-collection/lib";
-
-let moment;
-async function lazyLoadMoment() {
-  if (moment) return;
-  moment = await import("moment").default;
-}
+import moment from "moment";
 
 /**
  * @param {Object} olderThan older than date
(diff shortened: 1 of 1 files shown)


aldeed:
@willmoss1000 Thanks for tackling this! The direct import of `moment` is fine, but the code that deletes the linked account if it has no email is no longer necessary and won't do anything. Anonymous carts no longer have an `accountId` on them, and carts that do have an `accountId` will never be linked to an account with no email. The use of "anonymous accounts" is effectively phased out, even though the old UI may still create them for now.

aldeed:
Probably the whole thing could be simplified to `Cart.remove({ updatedAt: { $lte: olderThan } }, { multi: true })` now.

spencern:
@aldeed @zenweasel @willmoss1000 what do we need to do to get this moving?

willmoss1000:
@aldeed what you've suggested makes this simple. I've updated the PR accordingly.

aldeed:
Thanks @willmoss1000 !

PR #4802, opened 2018-11-13
ECS deployment: move TLS certificate ARN from propel.yaml to ENV vars

griggheo:
Resolves #issueNumber  
Impact: **breaking**  
Type: **refactor**

## Issue
The syntax for the configuration file for the CLI tool propel has been changed so that the TLS Certificate ARN can be specified via an ENV variables and not hardcoded in the configuration file.

## Solution
One-line change to the propel.yaml file.

## Breaking changes
This is a breaking change in that old propel configuration files will break the deployment process.

## Testing
Make sure https://reaction-core.staging.reactioncommerce.com is accessible after a deployment.

### DIFF
--- a/propel-feat.yaml
+++ b/propel-feat.yaml
@@ -1,6 +1,5 @@
 services:
 - name: reaction-core
-  certificate_arn: arn:aws:acm:us-west-2:773713188930:certificate/c2979a7a-7b84-43ed-b3a7-24b0256a1b9f
   root_domain: staging.reactioncommerce.com
   dns_name: reaction-core
   desired_task_count: 2
@@ -21,7 +20,7 @@ services:
       - container_port: 3000
         host_port: 3000
       image: reactioncommerce/reaction
-      image_tag: release-2.0.0-rc.5
+      image_tag: release-2.0.0-rc.7
       env_params:
       - name: REACTION_AUTH
       - name: REACTION_EMAIL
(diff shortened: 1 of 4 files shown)


propel-feat.yaml
```diff
@@ -31,27 +30,40 @@ services:
...
-    name: staging-storefront
+    name: staging-storefront-core
     containers:
+    - name: nginx
+      cpu: 128
+      memory: 256
+      port_mappings:
+      - container_port: 8082
+        host_port: 8082
+      image: sportsdirect/nginx
```
aldeed: Is `sportsdirect/nginx` correct for this repo?

aldeed:
Seems OK as far as I can tell. Just one question, and I'm rerunning app tests that failed. Probably just a random failure.

propel-feat.yaml
griggheo: Good catch, I fixed it.

spencern:
Failed with 
```
Waiting for CF stack "staging-service-reaction-core" to be updated
CreateRelease: DeployReleaseForService: Error waiting for CF stack "staging-service-reaction-core" to be updated: ResourceNotReady: failed waiting for successful resource state
```
during deploy to ecs release branch

griggheo:
Yup looking at it. Looks like AWS user permissions.

PR #4803, opened 2018-11-14
Use no-meteor functions for payment capture and refund methods

ajporlante:
Impact: **major**  
Type: **refactor**

## Issue
Payment capture and refund are currently done by creating Meteor methods named using `processorName + methodName` within a payment plugin . As we intend to minimize using Meteor, we want to use no-meteor functions instead by registering them into the `functionsByType` parameter in a plugin's `register.js` file.

## Solution
Instead of creating `processor/method-name` methods, payment plugins will need to register no-meteor functions in the `register.js` using the `functionsByType` parameter. In particular, payment plugins need to create the following functions:

```
${processor}CapturePayment
${processor}RefundPayment
${processor}ListRefund
${processor}DeAuthorizePayment
```

## Breaking changes
Payment plugins that use Meteor methods for capture and refund will not be compatible with this PR.

## Testing

1. 
As admin, enable example payment method in the admin dashboard. 
As customer, create an order and pay with the example payment method (this can be done in the classic UI). 
As admin, open the  newly created order in the dashboard panel for Orders.
Click Approve, and Capture. See that Captured Total is updated.
In the same order, enter a refund amount and click Refund. Check that the order in the database has additional success transactions.

2
As admin, enable Stripe payment method in the admin dashboard and enter correct credentials. 
As customer, create an order and pay with Stripe payment method (this can be done in the classic UI). 
As admin, open the  newly created order in the dashboard panel for Orders.
Click Approve, and Capture. See that Captured Total is updated.
In the same order, enter a refund amount and click Refund. See that the refunded amounts are displayed in the order.
For Stripe, you can verify the amounts captured and refunded in the Stripe merchant dashboard.


zenweasel:
@ajporlante I don't understand this part

> need to create the following functions:

${processor}CapturePayment
${processor}RefundPayment
${processor}ListRefund
${processor}DeAuthorizePayment

Why do we care about what the methods are called? Couldn't they just be in a nested object that defines capture/refund/list etc. under a key for the payment provider? I would like to get away from using the names as keys because this has always brittle and causes problems with payment providers who have non-javascript safe characters in their names

imports/plugins/included/payments-stripe/register.js
```diff
@@ -2,12 +2,20 @@
...
 Reaction.registerPackage({
   label: "Stripe",
   name: "reaction-stripe",
   icon: "fa fa-cc-stripe",
   autoEnable: true,
+  functionsByType: {
+    stripeCapturePayment: [stripeCapturePayment],
+    stripeCreateRefund: [stripeCreateRefund],
+    stripeListRefund: [stripeListRefund]
+  },
```
aldeed: Instead of this, add a `functions` object to the `paymentMethods` object below.

```js
paymentMethods: [{
  name: "stripe_card",
  displayName: "Stripe Card",
  functions: {
    capturePayment: stripeCapturePayment,
    createRefund: stripeCreateRefund,
    listRefund: stripeListRefund
  }
}],
```

Then when you need to use one of the functions, you can get the registered payment method by name based on what's in the order payment data, and call `paymentMethod.functions.capturePayment()` for example.

ajporlante:
@aldeed @zenweasel Reworked it as you requested. You can check, then I can proceed with removing the Meteor method files and rewriting the tests.

imports/plugins/core/orders/server/methods/capturePayments.js
```diff
@@ -35,16 +38,14 @@ export default function capturePayments(orderId) {
...
-    const processorLowercase = processor.toLowerCase();
+  const { mode, name, status, transactionId } = payment;
 
+  if (mode === "capture" && status === "approved" && name) {
     let result;
     let error;
     try {
-      result = Meteor.call(`${processorLowercase}/payment/capture`, payment);
+      const context = Promise.await(getGraphQLContextInMeteorMethod(Reaction.getUserId()));
+      result = Promise.await(paymentMethods[name].functions.capturePayment(context, payment));
```
aldeed: @ajporlante This is correct, but it's likely that at some point people could switch payment providers and end up removing the old one, in which case `paymentMethods[name]` will be undefined and you'll get a cryptic error. I'd rather give them a nice error.

You could add a function in `pluginRegistration.js`:

```js
export function getPaymentMethodConfigByName(name) {
  const config = paymentMethods[name];
  if (!config) {
    throw new Error(`Configuration not found for ${name} payment method. Did you remove the plugin that provides this payment method?`);
  }
  return config;
}
```

And then import and use `getPaymentMethodConfigByName(name)` in these methods.

aldeed:
@ajporlante See my one minor comment but otherwise looking beautiful

ajporlante:
@zenweasel This is ready for another review. I have reworked as @aldeed suggested, removed the unnecessary Meteor methods and tests, and rewrote the unit tests for the new no-meteor functions.

ajporlante:
@aldeed Fixed.

PR #4805, opened 2018-11-14
feat: Put mongo on the streams network

focusaurus:
Impact: **minor**  
Type: **feature**

## Issue

We are building some data integrations for a separate project and need connectivity to mongo, so putting it on the streams network for local development. This doesn't affect production deployments. The related PR into reaction-platform for the `streams` network is [PR 28](https://github.com/reactioncommerce/reaction-platform/pull/28). It would be best to merge the PR into reaction-platform first then merge this PR.

## Solution
Put mongo onto the `streams` network, adjust as needed in the config and code. I also removed an additional hard coding in `waitForMongo.js` so more code is properly using the `MONGO_URL` environment variable.

## Breaking changes

This will require the docker network `streams.reaction.localhost` to be created, which developers can do by pulling down the latest `reaction-platform` and running `make` (or `make network-create` if they want to be surgical about it).

## Testing

- Unit tests still pass
- I manually ran eslint on `waitForMongo.js` as it's normally excluded
- I ran reaction locally and confirmed I could still read and write to mongo via the reaction web app
- I manually tested `waitForMongo.js`


docker-compose.yml
```diff
@@ -22,15 +25,16 @@ services:
     depends_on:
       - mongo
     environment:
-      MONGO_URL: "mongodb://mongo:27017/reaction"
-      MONGO_OPLOG_URL: "mongodb://mongo:27017/local"
+      MONGO_URL: "mongodb://mongo.streams.reaction.localhost:27017/reaction"
+      MONGO_OPLOG_URL: "mongodb://mongo.streams.reaction.localhost:27017/local"
```
ticean: These URLS could be left unchanged. Mongo is still on the default network Better if the network addition doesn't bleed into this projects internals?

docker-compose.yml
```diff
@@ -22,15 +25,16 @@ services:
...
       ROOT_URL: "http://localhost:3000"
       HYDRA_ADMIN_URL: "http://hydra:4445"
       HYDRA_OAUTH2_INTROSPECT_URL: "http://hydra:4445/oauth2/introspect"
     networks:
       default:
       api:
       auth:
+      streams:
```
ticean: If the default network is used for the Mongo connection then this service doesn't need to be on the streams network.

package.json
```diff
@@ -200,7 +200,7 @@
     }
   },
   "scripts": {
-    "devserver": "ROOT_URL=http://localhost:3030 MONGO_URL=mongodb://localhost:27017/reaction NODE_ENV=reaction-node BABEL_DISABLE_CACHE=1 nodemon ./imports/node-app/devserver/index.js",
+    "devserver": "ROOT_URL=http://localhost:3030 MONGO_URL=mongodb://mongo.streams.reaction.localhost:27017/reaction NODE_ENV=reaction-node BABEL_DISABLE_CACHE=1 nodemon ./imports/node-app/devserver/index.js",
```
ticean: The default network can be used here, too.

docker-compose.yml
focusaurus: Testing now...

focusaurus: OK I made those changes and things still work. Pushed amended commit.

aldeed:
LGTM if @ticean approves of the changes

aldeed:
May also need to look into why the feature branch deploy step failed on CI. I don't know if deploy will fail when merged in the same way. @griggheo

focusaurus:
@griggheo Looks like errors running propel

```
download: s3://reaction-staging-propel-artifacts/propel-linux-amd64 to ./propel
Error: unknown flag: --suffix
Usage:
  propel param copy [flags]
```

griggheo:
@focusaurus  I think you can go ahead and merge if you want, and then I can fix the propel part via another PR.

> @griggheo Looks like errors running propel
> 
> ```
> download: s3://reaction-staging-propel-artifacts/propel-linux-amd64 to ./propel
> Error: unknown flag: --suffix
> Usage:
>   propel param copy [flags]
> ```

focusaurus:
@spencern Do you have admin to merge this despite the CI error?

PR #4808, opened 2018-11-16
(feat) Add meteor caching

zenweasel:
Resolves #NA 
Impact: **minor**  
Type: **performance**

## Issue
Many developers use the `METEOR_DISABLE_OPTIMISTIC_CACHING` variable, but it's not set when using Docker

## Solution
Setting the environment variable `METEOR_DISABLE_OPTIMISTIC_CACHING` tends to make restarts faster and reduces CPU usage. Setting this in the dockerfile so that `reaction-platform` users can benefit from settings this variable

## Breaking changes
None

## Testing
1. Start `reaction-platform` and observe that this setting is in effect.


aldeed:
Closing this because I did it in #4826 now, which would overwrite the change in this PR anyway.

PR #4811, opened 2018-11-19
Fix tax calculation arguments, other tax fixes

### LINKED ISSUE #4807: Different params passed to tax-plugin.
## Issue Description
When the function `calculateOrderGroupTaxes` is called on the checkout screen, the cart(modified) is passed in which the `items` has the property `priceWhenAdded`. But when the order is placed, the same function is called with parameters in which the `items` has the property `price`. This is causing the plugin to break.

## Steps to Reproduce
1. Add a taxed-plugin with following function to calculate the tax:
```js
export default async function calculateOrderGroupTaxes({ context, group }) {
  if (!group || !group.items || group.items.length === 0) {
    return {
      itemTaxes: [],
      taxSummary: {
        calculatedAt: new Date(),
        referenceId: Random.id(),
        tax: 0,
        taxableAmount: 0,
        taxes: []
      }
    };
  }
  console.log("price", group.items[0].price, "priceWhenAdded", group.items[0].priceWhenAdded);
  return {
      itemTaxes: [],
      taxSummary: {
        calculatedAt: new Date(),
        referenceId: Random.id(),
        tax: 0,
        taxableAmount: 0,
        taxes: []
      }
  }
}
```
2. Checkout with a product and see the logs in the console.

### Versions
I checked this with `origin/refactor-aldeed-split-tax-rates-plugin`

aldeed:
Resolves #4807   
Impact: **minor**  
Type: **feature**

## Changes
The order data being passed to tax calculation functions was not consistent due to differences between Cart and Order schemas. This PR introduces a new schema CommonOrder, which represents one actual or potential order for one shop. New transforms allow a cart fulfillment group or an order fulfillment group to be converted to a CommonOrder. Tax calculation functions only need to understand and deal with CommonOrders.

This was implemented in a generic way because we expect to need this same transformation for shipping restrictions, surcharges, and other similar calculations that do not need to care whether they're calculating for a Cart or an Order.

## Breaking changes for custom tax plugins
There are breaking changes only within the current RC7 release, since prior PRs were merged. Custom tax plugins should do the following:
- Change function name from `calculateOrderGroupTaxes` to `calculateOrderTaxes`
- Expect that `calculateOrderTaxes` will be called with `{ context, order }` rather than `{ context, group }`, where `order` is similar to what `group` was, but is guaranteed to be in the new `CommonOrder` schema, whether being called for a Cart or an Order. Among other things, a common order has `originAddress` and `currencyCode` properties, and the items are guaranteed to have `price` property set.
- Expect `calculateOrderTaxes` to now be called even when `shippingAddress` is `null`. Return `null` if you cannot calculate taxes. In most cases, tax plugins will immediately check whether `shippingAddress` is `null` and if so, return `null`.
- A `calculateOrderTaxes` function may now set `calculatedByTaxServiceName` property in the `taxSummary` that is returned. Set this to the name of your tax service (same as in `taxServices` array in `registerPackage`). One good reason to set this is to determine whether tax was calculated by your tax service after the order is placed. Depending on the type of tax integration, you may need to inform the third-party tax service that the order was placed. Use `referenceId` to store the corresponding calculation ID from the third-party system if necessary.

## Testing
- Test all aspects of taxation. Primarily, test adding taxable items to your cart and checking them 
out.
- To test the specific bug fix, you can log in `calculateOrderTaxes` function for the included "custom rates" service and verify that the `order` argument is the same whether being called for a cart (such as when adding items) or for an order (such as when completing checkout).

### DIFF
--- a/imports/collections/schemas/cart.js
+++ b/imports/collections/schemas/cart.js
@@ -64,6 +64,10 @@ export const CartItem = new SimpleSchema({
     optional: true
   },
   "attributes.$": CartItemAttribute,
+  "compareAtPrice": {
+    type: Money,
+    optional: true
+  },
   "createdAt": Date,
   "metafields": {
     type: Array,
@@ -114,6 +118,7 @@ export const CartItem = new SimpleSchema({
     index: 1,
     label: "Cart Item shopId"
   },
+  "subtotal": Money,
   "title": {
     type: String,
     label: "CartItem Title"
--- a/imports/collections/schemas/cart.js
+++ b/imports/collections/schemas/cart.js
@@ -42,6 +42,7 @@ const CartItemAttribute = new SimpleSchema({
  * @property {Metafield[]} metafields
  * @property {String} optionTitle optionTitle from the selected variant
  * @property {ShippingParcel} parcel Currently, parcel is in simple product schema. Need to include it here as well.
+ * @property {Money} price The current price of this item
  * @property {Money} priceWhenAdded The price+currency at the moment this item was added to this cart
  * @property {String} productId required
  * @property {String} productSlug Product slug
@@ -77,6 +78,7 @@ export const CartItem = new SimpleSchema({
     type: ShippingParcel,
     optional: true
   },
+  "price": Money,
   "priceWhenAdded": Money,
   "productId": {
     type: String,
--- a/imports/plugins/core/cart/server/no-meteor/mutations/convertAnonymousCartToNewAccountCart.test.js
+++ b/imports/plugins/core/cart/server/no-meteor/mutations/convertAnonymousCartToNewAccountCart.test.js
@@ -1,3 +1,4 @@
+import Factory from "/imports/test-utils/helpers/factory";
 import mockContext from "/imports/test-utils/helpers/mockContext";
 import convertAnonymousCartToNewAccountCart from "./convertAnonymousCartToNewAccountCart";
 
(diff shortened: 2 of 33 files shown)


aldeed:
@Akarshit Please test your plugin against this branch, and note the changes you'll need to make described in the PR description.

Akarshit:
@aldeed Would it make sense to add the `cartId` in the CommonOrder. Since some kind of a `id`, that can be used to relate the cart/order and request to get the tax, should be sent with the request.

aldeed:
@Akarshit 

- I can add `cartId` and `orderId` but either one might be `null` most of the time. Orders can be created without any cart, too. 
- Billing address will generally never be available when it runs for a cart, and some payment methods may not even collect a billing address, in which case it will be `null` when it runs for an order, too. But I can try to include it if we have one.
- I can add item titles and applied discounts

aldeed:
I was curious why billing address would be required, but this quote from the Radial docs explains it.

> The tax API requires an address to properly assess charges. In nearly all cases, the shipping address is used to determine taxes. However, there are certain cases when shipping address is not available, such as the purchase of a online gift certificate or an eBook. In cases where a shipping address is not available, the billing address should be used instead. For this reason, billing address is required.

aldeed:
@Akarshit I added all except discounts because the current discounts handling is janky and not calculated by item or by group. A proper promotions engine is on the roadmap very soon, so we can make sure the new promotions get added to CommonOrder at that time, and then tax plugins can be updated to handle them.

Akarshit:
@aldeed While checking this, I found that billingAddress is always `null`.
I checked this on the classic Meteor app, testing it in the starter-kit now.

aldeed:
@Akarshit I can't reproduce that. Tested with example and Stripe payment from classic UI, and `billingAddress` is correctly set when `order.sourceType` is "order" (when it's called when you place the final order). Prior to that, it's called from a Cart, and `billingAddress` will always be `null` for a cart.

If Radial requires `billingAddress` and you want estimated taxes calculated for carts, I'd just pass the shipping address as the billing address.

PR #4820, opened 2018-11-21
Add node_modules/.bin to PATH in docker

focusaurus:
Impact: **minor**  
Type: **chore**

## Issue
In the docker container I can't just run npm tools with CLIs like `eslint` outside of the defined npm scripts from `package.json`. Adding this directory to the PATH allows running such programs normally.

Just a developer convenience.


focusaurus:
OK I used `$APP_SOURCE_DIR` and ammended the commit.

PR #4821, opened 2018-11-21
feat: shipping method restrictions

kieckhafer:
Impact: **major**  
Type: **feature**

## Issue
Shops should be able to restrict available shipping methods based on shipping location and attributes of items inside an order (price, brand, type of item, etc).

## Solution
Create both universal and per-method ways to allow or deny a shipping method based on the address attached to the an order, and attributes attached to items in an order. This is done by creating a FlatRateFulfillmentMethodRestrictions collection, and creating allow and deny restrictions for each method inside this collection.

## Breaking changes
None. These are add-on features. If no Allow or Deny restrictions are found on a method, the method will be allowed in the same way it currently is. If no surcharges are found, nothing is different.

## Testing
- Use [these mongo dumps](https://drive.google.com/drive/u/0/folders/1CZg_abU9VBd7tKGuq8JQc6FXxrJ4oG7j) to load various shipping methods and restrictions into the database. This will update your shipping methods to include 10 flat rate methods to check the restrictions against, as well as creating a `FlatRateFulfillmentRestrictions` database, that includes various restrictions. 

Use the dumps with the following commands:

`mongorestore --db reaction --collection Shipping Shipping.bson`
`mongorestore --db reaction --collection FlatRateFulfillmentRestrictions FlatRateFulfillmentRestrictions.bson`

- Change the vendor of the Basic Reaction Product to `Reaction` to see a universal restriction

This video shows using a California address (contiguous 48 states), a Hawaii address, an Alaska address, and a New York Address, combined with the vendor being `Reaction`, which creates a universal restriction:
shipping restrictions mov

Examples that are built in to the data in the mongo dump:
Destination restrictions go down to a zip code level. Try changing the `newYorkReactionVendor` restriction to the following to only restrict certain postal codes, instead of all of New York State:
```
destination: {
  postal: [
    "10001", 10019"
  ]
}
```

This is easiest to test in the existing app, as you can add multiple addresses and easily switch between them to see the options change, but this will also work on the starterkit, with a little more effort to change the location when checking out.

There are also new GraphQL queries / mutations to `add`, `delete`, `update`, and `get` restrictions from the database.


imports/plugins/included/shipping-rates/server/no-meteor/queries/getFlatRateFulfillmentRestriction.js
```diff
@@ -0,0 +1,19 @@
...
+ * @return {Promise<Object>|undefined} - A restrictions document, if one is found
+ */
+export default async function getFlatRateFulfillmentRestriction(context, { restrictionId, shopId } = {}) {
+  const { collections } = context;
+  const { FlatRateFulfillmentRestrictions } = collections;
+
+  return FlatRateFulfillmentRestrictions.find({
+    _id: restrictionId,
+    shopId
+  });
```
aldeed: This will always return a single doc, so it should not return a cursor and the GraphQL should not be a connection. Just change `find` to `findOne` here.

imports/plugins/included/shipping-rates/server/no-meteor/resolvers/Query/getFlatRateFulfillmentRestriction.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+export default async function getFlatRateFulfillmentRestriction(parentResult, args, context) {
+  const { restrictionId, shopId, ...connectionArgs } = args;
+
+  const cursor = await context.queries.getFlatRateFulfillmentRestriction(context, {
+    restrictionId: decodeFulfillmentRestrictionOpaqueId(restrictionId),
+    shopId: decodeShopOpaqueId(shopId)
+  });
+
+  return getPaginatedResponse(cursor, connectionArgs);
```
aldeed: The query should be returning only the one document, so you can do simply `return context.queries.getFlatRateFulfillmentRestriction(/**/)` and you don't need `connectionArgs`.

The definition in the GraphQL needs to be updated to match

imports/plugins/included/shipping-rates/server/no-meteor/schemas/restrictions.graphql
```diff
@@ -0,0 +1,217 @@
+####
+# Custom data types for FlatRateFulfillmentRestriction data
+####
+type attributeRestrictions {
```
aldeed: All types, enums, inputs, etc. should start with a capital letter by GraphQL convention

imports/plugins/included/shipping-rates/server/no-meteor/schemas/restrictions.graphql
```diff
@@ -0,0 +1,217 @@
...
+  getFlatRateFulfillmentRestriction(
+    shopId: ID!
+    restrictionId: ID!
+    after: ConnectionCursor,
+    before: ConnectionCursor,
+    first: ConnectionLimitInt,
+    last: ConnectionLimitInt,
+    sortOrder: SortOrder = desc,
+    sortBy: FlatRateFulfillmentRestrictionSortByField = createdAt
+  ): FlatRateFulfillmentRestrictionConnection!
```
aldeed: This returns one document so should be

```
getFlatRateFulfillmentRestriction(restrictionId: ID!, shopId: ID!): FlatRateFulfillmentRestriction
```

- No connection args are needed because it's not a connection
- No exclamation on the response type because it might be `null` if the IDs you provide don't match anything.

imports/plugins/included/shipping-rates/server/no-meteor/schemas/restrictions.graphql
```diff
@@ -0,0 +1,217 @@
...
+
+"Defines a restriction for a single flat rate fulfillment method, or universally all flat rate fulfillment methods."
+type FlatRateFulfillmentRestriction implements Node {
+  "The restriction ID."
+  _id: ID!
+
+  "The shop ID"
+  shopId: ID!
+
+  "Method IDs to apply this restriction to."
```
aldeed: Append to the description to clarify: "If none, applies to all methods as a universal restriction."

imports/plugins/included/shipping-rates/server/no-meteor/schemas/restrictions.graphql
```diff
@@ -0,0 +1,217 @@
...
+  methodIds: [String]
+
+  """
+  The type of this restriction. Allowed types are "allow" or "deny".
+  """
+  type: restrictionTypeEnum!
+
+  "Attribute restrictions."
```
aldeed: Clarify in description whether multiple are evaluated with AND or OR, and whether they combine with destination restrictions with AND or OR

imports/plugins/included/shipping-rates/server/no-meteor/schemas/restrictions.graphql
```diff
@@ -0,0 +1,217 @@
...
+  attributes: [attributeRestrictions]
+
+  "Destination restrictions."
```
aldeed: Clarify in description whether multiple are evaluated with AND or OR, and whether they combine with attribute restrictions with AND or OR

imports/plugins/included/shipping-rates/server/no-meteor/schemas/restrictions.graphql
```diff
@@ -0,0 +1,217 @@
...
+####
+# Create restriction mutation
+####
+extend type Mutation {
+  "Create a flat rate fulfillment restriction"
+  createFlatRateFulfillmentRestriction(input: CreateFlatRateFulfillmentRestrictionInput!): CreateFlatRateFulfillmentRestrictionPayload!
+}
+
+# Inputs
+"Defines a fulfillment method that has a fixed price. This type is provided by the `flat-rate` fulfillment plugin."
```
aldeed: Seems like incorrect description here

imports/plugins/included/shipping-rates/server/no-meteor/schemas/restrictions.graphql
aldeed: Also update the descriptions in the `input` version of this type to match

imports/plugins/included/shipping-rates/server/no-meteor/schemas/restrictions.graphql
```diff
@@ -0,0 +1,217 @@
...
+}
+
+
+# Payloads
+"Response from the `CreateFlatRateFulfillmentRestriction` mutation"
+type CreateFlatRateFulfillmentRestrictionPayload {
+  "The same string you sent with the mutation params, for matching mutation calls with their responses"
+  clientMutationId: String
+
+  "The created fulfillment method"
```
aldeed: Update description

imports/plugins/included/shipping-rates/server/no-meteor/schemas/restrictions.graphql
```diff
@@ -0,0 +1,217 @@
...
+  deleteFlatRateFulfillmentRestriction(input: DeleteFlatRateFulfillmentRestrictionInput!): DeleteFlatRateFulfillmentRestrictionPayload!
+}
+
+# Inputs
+"Input for the `deleteFlatRateFulfillmentRestriction` mutation"
+input DeleteFlatRateFulfillmentRestrictionInput {
+  "An optional string identifying the mutation call, which will be returned in the response payload"
+  clientMutationId: String
+
+  "The ID of the flat rate fulfillment method you want to delete"
```
aldeed: Update description

imports/plugins/included/shipping-rates/server/no-meteor/schemas/restrictions.graphql
```diff
@@ -0,0 +1,217 @@
...
+  restrictionId: ID!
+
+  "The shop that owns the method"
```
aldeed: Update description

imports/plugins/included/shipping-rates/server/no-meteor/schemas/restrictions.graphql
```diff
@@ -0,0 +1,217 @@
...
+  shopId: ID!
+}
+
+# Payloads
+"Response from the `deleteFlatRateFulfillmentRestriction` mutation"
+type DeleteFlatRateFulfillmentRestrictionPayload {
+  "The same string you sent with the mutation params, for matching mutation calls with their responses"
+  clientMutationId: String
+
+  "The removed fulfillment method"
```
aldeed: Update description

imports/plugins/included/shipping-rates/server/no-meteor/schemas/restrictions.graphql
```diff
@@ -0,0 +1,217 @@
...
+  "An optional string identifying the mutation call, which will be returned in the response payload"
+  clientMutationId: String
+
+  "The ID of the flat rate fulfillment restriction you want to update"
+  restrictionId: ID!
+
+  "The updated restriction. Pass the whole updated restriction object without the ID."
+  restriction: FlatRateFulfillmentRestrictionInput!
+
+  "The shop that owns the method"
```
aldeed: Update description

imports/plugins/included/shipping-rates/server/no-meteor/resolvers/Mutation/updateFlatRateFulfillmentRestriction.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {String} args.input.shopId - The shop to update this flat rate fulfillment restriction for
+ * @param {String} [args.input.clientMutationId] - An optional string identifying the mutation call
+ * @param {Object} context - an object containing the per-request state
+ * @return {Promise<Object>} updateFlatRateFulfillmentRestrictionPayload
+ */
+export default async function updateFlatRateFulfillmentRestriction(parentResult, { input }, context) {
+  const { clientMutationId = null, restriction, restrictionId: opaqueRestrictionId, shopId: opaqueShopId } = input;
+
+  const shopId = decodeShopOpaqueId(opaqueShopId);
+  const restrictionId = decodeFulfillmentRestrictionOpaqueId(opaqueRestrictionId);
```
aldeed: Probably need to decode methodIds, too, right?

imports/plugins/included/shipping-rates/server/no-meteor/resolvers/Mutation/createFlatRateFulfillmentRestriction.js
```diff
@@ -0,0 +1,31 @@
...
+ * @param {Object} args.input.method - The method object
+ * @param {String} args.input.shopId - The shop to create this flat rate fulfillment method for
+ * @param {String} [args.input.clientMutationId] - An optional string identifying the mutation call
+ * @param {Object} context - an object containing the per-request state
+ * @return {Promise<Object>} CreateFlatRateFulfillmentMethodPayload
+ */
+export default async function createFlatRateFulfillmentRestriction(parentResult, { input }, context) {
+  const { clientMutationId = null, restriction, shopId: opaqueShopId } = input;
+
+  const shopId = decodeShopOpaqueId(opaqueShopId);
```
aldeed: Probably need to decode methodIds, too, right?

imports/plugins/included/shipping-rates/server/no-meteor/collections.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * @name getFlatRateFulfillmentRestrictionsCollection
+ * @summary Get shipping restrictions collection from Mongo
+ * @param {Object} context -  an object containing the per-request state
+ * @returns {Object|null} shipping restriction collection for the provided fulfillment group
+ */
+export function getFlatRateFulfillmentRestrictionsCollection(context) {
+  return context.app.db.collection("FlatRateFulfillmentRestrictions");
+}
```
aldeed: This file isn't used.

.circleci/bin/ecs-deploy-feature-branch.sh
```diff
@@ -7,8 +7,8 @@ export ENVIRONMENT=staging
 export CLUSTER=core
 # remove spaces from branch name
 export SERVICE_FEATURE=`echo $CIRCLE_BRANCH | sed 's/ //g'`
-export SERVICE=reaction-core
```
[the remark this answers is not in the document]
kieckhafer: @griggheo is creating a service to auto deploy branches that start with `feat`. This is his change pulled in. @griggheo can you confirm this is a correct change please?

imports/plugins/included/shipping-rates/server/no-meteor/util/getShippingRestrictionAttributes.js
```diff
@@ -0,0 +1,60 @@
...
+      if (tag.productId === productAndVariants.product.productId) {
+        flattenProduct.tags = tag.tags;
+      }
+      return null;
+    });
+
+    // Fetch custom attributes
+    getFunctionsOfType("addShippingRestrictionCustomAttributes").forEach((customAttributesFunc) => {
+      customAttributesFunc(flattenProduct, productAndVariants);
+    });
```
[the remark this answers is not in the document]
kieckhafer: Update the code to use a for-of, with comments on why.

imports/plugins/included/shipping-rates/server/no-meteor/util/getShippingRestrictionAttributes.js
```diff
@@ -0,0 +1,60 @@
...
+
+    products.push(flattenProduct);
+  }
+
+  return {
+    address,
+    discountTotal: summary.discountTotal.amount,
+    items: products,
+    itemTotal: summary.itemTotal.amount,
+    total: summary.total.amount
```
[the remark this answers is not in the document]
kieckhafer: Where would that different pricing data come from, the Group vs. total prices? I only see one piece of price data, the  on anything that's available in the function.

imports/plugins/core/orders/server/no-meteor/mutations/createOrder.js
```diff
@@ -261,9 +263,11 @@ export default async function createOrder(context, input) {
       workflow: { status: "new", workflow: ["coreOrderWorkflow/notStarted"] }
     };
 
+    const cart = await getCartById(context, cartId, { cartToken, throwIfNotFound: true });
+    const cartWithSummary = await xformCartCheckout(collections, cart);
     // Verify that the price for the chosen shipment method on each group matches between what the client
     // provided and what the current quote is.
-    const rates = await context.queries.getFulfillmentMethodsWithQuotes(finalGroup, context);
+    const rates = await context.queries.getFulfillmentMethodsWithQuotes(finalGroup, cartWithSummary, context);
```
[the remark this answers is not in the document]
kieckhafer: Point 1 above says that `createOrder` shouldn't know anything about carts, however `cartId` is passed into `xFormOrderGroupToCommonOrder`. Is it OK since it's just the ID and not the cart itself? It's also not manipulated or used in `xFormOrderGroupToCommonOrder`, it just gets passed through it.

imports/plugins/included/shipping-rates/server/no-meteor/getFulfillmentMethodsWithQuotes.js
```diff
@@ -73,34 +77,48 @@ export default async function getFulfillmentMethodsWithQuotes(context, fulfillme
...
-        method.handling = 0;
-      }
-      // Store shipping provider here in order to have it available in shipmentMethod
-      // for cart and order usage
-      if (!method.carrier) {
-        method.carrier = carrier;
+
+  // Get hydrated cart, an object of current order data including item and destination information
+  const hydratedCart = await getShippingRestrictionAttributes(context, cartWithSummary, fulfillmentGroup); // TODO: possibly change function name
+  const isCartShippingRestricted = await cartShippingRestricted(context, hydratedCart);
```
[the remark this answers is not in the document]
kieckhafer: We pass `hydratedCart` (renamed to `hydratedOrder`) into a few of the shipping restriction functions. They will be updated when the surcharge PR is added, i can try to see if I can move some of that info into `commorOrder` at that time.

imports/plugins/included/shipping-rates/server/no-meteor/util/filterShippingMethods.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    // Find all restrictions for this shipping method
+    const methodRestrictions = await flatRateFulfillmentRestrictionsCollection.find({ methodIds: method._id }).toArray();
+
+    // Check method against location allow check
+    const allowedMethodBasedOnShippingLocationsAllowList = await locationAllowCheck(methodRestrictions, method, hydratedOrder);
+
+    // Check method against location deny check
+    const allowedMethodsBasedOnShippingLocationsDenyList = await locationDenyCheck(methodRestrictions, method, hydratedOrder);
```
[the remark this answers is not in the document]
aldeed: Also, it will be more efficient if you exit as soon as you know it's not allowed. So I would add an early exit line after each of these, and then remove the IF that's currently around `awaitedValidShippingMethods.push`.

```js
const methodIsAllowedBasedOnShippingLocationsAllowList = await locationAllowCheck(methodRestrictions, method, hydratedOrder);
if (!methodIsAllowedBasedOnShippingLocationsAllowList) return awaitedValidShippingMethods;
// and so on
```

imports/plugins/included/shipping-rates/server/no-meteor/util/getShippingRestrictionAttributes.js
```diff
@@ -0,0 +1,64 @@
...
+  return {
+    address,
+    items: products,
+    groupDiscountTotal: totals.groupDiscountTotal.amount,
+    groupItemTotal: totals.groupItemTotal.amount,
+    groupTotal: totals.groupTotal.amount,
+    orderDiscountTotal: totals.orderDiscountTotal.amount,
+    orderItemTotal: totals.orderItemTotal.amount,
+    orderTotal: totals.orderTotal.amount
+  };
```
aldeed: As I mentioned in another comment, you can add the totals to CommonOrder. I would also keep the address property name as `shippingAddress` rather than changing to `address`. So I think what that means is that this function should essentially just alter each `commonOrder.items` and then return the `commonOrder`. It could maybe just add all props from `flattenProduct` to the order item object? I don't know if there are any conflicts there.

imports/plugins/included/shipping-rates/server/no-meteor/util/getShippingRestrictionAttributes.js
```diff
@@ -0,0 +1,64 @@
...
+ * @param {Object} context -  an object containing the per-request state
+ * @param {Object} totals - The totals object with discounts, item, and group totals
+ * @param {Object} commonOrder - details about the purchase a user wants to make.
+ * @param {Array} fulfillmentGroup.items - the items in the cart
+ * @returns {Object|null} shipping restriction attributes for the provided fulfillment group
+ */
+export default async function getShippingRestrictionAttributes(context, totals, commonOrder) {
+  const { collections, getFunctionsOfType } = context;
+  const { shippingAddress: destination, items: orderItems } = commonOrder;
+  const address = pick(destination, ["address1", "address2", "city", "country", "postal", "region"]);
```
aldeed: Is there a reason we're picking certain props from address? I'd just keep the whole `shippingAddress` as-is. Then you could also delete the `pick` function. (Should by using `_.pick` anyway)

imports/plugins/included/shipping-rates/server/no-meteor/util/helpers.js
```diff
@@ -0,0 +1,132 @@
...
+  lt(a, b) { return a < b; },
+  ne(a, b) { return a !== b; }
+};
+
+/**
+ * @constant
+ * @type {Object}
+*/
+export const propertyTypes = {
+  bool(a) { return a === "true"; },
```
aldeed: Maybe a little bit more freedom for bool conversion:

```js
bool(a) { return a.trim().toLowerCase() === "true"; }
```

It's also an eslint warning to use 1-character variable names

imports/plugins/included/shipping-rates/server/no-meteor/util/locationAllowCheck.js
```diff
@@ -0,0 +1,50 @@
...
+ * @param {Object} method - current method to check restrcictions against
+ * @param {Object} hydratedOrder - hydrated order for current order
+ * @returns {Bool} true / false as to whether method is still valid after this check
+ */
+export async function locationAllowCheck(methodRestrictions, method, hydratedOrder) {
+  // Get method specific allow restrictions
+  const allowRestrictions = methodRestrictions.filter((methodRestriction) => methodRestriction.type === "allow");
+
+  // Check to see if any restrictions for this method are destination restrictions
+  const destinationRestrictions = allowRestrictions.some((restriction) => restriction.destination !== null);
```
aldeed: Same comment as in `attributeDenyCheck` about reducing the number of times we're looping by filtering and then using the filtered array

imports/plugins/included/shipping-rates/server/no-meteor/util/locationDenyCheck.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * @summary Filter shipping methods based on per method allow location restrictions
+ * @param {Object} methodRestrictions - method restrictions from FlatRateFulfillmentRestrcitionsCollection
+ * @param {Object} method - current method to check restrcictions against
+ * @param {Object} hydratedOrder - hydrated order for current order
+ * @returns {Bool} true / false as to whether method is still valid after this check
+ */
+export async function locationDenyCheck(methodRestrictions, method, hydratedOrder) {
+  // Get method specific allow restrictions
+  const denyRestrictions = methodRestrictions.filter((methodRestriction) => methodRestriction.type === "deny");
```
aldeed: Same comment as in `attributeDenyCheck` about reducing the number of times we're looping by filtering and then using the filtered array

imports/plugins/included/shipping-rates/server/no-meteor/util/getShippingRestrictionAttributes.js
kieckhafer: @willopez please see @aldeed's question above. I do remember discussing at some point getting rid of that `pick`, but just want to make sure there isn't a reason we kept it in.

willopez: @kieckhafer that was my oversight, I was not aware that we had _.pick from lodash,  and I agree with @aldeed that the whole object should be be passed and the picking is not necessary. The custom pick function should be removed also.

imports/plugins/core/orders/server/no-meteor/mutations/createOrder.js
```diff
@@ -250,6 +387,20 @@ export default async function createOrder(context, input) {
 
   const orderId = Random.id();
 
+  // Loop through each fulfillment group, get all items and subtotals from each group and sum them up
+  const orderItemTotals = await Promise.all(fulfillmentGroups.map(async (groupInput) => {
+    const { items } = groupInput;
+
+    // Build order item for all items in fulfillment group
+    const builtItems = await Promise.all(items.map((item) => buildOrderItem(item, currencyCode, context)));
```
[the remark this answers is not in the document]
kieckhafer: After looking over this again, we moved the code that needs `orderItemTotal` into the `xformCart` and `xformOrder` functions, so we don't even need this whole block here, which takes care of the duplicate `buildOrderItems`: https://github.com/reactioncommerce/reaction/pull/4821/commits/3791e3a402acd32b5c272ea5c9ceb9efd012f692

kieckhafer:
@aldeed Round two of comments addressed, good for a look. Thanks~

imports/plugins/core/taxes/server/no-meteor/startup.js
```diff
@@ -9,6 +9,10 @@ import xformCartGroupToCommonOrder from "/imports/plugins/core/cart/server/no-me
  */
 async function getUpdatedCartItems(cart, context) {
   const taxResultsByGroup = await Promise.all(cart.shipping.map(async (group) => {
+    // Map the items onto the fulfillment groups
+    group.items = group.itemIds.map((itemId) => cart.items.find((item) => item._id === itemId));
```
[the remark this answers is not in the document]
kieckhafer: I believe we actually already do this [inside the xform](https://github.com/reactioncommerce/reaction/blob/feat-kieckhafer-shippingRestrictions/imports/plugins/core/cart/server/no-meteor/util/xformCartGroupToCommonOrder.js#L12), so I don't think we need to move it, just remove it from here.

PR #4822, opened 2018-11-21
Feat env file

rosshadden:
Impact: **minor**  
Type: **feature**

## Description
Added support for a `.env` file.

## Breaking changes
None. All this does is add additional support for a `.env` file. I did not move the existing variables from `docker-compose` over to it. I would like to and think we should, but this would mean all users would need to worry about copying/symlinking/creating a `.env` file so I chose not to make this change here or now.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,11 +1,15 @@
 REACTION_EMAIL=me@example.com
 REACTION_AUTH=somepassword
 REACTION_USER=ReactionAdmin
+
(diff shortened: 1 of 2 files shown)


aldeed:
Sorry, @rosshadden I didn't see this until after I did #4826. There was a GH issue for this, which wasn't assigned. Mine makes some additional related changes, so I'll close this one.

PR #4826, opened 2018-11-26
Update main Reaction app to use `.env` file

### LINKED ISSUE #4706: Use .env file

aldeed:
Resolves #4706   
Impact: **minor**  
Type: **feature**

## Issue
Most services that make up the Reaction platform use a `.env` file in the root of the service folder to define environment variables that should be set while running. They also have a `pre-build` script that the reaction-platform tool runs to create or update the `.env` file from a `.env.example` file, which is committed.

The Reaction API service does not have these.

## Solution
Add the `.env`-related files.

Also added two environment variables to `.env.example`, based on discussions elsewhere:

```
METEOR_DISABLE_OPTIMISTIC_CACHING=1
METEOR_WATCH_POLLING_INTERVAL_MS=10000
```

## Breaking changes
If you run Reaction locally, such as for development, you will now need to be sure there is a `.env` file with correct environment variables set in it. The `.env.example` file, with no changes, should work for most people. When running with `reaction-platform`, this should happen automatically. But if you've already been developing locally and you pull in this change, you'll need to run `bin/setup` once. You can also run `bin/setup` anytime you pull in the future, to add any new ENV variables.

## How to run locally outside of Docker
While we recommend always developing within Docker (as part of Reaction Platform) it is possible to run locally directly on your host machine (Linux or OSX). If you do this, you'll need a `.env` file with slightly different URLs. Copy `.env` to `.env.local` and update `.env.local` to look like this:

```
HYDRA_ADMIN_URL=http://localhost:4445
HYDRA_OAUTH2_INTROSPECT_URL=http://localhost:4445/oauth2/introspect
HYDRA_TOKEN_URL=http://localhost:4444/oauth2/token
MONGO_OPLOG_URL=mongodb://localhost:27017/local
MONGO_URL=mongodb://localhost:27017/reaction
OAUTH2_CLIENT_DOMAINS=http://localhost:4000
REACTION_AUTH=r3@cti0n
REACTION_EMAIL=admin@localhost
REACTION_SECURE_DEFAULT_ADMIN=false
REACTION_USER_NAME=admin
REACTION_USER=Admin
ROOT_URL=http://localhost:3000
```

We've simply changed `hydra` and `mongo` hostnames to `localhost`. Everything else is the same.

You can now run directly on your host machine, after starting `mongo` and `hydra` services in Docker, with this command:

```
export $(cat .env.local | xargs) && reaction
```

## Testing
Follow the instructions for creating the `.env`, and then make sure that the app starts, and run some simple smoke tests.


aldeed:
The one CI check failure is a known issue with feature deployment script that @griggheo is fixing separately.

.reaction/scripts/setup
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
ticean: The path for this file is different than the other projects, where it's at `bin/setup`. Is there a reason to put it in a nested directory? I'd favor a consistent path if possible.

.reaction/project-hooks/pre-build
```diff
@@ -0,0 +1,24 @@
...
+#  - It is good practice to keep this script lightweight and invoke setup
+#    scripts in your project.
+
+__current_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+__root_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+__root_name=$(basename "${__root_dir}")
+
+echo "${__root_name} post-project-start script invoked." 2>&1
+
+"${__root_dir}/.reaction/scripts/setup"
```
ticean: Another reference to the setup script path.

.reaction/scripts/setup
aldeed: @ticean Meteor eagerly loads all folders that aren't hidden, so that's why I changed it. However, as long as nobody puts a `.js` file in `bin/setup`, it probably won't cause problems. Might make Meteor a tiny bit slower having another folder to watch, but it's probably negligible, so I could try going back to bin/setup.

imports/plugins/core/core/server/Reaction/core.js
```diff
@@ -1029,10 +1020,10 @@ export default {
         \n ********************************* \n\n`);
 
     // run hooks on new user object
-    const user = Meteor.users.findOne(accountId);
+    const user = Meteor.users.findOne({ _id: userId });
```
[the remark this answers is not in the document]
aldeed: We happen to set them to the same generated ID right now so it doesn't cause problems, but I try to fix when I notice it.

aldeed:
@spencern All comments resolved

PR #4829, opened 2018-11-26
feat: shipping surcharges

kieckhafer:
Impact: **minor**  
Type: **feature**

_There is a matching PR in the starterkit to go along with this PR. It might be beneficial to test them both at the same time, as the starterkit PR requires the changes made in this PR: https://github.com/reactioncommerce/reaction-next-starterkit/pull/449_

## Issue
Shops should be able to apply surcharges based on shipping location and attributes of items inside an order (price, brand, type of item, etc). This is a similar feature, and could be seen as an extension of the work done for shipping restrictions:  #4821

## Solution
- Create both universal and per fulfillment method ways to add surcharges to an order based on the order destination, and / or attributes attached to items in an order. This is done by creating a `Surcharges` collection, and applying the surcharges to an order if criteria matches.

## Breaking changes
None. These are add-on features. If no surcharges are found, nothing is different.

## Testing

By default, surcharges can be tested against tags, and any product attribute listed [here](https://github.com/reactioncommerce/reaction/blob/feat-kieckhafer-surcharges/imports/plugins/core/cart/server/no-meteor/util/xformCartGroupToCommonOrder.js#L16-L34), any any user info such as shipping address.

### Creating / modifying surcharges
Use the following GQL queries / mutations to test `creating`, `deleting`, `updating` and `getting` a single surcharge, and `getting` all surcharges.

#### Get all surcharges for a shop
```
query {
  getSurcharges(shopId: "shopId-GOES-HERE") {
    edges {
      node {
        _id
        shopId
        message
        methodIds
        type
        amount {
          amount
          displayAmount
        }
        message(language: "LANGUAGE-CODE")
      }
    }
  }
}
```

#### Get a single surcharge
Use the `Get all surcharges for a shop` query to get the ID of a surcharge, then preform the following:
```
query {
  getSurcharge(shopId: "shopId-GOES-HERE", surchargeId: "surchargeId-GOES-HERE") {
    _id
    shopId
    message
    methodIds
    type
    amount {
      amount
      displayAmount
    }
    message(language: "LANGUAGE-CODE")
  }
}
```

#### Get a single surcharge in a non-default language (be sure to have added a surcharge message with a second language)
Use the `Get all surcharges for a shop` query to get the ID of a surcharge, then preform the following:
```
query {
  getSurcharge(shopId: "shopId-GOES-HERE", surchargeId: "surchargeId-GOES-HERE") {
    _id
    shopId
    message
    methodIds
    type
    amount {
      amount
      displayAmount
    }
    message(language: "LANGUAGE-CODE")
  }
}
```

#### Create a surcharge
```
mutation createSurcharge($input: CreateSurchargeInput!) {
  createSurcharge(input: $input) {
    clientMutationId
    surcharge {
      _id
      amount {
        amount
        displayAmount
      }
      attributes {
        property
        value
        propertyType
        operator
      }
      destination {
        region
      }
      message
      methodIds
    	type
    }
  }
}


Variables:
{
  "input": {
    "shopId": "shopId-GOES-HERE",
    "surcharge": {
      "amount": "amount": 999.99,
      "messagesByLanguage": [
        {
        	"content": "This is a message in english",
        	"language": "en"
        },{ 
          "content": "Este es un mensaje en espanol",
          "language": "es"
        }
      ],
      "methodIds": ["a","b"],
      "type": "surcharge",
      "attributes": [
        { "property": "productVendor", "value": "reaction", "propertyType": "string", "operator": "eq" },
        { "property": "variantTitle", "value": "Option 1 - Red Dwarf", "propertyType": "string", "operator": "eq" }
      ],
      "destination": { "region": ["CO", "NY"] }
    }
  }
}
```

#### Update a surcharge
```
mutation updateSurcharge($input: UpdateSurchargeInput!) {
  updateSurcharge(input: $input) {
    clientMutationId
    surcharge {
      amount {
        amount
        displayAmount
      }
      attributes {
        property
        value
        propertyType
        operator
      }
      destination {
        region
      }
      message
      methodIds
    	type
    }
  }
}


Variables:
{
  "input": {
    "shopId": "shopId-GOES-HERE",
    "surchargeId": "surchargeId-GOES-HERE"
    "surcharge": {
      "amount": 888.99,
       "messagesByLanguage": [
        {
        	"content": "This is an updated message in english",
        	"language": "en"
        },{ 
          "content": "Este es un nuevo mensaje en espanol",
          "language": "es"
        }
      ]
      "methodIds": ["a","b"],
      "type": "surcharge",
      "attributes": [
        { "property": "productVendor", "value": "reaction", "propertyType": "string", "operator": "eq" },
        { "property": "variantTitle", "value": "Option 1 - Red Dwarf", "propertyType": "string", "operator": "eq" }
      ],
      "destination": { "region": ["NJ", "TX"] }
    }
  }
}
```

#### Delete a surcharge
```
mutation deleteSurcharge($input: DeleteSurchargeInput!) {
  deleteSurcharge(input: $input) {
    clientMutationId
    surcharge {
      _id
      amount {
        amount
        displayAmount
      }
      message
      methodIds
      type
    }
  }
}


Variables:
{
  "input": {
    "shopId": "shopId-GOES-HERE",
    "surchargeId": "surchargeId-GOES-HERE"
  }
}
```

### Applying surcharges to checkout
1. Use the above instructions to create a few surcharges that will apply to different scenarios. For example: 
- - Create one that will apply with `destination: { region: ["CA"] }`, and then use a California address for your shipping address.
- - Create one that will apply with `attributes: [ { "property": "productVendor", "value": "Surcharge", "propertyType": "string", "operator": "eq" },]`, and then add a product to your cart that has a vendor of "Surcharge"
1. Add items to cart, and select a shipping method during checkout
1. See that appropriate surcharges are applied, and display in step 4 (Review) in checkout
1. Make sure you can complete checkout with surcharges applied

_Note that a surcharge must match **ALL** attributes **AND** destinations for it to apply. If you want to test just a destination, leave the attributes array empty. If you want to test just an attribute, leave the destination object empty. If you want to test both, then provide both, but note that they must both match (i.e. the product must have `title` value as `Basic Reaction Product` AND it must be shipped to `CA`)_


kieckhafer:
I'll keep this review inter-pod.

@impactmass if you could please be the primary reviewer here, with @aldeed as a secondary.

Let me know if you have any questions.

kieckhafer:
> I'm running into an issue where the second or third time I make a updateSurcharge mutation, multiple Surcharge documents get updated. It has happened twice. I'm trying to be sure this is not a fault in my testing.

That is strange, I will try to reproduce. You have to provide the ID, so it's odd it's happening, especially after the second or third time and not the first.

> Also, if I have a surcharge with destination: { region: ["CA", "OR"] } and a checkout address of 2110 Main Street, Santa Monica, CA 90405 US, I should see the surcharge applied to the cart right? Or must the surcharge doc region be one item ["CA"]?

Yes, the surcharge should apply if it meets any of the regions in the array. Take a note of the updated testing docs, it's possible that you have an attribute on that surcharge that it's not meeting as well. It must meet all attributes AND the destination for it to apply.

imports/plugins/included/surcharges/server/no-meteor/util/surchargeCheck.js
```diff
@@ -0,0 +1,77 @@
...
+ * @param {Object} extendCommonOrder - details about the purchase a user wants to make.
+ * @returns {Bool} true / false as to whether method is still valid after this check
+ */
+export async function surchargeCheck(surcharge, extendCommonOrder) {
+  const { items, shippingAddress } = extendCommonOrder;
+  const { attributes, destination } = surcharge;
+
+  const validSurcharge = items.some((item) => { // eslint-disable-line
+    // For each item, check against surcharge
+    const { country, postal, region } = destination;
```
impactmass: Testing with a surcharge with attributes, but without destination. When I tried adding to cart, there was an error that I think might be from this line. The error:

```
Cannot destructure property `country` of 'undefined' or 'null'.
```

kieckhafer:
> When I update the amount of one, I then see that both surcharges now have same regions ["NY", "NJ"]. This time, it was the first update after creating the surcharge.

I am seeing an issue... what I am seeing is that no matter what, the first Surcharge is the one that updates, and the others don't, even the one your provide the ID for.

Can you try testing one more thing to confirm... Add a 3rd / 4th / 5th surcharge, and then update the 5th one using the mutation above, and see if it's the first surcharge that updates

impactmass:
Your observation is right. I just saw the same thing with 4 surcharges. I tried updating the last, but it updated the first one created. Maybe that's the real issue I was trying to interpret

kieckhafer:
Interesting... I'm wondering if maybe the surchargeID gets obscured somewhere, so when it does the mongo update the ID is bad and it just updates the first one.

imports/plugins/included/surcharges/server/no-meteor/util/surchargeCheck.js
kieckhafer: Updated: moved the destructuring to after we check if destination exists.

impactmass:
Testing after the latest push, the queries and mutations work. I get surcharges on the cart and was able to checkout:
- When there's a matching surcharge-destinations for the shipping address 
- When there's a matching attribute for the cart item. I tested with the item `title`. From discussion with Erik, there are more fields we'd like to enable this for. So there'll be some more code updates.

kieckhafer:
@impactmass This has been updated to include a few more fields, including `productVendor`. You can retest with that field.

In addition, we've update the way the `message` field works, to handle multiple languages.

Now, the message input is called `messagesByLanguage`, and you input them as an object inside an array. They return a string of the current shop (or provided) languages translation. If no language is provided, we return the first translation in the array.

Testing docs have been updated.

kieckhafer:
Thanks @impactmass ! Working through the conflicts now, we knew there would be a few merging #4908 in. As for the `price` propertyType, I am writing up full documentation that will live in our docs.

kieckhafer:
@impactmass conflicts have been resolved. I am writing up full documentation for how to use all the different `propertyTypes` in surcharges and shipping restrictions that'll live in our developer / operator docs.

imports/plugins/included/surcharges/server/no-meteor/util/getSurchargeMessageInShopLanguage.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @name getSurchargeMessageInShopLanguage
+ * @summary Returns translated surcharge message
+ * @param {String} language Language to filter items by
+ * @param {Array} messagesByLanguage Surcharge messages by language array
+ * @return {String} Translated surcharge message, or first surcharge message if translation in
+ *  provided language is not available.
+ */
+export default function getSurchargeMessageInShopLanguage(language, messagesByLanguage) {
```
aldeed: I would remove "Shop" from this function name. It's not the shop language; it's whatever language the client asked for in the query.

imports/plugins/included/surcharges/server/no-meteor/util/getSurchargeMessageInShopLanguage.js
```diff
@@ -0,0 +1,15 @@
...
+  const translatedMessage = messagesByLanguage.find((message) => message.language === language);
+  if (translatedMessage) {
+    return translatedMessage.content;
+  }
+  return messagesByLanguage[0].content || "";
```
aldeed: This seems wrong since we don't even know what language will be the first one in the array. I would just return null (`return translatedMessage && translatedMessage.content || null;`)

imports/plugins/included/surcharges/server/no-meteor/util/helpers.js
```diff
@@ -0,0 +1,23 @@
+// TODO: EK - merge this into a central location
```
aldeed: Looks like you have a note about merging these, but they still exist in two different files.

imports/collections/schemas/core.js
```diff
@@ -0,0 +1,58 @@
...
+export const Money = new SimpleSchema({
+  amount: {
+    type: Number,
+    min: 0
+  },
+  currencyCode: String,
+  displayAmount: {
+    type: String,
+    optional: true
+  }
```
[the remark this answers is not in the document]
kieckhafer: I did not add `displayAmount`, it was already in the existing `Money` type, I just moved the `Money` schema into the `core` file so it was easily shared, as it was in either the `Cart` or `Order` before. I can update it if we think it's worth it, but I don't know if it will break one of the other places that use the `Money` schema.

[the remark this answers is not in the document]
aldeed: When I look in `develop` branch I see `Money` schema in two places and neither has `displayAmount` field.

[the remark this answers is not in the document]
kieckhafer: Hmm... Ok, trying to think back on it, I definitely moved it, maybe I did add it as well, I thought I had kept it exactly the same. This was one of the initial things I changed months ago, guess I forgot exactly what I did.

kieckhafer:
@impactmass 

> PS: Considering there's quite a number of commits here, what's the plan to resolved the DCO check?

I've tried to go back and rebase with all the commits, but ran into too many conflicts. Considering all those commits were made before we made the decision to use DCO, I was planning on manually approving DCO on this PR.

> I saw a few TODO comments. It may be better to not that in an issue somewhere so they can be revisited. TODOs have a way of hanging around in code for long.

I will update, as @aldeed mentioned this as well.

imports/plugins/core/cart/server/no-meteor/schemas/cart.graphql
```diff
@@ -174,6 +177,26 @@ type CartItem implements Node {
...
+  cartId: ID
+
+  "The fulfillmentGroupId (for reference)"
+  fulfillmentGroupID: ID
+
+  "The message to explain the surchage to customers, translated (if available) based on shop language"
+  message: String
+
+  "The surchargeId from the surchages collection (for reference)"
+  surchargeId: ID
```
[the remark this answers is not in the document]
kieckhafer: Good idea. And there should always be an ID, I'll make it required.

imports/plugins/core/orders/server/no-meteor/schemas/schema.graphql
```diff
@@ -141,6 +144,26 @@ type OrderItem implements Node {
...
+  cartId: ID
+
+  "The fulfillmentGroupId (for reference)"
+  fulfillmentGroupID: ID
+
+  "The message to explain the surchage to customers, translated (if available) based on shop language"
+  message: String
+
+  "The surchargeId from the surchages collection (for reference)"
+  surchargeId: ID
```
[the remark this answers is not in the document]
kieckhafer: https://github.com/reactioncommerce/reaction/pull/4829/commits/b804d7875f3fc50698e0efe3f4a2d261043cc8b7

imports/plugins/core/orders/server/no-meteor/schemas/schema.graphql
```diff
@@ -166,6 +189,9 @@ type OrderFulfillmentGroup implements Node {
   "A summary of the totals for this group"
   summary: OrderSummary!
 
+  "Surcharges applied to this order"
+  surcharges: [OrderSurcharge]
+
```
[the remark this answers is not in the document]
kieckhafer: > Why do we need them in both the fulfillment group and on the order itself? Are some surcharges specific to a group and others generic to the whole order?

No, I guess we don't. I was modeling this after shipping restrictions, when we did need it.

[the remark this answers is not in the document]
kieckhafer: removed from fulfillment group: https://github.com/reactioncommerce/reaction/pull/4829/commits/87a7ceb19db196f729b192c7640d00faf65498eb

imports/plugins/included/surcharges/server/no-meteor/util/getSurchargeMessageInShopLanguage.js
kieckhafer: That's fair. I modeled this after how we do it for the navigation, should we update it there as well? Or is this a different circumstance where we definitely need something to show for navigation?

imports/plugins/included/surcharges/server/no-meteor/util/helpers.js
kieckhafer: Yes... I was wondering thoughts on where we should put something like this? We use it three different places, do we have some sort of `lib` folder we use / should I create one for something like this?

imports/plugins/included/surcharges/server/no-meteor/util/getSurchargesTotalForCart.js
```diff
@@ -0,0 +1,26 @@
+import ReactionError from "@reactioncommerce/reaction-error";
+
+/**
+ * @summary Calculates total surcharge amount for a cart based on all surcharges
+ *   that have been applied to it
+ * @param {Object} context Context object
+ * @param {String} cartId Cart ID
+ * @returns {Object} Object with `surcharges` array and `total`
+ */
+export default async function getSurchargesTotalForCart(context, cartId) {
```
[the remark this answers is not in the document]
kieckhafer: So if no cart is available, should I throw an error similar to [how discounts is doing it](https://github.com/reactioncommerce/reaction/blob/feat-kieckhafer-surcharges/imports/plugins/core/discounts/server/no-meteor/util/getDiscountsTotalForCart.js#L14-L17)?

I'm not following how we would attach surcharges if no cart is available.

imports/plugins/included/surcharges/server/no-meteor/schemas/schema.graphql
```diff
@@ -0,0 +1,232 @@
...
+  type: surchargeTypeEnum!
+
+  "Attribute restrictions."
+  attributes: [surchargeAttributeRestrictionsInput]
+
+  "Destination restrictions."
+  destination: surchargeDestinationRestrictionsInput
+
+  "Amount."
+  amount: MoneyInput!
```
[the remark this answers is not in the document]
kieckhafer: So you're suggesting updating the mongo schema to use `Number`, and updating this to use `Float!`, is that correct?

imports/plugins/included/surcharges/server/no-meteor/getSurcharges.js
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * @summary Returns a list of surcharges to apply based on the cart.
+ * @param {Object} context - Context
+ * @param {Object} cart - the user's cart
+ * @return {Array} - an array that surcharges to apply to cart / order
+ * @private
+ */
+export default async function getSurcharges(context, { cart }) {
+  const { Surcharges } = context.collections;
+  const surcharges = await Surcharges.find({}).toArray();
```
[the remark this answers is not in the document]
kieckhafer: `map` only works with `toArray()` added as well... `await Surcharges.find({}).map((surcharge) => surcharge).toArray()`. Will this produce the same memory issue?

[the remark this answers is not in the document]
kieckhafer: I ended up going with the `forEach` based on seeing that's the method we use other places in the app

imports/plugins/included/surcharges/server/no-meteor/util/getSurchargesTotalForCart.js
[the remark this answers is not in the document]
aldeed: @kieckhafer Don't look at the way discounts is done because that was a hack to keep it working for now.

You should think of an order as its own thing that you calculate surcharges for. Actually the primary thing we calculate surcharges for. It helps to remember that when "orders" is someday its own microservice, it probably won't even be aware of what a cart is. If your system does not have the concept of a cart, you could remove cart plugin and still place orders.

In practical terms, it would be similar to `addShipmentMethodToGroup` and `addTaxesToGroup` functions in `placeOrder`. Add a new similar function, and in it you will convert order to commonOrder, and then call `await getSurcharges(context, { commonOrder })`. Tweak the existing `getSurcharges` function to take `commonOrder` rather than `cart`, making it useable for either cart or order. You can convert cart to commonOrder in the afterCartUpdate function.

Also, though, `placeOrder.js` should not import anything that's in `included` because the plugin might be removed. So you can instead have the surcharges plugin register a function of type "getSurcharges" and use `getFunctionsOfType` to run all such functions and merge the resulting arrays.

imports/plugins/included/surcharges/server/no-meteor/schemas/schema.graphql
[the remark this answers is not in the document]
aldeed: Yes, and add resolvers so that it's still in proper `Money` format when you query it after saving.

imports/plugins/included/surcharges/server/no-meteor/util/getSurchargeMessageInShopLanguage.js
aldeed: Navigation seems less clear-cut than this, but I'd lean toward that returning `null`, too.

I also think it would be much more straightforward and versatile to move the `language` argument to be directly on the `message` field. Then you don't need to attach it to the surcharge object in the query resolver.

aldeed: To be clear, I mean you'd query like this instead:

```
query {
  getSurcharge(shopId: "shopId-GOES-HERE", surchargeId: "surchargeId-GOES-HERE") {
    _id
    shopId
    message(language: "en")
  }
}
```

imports/plugins/included/surcharges/server/no-meteor/util/getSurchargesTotalForCart.js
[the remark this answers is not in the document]
kieckhafer: Should this be done inside the map of each `fulfillmentGroup`, or should it be outside of this since the surcharge applies to the order?

https://github.com/reactioncommerce/reaction/blob/feat-kieckhafer-surcharges/imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js#L466

If outside, what would we pass to the `xformOrderGroupToCommonOrder` function as the `group`?

aldeed:
@kieckhafer I'm taking a look now, but can you merge develop in one more time?

imports/collections/schemas/core.js
```diff
@@ -0,0 +1,54 @@
...
+ * @type {SimpleSchema}
+ * @property {String} _id
+ * @property {Money} amount
+ * @property {String} cartId optional
+ * @property {String} fulfillmentGroupId optional
+ * @property {String} messagesByLanguage optional
+ * @property {String} reason optional
+ * @property {String} surchargeId optional
+ */
+export const Surcharge = new SimpleSchema({
```
aldeed: Should we call this `AppliedSurcharge` to match the GraphQL schema and to be clearer?

imports/plugins/included/surcharges/server/no-meteor/schemas/schema.graphql
```diff
@@ -0,0 +1,286 @@
...
+  "The updated surcharge. Pass the whole updated surcharge object without the ID."
+  surcharge: SurchargeInput!
+}
+
+
+####
+# Queries
+####
+extend type Query {
+  "Get the full list of flat rate fulfillment method restrictions."
```
aldeed: This should say "of surcharges"

imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
```diff
@@ -488,7 +495,31 @@ export default async function placeOrder(context, input) {
...
+      collections,
+      currencyCode,
+      group: finalGroup,
+      orderId,
+      discountTotal
+    });
+
+    for (const func of getFunctionsOfType("getSurcharges")) {
+      const appliedSurcharges = await func(context, { commonOrder }); // eslint-disable-line
+      appliedSurcharges.forEach((appliedSurcharge) => {
```
aldeed: Can you set `appliedSurcharge.fulfillmentGroupId` before pushing? It looks like the schema already allows it, and it would be good to be able to see which group each is for when looking at the array on `order`

imports/plugins/included/surcharges/server/no-meteor/schemas/schema.graphql
```diff
@@ -0,0 +1,286 @@
...
+
+extend type Mutation {
+  "Delete a flat rate fulfillment restriction"
+  deleteSurcharge(input: DeleteSurchargeInput!): DeleteSurchargePayload!
+}
+
+extend type Mutation {
+  "Update a flat rate fulfillment surcharge"
+  updateSurcharge(input: UpdateSurchargeInput!): UpdateSurchargePayload!
+}
```
aldeed: All three of these fields can be wrapped in a single `extend type Mutation` block

imports/plugins/included/surcharges/server/no-meteor/startup.js
```diff
@@ -0,0 +1,47 @@
...
+ * @summary Called on startup
+ * @param {Object} context Startup context
+ * @returns {undefined}
+ */
+export default function startup(context) {
+  const { appEvents, collections } = context;
+  const { Cart } = collections;
+
+  // Adds Surcharges collection to context
+  collections.Surcharges = context.app.db.collection("Surcharges");
```
aldeed: Did you look at the queries being made to see if you should call `collectionIndex` here to add some?

imports/plugins/included/surcharges/server/no-meteor/xforms/xformSurchargeMessage.js
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * @name xformSurchargeMessage
+ * @summary Returns surcharge message in specified language
+ * @param {String} language Language to filter items by
+ * @param {Array} messagesByLanguage Array to check language against
+ * @return {String} Translated message to return to client
+ */
+export default async function xformSurchargeMessage(language, messagesByLanguage) {
+  const translatedMessage = getSurchargeMessageForLanguage(language, messagesByLanguage);
```
aldeed: Looks like `xformSurchargeMessage` args and return value are identical to `getSurchargeMessageForLanguage` and so providing no additional value. You could delete it and call `getSurchargeMessageForLanguage` directly.

imports/plugins/included/shipping-rates/server/no-meteor/util/isShippingRestricted.js
```diff
@@ -1,4 +1,4 @@
-import { operators, propertyTypes } from "./helpers";
+import { operators, propertyTypes } from "/lib/api";
```
aldeed: Should be from `"/imports/utils"` right?

lib/collections/transform/cartOrder.js
```diff
@@ -166,7 +184,8 @@ export const cartOrderTransform = {
     const taxes = parseFloat(this.getTaxTotal());
     const discount = parseFloat(this.getDiscounts());
     const discountTotal = Math.max(0, subTotal - discount);
-    const total = discountTotal + shipping + taxes;
+    const surcharges = parseFloat(this.getSurcharges());
+    const total = discountTotal + shipping + taxes + surcharges;
     return accounting.toFixed(total, 2);
   },
```
aldeed: As far as I can see, `getTotal` is never called anymore, so it could be deleted and `getSurcharges` isn't needed.

imports/plugins/included/surcharges/server/no-meteor/xforms/xformSurchargeMessage.js
kieckhafer: Good call. This wasn't the case originally, more was being done in xformSurchargeMessage before we updated the language schema.

lib/collections/transform/cartOrder.js
kieckhafer: `getTaxTotal()` is also unused outside of `getTotal`, so I've removed that as well.

imports/plugins/included/surcharges/server/no-meteor/startup.js
kieckhafer: It seems like `_id` might be the only index needed... I'm assuming that's probably already indexed though?

aldeed:
@kieckhafer That change looks good, but a few more things I see now (so close!):
- I forgot that we had long ago decided that we should not prefix GraphQL queries with `get`. So `getSurcharges` should be `surcharges` and `getSurcharge` should be `surchargeById`. The restriction queries actually have the same issue, but I’m not sure how we feel about changing them now. Have they been included in a release yet?
- The default sort is by `createdAt`, but I noticed that neither `createdAt` nor `updatedAt` are being set. Add `createdAt: new Date()` to the insert and `updatedAt: new Date()` to the `$set` in the update.
- I’m seeing that `methodIds` is not being respected. I created a surcharge for a single method ID, but it shows up in storefront checkout totals regardless of which fulfillment method I choose.

kieckhafer:
> * I forgot that we had long ago decided that we should not prefix GraphQL queries with `get`. So `getSurcharges` should be `surcharges` and `getSurcharge` should be `surchargeById`. The restriction queries actually have the same issue, but I’m not sure how we feel about changing them now. Have they been included in a release yet?

Updated. Restrictions were merged in ~November, they are in a release.

> * The default sort is by `createdAt`, but I noticed that neither `createdAt` nor `updatedAt` are being set. Add `createdAt: new Date()` to the insert and `updatedAt: new Date()` to the `$set` in the update.

Added

> * I’m seeing that `methodIds` is not being respected. I created a surcharge for a single method ID, but it shows up in storefront checkout totals regardless of which fulfillment method I choose.

Fixed

PR #4830, opened 2018-11-26
fix: fb appsecret vuln

spencern:
A security vulnerability was discovered which would permit an unauthenticated user to access the App Secret for the Facebook app associated with the Reaction Commerce social plugin.
This fix includes a migration to remove existing app secret data from the database.


pmn4:
thank you!

PR #4795, opened 2018-11-07
Fix identity provider login page not found due to missing permission

impactmass:
Resolves #4794  
Impact: **major**  
Type: **bugfix**

## Issue
OAuth IDP login pages created by the Hydra plugin are not showing up after recent  updates. So `/account/login` returns Not Found.

This started happening after the permission update to Package routes publication.

## Solution
Add the permission for the Hydra plugin to guest groups. This allows the route to be published for anyone visitor to the meteor app trying to login from the Starterkit. 

This should be done for plugins that add public routes.

## Breaking changes
N/A

No migration needed.

## Testing
1. Start up Starterkit (develop), Hydra (master) and Reaction API (this branch)
2. On the Starterkit, try to login (i.e click the login button)
3. You should be redirected to the Meteor app; and you should be able to see the login form (that's the fix).
4. Test this in incognito as well.

## Docs
I'll go through the docs and creating a public that adds a public route. It should say that the permission for the plugin needs to be granted to anonymous users, otherwise the route will not be accessible to non-admin users.

**Update:** In fact, there is no docs concern. Because in a post 2.0 world, our process now is that UI should be driven separately from the API (as seen in the warning message [here](https://docs.reactioncommerce.com/docs/1.16.0/swag-shop-initialization#getting-started-creating-a-plugin)

PR #4835, opened 2018-11-27
Fix: 404 on Hydra Oauth page

### LINKED ISSUE #4794: OAuth IDP Login Page returns a Not Found Page
Type: **bug|regression**

## Issue Description
OAuth IDP login pages created by the Hydra plugin are not showing up in the latest release candidate 6. So `/account/login` returns Not Found.

Going back a few commits (e.g `d37e12b2ea92`), the page shows up fine.

## Steps to Reproduce

1. Run the Starterkit latest, Reaction rc-6, and Hydra master
2. Try to login (or create account) on the Starterkit
3. You'll be shown a "Not Found" page on the Meteor app instead of Login form

## Possible Solution
My guess is that a new update to the Router is making it not find the needed template/component. (or maybe caused by the latest released patches?)

### Versions
Node: 8.11.3
NPM: 6.3.0
Meteor Node: 8.11.4
Meteor NPM: 6.4.1
Reaction CLI: 0.29.0
Reaction: 2.0.0-rc.6
Reaction branch: release-2.0.0-rc.6

dancastellon:
Resolves #4794
Impact: **major**
Type: **bugfix**

## Issue Description
OAuth IDP login pages created by the Hydra plugin are not showing up in the latest release candidate 6. So `/account/login` returns Not Found. Going back a few commits (e.g `d37e12b2ea92`), the page shows up fine.

## Solution
Calling `Reaction.addRolesToGroups()` to add the "account/login" permission to guest & customer groups was the first step. This was called in an `afterCoreInit` event callback in the hydra-oauth core plugin (see @impactmass's original PR - https://github.com/reactioncommerce/reaction/pull/4795/files)

Even with that, it was taking 2 server restarts before the account/login view was working (not returning a 404). I traced this down to Migration 5 - which was deleting and recreating all groups. Because we now have a core `createGroups` function that does this before the migrations run, this migration is no longer needed.

## Breaking changes
None

## Testing
1. Run `reaction reset -n`
2. Startup Reaction via the platform and confirm you can login and signup through clicking the buttons in the starterkit
3. Confirm you can still login as admin in Reaction and have correct permissions (update shop info, enable payment & shipping methods, etc)
4. Confirm you can still checkout in Starterkit & Reaction


aldeed:
Generally we never want to change a migration after it's been created. This may be a rare case when it's ok, but I'll have to give it some thought. If there's any way to just fix the migration to still do the same thing it was originally needed for, but without interfering with new groups, that would be preferable.

impactmass:
I share the same concern about not removing the migration. Flashing back, I see that the migration was added as a path for Reaction apps that existed before the introduction of Groups. I'll take a look now at how we can still keep that without the side effect we are currently getting

impactmass:
I find that: Migration 5 needs to be smarter. It currently assumes that every install it gets called on is coming from a pre-Groups version of Reaction. That is not always true (e.g in the case of this issue). 

It needs to confirm that there are no correctly setup Groups, before it tries to run. @dancastellon we can sync on this.

aldeed:
This is partly a flaw with the current migration system. It should be possible to mark the current DB version while seeding it, so that migrations that are unnecessary don't run. However, we're replacing the migration system soon so if we can work around this flaw for now, that's the best solution.

dancastellon:
@aldeed - @impactmass and I discussed what to do to keep the migration, and he just pushed a fix. Can you take another look? Thanks!

imports/plugins/core/versions/server/migrations/5_update_defaultRoles_to_groups.js
```diff
@@ -86,6 +72,7 @@ Migrations.add({
 
     // finds all accounts with a permission set and assigns them to matching group
     function updateAccountsInGroup({ shopId, permissions = [], groupId }) {
+      if (!groupId) return null;
```
aldeed: Seems like `return [];` would be better since the result is passed to `.concat` above.

impactmass: Good catch 💯

spencern:
Merging on @impactmass approval and @aldeed's review with a single suggestion.

PR #4850, opened 2018-12-04
[WIP] Graphql schema for packages

ajporlante:
Impact: **major**  
Type: **feature**

## Issue
Dashboard forms rely on Meteor `Package` subscriptions to pass data to the client. We wish to replace this by using graphql query instead.

## Solution
I created a `packageSettings` query that takes the name of the needed package that resolves to `requestedPackage.settings`. But to accomplish this, a plugin must have an extension of the `PackageSettingsData` by using `extend union` to specify the schema of the plugin's settings.

## Breaking changes
None

## Testing
To follow


ajporlante:
@aldeed @zenweasel @Akarshit I have created this PR to transition Meteor package subscriptions to graphql. Would like to request for your inputs on the approach I have taken so far before I proceed further.

imports/plugins/core/core/server/no-meteor/resolvers/Query/packageSettings.js
```diff
@@ -0,0 +1,35 @@
...
+export default async function packageSettings(_, args, context) {
+  const { name, shopId: opaqueShopId } = args;
+  const shopId = decodeShopOpaqueId(opaqueShopId);
+
+  const { collections, userHasPermission } = context;
+  const { Packages } = collections;
+
+  if (userHasPermission(["admin"], shopId) === false) {
+    throw new ReactionError("access-denied", "User does not have permission to update Bronto settings");
+  }
```
aldeed: If we want to replicate what the publication currently does (pretty sure we do), we should not throw an error here. The `public` field of `settings` has config that is intended to be public. So something more like this below:

```js
let { settings } = await Packages.findOne({ name, shopId });
if (!userHasPermission(["admin"], shopId)) {
  settings = { public: settings.public };
}
```

Also should check to make sure that `findOne` doesn't return `null`. Should throw a not-found in that case.

imports/plugins/core/core/server/no-meteor/schemas/packageSettings.graphql
```diff
@@ -0,0 +1,15 @@
+type MailData {
+  user: String!
+  password: String!
+}
+
+type CoreSettingsData {
+  mail: MailData!
+}
+
+union PackageSettingsData = CoreSettingsData
```
aldeed: Since there's a standard where everything under `public` is public, it might be better to make this an interface and make public a union. If that ends up being too weird, then what you have is fine.

```js
interface PackageSettingsData {
  "Settings that should be public"
  public: PackageSettingsPublicData
}

union PackageSettingsPublicData = JSONObject
```

aldeed:
@ajporlante So, when I originally reviewed this, I was forgetting that we at some point decided that plugins that need settings should create their own custom query and mutation for setting and getting them. You don't even necessarily need to store them in `Packages` collection; you could make a new `RadialSettings` collection for example, to keep things more separated from core.

This is because we are moving slowly toward a world in which plugins will actually be external http services and will likely not even share the same database, if they even have a database connection.

Does that approach work for you? That would essentially mean that this PR isn't necessary, but something similar could be added for a specific plugin if you need it. Which should then make it easy to define a proper schema without needing `JSONObject`.

(In general we are trying VERY hard to avoid ever using `JSONObject` anywhere.)

aldeed:
Closing. See my earlier comment.

PR #4853, opened 2018-12-04
[WIP] - Use base image directly in dev docker-compose

### LINKED ISSUE #4770: In Docker Compose, use Reaction base image directly
## Issue Description
In `docker-compose.yml`, which is used only for local development, there is

```yml
reaction:
    build:
      context: .
      target: meteor-dev
```

This builds an image from the local `Dockerfile`, running only the `meteor-dev` stage.

But this stage simply runs `npm install` and copies all development files into the image, neither of which are necessary.
- The `docker-compose` command already runs `npm install` every time you start the app.
- The `docker-compose` config already links local development files into the container, effectively overwriting what was copied for the initial image build anyway.

Instead we can do this:

```yml
reaction:
  image: reactioncommerce/base:v1.8.0-meteor
```

There are a couple good reasons to make this change:
- By not having to build a local image, first-time startup and startup after pruning local images will be faster.
- There are occasionally confusing errors caused by having files from the initial local image build still lurking in the image, in rare cases where the linked local directories do not overwrite them.

Prereq:
- One build step that is currently in the Reaction Dockerfile will need to be moved to the base Dockerfile. See https://github.com/reactioncommerce/base/issues/37

impactmass:
Resolves #4770  
Impact: **minor**  
Type: **chore**

## Issue
#4770 

## Solution
- Update reactionbase/base reactioncommerce/base#38
- Use new image in docker-compose.yml

## Breaking changes
N/A

## Testing
- Confirm that starting the app in docker works in dev
- Confirm that building an image from the Dockerfile works (this is already confirmed in the CI step https://circleci.com/gh/reactioncommerce/reaction/33891).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,8 @@ FROM reactioncommerce/base:v1.8-meteor as builder
 
 COPY --chown=node . $APP_SOURCE_DIR
 
+RUN meteor npm install
+
 RUN printf "\\n[-] Running Reaction plugin loader...\\n" \
  && reaction plugins load
 RUN printf "\\n[-] Building Meteor application...\\n" \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,9 @@
 ##############################################################################
-# meteor-dev stage - builds image for dev and used with docker-compose.yml
-##############################################################################
-FROM reactioncommerce/base:v1.8-meteor as meteor-dev
-
-LABEL maintainer="Reaction Commerce <architecture@reactioncommerce.com>"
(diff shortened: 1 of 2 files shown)


aldeed:
We are still discussing doing something similar, but closing this PR.

impactmass:
is the linked issue in this PR https://github.com/reactioncommerce/reaction/issues/4770 still valid as well, or we close that too?

aldeed:
@impactmass Still relevant. I posted an update comment in the issue.

PR #4859, opened 2018-12-10
refactor: updates to inventory counts and statuses 

### LINKED ISSUE #4337: Product inventory should update when a new order is created
## Prerequisites

## Issue Description
Currently, RC updates a product's inventory count after an admin processes an order. A product's inventory count should update when a new order is created.

## Steps to Reproduce
1. Spin up a fresh RC shop, log in as an admin and create a new product with a quantity of 2.
2. In another browser as a customer purchase 1 of the newly created product.
3. As an admin view the new product's quantity to see it still at 2
4. As an admin process the new order to completion then check the new product's quantity. It should now show 1.

## Possible Solution
N/A

### Versions
```sh
Node: 8.9.4
NPM: 5.6.0
Meteor Node: 8.9.4
Meteor NPM: 5.6.0
Reaction CLI: 0.29.0
Reaction: 1.12.1
Reaction branch: master
```

kieckhafer:
Resolves #4337 
Impact: **minor**  
Type: **refactor**

## Issue
Some inventory statuses are confusing and cause issues when trying to easily determine a current status. Inventory also does not correctly update on variants, depending on if a variant has options or not.

## Solution
This PR helps to clarify some statuses by adding new statuses, and defining older statuses better. It also fixes the issues concerning out dated inventory numbers on variants that have option children.

- Add `canBackorder` `boolean`. canBackorder represents the ability to backorder products: `inventoryPolicy === false && inventoryManagement === true`. Added to Variants and Options in the Catalog Schema and the Product Schema.

- Better define (via comments) `isBackorder` as being the current state of backorder, not the ability to backorder (which is `canBackorder`). Previously `inventoryPolicy === false && inventoryManagement === true && currentQuantity === 0`, now `canBackorder === true && inventoryAvailableToSell === 0`
- Depreciate `currentQuantity` from cart, and add `inventoryAvailableToSell` and `inventoryInStock`

- Add `inventoryAvailableToSell` inventory count to determine how many products are available for sale, i.e. not reserved. `inventoryAvailableToSell` is updated when a customer completes their order. This field is added on Options, Variants, and Product levels, in both the Catalog and the Products collections. If a `variant` or `product`, this is not an input number, but rather a number calculated based of the summed up total of its children's inventory.

- Add `inventoryInStock` in the Catalog collection. This is an existing field in the Products collection,  called `inventoryQuantity`. This is updated when an operator processes an order.

- Add calculations that automatically update the `inventoryAvailableToSell` and `inventoryInStock` numbers on Products and Variants, based on the summed inventory count of it's children. These functions are `updateParentVariantsInventoryAvailableToSellQuantity` and `updateParentVariantsInventoryInStockQuantity`. For example, `Variant A` has children `Option A` and `Option B`. In the past, we would stop updating inventory on `Variant A`, and in the UI show a client side calculation of the sum of inventory of `Option A` and `Option B`. This update calculates the sum of `Option A` and `Option B` whenever they are updated, and saves the number as an actual inventory number on `Variant A` in the database. This way we can trust the number we are provided by the data, instead of doing a client side calculation at runtime.

## Breaking changes
- Migration 51 has been added to attach `inventoryAvailableToSell` to all products / variants, to correctly calculate the numbers on parent products / variants, and to publish this data to already published Catalog items.
- `currentQuantity` has been marked with `depreciated` in the cart. This isn't a breaking change at the moment, but lays the path to remove this field and replace with `inventoryAvailableToSell` and `inventoryInStock` in the future.
- `Catalog.getVariantQuantity` and `ReactionProduct.getVariantQuantity` have been removed. Custom plugins using these methods will need to be updated. The same data returned by these methods is now on the object that was being passed into these methods as the field `inventoryQuantity` or `inventoryAvailableToSell`
- Moved `isBackorder`, `isLowQuantity`, and `isSoldOut` functions from the `catalog` plugin to the new `inventory` plugin. Custom plugins using these methods will need to update their import path.

## Testing
- Hook up to the Catalog endpoint via GraphiQL and see that `canBackorder`, `inventoryAvailableToSell` and `inventoryInStock` are present on products, variants, and options on our `catalogItemProduct` query.
You can use this GQL query to only see the relevent info and get rid of the noise in GraphiQL:
```
query {
  catalogItemProduct (slugOrId: "basic-reaction-product") {
    _id
    product {
			inventoryInStock
      inventoryAvailableToSell
      isSoldOut
      isLowQuantity
      isBackorder
      canBackorder
      variants {
        inventoryInStock
        inventoryAvailableToSell
        isSoldOut
        isLowQuantity
        isBackorder
        canBackorder
        options {
          inventoryInStock
          inventoryAvailableToSell
          isSoldOut
          isLowQuantity
          isBackorder
          canBackorder
        }
      }
    }
  }
}
```
- Add various combinations of products with only variants, only options, etc, and see that the `inventoryAvailableToSell` is correctly decremented when an order is placed.
- Process the order, and see that the same / correct updates are made to `inventoryInStock`.
** Inventory numbers will be different between the time an order is placed, and it is processed. They should be the same if there are no current orders waiting to be processed.
- When an order is cancelled BEFORE it's approved, only the `inventoryAvailableToSell` number should be adjusted.
- When an order is cancelled AFTER it's been approved, both inventory counts should be adjusted, IF the user chooses "restock" on the cancellation.
- As a customer, add `x` products to your cart and complete checkout. Go to PDP as admin, Adjust inventory via the operator UI, and see that the both inventory numbers adjust appropriately on all options, variants, and top level products.

- Hook up to the Cart endpoint via GraphiQL and see that `inventoryAvailableToSell` and `inventoryInStock` have been added.

- Migration 51 has two parts: the first adds the new inventory field to the `Products` collection product, and the second publishes those changes to _published_ catalog products. If you are running Reaction from scratch, this second part will not do anything on the first run, since the _Basic Reaction Product_ isn't published to the catalog by default. The best way to test the second part is to run the app from scratch, go to the PDP and publish the product, then go into Mongo and delete `inventoryAvailableToSell` and `inventoryInStock` fields from all products / variants / options inside the `Catalog` collection. Then, inside the `Migrations` collection, change the migration version to `50`. Then, restart the core docker container.


imports/plugins/core/inventory/server/no-meteor/utils/getVariantInventoryNotAvailableToSellQuantity.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param {Object} collections - Raw mongo collections.
+ * @return {Promise<number>} Reserved variant quantity.
+ */
+export default async function getVariantInventoryNotAvailableToSellQuantity(variant, collections) {
+  // Find orders that are new or processing
+  const orders = await collections.Orders.find({
+    "workflow.status": { $in: ["new", "coreOrderWorkflow/processing"] }
+  }).toArray();
+
+  const reservedQuantity = orders.reduce((sum, order) => {
```
kieckhafer: @aldeed 

Two questions here:

1) I use a lot of `reduce`s in here. Is this the right way to go about it, or is there a better way to filter down the  different levels arrays?

2) We discussed using the `shipping` or `item` status, instead of the order status, to determine what counts and doesn't count as a `not available` item. I've gone through the different workflows, and here are all the workflows for `order`, `shipping`, and `item`, at different steps. The `inventoryInStock` is decremented at the order step of `click approve`, even though the workflow isn't updated at that point. What do you think is the best status to use for this? @spencern what do you think?

notes

imports/plugins/core/inventory/server/no-meteor/utils/updateParentVariantsInventoryAvailableToSellQuantity.js
```diff
@@ -0,0 +1,75 @@
...
+    },
+    {
+      publish: true,
+      selector: {
+        type: "variant"
+      }
+    }
+  );
+
+  return;
```
kieckhafer: @aldeed Should we return anything here? I wasn't sure what to return.

imports/plugins/core/inventory/server/no-meteor/utils/updateParentVariantsInventoryInStockQuantity.js
```diff
@@ -0,0 +1,74 @@
...
+    },
+    {
+      publish: true,
+      selector: {
+        type: "variant"
+      }
+    }
+  );
+
+  return;
```
kieckhafer: @aldeed Should we return anything here? I wasn't sure what to return.

imports/plugins/core/orders/server/no-meteor/mutations/createOrder.js
```diff
@@ -458,7 +458,7 @@ export default async function createOrder(context, input) {
   OrderSchema.validate(order);
   await Orders.insertOne(order);
 
-  appEvents.emit("afterOrderCreate", order).catch((error) => {
+  appEvents.emit("afterOrderCreate", order, context).catch((error) => {
```
aldeed: You don't need to pass the context since it's already passed to the startup function, and that's what you're using in your inventory startup function anyway.

imports/plugins/core/ui/client/containers/inventoryBadge.js
```diff
@@ -19,9 +18,8 @@ const composer = (props, onData) => {
 
   // Admins pull variants from the Products collection
   if (Reaction.hasPermission(["createProduct"], Reaction.getShopId())) {
-    const inventoryQuantity = ReactionProduct.getVariantQuantity(variant);
     // Product collection variant
-    if (inventoryManagement && !inventoryPolicy && inventoryQuantity === 0) {
+    if (inventoryManagement && !inventoryPolicy && variant.inventoryAvailableToSell === 0) {
```
aldeed: The server does `<=` and I think you need to in case it drops below zero due to the `$inc`s

imports/plugins/core/ui/client/containers/inventoryBadge.js
```diff
@@ -30,7 +28,7 @@ const composer = (props, onData) => {
       label = "Sold Out!";
       i18nKeyLabel = "productDetail.soldOut";
     } else if (inventoryManagement) {
-      if (lowInventoryWarningThreshold >= inventoryQuantity) {
+      if (lowInventoryWarningThreshold >= variant.inventoryAvailableToSell) {
```
aldeed: The server badge code does `<=`, so this should be `>` right? Or reverse these and do `<=` to match server

imports/plugins/core/versions/server/util/convert51.js
```diff
@@ -0,0 +1,484 @@
...
+        $set: {
+          inventoryAvailableToSell: variantInventoryAvailableToSellQuantity,
+          inventoryQuantity: variantInventoryInStockQuantity
+        }
+      },
+      {
+        publish: true,
+        selector: {
+          type: "variant"
+        }
```
aldeed: Rather than `selector` option, all `update` calls in migrations should pass `bypassCollection2: true` option. Otherwise when the schema continues to change in the future, these updates will eventually throw errors.

imports/plugins/included/product-variant/components/variantForm.js
```diff
@@ -261,44 +262,84 @@ class VariantForm extends Component {
...
             type="number"
             name="inventoryQuantity"
             ref="inventoryQuantityInput"
-            value={this.props.onUpdateQuantityField(this.variant)}
+            value={this.variant.inventoryQuantity}
+            onChange={this.handleFieldChange}
+            onBlur={this.handleFieldBlur}
+            onReturnKeyDown={this.handleFieldBlur}
+            validation={this.props.validation}
+            helpText={"Inventory in stock"}
```
aldeed: Don't need `{}` around these strings

imports/plugins/included/product-variant/containers/gridItemNoticeContainer.js
```diff
@@ -23,9 +23,7 @@ const wrapComponent = (Comp) => (
 
       for (const topVariant of topVariants) {
         const inventoryThreshold = topVariant.lowInventoryWarningThreshold;
-        const inventoryQuantity = ReactionProduct.getVariantQuantity(topVariant);
-
-        if (inventoryQuantity !== 0 && inventoryThreshold >= inventoryQuantity) {
+        if (topVariant.inventoryAvailableToSell !== 0 && inventoryThreshold >= topVariant.inventoryAvailableToSell) {
```
aldeed: It should be `inventoryAvailableToSell > 0` in case it's negative.

Also it should be `inventoryThreshold > inventoryAvailableToSell` to match server, assuming server calculation is correct.

imports/plugins/core/catalog/server/no-meteor/utils/canBackorder.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @method canBackorder
+ * @summary If all the products variants have inventory policy disabled and inventory management enabled
+ * @memberof Catalog
+ * @param {Object[]} variants - Array with product variant objects
+ * @return {boolean} is backorder allowed or not for a product
+ */
+export default function canBackorder(variants) {
+  const results = variants.map((variant) => !variant.inventoryPolicy && variant.inventoryManagement);
+  return results.every((result) => result);
```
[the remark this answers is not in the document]
kieckhafer: We'll still need this for variants though to use to check if their options are backordered. Wont' need it on Product level though.

imports/plugins/core/inventory/server/no-meteor/startup.js
```diff
@@ -0,0 +1,46 @@
...
+        {
+          publish: true,
+          selector: {
+            type: "variant"
+          }
+        }
+      );
+
+      // Update `inventoryQuantity` on all parents of this variant / option
+      Promise.await(updateParentVariantsInventoryAvailableToSellQuantity(item, rawCollections));
```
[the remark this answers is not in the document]
kieckhafer: Yes, negative `inventoryAvailableToSell` is expected if an item is able to be backordered. `invenotryInStock`/`inventoryQuantity` should never go negative.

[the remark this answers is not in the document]
kieckhafer: 👍 Good solution, this is much cleaner this way

imports/plugins/core/inventory/server/no-meteor/utils/getVariantInventoryNotAvailableToSellQuantity.js
kieckhafer: @aldeed I think this should take care of what you mentioned above: https://github.com/reactioncommerce/reaction/pull/4859/commits/9c37bf6d07562ab9fe2716a8d21b7cd2ae0328dd

One other addition could be to add another status, `coreOrderItemWorkflow/removedFromInventoryQuantity`, and do `item.variantId === variant._id && item.workflow.workflow.includes("coreItemWorkflow/removedFromInventoryAvailableToSell") && !item.workflow.workflow.includes("coreItemWorkflow/removedFromInventoryQuantity")`, is that overkill though?

imports/plugins/core/inventory/server/no-meteor/startup.js
```diff
@@ -0,0 +1,221 @@
...
+    const { collections } = context;
+    const orderItems = order.shipping.reduce((list, group) => [...list, ...group.items], []);
+
+    // Create a new set of unique productIds
+    // We do this because variants might have the same productId
+    // and we don't want to update the product each time a variant is it's child
+    // we can map over the unique productIds at the end, and update each one once
+    orderItems.forEach(async (item) => {
+      // Update supplied item inventory
+      const updatedItem = await collections.Products.findOneAndUpdate(
```
aldeed: Unless you set the `returnOriginal` option to `false`, `findOneAndUpdate` default behavior is to return the original, not the updated. In this case it doesn't cause an issue, but `updatedItem` is an inaccurate variable name

aldeed: Also the return is technically an object where `value` is the document, either original or updated. So it would be better to do

```js
const { value: originalProduct } =
```

aldeed: Same comment for the other similar calls in this PR, but it doesn't seem to be anything more than a variable name issue in all places.

kieckhafer: It doesn't really matter, but do you think it might be best to have the updated document for use in the future, to expand on this, if needed: https://github.com/reactioncommerce/reaction/pull/4859/commits/09e2b4c0a9ace7f442069bf197d9222f1aba318e

aldeed: Agree that's best, but the option is called `returnOriginal` and needs to be set to `false`: http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#findOneAndUpdate

kieckhafer: Got it...updated. Was running off of the MongoDB Shell docs: https://docs.mongodb.com/manual/reference/method/db.collection.findOneAndUpdate/

PR #4869, opened 2018-12-12
feat(tags): A Tag can be active or inactive.

### LINKED ISSUE #4848: Tags: Backend - A Tag and its listing page can be marked as `inactive` and `active`
- An admin user can create a tag and a tag landing page that is not in the Nav, invisible to site maps and search engines. When a tag and its landing page are set as `inactive`, it should be invisible to site maps and search engines. Currently when an admin creates a tag, it automatically creates a route. We will no longer automatically create a route. A page has to be marked as `active`.

Front-end ticket for this feature: https://github.com/reactioncommerce/reaction/issues/4849

machikoyasuda:
Resolves #4848   
Impact: **minor**  
Type: **feature**

## Issue
An admin needs a way to make a Tag's Tag Listing Page hidden from the public. A Tag's `isActive` boolean field will allow an admin to mark a Tag Listing Page (TLP) as active, to be viewable to the public, or inactive, hidden from the public.

## Solution
- Adds a new `isActive` field: A Tag's `active` boolean field will allow an admin to mark a Tag as `true` or `false`:
- `isActive = true` - _active_
- `isActive = false` - _hidden_

Currently, the Tag's `isActive` field is not used in any logic. 

After this PR is merged, the Storefront Starter-Kit will be modified to only display TLP if the Tag's `isActive` field is set to `true`, and if a Tag has `isActive` set to `false`, the TLP will be removed from the site map and will not render.

## Database changes

- Migration: All legacy Tags will be migrated to have `isActive` set to `true`.

I've tested the migration locally:
screen shot 2018-12-12 at 3 15 20 pm
screen shot 2018-12-12 at 3 15 15 pm

## Breaking changes / Questions
* How does this affect Tag Listing Pages (TLP) that are part of the Navigation? Or should it not be related at all to NavBuilder logic? @dancastellon 

## Testing
1. Test the migration: Test that all Tags have `isActive` set
2. Test that a Tag can be set to `isActive`

### DIFF
--- a/imports/collections/schemas/tags.js
+++ b/imports/collections/schemas/tags.js
@@ -91,7 +91,7 @@ export const Tag = new SimpleSchema({
   },
   "isActive": {
     type: Boolean,
-    optional: true
+    optional: false
   }
 });
 
--- a/imports/collections/schemas/tags.js
+++ b/imports/collections/schemas/tags.js
@@ -89,7 +89,7 @@ export const Tag = new SimpleSchema({
     type: String,
     optional: true
   },
-  "active": {
+  "isActive": {
     type: Boolean,
     optional: true
   }
--- a/imports/collections/schemas/tags.js
+++ b/imports/collections/schemas/tags.js
@@ -88,6 +88,10 @@ export const Tag = new SimpleSchema({
   "displayTitle": {
     type: String,
     optional: true
+  },
+  "active": {
+    type: Boolean,
+    optional: true
   }
 });
 
--- a/imports/plugins/core/catalog/server/methods/catalog.js
+++ b/imports/plugins/core/catalog/server/methods/catalog.js
@@ -1087,7 +1087,8 @@ Meteor.methods({
 
     const newTag = {
       slug: Reaction.getSlug(tagName),
-      name: tagName
+      name: tagName,
+      isActive: true
     };
 
     const existingTag = Tags.findOne({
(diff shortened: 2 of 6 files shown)


imports/collections/schemas/tags.js
```diff
@@ -88,6 +88,10 @@ export const Tag = new SimpleSchema({
   "displayTitle": {
     type: String,
     optional: true
+  },
+  "active": {
+    type: Boolean,
+    optional: true
```
machikoyasuda: @mikemurray Should we require `active` to be set, and set it as `false` by default? And does this mean.. we have to make a migration for all past Tags and set them as `active` = `true`?

mikemurray: Yes, `active` should be required, and we would need a migration to set all previous tags to active.

mikemurray: One observation though, it seems we have a few different "enabled/visible" status fields like  `enabled`, `active`, `isVisible`, that we may need to start considering what the status of an "enabled" thing is vs something that's "hidden". Are they the same thing or different?

If nothing else, it should be `isActive` or `isEnabled`

machikoyasuda: Made the name change to `isActive` + added a Migration!

machikoyasuda:
Closing this PR because I realized we already have all of these things...

Thanks https://github.com/reactioncommerce/reaction/pull/2747

PR #4870, opened 2018-12-12
Support remote graphql schemas in plugins

focusaurus:
Resolves [reaction-feature-requests/issues/53](https://github.com/reactioncommerce/reaction-feature-requests/issues/53)

Impact: **minor**  
Type: **feature**

## Issue

Allow a plugin to link a remote graphql API into the main reaction graphql API via remote schema delegation/stitching. This capability will support our long-term architecture vision of independent-but-integrated services.

## Solution

The main implementation comes from `graphql-tools` which provides `mergeSchemas()` and other functions that support the base graphql schema delegation capability.

My solution here is to widen the feature set of our plugin `options.graphQL.schemas`. Currently, this option supports javascript strings which are GraphQL Schema Definition Language (SDL) format. This is widely used in the existing codebase. I've allowed this same plugin option to also support instances of graphql schema. This is implemented with 2 lines of duck typing. Strings are handled as before and become part of the local graphql schema. Non-strings are merged together via `mergeSchemas()`. The end result is our existing functionality still works but the new support for remote schema stitching is also available.

## Breaking changes

I believe there are no breaking changes here. Quick local testing looks like the reaction meteor app still functions fine.

If an existing plugin in the ecosystem is passing something other than a javascript string to `options.graphQL.schemas`, then this implementation would most likely be a breaking change and we would need more thorough duck-typing to handle that correctly, **or** to introduce a new plugin option property dedicated to this like `options.graphQL.remoteSchemas`, for example.

### Caveats

My test case for this is using a pre-generated schema for the remote service, and thus I can load the plugin synchronously into reaction. It is likely that enabling plugins to perform asynchronous initialization tasks would be convenient, but it's not strictly necessary for the work at hand, so I'm leaving it as a separate feature to implement, but it's worth thinking about a bit now. For example, a plugin might want to introspect the remote graphql service's schema over the network at runtime instead of generating an SDL file and bundling that with the plugin. At the moment, that configuration would not be supported.

## Testing

* Existing unit tests and lint pass
* I have a sample plugin I'm using to build this out and the remote graphql delegation is working for that
* I have manually smoke tested the reaction meteor UI is not obviously broken when a remote plugin is loaded.
* I have manually smoke tested the starter kit storefront is not obviously broken when a remote plugin is loaded.
* There is a unit test that stands up a mock remote/loopback graphql service with a single query, integrates it with TestApp, and sends a round trip graphql query.

## Docs

[Documentation pull request](https://github.com/reactioncommerce/reaction-docs/pull/718)


imports/node-app/core/createApolloServer.js
```diff
@@ -54,6 +56,12 @@ export default function createApolloServer(options = {}) {
 
       addCallMeteorMethod(context);
 
+      const localSchema = makeExecutableSchema({
+        typeDefs,
+        resolvers,
+        resolverValidationOptions
+      });
+      const fullSchema = mergeSchemas({ schemas: [localSchema, ...schemas] });
```
aldeed: Probably best to skip calling `mergeSchemas` if `schemas.length === 0`.

I would think we'd want to make both of these calls above the `expressServer.use` line, too. This function is called at the beginning of every single request that comes in, so it's only for things that need to be regenerated per-request.

focusaurus: OK making these changes and testing now. Still WIP for unit tests.

aldeed: Updated code looks good

imports/node-app/devserver/schemas.js
```diff
@@ -15,6 +15,9 @@ import paymentsExample from "/imports/plugins/included/payments-example/server/n
 import paymentsStripe from "/imports/plugins/included/payments-stripe/server/no-meteor/schemas";
 import shippingRates from "/imports/plugins/included/shipping-rates/server/no-meteor/schemas";
 
+// UNIT TEST
+import remoteGraphQL from "/imports/plugins/unit-test/remote-graphql/server/no-meteor/schemas";
```
aldeed: This will someday be the actual app, and not just for tests. In general I try to keep it working as an alternative to running the Meteor app. So it would be good if we can find a way to not import test schemas into it.

Perhaps you can extend it where it's used in the `TestApp` constructor? If you take `additionalSchemas` array as a constructor option, then you can pass that in during the test and keep it out of non-test code.

imports/node-app/core/createApolloServer.js
```diff
@@ -35,9 +35,19 @@ const resolverValidationOptions = {
 export default function createApolloServer(options = {}) {
   // the Meteor GraphQL server is an Express server
   const expressServer = express();
-
-  const { addCallMeteorMethod, context: contextFromOptions, resolvers, typeDefs } = options;
+  let typeDefs = options.typeDefs || [];
```
aldeed: Since it can now be schemas or typeDefs, maybe rename the `typeDefs` option to `schemas`? That term seems more generic.

imports/plugins/unit-test/remote-graphql/server/no-meteor/schemas.js
```diff
@@ -0,0 +1,14 @@
+import {
+  makeExecutableSchema,
+  makeRemoteExecutableSchema
+} from "graphql-tools";
+import { HttpLink } from "apollo-link-http";
+import fetch from "node-fetch";
+import { PORT } from "../../../../../../tests/mocks/mockRemoteGraphql";
+
+const graphqlUrl = `http://localhost:${PORT}/graphql`;
```
aldeed: The hard coded port and the extra complexity makes me nervous. Did you try just mocking the HTTP request with `nock` package? It is already a dependency.

We already have random issues with findFreePort when running on CI, so I'm planning to update `TestApp` to use https://www.apollographql.com/docs/apollo-server/features/testing.html#createTestClient rather than starting an actual server.

focusaurus: OK switched to `nock` and it's working fine. Awesome change. I :heart: `nock`.

imports/node-app/core/createApolloServer.js
focusaurus: OK are these options part of our external API surface? If I just make sure the reaction codebase is updated, all good?

PR #4871, opened 2018-12-12
feat: Add support for fallback tax service

### LINKED ISSUE #4817: Support a fallback tax service
Some users have requested the ability to have two tax services configured and enabled, one primary and one fallback. The fallback should be used only when the primary is down, to ensure that orders can still be placed. This is relatively easy to do:

1. Update tax settings form to allow selection of a fallback tax service. Store this in the same place where the active tax service is stored.
2. Update `getFulfillmentGroupTaxes` function to call the fallback service if the first service `calculateOrderGroupTaxes` function throws an error.

impactmass:
Resolves #4817   

## Issue:

Add ability to have two tax services configured and enabled, one primary and one fallback. The fallback should be used only when the primary is down, to ensure that orders can still be placed.

## Changes
1. Updates `Shop Tax Settings` form to allow selection of a fallback tax service. Stores the value in the package DB entry under `settings` field as `fallbackTaxServiceName`.

2. Changed previous `active`-tax service nomenclature to `primary`, to reflect the addition of a fallback

3. Changed existing `getActiveTaxServiceForShop` to `getTaxServicesForShop`. Now, getTaxServicesForShop returns both the primary and fallback tax services enabled for the shop.

4. Add tests for `getTaxServicesForShop` capturing how it should behave with different inputs

5. Update tax calculation step to use fallback config when primary returns `null`. This happens before falling back to default 0 tax (if forceZeroes is set)

## Breaking Changes
None. From my checks, the modified functions and names are all private to the core tax setup.

A migration is included for the schema field name change. An app's tax setting running previous release should work same even after getting this update.

## Testing
Requirements: Two tax services should be provided. Reaction already includes a "custom-rates" tax plugin. I made a clone of the custom rates plugin and updated parts of it to enable using it as an extra tax service. You can pull that from [here](https://github.com/impactmass/reaction-test-tax-plugin) and add it as a custom plugin.
1. Visit the Tax Admin UI. Select "Rapid" as primary tax service. Select "Custom Rates" as the fallback; then save. 
2. Configure the tax services:
    - For `custom tax`, configure a rate with values like these: `Country: US, Region: CA, Postal: 90405, Rate: 6`
    - For `rapid-tax` (i.e the custom test plugin), configure a rate with values like these: `Country: US, Region: CA, Postal: 90405, Rate: 100`
3. On the storefront app, add a product* to cart and proceed to checkout (either as guest or logged in).  [* Use a product that doesn't have a specific tax code set. More on that below]
4. Add in a shipping address. Use an address with the postal code you set for the tax service. Save. Confirm that the tax shown reflects what rate is set in the primary tax service
5. Open the `calculateOrderTaxes` function in the custom plugin. Here you will change this [line of code](https://github.com/impactmass/reaction-test-tax-plugin/blob/master/server/no-meteor/util/calculateOrderTaxes.js#L51) to make the plugin return null. This is so that the fallback tax config can kick in. After changing that, wait for the reaction app to restart.
6. Visit the storefront again. Click "change" to edit the shipping address. Change only the name and save. The tax calculation should reload. Confirm that the new tax reflects the fallback rate/setting.

## Note about Tax Codes
- We may have a situation where when editing a product, the primary tax service provided tax codes, then during a checkout the fallback might be the one providing tax calculation. This means that when Tax Codes are active, both the primary and fallback tax services must provide (and process) a similar tax code list.


Akarshit:
@impactmass So there could be a case when the tax plugin is triggered but there was no shipping entered because of which the tax-service returned `null`, will the fallback service calculate a tax in this case too?

zenweasel:
@impactmass So I pulled this down and trying to checkout using the Classic UI. I have custom rates enabled and no fallback tax rates and I'm not gettings tax in the cart.

The only thing that might be special is that I am migrating forward from a previous version. I didn't reset the db before testing (on purpose).

impactmass:
Merging this though I see a CI test error `timeout of 2000ms exceeded`. The test failure showed up after updating this branch with develop. I see the issue is happening on develop. I'll see it I can open a separate PR to fix it there.

PR #4879, opened 2018-12-14
feat(#4848) - Return only isVisible Tags, unless admin

### LINKED ISSUE #4848: Tags: Backend - A Tag and its listing page can be marked as `inactive` and `active`

machikoyasuda:
Resolves #4848   
Impact: **minor**  
Type: **feature**

## Issue
The current Tags query returns all in the collection, even if a tag's `isVisible` is set to false. To make sure the public cannot view the tag listing pages of certain tags, the Tags query should default to only returning `isVisible` tags.

## Solution
- [x] Adds a `shouldIncludeInvisible` parameter to the main query, set to `false` by default
- [x] Add a check for Admin status before `shouldIncludeInvisible` can return all tags
- Unit tests

## Breaking changes
None

## Testing
0. Mark some tags as isVisible == false
1. Test that an Admin can get all Tags, including Invisible tags
2. Test that a non-Admin can only get Visible tags
3. Test  `tag` and `tags` queries

### DIFF
--- a/imports/plugins/core/core/server/fixtures/products.js
+++ b/imports/plugins/core/core/server/fixtures/products.js
@@ -156,6 +156,7 @@ export default function () {
    * @property {String} name - `"Tag"`
    * @property {String} slug - `"tag"`
    * @property {Number} position - `_.random(0, 100000)`
+   * @property {Boolean} isActive - `true`
    * @property {Boolean} isTopLevel - `true`
    * @property {String} shopId - `getShop()._id`
    * @property {Date} createdAt - `faker.date.past()`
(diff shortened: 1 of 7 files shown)


imports/plugins/core/core/server/fixtures/products.js
```diff
@@ -166,6 +167,7 @@ export default function () {
     slug: "tag",
     position: _.random(0, 100000),
     //  relatedTagIds: [],
+    isActive: true,
```
mikemurray: Should `isActive` be removed?

machikoyasuda: Ooooops.... 😅

mikemurray:
👍

PR #4891, opened 2018-12-30
#4890 Use Request Origin to determine rootURL #multishop

### LINKED ISSUE #4890: Marketplace Shops on Different Domains
## Prerequisites
* [x] Are you running the latest version? (`master`)

## Issue Description
The Shop data model allows for shops to live on their own domains. As written, a user must have their preferences set in order to see a marketplace shop.

## Steps to Reproduce
1. Create a Marketplace Shop.
2. give this shop a named domain (requires a manually write to the DB)
3. Run your NextJS server on that named domain (suggestion on how to do that below)
4. visit named domain

expected: Marketplace Shop data
actual: Primary Shop data

## Possible Solution
Graph QL API to inspect the request's `"Origin"` header to determine `shopId`

### Versions
```
Node: 8.11.3
NPM: 6.5.0
Meteor Node: 4.8.4
Meteor NPM: 4.6.1
Reaction CLI: 0.29.0
Reaction: 2.0.0-rc.7
Reaction branch: master
Docker: 18.09.0
```

pmn4:
Resolves #4890
Impact: **minor|major**  
Type: **feature|bugfix**

## Issue
Shops have a feature which allows them to be associated with a domain. Given the new API/client relationship of Reaction v2, the API only inspects `process.env.ROOT_URL` and `request.hostname` which both refer to the API's domain, not the client's. Superceeding those values is a User preference for activeShop, but it is unclear how that would be set on the initial visit to a marketplace shop on its own domain.

## Solution
Inspect the request's `"Origin"` header if no user preference is set (although it is debatable whether that User preference should take precedence... that is, if my active shop is Marketplace Shop #1 and I visit Marketplace Shop #2, I should see Marketplace Shop #2's products).

I see that there is already [a PR](https://github.com/reactioncommerce/reaction/pull/4862/files) for falling back to the primary shop Id, so I'm sure we will have some merge-conflicts. Once approved, I will rebase against that commit

## Testing
1. Create a Marketplace Shop.
2. give this shop a named domain (requires a manually write to the DB)
3. Run your NextJS server on that named domain (suggestion on how to do that below)
4. visit named domain

expected: Marketplace Shop data
actual: Primary Shop data

### DIFF
--- a/imports/node-app/core/util/buildContext.js
+++ b/imports/node-app/core/util/buildContext.js
@@ -11,7 +11,7 @@ import getAbsoluteUrl from "/imports/plugins/core/core/server/util/getAbsoluteUr
  *   `userHasPermission` properties.
  * @param {Object} context - A context object on which to set additional context properties
  * @param {Object} request - Request object
- * @param {String} request.hostname - Hostname derived from Host or X-Forwarded-Host heaer
+ * @param {String} request.hostname - Hostname derived from Host or X-Forwarded-Host header
  * @param {Object} request.protocol - Either http or https
  * @param {Object} [request.user] - The user who authenticated this request, if applicable
  * @returns {undefined} No return
@@ -29,12 +29,16 @@ export default async function buildContext(context, request) {
     context.accountId = (account && account._id) || null;
   }
 
-  // Add the shopId for this request, either from the authenticated user's preferences or based on the ROOT_URL domain name
+  context.rootUrl = getRootUrl(request);
+  context.getAbsoluteUrl = (path) => getAbsoluteUrl(context.rootUrl, path);
+
+  // Add the shopId for this request, either from the authenticated user's
+  // preferences or based on the rootUrl domain name.
+  // *** important ***
+  //   context.rootUrl must be set
   context.shopId = await getShopIdForContext(context);
+  // TODO: fallback to primaryShopId if shopId is null
 
   // Add a curried hasPermission tied to the current user (or to no user)
   context.userHasPermission = getHasPermissionFunctionForUser(context.user);
-
-  context.rootUrl = getRootUrl(request);
-  context.getAbsoluteUrl = (path) => getAbsoluteUrl(context.rootUrl, path);
 }
--- a/imports/node-app/core/util/buildContext.test.js
+++ b/imports/node-app/core/util/buildContext.test.js
@@ -1,49 +1,158 @@
-import mockContext from "/imports/test-utils/helpers/mockContext";
+import mockContext, { resetContext } from "/imports/test-utils/helpers/mockContext";
 import buildContext from "./buildContext";
 
-const fakeUser = {
-  _id: "FAKE_BUILD_CONTEXT_USER_ID"
-};
+describe("#buildContext", () => {
+  let mockAccount;
+  let fakeUser;
 
-test("properly mutates the context object without user", async () => {
-  process.env.ROOT_URL = "http://localhost:3000";
-  const context = { collections: mockContext.collections };
-  await buildContext(context, { user: undefined });
-  expect(context).toEqual({
-    collections: mockContext.collections,
-    shopId: null,
-    user: null,
-    userHasPermission: jasmine.any(Function),
-    userId: null,
-    rootUrl: "http://localhost:3000/",
-    getAbsoluteUrl: jasmine.any(Function)
-  });
-});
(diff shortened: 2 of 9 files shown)


imports/node-app/core/util/buildContext.js
```diff
@@ -29,12 +29,16 @@ export default async function buildContext(context, request) {
     context.accountId = (account && account._id) || null;
   }
 
-  // Add the shopId for this request, either from the authenticated user's preferences or based on the ROOT_URL domain name
+  context.rootUrl = getRootUrl(request);
```
pmn4: moved these two lines up from below

imports/node-app/core/util/buildContext.test.js
```diff
@@ -1,49 +1,158 @@
-import mockContext from "/imports/test-utils/helpers/mockContext";
+import mockContext, { resetContext } from "/imports/test-utils/helpers/mockContext";
 import buildContext from "./buildContext";
 
-const fakeUser = {
```
pmn4: the diff isn't great here, I added a `describe` (made running tests in isolation easier) causing everything to be indented. new tests are at the bottom.

imports/plugins/core/accounts/server/no-meteor/getShopIdByDomain.js
```diff
@@ -1,12 +1,12 @@
 import url from "url";
 
-export default async function getShopIdByDomain(collections) {
+export default async function getShopIdByDomain(context) {
```
pmn4: I changed the parameter here to inspect the `context`, which should include the `rootURL`

imports/plugins/core/accounts/server/no-meteor/getShopIdByDomain.test.js
```diff
@@ -1,26 +1,41 @@
 import getShopIdByDomain from "./getShopIdByDomain";
-import mockContext from "/imports/test-utils/helpers/mockContext";
```
pmn4: once again, I added the `describe` block which makes it appear as though the whole file is modified.
I added a `resetContext` helper method as well, to set strings back to default values to reset mock functions

imports/plugins/core/accounts/server/no-meteor/getShopIdForContext.js
```diff
@@ -12,8 +12,15 @@ export default async function getShopIdForContext(context) {
 
   // if still not found, look up the shop by domain
   if (!shopId) {
-    shopId = await getShopIdByDomain(collections);
+    shopId = await getShopIdByDomain(context);
   }
 
+  // if (shopId) {
```
pmn4: would love opinions here. seems odd to _set_  data in a get function, but also makes sense to do it here

imports/plugins/core/core/server/util/getRootUrl.js
```diff
@@ -3,19 +3,52 @@
...
  * @returns {String} URL
  */
 export default function getRootURL(request) {
-  const { ROOT_URL } = process.env;
-  if (ROOT_URL) {
-    if (ROOT_URL.endsWith("/")) {
-      return ROOT_URL;
-    }
-    return `${ROOT_URL}/`;
+  let rootUrl = getOriginUrl(request);
```
pmn4: this is the main change in the PR

aldeed:
Thanks for the PR @pmn4 . This is something that definitely needs to be decided on and solved once and for all. I'm not sure if using origin is the correct approach, but let me clarify a few things and see if that helps you resolve.

- The goal with 2.0 / GraphQL is for there to be no server-side logic determining shop ID. Instead, every GraphQL query or mutation typically asks for `shopId` param.
- Due to the above, there should not really be `context.shop`. It's there only for backwards compatibility until we get everything into the new paradigm.
- A UI client is expected to know its shop ID in order to query. Our thought is that the ID would be taken from built-in client config (i.e., in the browser JS bundle) for a single-shop UI. And for multi-shop UI, the shop ID would be expected to be in the URL path, or a shop slug in the URL path and use that to get the ID.
- The NextJS starter app currently does a `primaryShopId` query on startup, but this was done temporarily, and most likely this will change to be done as described above (shop ID in .env)

It would be good to hear from you whether this approach makes sense in your case or if not, what specific example issues come up.

aldeed:
Closing this for now but we can discuss this in the future if it's still an issue. See my prior comments.

PR #4894, opened 2019-01-04
Restore CORS for 401s

aldeed:
Impact: **minor**  
Type: **bugfix**

## Issue
When I updated to Apollo Server 2.0, I switched to using their built-in CORS middleware. Unfortunately, I've now discovered that this only adds the CORS header for requests that make it past our auth middleware. This could be OK, but it turns out that Apollo Client behavior changes when the 401 response doesn't allow CORS, and this prevents the starterkit auto-token-refresh logic from working properly.

## Solution
Add back explicit CORS middleware, prior to the auth middleware.

## Breaking changes
None

## Testing
Try a GraphQL request with an expired or invalid token, and examine the headers of the 401 response. Verify that you see `Access-Control-Allow-Origin: *` header.

### DIFF
--- a/imports/node-app/core/createApolloServer.js
+++ b/imports/node-app/core/createApolloServer.js
@@ -1,3 +1,4 @@
+import cors from "cors";
 import express from "express";
 import { makeExecutableSchema, mergeSchemas } from "apollo-server";
 import { ApolloServer } from "apollo-server-express";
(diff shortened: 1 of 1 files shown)


aldeed:
@kieckhafer Yes I've noticed that failure happening more often, but it's not related. I'll dig in to see if we can remove or disable that test, but it doesn't need to block this PR.

kieckhafer:
Failing test is a timeout error that has been happening intermittently on various PR's, it's not isolated here, therefore it shouldn't block this PR from being merged.

PR #4895, opened 2019-01-04
Record plugin versions in DB and show in Shop panel

### LINKED ISSUE #4874: Create a standard for versioning "non-builtin" plugins
I would like to propose a change for all non-built-in plugins that they provide an additional field called `version` in `register.js` that would get written to the db with other “package” info. These would be semantic version numbers maintained by the plugin owner. It would be visible from the dashboard in the “basic” config section.

This would provide us a quick and easy way to know what version of all the plugins we have running in an environment where we don’t have quick and easy access to the code. I am proposing `register.js` rather than `package.json` because it makes it easy to parse that data and add it into the `Packages` collection at startup

There was also a suggestion that this info be included in a health check plugin. It should probably be available to anything that queries for package info

zenweasel:
Resolves #4874 
Impact: **minor**  
Type: **feature**

## Issue
Have some standard way to record versions of plugins and make them visible to an admin so that you know what code is deployed

## Solution
Plugins can add a "version" key to `register.js` and they will be stored in the db and displayed in the shop panel

## Breaking changes
None

## Testing
1. Add a new plugin with a `register.js` that looks like this:
```

Reaction.registerPackage({
  label: "Plugin-Versions",
  name: "reaction-plugin-versions",
  version: "1.0",
  icon: "fa fa-vines",
  autoEnable: true,
  registry: [{
    provides: ["dashboard"],
    label: "Plugin Versions",
    description: "Plugin Versions",
    icon: "fa fa-vines",
    template: "PluginVersionsSettings"
  }]
});
```
2. Restart reaction
3. Go to the dashboard. shop settings and click on "plugin version" and observe that the version number is available there.


imports/plugins/core/dashboard/client/components/plugin-versions.js
```diff
@@ -0,0 +1,18 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+
+
+export default class PluginVersion extends Component {
+  static propTypes = {
+    versionedPackage: PropTypes.object
```
aldeed: I recommend shape since you rely on the object structure. It should also be required since you don't check for existence.

```js
versionedPackage: PropTypes.shape({
  name: PropTypes.string.isRequired,
  version: PropTypes.string
}).isRequired
```

imports/plugins/core/dashboard/client/containers/plugin-versions.js
```diff
@@ -0,0 +1,22 @@
...
+      <div>
+        {versionedPackages.map((versionedPackage) => {
+          return <PluginVersion versionedPackage={versionedPackage} key={versionedPackage._id} />;
+        })}
+      </div>
+    );
+  }
+}
+
+export default PluginVersionsContainer;
```
aldeed: I wouldn't call this a container. Containers mean state storage and/or data hookup.

imports/plugins/core/dashboard/client/containers/plugin-versions.js
```diff
@@ -0,0 +1,22 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import PluginVersion from "../components/plugin-versions";
+
+class PluginVersionsContainer extends Component {
```
aldeed: In new code we name the file the same as the component, in the same case: `PluginVersionsContainer.js`

imports/plugins/core/dashboard/client/components/plugin-versions.js
```diff
@@ -0,0 +1,18 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+
+
+export default class PluginVersion extends Component {
```
aldeed: In new code we name the file the same as the component, in the same case: `PluginVersion.js`

imports/plugins/core/dashboard/client/templates/shop/settings/settings.js
```diff
@@ -158,12 +159,31 @@ Template.optionsShopSettings.helpers({
...
   },
 
   SitemapSettingsContainer() {
     return SitemapSettingsContainer;
+  },
+
+  PluginVersionsContainer() {
+    return PluginVersionsContainer;
+  }
+});
```
aldeed: It doesn't seem like you would need the two new functions in `optionsShopSettings` helpers. They're only used in `shopSettings` template.

aldeed:
Also there is one eslint error in `plugin-versions.js`

aldeed:
Added a version to a plugin and it showed up. 👍

PR #4903, opened 2019-01-08
feat: Add flag to enable only IDP routes

### LINKED ISSUE #4852: Meteor frontend UI 'accessible' during OAuth flow
## Prerequisites

## Issue Description
When performing OAuth login flow from a storefront using the Hydra IDP setup in the Meteor app, it is _possible_ to see other pages of the Meteor app (i.e the classic frontend). This can cause user confusion if they manage to navigate away from the login form to other pages on the Meteor app's classic UI.

For example, in the case of a startup issue where the login page can not be found, the Meteor app shows a 404 page with a nav bar that links to other pages in it's app. Clicking on the nav takes users to the classic UI (not the Storefront earlier visited).

First reported by @ticean.

## Steps to Reproduce
1. Try to login with the Starterkit (which starts the OAuth flow and takes you to the Meteor app which is serving as the OAuth IDP).
2. Once you see the login form, edit the URL (e.g change /account/login to a random string)
3. Press enter to visit that. Note that if you get a 404 page, it comes with a 

## Possible Solution
- Can other parts of the app be hidden by an ENV var configuration (either at access time or at deploy time)?

### Versions
2.0-rc-5 and newer

impactmass:
Resolves #4852  
Impact:
Type: **feature**

## Issue
See #4852

### Goals for this PR:
- Ability to run the app in a mode that allows only routes from the Identity Provider Plugin

### Solution Overview
- New ENV var `IDENTITY_PROVIDER_MODE` that is used to filter published routes

### Details
- There are three values the ENV can be set to. Here's what they do:
  - Showing all routes (no restriction) (ENV value `all`)
  - Showing only Identity provider routes (ENV value `idp-only`)
  - Showing all routes excluding identity provider routes (ENV value `exclude-idp`)
- In the client router, the code logic that creates the routes array from package registry items is now wrapped as a function, so that it can be called with different args.

## Breaking changes
None

## Testing
Note: Make sure to remove existing `reaction` container after changing env value (e.g with `docker-compose down` and then up again).
1. Set IDENTITY_PROVIDER_MODE=all in your ENV.
Once app is started, we want to confirm that all routes are available. Home page should load the products and `/account/login` should show login form.

2. Set IDENTITY_PROVIDER_MODE=idp-only in your ENV.
Once app is started, we want to confirm that only IDP routes show. Viewing `/account/login` should show login form. Other usual pages (including home page) show Not Found.

2. Set IDENTITY_PROVIDER_MODE=exclude-idp in your ENV.
Once app is started, we want to confirm that IDP routes do not show. Viewing `/account/login` should not show up. Other usual pages (including home page) should show.


imports/plugins/core/hydra-oauth/register.js
```diff
@@ -21,6 +21,17 @@ Reaction.registerPackage({
...
+    route: "/not-found",
+    name: "not-found",
+    label: "not-found",
+    meta: {
+      noAdminControls: true,
+      oauthLoginFlow: true
+    },
+    description: "Not Found Page",
+    workflow: "hydraOauthLogin",
+    template: "notFound"
```
impactmass: IDP's Not Found route definition. Here it's using the existing/default template. It can be updated to use a custom template as needed

imports/plugins/core/hydra-oauth/server/init.js
```diff
@@ -7,6 +7,6 @@ Hooks.Events.add("afterCoreInit", () => {
   Reaction.addRolesToGroups({
     allShops: true,
     groups: ["guest", "customer"],
-    roles: ["account/login"]
+    roles: ["account/login", "not-found"]
```
impactmass: Since the IDP defines the Not Found registry route, it needs to grant the access to it.

imports/plugins/core/core/server/publications/collections/packages.js
```diff
@@ -112,10 +119,19 @@ Meteor.publish("Packages", function (shopId) {
       ], Roles.GLOBAL_GROUP))) {
         options = {};
       }
+
+      const query = { shopId: myShopId };
+
+      // This is to ensure only needed Identity-provider-related routes are published
+      // The env can be one of three: "all", "idp-only", "exclude-idp". Default behavior is "all"
```
impactmass: @ticean I added you for comments about the ENV values. Is the naming good, unambiguous etc?

mikemurray:
@impactmass approved on my end. You're free to merge whenever you get the 👍 from @ticean.

PR #4908, opened 2019-01-09
Rewrite placeOrder and support multiple payments for an order

### LINKED ISSUE #4860: Add ability to have Loyalty or Gift Card payment method as second payment method
We need to be able to write a payment plugin that would store and manage loyalty rewords points that could be redeemed for dollars off an order.

This differs from a discount in that it does not lower the amount to be charged for tax purposes but is an actual payment against the order and can be used to pay the entire amount of the order.

We would also need to have it treated slightly differently than a normal payment in that a return on an order should always be credited to the Rewards first and then a tendered payment method second.

I _think_ these major changes need to be made:

1. Schema and other changes to allow more than one payment to be reflected on an order.
1. Allow multiple payment methods during checkout, and allow amount tendered to decrease balance of order and when balance reaches $0 the order is placed (as opposed to the current method of having the payment method responsible for placing the order)
1. Introduce the concept of a "store-credit" payment type and have the main core arbitrate when a cancellation/refund comes in that store-credit payment types are credited first over tendered payment methods.
1. Provide some way to communicate to a payment method that an amount has changed.

According to @aldeed there were discussions about how to handle this in the past with @spencern and @ticean and we would need to determine how to best accomplish this in the near term while remaining aligned with those plans.

aldeed:
Resolves #4860  
Impact: **breaking**  
Type: **feature**

## Changes
The primary change is that when placing an order, you can now pay with multiple payment methods. A variety of other changes needed to be made as a result.

MongoDB schema changes:
- There is no longer a `payment` property on each order fulfillment group. Rather there is a `payments` array on the order itself, and payments are not associated with any particular fulfillment group.
- `currencyCode` property is added to `invoice` for use by that resolver
- `captureErrorCode` and `captureErrorMessage` properties are added to `payment` so that the operator UI can display capture error messages
- `invoice` was previously on each fulfillment group and also on each `payment`. It's no longer on `payment` because payments no longer correspond to just one group.
- removed `payment.data.billingAddress` because it's already on `payment.address`

GraphQL schema changes:
- A new `placeOrder` GraphQL mutation supports placing an order with multiple payments for it.
- `_id` is now optional on `Address` type
- `shippingAddress` and `shop` fields added to `FulfillmentGroup` type
- `payments` field added to `Order` type
- `status` field is added to `Payment` type and is defined by new `PaymentStatus` enum
- [BREAKING] `placeOrderWithExampleIOUPayment` mutation is removed
- [BREAKING] `placeOrderWithStripeCardPayment` mutation is removed
- [BREAKING] Created `captureOrderPayments` mutation to replace the "orders/capturePayments" Meteor method. It has a similar API but allows you to pass an array of payment IDs to capture only certain payments.
- Moved some IOU Example types from core payments schema to the included plugin's schema

Meteor method changes:
- [BREAKING] `orders/approvePayment` method signature changed from `(order)` to `(orderId, paymentId)`

Refactoring:
- registration of payment methods is now handled within the `payments` plugin rather than in the core plugin's startup code.
- The "Currency exchange rates..." message that appears a bunch of times each time the operator UI loads a catalog listing page is change from "warn" level to "debug" level
- [BREAKING] The `createOrder` internal mutation function is removed
- Duplicate stripe functions into the `marketplace` plugin. That plugin should not require the `stripe` plugin.
- Improved the Stripe payment capture function to better handle errors. The "charge_already_captured" will now mark the charge as captured in the system rather than resulting in an error, and other errors appear in the operator UI so that you know why the capture didn't work.

Operator UI:
- The payment method settings for "supports" (capture, de-auth, etc.) are no longer respected and the check boxes have been removed from the operator UI
- The order summary view is rearranged to accommodate showing multiple payments and to keep similar information together.
    - Created a new `OrderPayment` React component
    - Order cancel button remains with the invoice UI, but approving, capturing, and refunding UI now appears once for each payment. Canceling the order still refunds all of the captured payments.
    - Some of the order info that was at the top has been moved to the shipping or payment sections based on what it is related to.
    - The Print button is now always visible and isn't coupled with payment buttons.

Bug fixes:
- One appEvent consumer throwing an error no longer prevents the remaining consumers from running 
- Fixed the permission check in "orders/refund/list" Meteor method

## Breaking changes
All of the individual `placeOrder*` GraphQL mutations provided by the built-in payment plugins are removed and replaced with a single `placeOrder` mutation that supports multiple payments. The `createOrder` internal mutation that these used is also gone, so this will break any custom payment method plugins.

See also any with BREAKING above.

## Testing
It's easiest to test this along with the starter kit changes. See the testing instructions in https://github.com/reactioncommerce/reaction-next-starterkit/pull/477


kieckhafer:
Migrations `24` and `32` both throw this error on startup: 
```
Error in migration 32, publishProductsToCatalog TypeError: Cannot read property 'emit' of undefined
      at Promise.asyncApply (imports/plugins/core/catalog/server/no-meteor/utils/publishProductToCatalog.js:80:15)
```

kieckhafer:
I don't think this is introduced in this PR - I will check with other branches to confirm, but just wanted to leave this as a note here: When I click the cart icon with an empty cart, i get an endless spinner:
reaction

imports/plugins/core/payments/server/no-meteor/schemas/schema.graphql
```diff
@@ -42,35 +59,42 @@ type Payment implements Node {
...
 # Use `extend enum` to add a plugin's payment method name
 "The name of a payment method, which is how payment methods are keyed"
 enum PaymentMethodName {
-  iou_example
-}
-
-"An example"
-type ExamplePaymentMethodData {
-  example: String!
+  none
```
kieckhafer: Do we not need other options for this `enum`?

imports/plugins/core/orders/client/components/orderTable.js
```diff
@@ -97,7 +98,8 @@ class OrderTable extends Component {
 
   renderOrderInfo(order) {
     const { displayMedia, moment } = this.props;
-    const { invoice } = getPaymentForCurrentShop(order);
+    const [payment] = order.payments || [];
+    const { amount } = payment || {};
```
kieckhafer: Can you explain how this is working to me? I think it's just a syntax I haven't seen before, but if this is a payment in an array, how are we extracting amount directly from payment?

aldeed: `const [payment]` takes the first payment and saves it in the `payment` const (array destructuring). Then I destructure the `amount` property off `payment`.

Ideally we shouldn't be using only the first payment anywhere, but I was trying to avoid doing too many changes in this one PR.

aldeed:
> Migrations 24 and 32 both throw this error on startup

That is something unrelated that I fixed in RC8 branch. I'll merge the fix into `develop` and this branch.

> Order cancellation is not available until all payments are approved and captured.

I thought this was always the case but maybe not. Either way, what you say makes sense. I should be able to update the cancellation method to skip refunding if the payment isn't captured yet.

> Order emails only display one payment

Good catch. I'll fix.

kieckhafer:
> I thought this was always the case but maybe not. Either way, what you say makes sense. I should be able to update the cancellation method to skip refunding if the payment isn't captured yet.

The cancellation button was hidden - you needed to click the down arrow next to the `Approve` button, but it was still available before the approval.

aldeed:
> The cancellation button was hidden

Right, but I seem to remember it didn't work for me until after approving. It just did nothing when I clicked it.

aldeed:
@kieckhafer All of the bugs you found should be fixed now.

@dancastellon I fixed sending the order emails when payments don't have a billing address, and I added a way for payment methods to indicate they don't support refunding. Add `canRefund: false` in the `paymentMethods` object in your `register.js`. The refund UI then will not appear. I think this was everything you had mentioned to me.

PR #4915, opened 2019-01-11
Replace all Hooks with appEvents

aldeed:
Impact: **breaking**  
Type: **feature**

## Changes
- Replaces all `Hooks` usage with the newer `appEvents`. This is a step on the path to heavier reliance on events as the primary communication mechanism among Reaction plugins and services.
- Also removed the `MethodHooks` feature.
- Change some `appEvents` events to emit with a different argument signature. All events now emit a single object. This makes it easier to add and remove properties without additional breaking changes in the future.

|Before (Hooks.Events) |After (appEvents) |
|---|---|
|Event Name: afterAccountsInsert<br>Arguments: userId, accountId | Event Name: afterAccountCreate<br>Arguments: { account, createdBy, isFirstOwnerAccount } |
|Event Name: afterAccountsUpdate<br>Arguments: userId, { accountId, updatedFields } | Event Name: afterAccountUpdate<br>Arguments: { account, updatedBy, updatedFields } |
|Event Name: afterAccountsRemove<br>Arguments: userId, accountId | Event Name: afterAccountDelete<br>Arguments: { account, deletedBy } |
|N/A | Event Name: afterAddUnverifiedEmailToUser<br>Arguments: { email, userId } |
|afterRemoveCatalogProduct was sometimes emitted for variants | Event Name: afterVariantSoftDelete<br>Arguments: { variant, deletedBy } |
|Event Name: afterRemoveProduct<br>Arguments: product | Event Name: afterProductSoftDelete<br>Arguments: { product, deletedBy } |
|N/A | Event Name: afterVariantUpdate<br>Arguments: { _id, field, value } |
|Event Name: afterPublishProductToCatalog<br>Arguments: product, catalogProduct | Event Name: afterPublishProductToCatalog<br>Arguments: { catalogProduct, product } |
|N/A | Event Name: afterOrderUpdate<br>Arguments: { order, updatedBy } |
|Event Name: onJobServerStart<br>Arguments: NONE | Event Name: jobServerStart<br>Arguments: NONE |
|Event Name: afterCoreInit<br>Arguments: NONE | Event Name: afterCoreInit<br>Arguments: NONE<br>**Deprecated. Do not use in new code. Put your code directly in a non-Meteor plugin startup function instead.** |
|N/A | Event Name: sendEmail<br>Arguments: { job, sendEmailCompleted, sendEmailFailed } |
|N/A | Event Name: afterShopCreate<br>Arguments: { createdBy, shop } |
|N/A | Event Name: afterMediaInsert<br>Arguments: { createdBy, mediaRecord } |
|N/A | Event Name: afterMediaUpdate<br>Arguments: { createdBy, mediaRecord } |
|N/A | Event Name: afterMediaRemove<br>Arguments: { createdBy, mediaRecord } |
|N/A | Event Name: afterOrderApprovePayment<br>Arguments: { approvedBy, order } |
|N/A | Event Name: afterOrderCancel<br>Arguments: { cancelledBy, order, returnToStock } |
|N/A | Event Name: afterOrderPaymentCapture<br>Arguments: { capturedBy, order, payment } |
|N/A | Event Name: afterOrderCreate<br>Arguments: { createdBy, order } |
|N/A | Event Name: afterNotificationCreate<br>Arguments: { createdBy, notification } |
|Event Name: afterCreateDefaultAdminUser<br>Arguments: user | No longer emitted. Instead "afterAccountCreate" is emitted and will have `isFirstOwnerAccount` set to `true`  |
|Event Name: beforeUpdateOrderWorkflow<br>Arguments: order, options | No longer emitted. |
|Event Name: afterUpdateOrderUpdateSearchRecord<br>Arguments: order | No longer emitted. Use `afterOrderUpdate`. |
|Event Name: afterInsertProduct <br>Arguments: product | No longer emitted. Was not used, but can be added back in the future if necessary. |
|Event Name: beforeUpdateCatalogProduct <br>Arguments: product, { userId, modifier } | No longer emitted. Was not used, but can be added back in the future if necessary. |
|Event Name: afterUpdateCatalogProduct <br>Arguments: productId, { modifier } | No longer emitted. Was not used, but can be added back in the future if necessary. |
|Event Name: beforeRemoveCatalogProduct <br>Arguments: product, { userId } | No longer emitted. Was not used, but can be added back in the future if necessary. |
|Event Name: afterRemoveCatalogProduct <br>Arguments: userId, productId | No longer emitted. Was not used, but can be added back in the future if necessary. |
|Event Name: beforeCoreInit <br>Arguments: NONE | No longer emitted. Was not used, but can be added back in the future if necessary. |
|Event Name: onCoreInit <br>Arguments: NONE | No longer emitted. Was not used, but can be added back in the future if necessary. |
|Event Name: onImport${collectionName} <br>Arguments: object | No longer emitted. Was not used, but can be added back in the future if necessary. |
|Event Name: onCreateUser <br>Arguments: user, options | No longer emitted. Was not used, but can be added back in the future if necessary. |
|Event Name: onLogin <br>Arguments: options | No longer emitted. Was not used, but can be added back in the future if necessary. |
|Event Name: afterSecurityInit <br>Arguments: options | No longer emitted. Was not used, but can be added back in the future if necessary. |
|Event Name: onOrderRefundCreated <br>Arguments: orderId | No longer emitted. Was not used, but can be added back in the future if necessary. Try using `afterOrderUpdate`. |
|Event Name: onOrderShipmentDelivered <br>Arguments: orderId | No longer emitted. Was not used, but can be added back in the future if necessary. Try using `afterOrderUpdate`. |
|Event Name: onOrderShipmentShipped <br>Arguments: orderId | No longer emitted. Was not used, but can be added back in the future if necessary. Try using `afterOrderUpdate`. |
|Event Name: onGenerateSitemap <br>Arguments: urls | Instead of using events, you can now register a function of type "getPageSitemapItems" to extend the sitemap. |

## Breaking changes
This does not break anything within the core and included plugins, but
- if you use community or custom plugins that depend on the `@reactioncommerce/hooks` package, you will need to update or obtain updated versions that use `context.appEvents` instead.
- if you have a plugin that uses `MethodHooks`, update it to implement those hooks a different way.
- review all `appEvents` consumed and emitted by custom plugins. Update expected and emitted arguments. See the table above.

## Testing
Look through code and test anything that looks like it might be affected by these changes, for example, anything that runs based on consuming an event.


kieckhafer:
@aldeed I took a look at the conflicted file (`imports/plugins/core/hydra-oauth/server/init.js`), but there didn't seem to be much info on what was in conflict. Perhaps it's the whole file that was deleted and then re-added here? Can you please take a look?

kieckhafer:
Came across this while searching for `Hooks`, can this line be deleted now?
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/index.js#L2

imports/plugins/included/search-mongo/server/methods/searchcollections.js
```diff
@@ -216,7 +216,7 @@ export function buildAccountSearchRecord(accountId, updatedFields) {
     // forceIndex is included to forceIndexing on startup, or when manually added
     const searchableFields = ["forceIndex", "shopId", "emails", "firstName", "lastName", "phone"];
 
-    const shouldRunIndex = updatedFields && updatedFields.some((field) => searchableFields.includes(field));
+    const shouldRunIndex = !Array.isArray(updatedFields) || updatedFields.some((field) => searchableFields.includes(field));
```
kieckhafer: Can you explain what we are doing here? It seems like we switched from checking if `updatedFields` is an array, to checking if it's not? Why?

Just trying to make sure I have it correct:
Before: If `updatedFields` exists and is an Array (`check` checks this) and includes one of the fields, `shouldRunIndex === true`

After: If `updatedFields` is not an `array` (is that possible at this point? doesn't `check` make sure it is an `array`?) or it is an array and includes one of the fields, `shouldRunIndex === true`

imports/plugins/included/search-mongo/server/hooks/search.js
```diff
@@ -51,36 +39,19 @@ Hooks.Events.add("afterUpdateOrderUpdateSearchRecord", (order) => {
...
   ProductSearch.remove({ _id: product._id });
   buildProductSearchRecord(product._id);
 });
-
-/**
- * after insert
- * @summary should fires on create new variants, on clones products/variants
- * @private
- */
-Hooks.Events.add("afterInsertProduct", (doc) => {
```
kieckhafer: Why are we getting rid of this `Hook` and not switching it to an 1appEvent1?

imports/plugins/core/orders/server/methods/updateHistory.js
```diff
@@ -18,20 +19,35 @@ export default function updateHistory(orderId, event, value) {
...
-  return Orders.update(orderId, {
+  const result = Orders.update({
+    _id: orderId
+  }, {
     $addToSet: {
       history: {
         event,
         value,
-        userId: Reaction.getUserId(),
+        userId: authUserId,
```
kieckhafer: Looks fine, just wondering for my own knowledge the reason we are setting this to a `const` instead of directly setting it here?

imports/plugins/core/core/server/startup/accounts.js
```diff
@@ -1,12 +1,12 @@
 import _ from "lodash";
-import Hooks from "@reactioncommerce/hooks";
 import Logger from "@reactioncommerce/logger";
 import Random from "@reactioncommerce/random";
 import { Meteor } from "meteor/meteor";
-import ReactionError from "@reactioncommerce/reaction-error";
 import { Accounts } from "meteor/accounts-base";
 import * as Collections from "/lib/collections";
```
kieckhafer: Should we update this to use the `rawCollections`, along with lines 160 & 162?

I also have this question on any file that imports `lib/collections`, which is a lot of them in this PR, just didn't want to overkill and comment on each on.

imports/plugins/core/accounts/server/no-meteor/mutations/addressBookAdd.js
```diff
@@ -90,13 +79,17 @@ export default async function addressBookAdd(context, address, accountUserId) {
...
-  throw new ReactionError("server-error", "Unable to add address to account");
+  const updatedAccount = await Accounts.findOne({ userId });
+  await appEvents.emit("afterAccountUpdate", {
+    updatedAccount,
+    updatedBy: userIdFromContext,
+    updatedFields: ["profile.addressBook"]
+  });
+
+  // Pull the updated address and return it
+  return updatedAccount.profile.addressBook.find((updatedAddress) => address._id === updatedAddress._id);
```
kieckhafer: Do we need to check that `updatedAccount.profile.addressBook` exists here?

imports/plugins/core/accounts/server/methods/addressBookUpdate.js
```diff
@@ -86,22 +90,14 @@ export default function addressBookUpdate(address, accountUserId, type) {
...
-      userId
-    });
-
-    // Pull the updated address and return it
-    return updatedAccount.profile.addressBook.find((updatedAddress) => address._id === updatedAddress._id);
-  }
-
-  throw new ReactionError("server-error", "Unable to update account address");
+  // Pull the updated address and return it
+  return updatedAccount.profile.addressBook.find((updatedAddress) => address._id === updatedAddress._id);
```
kieckhafer: Do we need to check that `updatedAccount.profile.addressBook` exists here?

imports/plugins/core/accounts/server/methods/addressBookRemove.js
```diff
@@ -38,17 +38,17 @@ export default function addressBookRemove(addressId, accountUserId) {
...
-  throw new ReactionError("server-error", "Unable to remove address from account");
+  const updatedAccount = Accounts.findOne({ userId });
+  Promise.await(appEvents.emit("afterAccountUpdate", {
+    updatedBy: authUserId,
+    updatedAccount
+  }));
+
+  // If the address remove was successful, then return the removed address
+  // Pull the address from the account before it was updated and return it
+  return account.profile.addressBook.find((removedAddress) => addressId === removedAddress._id);
```
kieckhafer: Do we need to check that `updatedAccount.profile.addressBook` exists here?

kieckhafer:
While not technically part of this PR, should we use this PR to update all the existing `appEvents.emit` to use an object to pass down variables?

**`afterCartUpdate`:** 
_One is already converted_:
 https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/taxes/server/no-meteor/startup.js#L79

_But many are not:_
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/accounts/server/methods/markAddressValidationBypassed.js#L22
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/methods/setPaymentAddress.js#L34
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/methods/setUserCurrency.js#L64
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/methods/unsetAddresses.js#L93
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/no-meteor/mutations/addCartItems.js#L66
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/no-meteor/mutations/reconcileCartsKeepAnonymousCart.js#L43
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/no-meteor/mutations/reconcileCartsMerge.js#L61
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/no-meteor/mutations/removeCartItems.js#L56
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/no-meteor/mutations/setEmailOnAnonymousCart.js#L41
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/no-meteor/mutations/setShippingAddressOnCart.js#L62
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/no-meteor/mutations/updateCartItemsQuantity.js#L79
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/no-meteor/startup.js#L66
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/shipping/server/no-meteor/mutations/selectFulfillmentOptionForGroup.js#L60
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/included/discount-codes/server/methods/methods.js#L140
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/included/discount-codes/server/methods/methods.js#L267

**`afterCartCreate`:**
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/no-meteor/mutations/convertAnonymousCartToNewAccountCart.js#L44
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/cart/server/no-meteor/mutations/createCart.js#L81

**`sendEmail`:**
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/core/server/Reaction/processEmailJobs.js#L88

**`afterMediaInsert`:**
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/files/server/methods.js#L33

**`afterMediaUpdate`:**
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/files/server/methods.js#L60
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/files/server/methods.js#L107

**`afterOrderApprovePayment`:**
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/orders/server/methods/approvePayment.js#L61

**`afterOrderCreate`:** 
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/core/orders/server/no-meteor/mutations/createOrder.js#L466

**`afterNotificationCreate`:**
https://github.com/reactioncommerce/reaction/blob/feat-aldeed-remove-hooks/imports/plugins/included/notifications/server/no-meteor/createNotification.js#L43

imports/plugins/included/search-mongo/server/methods/searchcollections.js
aldeed: This is one of those changes I made months ago when I started this, so I don't remember it that well. But I think I changed this intentionally because this function is called in a `afterAccountUpdate` event consumer function, and I believe sometimes those events do not pass the `updatedFields` property now, or at least we can't rely on them to. My thinking was that if we don't know which fields were updated (i.e., if `updatedFields` isn't an array), we should reindex to be safe.

You make a good point about the `check` above, though. I didn't see that. I'll make it optional so that we actually get to this point when `updatedFields` is undefined. I'll also add a comment here to explain what I just said.

imports/plugins/core/orders/server/methods/updateHistory.js
aldeed: Because it's also passed into `appEvents.emit` below and it's always better to avoid running a function twice.

imports/plugins/core/core/server/startup/accounts.js
aldeed: This file is Meteor code (assumes it's running in Fibers), so I think it's best to keep using the Meteor collections from `/lib/collections` in such code until we convert it all. You absolutely could import and use `rawCollections` instead, but then you'd need to wrap every call in `Promise.await` because all of that non-Meteor stuff uses Promises instead of Fibers. (`Promise.await` is Meteor's interface between Promises and Fibers. They're just two different ways of making async JavaScript appear to be synchronous.)

imports/plugins/core/accounts/server/no-meteor/mutations/addressBookAdd.js
aldeed: Given that we just did `Accounts.updateOne` to add an address, it's very unlikely to not exist. I think the only way it wouldn't is if another process altered the account in the milliseconds between the `updateOne` and the `findOne`.

That said, zero chance of error is better than minuscule chance of error. 😄 This code is from before we started using `findOneAndUpdate`. I'll update it to use that, and then it will be guaranteed to exist.

imports/plugins/core/accounts/server/methods/addressBookUpdate.js
aldeed: Same answer

imports/plugins/core/accounts/server/methods/addressBookRemove.js
aldeed: In this case, yes we should be checking because it's coming from `account` rather than `updatedAccount`. We should actually be checking much earlier in this function. I rewrote this a bit and pushed a commit.

aldeed:
> can this line be deleted now?

That comment might actually be referring to MethodHooks, but I don't see those being used anymore either, so I can remove those and the comment, too.

> should we use this PR to update all the existing appEvents.emit to use an object to pass down variables?

Yes! I was thinking of doing that, too. I'm still planning to go through this once more and document all of the changes in a before/after table. So I'll verify they are all using object arg at that time. I'm leaning toward also adding a way to define a SimpleSchema for each event, which we can then validate the objects. That way nobody can mess up.

aldeed:
@kieckhafer This is updated and ready for another look. Note the table I added to the PR description, which shows how each event has changed.

kieckhafer:
Most everything looks good. The only issue I came across is when an order is cancelled before the payment is processed, the inventory updates incorrectly. I do not think this is an issue related to this PR (although this is the first time I've come across the issue), and I have created a ticket to look into: https://github.com/reactioncommerce/reaction/issues/4931. My guess is this was overlooked in the multiple payments PR ( #4908 ).

PR #4925, opened 2019-01-18
(fix) Make Publish All work correctly

### LINKED ISSUE #4823: "Publish All" only publishes some
## Issue Description
When clicking on "Publish All" it seems to only seems to publish the products that are visible in the Grid.

## Steps to Reproduce
1. Add a lot a products (say 10k)
1. Click on "Publish All"
1. Observe that only about ~20 products are published (I don't know that it's actually only what is in the Grid. But it's def not *all*)

## Possible Solution
Just do a query for all products and publish those

### Versions
Release RC6

zenweasel:
Resolves #4823 
Impact: **major**  
Type: **bugfix**

## Issue
Currently the "Publish All" button only publishes the visible products. This makes it so all products are published.

## Solution
Create a `publishAllToCatalog` method and call that when there is more than one product in the scope.

## Testing
1. Create multiple products with unpublished changes (> 100)
1. Hit "Publish All"
1. Verify that all products all now in the Catalog

### DIFF
--- a/imports/plugins/core/catalog/server/no-meteor/mutations/publishAllProducts.js
+++ b/imports/plugins/core/catalog/server/no-meteor/mutations/publishAllProducts.js
@@ -1,0 +1,10 @@
+import publishProducts from "./publishProducts";
+
+
+export default async function publishAllProducts(context) {
+  const { collections } = context;
+  const { Products } = collections;
+  const allProducts = await Products.find({ type: "simple" }, { _id: 1 }).toArray();
+  const allProductIds = allProducts.map((product) => product._id);
+  return publishProducts(context, allProductIds);
+}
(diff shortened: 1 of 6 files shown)


imports/plugins/core/catalog/server/no-meteor/mutations/publishAllProducts.js
```diff
@@ -0,0 +1,10 @@
+import publishProducts from "./publishProducts";
+
+
+export default async function publishAllProducts(context) {
+  const { collections } = context;
+  const { Products } = collections;
+  const allProducts = await Products.find({ type: "simple" }, { _id: 1 }).toArray();
```
spencern: Does this fetch the entire product collection to the server and then work with the entire set in memory?

aldeed:
Why do we need this? Publishing every product would be very intensive so I don’t think we should support doing it this way. This is the kind of thing you would do with queueing, not in memory or in a method

zenweasel:
If publish all doesn't publish all, then what exactly is it supposed to do?

zenweasel:
Right now you cant even publish more than 50 products

spencern:
@zenweasel there may be (probably is) a need for publishing products as a group. This implementation seems like it may not scale to large collections of products without causing performance issues.

PR #4930, opened 2019-01-22
refactor: Update `inventoryQuantity` field to be `inventoryInStock`

kieckhafer:
Impact: **breaking|minor**  
Type: **refactor**

## Issue
A recent update to how we calculate inventory added a new field, `inventoryAvailalbeToSell`. This left us with two inventory fields, `inventoryAvailalbeToSell` and `inventoryQuantity`, in the `Products` collection.

In the `Catalog` collection, we renamed `inventoryQuantity` to `inventoryInStock`, as it was a better description of what the field actually is.

We initially left the `Products` collection alone, but feel like we should keep these field names in sync to avoid confusion.

## Solution
Rename the field `inventoryQuantity` to `inventoryInStock` in the `Products` collection.

## Breaking changes
No breaking changes in Reaction code. If you use the `inventoryQuantity` field in any custom code, it will need to be updated.

## Testing
1. Create orders with various amounts of products, and see that inventory correctly increments / decrements
1. See that Sold Out, Low Inventory badges work as they were


imports/plugins/core/catalog/server/no-meteor/utils/createCatalogProduct.js
```diff
@@ -108,7 +108,7 @@ export async function xformProduct({ collections, product, shop, variants }) {
         variantInventory = {
           canBackorder: canBackorder(variantOptions),
           inventoryAvailableToSell: variant.inventoryAvailableToSell || 0,
-          inventoryInStock: variant.inventoryQuantity || 0,
+          inventoryInStock: variant.inventoryInStock || 0,
```
aldeed: This may not be the time to correct this, but `inventoryInStock` and `inventoryAvailableToSell` are optional in the catalog GraphQL schema, and it seems to me that `null` would mean something different from `0`. If inventory isn't tracked in the product (`null`) but yet that becomes `0` in the catalog, that seems to be changing the meaning from "unknown" to "none".

So essentially I'm proposing that we remove the `|| 0` from all of these

imports/plugins/core/versions/server/migrations/56_change_inventoryQuantity_to_inventoryInStock.js
```diff
@@ -0,0 +1,55 @@
...
+      {
+        type: "variant"
+      },
+      {
+        $rename: { inventoryQuantity: "inventoryInStock" }
+      },
+      {
+        multi: true
+      }
+    );
```
aldeed: You could merge these into one update call with `{ type: { $in: ["simple", "variant"] }`

kieckhafer:
@aldeed can this be merged?

PR #4933, opened 2019-01-23
fix: tag ui bugs

### LINKED ISSUE #4921: tag manager operator UI bugs
- [ ]  Check marks in table aren't using our design system checks (http://zpl.io/bzYjeKA)
- [ ]  Edit buttons (pencil icon) are left aligned in column and should be centered
- [ ]  The visible/hidden column should have a header label = "Status"
- [ ]  Status column - Hidden status dots should be grey not red, red connotes error
- [ ]  Bulk actions button should be enabled - so the user can click it for context (before having anything selected). If they try to take an action w/out anything selected then we should show a warning error saying "No items selected"
- [ ]  The status and the action dropdown need to be consistent in their wording (ie make visible/hidden vs enable/disable)

mikemurray:
Resolves #4921  
Impact: **minor**  
Type: **bugfix**

## Issue

Fixing the following UI issues

List [TBD] some may be moved out of this PR to keep the complexity down

- [ ]  Check marks in table aren't using our design system checks (http://zpl.io/bzYjeKA)
- [ ]  Edit buttons (pencil icon) are left aligned in column and should be centered
- [ ]  The visible/hidden column should have a header label = "Status"
- [ ]  Status column - Hidden status dots should be grey not red, red connotes error
- [ ]  Bulk actions button should be enabled - so the user can click it for context (before having anything selected). If they try to take an action w/out anything selected then we should show a warning error saying "No items selected"
- [ ]  The status and the action dropdown need to be consistent in their wording (ie make visible/hidden vs enable/disable)

## Breaking changes

none.

## Testing

Ensure the following changes have been made to 

- Edit buttons (pencil icon) are center aligned in their column

### DIFF
--- a/imports/plugins/core/tags/client/components/TagDataTableColumn.js
+++ b/imports/plugins/core/tags/client/components/TagDataTableColumn.js
@@ -11,6 +11,13 @@ const Cell = styled.div`
   height: 100%;
 `;
 
+const CenteredCell = styled.div`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100%;
+`;
+
 const EnabledLabel = styled.span`
   padding-left: 8px;
 `;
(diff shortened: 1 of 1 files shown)


mikemurray:
@cassytaylor updated text for the bulk actions

image

PR #4934, opened 2019-01-24
fix: migration 56 throwing an error

kieckhafer:
Impact: **minor**  
Type: **bugfix**

## Issue
Migration 56 is throwing an error, `Error while migrating TypeError: Cannot read property 'namedContext' of null`. It seems this error stems from the issue of having two different schemas for `variant` and `simple`, causing the issue with the update.

There are various cases of this same error reported in github issues:
https://github.com/aldeed/meteor-collection2/issues/310
https://github.com/aldeed/meteor-collection2-core/issues/12

## Solution
Split the update into two different database calls.

## Breaking changes
None

## Testing
1. Start reaction from a fresh install
1. See all migrations worked

### DIFF
--- a/imports/plugins/core/versions/server/migrations/56_change_inventoryQuantity_to_inventoryInStock.js
+++ b/imports/plugins/core/versions/server/migrations/56_change_inventoryQuantity_to_inventoryInStock.js
@@ -6,23 +6,13 @@ Migrations.add({
   up() {
     collections.Products.update(
       {
-        type: "simple"
+        type: { $in: ["simple", "variant"] }
       },
       {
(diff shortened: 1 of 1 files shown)


aldeed:
@kieckhafer I would instead fix by adding `bypassCollection2: true` option to all of the `update` calls. If you look at other migrations, they do this. It's important to do this for all migrations anyway because if the schema ever changes, we do NOT want these updates to fail based on not matching the new schema. They should always and forever do the same thing because presumably there will be a later migration that handles the later schema change.

Sorry I didn't catch this when reviewing

kieckhafer:
@aldeed updated

PR #4937, opened 2019-01-25
chore Update to base image 1.8.0.2 to include Kafka binary libs

impactmass:
Type: **chore?**

## Changes
- We published an updated base image in order to add new dependencies useful for kakfa development (reactioncommerce/base#39). This update is particularly useful for data import features.
- The base image versioning is currently tied to Meteor versions, hence the new base image for 1.8.0.2
- Meteor 1.8.0.2 only contains a minor bug fix https://github.com/meteor/meteor/pull/10403 
- Meteor's Atmosphere packages also got some patch updates

## Testing
- Run basic tests in the meteor app (guest browsing, add to cart, login, checkout)
- You can merge this into a branch you're using in development and test it out for a few hours
- Confirm that the kafka-related bins (librdkafka-dev, libsasl2-dev) are present (@dancastellon I assume you know how to confirm this, since you use them)

### DIFF
--- a/.meteor/release
+++ b/.meteor/release
@@ -1,1 +1,1 @@
-METEOR@1.8
+METEOR@1.8.0.2
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -11,11 +11,11 @@ meteor-base@1.4.0             # Packages every Meteor app needs to have
 mobile-experience@1.0.5       # Packages for a great mobile UX
 blaze-html-templates@1.0.4    # Compile .html files into Meteor Blaze views
 es5-shim@4.8.0               # ECMAScript 5 compatibility for older browsers.
-ecmascript@0.12.0              # Enable ECMAScript2015+ syntax in app code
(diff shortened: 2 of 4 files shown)


.meteor/release
```diff
@@ -1 +1 @@
-METEOR@1.8
+METEOR@1.8.0.2
```
zenweasel: @impactmass did you intend to upgrade Meteor here?

impactmass:
> Technically they aren't sinek binaries, they are the Kafka binary libs that would work with any library that wraps the binaries. Sinek is the only one I know of though

Thanks, I've updated the copy for correctness.

.meteor/release
impactmass: Yes. 

Here's a line from the PR description (with some more emphasis):
> The base image versioning is currently tied to Meteor versions, hence the new base image for 1.8.0.2

The versioning plan in the `reaction/base` repo is that we cut releases with meteor versions. So the binaries needed here are in the new version cut with the updated meteor. Hence this change.

PR #4938, opened 2019-01-25
GraphQL subscriptions

aldeed:
Impact: **minor**  
Type: **feature**

## Changes
Implemented Apollo GraphQL subscriptions support, with a `tick` subscription for testing.

## Breaking changes
None

## Testing
Test all of the following:
- You can subscribe to `tick` in GraphQL Playground on `localhost:3000`
- You can subscribe to `tick` in GraphQL Playground on `localhost:3030`
- You can subscribe to `tick` in the operator UI
- You can subscribe to `tick` in the storefront UI
- When running the operator UI, there are no errors in the console about websocket handshakes
- Normal GraphQL queries and mutations continue to work as before

### Subscribe to "tick" in GraphQL Playground on localhost:3000
1. Run the Meteor API as normal.
1. After you see the "listening" message in the server logs, go to http://localhost:3000/graphql-alpha.
1. Enter `http://localhost:3000/graphql-alpha` as the URL and the following query:
    ```
    subscription {
      tick
    }
    ```
1. You should see multiple responses tick in with incrementing number.

### Subscribe to "tick" in GraphQL Playground on localhost:3030
Same instructions as above, but run the pure Node app with `npm run devserver` and test on http://localhost:3030 instead of 3000. You can run the pure Node app at the same time as the Meteor app.

Note: The pure Node app runs on your host machine so you'll need to `meteor npm install` on the host if you haven't already.

### Subscribe to "tick" in operator UI and the storefront UI
This PR configured the operator UI already. Storefront UI will need to be configured in a similar way. To test either one, you can use this same test component:

```js
import gql from "graphql-tag";
import { Subscription } from "react-apollo";

const TICK_SUBSCRIPTION = gql`
  subscription tick {
    tick
  }
`;

const SubscriptionTest = () => (
  <Subscription subscription={TICK_SUBSCRIPTION} >
    {({ data: { tick } = {}, loading }) => (
      <div>{loading ? "Loading..." : tick}</div>
    )}
  </Subscription>
);
```

Put that component anywhere in the client code, import it onto an existing page, and add `<SubscriptionTest />` to the page render. When the page loads, you should see "Loading..." followed by incrementing numbers.

### DIFF
--- a/imports/node-app/core/ReactionNodeApp.js
+++ b/imports/node-app/core/ReactionNodeApp.js
@@ -45,7 +45,8 @@ export default class ReactionNodeApp {
 
     const {
       apolloServer,
-      expressApp
+      expressApp,
+      path
     } = createApolloServer({
       addCallMeteorMethod: this.options.addCallMeteorMethod || defaultAddCallMethod,
       context: this.context,
@@ -56,10 +57,10 @@ export default class ReactionNodeApp {
 
     this.apolloServer = apolloServer;
     this.expressApp = expressApp;
+    this.graphQLPath = path;
 
-    // Wire up GraphQL subscription websocket handlers
+    // HTTP server for GraphQL subscription websocket handlers
     this.httpServer = options.httpServer || createServer(this.expressApp);
-    apolloServer.installSubscriptionHandlers(this.httpServer);
   }
 
   setMongoDatabase(db) {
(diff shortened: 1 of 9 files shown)


kieckhafer:
Works as expected, and doesn't break anything existing.

To get it working on starterkit, we had to do a few npm updates which were breaking a lot of things, but it works well after those updates. PR for those here, will merge after this is merged: https://github.com/reactioncommerce/reaction-next-starterkit/pull/492

PR #4942, opened 2019-01-28
Use stripe.setAppInfo to identify ReactionCommerce to Stripe (fixes #57)

### LINKED ISSUE #57: Cart drawer image should pull from variant, parent, first
Cart drawer images should have similar logic to the product detail page.

If variant has image display, else display pariant variant image, else display variant[0] image, else placeholder.

rattrayalex-stripe:
Resolves https://github.com/reactioncommerce/reaction-feature-requests/issues/57 
Impact: **minor**  
Type: **feature**

## Issue
https://github.com/reactioncommerce/reaction-feature-requests/issues/57

## Solution
I hardcode the name `"ReactionCommerce"` in a few places, which isn't great - do you have a suggestion for a place to reference that as a constant? 

Otherwise, I pull version and url from `package.json`, which seems to be what reaction does elsewhere as well.

## Breaking changes
none.

## Testing
none; do you have suggestions? I could write a nock test and copy/paste it in 5 places, but that doesn't feel great. I could also add a unit test for `getStripeInstanceForShop` or add a case to an existing test file, such as `stripeCapturePayment.test.js`.


rattrayalex-stripe:
Thanks so much, @aldeed !

Staring at this a bit more, I had a few thoughts: 

- It seems that the "marketplace" plugin and the "payments-stripe" plugin are fairly distinct. We should probably identify them as such (we can implement logic on our side to classify "ReactionCommerceMarketplace" as a derivative of "ReactionCommerce"). 
- Thus, probably best not to share code directly, especially given the preexising pattern of similar functions duplicated across the two plugins.
- I was also a bit nervous about using the same `appInfo` in non-Stripe areas, since it may lead to abuse/misuse such as additional fields getting passed in to Stripe because they were useful in other areas.
- It should not be the path of least resistance for new usages of `stripe` within either plugin to be called without `setAppInfo`; duplicating that function call in many places made me nervous. Relatedly, Reaction may want to add other settings for each stripe instance in the future, such as `stripe.setApiVersion()`, `stripe.setHttpAgent()`, or `stripe.setClientId()`. 

Accordingly, I centralized things a bit with one `getStripeInstance` function per plugin. Hopefully this should make it easier in the future to keep things somewhat centralized.  In preparation, I found `getStripeApi` to be a slightly confusing name when in parallel to `getStripeInstance`, and renamed it to `getStripeApiKey`.

Additionally, I'm considering a follow-on refactor to deduplicate/unify `getStripeInstanceFromShop` and `getStripeApiKey`, though I'm not yet 100% sure of the implications/benefits of that. Curious for your thoughts on that!

Totally happy to go with your originally suggested solution of `/imports/utils/appInfo.js` if preferred! 

Mind taking another look @aldeed ?

aldeed:
@rattrayalex-stripe People not running marketplace should never have `connectAuth` set because it's only set in the marketplace plugin BUT in the interest of not trying to do too much in one PR, let's leave those lines as-is.

You can safely make the `paymentPluginName` change, though. Use the `PACKAGE_NAME` constant. There are actually several files with that constant defined, so you could instead define it in a single file (within each plugin) and import everywhere.

aldeed:
I switched the base to `develop` and fixed conflicts. Tested to verify checkout with Stripe still works. This will be merged if all the checks pass again.

For future PRs be sure to branch from and direct PRs to `develop` unless it's a critical fix to the current release.

rattrayalex-stripe:
Awesome, thanks so much @aldeed ! 

Follow-up PR at https://github.com/reactioncommerce/reaction/pull/4951

rattrayalex-stripe:
By the way, your contribution guide was super helpful and clear - it looks like the use of a `develop` branch isn't documented at https://docs.reactioncommerce.com/docs/next/contributing-to-reaction#step-3-prepare-a-pull-request-for-review - might be nice to add it there too.

PR #4947, opened 2019-01-30
Remove the checkout UI

aldeed:
Impact: **major**  
Type: **feature**

## Changes
BREAKING CHANGE

All links, buttons, routes, and components related to checkout are now removed. This helps clarify the fact that in 2.0, this app is only to be used as an API and a UI for shop operators. Storefront UI can be built from scratch or by starting from https://github.com/reactioncommerce/reaction-next-starterkit.

Additional PRs will be coming to remove other pieces of the storefront UI bit by bit until only an operator UI is left.

## Testing
- Verify that the storefront still works (for now), but there are no "checkout" buttons anywhere.
- Verify that doing things logged in as an operator still works.
- Verify that the new `/operator` page still works.


zenweasel:
Bye Felicia

aldeed:
@mikemurray I added that package back until we get component library tree shaking working properly. I had that error, too, and I thought I had fixed, but must not have committed it. Sorry for the trouble.

PR #4949, opened 2019-01-30
fix: taxes no updating reactively in starterkit

kieckhafer:
Resolveshttps://github.com/reactioncommerce/reaction-next-starterkit/issues/469
Impact: **minor**  
Type: **bugfix**

## Issue
When changing the quantity of the items or shipping address, the tax calculated is not updated reactively, the page has to be refreshed to get the taxes.

## Solution
Return an updated Cart to the client after `appEvents` has run. We were previously returning the same `Cart` object as before it had run, therefore not getting any of the updates.

## Breaking changes
None

## Testing
1. Start up reaction platform - most of the testing is in the starterkit
1. Go to Admin, enable "Example tax service"
1. Create a new product.
1. Add it to the cart, go to the checkout page, see that taxes are calculated.
1. Change the quantity of the item. See that taxes are recalculated.


imports/plugins/core/cart/server/no-meteor/mutations/updateCartItemsQuantity.js
```diff
@@ -81,5 +81,8 @@ export default async function updateCartItemsQuantity(context, input) {
     updatedBy: userId
   });
 
-  return { cart: updatedCart };
+  // Re-fetch cart with updated data
+  const updatedCartAfterAppEvents = await Cart.findOne({ _id: cartId });
```
zenweasel: I guess the only real question here is what is the performance impact of constantly refetching the cart. Certainly that's going to significantly impact how many times we grab from that db.

kieckhafer: I see where that could be an issue... although this is only called when items are updated once they are in the cart, not when items are added or removed from that cart, those are other mutations. Not saying that definitely makes it OK, but I think the frequency of this mutation being hit is much less than any other cart actions.

That being said, anytime we run this mutation pre-my update here, we hit `appEvents` which then hits the DB a few times throughout the different events. Is hitting the DB inside an `appEvent` with an `update` or `findOneAndUpdate` any different than what I'm doing here?

zenweasel: So if you update the cart above, and an external process needs to then update the cart with additional/new data, what ensures that your new fetch will run **after** that data is completed, especially when that external process may be hitting an external API. It may work locally with custom rates but with something that takes longer it might not?

PR #4955, opened 2019-02-01
feat: allow tax services to add custom data to taxes

### LINKED ISSUE #4945: Provision for storing custom data in order's taxes
Reaction plugin might need to store custom data along with the data returned by the `calculateOrderTax` function.
Currently there is no way to store this data. So we need a way that plugins can store custom data in the order.

aldeed:
Resolves #4945   
Impact: **minor**  
Type: **feature**

## Changes
A tax service's `calculateOrderTaxes` function may now return a black box `customData` object on each item in the `taxes` arrays.

This is not exposed through GraphQL. If you need any of the fields, you can create a custom plugin to `extend type CalculatedTax` with your properly typed custom fields and add resolvers as necessary.

Primarily, though, this is intended to be used to store extra data that the third-party tax integrations need for later API calls.

## Breaking changes
None

## Testing
This isn't currently used in the built-in tax service. You'd have to modify that `calculateOrderTaxes` function to include `customData` in its return, and then verify that it's stored in the database.

### DIFF
--- a/imports/plugins/core/taxes/lib/simpleSchemas.js
+++ b/imports/plugins/core/taxes/lib/simpleSchemas.js
@@ -2,6 +2,16 @@ import SimpleSchema from "simpl-schema";
 
 export const Taxes = new SimpleSchema({
   _id: String,
+  /**
+   * Custom key/value data that you need to store.
+   * You'll need to extend GraphQL schemas if you
+   * want to expose any of this data through the API.
+   */
+  customData: {
+    type: Object,
+    blackbox: true,
+    optional: true
+  },
   jurisdictionId: {
     type: String,
     optional: true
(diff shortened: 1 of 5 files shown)


imports/plugins/core/taxes/lib/simpleSchemas.js
```diff
@@ -2,6 +2,16 @@ import SimpleSchema from "simpl-schema";
...
+  /**
+   * Custom key/value data that you need to store.
+   * You'll need to extend GraphQL schemas if you
+   * want to expose any of this data through the API.
+   */
+  customData: {
+    type: Object,
+    blackbox: true,
+    optional: true
+  },
```
aldeed: This is the only actual change. All other changes were in support of adding the tests.

aldeed:
@Akarshit Does this correctly solve #4945 for you?

Akarshit:
@aldeed I am testing this.

Akarshit:
Works!

Akarshit:
@aldeed One thing I just discovered is that this only lets me store data per tax item.
I still don't have a way to store something at a higher level, for example the what `taxClass` was the product taxes according to the API(which I get in the response.)

aldeed:
@Akarshit Would you want to store that on `taxSummary`? Can you submit a new issue with your suggestion?

PR #4965, opened 2019-02-06
feat: custom tax data part 2

### LINKED ISSUE #4945: Provision for storing custom data in order's taxes

### LINKED ISSUE #4964: Allow to store customFields on taxSummary.
There is no way to store something at cart/item level, for example to store API response with fields that concern the whole cart/item.

aldeed:
Resolves #4945   
Resolves #4964
Impact: **minor**  
Type: **feature**

## Issue
#4955 added ability to store custom fields on `taxes` array items, but there is still a need to store item-level and order-level custom tax data.

## Solution
It now allows `customFields` objects on not only `taxes` array items, but also on each item tax object, and on `taxSummary`. Note also that based on discussions since the first PR, the field name has been changed to `customFields` rather than `customData`.

## Breaking changes
The field name has been changed to `customFields` rather than `customData`. This is only breaking since the previous PR, which has not yet been released.

## Testing
See #4955 testing instructions

### DIFF
--- a/imports/plugins/core/taxes/lib/simpleSchemas.js
+++ b/imports/plugins/core/taxes/lib/simpleSchemas.js
@@ -7,7 +7,7 @@ export const Taxes = new SimpleSchema({
    * You'll need to extend GraphQL schemas if you
    * want to expose any of this data through the API.
    */
-  customData: {
+  customFields: {
     type: Object,
     blackbox: true,
     optional: true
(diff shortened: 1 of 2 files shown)


aldeed:
@Akarshit Hopefully this should be everything you need

kieckhafer:
@aldeed Just noticed this is causing issues in the starterkit when placing an order... `customFields` needs to be added to the `Order` schema as well.

I'm adding it in the surcharges PR #4829 

https://github.com/reactioncommerce/reaction/pull/4829/commits/43186e20f991f801d5187709e35437ed55c7d758

PR #4966, opened 2019-02-06
feat: shipping Operator into 2.0

kieckhafer:
Impact: **minor**  
Type: **feature**

## Issue
Shipping operator was not moved into Operator 2.0

## Solution
As a temporary solution, before updating to React / adding more features, simply include the existing Blaze template in Operator 2.0 so that we have something instead of a blank screen

## Breaking changes
None

## Testing
1. Start the app
1. Login as admin
1. Head to `/operator/shipping`
1. See that the existing Shipping operator UI is ported and functional


kieckhafer:
Closing in favor of $4967

PR #4979, opened 2019-02-11
Add custom fields to order schema

### LINKED ISSUE #4978: customFields is not allowed by the schema
## Issue Description
We are unable to place any order because `customFields` is not in `OrderSchema`. This is since merging https://github.com/reactioncommerce/reaction/pull/4962.

## Steps to Reproduce
1. Place an order in the starter kit.
2. See that placing the order will fail with this message: `customFields is not allowed by the schema`.

## Possible Solution
Just add `customFields` to `OrderSchema`.

ajporlante:
Resolves #4978
Impact: **critical**  
Type: **bugfix**

## Issue
`customFields` is not in the order schema causing placing any order to fail.

## Solution
Add `customFields` to order schema.

## Breaking changes
None

## Testing
1. Place an order from the starter kit.
2. The order should be successfully placed.

### DIFF
--- a/imports/collections/schemas/orders.js
+++ b/imports/collections/schemas/orders.js
@@ -375,6 +375,7 @@ export const OrderFulfillmentGroup = new SimpleSchema({
  * @property {String} cartId optional For tracking which cart created this order
  * @property {Date} createdAt required
  * @property {String} currencyCode required
+ * @property {Object[]} customFields optional
  * @property {Document[]} documents optional
  * @property {String} email optional
  * @property {Object[]} exportHistory optional
--- a/imports/collections/schemas/orders.js
+++ b/imports/collections/schemas/orders.js
@@ -413,6 +413,11 @@ export const Order = new SimpleSchema({
     type: String,
     optional: true
   },
+  "customFields": {
+    type: Object,
+    blackbox: true,
+    optional: true
+  },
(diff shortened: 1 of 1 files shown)


rosshadden:
Don't you need to add it to the GraphQL schema as well?

ajporlante:
Im a little confused on what's the intention here. Reading through this PR https://github.com/reactioncommerce/reaction/pull/4962, I think the intention was to let the plugins extend the `Order` schema and be specific on the structure of `customFields`. But the problem is that for plugins that do not require `customFields`, like Stripe or Example IOU, placing an order is broken because the `placeOrder` function always saves `customFields` and by default, the order schema does not have `customFields`. So there are 2 approaches here: add `customFields` in Stripe or Example IOU or do as I did in this PR.

ajporlante:
I see here that @kieckhafer will be adding my suggested changes here to this PR: https://github.com/reactioncommerce/reaction/pull/4829/files

So, we can close this.

kieckhafer:
@ajporlante It might be better to leave this open and get it merged quickly. The PR of mine you are referring to is a huge PR, and while it’s nearing the end of being reviewed, there might still be some snags before it gets merged.

Considering this is a breaking bug, we should fix it ASAP.

PR #4981, opened 2019-02-12
feat: add ordersByAccountId query

### LINKED ISSUE #4579: GraphQL: Implement ordersByAccountId query
Implement ordersByAccountId query

kieckhafer:
Resolves #4579
Impact: **minor**  
Type: **feature**

## Issue
While we had `ordersByAccountId` outlined in our GraphQL schema, we had not yet built out the query to use this.

## Solution
Add the needed query to be able to use `ordersByAccountId`

## Testing
1. Create orders from multiple accounts
1. Add the following into the `Shops` collection, if you'd like to see translations of order statuses
```
"orderStatusLabels": {
    "new" : [ 
        {
            "language" : "en",
            "label" : "Order received"
        }, 
        {
            "language" : "es",
            "label" : "Recibido"
        }
    ],
        {
            "language" : "en",
            "label" : "Order received"
        }, 
        {
            "language" : "es",
            "label" : "Recibido"
        }
    ],
        {
            "language" : "en",
            "label" : "Shipped"
        }, 
        {
            "language" : "es",
            "label" : "Enviado"
        }
    ],
        {
            "language" : "en",
            "label" : "Canceled"
        }, 
        {
            "language" : "es",
            "label" : "Cancelado"
        }
    ]
}
```
1. Use the following `query`, and make sure you only get back orders from the correct supplied accountId

```
{
  ordersByAccountId(accountId: "PUT_ACCOUNT_ID_HERE") {
    edges {
      node {
        _id
        status(language: "es") {
	  label
          status
        }
        ...anyOtherFieldsYouWantToTest
      }
    }
  }
}
```

```
{
  ordersByAccountId(accountId: "PUT_ACCOUNT_ID_HERE", shopIds: ["PUT_SHOP_ID(s)_HERE]) {
    edges {
      node {
        _id
        status(language: "es") {
	  label
          status
        }
        ...anyOtherFieldsYouWantToTest
      }
    }
  }
}
```

1. Remove translations and make sure query still works and returns raw `status` data if translations are not available at all (not in collection), or for the requested language (request `fr`)
1. If you'd like to see this in the UI, you can test against this starterkit branch: https://github.com/reactioncommerce/reaction-next-starterkit/pull/393 - however that one isn't quite ready so any UI errors please ignore, only use it to visualize the API as well as it can do for now


imports/plugins/core/orders/server/no-meteor/resolvers/Order/orderStatus.js
```diff
@@ -0,0 +1,44 @@
...
+ * @name "Order.orderStatus"
+ * @method
+ * @memberof Order/GraphQL
+ * @summary Displays a human readable status of order state
+ * @param {Object} order - Result of the parent resolver, which is a Order object in GraphQL schema format
+ * @return {String} A string of the order status
+ */
+export default async function orderStatus(order) {
+  const { workflow: { status } } = order;
+
```
kieckhafer: @aldeed what are your thoughts on different labels / workflow displays here?

What this does is take our `workflow.status` data, and xform it into a more human readable label to go into the Orders account page. We (@rymorgan @cassytaylor and I) were thinking that "new" doesn't really mean much to anyone, so we'd go directly with "Order Received". "processing" could also be iffy to an end user, so we'll keep it "Order Received at that point".

imports/plugins/core/orders/server/no-meteor/queries/ordersByAccountId.js
```diff
@@ -0,0 +1,39 @@
...
+  const { Orders } = collections;
+
+
+  if (!accountId) {
+    throw new ReactionError("invalid-param", "You must provide accountId arguments");
+  }
+
+  // TODO: EK - Permissions check here
+  // Unless you are an admin with orders permission, you are limited to seeing it if you placed it
+  // if (!userHasPermission(["orders"], shopId)) {
```
kieckhafer: @aldeed is there a way to use multiple shopIds here for this permissions check?

imports/plugins/core/orders/server/no-meteor/resolvers/Query/ordersByAccountId.js
```diff
@@ -0,0 +1,29 @@
...
+  const { accountId, shopIds: opaqueShopIds } = args;
+
+  const shopIds = opaqueShopIds && opaqueShopIds.map(decodeShopOpaqueId);
+
+  return context.queries.ordersByAccountId(context, {
+    accountId: decodeAccountOpaqueId(accountId),
+    shopIds
+  });
+
+  // TODO: EK - Get paginated response
```
kieckhafer: @aldeed is this somewhere where we might want paginated responses? Or do we just want to display all orders for a single user?

imports/plugins/core/orders/server/no-meteor/queries/ordersByAccountId.js
aldeed: No, you'll have to do the `userHasPermission` check in a `shopIds.forEach` loop.

On line 15, it should be `accountId: contextAccountId`

I would not change `accountId` since that could produce unexpected results. You could just wrap the whole permissions check loop in `if (accountId !== contextAccountId)`, like this:

```js
if (accountId !== contextAccountId) {
  shopIds.forEach((shopId) => {
    if (!userHasPermission(["orders"], shopId)) {
      throw new ReactionError("access-denied", "Access Denied");
    }
  });
}
```

kieckhafer: In this case, if `accountId === contextAccountId`, will it not just skip the check then?

aldeed: Actually I don't think my example handles the case where they have "orders" permission and they request all orders for another account, without passing any `shopIds`. I can think of two possible solutions to that:
- Look up the orders, get all the shopIds on the orders, and then run permission check for each
- Set `shopIds` to the list of IDs for which they user has "orders" permission.

The first seems like a lot of work, so I like the second. You'd have to add a new permission function like `shopsUserHasPermissionFor` and it would become something like this:

```js
if (accountId !== contextAccountId) {
  // If an admin wants all orders for an account, we force it to be limited to the
  // shops for which they're allowed to see orders.
  if (!shopIds) shopIds = shopsUserHasPermissionFor("orders");

  shopIds.forEach((shopId) => {
    if (!userHasPermission(["orders"], shopId)) {
      throw new ReactionError("access-denied", "Access Denied");
    }
  });
}
```

imports/plugins/core/orders/server/no-meteor/resolvers/Order/orderStatus.js
aldeed: @kieckhafer I'm all for readable labels but we probably should not do it with hard-coded values. Ideally these would be customizable and language-specific. The `label` resolver could take a `language` arg as we've done elsewhere, and values could come from a language-to-label map somewhere in shop settings?

```js
orderStatusLabels: {
  new: [
    { language: "en", label: "Order received" }
  ]
}
```

Otherwise it could be left up to frontend code, but that's not as nice since there could be inconsistencies in what label each frontend uses.

imports/plugins/core/orders/server/no-meteor/resolvers/Query/ordersByAccountId.js
aldeed: @kieckhafer Yes definitely pagination/connection

imports/plugins/core/orders/server/no-meteor/queries/ordersByAccountId.js
aldeed: @kieckhafer If `accountId === contextAccountId` why would any checks be necessary? You should be able to see all of your own orders regardless of what permissions you have.

kieckhafer:
@aldeed addressed all comments made in this PR (they were just answers to questions I had, nothing has been reviewed officially yet), and this is ready for an actual review.

Translations are living on the `Shops` collection, on the `orderStatusLabels` object for now. I have a DR open to further investigate this, and they can be easily moved. I've made a ticket to build out the UI for this, but for the sake of getting this pulled into forks that need it, I would prefer to get this base merged prior to that buildout: https://github.com/reactioncommerce/reaction/issues/5000

In other scenarios (Surcharges), we decided that if a language was not provided for translations, we would show nothing. In this case, I've made the decision to show the raw order status (i.e. `coreOrderWorkflow/processing`) if there is no translation available. What are your thoughts on that? Can easily be changed to show nothing if no label is present.

imports/plugins/core/orders/server/no-meteor/resolvers/Order/orderSummary.js
```diff
@@ -0,0 +1,80 @@
...
+  const totalDiscountsAmount = totalDiscounts.reduce((acc, value) => acc + value, 0);
+  const totalShippingAmount = totalShipping.reduce((acc, value) => acc + value, 0);
+  const totalSubtotalAmount = totalSubtotal.reduce((acc, value) => acc + value, 0);
+  const totalSurchargesAmount = totalSurcharges.reduce((acc, value) => acc + value, 0);
+  const totalTaxableAmountAmount = totalTaxableAmount.reduce((acc, value) => acc + value, 0);
+  const totalTaxesAmount = totalTaxes.reduce((acc, value) => acc + value, 0);
+  const totalTotalAmount = totalTotal.reduce((acc, value) => acc + value, 0);
+
+  // Get currencyCode from first fulfilllmentMethod
+  const { currencyCode } = fulfillmentMethods[0];
```
kieckhafer: @aldeed Not sure if this is the best way to get `currencyCode`. Was thinking we could also query and get the entire shop, and then take it from `shop.currency.code`, but didn't really want to add another query for the `Shops` collection when this data is available here.

My only qualm with it is what if we have more than one fulfillmentMethod. They both should probably be the same currency code, right?

aldeed: I don't think it's possible to have groups with different currencies, but to avoid the `[0]`, you can use `order.currencyCode`, which is required so should always be set.

imports/plugins/core/orders/server/no-meteor/queries/ordersByAccountId.js
```diff
@@ -0,0 +1,70 @@
...
+  if (!accountId) {
+    throw new ReactionError("invalid-param", "You must provide accountId arguments");
+  }
+
+  let query = {
+    accountId
+  };
+
+  // Limit which orders are returned: `all`, `canceled`, `completed`, or open`
+  if (orderStatus !== "all" || null || undefined) {
```
aldeed: JS logic doesn't work this way. You would need `orderStatus !== "all" || orderStatus === null || orderStatus === undefined`, but I don't think you actually need the outer `if` at all. You can remove it and do `if/else if` or a `switch` on `orderStatus`.

imports/plugins/core/orders/server/no-meteor/queries/ordersByAccountId.js
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * @name ordersByAccountId
+ * @method
+ * @memberof Order/NoMeteorQueries
+ * @summary Query the Orders collection for orders made by the provided accountId and (optionally) shopIds
+ * @param {Object} context - an object containing the per-request state
+ * @param {Object} params - request parameters
+ * @param {String} params.accountId - Account ID to search orders for
+ * @param {String} params.orderStatus - Workflow status to limit search results
```
aldeed: It will be more useful and not much extra work to do `orderStatuses` array rather than just one. Return orders with any of the statuses in the array. (So in UI the filter could be check boxes rather than a single status select.)

imports/plugins/core/orders/server/no-meteor/schemas/schema.graphql
```diff
@@ -213,6 +259,12 @@ type Order implements Node {
   "The shop through which the order was placed"
   shop: Shop!
 
+  "The status of the order. Can be new, processing, shipped, complete."
+  status(language: String!): OrderStatus
```
aldeed: Looking at this all put together, I'm having second thoughts about needing the `OrderStatus` object. It would be simpler to query and access if we put both directly on the order. Thoughts?

```graphql
"The machine-readable order status"
status: String!

"The order status for display in UI"
displayStatus(language: String!): String!
```

Plus then you need not specify a language if you need only the status enum without needing to display it.

aldeed: Note that I also added `!` to the responses because I believe they will never be `null`, and I called it `displayStatus` rather than `statusLabel` to better match the naming we've used elsewhere in GraphQL.

imports/plugins/core/orders/server/no-meteor/queries/ordersByAccountId.js
kieckhafer: Good idea. How do we want to go about the `All` selection then? Just have all the statuses checked by default?

kieckhafer: @rymorgan See above. What are your thoughts on using a multi-select instead of a single dropdown for the Order Status filter?

I don't think it looks that great, might get confusing for the user too.

I think @aldeed's idea of passing an array is best for the server side, but on the client maybe we have pre-set arrays that we pass (at least in our starterkit, other people can do what we want), and continue to use the single select. Thoughts?

account_profile___reaction

rymorgan: Agreed that a multi-select would likely be confusing to a user. I think it adds an extra bit of complexity to the user experience for not a huge benefit to the user. And actually a detriment to the user if they are confused by the UI. I don't remember seeing this type of behavior on brands that I visit which makes me think we'd be designing something almost all storefronts would eliminate.

aldeed: I agree, too. I was thinking more of future operator UI needs.

PR #4982, opened 2019-02-12
Fix: Don't crash when mediaItem.URLs is null

focusaurus:
Impact: **minor**  
Type: **bugfix**

## Issue
With some imperfect development data we have some catalog products that have a `primaryImage.URLs: null`. This situation causes the destructuring to fail with this error:

```
GraphQL error: Cannot destructure property `large` of 'undefined' or 'null'.
    at new ApolloError (/usr/local/src/node_modules/apollo-client/errors/ApolloError.js:25:1)
```

## Solution
Return null early and avoid the exception.

## Testing

This module was previously barren of unit tests. I added some basic tests that confirm this fix plus the base case behavior.

### DIFF
--- a/imports/plugins/core/graphql/server/no-meteor/xforms/catalogProduct.js
+++ b/imports/plugins/core/graphql/server/no-meteor/xforms/catalogProduct.js
@@ -15,7 +15,7 @@ export const encodeCatalogProductOpaqueId = encodeOpaqueId(namespaces.CatalogPro
  * @return {Object} transformed product media item
  */
 export function xformProductMedia(mediaItem, context) {
-  if (!mediaItem) return null;
+  if (!(mediaItem && mediaItem.URLs)) return null;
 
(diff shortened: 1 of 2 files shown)


aldeed:
@focusaurus Very happy to merge unit tests, but as a general rule I don't think we should have to guard against values that the schema doesn't allow. In the `ImageInfo` SimpleSchema, `URLs` is required (but can be an empty object). So my bigger concern would be finding out where we are not validating against that schema and fixing it to validate. We should not let such bad data be saved in the first place.

dancastellon:
Following up here - `URLs` isn't present for Cloudinary images because that object is build in the GraphQL resolver.

aldeed:
@dancastellon It would still need to be set to `{}`, though, or it will fail the schema validation. Otherwise we can make that field optional in the schema, but I'm not sure that makes sense to do outside of the Cloudinary integration use case.

PR #4986, opened 2019-02-14
fix: properly save all customFields from tax service result

aldeed:
Impact: **minor**  
Type: **bugfix**

## Issue
https://github.com/reactioncommerce/reaction/pull/4955 and https://github.com/reactioncommerce/reaction/pull/4965 added the ability for tax services to return `customFields` in the response. A couple bugs were found after merging.
- When calculating for a `cart`, `customFields` on cart items and on `taxSummary` were not saved to the database
- When calculating for an `order`, `customFields` on cart items were not saved to the database

## Solution
All `customFields` returned by a tax service at any level are now saved in the database, for both cart and order.

On cart and order items, the field name in the database is `customTaxFields` as opposed to `customFields` because otherwise it would be unclear that they're related to taxes since they're saved directly on the item object. The tax service should still return them as `customFields`; the field name change is only in the database.

## Breaking changes
None

## Testing
1. Temporarily update the `calculateOrderTaxes` function in the `taxes-rates` plugin to include `customFields` objects in the `taxSummary`, in one of the `itemTaxes` objects, and in one of the `taxes` objects.
2. Add items to a cart and begin checking out, using a destination address for which the shop collects taxes.
3. Once you see a tax amount on the checkout page, look in MongoDB to verify that all three expected `customFields` objects were saved in the `Cart` document.
4. Finish checking out. Look in MongoDB to verify that all three expected `customFields` objects were saved in the `Order` document.

### DIFF
--- a/imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
+++ b/imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
@@ -470,7 +470,7 @@ export default async function placeOrder(context, input) {
     await addShipmentMethodToGroup(context, finalGroup, cleanedInput, groupInput, discountTotal, orderId);
 
     // Apply Taxes
-    await addTaxesToGroup(context, finalGroup, cleanedInput, groupInput, discountTotal, orderId);
+    await addTaxesToGroup(context, finalGroup, orderInput, discountTotal, orderId);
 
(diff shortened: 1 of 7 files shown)


aldeed:
@Akarshit This should fix the remaining issues with saving `customFields` for taxes. I have unit tests to prove it this time! 😄

PR #4990, opened 2019-02-15
chore: delete unused files

aldeed:
Impact: **minor**  
Type: **chore**

## Changes
Deletes some files that are no longer used for anything

## Testing
No testing necessary since none of the files are part of the codebase.


aldeed:
@spencern Can you confirm that none of these files are needed and merge?

PR #4992, opened 2019-02-15
Meteor and Docker performance tweaks

aldeed:
Impact: **minor**  
Type: **feature**

## Changes
- Removed CLI dependency. The same Node code from the CLI, with a few unnecessary things removed, is now in `.reaction/run`. I added logging showing how long each step takes.
- In `docker-compose.yml`, added `no-audit` flag on the `meteor npm install` command to speed that up a little. It's now consistently around 20 seconds for me when it doesn't need to install anything.
- In `docker-compose.yml`, added `:cached` for the volume that links the host files in, which should help with file watching/syncing a bit. See https://docs.docker.com/compose/compose-file/#caching-options-for-volume-mounts-docker-desktop-for-mac and https://docs.docker.com/docker-for-mac/osxfs-caching/
- Added `--no-release-check` and `--no-lint` flags for the `meteor run` command. Added `--no-release-check` for the `meteor test` commands. Not sure I see much of a difference from this on my machine but I don't think it hurts.
- After the Meteor startup finishes and the Reaction startup begins, it now logs how many milliseconds the Meteor startup took. This is just helpful for comparing and debugging performance issues.

## Breaking changes
I removed the code that parsed a `settings.json` file and passed it in as the Meteor `--settings` flag. Not sure if this is still something we use or support? It could be added back but would mean porting some additional code from the CLI.

## Testing
Verify the app starts up and runs. Ideally starts and/or restarts faster now.


aldeed:
@kieckhafer I'm interested to know whether you see any improvement with these changes.

kieckhafer:
@aldeed Startup time is a slight improvement, but reload time after a change is HUGE. @mikemurray check this out, see how it works for you.

**Initial startup time**
This PR: 6m45s on average
Previously: Over 8 minutes, usually closer to 10 minutes

**After making a change**
This PR: 20 seconds!
Previously: 3-10 minutes, depending on how long the app has been running

One odd thing, not sure if this has to do with the caching or not. After a change, I am able to reload the page right away, which is nice, as in the past if you try that it just spins forever while the app is restarting. However, if you reload immediately, you don't see the changes. If you wait about 20 seconds to reload, then you do see the changes.

focusaurus:
Can we make this support debugging? Maybe add these to package.json scripts?

```
"inspect": "node --experimental-modules --inspect ./.reaction/run/index.mjs",
"inspect-brk": "node --experimental-modules --inspect-brk ./.reaction/run/index.mjs",
"inspect-docker": "node --experimental-modules --inspect=0.0.0.0:9229 ./.reaction/run/index.mjs",
"inspect-brk-docker": "node --experimental-modules --inspect-brk=0.0.0.0:9229 ./.reaction/run/index.mjs",
```

I was working on getting [this reaction-cli PR](https://github.com/reactioncommerce/reaction-cli/pull/79) released but this would make that obsolete.

focusaurus:
CC: @loan-laux Heads up FYI. This might provide an alternative way to get debugging working. I might test my proposed changes to this setup locally and see if that works.

focusaurus:
@aldeed What do you think about [these changes to support debugging](https://github.com/reactioncommerce/reaction/compare/chore-aldeed-perf-tweaks...chore-pete-perf-tweaks?expand=1)

focusaurus:
@aldeed Did you try `delegated` docker volume setting? My reading of the docs would indicate that would be more appropriate for us instead of `cached`.

> However, by relinquishing consistency, delegated mounts offer significantly better performance than the other configurations. Where the data written is ephemeral or readily reproducible, such as from scratch space or build artifacts, delegated may be the right choice.

ticean:
@focusaurus re: `delegated`, the definitions are:

> * consistent: perfect consistency (host and container have an identical view of the mount at all times)
> * cached: the host’s view is authoritative (permit delays before updates on the host appear in the container)
> * delegated: the *container’s view is authoritative* (permit delays before updates on the container appear in the host)

Delegated sematics section reads:

> If changes to the mount source directory are present on the host file system, those changes may be lost when the delegated mount synchronizes with the host source directory.

Doesn't that mean that developers could lose changes to the`src` directory?

---

Quicklink to docs: https://docs.docker.com/docker-for-mac/osxfs-caching/

focusaurus:
I see. I thought it was just "delays" not losing changes. Do we want a `delegated` mount just for the `.meteor` directory though?

loan-laux:
@focusaurus Got it. Would definitely be nice having debugging on here. Really excited about these perf improvements — nice job!

mikemurray:
@kieckhafer @aldeed Here are my results. Seems to shave off 1-2 mins

**Initial startup time**
```
docker-compose down -v --rmi local
docker-compose up -d
```

This PR: ~5m57s
Previously: ~6m50s

**Stop 'n start Reaction container**
```
docker-compose stop reaction
docker-compose up -d reaction
```

This PR: ~2m42s
Previously: ~4m28s

**After making a change**
This PR: ~50s
Previously: ~1m41s

machikoyasuda:
@ticean oops! i didn't mean to re-request a review from you. clicked that mysterious button on accident!!

aldeed:
@focusaurus I pulled in the debugging changes from your branch. Thanks!

`cached` makes the most sense to me for the project files, but I hadn't thought about `.meteor`. The files directly in `.meteor` are probably best treated as part of the project and therefore `cached`, but `.meteor/local`, where all of the temporary builds are done, seems right for `delegated`. I'm a bit worried about what might happen when mixing `cached` and `delegated` in the same tree like that, but I guess there's one way to find out...

aldeed:
Also based on my reading I think these control both whether delays are OK _and_ which is the source of truth when updates conflict. The host is SOT for cached and the container is SOT for delegated. Otherwise in normal mode they jump through hoops to ensure there can never be a conflicting view, which is why it leads to poor performance on Mac.

aldeed:
Using `delegated` for `.meteor/local` seems to work for me. I'm going to get this merged, but if anyone encounters related issues once this is in `develop` we can discuss and fix or revert as necessary.

PR #4998, opened 2019-02-19
De-Meteorize emailing code in the API

### LINKED ISSUE #4997: De-Meteorize emailing code in the API
See epic https://github.com/reactioncommerce/reaction/issues/4996

## Work
- Update core and SMTP email sending code so that it does not use any Meteor dependencies, except for the JobCollection job queue. De-Meteorizing the job queue system will be done in a follow-up task, at which point emails will be sent when running the API as a pure Node app.
- Move updated/new code to proper locations if the current location isn't in keeping with plugin responsibilities and newer best practices

### LINKED ISSUE #5097: Emails: Order confirmation email should be overrideable.
## Issue

A plugin should be able to block the Core API from sending Order confirmation emails. Otherwise, if a plugin sends a custom Order confirmation email, and the API sends one as well, the customer will get 2 emails.

## Solution

- Update `sendOrderEmail` function to have it call a mutation function at the end instead of directly calling `Reaction.Email.send`, so that plugins can overwrite that mutation function.

aldeed:
Resolves #4997   
Resolves #5097 
Impact: **major**  
Type: **feature**

## Changes
Internally, the server code related to sending emails in `core`, `email`, `email-smtp`, and `email-templates` plugins no longer relies on any Meteor globals or imports.

Exceptions:
- `/imports/utils/jobs.js` imports Meteor dependencies to inject them into job-collection for job queueing. This is scheduled to be addressed separately.
- The `emails/retryFailed` Meteor method is still a Meteor method. It is currently only called by operator UI, so it isn't important to convert and may change in the future anyway.
- The `email/saveSettings` and `email/verifySettings` Meteor methods in the `email-smtp` plugin are still Meteor methods.  They are currently only called by operator UI, so it isn't important to convert them yet.

Specifics:
- A new `context.mutations.sendEmail` function is now the way to send email rather than `Reaction.Email.send`.
- Everything on the `Reaction.Email` object except `Reaction.Email.send` is removed. `Reaction.Email.send` is deprecated but available, though it now passes through to `context.mutations.sendEmail` function
- In `email-templates` plugin, the built-in seed HTML templates are now `.js` files instead of `.html` files. They export only an HTML string, and that HTML has not changed, but this was easier than keeping them HTML files as far as being able to import them for seeding without jumping through too many hoops (Babel, Meteor file loading and static imports, etc.)
- The `sendEmail` mutation now calls `context.mutations.renderEmail` and expects back a Promise that resolves with `{ html, subject }`. This allows any plugin to manage email templates and rendering responsibilities. By default, the `email-templates` plugin does this. If you want your own solution, remove the `email-templates` plugin and register a mutation named `renderEmail` in your own custom plugin.
- [BREAKING] In `email-templates` plugin, the `renderEmail` mutation now uses `handlebars` NPM package to compile the HTML templates, rather than Meteor Blaze/Spacebars. This is not technically a breaking change because email templates have always been tagged as `handlebars` templates. However, Spacebars supports a few templating features that Handlebars does not, so it would be a good idea to test and adjust all of your email templates before deploying this change to production.

# Below here is from PR #5102 merged into this one

## Issue
Plugins can trigger emails, like the Order Confirmation email, to be sent after a certain `appEvent`, but they currently cannot cancel the API's default email from being sent. As a result, in the current app, if a custom plugin sends a custom Order Confirmation email, the customer will get 2 emails: Reaction's default, and the custom one. A plugin developer should be able to _override_ the default Order Confirmation email method, rather than add its own.

## Solution
- Refactor the current `core/orders/server/util/sendOrderEmail.js` method to use a new `core/orders/server/no-meteor` mutation method instead.
- Plugins can now override this `sendOrderEmail` mutation.

## Breaking changes
None

### Setup
1. Log in as operator.
2. Edit Shop settings and make sure the shop email address is testing@reactioncommerce.com. This address is verified to send emails through the Reaction testing SES account. Other addresses may not work.
3. Go to the operator user profile page and make sure you have a real email address. You can use something like `<yourname>+admin@reactioncommerce.com`.
4. Enter correct email service credentials on Email page. Make sure the indicator is green.

### Reset Password Email
6. Log out.
7. Click Reset password link, enter admin email, and click Reset your password.
8. Verify that you get the password reset email and it looks correct. Make sure the Reset Password button in the email goes to the correct page and allows you to reset your password.

### User Invite Email
1. Log in to operator UI
2. Go to Accounts screen
3. On "Add Admin User" card, add a new user with a real email address. For example, `<yourname>+admin2@reactioncommerce.com`
4. Verify that you get the "You are invited to join" email and that everything in it looks correct and clicking the button goes to the correct page.

### Verify Operator Email Address Email
1. Log in to operator UI
2. Go to Profile page
3. Change your email address to a different real address.
4. Verify that you get the "Verify your email address" email and that everything in it looks correct and clicking the button goes to the correct page.

### Welcome Email
1. Register a new user from storefront UI
2. Verify that you get the welcome email and the Get Started button goes to the storefront.

### New Order Email
1. Place an order from storefront.
2. Verify that you get the "order received" email and everything in it looks correct.
3. Ensure order link goes to the storefront order page

### Order Item Refund
1. Log in to operator UI.
2. On Orders page, view the order.
3. Approve and capture payment if not already done
4. Click on an item in the "Invoice" section.
5. In the popup, select the check box next to an item and click Refund items.

### Cancel Order Email
1. Log in to operator UI.
2. On Orders page, view the order.
3. Click Cancel Order and cancel it.
4. Verify you get the refund email.

### Order Shipped Email
1. Log in to operator UI.
2. On Orders page, view the order.
3. In the Shipping section, click "All items packed" and then "Shipped"
4. Verify you get "order on its way" email, with correct data in it
5. Click Resend Shipment Notification
6. Verify you get the email again.

### Known Issues, Unrelated to PR, You Will See While Testing
- Reset Password button in email doesn't go to reset password form
- Button from "invited" email doesn't bring you to a page to accept the invitation or explain how to register and accept.
- Button in welcome email goes to the IDP sign-in form, but signing in does not work and does not redirect you to the storefront URL as expected.
- Order link in order emails does not go to the storefront order page. Goes to the old page on operator UI


aldeed:
I'm still testing this but moved to review in case anyone else wants to test or review in the meantime.

PR #5003, opened 2019-02-21
feat: Use .env.example files from custom plugins

focusaurus:
- Combine the main .env.example file with any .env.example
  files from custom plugins during setup
- Also support ignoring commented-out lines in `.env.example` files
- Also ignore blank lines
- also format with shfmt
- also fix one shellcheck lint error

Impact: **minor**  
Type: **feature**

## Issue

Sometimes reaction custom plugins need environment variable settings, and need to provide examples. Previously only the main reaction `.env.example` file was used to initially populate the user's `.env` file. Any settings from custom plugins would require a documented manual process to initially setup.

## Solution

Search the `imports/plugins/custom` directory for any files named `.env.example` and incorporate those into the main `.env`.

## Breaking changes

The script has always avoided overwriting any values that already exist in the file, and that behavior is retained.

## Testing

I locally tested a few scenarios

- no extra `.env.example` files
- several custom plugins with `.env.example` files
- a custom plugin with an `.env.example`in a `docs` subdire

**How to test this yourself**
- check out this feature branch
- put a bunch of varying `.env.example` files in your directory tree including the reaction project directory and anywhere under `imports/plugins/custom`
- Populate those files with a mix of `KEY=value` vars and commented out lines
- Run `bin/setup`
- Examine your project root `.env` file and make sure it has the values you expect

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -2,29 +2,37 @@
 
 __dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 env_file=${__dir}/../.env
-env_example_file=${__dir}/../.env.example
 
-function main {
+function main() {
   set -e
 
   add_new_env_vars
 }
 
-function add_new_env_vars {
+function add_new_env_vars() {
   # create .env and set perms if it does not exist
-  [ ! -f "${env_file}" ] && { touch "${env_file}" ; chmod 0600 "${env_file}" ; }
+  [[ ! -f "${env_file}" ]] && {
+    touch "${env_file}"
+    chmod 0600 "${env_file}"
+  }
 
-  export IFS=$'\n'
-  for var in $(cat "${env_example_file}"); do
-    key="${var%%=*}"     # get var key
-    var=$(eval echo "$var") # generate dynamic values
+  find . imports/plugins/custom -name .env.example -type f -print0 |
+    xargs -0 grep -Ehv '^\s*#' |
+    {
+      while IFS= read -r var; do
+        if [[ -z "${var}" ]]; then
+          continue
+        fi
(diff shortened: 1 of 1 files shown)


zenweasel:
@focusaurus I think the "testing" portion of the PR template refers to how I would be able to test it to ensure it works

focusaurus:
OK @zenweasel I updated the description to have that info too

focusaurus:
Ah crap. That's a GNU xargs vs BSD xargs thing. BSD xargs doesn't support `--no-run-if-empty`. I think it can be eliminated though.

focusaurus:
OK I removed the `--no-run-if-empty`. It was belt and suspenders anyway. Should be portable now.

zenweasel:
IMHO I think generally we want to avoid doing a `--force-push` as this means anybody else working on a branch (like a reviewer) cannot fast-forward.

zenweasel:
So I if I have a commented out line in one of my `.env.example` it adds a blank line, if I rerun it the other entries are fine but it continues to add blank lines for every time I run it.

focusaurus:
- OK I can't reproduce the blank lines on linux. Will try on my mac.
- Regarding amending commits and force pushing, this seems like a 1-commit change is what we want long term but sounds like a process we should define in Notion. Dunno, git seems like it's not giving us any great options in this scenario.

focusaurus:
Ah, OK it's not comment lines, it's blank lines which get carried forward.

focusaurus:
OK @zenweasel I think the latest change should prevent blank line propagation. Sorry about the force pushes but I didn't want to change process mid-PR.

zenweasel:
Works great. Great work. 👍

PR #5010, opened 2019-02-26
Add cancelOrderItem mutation

### LINKED ISSUE #2382: Cancel a line item from an order
Users with orders permissions should be able to cancel a line item from an order.

Cancelling a line item or reducing the quantity of a line item before an order has been processed should cause the following.
1. Reduce the amount of the credit authorization or cancel and reauthorize for the correct amount.
2. Trigger a Line Item Cancelled event
3. Send the customer an email and a notification letting them know that an item has been removed from their order.

aldeed:
Resolves #2382
Impact: **minor**  
Type: **feature**

## Changes
A new `cancelOrderItem` mutation allows you to cancel a single order item or a partial quantity of a single order item.

The operator UI does not yet use this. That will be done in a separate PR, at which time the `cancelOrder` Meteor method will also be removed.

- `cancelOrderItem` can be called internally to synchronize from an external system by setting `context.isInternalCall` to `true`
- `cancelOrderItem` can be called through GraphQL by any user with "orders" permission for the shop that owns the order
- `cancelOrderItem` can be called through GraphQL by the user who placed the order. In this case, the cancellation will only succeed if both the item status and the order status are "new".
- Unlike with the existing Meteor method, neither restocking nor refunds are handled by this mutation. These are separate things with separate systems, so we'll now expect clients to separately ask for restocking or refunds as necessary.
- When the last item in a group is canceled, the group is also canceled. When the last group in an order is canceled, the order is also canceled.

## Breaking changes
None

## Testing
1. Place an order.
1. Verify that you can cancel an order item when authenticated as a user with "orders" role or as the user who placed the order, but not when authenticated as someone else or when not authenticated.


imports/plugins/core/orders/server/no-meteor/resolvers/Mutation/cancelOrderItem.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @name Mutation/cancelOrderItem
+ * @method
+ * @memberof Payments/GraphQL
+ * @summary resolver for the cancelOrderItem GraphQL mutation
+ * @param {Object} parentResult - unused
+ * @param {Object} args.input - an object of all mutation arguments that were sent by the client
+ * @param {String} args.input.orderId - The order ID
+ * @param {String} args.input.itemId - Payment info
```
kieckhafer: This should be a comment concerning `item Id`, not `Payment info`

imports/plugins/core/orders/server/no-meteor/schemas/schema.graphql
```diff
@@ -23,6 +23,17 @@ extend type Query {
 }
 
 extend type Mutation {
+  """
+  Use this mutation to cancel one item of an order, either for the full ordered quantity
+  or for a partial quantity. If partial, the item will be split into two items and the
+  original item will have a lower quantity and will be canceled.
```
kieckhafer: This is a good comment to use inline in the code where I asked for more commenting in the mutation. This makes sense, and having it inline there would be helpful.

Akarshit:
The mutation works and does what I had in my mind so 👍

imports/plugins/core/orders/server/no-meteor/mutations/cancelOrderItem.js
```diff
@@ -0,0 +1,159 @@
...
+  if (modifiedCount === 0 || !updatedOrder) throw new ReactionError("server-error", "Unable to update order");
+
+  await appEvents.emit("afterOrderUpdate", {
+    order: updatedOrder,
+    updatedBy: userId
+  });
+
+  if (fullOrderWasCanceled) {
+    await appEvents.emit("afterOrderCancel", {
+      cancelledBy: userId,
```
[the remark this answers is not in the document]
aldeed: Gah, I hate that word. There isn't a correct spelling afaik, but at one place I worked they always used one "l" based on the recommendation of Microsoft's manual of style. So I agree and I'll change it.

imports/plugins/core/orders/server/no-meteor/mutations/cancelOrderItem.js
```diff
@@ -0,0 +1,159 @@
...
+    let itemToAdd;
+    const updatedItems = group.items.map((item) => {
+      if (item._id !== itemId) return item;
+      foundItem = true;
+
+      if (accountIsOrderer && !itemStatusesThatOrdererCanCancel.includes(item.workflow.status)) {
+        throw new ReactionError("invalid", `Item status (${item.workflow.status}) is not one of: ${itemStatusesThatOrdererCanCancel.join(", ")}`);
+      }
+
+      if (item.quantity > cancelQuantity) {
```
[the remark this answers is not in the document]
aldeed: Yes, we don't need it if equal. I'll add a comment.

imports/plugins/core/orders/server/no-meteor/mutations/cancelOrderItem.js
```diff
@@ -0,0 +1,185 @@
...
+        throw new ReactionError("invalid-param", "cancelQuantity may not be greater than item quantity");
+      }
+
+      // If we make it this far, then we've found the item that they want to cancel.
+      // We set the status and the cancel reason if one was provided.
+      // This will also decrement the quantity to match the quantity that is being
+      // canceled, which will be offset by pushing `itemToAdd` into the array later.
+      return {
+        ...item,
+        cancelReason: reason,
```
kieckhafer: Not sure if this is a big deal, as it doesn't seem to do anything aside from add a new item to the array, but I can run the same mutation multiple times and the workflow updates each time:

robo_3t_-_1_1

aldeed: It's probably nicer if we avoid doing an update when it's already canceled. I'll figure out a fix.

aldeed: I pushed a fix. It will still allow the cancelation, so `cancelReason` for example would change if called a second time with a different reason, but it won't change the status or add to workflow array.

imports/plugins/core/orders/server/no-meteor/mutations/cancelOrderItem.js
```diff
@@ -92,11 +92,15 @@ export default async function cancelOrderItem(context, input) {
...
+        const newItemQuantity = item.quantity - cancelQuantity;
         itemToAdd = {
           ...item,
           _id: Random.id(),
-          quantity: item.quantity - cancelQuantity
+          quantity: newItemQuantity
         };
+
+        // Update the subtotal since it is related to the quantity
+        itemToAdd.subtotal = item.price.amount * newItemQuantity;
```
kieckhafer: Haven't tested this in action yet, but based on fixing a similar issue when adding items to cart, I think that `subtotal` is an object that contains an `amount` and `currencyCode`, not just a `number`.

Will confirm when testing...

imports/plugins/core/orders/server/no-meteor/mutations/cancelOrderItem.js
```diff
@@ -107,6 +111,9 @@ export default async function cancelOrderItem(context, input) {
         quantity: cancelQuantity
       };
 
+      // Update the subtotal since it is related to the quantity
+      updatedItem.subtotal = item.price.amount * cancelQuantity;
```
kieckhafer: Haven't tested this in action yet, but based on fixing a similar issue when adding items to a cart, I believe that `subtotal` is an object with `amount` and `currencyCode` on it, not just a `number`.

Will confirm / disprove when testing.

aldeed: Yes, this confused me at first, too. Love these little quirks!

kieckhafer:
Updates resolve existing issues, and can confirm that subtotals and itemIds are updated properly when cancelling.

Issue #4999, opened 2019-02-19
Backend method to fulfil order(shipment) after order is placed.


Akarshit:
Since most of the clients have a separate order management/tracking system, the acknowledgement of shipping items in the order comes in later, after the order is placed.
These tracking numbers, status can be at unit level.
We need a method that a plugin can just call which will split the order based on the shipping information and also update the status of the units/items/orders.

kkuzemka:
@spencern - this is directly blocking some work on the Radial order imports - let's prioritize this to the top of the list.

Akarshit:
A part of this is already implemented in https://github.com/reactioncommerce/reaction/pull/5010
We have to implement the code now that can change status to shipping and add shipping information, change status to "open"(which means start).
I have doc withe a couple of use cases which would be helpful to look at. Let me know who is working on this and I will pass the doc to him/her.

spencern:
Will you outline the use cases here in this ticket? GitHub issues are the preferred way to discuss these things

What part of this is implemented in #5010? - what remains?

Akarshit:
What remains: Implementing Open, Shipped, Adjust(only for display purposes) status change to the orders.

Here is an explanation of the issue and use cases in detail.
Most of the clients provide shipping fulfilment information after the order has been placed. It is at this time only that we can know if the order would be split is multiple shipments or not. 

This is not for any calculation/payment/or any sensitive purposes. It is **just for displaying order's status to customers** and to **send email on update.**

**Items**: The final product that gets added to the cart. Example: A red 'M' jacket

**Units:** The number of units of the `item` added to the cart. Example: 5 units of red 'M' jacket.

**Tracking#:** The tracking information about the product. Every "shipment" in reaction should have the same `tracking#`.

**Order Status:** The status of the complete order. It is `new` by default, then should go to `coreOrderWorkflow/processing``, then to `coreOrderWorkflow/completed` when all items are shipped(or atleast 1 shipped and rest are canceled) or to `coreOrderWorkflow/canceled` when all the items have been canceled.

The current order is stored in reaction in the following way(simplified for readabbility):

    {
    	"shipping" : [
    		{
    			"_id" : "Ship1",
    			"items" : [
    				{
    					"_id" : "A",
    					"subtotal" : 64.95,
    					"workflow" : {
    						"status" : "coreOrderWorkflow/processing",
    						"workflow" : [
    							"coreOrderWorkflow/created",
[...]

PR #5018, opened 2019-03-05
Add moveOrderItems mutation

aldeed:
Part of #4999
Impact: **minor**  
Type: **feature**

## Changes
A new `moveOrderItems` mutation allows you to move one or more order items from one fulfillment group to another existing fulfillment group on the same order.

- `moveOrderItems` can be called internally to synchronize from an external system by setting `context.isInternalCall` to `true`
- `moveOrderItems` can be called through GraphQL by any user with "orders" permission for the shop that owns the order
- `moveOrderItems` can be called through GraphQL by the user who placed the order. In this case, the move will only succeed if both the item status and the order status are "new".

## Breaking changes
None

## Testing
1. Place an order.
1. Verify that you can move order items when authenticated as a user with "orders" role or as the user who placed the order, but not when authenticated as someone else or when not authenticated.


aldeed:
@kieckhafer I did intentionally want an error if you try to have no items in a group. It's something we could consider supporting in the future, but I think it leads to a lot of unanswered questions and potential confusion as far as what to do about group status, totals, etc.

I will fix the `itemIds` issue. Beyond that, I think there is a whole can of worms here that I hadn't realized until now. We probably need to recalculate totals after moving, or they won't make sense. The group totals will change, which means discounts, shipping, taxes, and surcharges could all change, too.

I think I'll need to pull out the placeOrder calculations to a util function, so we can redo them as the order changes. (The payments will then not necessarily match the total, but that can be handled by refunds or adding additional payments, something we'll add support for in a future PR.)

aldeed:
@kieckhafer This should be ready for another look

PR #5019, opened 2019-03-06
Add updateOrder mutation

aldeed:
Part of #4999
Impact: **minor**  
Type: **feature**

## Changes
A new `updateOrder` mutation allows you to change the `email`, `status`, or `customFields` of an order after it has been placed. (Saving `customFields` is not enabled by default. See https://docs.reactioncommerce.com/docs/how-to-store-custom-order-fields)

- `updateOrder` can be called internally to synchronize from an external system by setting `context.isInternalCall` to `true`
- `updateOrder` can be called through GraphQL by any user with "orders" permission for the shop that owns the order

## Breaking changes
None

## Testing
Note: Do not worry about testing the `customFields` update since it requires customization to opt in.

1. Place an order.
1. Verify that you can update the order when authenticated as a user with "orders" role, but not when authenticated as someone else or when not authenticated.

### DIFF
--- a/imports/plugins/core/orders/server/no-meteor/resolvers/Mutation/updateOrder.js
+++ b/imports/plugins/core/orders/server/no-meteor/resolvers/Mutation/updateOrder.js
@@ -1,0 +1,38 @@
+import { decodeOrderOpaqueId } from "@reactioncommerce/reaction-graphql-xforms/order";
+
+/**
+ * @name Mutation/updateOrder
+ * @method
+ * @memberof Payments/GraphQL
+ * @summary resolver for the updateOrder GraphQL mutation
+ * @param {Object} parentResult - unused
+ * @param {Object} args.input - an object of all mutation arguments that were sent by the client
+ * @param {Object} [args.input.customFields] - Updated custom fields
+ * @param {String} [args.input.email] - Set this as the order email
+ * @param {String} args.input.orderId - The order ID
+ * @param {String} [args.input.status] - Set this as the current order status
+ * @param {String} [args.input.clientMutationId] - An optional string identifying the mutation call
+ * @param {Object} context - an object containing the per-request state
+ * @return {Promise<Object>} UpdateOrderPayload
+ */
+export default async function updateOrder(parentResult, { input }, context) {
+  const {
+    clientMutationId = null,
+    customFields,
+    email,
+    orderId,
+    status
+  } = input;
+
+  const { order } = await context.mutations.updateOrder(context, {
+    customFields,
+    email,
+    orderId: decodeOrderOpaqueId(orderId),
+    status
+  });
+
+  return {
+    clientMutationId,
+    order
+  };
+}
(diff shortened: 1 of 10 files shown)


tests/order/UpdateOrderMutation.graphql
```diff
@@ -0,0 +1,12 @@
+mutation ($email: String, $orderId: ID!, $status: String) {
```
kieckhafer: Does `customFields` need to be added in the mutation here, or is that going to ben an extension in its' own package?

imports/plugins/core/orders/server/no-meteor/schemas/schema.graphql
```diff
@@ -401,6 +406,30 @@ input PlaceOrderInput {
   payments: [PaymentInput]
 }
 
+"Input for the updateOrder mutation"
+input UpdateOrderInput {
```
kieckhafer: Does `customFields` need to be added in the mutation here, or is that going to ben an extension in its' own package?

kieckhafer:
Works as described.

Can update order only from `orders` permission user. Cannot update even by the user that placed the order.

A few comments, but overall good to go.

tests/order/UpdateOrderMutation.graphql
aldeed: It will need to be added if used because we don't know what the schema should be.

imports/plugins/core/orders/server/no-meteor/schemas/schema.graphql
aldeed: It will need to be added if used because we don't know what the schema should be.

aldeed:
@kieckhafer I do think eventually we should check against a status list, but we'll soon have to do some work to make that list configurable. I thought about adding a check now with a hard-coded array but it seems low risk to leave it wide open for now, and I don't know for sure that people aren't already storing custom statuses that would not be valid according to our list.

imports/plugins/core/orders/server/no-meteor/resolvers/Mutation/updateOrder.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @name Mutation/updateOrder
+ * @method
+ * @memberof Payments/GraphQL
+ * @summary resolver for the updateOrder GraphQL mutation
+ * @param {Object} parentResult - unused
+ * @param {Object} args.input - an object of all mutation arguments that were sent by the client
+ * @param {Object} [args.input.customFields] - Updated custom fields
+ * @param {String} [args.input.email] - Set this as the order email
```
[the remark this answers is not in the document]
aldeed: Are you referring to the `[]` around them? That means optional in [jsdoc](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values) and they are in fact all optional except `orderId`. Ideally you would provide as least one thing to update, but if you don't, I just have it return the un-updated order without throwing an error.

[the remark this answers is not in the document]
kieckhafer: Got it... Yes that is what I was referring to.

aldeed:
Also, @kieckhafer all of your commits and conflict fixes LGTM. Thanks!

PR #5020, opened 2019-03-06
Add updateOrderFulfillmentGroup mutation

aldeed:
Part of #4999
Impact: **minor**  
Type: **feature**

## Changes
A new `updateOrderFulfillmentGroup` mutation allows you to change the `tracking` and `status` fields of an order fulfillment group after the order has been placed.

The operator UI now uses this mutation for the `tracking` update only. Changing the operator UI to use GraphQL for group status updates will be done in a later PR.

- `updateOrderFulfillmentGroup` can be called internally to synchronize from an external system by setting `context.isInternalCall` to `true`
- `updateOrderFulfillmentGroup` can be called through GraphQL by any user with "orders" permission for the shop that owns the order
- The "orders/updateShipmentTracking" Meteor method is removed in favor of the `updateOrderFulfillmentGroup` GraphQL mutation

## Breaking changes
None

## Testing
1. Place an order.
1. Verify that you can update the order fulfillment group status and tracking when authenticated as a user with "orders" role, but not when authenticated as someone else or when not authenticated.
1. Verify that saving the Tracking field in the operator UI works.


aldeed:
@kieckhafer I add `trackingUrl` in https://github.com/reactioncommerce/reaction/pull/5020/commits/1c86ee9da539b45b96ec2b0368e8636a61bb038d per discussion in the other PR

imports/plugins/core/orders/client/graphql/index.js
```diff
@@ -28,3 +31,28 @@ export const captureOrderPayments = async ({ orderId, paymentIds, shopId }) => {
     }
   });
 };
+
+export const updateOrderFulfillmentGroup = async ({
+  orderId,
+  orderFulfillmentGroupId,
+  status,
+  tracking
```
kieckhafer: See that you noted trackingUrl is not updated here yet. Should we go ahead and add it already to keep this in sync with what the other mutation can do?

aldeed: I'd rather wait until we actually add a field in UI for it and need it here.

PR #5024, opened 2019-03-07
Add splitOrderItem mutation

aldeed:
Part of #4999
Impact: **minor**  
Type: **feature**

## Changes
A new `splitOrderItem` mutation allows you to split one order item into two items, such that the two items are identical other than their quantity and subtotal, and their quantities add up to the original item's quantity.

- `splitOrderItem` can be called internally to synchronize from an external system by setting `context.isInternalCall` to `true`
- `splitOrderItem` can be called through GraphQL by any user with "orders" permission for the shop that owns the order

## Breaking changes
None

## Testing
1. Place an order.
1. Verify that you can split an order item when authenticated as a user with "orders" role, but not when authenticated as someone else or when not authenticated.


aldeed:
@kieckhafer The tax and other total recalculation issues should be fixed now

PR #5029, opened 2019-03-11
(feat) Allow plugins to override Order Id generation

### LINKED ISSUE #5052: Use single `referenceId` from cart => order
It is a common request that we need to have a single transaction ID that remains consistent from payment all the way through to order export.

It has been decided that the way to do this is to add a `referenceId` to cart (the creation of which can be customized from a plugin (similar to [here](https://github.com/reactioncommerce/reaction/pull/5029)). Then when an order is created the `order.referenceID` would be the cart `referenceId`.

Then the app needs to be audited to make sure that wherever we display a reference ID to the customer or to an admin (e.g. in confirmation emails or the order admin) it is always the same `order.referenceId`.

zenweasel:
Resolves #5052 
Impact: **major**  
Type: **feature**

## Issue
Currently we use the Random id to generate Order Ids. A lot of systems have requirements for IDs to be of a different type or in a specific order (e.g. consecutive integers). This allows a plugin to override this by providing a `createOrderReferenceId` and a `createCartReferenceId` function. The default method of using `Random.id()` remains the same.

It also adds a `referenceId` to the cart

## Solution
Allow for a `createOrderId` function to be provided by a plugin and use that instead of Random.id

## Breaking changes
None

## Testing
1. Create an order and verify it has an order id
1. Create a test plugin in `imports/plugins/custom/fake` with just this file

register.js
```
import _ from "lodash";

function createNumericId() {
  console.log("createdNumericId");
  return _.random(0, 1000).toString();
}

Reaction.registerPackage({
  label: "Fake",
  name: "reaction-fake",
  icon: "fa fa-fake",
  autoEnable: true,
  functionsByType: {
    createCartReferenceId: [createNumericId]
  }
});
```

3. Restart Reaction
4. Create an order
5. Verify that the order Id is a number between 0 and 1000.


zenweasel:
@aldeed Made the change

zenweasel:
@aldeed Changes made

manueldelreal:
Screen Shot 2019-03-15 at 3 52 54 PM
regular reaction order id


Screen Shot 2019-03-15 at 4 51 05 PM
@zenweasel changes order id

No issues seem to be introduced by this

zenweasel:
I need to change this to pass in the cart and order to the function in case people want to generate this ID from either of those entities.

zenweasel:
@aldeed I've added overriding the cart ID. I had to add code to fetch cart in placeOrder. Let me know if that works. Thanks!

imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
```diff
@@ -261,6 +262,27 @@ export default async function placeOrder(context, input) {
     }
   };
 
+  let referenceId;
+  // Grabbing cart to get the cart.referenceId. It's a shame since we are fetch cart twice before
+  const cart = await Cart.findOne({ _id: cartId }, { _id: 0, referenceId: 1});
```
aldeed: I would move this `findOne` to above the discounts block, and update `getDiscountsTotalForCart` function to have the `cart` passed in rather than just the cart ID. `getDiscountsTotalForCart` is called only in two places, and in the other place (discounts startup code), the full `cart` is already available. So this change will prevent a duplicate cart lookup in two places.

Also, you could check but I don't think `_id: 0` does anything when combined with other fields that are `1`. They may have changed this in more recent Mongo versions.

aldeed: Two more things:

Add this after the `findOne` line:

```js
if (!cart) {
  throw new ReactionError("not-found", "Cart not found");
}
```

And `cartId` isn't required so you need to check it. If falsy, you can skip the `findOne`. And do something like `if (cart && cart.referenceId)` below, or somehow make sure `cart` is defined.

imports/plugins/core/versions/server/migrations/58_add_reference_id_to_order.js
```diff
@@ -0,0 +1,18 @@
...
+import * as collections from "/lib/collections";
+
+
+Migrations.add({
+  version: 58,
+  up() {
+    // In case any orders don't have a reference ID
+    collections.Orders.find({ "referenceId" : { "$exists" : false }}).forEach((order) => {
+      collections.Orders.findOneAndUpdate({ _id: order._id }, { $set: { referenceId: order._id }})
+    });
```
aldeed: This was already done in migration 47.

imports/plugins/core/versions/server/migrations/58_add_reference_id_to_order.js
```diff
@@ -0,0 +1,18 @@
...
+
+    // In case we have existing carts
+    collections.Cart.find({ "referenceId" : { "$exists" : false }}).forEach((cart) => {
+      collections.Cart.findOneAndUpdate({ _id: cart._id }, { $set: { referenceId: cart._id }})
+    })
```
aldeed: It's best to use our `findAndConvertInBatches` util function since these collections can be large. This was already done for orders in migration 47. I would do only cart here, which means you could simply copy the `47_order_ref.js` file contents into here and change `Orders` to `Cart` and `order` to `cart` everywhere.

aldeed:
@zenweasel One more thing. There is a second place where carts are created: `/imports/plugins/core/cart/server/no-meteor/mutations/convertAnonymousCartToNewAccountCart.js`

In this file I think you will want to do `referenceId: anonymousCart.referenceId` if that sounds right to you.

imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
zenweasel: @aldeed I had considered just leveraging the `getDiscountsTotalsForCart` but it felt like a bad separations of concerns, and somebody could eliminate or change that method and not realize the side effects. It seems like the better change would be to grab the cart and modify that method to take a passed in cart.

zenweasel: I don't understand, how is `cartId` not required here

zenweasel: As far as I know `_id: 0` has worked with every version of Mongo I have used at least

zenweasel:
Closing in favor of https://github.com/reactioncommerce/reaction/pull/5054

imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
aldeed: You're right. I just tried it. I know there was an early version where if you had any `1` fields, the `0` fields were ignored. But I don't think I've tested it since 5 or 6 years ago.

PR #5030, opened 2019-03-11
[WIP] (feat) Allow disabling search via ENV variable

zenweasel:
Resolves #NA
Impact: **minor**  
Type: **feature**

## Issue
Even if you have another method enabled the Mongo search is still listening to events and keeping the Search collections up to date. This could involve a lot of extra disk ops and extra storage space that is not needed

## Solution
Add an env variable called `DISABLE_MONGO_SEARCH` that acts like a feature flag and disables all Mongo Search operations

## Breaking changes
None. Mongo Search remains on by default until you disable it.

## Testing
1. Start app with the env not set, ensure that everything operates as normal.
1. Set `DISABLE_MONGO_SEARCH` to `true` and reset the database.
1. Ensure that the `ProductSearch` collection is not created.

### DIFF
--- a/imports/plugins/included/search-mongo/server/config.js
+++ b/imports/plugins/included/search-mongo/server/config.js
@@ -1,0 +1,5 @@
+import envalid, { bool } from "envalid";
+
+export const env = envalid.cleanEnv(process.env, {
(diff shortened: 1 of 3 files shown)


zenweasel:
Close in favor of disabling the plugin at the import level

Issue #4996, opened 2019-02-18
[EPIC] Moving the API off Meteor


aldeed:
We have been slowly removing Meteor dependencies from Reaction API server code. We are now close enough to being done on the server side, that we can create this Epic to track the remaining work.

To be clear, the goal here is to get the _API server_ running as a non-Meteor Node app. This means all GraphQL code, all events, all background jobs, all relevant startup code. It does _not_ include necessarily all Meteor methods or publications, nor does it include operator UI client code. There will remain a Meteor app to run those things for some time yet, but it will be needed only for running the operator UI. The operator UI will eventually not need Meteor, but there is much work to be done there and no timeline yet.

Here is a preliminary list of the work that this Epic entails:
- [x] Move all SimpleSchema indexes to use `collectionIndex` function, and remove `aldeed:schema-index` Meteor package
    - Issue #5080
- [x] Add a standard way to register collections and their indexes
    - Issue #4268
    - PR #5196
- [x] Add SimpleSchemas on `context`
    - Issue #5335 
- [x] De-Meteorize emailing
    - PR #4998
- [x] De-Meteorize the job-collection job queue
    - Issue #5288 
- [x] Review core startup code
    - [x] PR #5212 
- [x] Non-meteor translations loading
  - [x] PR #5514 
- [x] Deal with template-type assets as in `core/server/hooks.js`
  - [x] PR #5430
- [x] Replace `setAppVersion` on startup with getting the data from the new `systemInfo` GraphQL
   - [x] PR #5429
- [x] Create a `registerPackage` equivalent with no Meteor deps
- [x] Update all plugins to use the non-Meteor `registerPackage`
- [x] Move calling of registerPackageHandlers into NodeApp, or part of `registerPackage` replacement.
- [x] Remove all use of `context.callMeteorMethod` and remove it
  - [x] Issue #5336
  - [x] Issue #5337 - PR #5520 
  - [x] Issue #5338 - PR #5492 
  - [x] Issue #5339 - PR #5493
  - [x] Issue #5340 - PR: #5488 
  - [x] Issue #5341 - PR: #5484 
  - [x] Removal PR: #5524
- [x] Remove all use of `getGraphQLContextInMeteorMethod` and remove it ( #5529 )
  - [x] `inviteShopMember.js` - #5520 
  - [x] `inviteShopOwner.js` - #5553 
  - [x] `sendResetPasswordEmail.js` - #5558 
  - [x] `sendVerificationEmail.js` - TODO
[...]

PR #5031, opened 2019-03-11
Allow users to disable plugins via config

zenweasel:
Resolves #NA
Impact: **minor**  
Type: **feature**

## Issue
There is no good and easy way to disable a plugin through configuration

## Solution
Keep a list of disabled plugins and don't load/import any of those

## Breaking changes
None

## Testing
1. Start Reaction and observe no problems
1. Edit `.reaction/pluginConfigs` and add `search-mongo` to the array for `disabledPlugins`.
1. Restart Reaction
1. Observe that the mongo search is disabled


.reaction/pluginConfig.js
```diff
@@ -0,0 +1,5 @@
+const disabledPlugins = [];
```
aldeed: Can you add a comment above here explaining what goes in the array? At first glance I would have thought the plugin's name, but it looks like actually the directory name.

Also, should we gitignore this file? (after this initial commit)

zenweasel:
@aldeed I wonder if sort of tangential to this if we should move all the collection creation stuff for `search-mongo` so that those Collections are never created if the plugin is disabled?

aldeed:
@zenweasel Yes as a general rule we need to get all collection creation moved into the related plugin startup. It would be done soon as part of https://github.com/reactioncommerce/reaction/issues/4996 if not now

zenweasel:
@aldeed Requested changes made

PR #5033, opened 2019-03-11
Fix: startup asset provisioning

### LINKED ISSUE #5032: Plugin assets are not copied over to private directory
## Prerequisites

## Issue Description
A custom plugin should be able to contain a `private` directory that is copied over to the application's `private` directory on startup. It looks like this no longer happens. The issue can be traced to `.reaction/run/provisionAssets.mjs`, line 39, which is calling `fs.copySync` - a function that does not exist.

## Steps to Reproduce
1. Create a custom plugin with a `private` directory
2. After line 35 in the file mentioned above, add `console.log("sourceDir", sourceDir)`
3. Startup Reaction. Observe your custom plugin's directory being listed, followed by an error about `fs.copyFile`

## Possible Solution
Node does have `fs.copyFileSync` but this doesn't work for directories. Write a function that recursively copies the plugin's `private` directory and files over.

dancastellon:
Resolves #5032 
Impact: **minor**  
Type: **bugfix**

## Issue
I discovered asset provisioning was broken when trying to create custom email templates by placing them in my plugin's `private` folder.

## Solution
[This commit](https://github.com/reactioncommerce/reaction/commit/57d1b3620601991679adadb6333b37dd0ab7d2bf) copied logic from CLI to allow Reaction to start without it. In doing so, the `fs-extra` package that CLI uses wasn't included. `fs-extra` provides the `copySync` function that recursively copies a directory. Without it, startup was throwing an error about `copySync` not existing, and custom plugin `private` directories weren't copied over to app's `private` directory. The fix was to simply add the package and change an import line.

## Breaking changes
None

## Testing
1. Create a custom plugin with a `private` directory
2. Startup Reaction and confirm no errors after `Provisioning assets...`


dancastellon:
@aldeed All set here

dancastellon:
@aldeed Signed off

PR #5035, opened 2019-03-11
Chore: Custom env setup

ticean:
Resolves: NA
Impact: **minor**  
Type: **chore**

## Issue
The `bin/setup` script is hardcoded for a certain use case - copying `.env.example` files to the `.env` file.

## Solution
Make the setup script more flexible. `bin/setup-env` now accepts inputs for defining the input and output files. `setup` now wraps this general script for plugins. This composes better for any use case. For example, it could be used for sourcing configuration from external config shared by a development team.

This implementation moves the functionality in `bin/setup` to `bin/setup-env` to preserve any documentation, so there will be no breaking changes anywhere, even in the docs. But, this means that the `setup` might end up being different than `setup` in other projects because we're `find`ing all `.env.example` files.

## Breaking changes
None.

## Testing
1. List the steps needed for testing your change in this section.
2. Assume that testers already know how to start the app, and do the basic setup tasks.
3. Be detailed enough that someone can work through it without being too granular


bin/setup
```diff
@@ -13,39 +15,6 @@ set -o posix    # more strict failures in subshells
...
-          echo "$var" >>"${env_file}"
-        fi
-      done
-    }
-}
-
-main
+find . -name .env.example -type f -print0 \
+  | xargs -0 -I '{source}' \
+    setup '{source}' .env
```
focusaurus: I'm worried about `PATH` here. Should we `cd` to a known dir and then run setup with a relative path like `./bin/setup`?

ticean: 👍 👍 I didn't battle harden this very well. Just tried to open it as quickly as I could for discussion.

ticean:
Closing as inactive.

PR #5042, opened 2019-03-14
FIX: Unable to checkout with taxes

ajporlante:
Impact: **major**  
Type: **bugfix**

## Issue
It is currently impossible to checkout because of schema validation issue here:

https://github.com/reactioncommerce/reaction/blob/b7c6c207a431e511857f66beec64755a59bc4409/imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js#L577

This validation always fails because the required fields here are not provided:

https://github.com/reactioncommerce/reaction/blob/b7c6c207a431e511857f66beec64755a59bc4409/imports/plugins/core/taxes/lib/extendCoreSchemas.js#L58-L81

 since the following lines were removed:

https://github.com/reactioncommerce/reaction/blob/b7c6c207a431e511857f66beec64755a59bc4409/imports/plugins/core/orders/server/util/addTaxesToGroup.js#L30-L35

Also, some fields needed by SE Team to be saved in an order were removed.

## Solution

I put back tax data per item and added a blackbox `taxData` field to invoice schema to allow saving of custom fields returned by tax service in a fulfillment group..

## Testing
You should be able to checkout.

### DIFF
--- a/imports/collections/schemas/payments.js
+++ b/imports/collections/schemas/payments.js
@@ -45,6 +45,11 @@ export const Invoice = new SimpleSchema({
     type: Number,
     min: 0
   },
+  taxData: {
+    type: Object,
+    blackbox: true,
+    optional: true
+  },
   total: {
     type: Number,
     min: 0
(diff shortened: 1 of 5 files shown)


aldeed:
@ajporlante There is a simpler fix. I have it in https://github.com/reactioncommerce/reaction/pull/5018 already, so we can either get that merged or cherry pick https://github.com/reactioncommerce/reaction/pull/5018/commits/fdc1cbfb16bc56588ec1de14d5a0e284cd6eae19

aldeed:
There is an alternative PR here: https://github.com/reactioncommerce/reaction/pull/5043

I know this fixes the main checkout error I was seeing, but I'll leave this open for now in case there's still some additional issue.

PR #5050, opened 2019-03-18
chore: add optional first and last name fields to schemas

kieckhafer:
Impact: **minor**  
Type: **chore**

## Issue
Some payment integrations require `firstName` and `lastName` fields to be separated, not just a `fullName` as we currently have set up.

## Solution
Add optional `firstName` and `lastName` fields to our meteor simpl schema for `Address`. This schema is also use don payments. Also, add the fields to the GQL schema for `Address` and `AddressInput`, which again is used for payments and all places `fullName` is required.

As per @spencern's comment... since `fullName` is required,  and `firstName` and `lastName` are optional, we leave it up to the client how to send the data, whether thats just using `fullName`, or taking the input from `firstName` and `lastName` and combining them to create a `fullName`.

## Breaking changes
None

## Testing
1. Use GQL to add an address with first name and last name
```
mutation {
  addAccountAddressBookEntry(
    input: {
      accountId: "cmVhY3Rpb24vYWNjb3VudDpvTldwRFdBa3hlcUxQU3hpUw==",
      address: {
        address1: "123 Main Street Street"
        firstName: "Erik"
        lastName: "Kieckhafer"
        city: "Santa Monica"
        region: "CA"
        country: "US"
        postal: "90405"
        phone: "555-666-7890"
        fullName: "Erik Kieckhafer"
      }
    }
  ) {
	address {
          fullName
          firstName
          lastName
  }
  }
}
```
1. See that they are both saved in the database
1. See that you can query `firstName`, `lastName`, and `fullName` wherever you need an address

### DIFF
--- a/imports/collections/schemas/address.js
+++ b/imports/collections/schemas/address.js
@@ -50,7 +50,7 @@ export const Address = new SimpleSchema({
   },
   "lastName": {
     type: String,
-    label: "Laast name",
+    label: "Last name",
     optional: true
   },
   "address1": {
--- a/imports/collections/schemas/address.js
+++ b/imports/collections/schemas/address.js
@@ -17,6 +17,8 @@ const withoutCodeCountries = ["AO", "AG", "AW", "BS", "BZ", "BJ", "BW",
  * @type {SimpleSchema}
  * @property {String} _id
  * @property {String} fullName required
+ * @property {String} fistName
+ * @property {String} lastName
  * @property {String} address1 required
  * @property {String} address2
  * @property {String} city required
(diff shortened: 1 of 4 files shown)


imports/collections/schemas/address.js
```diff
@@ -41,6 +43,16 @@ export const Address = new SimpleSchema({
...
     label: "Full name"
   },
+  "firstName": {
+    type: String,
+    label: "First name",
+    optional: true
+  },
+  "lastName": {
+    type: String,
+    label: "Laast name",
```
mikemurray: `Last`

imports/collections/schemas/address.js
```diff
@@ -17,6 +17,8 @@ const withoutCodeCountries = ["AO", "AG", "AW", "BS", "BZ", "BJ", "BW",
  * @type {SimpleSchema}
  * @property {String} _id
  * @property {String} fullName required
+ * @property {String} fistName
```
aldeed: @kieckhafer Just pointing out this misspelling in case you want to fix in the next PR you do.

kieckhafer: Thanks, fixed it in https://github.com/reactioncommerce/reaction/pull/5071

PR #5053, opened 2019-03-19
feat: remove search plugins

spencern:
Impact: **breaking**  
Type: **feature**

## Issue
This PR removes the `search-mongo` and `ui-search` plugins from the `imports/plugins/included` directory of `reaction`. It also removes the search bar from the operator orders UI that depended upon the `search-mongo`. It cleans up the util functions run by some of the migrations used to setup the mongo search collections.

We're doing this because we've found that some of the app event hooks within this plugin fire even when the plugin is disabled. Having these events run can cause considerable slowness when publishing a product to the catalog. This issue is exacerbated when importing and publishing large sets of products simultaneously.

When updating our storefront and operator UIs, we've moved away from using this search package. While there are better search options for most users of Reaction, such as ElasticSearch, we do plan to support basic search functionality out of the box for both the storefront and for the operator UIs. 

The operator order management UI is very sluggish. This PR does _not_ attempt to resolve any issues related to the order management UI. 

## Breaking changes
Removes order search bar from operator order UI.
Removes `search-mongo` and `ui-search` plugins from `reaction`.

## Testing
1. Ensure the app builds successfully from a fresh install or from an existing install.
2. Check the operator orders panel. Ensure orders load and that the filters continue to work.


MarcusSorealheis:
We should consider adding [MongoDB Atlas Search](https://www.mongodb.com/atlas/search) to the framework. It's free for most people and is a fast-growing and simple to use alternative to text matching in MongoDB.

PR #5054, opened 2019-03-19
(feat) Allow extend generate orderId and cartId

### LINKED ISSUE #5052: Use single `referenceId` from cart => order

zenweasel:
**Branch Recreated in order to Sign Commits**
** Replaces PR https://github.com/reactioncommerce/reaction/pull/5029 **

Resolves #5052
Impact: **minor**  
Type: **feature**

## Issue
Currently we use the Random id to generate Order Ids. A lot of systems have requirements for IDs to be of a different type or in a specific order (e.g. consecutive integers). This allows a plugin to override this by providing a `createOrderId` function. The default method of using `Random.id()` remains the same.

## Solution
Allow for a `createOrderReferenceId` and `createCartReferenceId` function to be provided by a plugin and use that instead of Random.id

## Breaking changes
None

## Testing
1. Create an order and verify it has an referenceId
1. Create a test plugin in `imports/plugins/custom/fake` with just this file

register.js
```
import _ from "lodash";

function createNumericId() {
  console.log("createdNumericId");
  return _.random(0, 1000).toString();
}

Reaction.registerPackage({
  label: "Fake",
  name: "reaction-fake",
  icon: "fa fa-fake",
  autoEnable: true,
  functionsByType: {
    createCartReferenceId: [createNumericId]
  }
});
```

3. Restart Reaction
4. Add something to cart
5. Verify it was a reference id that is between 0 and 1000
6. Checkout with this cart
7. Verify that `order.referenceId` is the same as the number above

### DIFF
--- a/imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
+++ b/imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
@@ -174,8 +174,10 @@ export default async function placeOrder(context, input) {
   } = orderInput;
   const { accountId, account, collections, getFunctionsOfType, userId } = context;
   const { Orders, Cart } = collections;
+
+  let cart;
   if (cartId) {
-    const cart = await Cart.findOne({ _id: cartId });
+    cart = await Cart.findOne({ _id: cartId });
     if (!cart) {
       throw new ReactionError("not-found", "Cart not found while trying to place order");
     }
@@ -275,7 +277,7 @@ export default async function placeOrder(context, input) {
   if (!createReferenceIdFunctions || createReferenceIdFunctions.length === 0) {
     // if the cart has a reference Id, and no custom function is created use that
     if (_.get(cart, "referenceId")) { // we want the else to fallthrough if no cart to keep the if/else logic simple
-      referenceId = cart.referenceId;
+      ({ referenceId } = cart);
     } else {
       referenceId = Random.id();
     }
--- a/imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
+++ b/imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
@@ -282,7 +282,7 @@ export default async function placeOrder(context, input) {
       referenceId = Random.id();
     }
   } else {
-    referenceId = await createReferenceIdFunctions[0](order, cart);
+    referenceId = await createReferenceIdFunctions[0](context, order, cart);
     if (typeof referenceId !== "string") {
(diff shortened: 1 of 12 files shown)


zenweasel:
I believed I resolved all sign-off issues but DCO is still failing 🤷‍♂️

aldeed:
I left comments on #5029

zenweasel:
Resolve code-signing issue by interactive rebase

zenweasel:
@aldeed Ready for another look

imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
```diff
@@ -261,6 +272,25 @@ export default async function placeOrder(context, input) {
...
+  const createReferenceIdFunctions = getFunctionsOfType("createOrderReferenceId");
+  if (!createReferenceIdFunctions || createReferenceIdFunctions.length === 0) {
+    // if the cart has a reference Id, and no custom function is created use that
+    if (_.get(cart, "referenceId")) { // we want the else to fallthrough if no cart to keep the if/else logic simple
+      ({ referenceId } = cart);
+    } else {
+      referenceId = Random.id();
+    }
+  } else {
+    referenceId = createReferenceIdFunctions[0](cart, order);
```
aldeed: Can you pass `order` arg before `cart` arg since `cart` may be undefined and some people may not need it?

aldeed: Actually a couple more things:

1. Do `= await createReferenceIdFunctions[0](order, cart)` (add `await`) in case the plugin needs to do async calls to generate an ID.
2. Specifically check `referenceId` after this and if it isn't a string, throw an error with a message something like `"createOrderReferenceId" function returned invalid ID`. The schema check later would catch this, but if that happened, it would be a generic error and would take some investigating to figure out that it's an issue with the registered function.

Do both of these things for the cart ID function, too.

aldeed:
There is also one unit test for createCart that needs `referenceId: jasmine.any(String)` added to the expected object to get the test passing again.

zenweasel:
@aldeed Ready for yet another review

PR #5055, opened 2019-03-19
Refactor #5025: Add env. vars. for graphql playground and introspection

### LINKED ISSUE #5025: Add ENV for GraphQL Playground and introspection
## Background

https://www.apollographql.com/docs/apollo-server/features/graphql-playground.html#Enabling-GraphQL-Playground-in-production

Currently Reaction uses default Apollo behavior:

> In development, Apollo Server enables GraphQL Playground on the same URL as the GraphQL server itself (e.g. http://localhost:4000/graphql) and automatically serves the GUI to web browsers. When NODE_ENV is set to production, GraphQL Playground (as well as introspection) is disabled as a production best-practice.

Reaction prefers to use individual env variables rather than NODE_ENV.

## Work
- In `createApolloServer.js`, add the `introspection` and `playground` options with value coming from new ENV variables
- Use `envalid` for parsing the variables. Default should be `false` for both.
- Add both variables to `.env.example` with values of `true` for both
- Variable names should be something like `GRAPHQL_INTROSPECTION_ENABLED` and `GRAPHQL_PLAYGROUND_ENABLED`

brunoLehnen:
Resolves #5025   
Impact: **minor**  
Type: **refactor**

## Issue
In development, Apollo Server enables GraphQL Playground on the same URL as the GraphQL server itself (e.g. http://localhost:4000/graphql) and automatically serves the GUI to web browsers. When NODE_ENV is set to production, GraphQL Playground (as well as introspection) is disabled as a production best-practice.

## Solution
Set environment variables with "envalid" in order to determine whether GraphQL Playground should be enabled or not in any environment. 

## Breaking changes
These changes are all pretty harmless, the only exception to that which has the capacity of causing an error is adding invalid boolean values to the newly added environment vars. (GRAPHQL_INTROSPECTION_ENABLED and  GRAPHQL_PLAYGROUND_ENABLED).

**FYI** - Following are the accepted values for boolean environment variables in the "envalid" lib. :  "0", "1", "true", "false", "t", "f"

## Testing
1. Set the GRAPHQL_INTROSPECTION_ENABLED and  GRAPHQL_PLAYGROUND_ENABLED;
2. Access GraphQL Playgrand on the same URL as the GraphQL server.


imports/node-app/core/createApolloServer.js
```diff
@@ -59,7 +64,9 @@ export default function createApolloServer(options = {}) {
     schema,
     subscriptions: {
       path
-    }
+    },
+    instrospection: env.GRAPHQL_INSTROSPECTION_ENABLED,
```
aldeed: `INTROSPECTION` and `introspection` are misspelled here, so have no effect.

imports/node-app/core/createApolloServer.js
```diff
@@ -31,6 +32,10 @@ export default function createApolloServer(options = {}) {
   const schemas = options.schemas || [];
   const schemasToMerge = schemas.filter((td) => typeof td !== "string");
   const typeDefs = schemas.filter((td) => typeof td === "string");
+  const env = envalid.cleanEnv(process.env, {
+    GRAPHQL_INTROSPECTION_ENABLED: bool({ default: false }),
+    GRAPHQL_PLAYGROUND_ENABLED: bool({ default: false })
+  });
```
aldeed: Please add a new file `/imports/node-app/core/config.js` and move the `cleanEnv` call plus the `envalid` imports into that file. Export the cleaned config from there.

Then do `import config from "./config";` in this file and `config.GRAPHQL_INTROSPECTION_ENABLED`, etc.

This pattern was decided recently but not yet added to the Reaction public docs so you could not have known.

.env.example
```diff
@@ -13,3 +13,5 @@ REACTION_USER_NAME=admin
 REACTION_USER=Admin
 ROOT_URL=http://localhost:3000
 IDENTITY_PROVIDER_MODE=all
+GRAPHQL_INTROSPECTION_ENABLED=true
+GRAPHQL_PLAYGROUND_ENABLED=true
```
aldeed: Please also move these up so that this file is alphabetized.

imports/node-app/core/createApolloServer.js
brunoLehnen: Oh, sorry about that. Will fix.

imports/node-app/core/createApolloServer.js
brunoLehnen: Will do.

.env.example
brunoLehnen: Will do.

imports/node-app/core/createApolloServer.js
brunoLehnen: Fixed

imports/node-app/core/createApolloServer.js
brunoLehnen: Fixed.

.env.example
brunoLehnen: Fixed.

brunoLehnen:
@aldeed, thanks a million for reviewing my PR and for letting me contribute to the project. 
Please let me know if there's anything I can help you guys with :+1:

aldeed:
This is ready to merge, but I need to figure out how we can update the integration tests to continue passing. Envalid does not like the value of NODE_ENV when tests are running it seems.

brunoLehnen:
That's Great @aldeed! Also, if you want, I can help you out with those failing tests too, just let me know :+1:

PR #5058, opened 2019-03-20
fix marketplace invitation validation

### LINKED ISSUE #5057: Invite markeplace owner - validation error: Order status labels is required
## Issue Description
When invitation to new markeplace owner is sent. We are getting error: Order status labels is required.

## Steps to Reproduce
1. Logged in as admin
2. Go to Shop menu
3. Turn on markeplace switch
4. Invite new markeplace owner

marketplace

## Possible Solution
In file reaction/imports/plugins/core/orders/lib/extendShopSchema.js there is missing parameter for "optional: true". This was assumed while property was described as optional before (@property {Object} orderStatusLabels optional).

```
orderStatusLabels: {
    type: Object,
    blackbox: true,
    optional: true
  }
```

### Versions
Node: 8.15.0
NPM: 6.7.0
Meteor Node: 8.11.4
Meteor NPM: 6.4.1
Reaction CLI: 0.29.0
Reaction: 2.0.0-rc.10
Reaction branch: develop
Docker: 18.06.1-ce

TilenKreca:
Resolves #5057
Impact: **major**  
Type: **bugfix**

## Issue
When invitation to new markeplace owner is sent. We are getting error: Order status labels is required.

## Solution
Is provided by this pull request.

## Breaking changes
No breaking changes or functional changes

## Testing
Probably this option need's to be added in tests: Inviting new marketplace owner.

### DIFF
--- a/imports/plugins/core/orders/lib/extendShopSchema.js
+++ b/imports/plugins/core/orders/lib/extendShopSchema.js
@@ -9,6 +9,7 @@ import { Shop } from "/imports/collections/schemas";
 Shop.extend({
   orderStatusLabels: {
     type: Object,
-    blackbox: true
+    blackbox: true,
+    optional: true
   }
 });


spencern:
Hi @TilenKreca 
Thanks for taking a look at this error and suggesting a fix. 
I'm curious why our marketplace owner invitation involves the Order schema at all. 
We've been less focused on our current implementation of the marketplace features lately, so it's possible something has slipped in that shouldn't have.

Would you mind investigating why this Schema is used when a marketplace owner invitation is sent?

TilenKreca:
Hi @spencern 

I see this was added like 1 month ago and according to commit message @kieckhafer  "extend Shops schema to allow for orderStatusLabel translations" I assume this has something to do with translations.

I am quite new in Reaction, so maybe it would be much better if someone from core developer's would look at.

spencern:
Thanks for following up. We'll take a look as time permits :)

mikemurray:
Confirmed with @kieckhafer that `orderStatusLabels` is indeed optional, so this change should be fine. I'll do a test of the app, and we should be good to go.

Issue #5062, opened 2019-03-22
Products in cart do not show correct image


willopez:
## Prerequisites
* [x] Are you running the latest version?
* [x] Are you able to consistently reproduce the issue?
* [x] Did you search the issue queue for existing issue? [Search issues](https://github.com/reactioncommerce/reaction/issues)

## Issue Description
The product in the cart is showing the image for the first variant of a product, even when I added a different variant.

## Steps to Reproduce
1. Add a variant other that the first one, to the shopping cart
2. Observe incorrect image in mini cart popover

willopez:
fix here: https://github.com/sportsdirect/reaction-plugin-cloudinary/pull/9

willopez:
closed via #5063

PR #5063, opened 2019-03-22
Set variant media correctly when added to cart

willopez:
Resolves: #5062
Impact: **Critical**
Type: **bugfix**

## Issue
Incorrect variant image rendered in mini cart popover and cart page

## Solution
Give priority to media in variant over media stored on the parent product object. 

## Testing
1. Add variant other than the first one to the shopping cart
2. Confirm the correct variant image is rendered in the cart popover
3. Confirm the correct variant images are rendered in the cart page

### DIFF
--- a/imports/plugins/core/graphql/server/no-meteor/xforms/cart.js
+++ b/imports/plugins/core/graphql/server/no-meteor/xforms/cart.js
@@ -51,20 +51,17 @@ async function xformCartItem(context, catalogItems, products, cartItem) {
   }
 
   const catalogProduct = catalogItem.product;
+
   const { variant } = findVariantInCatalogProduct(catalogProduct, variantId);
   if (!variant) {
     throw new ReactionError("invalid-param", `Product with ID ${productId} has no variant with ID ${variantId}`);
   }
 
-  const variantSourceProduct = products.find((product) => product._id === variantId);
-  const catalogVariantId = variantSourceProduct.ancestors[1];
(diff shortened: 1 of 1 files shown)


imports/plugins/core/graphql/server/no-meteor/xforms/cart.js
```diff
@@ -56,8 +56,17 @@ async function xformCartItem(context, catalogItems, products, cartItem) {
     throw new ReactionError("invalid-param", `Product with ID ${productId} has no variant with ID ${variantId}`);
   }
 
+  const variantSourceProduct = products.find((product) => product._id === variantId);
+  const catalogVariantId = variantSourceProduct.ancestors[1];
+  const catalogVariant = catalogProduct.variants.find((cVariant) => cVariant._id === catalogVariantId);
```
aldeed: `ancestors[1]` is not correct because sometimes there's only one item in the `ancestors` array. `variantId` here may be referring to a variant or option ID, and we've already found `variant` (which may be an option) with the `findVariantInCatalogProduct` call above.

So I think removing everything above here and changing `catalogVariant` to `variant` below will make this correct.

This would be a good time to add a unit test to prove this is working, too.

aldeed:
Also sign your commits

willopez:
@aldeed I have addressed your comments. In the interest of time, I would like to get this merged. I have created another issue for myself to add a unit test in another PR. https://github.com/reactioncommerce/reaction/issues/5086

PR #5064, opened 2019-03-23
feat: registerBlock API for registering blocks and rendering components for regions

### LINKED ISSUE #5038: Allow users to customize the PDP to include new fields w/o modifying core
# Summary

Users need to be able to add fields to the PDP so they can be modified by admins w/o having to edit the code 

## Justification

Currently users can extend the DB schema and the GraphQL schema without modifying core but these fields are not editable in the PDP. This feature used to exist but was removed at some point. Adding custom fields is one of the most common use cases and is no longer fully covered.

## Detail Requirements

A developer should be able to specify through code additional fields that should be added to the PDP. They should be able to specify:

1. The name of the field (going to assume that this is a field that's on an extended schema on `Product`)
1. The "Label" for the field (could also be derived from the simple-schema label)
1. What sort of control should be used (text box, select, etc)
1. If using a Select, what valid available options there are
1. For an initial implementation all custom fields could just be displayed in a separate "Custom Fields" are similar to Metadata

This is a temporary implementation until Catalog 2.0 is built out

mikemurray:
Resolves #5038  
Impact: **major**  
Type: **feature**

## Solution

- Using `registerComponent` as a starting point
- I created the `registerBlock` API which registers components.
- Each Block is assigned to a region, with a priority for ordering.
- Regions are then rendered with all the blocks registered to that region

**Register a block**
```js
// ProductDetail Block Region: Main
registerBlock({
  region: "ProductDetailMain", // Region name (Required)
  name: "ProductDetailForm", // Name of block, which will help with overriding  (Required)
  component: ProductDetailForm, // React component or function (Required)
  hocs: [withProductForm], // Any higher-order component you want to compose with this component
  priority: 10 // Position in region. All built-in blocks have priority increments of 10. Decimals work too.
});
```
**Render blocks**
```jsx
<Blocks
  region="ProductDetailMain" // Block region to render
  blockProps={props} // Props to pass to all blocks
/>
```

**Render blocks using the render prop**
 This allows you to wrap each block in a container and do any additional setup if you need
```jsx
<Blocks region="ProductDetailMain" blockProps={blockProps}>
  {(blocks) =>
    blocks.map((block, index) => (
      <div className={classes.block} key={index}>
        {block}
      </div>
    ))
  }
</Blocks>
```

**Docs will be writtien before this PR is ready for review**

## Breaking changes

- Green field flash on field save removed from most fields as `refs` are not allowed for functional components
- Product components have been broken into smaller peices and the old files removed. This should be an internal breaking change as prior to this PR, those components couldn't be modified without touching core.

## Testing
1. Ensure the product and variant forms still work in their entirety


machikoyasuda:
Made this Issue to track the docs for this ticket: https://github.com/reactioncommerce/reaction-docs/issues/758

mikemurray:
@aldeed 

Should I be registering blocks to be reliant on some props from the page, or register them to be more independent?

Right now theres a `withProduct` HOC that wraps the page and every registered block would have access to props from `withProduct`?

- `withProduct` contains the current product, and a list of all top-level variants.

```js
registerOperatorRoute({
  isNavigationLink: false,
  isSetting: false,
  path: "/products/:handle",
  mainComponent: ProductDetailLayout,
  hocs: [withProduct]
});
```

Or should each block be required to define it's own HOCs making them more independent.

- `withProduct` contains the current product, and a list of all top-level variants.
- `withProductForm` contains all of the state management and mutations for editing a product.

// ProductDetail Block Region: Main
registerBlock({
  region: "ProductDetailMain",
  name: "ProductDetailForm",
  component: ProductDetailForm,
  hocs: [withProduct, withProductForm],
  priority: 10
});

I can see the merit of both approaches, though I feel like total independence makes it easier to understand where props are coming from, especially if blocks are moved around to different regions.

aldeed:
@mikemurray In terms of functionality, I think it's important that it support either approach. The person adding `<Blocks />` to their component render should be able to decide to pass down `product` prop if they want, and they would document this fact (e.g. "All blocks rendered into the `ProductDetailMain` region will have `product` and `isLoadingProduct` props passed to them").

But I think we should also keep the `hocs` option, allowing any combination of the approaches.

Now that being said, we should have guidance about what you _should_ do, even if you _can_ do either. I'm not sure it's an easy answer. As you said, keeping things isolated is great, but replicating the same HOC in 10 places is not, especially if there isn't something like Apollo caching ensuring that the same data isn't actually fetched twice in a row. So our "guidance" will have to be something like "think carefully about the implications of the decision", plus some example cases to illustrate.

Otherwise this looks good. Do you think we should also add a `props` option, for any props to pass down that are known at the time of registering? If we don't have it, we could work around with a wrapper component, but might be nice.

aldeed:
For the product detail page example in particular, my vote would be to fetch product at the page level and always provide to all regions on the page. The reason is because it's a product page, so it's reasonable to assume that at least one if not all blocks will need the product. Whereas if a particular block also needs some other entity that others won't need, that's a good use case for block-level data loading with `hocs`.

mikemurray:
@aldeed Would the `props` option be applied before or after the HOCs?

mikemurray:
Added @aldeed to review the Block API
Added @manueldelreal to test and verify that the product form all still work, that you can create, edit, archive products, variants, and options.

imports/plugins/core/components/lib/blocks.js
```diff
@@ -0,0 +1,274 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Logger } from "/client/api";
+import { compose, setDisplayName } from "recompose";
+
+// export const Blocks = () ; // populated with final wrapped blocks
```
aldeed: What's this? Can it be removed?

imports/plugins/core/components/lib/blocks.js
```diff
@@ -0,0 +1,274 @@
...
+ * @summary Populate the final BlockComponents object with the contents of the lookup table.
+ * This should only be called once on app startup.
+ * @returns {Object} An object containing all of the registered blocks by region
+ * @memberof Components/Helpers
+ **/
+export function loadRegisteredBlocks() {
+  Object.keys(BlocksTable).forEach((regionName) => {
+    BlockComponents[regionName] = getBlocks(regionName);
+  });
+  return BlockComponents;
```
aldeed: Is there a reason this needs to be run on startup? Couldn't everything in here and `getBlocks` just happen within `registerBlock` directly? I'd like to avoid adding startup code whenever possible.

imports/plugins/included/product-admin/client/blocks/VariantTaxForm.js
```diff
@@ -0,0 +1,103 @@
+import React from "react";
```
aldeed: Technically this block should be added by the core `taxes` plugin because that's what extends the product schema to add these fields. But if it isn't easy to move it right now, it's fine to punt that.

imports/plugins/included/product-admin/client/hocs/withProduct.js
```diff
@@ -138,21 +134,6 @@ const wrapComponent = (Comp) => {
...
-            Meteor.call("products/updateMetaFields", productId, metafield, index);
-          } else if (metafield.key && metafield.value) {
-            Meteor.call("products/updateMetaFields", productId, metafield);
-          }
-
-          setNewMetaField({
-            key: "",
-            value: ""
-          });
-        }}
```
aldeed: Did you intend to remove all this? Seems like `onMetaSave` is still called by the metadata component.

imports/plugins/core/components/lib/blocks.js
mikemurray: I believe I tried without this load components function but it resulted in the blocks being rebuilt, with their HOCs every time a change was saved to a field. That meant you couldn't alt-tab to the next field as the component would re-render and lose focus. This change + using JSX in the `Blocks` component seems to solve that issue.

imports/plugins/included/product-admin/client/hocs/withProduct.js
mikemurray: This was moved to the `withProductForm` HOC

imports/plugins/core/components/lib/blocks.js
mikemurray: removed.

imports/plugins/core/components/lib/blocks.js
aldeed: OK it's something we could look into and adjust later anyway.

imports/plugins/included/product-admin/client/hocs/withProduct.js
aldeed: 👍 I missed that but I see it now.

imports/plugins/included/product-admin/client/blocks/VariantTaxForm.js
mikemurray: I'd rather leave it for now so this change can be isolated in its own PR.

mikemurray: Added a ticket here: https://github.com/reactioncommerce/reaction/issues/5103

PR #5065, opened 2019-03-25
Add cart referenceId in graphql schema

ajporlante:
Impact: **minor**  
Type: **chore**

## Issue
Cart's referenceId, introduced here https://github.com/reactioncommerce/reaction/pull/5054, needs to be passed to the client.

## Solution
Add `referenceId` to cart's graphql schema.

## Breaking changes
None

### DIFF
--- a/imports/plugins/core/cart/server/no-meteor/schemas/cart.graphql
+++ b/imports/plugins/core/cart/server/no-meteor/schemas/cart.graphql
@@ -18,6 +18,9 @@ type Cart implements Node {
   "An email address that has been associated with the cart"
   email: String
 
+  "A string to reference the cart"
+  referenceId: String
+
   "The items that have been added to the cart. A cart is not created until the first item is added. Items can be removed from a cart, and a cart is not deleted if all items are removed from it. Because all items may have been removed, this may be an empty array."
   items(after: ConnectionCursor, before: ConnectionCursor, first: ConnectionLimitInt, last: ConnectionLimitInt, sortOrder: SortOrder = desc, sortBy: CartItemsSortByField = addedAt): CartItemConnection
 
--- a/imports/plugins/core/cart/server/no-meteor/schemas/cart.graphql
+++ b/imports/plugins/core/cart/server/no-meteor/schemas/cart.graphql
@@ -21,10 +21,12 @@ type Cart implements Node {
   "The items that have been added to the cart. A cart is not created until the first item is added. Items can be removed from a cart, and a cart is not deleted if all items are removed from it. Because all items may have been removed, this may be an empty array."
   items(after: ConnectionCursor, before: ConnectionCursor, first: ConnectionLimitInt, last: ConnectionLimitInt, sortOrder: SortOrder = desc, sortBy: CartItemsSortByField = addedAt): CartItemConnection
 
-  "If you integrate with third-party systems that require you to send the same ID for order
+  """
+  If you integrate with third-party systems that require you to send the same ID for order
   calculations as for cart calculations, you may use this ID, which is the same on a `cart` as on
   the `order` placed from that cart. This ID can also be customized by plugins and is the best
-  ID to use if it is necessary to show a cart ID in the user interface."
+  ID to use if it is necessary to show a cart ID in the user interface.
+  """
   referenceId: String
 
   "The shop that owns the cart."
--- a/imports/plugins/core/cart/server/no-meteor/schemas/cart.graphql
+++ b/imports/plugins/core/cart/server/no-meteor/schemas/cart.graphql
@@ -18,11 +18,14 @@ type Cart implements Node {
   "An email address that has been associated with the cart"
   email: String
 
-  "A string to reference the cart"
-  referenceId: String
-
   "The items that have been added to the cart. A cart is not created until the first item is added. Items can be removed from a cart, and a cart is not deleted if all items are removed from it. Because all items may have been removed, this may be an empty array."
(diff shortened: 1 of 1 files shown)


imports/plugins/core/cart/server/no-meteor/schemas/cart.graphql
```diff
@@ -18,6 +18,9 @@ type Cart implements Node {
   "An email address that has been associated with the cart"
   email: String
 
+  "A string to reference the cart"
+  referenceId: String
+
```
ticean: Why is`referenceId`, "a string to reference the cart", required when there is already `_id`?

ajporlante: The intention for `referenceId` is to be an ID that can be customized. It can be a running sequence of integers in contrast to `_id` which is always a random string.

ticean: Is it indexed and used for lookup? Is it intended as an external reference and settable by the client? Is this widely applicable and needed in core; is a sequence of integers something that will apply to most Reaction users?

This seems like a customization for a specific need, but if it does need to go into core could you please add more detail to the description that states it's reason for existence and the intended use case(s)?

ajporlante: it is just used for lookup. It's settable only in the backend with a through a plugin. Whether or not it should be in core, I think the decision has been made here https://github.com/reactioncommerce/reaction/pull/5054. @zenweasel is probably the better person to explain. But as I understand, we needed a custom identifier for an order which can't be done without touching core code. Having the referenceId will have the most minimal impact to core for us to be able to implement such a requirement.

zenweasel:
@ticean The need here is to have a single identifier that follows the process from payment. Also we have added the ability to customize this so that we could for example use sequential digits. I find this to be an extremely common if not ubiquitous requirement

zenweasel:
Furthermore, `referenceId` already exists in core, this change is just exposing it to the client so clients that use client-only payment methods can use this number in their requests.

imports/plugins/core/cart/server/no-meteor/schemas/cart.graphql
zenweasel: @ticean The need here is to have a single identifier that follows the process from payment to order completion. Also we have added the ability to customize this so that we could for example use sequential digits. I find this to be an extremely common if not ubiquitous requirement

zenweasel: Furthermore, `referenceId` already exists in core, this change is just exposing it to the client so clients that use client-only payment methods can use this number in their requests.

aldeed: Correct that this is already approved for core. I had my doubts about the need, too, but it is a real need and I haven't thought of any other solution. That said, I'd like for the documentation here to clarify the use more. Instead of "a string to reference the cart", maybe:

```graphql
"""
If you integrate with third-party systems that require you to send the same ID for order
calculations as for cart calculations, you may use this ID, which is the same on a `cart` as on
the `order` placed from that cart. This ID can also be customized by plugins and is the best
ID to use if it is necessary to show a cart ID in the user interface.
"""
```

@ticean To answer some of your other questions, `referenceId` on both `cart` and `order` sometimes needs to be used for lookup due to limitations of third-party integrations. We have a unique index on both fields in Mongo. It seems like this will be a more common need than we hoped, so that's why it's in core.

aldeed: You also asked if it is settable by the client. It is not. It is currently a `Random.id()` by default, but a server-side plugin can also register a different function for generating any unique string. If `referenceId` is present on `cart` and if `placeOrder` receives a `cartId`, then the `referenceId` from the cart is copied to the order rather than generating a new one.

aldeed:
Please also move the field down so that the fields remain alphabetized

imports/plugins/core/cart/server/no-meteor/schemas/cart.graphql
ticean: Thanks @aldeed! This is the info I was looking for. To me, it seems very specific to put this into core. We've had a single request for this? I'll won't continue to press that, though.

My main concern is that the field is not well documented and a doc description like you posted above would greatly improve that. Thanks again!

aldeed: @ticean You don't necessarily need to drop the argument, but an offline conversation with @zenweasel will probably help illuminate the complexities. The one aspect that still bothers me on some levels is that `referenceId` is now serving two purposes: first, as a non-opaque'd ID by which customers and operators can reference the same order; second, as an ID that matches the assumptions made by external systems that need an ID passed in. But I also didn't like the idea of introducing a third type of ID.

ticean: Yes, sounds good. @zenweasel I will always welcome your enlightenment! :smile: The improved doc string will resolve my comments here.

aldeed:
@zenweasel I'm pretty sure multi-line description wrapped in `""` is not valid GraphQL, but I may be wrong. If it shows up fine in GraphQL Playground schema docs then it's fine. For multi-line we normally use triple double (?!) quotation marks above and below. https://graphql.github.io/graphql-spec/draft/#sec-String-Value

zenweasel:
@aldeed Made that change

PR #5068, opened 2019-03-26
(fix): GraphQL cart methods should return variantTitle of parent variant when options present

cshepherd:
Resolves issue with storefront not getting proper variantTitle (no issue number)
Impact: critical  
Type: bugfix

## Issue
When items are added to the cart as through createCartMutation or addCartItemsMutation, when options for a variant are present, the variantTitle returned is the title of the option, not the variant.

## Solution
Updated imports/plugins/core/cart/server/no-meteor/util/addCartItems.js to save parentVariant.title when it is present, or to retain chosenVariant.title otherwise, and pass this back as variantTitle in the reply.

## Testing
If RC is staged locally, add an item with variants and options to your cart, with a network inspector running.
Note that in the reply from graphql-alpha, the variantTitle in the reply is a number (ie '6' instead of '0007-DK-INDIGO-MOON') before this is applied, and is set properly (ie to '0007-DK-INDIGO-MOON') after it is applied.

### DIFF
--- a/imports/plugins/core/cart/server/no-meteor/util/addCartItems.js
+++ b/imports/plugins/core/cart/server/no-meteor/util/addCartItems.js
@@ -95,7 +95,9 @@ export default async function addCartItems(context, currentItems, inputItems, op
     // Until we do a more complete attributes revamp, we'll do our best to fudge attributes here.
     // The main issue is we do not have labels.
     const attributes = [];
+    let variantTitle = chosenVariant.title;
     if (parentVariant) {
+      variantTitle = parentVariant.title;
       attributes.push({
         label: null, // Set label to null for now. We expect to use it in the future.
         value: parentVariant.title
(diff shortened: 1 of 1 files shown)


aldeed:
@cshepherd @kieckhafer @spencern On a general level, this change doesn't seem correct to me. In cart and order, when we use the word `variant`, we mean it generically. The "variant" chosen for the cart may have been a variant in the catalog, or may have been an "option". In the future, we hope to merge "variant" and "option" ideas in the catalog, but for now it leads to confusion.

So in the operator UI, I can edit the "option", and enter something in Title field. I would expect that `variantTitle` is equal to this, and not equal to the parent variant's title. You didn't add the parent variant to the cart, you added the option variant.

So to me this seems like it is "fixing" what is actually a data issue. Product import should be setting a correct `title` in the catalog for all options.

A further confusion is the term `optionTitle`, which is not referring to a catalog option but rather to a select option. For example, for a set of options that represent color choices within a "Small" variant of "T-Shirt", `optionTitle` would be "Red" while `option.title`/`cartItem.variantTitle` would be "Small Red T-Shirt". By contrast, the `parentVariant.title` is probably just "Small T-Shirt" and the product title is "T-Shirt". So at each level `title` should become more specific in the data.

cshepherd:
Thank you, I agree and have a better insight now on how to resolve this.

PR #5070, opened 2019-03-27
feat: replace `reaction plugins` usage

### LINKED ISSUE #5036: Remove all usages of `reaction-cli` from Dockerfile and elsewhere
## Issue Description
We are deprecating the `reaction-cli` as of 2.0 official but it is still being used in the `Dockerfile` in the project and possibly elsewhere.

## Steps to Reproduce
View the `Dockerfile` included in the project

## Possible Solution
Use the `loadPlugins.mjs` instead of `reaction plugins load` (just a guess)

### Versions
RC10

aldeed:
Resolves #5036   
Impact: **minor**  
Type: **chore**

## Issue
Reaction CLI is deprecated. We previously updated development runs to use Node scripts copied into this repo instead. Dockerfile and CircleCI had not yet been updated.

## Solution
Dockerfile and CircleCI are now updated to have no remaining dependency on the Reaction CLI being installed. Actually removing the CLI from the base Docker image will be done later.

## Breaking changes
None

## Testing
Test a Docker build and verify that you can run the resulting image and use the API/operator UI as expected. The CircleCI config changes will be tested automatically based on tests passing on CI.


zenweasel:
@aldeed If this doesn't already I wonder if this could respect the settings in `pluginConfig` so that excluded plugins are not included in the build.

aldeed:
@zenweasel It runs the same `loadPlugins` function so it will respect excluded plugins.

PR #5074, opened 2019-03-27
Fix duped cart referenceId

### LINKED ISSUE #5073: Unable to reconcile cart when logging in
## Issue Description

When you create a cart anonymously, then you log in, `reconcileCarts` will error:
```
Message: E11000 duplicate key error collection: reaction.Cart index: referenceId_1 dup key: { : "44000000008" }
```
Here:
https://github.com/reactioncommerce/reaction/blob/969fc6efee092e99bf133084cd61e8c78fc8a6db/imports/plugins/core/cart/server/no-meteor/mutations/convertAnonymousCartToNewAccountCart.js#L1-L59

you will see that a new cart is being created with the same referenceId as the anonymous cart so the database returns an error on uniqueness of referenceId.

## Steps to Reproduce
1. Clear all documents in Cart collection.
2. Make sure you are logged out in starterkit.
3. Add an item to cart.
4. Log in.
5. See the console error that referenceId has been duplicated.

## Possible Solution
Not sure

ajporlante:
Resolves #5073 
Impact: **critical**  
Type: **bugfix**

## Issue

`convertAnonymousCartToNewAccountCart` tries to create a new cart with referenceId that is the same as an existing cart but referenceId needs to be unique as defined in the Cart schema.

## Solution
Before creating a new cart, update the old cart (the anonymous cart) by removing the referenceId field so the new cart can be created with the same referenceId without having uniqueness issues.

## Breaking changes
None

## Testing
- Clear all documents in Cart collection.
- Make sure you are logged out in starterkit.
- Add an item to cart.
- Take note of the _id and referenceId of the cart in the database.
- Log in.
- See that a new account cart is created in the database with a different _id but same referenceId.


imports/plugins/core/cart/server/no-meteor/mutations/convertAnonymousCartToNewAccountCart.js
```diff
@@ -53,7 +56,12 @@ export default async function convertAnonymousCartToNewAccountCart({
   });
 
   const { deletedCount } = await Cart.deleteOne(anonymousCartSelector);
-  if (deletedCount === 0) throw new ReactionError("server-error", "Unable to delete anonymous cart");
+  if (deletedCount === 0) {
+    if (referenceId) {
+      await Cart.findOneAndUpdate({ _id }, {set: { referenceId } });
```
aldeed: Since you are not even using the `findOne` result, these should both be `Cart.updateOne` instead of `Cart.findOneAndUpdate`.

imports/plugins/core/cart/server/no-meteor/mutations/convertAnonymousCartToNewAccountCart.js
```diff
@@ -39,7 +41,8 @@ export default async function convertAnonymousCartToNewAccountCart({
   };
 
   if (anonymousCart.referenceId) {
```
aldeed: Since you have it destructured now, this `if` could be changed to `if (referenceId)`

imports/plugins/core/cart/server/no-meteor/mutations/convertAnonymousCartToNewAccountCart.js
```diff
@@ -39,7 +41,8 @@ export default async function convertAnonymousCartToNewAccountCart({
...
-    newCart.referenceId = anonymousCart.referenceId;
+    await Cart.findOneAndUpdate({ _id }, {$unset: { referenceId: 1 } });
```
aldeed: Can you add a comment above this line about why this is being done? i.e., because referenceId has a uniqueness constraint but we want to copy the same value from anonymous cart to account cart.

ajporlante:
This is ready for another round of review.

ajporlante:
@aldeed Done.

PR #5079, opened 2019-03-28
feat: add name related data to viewer Account query

kieckhafer:
Impact: **minor**  
Type: **feature**

## Issue
The API does not provide all data needed for display on the account profile page.

## Solution
Add new `firstName`, `lastName`, and `name` fields to resolver.

## Breaking changes
None

## Testing
1. Create an account with `firstName`, `lastName`, and `name` fields inside the `profile`. There is no UI inside core reaction or starterkit to do this, you'll need to do it via Mongo.
1. use the `viewer` GQL query and see that these fields are returned when requested:

```
query {
  viewer {
    name
    firstName
    lastName
  }
}
```

### DIFF
--- a/imports/plugins/core/accounts/server/no-meteor/resolvers/Account/index.js
+++ b/imports/plugins/core/accounts/server/no-meteor/resolvers/Account/index.js
@@ -9,6 +9,9 @@ export default {
   addressBook,
   currency: (account) => getXformedCurrencyByCode(account.profile && account.profile.currency),
   emailRecords: (account) => account.emails,
+  firstName: (account) => account.profile.firstName,
+  lastName: (account) => account.profile.lastName,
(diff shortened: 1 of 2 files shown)


kieckhafer:
@aldeed @spencern Should we be providing the `firstName`, `lastName`, and `name` fields on the top level of the `viewer` query? Or should we put them inside a `profile` object, and then use `profile.name` (etc) when using them in the UI?

We currently have `name` at the top level of the query (even though it currently doesn't work):

```
query viewerQuery {
  viewer {
    name
    emailRecords {
      address
    }
  }
}
```

kieckhafer:
doesn't seem like the test issues are related, re-running to check.

kieckhafer:
@aldeed issues fixed.

PR #5084, opened 2019-04-01
adding indexes

Akarshit:
While publishing the product to the database, the following queries are made:
1. ```const result = await Catalog.updateOne({ "product.productId": catalogProduct.productId}, modifier, { upsert: true });```
2. `catalog = await Catalog.findOne({ "product.variants.options._id": optionId });`

This PR adds indexes for these queries.

### DIFF
--- a/imports/plugins/core/catalog/server/no-meteor/startup.js
+++ b/imports/plugins/core/catalog/server/no-meteor/startup.js
@@ -1,4 +1,5 @@
 import Logger from "@reactioncommerce/logger";
+import collectionIndex from "/imports/utils/collectionIndex";
 import hashProduct from "./mutations/hashProduct";
 
 /**
@@ -31,6 +32,9 @@ async function hashRelatedProduct(media, collections) {
  */
 export default function startup(context) {
   const { appEvents, collections } = context;
+
+  collectionIndex(collections.Catalog, { "product.productId": 1 }); // Used to update products
+  collectionIndex(collections.Catalog, { "product.variants.options._id": 1 }); // Used while publishing products
 
   appEvents.on("afterMediaInsert", ({ mediaRecord }) => {
     hashRelatedProduct(mediaRecord, collections).catch((error) => {
--- a/lib/collections/collections.js
+++ b/lib/collections/collections.js
@@ -210,6 +210,8 @@ export const MediaRecords = new Mongo.Collection("cfs.Media.filerecord");
 // Index on the props we search on
 if (Meteor.isServer) {
   collectionIndex(Catalog.rawCollection(), { updatedAt: 1 });
+  collectionIndex(Catalog.rawCollection(), { "product.productId": 1 }); // Used to update products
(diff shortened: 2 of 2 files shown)


imports/plugins/core/catalog/server/no-meteor/startup.js
```diff
@@ -32,6 +33,9 @@ async function hashRelatedProduct(media, collections) {
 export default function startup(context) {
   const { appEvents, collections } = context;
 
+  collectionIndex(collections.Catalog, { "product.productId": 1 }); // Used to update products
+  collectionIndex(collections.Catalog, { "product.variants.options._id": 1 }); // Used while publishing products
+
```
ticean: We had a team conversation in Slack where it was proposed to combine the id's and slugs from all the levels into a single field that could be indexed. We had a technical MongoDB review from the Atlas team that strongly recommended that we reduce the number of indexes. It was the highest priority recommendation from the review.

Could we discuss and determine a plan for creating the single index field before merging this?

cc @spencern @mikemurray

aldeed: I reviewed indexes yesterday while doing #5090. Although my PR does not add or remove any indexes yet, I did note many indexes that don't seem to be necessary in my view. These two do seem to be necessary. I think we should add these and then remove all of the blatantly useless ones and re-evaluate after that.

spencern: If adding this index solves a problem that we're having now without having to rewrite code (single index field), I'd like to get this merged and revisit optimizations to our queries tomorrow, from there we can determine and remove unnecessary indexes.

Ticean do you feel we should hold up this PR until we make a decision about indexing better? or can we create a follow up issue for someone to work this week?

spencern:
@Akarshit you need to sign your commits for this PR as well.

imports/plugins/core/catalog/server/no-meteor/startup.js
ticean: @spencern depends on what you're thinking when you say "a problem that we're having now". We have an implementation in progress but indexes can be added to that project's database manually, which is what the team is doing for now. I don't see why there's a rush to make a merge to Reaction core where we have the potential to churn indexes on the open source users.

The reason I asked that we hold is that the design of this particular merge is that these two indexes don't provide full coverage over the app queries. Scans from the Catalog collections were big part of why we had so many alerts from Atlas on the project.

zenweasel: So @ticean, @spencern @aldeed where are we with this? Are we closing this, are we doing anything to move this forward?

Akarshit:
@spencern Can we merge this?

ticean:
This is not a blocker in any way so I'm going to close this one for the reasons we've discussed above. 

@aldeed and I discussed alternative designs that would be more efficient on Slack and provide full index coverage. I'll open a new issue with that info.

zenweasel:
@ticean 
I think this is a blocker in the sense that this is one of the reasons that publish is so slow. If there is an alternative solution we need to get it underway ASAP

Issue #1948, opened 2017-03-08
Show app version in Dashboard


zenweasel:
Once the Avalara PR is merged we will have the Reaction version in the Shops collection. Seems like it would be handy to have this display somewhere in the Dashboard for both us and customers.

hlowso:
Hey there! I'm super new to github but I'd like to try to take on this issue. At the moment I'm just trying to get my bearings in the code and figure out where the dashboard is being generated and how it's getting its data. I'm happy to keep looking around myself but I'd also take any pointers. Cheers!

zenweasel:
There is a `Reaction.getAppVersion` method available and you could just add this as a read-only property in Shop settings.

aguscha333:
Hey has anyone taken this? I'd be happy to.

zenweasel:
@aguscha333 Sure. That would be great. Let me know if you need any help.

aguscha333:
I was able to install the repo but not get it up and running apparently. I've been a long time at it already, don't think it's still worth It to keep trying for such a small contribution

yellowbrickc:
Hi @all 
I see that you have a lot of "help-wanted" issues and I have some time and would like to help. On the other hand I have a few years of ecommerce experience so that I really would like to see this app up-and-running.
I installed and executed everything written in the documentation for react-platform but I'm failing to start the platform due the message "You must set MONGO_URL environment variable.". Even if try to set this (i.e. localhost:27017,  mongodb://mongo:27017/reaction and other combinations) I get the message "ERROR: MongoDB not reachable in time.". Can you give me a hint?

focusaurus:
@yellowbrickc Let's make sure your mongo started. `docker-compose -f reaction-platform/reaction/docker-compose.yml logs mongo` might have some info on whether it's up or down or has an error, etc. It'll be easier for us to assist with your setup in [our gitter chat](https://gitter.im/reactioncommerce/reaction) vs cluttering up this github issue.

chrispotter:
I started looking into this issue. 

My thought was to add it to the bottom of the sidebar as a universal reference from the admin, kind of how Magento has in their footer... Maybe something like `Reaction Commerce© v 2.1.3.4.5rev12`

<img width="393" alt="Screen Shot 2019-03-28 at 2 27 40 PM" src="https://user-images.githubusercontent.com/7069752/55183284-dfcfe000-5165-11e9-9119-7b9aace12a07.png">


[...]

PR #5087, opened 2019-04-01
Feature/system info

chrispotter:
Initial Solution to #1948 
Impact: **major**  
Type: **feature**

## Issue
This is the start to a `system-info` plugin that will offer information related to system 

## Solution
This is step one from Issue #1948 to add a new graphql schema to be referenced from the dashboard to show appVersion.

## Testing
`http://localhost:3000/graphql-alpha`
```
{
  systemInformation(shopId: "cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg==") {
    apiVersion
    mongoVersion {
      version
    }
    plugins {
      name
      version
    }
  }
}
```

Use admin auth token and your correct primary shop ID.

### DIFF
--- a/imports/plugins/core/system-info/register.js
+++ b/imports/plugins/core/system-info/register.js
@@ -8,7 +8,6 @@ Reaction.registerPackage({
   name: "reaction-systeminfo",
   icon: "fa fa-info",
   autoEnable: true,
-  version: "1.0.0",
   graphQL: {
     resolvers,
     schemas
--- a/imports/plugins/core/system-info/register.js
+++ b/imports/plugins/core/system-info/register.js
@@ -1,0 +1,17 @@
+import Reaction from "/imports/plugins/core/core/server/Reaction";
+import queries from "./server/no-meteor/queries";
+import resolvers from "./server/no-meteor/resolvers";
+import schemas from "./server/no-meteor/schemas";
+
+Reaction.registerPackage({
+  label: "System Information",
+  name: "reaction-systeminfo",
+  icon: "fa fa-info",
+  autoEnable: true,
+  version: "1.0.0",
+  graphQL: {
+    resolvers,
+    schemas
+  },
+  queries
+});
--- a/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
+++ b/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
@@ -5,10 +5,13 @@ import packageJson from "/package.json";
  * @method
  * @memberof SystemInformation/GraphQL
  * @summary get systemInformations
+ * @param {Object} context - an object containing the per-request state
  * @return {<Object} System Information
  **/
-export default async function systemInformation() {
+export default async function systemInformation(context) {
+
   return {
-    apiVersion: packageJson.version
+    apiVersion: packageJson.version,
+    plugins: await context.collections.Packages.find({ version: { $exists: true }}).toArray()
   };
 }
--- a/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
+++ b/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
@@ -1,0 +1,21 @@
+import ReactionError from "@reactioncommerce/reaction-error";
+import packageJson from "/package.json";
+
+/**
+ * @name queries.systemInformation
+ * @method
+ * @memberof Tags/GraphQL
+ * @summary get systemInformations
+ * @param {Object} context - an object containing the per-request state
+ * @param {String} [params.shopId] - Shop ID
+ * @return {Promise<Array<Object>>} array of TagProduct
+ **/
+export default async function systemInformation(context) {
+  // if (!context.userHasPermission(["owner", "admin"], shopId)) {
+  //   throw new ReactionError("access-denied", "Access denied");
+  // }
+
+  return {
+    apiVersion: packageJson['version']
+  };
+}
--- a/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
+++ b/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
@@ -1,16 +1,21 @@
 import packageJson from "/package.json";
+import ReactionError from "@reactioncommerce/reaction-error";
+
 
 /**
  * @name queries.systemInformation
  * @method
  * @memberof SystemInformation/GraphQL
  * @summary get systemInformations
+ * @param {String} args.shopId The ID of the shop to load settings for
  * @param {Object} context - an object containing the per-request state
  * @return {<Object} System Information
  **/
-export default async function systemInformation(context) {
+export default async function systemInformation(id, context) {
+  if (!context.userHasPermission(["owner", "admin"], id)) throw new ReactionError("access-denied", "User does not have permission");
+  const plugins = await context.collections.Packages.find({ shopId: id, version: { $exists: true } }).toArray();
   return {
     apiVersion: packageJson.version,
-    plugins: await context.collections.Packages.find({ version: { $exists: true } }).toArray()
+    plugins: plugins.map(({ name, version }) => ({ name, version }))
   };
 }
--- a/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
+++ b/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
@@ -11,13 +11,13 @@ import ReactionError from "@reactioncommerce/reaction-error";
  * @return {Promise<Object>} System Information
  **/
 export default async function systemInformation(context, shopId) {
-  const { userHasPermission, collections: { Packages }, app: {db} } = context;
+  const { userHasPermission, collections: { Packages }, app: { db } } = context;
   // sensitive information should be accesible to admins only
-  if (!context.userHasPermission(["admin"], shopId)) throw new ReactionError("access-denied", "User does not have permission");
+  if (!userHasPermission(["admin"], shopId)) throw new ReactionError("access-denied", "User does not have permission");
 
   const mongoAdmin = await db.admin();
   const mongoInfo = await mongoAdmin.serverStatus();
-  const plugins = await Packages.find({ shopId: shopId, version: { $exists: true }}, {projection: { name: 1, version: 1 } }).toArray();
+  const plugins = await Packages.find({ shopId: { shopId }, version: { $ne: null } }, { projection: { name: 1, version: 1 } }).toArray();
   return {
     apiVersion: packageJson.version,
     mongoVersion: { version: mongoInfo.version },
--- a/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
+++ b/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
@@ -1,4 +1,3 @@
-import ReactionError from "@reactioncommerce/reaction-error";
 import packageJson from "/package.json";
 
 /**
@@ -6,11 +5,10 @@ import packageJson from "/package.json";
  * @method
  * @memberof SystemInformation/GraphQL
  * @summary get systemInformations
- * @param {Object} context - an object containing the per-request state
  * @return {<Object} System Information
  **/
-export default async function systemInformation(context) {
+export default async function systemInformation() {
   return {
-    apiVersion: packageJson['version']
+    apiVersion: packageJson.version
   };
 }
--- a/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
+++ b/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
@@ -1,4 +1,3 @@
-import packageJson from "/package.json";
 import ReactionError from "@reactioncommerce/reaction-error";
 
 /**
@@ -11,15 +10,15 @@ import ReactionError from "@reactioncommerce/reaction-error";
  * @return {Promise<Object>} System Information
  **/
 export default async function systemInformation(context, shopId) {
-  const { userHasPermission, collections: { Packages }, app: { db } } = context;
+  const { userHasPermission, app: { db } } = context;
   // sensitive information should be accesible to admins only
   if (!userHasPermission(["admin"], shopId)) throw new ReactionError("access-denied", "User does not have permission");
 
   const mongoAdmin = await db.admin();
   const mongoInfo = await mongoAdmin.serverStatus();
-  const plugins = await Packages.find({ shopId: { shopId }, version: { $ne: null } }, { projection: { name: 1, version: 1 } }).toArray();
+  const plugins = await Object.values(context.app.registeredPlugins).filter((plugin) => plugin.version);
   return {
-    apiVersion: packageJson.version,
+    apiVersion: context.appVersion,
     mongoVersion: { version: mongoInfo.version },
     plugins: plugins.map(({ name, version }) => ({ name, version }))
   };
--- a/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
+++ b/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
@@ -9,9 +9,8 @@ import packageJson from "/package.json";
  * @return {<Object} System Information
  **/
 export default async function systemInformation(context) {
-
   return {
     apiVersion: packageJson.version,
-    plugins: await context.collections.Packages.find({ version: { $exists: true }}).toArray()
+    plugins: await context.collections.Packages.find({ version: { exists: true } }).toArray()
   };
 }
--- a/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
+++ b/imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
@@ -4,17 +4,12 @@ import packageJson from "/package.json";
 /**
  * @name queries.systemInformation
  * @method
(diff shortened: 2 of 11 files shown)


imports/plugins/core/system-info/server/no-meteor/schemas/schema.graphql
```diff
@@ -0,0 +1,9 @@
+extend type Query {
+  systemInformation: SystemInfo
+}
+
+"Represents Reaction System Infomation"
+type SystemInfo {
+  # core api version
+  apiVersion: String
+}
```
focusaurus: I think it would be nice to add some comments here about other fields we plan to perhaps add to this type in the future as per Eric's sketch, just so someone else implementing them will know for sure they are going in the right spot.

imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
```diff
@@ -0,0 +1,16 @@
...
+ * @method
+ * @memberof SystemInformation/GraphQL
+ * @summary get systemInformations
+ * @param {Object} context - an object containing the per-request state
+ * @return {<Object} System Information
+ **/
+export default async function systemInformation(context) {
+  return {
+    apiVersion: packageJson.version,
+    plugins: await context.collections.Packages.find({ version: { exists: true } }).toArray()
```
chrispotter: @aldeed is this what you meant by await with the `toArray()`? testing through graphql-alpha still returns plugins{}

imports/plugins/core/system-info/register.js
```diff
@@ -0,0 +1,17 @@
...
+import queries from "./server/no-meteor/queries";
+import resolvers from "./server/no-meteor/resolvers";
+import schemas from "./server/no-meteor/schemas";
+
+Reaction.registerPackage({
+  label: "System Information",
+  name: "reaction-systeminfo",
+  icon: "fa fa-info",
+  autoEnable: true,
+  version: "1.0.0",
```
aldeed: So far we have not been versioning core plugins separately. `version` field is optional and can be removed.

imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
```diff
@@ -0,0 +1,16 @@
...
+ * @method
+ * @memberof SystemInformation/GraphQL
+ * @summary get systemInformations
+ * @param {Object} context - an object containing the per-request state
+ * @return {<Object} System Information
+ **/
+export default async function systemInformation(context) {
+  return {
+    apiVersion: packageJson.version,
+    plugins: await context.collections.Packages.find({ version: { $exists: true } }).toArray()
```
aldeed: Hmm, just thinking this through a bit more. There are `settings` objects on many package docs that may have sensitive settings. Although this won't make it through GraphQL back to clients, I'd rather be safe and filter out unneeded fields here. So add a `.map` on the array and keep only `name` and `version` properties for now.

Another thing I forgot is that there is one Package doc per shop ID in a multi-shop scenario. So the GraphQL will need to take a `shopId` argument and pass it into this function, and then this function should include it in the `find` query.

Finally, `version` could be `null`, which passes `{ $exists: true }` check, so `{ $ne: null }` is better (ref: https://docs.mongodb.com/manual/tutorial/query-for-null-fields/).

So putting all that together, something like this:

```js
export default async function systemInformation(context, { shopId }) {
  const plugins = await context.collections.Packages.find({ shopId, version: { $ne: null } }).toArray();
  return {
    apiVersion: packageJson.version,
    plugins: plugins.map(({ name, version }) => ({ name, version }))
  };
}
```

imports/plugins/core/system-info/server/no-meteor/resolvers/Query/systemInformation.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @name Query.systemInformation
+ * @method
+ * @memberof SystemInformation/GraphQL
+ * @summary get system information for reaction site
+ * @param {ConnectionArgs} args An object of all arguments that were sent by the client
+ * @param {String} args.shopId The ID of the shop to load navigation items for
```
aldeed: This doc comment needs updating.

imports/plugins/core/system-info/server/no-meteor/schemas/schema.graphql
```diff
@@ -0,0 +1,16 @@
+extend type Query {
+  systemInformation: SystemInfo
+}
+
+"Represents Reaction Plugin"
+type Plugin {
+  name: String,
```
aldeed: Every field should have a string above it with the field description. These are shown in GraphQL Playground when viewing the schema docs.

Also, I think the commas are ignored, but they aren't necessary and we don't include them.

imports/plugins/core/system-info/server/no-meteor/schemas/schema.graphql
```diff
@@ -0,0 +1,16 @@
+extend type Query {
+  systemInformation: SystemInfo
```
aldeed: Add a string literal description of this query above it. Also add `!` after `SystemInfo` because there will always be a non-null return if there isn't an error thrown.

And as I mentioned in another comment, you'll need to add `(shopId: ID!)` param for this query.

imports/plugins/core/system-info/server/no-meteor/schemas/schema.graphql
```diff
@@ -0,0 +1,16 @@
...
+}
+
+"Represents Reaction Plugin"
+type Plugin {
+  name: String,
+  version: String
+}
+
+"Represents Reaction System Infomation"
+type SystemInfo {
```
aldeed: Might as well call this `SystemInformation`, too. Best to not abbreviate if we don't have to.

mikemurray:
Maybe you're already thinking about this, but is this API going to remain public, or will it be behind auth of some kind? Since this is system info, it feels like it should be secured somehow.

chrispotter:
@mikemurray I was considering throwing in
`if (!context.userHasPermission(["owner", "admin"], id)) throw new ReactionError("access-denied", "User does not have permission");` but while I was testing my graphql-fu wasn't high enough to pass in an acceptable user through graphql-alpha and I forgot to add back in after.

@aldeed do you agree this should be behind authentication?

imports/plugins/core/system-info/register.js
chrispotter: oops added for testing and never removed it

chrispotter:
I think there is enough here for a PR with final review. 

This can then be used when refactoring the admin page issue in #1948 when that is reconfigured to use updated API techniques. what do you think @focusaurus @aldeed ?

imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
```diff
@@ -0,0 +1,21 @@
...
+ * @method
+ * @memberof SystemInformation/GraphQL
+ * @summary get systemInformations
+ * @param {String} args.shopId The ID of the shop to load settings for
+ * @param {Object} context - an object containing the per-request state
+ * @return {<Object} System Information
+ **/
+export default async function systemInformation(id, context) {
+  if (!context.userHasPermission(["owner", "admin"], id)) throw new ReactionError("access-denied", "User does not have permission");
+  const plugins = await context.collections.Packages.find({ shopId: id, version: { $exists: true } }).toArray();
```
aldeed: I still think you should replace `{ $exists: true}` with `{ $ne: null }` to avoid any null version packages being included.

Second, stylistically, you should change `id` argument name to `shopId` and change `shopId: id` to just `shopId` in the query.

Also, stylistically, we destructure whenever using keys from an object more than one place in a function, i.e. put this at the top of the function:

```js
const { collections, userHasPermission } = context;
```

In the jsdocs, `args.shopId` should be just `shopId`.

Whenever we pass in `context`, we do it as the first argument, so flip your args. There may be some older functions that don't follow this pattern but we try to always do it now.

Lastly, I don't necessarily agree with restricting this to admins. Currently it is only versions, and this information could be useful to a storefront app, or to anyone who is able to do any queries (which is anyone). If we later add fields that only admins should see, we could implement field-level security at that time.

imports/plugins/core/system-info/server/no-meteor/resolvers/Query/systemInformation.js
```diff
@@ -0,0 +1,15 @@
+import { decodeShopOpaqueId } from "@reactioncommerce/reaction-graphql-xforms/shop";
+
+/**
+ * @name Query.systemInformation
+ * @method
+ * @memberof SystemInformation/GraphQL
+ * @summary get system information for reaction site
+ * @param {ConnectionArgs} args An object of all arguments that were sent by the client
+ * @param {Object} context An object containing the per-request state
+ * @return {<Object>} System Information Object
```
aldeed: In this function and the other query function, the `@return` type is technically `Promise<Object>` since they are `async`.

imports/plugins/core/system-info/server/no-meteor/schemas/schema.graphql
```diff
@@ -0,0 +1,21 @@
+extend type Query {
+  systemInformation(shopId: ID!): SystemInformation!
+}
+
+"Represents Reaction Plugin"
+type Plugin {
+  # name of plugin
+  name: String
+  # version of plugin
+  version: String
```
aldeed: `#` is a comment and we only use those sparingly for information that is useful in the file but need not be in the public docs. String literals are descriptions, which are used to build API docs, and we require those for everything. The descriptions should also start with a capital letter and be in sentence case. In this case:

```graphql
"Information about an API plugin"
type Plugin {
  "The plugin name"
  name: String!

  "The installed version of this plugin"
  version: String
}

"Information about the Reaction system"
type SystemInformation {
  "The core API version"
  apiVersion: String!

  # TODO: data version
  # TODO: mongo version

  "The installed API plugins"
  plugins: [Plugin]!
}
```

Note I also added `!` after the `String` type for `name` and `apiVersion`. This is because all plugins always have a name, and we want to inform API users that they can rely on that value being there without testing for `null`. I added one after `[Plugin]`, too, to indicate that we'll always return an array rather than `null`, even though it may be an empty array.

We also keep an empty line between each field as a matter of style.

You can add database version right now by doing this in the query function:

```js
const mongoInfo = await context.app.db.admin.serverStatus();
return {
  // ...
  mongoVersion: mongoInfo.version
};
```

There may be other useful info in `mongoInfo`, in which case you could make a separate `DatabaseInformation` type for it in the schema.

imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
chrispotter: 🤦‍♂️ darn i thought i had updated `{ $ne: null }` ! thanks!

I'll remove admin stuff!

focusaurus: Could we use a mongo field selection/project to just ask for `name` and `version` from mongodb? I think in general it's good to be precise about that when you know exactly which fields you need, but in this case specifically I think sometimes packages have a black-box `settings` object and in some cases they might be large-ish.

imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
```diff
@@ -0,0 +1,21 @@
...
+  return {
+    apiVersion: packageJson.version,
+    plugins: plugins.map(({ name, version }) => ({ name, version }))
```
focusaurus: Noice!

imports/plugins/core/system-info/server/no-meteor/schemas/schema.graphql
```diff
@@ -0,0 +1,21 @@
...
+
+"Represents Reaction System Infomation"
+type SystemInformation {
+  # core api version
+  apiVersion: String
+  # TODO: data version
+  # TODO: mongo version
+  #plugins installed with name, version information
+  plugins:[Plugin]
+}
```
focusaurus: @aldeed thanks for all the attention to detail in here. This part of the code is still mostly unfamiliar to me so my input is pretty cursory. Since this is going into the core API, I think it'd be worth having another set of eyes too. Maybe @mikemurray ?

imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
focusaurus: On the topic of whether to make this public or restricted, my personal vote is public. My reasoning is it has a high level of utility for troubleshooting, tech support, and general deployment management instrumentation/introspection. 

However, this type of thing is usually disabled in the first round of security hardening. For example, the npm `helmet` security package removes the equivalent header from `express` app servers, and it's common to strip this stuff from Internet-facing web servers. It's a countermeasure to defeat fingerprinting and I think from a security perspective it's a valid practice. Would it make sense to add a configuration setting for this like `[public, admin, disabled]`?

imports/plugins/core/system-info/server/no-meteor/schemas/schema.graphql
chrispotter: i like the Idead of a databaseInformation Object as a child of the SystemInformation, i'll add that too

ticean:
This should definitely be protected. We **should not** expose information like installed plugins to the public API. That would expose business secrets (ie which payment provider is used) as well as provide valuable information to attackers.

imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
aldeed: Agree with @focusaurus that you should add `{ projection: { name: 1, version: 1 } }` as second arg of the collection `find` call.

aldeed:
@ticean In most cases it's going to be pretty obvious what payment provider is used. Frontend code will have `stripe` library, or you'll be redirected to paypal payment page, etc. But I'm fine with hiding plugin information from public API to be safe. What do you think about `apiVersion` field? Can we keep that public? (`databaseInformation` field should be admin-only.)

ticean:
I just used payment provider as an example. It doesn't matter if the users see which payment is used on the frontend. That doesn't expose which _plugin_ and _plugin version_ are providing the functionality. A plugin list does. Such a list provides mineable data that can be analyzed to describe all the functionalities the site has, the vulnerabilities available (via cross reference), and generally how up to date the site is keeping their dependencies.

I don't think any part of this endpoint should be public.

chrispotter:
I think [this](https://github.com/reactioncommerce/reaction/pull/5154) will effect this PR

chrispotter:
@aldeed i think this is ready for your review again!

imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
```diff
@@ -0,0 +1,26 @@
+import packageJson from "/package.json";
```
aldeed: I've started thinking about how we're soon wanting to move plugins to be npm packages. An import from outside the plugin like this one won't work in that case. At the moment, we only need the version. Instead, there are two files where `new ReactionNodeApp()` is done. You can import `packageJson` in those two files, and add the version to context in the `ReactionNodeApp` constructor options: `appVersion: packageJson.version`.

Then here you can use `context.appVersion`.

imports/plugins/core/system-info/server/no-meteor/queries/systemInformation.js
```diff
@@ -0,0 +1,26 @@
...
+ * @return {Promise<Object>} System Information
+ **/
+export default async function systemInformation(context, shopId) {
+  const { userHasPermission, collections: { Packages }, app: { db } } = context;
+  // sensitive information should be accesible to admins only
+  if (!userHasPermission(["admin"], shopId)) throw new ReactionError("access-denied", "User does not have permission");
+
+  const mongoAdmin = await db.admin();
+  const mongoInfo = await mongoAdmin.serverStatus();
+  const plugins = await Packages.find({ shopId: { shopId }, version: { $ne: null } }, { projection: { name: 1, version: 1 } }).toArray();
```
aldeed: Since initially reviewing this, we've been talking more about trying to stop using the `Packages` collection. The problem is that the collection can get out of sync with what plugins are truly in the codebase. So it would be better to access the actual list of plugins that were registered when the app started up.

On context, `context.app` is the `ReactionNodeApp` instance. On that you can find the plugin config for all plugins on `app.registeredPlugins` object. So I think it would be something like this:

```js
const plugins = Object.values(context.app.registeredPlugins);
```

The rest should be the same.

aldeed: Note that unlike the `Packages` collection, `registeredPlugins` is NOT different per shopId. But I think you should keep the shopId param in GraphQL for the permission check and potential future use.

imports/plugins/core/system-info/server/no-meteor/schemas/schema.graphql
```diff
@@ -0,0 +1,27 @@
+extend type Query {
+  "SystemInformation object"
+  systemInformation(shopId: ID!): SystemInformation!
+}
+"Represents Reaction Plugin"
+type Plugin {
+  "name of plugin"
```
aldeed: All field descriptions should begin with a capital letter by convention. I guess our GraphQL linter does not check this.

aldeed:
Merged in latest develop to fix the startup bug, and then tested. Looks good!

PR #5088, opened 2019-04-01
Plugin: Prometheus Metrics

focusaurus:
Resolves [reaction-feature-requests/60](https://github.com/reactioncommerce/reaction-feature-requests/issues/60)  
Impact: **minor**  
Type: **feature**

## Issue

Currently reaction is not instrumented for application performance monitoring.

## Solution

- Add a new included plugin which Incorporates the npm package `epimetheus`
- This will expose prometheus metrics about the node process
  - It's disabled by default
  - Can be enabled in the Catalyst UI
  - The HTTP path it uses defaults to `/metrics` but is configurable in the catalyst UI

## Breaking changes

Potential for http routing conflicts on `/metrics`.

## Testing

- Fire up reaction core with this branch
- Navigate to the catalyst (operator UI)
- Click on "Metrics Prometheus"
- Check "enabled"
- Click "Save Changes"
- Restart reaction
  - (Although technically we could probably make this changeable without a restart, it would be a lot more work)
- `curl localhost:3000/metrics` should spit out some plain text prometheus metrics
- Go back and disable this plugin
- restart reaction
- `curl localhost:3000/metrics` should give you a 404 status code and the "Not Found" HTML page


focusaurus:
Will handle snyk security CI failure by rebasing after #5089 is resolved.

focusaurus:
Erik, Click on "Shop Localization" then click the blue refresh arrows
there. That will rebuild the translations in your db from source code, and
then you should see the proper labels.

kieckhafer:
👍 Looks good with translations. Thanks for putting `Helmet` there too to get the title.

rymorgan:
@focusaurus Totally appreciate the 'default to action' on this PR but anything going into core needs some design or design consultation. The UX on this is confusing. We're really aiming to contain the left nav. Not all functions and settings are created equally. Furthermore, the form elements are not following proper UI patterns. 

I realize that proper UI patterns are being communicated clearly at this stage and is something we're going to set a foundation for during the product cycle. For now anything that's going into the core UI needs some design considerations.

focusaurus:
Roger roger. Let me know when you have time to take a pass on the design and I can implement that.

focusaurus:
I think I might roll this back and implement as a separate repo in the interim.

PR #5090, opened 2019-04-02
Remove aldeed:schema-index Meteor package, use collectionIndex

### LINKED ISSUE #5080: Remove aldeed:schema-index Meteor package
## Context
This is part of the work necessary to split the Reaction GraphQL API fully off Meteor. See #4996 

## Work
- [x] Move all SimpleSchema indexes to use `collectionIndex` function. Remove `index` and `unique` options from all SimpleSchemas.
- [x] Remove `aldeed:schema-index` Meteor package.

Note: Be sure to change the index names to match how `aldeed:schema-index` names them, for backward compatibility with systems that already have the indexes created.

aldeed:
Resolves #5080   
Impact: **minor**  
Type: **feature**

## Changes
The `aldeed:schema-index` Meteor package takes SimpleSchema options like `index: 1` and `unique: true` and calls MongoDB `createIndex` based on them. In an effort to reduce Meteor dependencies, all usage of this and the `aldeed:schema-index` dependency itself has now been removed.

Indexes are now set up by the plugins that care about them. They call `collectionIndex` in a startup function to do this. This may not be the final way this is done, but this PR is one step out of many to get to where we are going.

The goal was not to make any decisions about the appropriateness of any indexes in this PR. I did notice some that seem to be outdated, but I did not remove them at this time. This is strictly an internal refactor.

**There is one way in which this PR is more than a refactor. There were several fields in the Catalog schema that were marked to be indexed, but because that schema was not attached to the Catalog collection in Meteor, the indexes were not actually created. I did copy the useful ones to the new code, so now Catalog collection will have additional indexes created.**

## Breaking changes
If you have custom plugins that use `index` or `unique` options in their schemas, add back the `aldeed:schema-index` dependency or update the plugins to use `collectionIndex` function.

## Testing
1. Run the app on `develop` branch. Verify that your local database has indexes. Save off the list of indexes. For example:

```
db.getCollectionNames().forEach(function(collection) {
   indexes = db[collection].getIndexes();
   print("Indexes for " + collection + ":");
   printjson(indexes);
});
```

2. Drop all collections.
3. Run the app on this PR branch. Verify that your local database has indexes. Compare the list with the one you saved off. It should have all of the same indexes as before, and perhaps a few additional. (As stated above, some indexes may not make sense anymore, but we will evaluate this in a future PR.)


imports/collections/schemas/tags.js
```diff
@@ -16,7 +16,7 @@ import { Metafield } from "./metafield";
  * @property {String[]} relatedTagIds optional
  * @property {Boolean} isDeleted default value: `false`
  * @property {Boolean} isTopLevel required
- * @property {Boolean} isVisible defalut value: `true`
```
kieckhafer: nice catch

imports/node-app/core/util/defineCollections.js
```diff
@@ -29,7 +29,6 @@ export default function defineCollections(db, collections) {
     ProductSearch: db.collection("ProductSearch"),
     roles: db.collection("roles"),
     SellerShops: db.collection("SellerShops"),
-    Shipping: db.collection("Shipping"),
```
kieckhafer: Assuming this is just because we are creating `Shipping` inside it's own package?

imports/plugins/core/discounts/lib/collections/schemas/discounts.js
```diff
@@ -91,7 +82,7 @@ export const Discounts = new SimpleSchema({
   },
   "conditions.order.min": {
     type: Number,
-    label: "Mininum",
+    label: "Minimum",
```
kieckhafer: 👍

imports/collections/schemas/tags.js
aldeed: Code Spell Checker extension caught it, but I'll take credit. 😄

imports/node-app/core/util/defineCollections.js
aldeed: Yep

Issue #4692, opened 2018-09-28
Document how we round and total


aldeed:
This issue is a spinoff of https://github.com/reactioncommerce/reaction/issues/4684#issuecomment-425473131

The outcome should be public documentation about how we round and total an order, and why. Development changes may be necessary after this discussion, in order for the system to match whatever documentation we write, so creating issues for those necessary changes is another outcome of this issue.

Considerations:
- At what points should rounding occur and how (floor/ceil/nearest)? E.g., does each item's tax get rounded to nearest and then totaled to get total tax? Or does tax get totaled and then rounded? These will yield slightly different amounts for total tax. Same question for discounts.
- What does "nearest cent" mean? There might be differences in which decimal place we should round to per currency? And some currencies do not have cents, but these could end up being fractional if cut in half by a discount or multiplied by a non-integer sales tax percentage.
- There may be legal requirements for how to apply and round taxes (multiply by each item subtotal vs multiply by order total)
- There may be business requirements for how to apply and round discounts
- We use `account.js` to format money values for display. It's readme lists "Add choice of rounding method for precision (up, down or nearest-neighbour)" as a future feature, but does not state the current rounding method. It appears to be rounding to nearest. It also has open issues about rounding errors (https://github.com/openexchangerates/accounting.js/issues/195). We may want to use something else if this library is not properly maintained.

cc @spencern @ticean I will add to architecture meeting agenda.

aldeed:
Related: https://docs.mongodb.com/manual/tutorial/model-monetary-data/

It may be best to develop a plan to convert all floats to `NumberDecimal` in MongoDB

zdanowiczkonrad:
Regarding problem with accounting.js that is unmaintained, consider using my lightweight alternative - [https://github.com/zdanowiczkonrad/number-currency-format](number-currency-format) - I developed it for my purpose initially, but seems like it gives more control about precision and rounding and there should be definitely a way to solve your issue (otherwise, I am open to extend it for your needs!).

PR #5091, opened 2019-04-02
fix:  Rounding Issue in Cart/Orders

nnnnat:
Impact: **critica**  
Type: **bugfix**

## Issue
When using Reaction with 3rd party pricing engines Reaction will sometimes incorrectly round the carts total/subtotal. For example adding 6 product with a price of $2.99 would result in a cart/order total of $17.93999999999994.

## Solution
In this PR I've wrapped any cart/order "money math" with [accounting-js](https://openexchangerates.github.io/accounting.js/) `toFixed` function to force the correct rounding. 

## Breaking changes
N/A

## Testing
1. Do the [order creation](https://docs.reactioncommerce.com/docs/order-creation) script to test these changes. Make sure you add multiples of the same product to your cart. We want to verify the `price * quantity` calculations are also working.

### DIFF
--- a/imports/plugins/core/cart/server/no-meteor/util/addCartItems.js
+++ b/imports/plugins/core/cart/server/no-meteor/util/addCartItems.js
@@ -1,5 +1,6 @@
 import Random from "@reactioncommerce/random";
 import SimpleSchema from "simpl-schema";
+import { toFixed } from "accounting-js";
 import ReactionError from "@reactioncommerce/reaction-error";
 import findProductAndVariant from "/imports/plugins/core/catalog/server/no-meteor/utils/findProductAndVariant";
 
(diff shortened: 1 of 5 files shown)


imports/plugins/core/cart/server/no-meteor/util/addCartItems.js
```diff
@@ -133,7 +134,7 @@ export default async function addCartItems(context, currentItems, inputItems, op
       shopId: catalogProduct.shopId,
       // Subtotal will be kept updated by event handler watching for catalog changes.
       subtotal: {
-        amount: variantPriceInfo.price * quantity,
+        amount: +toFixed(variantPriceInfo.price * quantity, 3),
```
spencern: Am I reading this right that we're fixing to 3 decimal places? If so, why 3?

kieckhafer: Based of @aldeed's previous work on related issues, some currency codes go up to 3 decimals, so we need to accommodate those.

Algeria, Bahrain, Iraq, Jordan, Kuwait, Libya and Tunisia

aldeed:
I want to also point out that while this should mitigate 99.999% of cases, it doesn't really FIX the issue of JS math errors. I've previously documented some of the gotchas here: https://github.com/reactioncommerce/reaction/issues/4692

@spencern It may be a good time to focus some effort on figuring out a 100% foolproof solution to storing and calculating with floats before this crops up again.

PR #5102, opened 2019-04-05
#5097: 📬 Order confirmation email can be overriden by a Plugin

### LINKED ISSUE #5097: Emails: Order confirmation email should be overrideable.

machikoyasuda:
Resolves #5097  
Impact: **critical**  
Type: **feature**

## Issue
Plugins can trigger emails, like the Order Confirmation email, to be sent after a certain `appEvent`, but they currently cannot cancel the API's default email from being sent. As a result, in the current app, if a custom plugin sends a custom Order Confirmation email, the customer will get 2 emails: Reaction's default, and the custom one. A plugin developer should be able to _override_ the default Order Confirmation email method, rather than add its own.

## Solution
- Refactor the current `core/orders/server/util/sendOrderEmail.js` method to use a new `core/orders/server/no-meteor` mutation method instead.
- Plugins can now override this `sendOrderEmail` mutation.

## Breaking changes
None

## Testing
0. Make sure your stock StarterKit supports email (SMTP) and checkout (payment). (Ask a coworker for SMTP config instructions).
1. In stock Reaction and stock Starterkit, make sure Reaction Order confirmation emails still work
2. Make sure a plugin can override this mail function successfully (to be tested later more fully in the plugin itself.)


aldeed:
@machikoyasuda To resolve the failing integration tests, you can create a file in `/imports/test-utils/__mocks__/meteor`, named `meteorhacks:ssr.js`, and put this in it:

```js
export const SSR = {
  compileTemplate: () => null,
  render: () => "TEST"
};
```

I didn't test it, so that might not be 100% correct, but something similar to that should work. The point is to mock the `meteorhacks:ssr` Meteor package since it's imported in non-Meteor code. I have a branch de-meteorizing that stuff, but I still need to clean it up and get it reviewed and merged.

aldeed:
Per discussion with @machikoyasuda I am going to merge this into my de-meteorize emailing PR and get the whole thing merged so that this can be done properly.

PR #5106, opened 2019-04-08
feat: drop unnecessary MongoDB indexes

aldeed:
Impact: **major**  
Type: **performance**

## Issue
Due to past usage of the schema-index Meteor package and because people have not always remembered to drop related indexes when removing fields from the schemas, there were many unnecessary MongoDB indexes being created in code.

## Solution
Remove the `createIndex` calls and add a migration to drop the indexes. If there is confusion about why any of the indexes are being dropped, leave a comment and I can explain.

## Breaking changes
None

## Other changes
There is one small unrelated change included. When an address validation service is enabled for a shop, and then the plugin for that service is removed, and an address validation request is made, there is now a proper error explaining the situation instead of a hard-to-debug `undefined` error.

## Testing
Start the app. Look in the database and verify that the dropped indexes have been dropped.


aldeed:
Merged in develop to fix the new conflict

imports/plugins/core/versions/server/migrations/59_drop_indexes.js
```diff
@@ -0,0 +1,125 @@
...
+      Shops.dropIndex("c2_layout.$.structure.notFound");
+      Shops.dropIndex("c2_layout.$.structure.template");
+      Shops.dropIndex("c2_shopType");
+      Shops.dropIndex("c2_workflow.status");
+
+      Translations.dropIndex("c2_i18n");
+      Translations.dropIndex("c2_shopId");
+    } catch (error) {
+      // This may fail if the index doesn't exist, which is what we want anyway
+      Logger.warn(error, "Caught error from dropIndex calls in migration 59");
```
ticean: Question, not a blocker: Can other error cases catch here? Would it be worth the effort to handle a specific "index doesn't exist" or is it OK to swallow all?

aldeed: Agree that would have been better but in this case any other error should be highly unlikely and it is still logged, so won’t be swallowed completely

PR #5112, opened 2019-04-10
Add a GraphQL schema linter, fix found issues

aldeed:
Impact: **minor**  
Type: **feature**

## Issue
People often forget or don't know about our conventions for the GraphQL schema files, such as always including field descriptions, alphabetizing fields, and more.

## Solution
I've added [graphql-schema-linter](https://github.com/cjoudrey/graphql-schema-linter) as a dev dependency, configured it, and defined a few additional custom rules for it.

It found 300+ style issues with the `.graphql` files, which I've also fixed in this PR. Mostly missing descriptions and fields not alphabetized.

I also added this linting as a CI task.

## Breaking changes
None

## Testing
1. Verify you can run the GraphQL linter and there are no issues found: `npm run lint:gql`
2. Remove a description from a field, re-run the linter, and verify that is complains about the missing description.
3. Make sure the API still starts up OK and you can run some GraphQL queries.


.reaction/graphql-linter-rules/type_fields_sorted_alphabetically.js
```diff
@@ -0,0 +1,38 @@
...
+              sortedFieldList.join(', '),
+            [node]
+          )
+        );
+      }
+    }
+  };
+}
+
+function arraysEqual(a, b) {
```
focusaurus: Maybe DRY this up a bit with a helper function/module. (The `arraysEqual` function and the slice/sort bit)

package.json
```diff
@@ -217,6 +218,7 @@
     "inspect-brk-docker": "node --experimental-modules  ./.reaction/scripts/run.mjs --inspect-brk=0.0.0.0:9229",
     "lint": "eslint .",
     "lint:errors": "eslint . --quiet",
+    "lint:gql": "graphql-schema-linter **/schemas/*.graphql",
```
focusaurus: Why are we globbing on `/schemas`? There's several `.graphql` files that don't match this glob I think.

aldeed: I couldn't really find a better way to do it. We need a glob that omits all `.graphql` files that are client/query files because this doesn't validate those. By convention our server schemas are almost always in a `schemas` folder so I landed on this. I'm open to any suggestions.

.reaction/graphql-linter-rules/type_fields_sorted_alphabetically.js
aldeed: Done

.reaction/graphql-linter/util/listIsAlphabetical.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+/**
+ * @summary Returns `true` if the list is in alphabetical order,
+ *   or an alphabetized list if not
+ * @param {String[]} list Array of strings
+ */
+function listIsAlphabetical(list) {
+  const sortedList = list.slice().sort();
+  if (arraysEqual(list, sortedList)) return true;
```
focusaurus: The return types seem inconsistent `true` and array. Do we want just `return arraysEqual(list, sortedList);`?

focusaurus: Oh hmm. I guess I don't really like this API. I expect an function with `is` in the name to always return a boolean. I get it now after closer inspection but maybe return `{isSorted: false, sortedList}` ?

aldeed: That does seem better. Updated.

PR #5119, opened 2019-04-12
ShopId can be any string

Akarshit:
Since clients would demand shopId based on their need, we should have flexibility in the type of string that can be a shopId.


zenweasel:
@Akarshit You need to sign your commits

PR #5120, opened 2019-04-12
fix: nnnnat 5113 password reset form

### LINKED ISSUE #5113: Emails: Reset Password button in email doesn't go to reset password form
## Prerequisites

## Issue Description
The email sent when you request to reset your password has a button. When you click that button, it goes to the sign in page rather than a reset password form.

### Setup
1. Log in as operator.
2. Edit Shop settings and make sure the shop email address is testing@reactioncommerce.com. This address is verified to send emails through the Reaction testing SES account. Other addresses may not work.
3. Go to the operator user profile page and make sure you have a real email address. You can use something like `<yourname>+admin@reactioncommerce.com`.
4. Enter correct email service credentials on Email page. Make sure the indicator is green.

### Reproduce
1. Log out.
2. Click Reset password link, enter admin email, and click Reset your password.
3. Click the button in the email you receive.

## Possible Solution
May need to include a query string in the URL and change the state of the sign-in form based on that.

### Versions
2.0.0 RC 11

nnnnat:
Resolves #5113   
Impact: **critical**  
Type: **bugfix**

## Issue
The email sent when you request to reset your password has a button. When you click that button, it goes to the sign in page rather than a reset password form.

## Solution
After talking with @mikemurray we believe this bug may have been introduced during the switch over to the newer `/operator` UI. Mike can correct me if I'm remembering wrong, but these changes cause the routes in the plugin's `register.js` file to be ignored and for the Meteor app to only display either the login view or the admin view. 

I've update the Account login component to set the `currentView` property to be `loginFormUpdatePassword` if the component is being displayed on the "reset-password" route. In the Account auth container I display the `updatePassword` component then redirect to the /account/login route once the password has been reset.

## Breaking changes
N/A

### Setup
1. Log in as operator.
2. Edit Shop settings and make sure the shop email address is testing@reactioncommerce.com. This address is verified to send emails through the Reaction testing SES account. Other addresses may not work.
3. Go to the operator user profile page and make sure you have a real email address. You can use something like `<yourname>+admin@reactioncommerce.com`.
4. Enter correct email service credentials on Email page. Make sure the indicator is green.

### Test Steps
**Admin Account**
1. Log out.
2. Click Reset password link, enter admin email, and click Reset your password.
3. Click the button in the email you receive.
4. Enter a new admin account password.
5. You should be redirected to the Signin form. Enter your admin email and **new** password.
6. You should be redirected back to the operator logged into your admin account.

**Customer Account**
1. Create a new customer account with a real email address. You can use something like `<yourname>+customer1@reactioncommerce.com`.
2. After you've created the account log out.
2. Click Reset password link, enter customer email, and click Reset your password.
3. Click the button in the email you receive.
4. Enter a new customer account password.
5. You should be redirected to the Signin form. Enter your customer email and **new** password.
6. You should be redirected back to the storefront homepage logged into your customer account.


imports/plugins/core/accounts/client/containers/auth.js
```diff
@@ -153,12 +149,26 @@ class AuthContainer extends Component {
...
     });
-  }
+  };
 
-  hasPasswordService = () => !!Package["accounts-password"]
+  hasPasswordService = () => !!Package["accounts-password"];
 
   renderAuthView() {
     if (this.props.currentView === "loginFormSignInView") {
+      if (this.props.currentRoute.route.name === "reset-password") {
```
nnnnat: @aldeed is this where & how I should be checking if we need to display the`UpdatePassword` form?

nnnnat:
After some more testing it seemed like the `core/account/auth` container isn't expecting to display a reset password form. I've done a simple change and now I'm able to reset my account password but the redirects between the storefront and login pages are buggy.
@mikemurray or @aldeed do either of you know how this was working before hydra? The resetting of the password seems to work but I'm unsure of my approach.

imports/plugins/core/accounts/client/containers/auth.js
aldeed: @nnnnat In the `registerPackage` call for the accounts plugin, I see this:

```
{
    route: "/reset-password/:token/:status?",
    template: "loginFormUpdatePassword",
    workflow: "none",
    meta: { noAdminControls: true },
    name: "reset-password",
    label: "reset-password"
  }, {
    route: `/${ENROLL_URI_BASE}/:token/:status?`,
    template: "loginFormUpdatePassword",
    workflow: "none",
    meta: { noAdminControls: true },
    name: ENROLL_URI_BASE,
    label: "Account Enroll"
  }
```

These may need to get updated and moved into the hydra-oauth plugin, but I'm not sure the exact solution without playing around with it a lot more. @impactmass may have a better idea of what to do. But essentially we need those routes to work again, and make sure the email is linking to the `reset-password` route.

Generally the way you're doing this here feels a little bit off. It seems like you could do this in the `Login` component where it's already examining the `currentRoute`, and then set a new value for `currentView`, which you can check here.

impactmass: @nnnnat, sorry Nat, I need to look a bit more to understand why this isn't working anymore and then be able to suggest approaches.  

My recollection is that those routes pasted above were loading up fine, and I remember there is a use of that props.currentRoute somewhere in the signup/in components to toggle the views. But, I'll need to check

impactmass:
@nnnnat 

> I'm able to reset my account password but the redirects between the storefront and login pages are buggy. mikemurray or aldeed do either of you know how this was working before hydra?

Redirecting back to storefront was not implemented before now. Reason: In cases where there are multiple client UIs that can perform a reset password flow, there is no easy way to know where the redirect should go to. An easy option is to include some bit of client's info in the URL sent to the user (not ideal based on discussions with team). If one of the client UIs is a mobile app, it becomes even more tricky. That was why we displayed a page with info on next steps for the user, as at the time we worked on this.

But implementers that are using **only 1** single client UI can hard set a redirect path.

aldeed:
@nnnnat @impactmass  After thinking about all of the email link issues, I think all of them have a common need, which is that we need to store several URLs on Shop and provide some simple text fields in operator UI for setting them. I'm thinking:

- `storefrontHomeUrl`
- `storefrontOrderUrl` (can have `:orderReferenceId` and `:orderToken` in the string)
- `storefrontOrdersUrl` (can have `:accountId` in the string)
- `storefrontAccountProfileUrl` (can have `:accountId` in the string)

Then these can be resolved and included in the email data. Also you will then be able to redirect to the shop's `storefrontHomeUrl` after password reset.

I should have some time tomorrow to work on this.

nnnnat:
@aldeed I've made a few updates to this branch to set the `currentView` prop in the `login` component and added the `updatePassword` component to the `auth` container. I worked with @mikemurray on trying to get the routes in the `register.js` to work again, but they're now being ignored and only the login or operator view are being displayed. We think checking for the route in one of the components might be the best way to do this now, what are your thoughts?

mikemurray:
If we need routes, we should avoid using the old routing API altogether and opt for a simpler solution like the `registerOperatorRoute` API. That would make generic routes configurable, million times less convoluted.

I opted to not include the old routes because there are so many of them that have no use or were never used.

aldeed:
I agree with @mikemurray that if we can make it work with new routing, we should. As far as having the proper URL available for redirecting, I wrote up my proposal in this issue: https://github.com/reactioncommerce/reaction/issues/5144

Even after that is done, there would still be the question of whether to redirect to admin UI or storefront UI after completing the reset. My thought is that if they have "admin" role, redirect to operator UI home, else storefront home. Even if an admin actually was trying to log into the storefront, they'll be smart enough to know how to get there. But it's more likely they meant to log in to operator.

aldeed:
@nnnnat The core PR with the URL data has been merged. UI is not done yet, but you can set directly in the database for testing. For redirect after successful password reset, you'll want to get `shop.storefrontUrls.storefrontHomeUrl` and use that for the redirect for non-admins, otherwise redirect to the operator home page for admins.

nnnnat:
@mikemurray I think this is ready for another review. I've had some issues getting the `coreLayout` component to work with this, I think I got it sorted out but would like for you to take a look.

imports/plugins/core/accounts/client/containers/updatePassword.js
```diff
@@ -76,7 +75,12 @@ const wrapComponent = (Comp) => (
         } else {
           // Now that Meteor.users is verified, we should do the same with the Accounts collection
           Meteor.call("accounts/verifyAccount");
-          Router.go(this.props.onCompleteRoute);
+          const { storefrontUrls } = Reaction.getCurrentShop();
+          if (Reaction.hasAdminAccess()) {
+            Router.go("/operator");
+          } else {
+            window.location.href = `${storefrontUrls.storefrontHomeUrl}/signin`;
```
mikemurray: Is it possible that `/signin` could be `/sign-in` or `/login` in a different storefront. Should there be another variable for `storefrontSignInUrl`?

nnnnat: I had the same thought, I think it make sense to have that URL as one of the `storefrontUrls`. @kieckhafer & @aldeed what are y'all thoughts on adding the login URL?

nnnnat:
@mikemurray aside form the comment above have you been able to successfully test this change? I'd like to merge this and update the `storefrontUrls` in a follow up PR if we can

imports/plugins/core/accounts/client/containers/updatePassword.js
kieckhafer: Yeah, I think that makes sense to add it, can't hurt.

mikemurray:
@nnnnat This is otherwise fine. Make a ticket for the remaining issues, and I'll approve these changes.

nnnnat:
storeforntUrls ticket [here](https://github.com/reactioncommerce/reaction/issues/5162)

PR #5121, opened 2019-04-12
Add correct item images to order email data

aldeed:
Impact: **minor**  
Type: **feature**

## Issue
The way image URLs were being added to the data object for order emails was not taking into account plugins with image transformations.

## Solution
Order items in the email data now go through the same transformations that we use for GraphQL, resulting in the correct resolved image URLs.

## Breaking changes
None

## Other changes
I included a migration that removes all docs from Assets collection that have the "template" type. Otherwise this interferes with resetting default templates because the files cached in Assets keep getting re-imported rather than the template files from code.

## Testing
Place an order and verify that the correct image still shows for all items in the email. Any further testing requires a custom plugin setup.


machikoyasuda:
@manueldelreal I tested that the plugins work. Can you confirm that this plugin doesn't break how current Core emails get images?

PR #5122, opened 2019-04-12
[WIP] fix: Fix critical performance issue caused by catalog items query for `featured` products

### LINKED ISSUE #5116: Catalog Items query "featured" sort performs poorly with large dataset
## Issue Description

Querying `catalogItems` with `sortBy featured` option has severe performance issues when there is a large number of products in the catalog.

## Possible Solution

Revert to the old aggregation, while it's not perfect, should perform better than the current implementation. This would be temporary while we reassess a more permanent solution.

mikemurray:
Resolves #5116   
Impact: **breaking,critical**  
Type: **feature,bugfix,performance,refactor**

## Issue

Querying `catalogItems` for `featured` products can cause Mongo/Reaction to become unresponsive as it tries to load and process every document in the `Catalog` collection.

## Solution

Cursor-based pagination with (first, last, before, and after) does not seem to work well with Mongo Aggregations as we have implemented it.

Cursor-based pagination has been thrown out for Mongo Aggregations for the time being. Instead, it's been replaced with offset based pagination (limit, and page).

The GQL query `catalogItems()` now supports `limit, page` as parameters for more traditional pagination for all `sortBy` types. This is in addition to the cursor based pagination parameters.

Sort by featured `sortBy: featured` **ONLY** supports offset based pagination.

If you use any combination of (first, last, before, after) and (limit, skip) together in the same query, an error will be thrown.

## Breaking changes

query `catalogItems(sortBy: featured, ...)` now only supports offset-based pagination. This means that pagination of featured products must be done with limit and page, like so... `catalogItems(sortBy: featured, limit: 20, page: 1, tagIds: [...] shopIds: [...])` 

## Testing

**Admin (offset-based-pagination-aggregation)**
This test is to ensure the productsByTagId, which has been updated to be exclusively offset-based pagination for a mongo aggregation works as intended.

1. Using a super-duper large dataset that caused this issue in the first place ;) ...
2. Open the operator UI to http://localhost:3000/operator/tags
3. Load a tag that has a lot of products. ( you might also get some from this link if your data set has it, http://localhost:3000/operator/tags/edit/cmVhY3Rpb24vdGFnOjJNWWdtU2NkZGdvb0tvTGpn )
3. Use the pagination controls and ensure you can go next, back, and increase the limits per page.

**Starterkit, (cursor-based-pagination)**
This is a backward compatibility check.

1. Ensure the product grids load and paginate properly.
2. Ensure the URL has some combination of `first, last, before, or after`

### DIFF
--- a/imports/plugins/core/catalog/server/no-meteor/queries/catalogItemsAggregate.js
+++ b/imports/plugins/core/catalog/server/no-meteor/queries/catalogItemsAggregate.js
@@ -36,13 +36,6 @@ export default async function catalogItemsAggregate(context, { isSoldOut, shopId
     }
   };
 
-  // defaultSort by id
-  const defaultSort = {
-    $sort: {
-      "product._id": 1
-    }
-  };
-
   const tag = await Tags.findOne({
     _id: tagId
   });
@@ -102,6 +95,6 @@ export default async function catalogItemsAggregate(context, { isSoldOut, shopId
 
   return {
     collection: Catalog,
-    pipeline: [match, defaultSort, addFields, projection, sort]
+    pipeline: [match, addFields, projection, sort]
   };
 }
--- a/imports/plugins/core/catalog/server/no-meteor/resolvers/Query/catalogItems.js
+++ b/imports/plugins/core/catalog/server/no-meteor/resolvers/Query/catalogItems.js
@@ -1,7 +1,11 @@
-import { getPaginatedResponse, getPaginatedAggregateResponse } from "@reactioncommerce/reaction-graphql-utils";
+import ReactionError from "@reactioncommerce/reaction-error";
 import { decodeShopOpaqueId } from "@reactioncommerce/reaction-graphql-xforms/shop";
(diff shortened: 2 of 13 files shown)


mikemurray:
@aldeed The GraphQL linter is mad since I removed `first, last, before, and after` from the `productByTagId` query. This is that error.

```
171:3 A field that returns a Connection Type must include forward pagination arguments (`first` and `after`), backward pagination arguments (`last` and `before`), or both as per the Relay spec.  relay-connection-arguments-spec-custom
```

The fix seems rather silly, I have to remove the word `Connection` from `TagProductConnection`, then it's fine. Do you have any recommendations like a new suffix, use "TagProducts" or something else?

aldeed:
Note that #5228 may supersede this now

rahbar8819:
vallagse mama

spencern:
@mikemurray are you still working on this? Should we update it? Close it?

machikoyasuda:
closing b/c we have https://github.com/reactioncommerce/reaction/pull/5228 now

PR #5124, opened 2019-04-15
(fix) remove billingAddress transformation in graphQL resolver for Payments

cshepherd:
Currently what is happening is that a customer implementation is using the withOrder/withOrders containers on the client (React) side which is relying on Payment.billingAddress to be populated.

The GraphQL output currently looks like this:
Screenshot from 2019-04-11 17-31-49

It looks like the transform that this PR references may not be necessary, as removing it would appear to give the desired results, ie:

Screenshot from 2019-04-11 17-47-32

Tagging @aldeed for review as you most recently worked on this part of the GQL code in reaction and will be able to dispense swift guidance.


cshepherd:
closing to resubmit properly

PR #5125, opened 2019-04-15
(fix) remove transform for billingAddress in gql resolver for Payment

cshepherd:
Currently what is happening is that a customer implementation is using the withOrder/withOrders containers on the client (React) side which is relying on Payment.billingAddress to be populated.

The GraphQL output currently looks like this:
Screenshot from 2019-04-11 17-31-49

It looks like the transform that this PR references may not be necessary, as removing it would appear to give the desired results, ie:

Screenshot from 2019-04-11 17-47-32

Tagging @aldeed for review as you most recently worked on this part of the GQL code in reaction and will be able to dispense swift guidance.

### DIFF
--- a/imports/plugins/core/payments/server/no-meteor/resolvers/Payment/index.js
+++ b/imports/plugins/core/payments/server/no-meteor/resolvers/Payment/index.js
@@ -1,6 +1,5 @@
 import { encodePaymentOpaqueId } from "@reactioncommerce/reaction-graphql-xforms/payment";
 
 export default {
-  _id: (node) => encodePaymentOpaqueId(node._id),
-  billingAddress: (node) => node.address
+  _id: (node) => encodePaymentOpaqueId(node._id)
 };


cshepherd:
`find . -name resolvers | xargs grep -ri payments` agrees that the only query resolver that sends payments is `/imports/plugins/core/orders/server/no-meteor/resolvers/Order/index.js`

To Reproduce Issue:

1. Login as admin@localhost from new/incognito window to ensure no previous session
2. Open Chrome Web Inspector
3. Visit account/orders page
4. Look at graphQL request for ordersByAccountIdQuery and observe null Payment.billingAddress as in first image above

Testing Procedure:

1. Login as admin@localhost from new/incognito window and ensure no previous session
2. Open Chrome Web Inspector
3. Visit account/orders page (or any other page that uses the withOrder or withOrders containers, thus invoking this graphQL resolver)
4. Look at graphQL request for ordersByAccountIdQuery and observe populated Payment.billingAddress as in second image above

PR #5126, opened 2019-04-16
feature: add Catalog query filters

willopez:
Impact: minor  
Type: feature

## Feature
Adds Boolean filtering support for the `catalogItems` query. Filters are specified as a key value pair of a supported filter field,(see below) and a boolean value. A `booleanFilters` param will need to be sent with requests to the `catalogItems` query. These filters are optional.

Example query
```
query catalogItemsQuery($shopId: ID!, $tagIds: [ID], $booleanFilters: [CatalogBooleanFilter], $first: ConnectionLimitInt, $last:  ConnectionLimitInt, $before: ConnectionCursor, $after: ConnectionCursor, $sortBy: CatalogItemSortByField, $sortByPriceCurrencyCode: String, $sortOrder: SortOrder) {
  catalogItems(shopIds: [$shopId], tagIds: $tagIds, booleanFilters: $booleanFilters, first: $first, last: $last, before: $before, after: $after, sortBy: $sortBy, sortByPriceCurrencyCode: $sortByPriceCurrencyCode, sortOrder: $sortOrder) {
.
.
.
  }
}

```

Example filters:

```
[
        {
          name: "isSoldOut",
          value: false
        },
        {
          name: "isBackorder",
          value: false
        }
]
```

Currently supported fields 
```
isDeleted | Boolean
isVisible | Boolean
isLowQuantity | Boolean
isSoldOut | Boolean
isBackorder | Boolean
```
## Testing
1. In the storefront, add filters as specified above
2. confirm products are correctly filtered


willopez:
@mikemurray FYI, there are some tests failing that were already failing. I fixed the one related to this PR.

willopez:
@mikemurray this is now ready for review, the underlying inventory issue has been resolved.

willopez:
@mikemurray I have refactored the filters to be of Boolean type and restricted the props on which to filter on with an enum.

imports/plugins/core/catalog/server/no-meteor/schemas/schema.graphql
```diff
@@ -27,6 +45,18 @@ type PriceRange {
   range: String!
 }
 
+"Represents a name of a filter to be applied to the Catalog collection"
+input CatalogFilter {
```
aldeed: Currently GraphQL does not support input unions, and we'll definitely need non-bool filtering eventually. So I think I'd call this `CatalogFilterBoolean` and change the input field to `booleanFilters: [CatalogFilterBoolean]`. That way we can add `stringFilters`, etc. later.

willopez:
@aldeed I have resolved all your previous comments. Wondering if I should created indices for 
```
isDeleted
isVisible
isLowQuantity
isSoldOut
isBackorder
```

aldeed:
I'm approving the code but I did not test. @willopez Indexes are generally not very useful for fields with boolean values. The more possible values a field has, the more an index helps. But by their nature boolean fields only have 2 possible values. So I think we can hold off adding indexes for these unless we find out they are indeed helpful in the wild.

Issue #5111, opened 2019-04-09
bin/setup should handle .env missing trailing newline


focusaurus:
## Prerequisites
* [x] Are you running the latest version?
* [x] Are you able to consistently reproduce the issue?
* [x] Did you search the issue queue for existing issue? [Search issues](https://github.com/reactioncommerce/reaction/issues)

## Issue Description

If your `.env` file ends without a trailing newline, `bin/setup` won't correct that and thus you'll end up with 2 variables jumbled on the same line.

**Example .env before**

```
A=a
B=b
OOPSNONEWLINE=oops
```

**Example .env after ./bin/setup**
```
A=a
B=b
OOPSNONEWLINE=oopsC=c
```

`C=c` should go on a new line, but it gets put on the end of the preceeding line.

## Steps to Reproduce
Please provide starting context, i.e. logged in as a user, configure a particular payment method.
1. Remove the final newline from your `.env` file.
  - This is actually somewhat tricky to do in normal editing since most editors add a final newline automatically
  - `echo -n FOO=foo >> .env`
2. Make sure your `.env` is missing at least one variable from `.env.example`
3. Run `./bin/setup`

## Possible Solution
`setup` could detect this explicitly and fix it.

### Versions
(run `reaction -v` from your reaction directory)

focusaurus:
Might make a separate function that runs at the beginning and ensures `.env` has a final newline (via `tail -1 | grep -E` or something).

aldeed:
Note that #5035 rewrites this script, but I didn't check to see whether that rewrite fixes this issue. I'm guessing probably not.

mikemurray:
Resolved in #5156

PR #5127, opened 2019-04-16
Fix issue: 5111 - fixing when variable is appended on the same line w…

sterowney:
Fix for the issue: https://github.com/reactioncommerce/reaction/issues/5111

Add each missing variable with a newline, then after that is finished, tidy up and remove any empty lines in the .env file.


aldeed:
Thanks. Looks like this was fixed elsewhere.

PR #5132, opened 2019-04-17
Feat: Support custom login error message

ajporlante:
Impact: **minor**  
Type: **feature**

## Issue
Login error messages are the default ones returned by the meteor login function. Some implementations will need custom error messages.

## Solution
I added a Meteor method that calls no-meteor functions registered through `functionsByType` with type `customizeLoginErrorMessage`.

## Breaking changes
None

## Testing
1. Without extra setup, you should be able to login. If you enter incorrect credentials, you should see default error messages.
2. In a plugin, register a function of type `customizeLoginErrorMessage`

For example:

In `register.js`:
```
import customizeLoginErrorMessage from "./server/no-meteor/customizeLoginErrorMessage";
...
functionsByType: {
    customizeLoginErrorMessage: [customizeLoginErrorMessage],
}
...
```

In `/server/no-meteor/customizeLoginErrorMessage.js`:
```
export default async function customizeLoginErrorMessage(context, username, error) {
  return {
    ...error,
    reason: "Due to a system upgrade, you must reset your password before you can access your account."
  };
}
```

Login with an invalid credential. You should see a custom error message.


ajporlante:
Sorry, I'm closing this because of messed up merges.

PR #5133, opened 2019-04-17
Feat: Support custom login error message

ajporlante:
Impact: **minor**  
Type: **feature**

## Issue
Login error messages are the default ones returned by the meteor login function. Some implementations will need custom error messages.

## Solution
I added a Meteor method that calls no-meteor functions registered through `functionsByType` with type `customizeLoginErrorMessage`.

## Breaking changes
None

## Testing
1. Without extra setup, you should be able to login. If you enter incorrect credentials, you should see default error messages.
2. In a plugin, register a function of type `customizeLoginErrorMessage`

For example:

In `register.js`:
```
import customizeLoginErrorMessage from "./server/no-meteor/customizeLoginErrorMessage";
...
functionsByType: {
    customizeLoginErrorMessage: [customizeLoginErrorMessage],
}
...
```

In `/server/no-meteor/customizeLoginErrorMessage.js`:
```
export default async function customizeLoginErrorMessage(context, username, error) {
  return {
    ...error,
    reason: "Due to a system upgrade, you must reset your password before you can access your account."
  };
}
```

Login with an invalid credential. You should see a custom error message.

### DIFF
--- a/imports/plugins/core/accounts/server/methods/customizeLoginErrorMessage.js
+++ b/imports/plugins/core/accounts/server/methods/customizeLoginErrorMessage.js
@@ -1,0 +1,27 @@
+import Logger from "@reactioncommerce/logger";
+import { check } from "meteor/check";
+import getGraphQLContextInMeteorMethod from "/imports/plugins/core/graphql/server/getGraphQLContextInMeteorMethod";
+
+/**
+ * @name accounts/customizeLoginErrorMessage
+ * @memberof Accounts/Methods
+ * @method
+ * @param {String} username - username/email entered by a user
+ * @param {Object} error - error object returned by Meteor.loginWithPassword
+ * @returns {Object} custom error object
+ */
+export default function customizeLoginErrorMessage(username, error) {
+  check(username, String);
+  check(error, Object);
+
+  let customError = error;
+  const context = Promise.await(getGraphQLContextInMeteorMethod());
+  const { getFunctionsOfType } = context;
+  const customizeErrorFunctions = getFunctionsOfType("customizeLoginErrorMessage");
+  if (customizeErrorFunctions.length > 0) {
+    customError = customizeErrorFunctions[0](context, username, error);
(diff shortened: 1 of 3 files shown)


imports/plugins/core/accounts/server/methods/customizeLoginErrorMessage.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+export default function customizeLoginErrorMessage(username, error) {
+  check(username, String);
+  check(error, Object);
+
+  let customError = error;
+  const context = Promise.await(getGraphQLContextInMeteorMethod());
+  const { getFunctionsOfType } = context;
+  const customizeErrorFunctions = getFunctionsOfType("customizeLoginErrorMessage");
+  if (customizeErrorFunctions.length > 0) {
```
focusaurus: Should we be calling all the functions in the array instead of just the first?

ajporlante: Not sure to be honest as I there needs to be only one result anyway. I just got inspired by this:

https://github.com/reactioncommerce/reaction/blob/dec31982854ca4b1e907319ac138d49c3eac6dc8/imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js#L275-L292

where the `getFuctionsOfType` only applies the first function.

aldeed: It's fine for code to choose to (and document that it will) only call the first function. But in this case, I think @focusaurus is right that it would make more sense to call them all in series and stop as soon as one returns something. That way different plugins can customize different errors.

ajporlante:
@aldeed @focusaurus Revised as requested and I also added `await` for the custom functions just in case they will be `async`.

aldeed:
@ajporlante Looks good and good idea awaiting it. There is one eslint issue you need to resolve before we can merge this.

PR #5134, opened 2019-04-17
fix: inventory quantity resets to 0 when any variant or option quantity is undefined

kieckhafer:
Impact: **major**  
Type: **bugfix**

## Issue 1
When auto-updating variant quantites, based off a variants options quantities, if any of the children have an `undefined` quantity, the count gets reset to `0`, causing an error and showing all inventory as `0`.

It's caused by not adding parens around the `||` clause of the inventory check. For example, evaluate
`50 + undefined || 0` vs `50 + (undefined || 0)`.

### Solution
Add parens where necessary.

-----

## Issue 2
When querying for reserved products, we check every order in a certain status. We need to filter this better to create less of a drain on memory.

### Solution
Add `"shipping.items.variantId": variant._id` filter to Orders query to only find orders with the variant we are querying.

-----

## Issue 3
Collection indexes were incorrect.

### Solution
Fix indexes, and add migration to remove old indexes


focusaurus:
@kieckhafer could you please add a test case for the fix for issue 1 with the `reduce` syntax mistake? Should be pretty straightforward to reproduce in tests against the unfixed code and then confirm the fix is correct.

PR #5136, opened 2019-04-17
Improved performance for products and catalog items queries with featured sort

aldeed:
Impact: **minor**  
Type: **performance**

## Issue
The "featured" sort for `productsByTagId` and `catalogItems` queries was slow with a large number of products.

## Solution
I've rewritten the code to be faster. Rather than trying to fix pagination of aggregation queries in a generic way, I've written a util function specifically for this type of aggregation. This is easier because it doesn't need to support arbitrary sortBy.

## Breaking changes
None.

## Testing
Verify that paging the `productsByTagId` and `catalogItems` queries is now faster with many products loaded, and the results are still correct.


mikemurray:
If I do `last: 20`, with or without `sortby: featued` on the `catalogItems` query, I get the memory limit error below. I can do `first: 200` no problem. This is with `1308` documents in the tag I'm using, 40k+ in the catalog collection.

**query. your tag ids may vary**
```
  catalogItems(
    shopIds: ["cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg=="],
    tagIds: [ "cmVhY3Rpb24vdGFnOjIyOTNXbktmUXg3RzNiZWVC"],
    sortBy: featured,
    last: 20
  )
```

**error**
```
Executor error during find command: OperationFailed: Sort operation used more than the maximum 33554432 bytes of RAM. Add an index, or specify a smaller limit.
```

aldeed:
The overall takeaway from @mikemurray's test results is that the queries that do NOT use the aggregation are slow when doing backwards pagination. (When doing featured sort but paginating backwards and `before` is outside the featured array, it just does normal cursor pagination.)

So technically this is something different than the original issue this PR is fixing, but it's still bad so I'm going to spend some time trying to find out what's wrong with this specific case.

mikemurray:
A similar issue is happening with offset based pagination from this PR: https://github.com/reactioncommerce/reaction/pull/5122

I hit the memory limit when I do `limit: 20, page: 33` total count is `1308`, so page `33` should be at product `660`. Page 32 and under work fine.

**sort options for my other query**
```
    sortBy: minPrice,
    sortByPriceCurrencyCode: "USD",
    limit: 20,
    page: 34
```

aldeed:
Here is a summary of the indexing situation:

- The `{ createdAt: 1 }` index should be `{ createdAt: 1, _id: 1 }` to be usable by standard `createdAt` sort of catalog items. Made that change in this PR. This has a big impact on sorting performance and gets rid of memory errors when backward sorting.
- The `{ updatedAt: 1 }` index should be `{ updatedAt: 1, _id: 1 }` to be usable by `updatedAt` sorting of catalog items. Made that change in this PR.
- We need to add an index to support `minPrice` sorting. `{ "product.pricing.USD.minPrice": 1, "_id": 1 }` works as long as prices are in USD. I don't want to make an index for every currency code, but I'm thinking we can create one just for the primary shop's default currency for now, pending a better rewrite of pricing.

aldeed:
Pushed a fix to the bug pointed out by @mikemurray and merged in `develop`

PR #5140, opened 2019-04-22
feat: Migrations & related checks can be disabled

ticean:
Resolves N/A
Impact: **minor**  
Type: **feature**

## Issue
The checks that verify the migration state are important to run the application in a safe and sane way. However, in some emergency circumstances it can be necessary to disable checks so that the application may be started when the migration versions don't line up.

An example scenario might be an emergency release rollback.

## Solution
Adds new environment variables to flag the migration feature. 

- `MIGRATION_BYPASS_ENABLED`, boolean, not required, default: false

Code changes disable migrations and checks when the flag is true.

:warning: **Running in such a mode is inherently unsafe!** It is the operator's responsibility to decide if it is OK to run in this mode. The difference between the target and existing migration state should be examined carefully.

## Breaking changes
None.

#### Migrations are Enabled
1. Ensure `MIGRATION_BYPASS_ENABLED` flag is false
2. Add a migration.
3. Migrations should run on startup.

#### Migration Checks are Enabled
1. Ensure `MIGRATION_BYPASS_ENABLED` flag is false
2. Add a migration.
3. Run migration.
4. Rollback code to older version.
5. Migration check should run at startup. App should not start.

#### Migrations are Disabled
1. Ensure `MIGRATION_BYPASS_ENABLED` flag is true
2. Add a migration.
3. Migrations should not run at startup.
4. Warning should appear in logs, `"DANGER: MIGRATION_BYPASS_ENABLED is true and all migration activity is bypassed. This is a dangerous mode and may result in data corruption."`

#### Migration Checks are Disabled
1. Ensure `MIGRATION_BYPASS_ENABLED` flag is true
2. Add a migration.
3. Run migration.
4. Rollback code to older version.
4. Migration check should not run at startup. App should start.

### DIFF
--- a/imports/node-app/core/config.js
+++ b/imports/node-app/core/config.js
@@ -3,6 +3,10 @@ import envalid, { bool, str } from "envalid";
 export default envalid.cleanEnv(process.env, {
   GRAPHQL_INTROSPECTION_ENABLED: bool({ default: false }),
   GRAPHQL_PLAYGROUND_ENABLED: bool({ default: false }),
+  MIGRATION_BYPASS_ENABLED: bool({
+    default: false,
+    desc: "Bypasses migration version checks and migration runs. Enables startup if migration state is not compatible. This can be dangerous enough to cause data inconsistencies. Use at your own risk!"
+  }),
   // This is necessary to override the envalid default
   // validation for NODE_ENV, which uses
   // str({ choices: ['development', 'test', 'production'] })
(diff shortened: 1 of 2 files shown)


ticean:
Thanks for the lint fix @aldeed! I can squash the commits down to a single, signed commit once the test/lint suites pass.

zenweasel:
@ticean Sign your commits and let's get this merged. 🎉

aldeed:
Oops, sorry I did the lint fix on github and forgot to add my signature. A signed squash sounds good @ticean

ticean:
All squashed with no changes. The DCO is passing.

PR #5142, opened 2019-04-22
refactor: use `getVariantPrice` to determine price in `xformCartGroupToCommonOrder`

kieckhafer:
Impact: **critical**  
Type: **refactor**

## Issue
We are breaking our pricing out from being stored on the `Product`. The consequences of this mean that we cannot reliably take the price currently stored on the product and use it in various cart transforms. This PR takes our function `getVariantPrice`, which takes the price from the `Catalog` item, and returns the correct price to use in our transform.

## Testing
- Mkae sure prices are showing up correctly on cart items after they are run through `xformCartGroupToCommonOrder`.

## Breaking changes
Older shops must confirm they are using our `simple-pricing` plugin, and that `pricing` is available on Catalog items.


kieckhafer:
@aldeed @rosshadden 

I'm opening this PR as a place for discussion. I don't think my solution here is the way this should be resolved. We shouldn't be importing a custom plugin function into core files like this.

@rosshadden mentioned that we should probably create some internal way to know if we are using the pricing engine or not, and use a single field on the `Catalog` item for the price, instead of having two different fields that may or may not be set.

What are your thoughts on the best way to go about this?

rosshadden:
I believe @aldeed is working on abstracting the current core pricing code into a core plugin. And I think one of the goals is for the interfaces of the two pricing plugins to be the same.

kieckhafer:
@rosshadden thanks.

@aldeed So can I move forward under the assumption that no matter if a price comes from the pricing engine, or another place, it will always be available in the same place on the `Catalog` item that I'm reading from? Something like `prices.currentPrice`?

aldeed:
@kieckhafer You should be able to use `context.queries.getVariantPrice` instead of importing it. The included `simple-pricing` plugin has it and any custom plugin will be expected to provide a replacement.

kieckhafer:
@aldeed so is what I've done here the right way to do this? All I need to change is to swap out the `getVariantPrice` helper with the `context` version?

aldeed:
@kieckhafer Pretty much what you have, except no need to check for `prices` or mention pricing engine in comments. Just `if (property === "price")` then value should be the `price` field from the object returned by `getVariantPrice`.

I would also add a unit test for it.

kieckhafer:
Oh Ok, I think that makes sense. So even if `Catalog.price` is present and not set to `0`, we won't use that, we'll still use `getVariantPrice`?

aldeed:
@kieckhafer I think it makes sense to ask the pricing plugin every time. We're moving away from having any prices stored on catalog at all, and we don't want to assume any particular pricing schema.

kieckhafer:
@aldeed `getVariantPrice` [looks for `pricing` on `catalogVariant`](https://github.com/reactioncommerce/reaction/blob/ref-kieckhafer-pricingEngineWithSurcharges/imports/plugins/included/simple-pricing/server/no-meteor/queries/getVariantPrice.js#L10), but that field isn't available on the item at this point.

In the cart, we are using the `extendedCommonOrder`, which has `items` on it, each of which have `price` from the `Products` collection, and optionally `prices` from the pricing engine... but where does `pricing` come from?

imports/plugins/core/cart/server/no-meteor/util/xformCartGroupToCommonOrder.js
```diff
@@ -1,4 +1,5 @@
 import { toFixed } from "accounting-js";
+import findVariantInCatalogProduct from "/imports/plugins/core/catalog/server/no-meteor/utils/findVariantInCatalogProduct";
```
aldeed: At some point recently I added `findVariantInCatalogProduct` as a context query. So you can remove this import and instead do `const { variant } = context.queries.findVariantInCatalogProduct(...)` below. That way we're not importing between plugins.

imports/plugins/core/cart/server/no-meteor/util/xformCartGroupToCommonOrder.js
```diff
@@ -13,27 +14,49 @@ export default async function xformCartGroupToCommonOrder(cart, group, context)
...
-    subtotal: {
-      amount: +toFixed(item.price.amount * item.quantity, 3),
-      currencyCode
-    },
-    taxCode: item.taxCode,
-    title: item.title,
-    variantId: item.variantId,
-    variantTitle: item.variantTitle
+  items = await Promise.all(items.map(async (item) => {
+    let itemPrice = item.price;
```
aldeed: If `item.price` is set, it would be the whole `{ amount, currencyCode }` object.

imports/plugins/core/cart/server/no-meteor/util/xformCartGroupToCommonOrder.js
```diff
@@ -13,27 +14,49 @@ export default async function xformCartGroupToCommonOrder(cart, group, context)
...
+
+    // Get the catalog version of the item to get pricing data from it
+    if (catalogItemsInGroup) {
+      const catalogProduct = catalogItemsInGroup.find((catalogItem) => catalogItem.product.productId === item.productId);
+      if (catalogProduct) {
+        const catalogVariant = findVariantInCatalogProduct(catalogProduct.product, item.variantId);
```
aldeed: The variant returned may be `null`. I think this should throw an error if that happens because it should never happen.

imports/plugins/core/cart/server/no-meteor/util/xformCartGroupToCommonOrder.js
```diff
@@ -13,27 +14,49 @@ export default async function xformCartGroupToCommonOrder(cart, group, context)
...
+        const variantPrice = await context.queries.getVariantPrice(context, catalogVariant, currencyCode);
```
aldeed: The variantPrice returned may be `null`. I think this should throw an error if that happens because it should never happen.

imports/plugins/core/cart/server/no-meteor/util/xformCartGroupToCommonOrder.js
kieckhafer: Yes... I made a note in that PR to update this one when it was available, wasn't sure if it was yet. Will update.

imports/plugins/core/cart/server/no-meteor/util/xformCartGroupToCommonOrder.js
kieckhafer: 👍 updated to be `itemPrice.amount`

kieckhafer:
@aldeed all updates made, ready for another look

kieckhafer:
@aldeed latest comment addressed, ready for another look

aldeed:
Approved but not merging since there are still other reviewers listed.

PR #5143, opened 2019-04-23
Finish simple-pricing plugin

### LINKED ISSUE #4969: Remove deprecated `price` field from Catalog schema
## Prerequisites

## Issue Description
There is a deprecated `price` field on the `CatalogProduct` and `CatalogProductVariant` schemas. There is also the newer `pricing` field and more pricing changes on the way, and having both is confusing. Now that classic storefront UI is being removed, we can remove `price` field from catalog.

## Work
- Remove `price` from the `CatalogProduct` and `CatalogProductVariant` schemas, both GraphQL and SimpleSchema.
- No longer set `price` in `createCatalogProduct.js`
- Migration to remove `price` from all documents in `Catalog` collection.
- Verify that these changes cause no issues in the operator and storefront UIs

The `Products` collection will still have `price`, which will still be converted to `pricing` when publishing to a catalog.

### LINKED ISSUE #4970: Move pricing fields and code to a simple-pricing included plugin
## Background
There are newer and better pricing models in development, but existing pricing code is currently spread across several core plugins. If a custom pricing plugin is added, it will have data alongside the current pricing fields, which would not be used. This would be confusing.

## Work
Move all code and schemas related to product and catalog pricing into an `included` plugin, which should be removable. The goal is to be able to remove it and swap in a different custom pricing plugin, with everything still working.

aldeed:
Resolves #4969 
Resolves #4970   
Impact: **major**  
Type: **feature**

## Big Picture Benefit
Because all pricing logic and schemas are encapsulated in the included `simple-pricing` plugin, installations that have their own custom pricing plugin do not have to jump through hoops to keep core pricing code happy. Default pricing behavior can be disabled by disabling the `simple-pricing` plugin.

## Changes
- Removed the deprecated `price` field from `CatalogProduct` and `CatalogProductVariant`. Ensured that no code uses it.  I opted not to do a migration to remove this from the database. It will be removed as each product is published, or you may manually unset it.
- Moved price-related functions, code, and schemas into the `simple-pricing` plugin. This plugin was created prior to this for this purpose, but we never finished migrating all pricing code to there.
- Some significant improvements to database lookups for pricing-related fields.

NOTE: Some operator UI client code has not yet been moved into the simple-pricing plugin. This will be done in a future PR.

## Breaking changes
The deprecated `price` field has been removed from `CatalogProduct` and `CatalogProductVariant`.

## Tangential Changes
Because it was needed for this work, I've added `app.registerPlugin`, which is a version of `Reaction.registerPackage` that can be called in a pure Node app. Soon we will update all plugins to use this, but for now `Reaction.registerPackage` still works and the API is the same for both functions.

## Testing
- Verify that prices look correct in UI everywhere
- Verify operator can update prices and publish updated prices to catalog
- Verify you can check out on storefront with correct prices
- Verify that you can disable the `simple-pricing` plugin, add a custom pricing plugin to replace it, and everything related to prices continues to work.


imports/plugins/core/core/server/startup/startNodeApp.js
```diff
@@ -48,20 +39,22 @@ export default async function startNodeApp() {
...
-      queries: finalQueries,
+      queries: coreQueries,
       rootUrl: ROOT_URL
     },
-    functionsByType,
     graphQL: {
       graphiql: Meteor.isDevelopment,
-      resolvers: finalResolvers,
-      schemas: [...coreSchemas, ...schemas]
+      resolvers: coreResolvers,
```
kieckhafer: I changed these fields a bit in a PR that was just merged, I can try to help resolve conflicts if they come up: https://github.com/reactioncommerce/reaction/pull/5145/files#diff-aec11b46e8437e7234ae8e5cc0c1a115

imports/plugins/core/catalog/server/no-meteor/queries/findVariantInCatalogProduct.js
```diff
@@ -1,4 +1,7 @@
 /**
+ * @summary Given a catalogProduct from the database, traverses it to find
```
kieckhafer: Just a note to update this PR with `context.queries.findVariantInCatalogProduct.js` once this is merged: https://github.com/reactioncommerce/reaction/pull/5142/files#diff-d442c3edf8473bfa06bd8fa3bf9d96ddR2

aldeed:
Merged in latest develop, fixed the conflicts @kieckhafer mentioned, and finished my testing of this. Adding @manueldelreal for any further testing you want to do before this is merged.

PR #5151, opened 2019-04-26
fix: race condition by awaiting each publish catalog function by type…

nnnnat:
Resolves 
Impact: **critical**  
Type: **bugfix**

## Issue
`publishProductToCatalog` function by type would be called synchronously, this would cause partial/incomplete Catalog publish if any of the functions by type were trying to do async operations.

## Solution
Await the custom function by type calls for the `publishProductToCatalog` function.

## Breaking changes
N/A

## Testing
1. As an admin "create" and "publish" a new Product.
2. As an Admin "edit" an existing Product and "publish"

### DIFF
--- a/imports/plugins/core/catalog/server/no-meteor/utils/createCatalogProduct.js
+++ b/imports/plugins/core/catalog/server/no-meteor/utils/createCatalogProduct.js
@@ -252,10 +252,11 @@ export default async function createCatalogProduct(product, context) {
   const catalogProduct = await xformProduct({ collections, product, shop, variants });
 
   // Apply custom transformations from plugins.
-  getFunctionsOfType("publishProductToCatalog").forEach((customPublishFunc) => {
+  for (const customPublishFn of getFunctionsOfType("publishProductToCatalog")) {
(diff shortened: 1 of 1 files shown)


imports/plugins/core/catalog/server/no-meteor/utils/createCatalogProduct.js
```diff
@@ -252,10 +252,11 @@ export default async function createCatalogProduct(product, context) {
   const catalogProduct = await xformProduct({ collections, product, shop, variants });
 
   // Apply custom transformations from plugins.
-  getFunctionsOfType("publishProductToCatalog").forEach((customPublishFunc) => {
+  for (const customPublishFn of getFunctionsOfType("publishProductToCatalog")) {
```
nnnnat: @aldeed you mentioned writing up a few test for this bug, what should I test for here? I'm looking to make sure each `customPublishFn` has been called before we return out of this func?

nnnnat:
I'm going to merge this PR and follow up  with another PR with test cases #5152.

PR #5154, opened 2019-04-26
Avoid unnecessary database queries for common GraphQL requests

aldeed:
Impact: **minor**  
Type: **performance**

## Changes
Some database queries frequently made for GraphQL requests are either removed or done only when necessary.

`applyPaginationToMongoCursor` and `getPaginatedResponse` util functions have been rewritten to do count queries for pagination only when requested. These queries are expensive on large collections and were often being done even though the client request did not even ask for paging fields. A new `wasFieldRequested` function is used everywhere to determine whether the `hasPreviousPage`, `hasNextPage`, and `totalCount` fields are included in a client request. Then the count queries are done only as necessary for these fields.

## Breaking changes
None

## Testing
These changes affect many GraphQL queries, but testing one or two such as `catalogItems` and `roles` should be sufficient.

When querying GraphQL and asking for `catalogItems` with no paging fields for a large Catalog collection, it should be much faster. When you DO ask for paging fields, they should be in the response and correct.

From a user perspective, some basic smoke tests of storefront and operator UI should be sufficient.


kieckhafer:
I'd suggest maybe we hold off a day or two on merging this - I've left a note in slack for all of our devs to double check any custom plugins that they are currently working on and make sure `context.shopId` is not a part of any of them.

imports/plugins/core/accounts/server/no-meteor/resolvers/Query/groups.js
```diff
@@ -11,19 +11,24 @@ import { decodeShopOpaqueId } from "@reactioncommerce/reaction-graphql-xforms/sh
...
  */
-export default async function groups(_, { shopId, ...connectionArgs }, context) {
+export default async function groups(_, { shopId, ...connectionArgs }, context, info) {
   // Transform ID from base64
   const dbShopId = decodeShopOpaqueId(shopId);
 
   const query = await context.queries.groups(context, dbShopId);
-  return getPaginatedResponse(query, connectionArgs);
+  return getPaginatedResponse(query, connectionArgs, {
+    includeHasNextPage: wasFieldRequested("pageInfo.hasNextPage", info),
```
kieckhafer: Would it be easier to maintain if we do this inside of `getPaginatedResponse.js` instead of passing it individually in every `getPaginatedResponse`?

aldeed: I did consider that, but I wanted to try to keep that function unaware of GraphQL. That way it could be used for some other interface, too. (In fact, GraphQL info isn't available in a couple places that call it.) But I agree it feels kind of yucky to repeat similar code everywhere. One idea might be to make a GraphQL-aware lightweight wrapper function. Maybe I'll do that.

aldeed:
Do not merge yet pending a few changes.

imports/plugins/core/accounts/server/no-meteor/resolvers/Query/groups.js
kieckhafer: Makes sense either way. I figured there was some reasoning behind it, just noticing the same code in 27 places I wanted to make sure the question was asked.

PR #5161, opened 2019-05-02
Fix: replaceBlock function

### LINKED ISSUE #5160: replaceBlock does not work
## Issue Description
`replaceBlock` function does not work.

## Steps to Reproduce
In a plugin, call the `replaceBlock` function, e.g.:

```
replaceBlock({
  region: "VariantDetailMain",
  block: "OptionTable",
  component: OptionTable
});
```
See that there is an error in the browser console:
```
Cannot read property 'undefined' of undefined
```

## Possible Solution
`replaceBlock` function just needs renaming of some parameter/variable names.

### Versions
`develop`

ajporlante:
Resolves #5160   
Impact: **critical**  
Type: **bugfix**

## Issue
`replaceBlock` function does not work due to inconsistent naming of variables.

## Solution
Just renamed the variables.

## Breaking changes
None

## Testing
In a plugin, call replaceBlock function. The block should be replaced by the new component.

### DIFF
--- a/imports/plugins/core/components/lib/blocks.js
+++ b/imports/plugins/core/components/lib/blocks.js
@@ -197,7 +197,7 @@ export function getBlocks(regionName) {
  * This function keeps track of the previous HOCs and wraps the new HOCs around previous ones
  * @param {Object} options Object containing block information
  * @param {String} options.region The region of the block that will be replaced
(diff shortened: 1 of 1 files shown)


mikemurray:
This should be good to go after the eslint issue is addressed.

PR #5169, opened 2019-05-06
Taxation based on origin

### LINKED ISSUE #5166: Taxation Location field is not used
## Issue Description
The `Taxes` schemas in `taxes-rates` and `taxes` provide fields to specify whether a tax should be applied based on the destination or origin of an order. This setting can neither be edited in the backend nor are taxes based on the origin applied.

## Steps to Reproduce
1. Go to the Taxes section in operator ui
2. Add or edit a custom rate
3. Note that there is no field to edit `taxLocale`

## Possible Solution
1. Add field to `custom.html`
2. Change calculation in `calculateOrderTaxes.js`

### Versions
Reaction: 2.0.0-rc.10

derBretti:
Resolves #5166   
Impact: **minor**  
Type: **feature**

## Issue
The Taxes schemas in taxes-rates and taxes provide fields to specify whether a tax should be applied based on the destination or origin of an order. This setting can neither be edited in the backend nor are taxes based on the origin applied.

## Solution
It is possible to chose between available values for `taxLocale` in tax settings (currently "destination" and "origin"). Both types of taxes, those applied based on the origin and those applied based on the destination are used when calculating taxes.

## Testing
1. Go to the Taxes section in operator ui
2. Add or edit a custom rate
3. Set the value of "Taxation Location" to "origin"
4. Make sure that the shop's address lies in a jurisdiction where the tax is applied
5. Buy an item for which the tax is applied
6. Note that the tax is calculated for the item during checkout

### DIFF
--- a/imports/plugins/included/taxes-rates/client/settings/custom.html
+++ b/imports/plugins/included/taxes-rates/client/settings/custom.html
@@ -32,6 +32,7 @@
             <span class="help-block">{{afFieldMessage name="region"}}</span>
           {{/if}}
           {{> afQuickField name='postal' class='form-control' placeholder="Postal Code"}}
+          {{> afQuickField name='taxLocale' options=localeOptions class='form-control'}}
           {{> afQuickField name='taxCode' class='form-control' placeholder="Tax only products with this tax code (optional)"}}
           {{> afQuickField name='rate' class='form-control' placeholder="Rate as a percentage"}}
         </div>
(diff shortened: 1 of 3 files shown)


aldeed:
@derBretti Sorry for the delay on this one. I've reimplemented your same API changes in the 3.0 stream here: https://github.com/reactioncommerce/reaction/pull/6076

We won't be merging this 2.x PR because 2.x is getting urgent fixes only and this is considered an unimplemented feature in 2.x rather than something that is broken.

PR #5171, opened 2019-05-09
Feat #63 accounts language email

### LINKED ISSUE #63: Responsive grid width / Long text formating
We should either make the grid elements responsive to the widest element ( which I think is preferable), or fix the formatting to take care of excessively long text elements.
reaction_demo

jm18457:
Resolves #63
Impact: **minor**  
Type: **feature**

## Issue
Currently all emails are sent based on language (field in mongodb) of the shop account belongs to. This means users cannot specify which language emails they want to receive. Similar for anonymous purchases.

## Solution
With this solution:
* Users can set their profile language
* All emails which are sent to accounts (not anonymous) are in language of their profile 
* Anonymous Orders can accept optional field language and then emails are sent based on this order email

## Overview
If user / order does not have specified account it still looks at shop language to find template.

* Added language field in Accounts.profile document
* Added language field in Order document
* Add graphQL query, mutation, resolvers to set and get language of account
* Modify getTemplate and getSubject to accept additional parameter language
* Modify functions where getTemplate and getSubject are used to get language for the email

## Breaking changes
There are no breaking changes.

## Testing
TBD

### DIFF
--- a/imports/collections/schemas/orders.js
+++ b/imports/collections/schemas/orders.js
@@ -458,7 +458,7 @@ export const Order = new SimpleSchema({
   "history.$": History,
   "language": {
     type: String,
-    defaultValue: "test"
+    optional: true
   },
   "notes": {
     type: Array,
--- a/imports/collections/schemas/orders.js
+++ b/imports/collections/schemas/orders.js
@@ -205,10 +205,6 @@ export const OrderItem = new SimpleSchema({
   "history.$": {
     type: History
   },
-  "language": {
-    type: String,
-    optional: true
-  },
   "optionTitle": {
     type: String,
     optional: true
@@ -460,6 +456,10 @@ export const Order = new SimpleSchema({
     optional: true
   },
   "history.$": History,
+  "language": {
+    type: String,
+    defaultValue: "test"
+  },
   "notes": {
     type: Array,
     optional: true
--- a/imports/collections/schemas/orders.js
+++ b/imports/collections/schemas/orders.js
@@ -205,6 +205,10 @@ export const OrderItem = new SimpleSchema({
   "history.$": {
     type: History
   },
+  "language": {
+    type: String,
+    optional: true
+  },
   "optionTitle": {
     type: String,
     optional: true
--- a/imports/plugins/core/accounts/server/methods/setProfileLanguage.js
+++ b/imports/plugins/core/accounts/server/methods/setProfileLanguage.js
@@ -1,4 +1,5 @@
 import { check, Match } from "meteor/check";
+import R from "ramda";
 import { Accounts, Shops } from "/lib/collections";
 import appEvents from "/imports/node-app/core/util/appEvents";
 import Reaction from "/imports/plugins/core/core/server/Reaction";
@@ -28,14 +29,12 @@ export default function setProfileLanguage(languageCode, accountId) {
     throw new ReactionError("access-denied", "Access denied");
   }
 
-  // Make sure this language code is in the related shop languages list
   const primaryShopLanguages = Reaction.getPrimaryShopLanguages();
 
-  // try to find language in shop
-  const language = primaryShopLanguages.find((shopLanguage) => shopLanguage.i18n === languageCode && shopLanguage.enabled);
+  const primaryShopLanguage = R.find((shopLanguage) => shopLanguage.i18n === languageCode && shopLanguage.enabled)(primaryShopLanguages)
 
-  if (!language) {
-    throw new ReactionError("invalid-argument", `The shop for this account does not define any language with code "${languageCode}"`);
+  if (!primaryShopLanguage) {
+    throw new ReactionError("invalid-argument", `Primary shop does not define any enabled language with code "${languageCode}"`);
   }
 
   Accounts.update({ userId }, { $set: { "profile.language": languageCode } });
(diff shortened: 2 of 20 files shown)


imports/plugins/core/orders/server/no-meteor/schemas/schema.graphql
```diff
@@ -437,6 +439,8 @@ input OrderInput {
   order, and those shop IDs may or may not match this one.
   """
   shopId: String!
+
+  language: String
```
aldeed: All fields must have a description and should be alphabetized

imports/plugins/core/orders/server/no-meteor/schemas/schema.graphql
```diff
@@ -315,6 +315,8 @@ type Order implements Node {
 
   "The date and time at which this order was last updated"
   updatedAt: DateTime!
+
+  language: String
```
aldeed: All fields must have a description and should be alphabetized

imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
```diff
@@ -1,4 +1,5 @@
 import SimpleSchema from "simpl-schema";
+import R from "ramda";
```
aldeed: Single character variables are not allowed by Reaction style rules. I'd recommend just using the standard array `.find` in this case, though.

imports/plugins/core/orders/server/no-meteor/mutations/placeOrder.js
```diff
@@ -170,8 +171,25 @@ export default async function placeOrder(context, input) {
...
+    // if non-primary shop does not have any languages use primary shop
+    shop = await Shops.findOne({ shopType: "primary" }, { languages: 1 });
+  }
+
+  const shopLanguages = (shop && shop.languages) || [];
+  const shopLanguage = R.find(R.whereEq({ enabled: true, i18n: language }))(shopLanguages);
+
+  // set to undefined value so order language will not be set
+  if (!shopLanguage) {
+    language = undefined;
```
aldeed: Throwing a helpful error would be better so they aren't confused why the language they specified isn't set.

imports/plugins/core/accounts/server/methods/setProfileLanguage.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * @name accounts/setProfileLanguage
+ * @memberof Accounts/Methods
+ * @method
+ * @param {String} languageCode - i18n language code
+ * @param {String} [accountId] - accountId of user to set language of. Defaults to current user ID
+ * @summary Sets users profile language
+ * @returns {Object} Account document
+ */
+export default function setProfileLanguage(languageCode, accountId) {
```
aldeed: Move this to a non-Meteor mutation function in `/imports/plugins/core/accounts/server/no-meteor/mutations/setProfileLanguage.js` and call it as `const updatedAccount = await context.mutations.setProfileLanguage(context, languageCode, dbAccountId);` in the resolver. Using `callMeteorMethod` is deprecated and soon going away.

imports/collections/schemas/orders.js
```diff
@@ -456,6 +456,10 @@ export const Order = new SimpleSchema({
     optional: true
   },
   "history.$": History,
+  "language": {
```
aldeed: I think this should be more specifically named, maybe `ordererPreferredLanguage`?

janus-reith:
Sending the current language per order seems to be a resonable solution to me.

In a typical scenario I'd expect the storefront to offer multiple languages - The currently used language would then be part of the order, regardless if it was placed anonymously or by an existing account.

Storing a language field on the account makes sense to me aswell, although I wouldn't treat that as specific for emails, but also to set defaults for whatever i18n provider is used in the frontend, if that account language might differ from the one requested by the browser.

First I was unsure about the part where the account language is used to determine the email language, as we can send it only with the order, regardless if we have an account or anonymous users.
But after rethinking this, it probably makes sense to use the account. A user might place an order in one language, later change the language and expect follow ups to that order like a shipping notice to also use that new language setting.

The last things missing for me:

- Where would an account get its inital language from?

We could make the storefront determine the language and send it over to the idp register form as part of the url query. The Login IDP shouldn't check language on its own, as the user might have set a different language in the storefront than the browser default.

 - How would the currently used storefront language affect the account language?

One option would be to just to give the user the choice like a dropdown for the current display language. Now every time the user selects a language here that differs from the account language a dialog could be shown asking if the default should be changed. The profile page could still have a seperate language preference field regardless of that, or do the same thing vice versa and ask if the current display language should be changed accordingly to avoid any confusion for users.

aldeed:
@janus-reith Reaction 1.x had such a language select which would set `profile.language` on Account. So the field is technically already supported, but it hasn't been ported over to GraphQL. If this PR is merged, setting the account language would once again be possible, so you could add back a language select to your storefront UI.

It would be nice to eventually tie it in with IDP and new account registration, but that can be done later, and it's a separate concern since the goal is to eventually make IDP+auth a third-party component.

aldeed:
@jm18457 Will you have time to address the comments?

jm18457:
> @jm18457 Will you have time to address the comments?

I will be busy for a certain period of time right now, but I will see if I can address them over the weekend or next weekend.

janus-reith:
It would probably be easier to redo this PR from scratch with the new meteor-free `sendEmail` function on context that is used now.

jm18457:
Hello, I will close this pull request and create a new pull request, since updated changes are in different repository.
I will also create a new issue in this repository with type feature so they are linked.

Issue #5144, opened 2019-04-24
Add URL fields to shop settings


aldeed:
## Background
This issue must be done before the following email issues can be resolved:
- https://github.com/reactioncommerce/reaction/issues/5115
- https://github.com/reactioncommerce/reaction/issues/5114
- https://github.com/reactioncommerce/reaction/issues/5113
- https://github.com/reactioncommerce/reaction/issues/4679

The common theme in these issues is that the API does not know what the storefront URLs are, so it can't use proper URLs when generating emails or redirecting after password resets, etc.

## Work
Add the following fields to `Shop` schema:
- `storefrontHomeUrl`
- `storefrontOrderUrl` (can have `:orderReferenceId` and `:orderToken` in the string)
- `storefrontOrdersUrl` (can have `:accountId` in the string)
- `storefrontAccountProfileUrl` (can have `:accountId` in the string)

Alternatively they could be grouped in a `storefrontUrls` object on `shop`. They are all `String` type and optional with no default value.

Then add a new GraphQL mutation, `updateShopUrls` that allows admins to update these new fields.

Finally, add a Storefront URLs card on the shop settings UI page, with one text field for each of these.

The ability to set and change these is the outcome of this work. Further updates to use the values as necessary will be done separately after this is merged.

aldeed:
@rymorgan Any UI design comments? It's just 4 new text fields. I'm thinking on their own card? Ideally some help text or placeholder about what variables are recognized in each.

rymorgan:
@aldeed Yes please it's own card. These fields are overrides URLs that will be used by our transactional emails? Will they be used for anything else? Do they have to be filled in for the transactional emails work? I'm not clear on what they are used for by an operator. But my main request is that they are grouped logically and that their usage is spelled out for the user. I'm happy to help with that but I don't know enough yet. So clear logical title for the card. Clear labels on the forms that help a user know what these fields are for.

kieckhafer:
This is going to live on the Shops setting page. 

Unfotunetely this uses some very old Blaze code to provide all these settings. I was just going to add a new accordion panel here. I do not think at this time I have the time to refactor this whole screen into the new format:

![Not-found](https://user-images.githubusercontent.com/4482263/56701251-ee50df00-66b2-11e9-9a93-ead9a864a4fc.png)

kieckhafer:
@rymorgan after our discussion, I moved this to the email card. 

[...]

PR #5174, opened 2019-05-09
feat: UI for adding Storefront Urls inside Catalyst

### LINKED ISSUE #5150: Add URL fields to shop settings - UI
Creating this ticket to do the UI work for #5144, as we need to merge the backend work for 5144 ASAP, and the UI seems to need a little discussion:

-----
From the original ticket, #5144:

## Background
This issue must be done before the following email issues can be resolved:
- https://github.com/reactioncommerce/reaction/issues/5115
- https://github.com/reactioncommerce/reaction/issues/5114
- https://github.com/reactioncommerce/reaction/issues/5113
- https://github.com/reactioncommerce/reaction/issues/4679

The common theme in these issues is that the API does not know what the storefront URLs are, so it can't use proper URLs when generating emails or redirecting after password resets, etc.

## Work
Add the following fields to `Shop` schema:
- `storefrontHomeUrl`
- `storefrontOrderUrl` (can have `:orderReferenceId` and `:orderToken` in the string)
- `storefrontOrdersUrl` (can have `:accountId` in the string)
- `storefrontAccountProfileUrl` (can have `:accountId` in the string)

Alternatively they could be grouped in a `storefrontUrls` object on `shop`. They are all `String` type and optional with no default value.

Then add a new GraphQL mutation, `updateShopUrls` that allows admins to update these new fields.

Finally, add a Storefront URLs card on the shop settings UI page, with one text field for each of these.

The ability to set and change these is the outcome of this work. Further updates to use the values as necessary will be done separately after this is merged.

kieckhafer:
Resolves #5150   
Impact: **minor**  
Type: **feature**

## Issue
The UI portion of #5144 was not completed at the time 5144 was done.

## Solution
This PR adds the UI for #5144.

Also added a `withPrimaryShop` HOC, which provides shop information. Currently, only fields I needed for this PR are provided, they should probably be added as needed going forward.

## Breaking changes
None

## Testing
1. Navigate to /operator/shop-settings
1. Try changing the various fields
1. Refresh the page to make sure they have saved
1. Go into the Shops collection of mongo, and look at the `storefrontUrls` object, and make sure they have saved

Not-found


kieckhafer:
@mikemurray this is ready-ish to go, with a caveat: It's a new MUI Card, inside of a Blaze template. That means some style issues are existing. What's the best way to remedy this? Is it to add a Theme provider here? And if so, where exactly does that get done in the Blaze component?

PR #5183, opened 2019-05-20
fix: #5182 add subtotal display amount to dataForOrderEmail

### LINKED ISSUE #5182: bug: getDataForOrderEmail should also include subtotal displayAmount
Custom template emails should be able to use the subtotal amount instead of just the item price.

## Solution

Add `subtotal.displayAmount` to the object returned in the `getDataForOrderEmail` method

machikoyasuda:
Resolves #5182  
Impact: **minor**  
Type: **feature|bugfix**

## Issue
The Order Email's `getDataForOrderEmail` method's return object should also return `subtotal`'s displayAmount.

## Solution
Adds displayAmount to subtotal
Updates tests

## Testing
1. Run the test
1. Use `subtotal.displayAmount` in an email

### DIFF
--- a/imports/plugins/core/orders/server/no-meteor/util/getDataForOrderEmail.js
+++ b/imports/plugins/core/orders/server/no-meteor/util/getDataForOrderEmail.js
@@ -106,6 +106,11 @@ export default async function getDataForOrderEmail(context, { order }) {
         // Add displayAmount to match user currency settings
         displayAmount: formatMoney(item.price.amount * userCurrencyExchangeRate, userCurrency)
       },
+      subtotal: {
+        ...item.subtotal,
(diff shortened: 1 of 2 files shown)


machikoyasuda:
@aldeed Any idea why these specs are failing? The error message is hard to decipher:  https://circleci.com/gh/reactioncommerce/reaction/43994

aldeed:
@machikoyasuda In the same way you added `displayAmount: jasmine.any(String)` in `combinedItems`, you also need to add it for the `subtotal` object in `order.shipping.items` further down. I think that's the only actual difference shown in the diff. (Annoyingly it shows the Anys as differences so you need to ignore those.)

PR #5186, opened 2019-05-22
Add a setting for whether options without inventory enabled are sellable

### LINKED ISSUE #5185: Add a setting for whether options without inventory enabled should be sellable
## Issue
The core `inventory` package, if no inventory plugin has inventory for an option, assumes that the option should be sellable. Essentially infinite inventory available.

This is a correct assumption in most cases because the core product should be useable out of the box without inventory tracking configured. However, in some complex inventory syncing situations, it is best to consider such options "out of stock" instead.

## Solution
In the core `inventory` package, add a boolean shop-specific setting `isVariantWithoutInventoryEnabledSoldOut`, which is `false` by default and has a toggle in the UI to change it.

aldeed:
Resolves #5185   
Impact: **minor**  
Type: **feature**

## Issue
The core `inventory` package, if no inventory plugin has inventory for an option, assumes that the option should be sellable. It is not shown as sold out, and although inventory available to sell is 0, `canBackorder` is set to `true` to allow selling always.

This is a correct default assumption because the core product should be useable out of the box without inventory tracking configured. However, in some inventory syncing situations, it is best to consider such options "out of stock" instead.

## Solution
There is a new shop-specific app setting, `canSellVariantWithoutInventory`, which is `true` by default. If this is changed to `false`, then variants/options without inventory tracking enabled will appear sold out and with back-ordering disabled.

## Other Changes
The StorefrontUrls React component had `ThemeProvider` and `MuiThemeProvider` temporarily added due to being used in `React` Blaze helper. This is removed in favor of putting the wrappers in the `React` helper itself, which is how other providers have been added in the past. This way each individual component doesn't have to solve the same issue.

A new Blocks region, "ShopSettings", is added to the bottom of the Shop Settings page. This allows React components to be added to that page with the newer and better API.

Rather than make custom GraphQL queries and mutations for this setting, I took the opportunity to implement a generic app settings pattern we have recently discussed. This is all implemented in a new `core/settings` plugin.
  - Any plugin can add any shop-specific setting by defining it in the `registerPlugin` call, using the `shopSettingsConfig` key, and extending `ShopSettings` and `ShopSettingsUpdates` in the GraphQL schema. (See core inventory plugin for example.)
  - Any plugin can add any global setting by defining it in the `registerPlugin` call, using the `globalSettingsConfig` key, and extending `GlobalSettings` and `GlobalSettingsUpdates` in the GraphQL schema. Nothing uses global settings yet.
  - GraphQL mutations `updateShopSettings` and `updateGlobalSettings` are used to set one or more registered settings
  - GraphQL queries `shopSettings` and `globalSettings` are used to get one or more registered settings

## Breaking changes
None

## Testing
1. Identify a sellable product variant that does not have inventory tracking enabled (or disable).
1. On Shop Settings page, ensure the check box on the Inventory card is enabled.
1. On the storefront, ensure that you can add any quantity of the identified variant to your cart and order it.
1. On Shop Settings page, ensure the check box on the Inventory card is disabled.
1. Refresh the storefront. The variant should now appear as sold out. Ensure that you cannot add any quantity of the identified variant to your cart, and if you already have in your cart, that you cannot order it.

### DIFF
--- a/imports/plugins/core/dashboard/client/components/StorefrontUrls.js
+++ b/imports/plugins/core/dashboard/client/components/StorefrontUrls.js
@@ -1,7 +1,7 @@
 import React, { Component, Fragment } from "react";
 import PropTypes from "prop-types";
 import gql from "graphql-tag";
-import styled from "styled-components"; // TODO: remove ThemeProvder when Blaze wrapper is removed
+import styled from "styled-components";
 import { Form } from "reacto-form";
 import { Mutation } from "react-apollo";
 import Button from "@material-ui/core/Button";
--- a/imports/plugins/core/dashboard/client/components/StorefrontUrls.js
+++ b/imports/plugins/core/dashboard/client/components/StorefrontUrls.js
@@ -1,8 +1,7 @@
 import React, { Component, Fragment } from "react";
 import PropTypes from "prop-types";
 import gql from "graphql-tag";
-import styled, { ThemeProvider } from "styled-components"; // TODO: remove ThemeProvder when Blaze wrapper is removed
-import MuiThemeProvider from "@material-ui/core/styles/MuiThemeProvider"; // TODO: remove when Blaze wrapper is removed
+import styled from "styled-components"; // TODO: remove ThemeProvder when Blaze wrapper is removed
 import { Form } from "reacto-form";
 import { Mutation } from "react-apollo";
 import Button from "@material-ui/core/Button";
@@ -16,9 +15,6 @@ import Field from "@reactioncommerce/components/Field/v1";
 import TextInput from "@reactioncommerce/components/TextInput/v1";
 import { i18next } from "/client/api";
 import withPrimaryShop from "/imports/plugins/core/graphql/lib/hocs/withPrimaryShop";
-import theme from "imports/plugins/core/router/client/theme"; // TODO: remove when Blaze wrapper is removed
-import muiTheme from "imports/plugins/core/router/client/theme/muiTheme"; // TODO: remove when Blaze wrapper is removed
-
 
 const PaddedField = styled(Field)`
   margin-bottom: 30px;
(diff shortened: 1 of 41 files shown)


willopez:
getting the following error: 
```
Message: Cannot query field "lowInventoryWarningThreshold" on type "CatalogProduct"., Location: [{"line":22,"column":13}], Path: undefined
```
I believe `lowInventoryWarningTheshold` is missing in the `CatalogProduct` query, and will also be necessary to add `inventoryReserved`.

imports/plugins/core/dashboard/client/components/StorefrontUrls.js
```diff
@@ -1,8 +1,7 @@
 import React, { Component, Fragment } from "react";
 import PropTypes from "prop-types";
 import gql from "graphql-tag";
-import styled, { ThemeProvider } from "styled-components"; // TODO: remove ThemeProvder when Blaze wrapper is removed
-import MuiThemeProvider from "@material-ui/core/styles/MuiThemeProvider"; // TODO: remove when Blaze wrapper is removed
+import styled from "styled-components"; // TODO: remove ThemeProvder when Blaze wrapper is removed
```
willopez: this TODO can be removed now?

willopez:
After, going thru steps 1-4 for a particular product and going back to step 1-2, the product seems to be stuck with an value of `inventoryInStock = 0`. I had expected it to be reset to the initial value, in my case 14 before toggling the inventory switch. If this is by design, it should be written in the description of the toggle switch.

aldeed:
@willopez I can't reproduce or maybe I'm not understanding. What exact steps will reproduce? If I am editing an option, enable inventory, enter 14 in stock, then disable inventory, it disabled the fields and shows 0. Then I re-enable and it enables the fields and shows 14 once again. Toggle back and forth goes 0, 14, etc. Do I have to do something additional to trigger it getting stuck at 0?

aldeed:
Or is it correct on the admin UI but stuck at 0 in the catalog?

aldeed:
I just tested everything again and fixed one bug, but it isn't related to the issue @willopez is seeing. Need to pair with him to see what's going on.

willopez:
I was misunderstanding the meaning of the inventory switch being off. When it's off, variants without inventory tracking will be set as sold out.

Issue #5163, opened 2019-05-02
Navigation editor UI tweaks


mikemurray:
## Issue Description

The **Edit Navigation Item** modal may hide the action buttons on some smaller screen sizes.

![image](https://user-images.githubusercontent.com/42217/57094554-aa1e9980-6cc5-11e9-81d8-f5c324c654b4.png)

**Saving/publishing** navigation items give no indication of success or failure. It's also impossible to tell if there are any unpublished changes.

![Not-found](https://user-images.githubusercontent.com/42217/57094814-43e64680-6cc6-11e9-98ac-fab4d0c8ef91.png)

## Possible Solution

- Edit dialog should be split into header/content/actions with the content region being the only scrollable one
- Add "Has unpublished changes" next to the discard button at the very least
- Have the `Save changes` buttons indicate that it is `[spinner] Saving changes` when clicked
- Display an alert toast when there are errors
- Display an indication of published status on the navigation tree items. (This may not be easily achieved)

spencern:
@mikemurray can we close this now?

mikemurray:
No. This is additional stuff not tackled in the other navigation PR.

PR #5190, opened 2019-05-24
feat: Update default navigation item visibility options when added to the navigation tree

### LINKED ISSUE #5189: Adjust navigation item visibility options default behavior
## Issue Description

When dragging navigation items into the tree and saving changes, the navigation item is instantly visible to all users since they all default to `isVisible: true`.

## Possible Solution

- default "Show in storefront" to `false` to prevent admins from accidentally publishing nav items
- maybe: add the options to change the default behavior back to `Show in storefront = true` if it becomes cumbersome.
- maybe: when dragging a nav item into the tree for the first time the edit modal, or a simpler dialog could ask the user if they want to show that item in the storefront.

mikemurray:
Resolves #5189   
Impact: **major**  
Type: **feature**

## Issue

Navigation items a visible by default when placed in a navigation tree. This could lead to admins publishing content not yet ready for customers.

## Solution

- Default the option "Show in storefront" to `false`
- Add shop settings to override the default behavior (API only)

## Breaking changes

- More like a breaking workflow. Every nav item added to the nav tree will need to be explicitly made visible on the storefront.
- Renamed plugin in register.js from "navigation" to "reaction-navigation". Added a migration.
- Removed "registry" and "layout" config from plugin registration

## Testing
1. Add a new navigation item
2. Drag it into the tree
3. Click "Save changes"
4. See the navigation item update and say "Hide from storefront"
5. Edit the item
6. check "Show in storefront"
7. Save changes in the modal
8. Save changes into the toolbar at the top
9. Refresh the page and make sure the item doesn't say "Hide from storefront"

(all other UI hiccups will be addressed in this ticket https://github.com/reactioncommerce/reaction/issues/5163)


mikemurray:
@kieckhafer Review code / test functionality
@aldeed Review code. I used the new settings API and the new plugin registration method.

imports/plugins/core/navigation/client/hocs/withNavigationUIStore.js
```diff
@@ -40,12 +49,18 @@ export default (Component) => (
     }
 
     navigationTreeToSortable(navigationTree) {
+      const {
+        shouldNavigationTreeItemsBeAdminOnly,
+        shouldNavigationTreeItemsBePublicallyVisible,
```
aldeed: Publically > Publicly

imports/plugins/core/navigation/server/no-meteor/schemas/settings.graphql
```diff
@@ -0,0 +1,40 @@
+extend type ShopSettings {
+  """
+  Whether a navigation item added to the navigation tree be visible only to
```
aldeed: be --> should be

imports/plugins/core/navigation/server/no-meteor/schemas/settings.graphql
```diff
@@ -0,0 +1,40 @@
...
+  admins by default
+  """
+  shouldNavigationTreeItemsBeAdminOnly: Boolean!
+
+  """
+  Whether a navigation item added to the navigation tree be
```
aldeed: be --> should be

imports/plugins/core/navigation/server/no-meteor/schemas/settings.graphql
```diff
@@ -0,0 +1,40 @@
...
+  public API/Storefront visible by default
+  """
+  shouldNavigationTreeItemsBePublicallyVisible: Boolean!
+
+  """
+  Whether a navigation item added to the navigation tree should show be a secondary
```
aldeed: should show be --> should be

imports/plugins/core/navigation/server/no-meteor/schemas/settings.graphql
```diff
@@ -0,0 +1,40 @@
...
+  navigation by default
+  """
+  shouldNavigationTreeItemsBeSecondaryNavOnly: Boolean!
+
+}
+
+extend input ShopSettingsUpdates {
```
aldeed: Same typo comments on these three descriptions

imports/plugins/core/navigation/server/no-meteor/util/setDefaultsForNavigationTreeItems.js
```diff
@@ -1,10 +1,28 @@
+// Fallback options setting every value to false.
+// This effectively means the item will not be published to the public API.
+const standardDefaults = {
+  shouldNavigationTreeItemsBeAdminOnly: false,
+  shouldNavigationTreeItemsBeSecondaryNavOnly: false,
+  shouldNavigationTreeItemsBePublicallyVisible: false
+};
```
aldeed: Maybe you could just make `defaultValues` required and remove these standard fallbacks? Just seems like too many layers. There should always be default values returned from the server because you specified default values in registerPlugin.

mikemurray:
@aldeed Yea, I think there should be a UI to change it in the future. Probably inline somewhere in the navigation editor, but I don't think we have a design for something like that.

cc @reactioncommerce/design

imports/plugins/core/navigation/server/no-meteor/util/setDefaultsForNavigationTreeItems.js
mikemurray: I'll make it required, add validation and throw an error if they're not defined.

rymorgan:
@mikemurray Can I see a demo? So I know where we need to go with this?

mikemurray:
@rymorgan At the moment there's nothing really new to see in the UI, just the fact that once you drag navigation items into the tree they're set to be hidden from the storefront by default, as well as some other options.

The 3 settings I have now determine whether new items are visible to users, visible to admins only, or are secondary nav items when they are added to the navigation tree. These settings could be considered default publish status for new nav items.

Where they live may be determined by how easily we want these settings changed, or what responsibility we want to put on the plugin to maintain it's own settings interface. It could be as simple as a switch in the page that's easily toggled. Or a settings dropdown/modal that shows up when you click a settings button in the toolbar. Or, maybe settings like these don't need quick access and could be put in a generic "Settings" section in the app.

mikemurray:
@kieckhafer @aldeed ready for re-review

imports/plugins/core/versions/server/migrations/65_update_navigation_tree.js
```diff
@@ -0,0 +1,43 @@
...
+const { NavigationTrees } = rawCollections;
+
+Migrations.add({
+  version: 65,
+  async up() {
+    // Find all navigation trees
+    const navigationTrees = await NavigationTrees.find({}).toArray();
+
+    // Update default values for each tree
+    Promise.all(navigationTrees.map((tree) => {
```
aldeed: Since this runs in a Meteor fibers context, you should have `Promise.await` around `Promise.all`. Otherwise it will think the migration is done before it's actually done.

```js
Promise.await(Promise.all(navigationTrees.map((tree) => {
```

aldeed: And then I think you'd use a normal try/catch around that if you want to catch and log the error.

mikemurray:
@aldeed Fixed. Waiting for tests to pass.

PR #5195, opened 2019-05-31
[Prototype] Add Storybook support and some examples

mikemurray:
Impact: **minor**  
Type: **docs**

## Issue

Documentation should live closer to home to make it easier to write docs while you code.

## Solution

Add Storybook with a few documentation examples.

This prototype has 3 different location for docs.

- ConfirmDialog as its story next to the implementation
image

- The register components package has a `docs/stories` folder as well as a `docs/developer` folder containing the stories, and markdown files.

image

- `/imports/stories`

image

## Breaking changes

none

## Testing
1. Run in docker `docker-compose up storybook` or...
2. Run outside doctor `npm install && npm run storybook`
3. Go to http://localhost:3100 if you're not automatically taken there (usually because of docker)

See something like this:

image

### DIFF
--- a/.storybook/addons.js
+++ b/.storybook/addons.js
@@ -1,0 +1,3 @@
+import '@storybook/addon-actions/register';
+import '@storybook/addon-links/register';
+import "@storybook/addon-knobs/register";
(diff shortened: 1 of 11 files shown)


machikoyasuda:
closing for now

PR #5196, opened 2019-06-03
feat: define collections using registerPlugin

### LINKED ISSUE #4268: Create a system that that allows you create indexes on Collections without modifying core
Users should be able to add indexes to collection based on their individual needs without having to modify core code. It would be great if this was something that could be checked into a repo.

aldeed:
Epic #4996
Resolves #4268   
Impact: **minor**  
Type: **feature**

## Issue
Most MongoDB collections are defined in a single place in core code, breaking the plugin pattern. Some newer collections are defined in plugin startup functions, but this pattern does not give good control over when the collections exist. Ideally they should already be available when all plugin startup functions run.

## Solution
Introduces a new `collections` option for `registerPlugin`, which allows plugins to define their MongoDB collections and indexes in a standard way, such that they'll exist already when plugin startup functions run.

## Breaking changes
None. Custom plugins may still define collections in a startup function, but you are encouraged to move to the new pattern.

## Additional changes
- The `TestApp` class was duplicating several blocks of code from `ReactionNodeApp` class. Now, `TestApp` wraps and calls through to a `ReactionNodeApp` instance.
- The collection changes caused timing failures in Meteor app tests. I added `Reaction.onAppStartupComplete` as a better way to stop all tests from running until after all collections are available and all startup code has run.
- In order for integration tests to continue working, I had to move all plugins with collections to the newer non-Meteor `registerPlugin` pattern.
- The collection pattern changes added to startup confusion around file-collections startup. To resolve this, I moved around some files plugin code and eliminated some code that was duplicated between non-Meteor and Meteor startup. None of the actual files code logic is changed; it's just called in better places and in a clearer order.

## Testing
- Smoke test the API and verify that everything continues to work.
- Smoke test the Catalyst UI and verify that everything continues to work.
- Smoke test the example storefront connected to the API in this branch and verify that everything continues to work.


aldeed:
Reviewer note: Unfortunately this was the type of change that cascaded into needing several changes that I couldn't pull apart. The diff looks worse than it is. All the register.js files are just copy/paste from the Meteor file into the non-Meteor file, plus add `collections` option. In many cases the `collections` were moved out of `startup.js`, so that means all of those files changed, too.

imports/plugins/core/accounts/server/no-meteor/register.js
```diff
@@ -0,0 +1,134 @@
...
+      groups: ["guest", "customer"],
+      roles: ["account/verify", "reset-password", ENROLL_URI_BASE]
+    }],
+    collections: {
+      Accounts: {
+        name: "Accounts",
+        indexes: [
+          // Create indexes. We set specific names for backwards compatibility
+          // with indexes created by the aldeed:schema-index Meteor package.
+          [{ groups: 1 }, { name: "c2_groups" }],
```
zenweasel: Is there no way to not use these old index names?

imports/plugins/core/discounts/server/no-meteor/register.js
```diff
@@ -0,0 +1,28 @@
+import startup from "./startup";
```
zenweasel: Just an aside that if I was to nominate a plugin to be removed this would def be one of them

imports/plugins/core/email/server/no-meteor/register.js
```diff
@@ -0,0 +1,62 @@
+import mutations from "./mutations";
```
zenweasel: I feel like moving these to `no-meteor/register` is confusing. Does moving it to this directory enable any different functionality or is it just a mnemonic?

imports/plugins/core/email/server/no-meteor/util/getShopLogo.js
```diff
@@ -8,6 +6,9 @@ import { Media } from "/imports/plugins/core/files/server";
  * @return {String} Absolute image URL
  */
 export default async function getShopLogo(context, shop) {
+  const { collections } = context;
```
zenweasel: Could these be combined into one line? I don't see us using `collections` anywhere else

imports/plugins/core/files/server/fileCollections.js
```diff
@@ -1,94 +1,80 @@
...
-// For backward-compatibility with code relying on how CFS did it, we'll put a
-// reference to the backing MongoDB collection on Media.files property as well.
-Media.files = MediaRecords;
+  const { stores, tempStore } = files;
+  if (!stores) throw new Error("Expected context.files.stores");
+  if (!tempStore) throw new Error("Expected context.files.tempStore");
 
-WebApp.connectHandlers.use("/assets/files", downloadManager.connectHandler);
+  WebApp.connectHandlers.use("/assets/uploads", (req, res) => {
+    req.baseUrl = "/assets/uploads"; // tus relies on this being set, which is an Express thing
```
zenweasel: `tus`?

imports/node-app/core/ReactionNodeApp.js
```diff
@@ -74,7 +79,32 @@ export default class ReactionNodeApp {
...
+              if (!collectionConfig || typeof collectionConfig.name !== "string" || collectionConfig.name.length === 0) {
+                throw new Error(`In registerPlugin, collection "${collectionKey}" needs a name property`);
+              }
+              if (this.collections[collectionKey]) {
+                throw new Error(`Plugin ${pluginName} defines a collection with key "${collectionKey}" in registerPlugin,` +
+                  " but another plugin has already defined a collection with that key");
+              }
+              this.collections[collectionKey] = this.db.collection(collectionConfig.name);
+              if (Array.isArray(collectionConfig.indexes)) {
+                for (const indexArgs of collectionConfig.indexes) {
```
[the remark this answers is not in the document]
aldeed: I don't remember why. I may have initially been awaiting the collectionIndex call?

imports/plugins/core/accounts/server/no-meteor/register.js
aldeed: If there's a way to change index names, I suppose we could do a migration. But without that, if you leave out the name and they have the old names already in the db, you get an error about trying to create the same index with a different name. There may be some way, but it would take experimentation.

imports/plugins/core/email/server/no-meteor/register.js
aldeed: It's mnemonic, but the new way of exporting a single `register` function that accepts `app` argument is important. It's a stepping stone toward NPM packages. Imagine the next step where everything in `no-meteor` moves to a separate package. Then the original, outer `register.js` file will `import register from "@reactioncommerce/plugin-email/register";` and pass it the `app` to connect it.

The step after that is removing the Meteor auto-register stuff and having people explicitly call the plugins in the order they wish, like is done in `registerPlugins.js`. That ultimately will be the pattern when the API is fully transitioned to pure Node app.

imports/plugins/core/email/server/no-meteor/util/getShopLogo.js
aldeed: Yes I do this by habit since we have other places where we need it.

imports/plugins/core/files/server/fileCollections.js
aldeed: https://tus.io/

It's what file-collections package uses for uploading.

zenweasel:
Not to clutter this up since this is already a big one, but having given it some more thought I am thinking we should migrate these indexes to a standard naming convention now.

It could also be a separate PR but I think if we don't do it now I feel like we will never do it.

I am not blocking this PR for that but it is my suggestion.

aldeed:
@zenweasel That makes sense to me. I'll look into it for either this PR or another.

aldeed:
@mikemurray After another merge from develop and a few random other fixes, I'm not seeing the error you mention.

mikemurray:
@aldeed 👍 I will run it again

Issue #5157, opened 2019-04-30
Refactor Catalyst Orders panel to work with multiple fulfillment methods


kieckhafer:
Catalyst's (Operator UI) Orders panel is out-of-date, and doesn't work as needed with our multiple fulfillment group updates (#5027, #5024, #5020, #5019, #5018, #5010).

To make the panel up-to-date, we need to do the following work:

- [ ] Display multiple fulfillment methods, and items inside the methods separately (Currently, we combine all items from all methods into a single group)

- [ ] Make sure the methods are displayed correctly (Currently, we display multiple method sections, but we just replicate a single method over and over, which is not correct)

- [ ] Hook up the new mutations to allow for order manipulation with the new methods

This list is not comprehensive, it is simply what was noticed at first pass through. This ticket will be updated with additional work when it's discovered.

kieckhafer:
@aldeed I have a few questions about how all the new mutations work:

1. we don't have a `cancelOrder` GQL mutation, we just have `cancelOrderItem`. Is this on purpose? Is the thought to cancel an order, we loop over all order items and cancel them individually?
1. it doesn't look like the `cancelOrderItem` mutation performs any refunds, or inventory updates. I remember you mentioning the inventory should be done independently, so maybe we need to call the inventoryUpdate separately after an item is cancelled? What about for refunding items that are cancelled? It looks like we might still need to create a mutation for that?
1. Language wise, is cancelling an item the same as refunding an item? As in, if I press the `cancel` button, should I automatically be refunded? If i press the `refund` button, should it automatically be cancelled? Or are these two separate actions that need two separate trigger buttons, one `cancel` and one `refund`?

jrphilo:
@kieckhafer Re: question #3, the difference between a cancellation and a refund is dependent on whether the item is pre- or post- shipping. Essentially, when a Customer places an order, a payment authorization is put in place, but no money is taken. Funds are only drawn against a Customers card when an item ships. So an item is "Cancelled" before it has shipped (usually because the item cannot be found in the warehouse). Once it has shipped, payment will have been taken, so any return of the item would now trigger a "Refund". Let me know if any of this is unclear or if you want to discuss any further.

rymorgan:
[...]

PR #5201, opened 2019-06-05
add approveOrderPayments graphql mutation

kieckhafer:
Part of #5157 
Impact: **minor**  
Type: **feature**

## Changes
A new `approveOrderPayments` mutation allows you to approve order payments, one by one, or multiple payments at one time. This is currently a required step in our workflow, and triggers a few changes with other packages, like inventory. Although we might remove this step in the future, this mutation can be used by existing users who use this step in the process.

## Breaking changes
None

## Testing
1. Place an order with one or more payments.
1. Verify that the mutation works as described, by updating payment status to `approved` for the corresponding payments.

### DIFF
--- a/imports/plugins/core/payments/server/no-meteor/mutations/__snapshots__/approveOrderPayments.test.js.snap
+++ b/imports/plugins/core/payments/server/no-meteor/mutations/__snapshots__/approveOrderPayments.test.js.snap
@@ -1,0 +1,13 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`throws if orderId isn't supplied 1`] = `"Order ID is required"`;
+
+exports[`throws if paymentIds isn't supplied 1`] = `"Payment ids is required"`;
+
+exports[`throws if permission check fails 1`] = `"Access denied"`;
+
+exports[`throws if shopId isn't supplied 1`] = `"Shop ID is required"`;
+
+exports[`throws if the order doesn't exist 1`] = `"Access denied"`;
+
+exports[`throws if the order payment doesn't exist 1`] = `"Cannot read property 'status' of undefined"`;
--- a/imports/plugins/core/payments/server/no-meteor/mutations/__snapshots__/approveOrderPayments.test.js.snap
+++ b/imports/plugins/core/payments/server/no-meteor/mutations/__snapshots__/approveOrderPayments.test.js.snap
@@ -8,6 +8,4 @@ exports[`throws if permission check fails 1`] = `"Access denied"`;
 
 exports[`throws if shopId isn't supplied 1`] = `"Shop ID is required"`;
 
-exports[`throws if the order doesn't exist 1`] = `"Access denied"`;
-
-exports[`throws if the order payment doesn't exist 1`] = `"Cannot read property 'status' of undefined"`;
+exports[`throws if the order doesn't exist 1`] = `"Order not found"`;
--- a/imports/plugins/core/payments/server/no-meteor/mutations/approveOrderPayments.test.js
+++ b/imports/plugins/core/payments/server/no-meteor/mutations/approveOrderPayments.test.js
@@ -92,16 +92,16 @@ test("updates an order status", async () => {
     shopId: "SHOP_ID"
   }));
 
+  mockContext.collections.Orders.findOneAndUpdate.mockReturnValueOnce(Promise.resolve({
+    modifiedCount: 1,
+    value: {}
+  }));
+
   await approveOrderPayments(mockContext, {
     orderId: "abc",
     paymentIds: ["1"],
     shopId: "SHOP_ID"
   });
-
-  mockContext.collections.Orders.findOneAndUpdate.mockReturnValueOnce(Promise.resolve({
-    modifiedCount: 1,
-    value: {}
-  }));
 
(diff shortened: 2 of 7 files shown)


imports/plugins/core/payments/server/no-meteor/mutations/approveOrderPayments.test.js
```diff
@@ -0,0 +1,120 @@
...
+
+  await approveOrderPayments(mockContext, {
+    orderId: "abc",
+    paymentIds: ["1"],
+    shopId: "SHOP_ID"
+  });
+
+  mockContext.collections.Orders.findOneAndUpdate.mockReturnValueOnce(Promise.resolve({
+    modifiedCount: 1,
+    value: {}
```
kieckhafer: @aldeed can you help me figure out why this test is failing?

I get the error:

```
TypeError: Cannot destructure property `value` of 'undefined' or 'null'.

      65 |
      66 |   return { order: updatedOrder };
    > 67 | }
         |  ^
      68 |

      at _approveOrderPayments (imports/plugins/core/payments/server/no-meteor/mutations/approveOrderPayments.js:67:2)
```

Which makes me think my mock isn't returning correctly? But if I'm just checkout what the `findOneAndUpdate` was called with, does that matter? I've followed the pattern from a lot of other tests here, but can't get it working.

aldeed: `findOneAndUpdate.mockReturnValueOnce` is after the `await approveOrderPayments` call, so it's mocked too late. Move it to right after the `findOne.mockReturnValueOnce`.

kieckhafer: So simple, thanks!

imports/plugins/core/payments/server/no-meteor/schemas/schema.graphql
```diff
@@ -68,6 +71,9 @@ type Payment implements Node {
   "The payment method"
   method: PaymentMethod!
 
+  "Risk level of payment"
+  riskLevel: String
```
aldeed: It looks like there are three possible values mentioned in the code: normal, elevated, highest. Maybe make this an enum value so clients know what to expect back here?

imports/plugins/core/payments/server/no-meteor/mutations/__snapshots__/approveOrderPayments.test.js.snap
```diff
@@ -0,0 +1,13 @@
...
+
+exports[`throws if orderId isn't supplied 1`] = `"Order ID is required"`;
+
+exports[`throws if paymentIds isn't supplied 1`] = `"Payment ids is required"`;
+
+exports[`throws if permission check fails 1`] = `"Access denied"`;
+
+exports[`throws if shopId isn't supplied 1`] = `"Shop ID is required"`;
+
+exports[`throws if the order doesn't exist 1`] = `"Access denied"`;
```
aldeed: This isn't the expected error. You might need to mock more things to be sure it gets past the permission check and hits the order exists check.

imports/plugins/core/payments/server/no-meteor/mutations/__snapshots__/approveOrderPayments.test.js.snap
```diff
@@ -0,0 +1,13 @@
...
+
+exports[`throws if the order payment doesn't exist 1`] = `"Cannot read property 'status' of undefined"`;
```
aldeed: This isn't the expected error.

kieckhafer: Updated

imports/plugins/core/payments/server/no-meteor/mutations/__snapshots__/approveOrderPayments.test.js.snap
kieckhafer: Updated

imports/plugins/core/payments/server/no-meteor/schemas/schema.graphql
kieckhafer: Updated

imports/plugins/core/payments/server/no-meteor/schemas/schema.graphql
```diff
@@ -105,6 +105,18 @@ enum PaymentStatus {
...
 
+"Valid payment risk levels"
+enum RiskLevels {
+  "A normal risk level for a payment"
+  normal
+
+  "An elevated risk level for a payment"
+  elevated
+
+  "The highest rist level for a payment"
```
aldeed: risk

imports/plugins/core/payments/server/no-meteor/schemas/schema.graphql
```diff
@@ -72,7 +72,7 @@ type Payment implements Node {
   method: PaymentMethod!
 
   "Risk level of payment"
-  riskLevel: String
+  riskLevel: RiskLevels
```
aldeed: It looks like we typically do singular for enums, so `RiskLevel`

kieckhafer:
what's the best way to do `graphql-lint` locally, so we can catch this before it's pushed?

aldeed:
@kieckhafer `npm run lint:gql` does it

PR #5027, opened 2019-03-08
Add addOrderFulfillmentGroup mutation

aldeed:
Part of #4999
Impact: **minor**  
Type: **feature**

## Changes
A new `addOrderFulfillmentGroup` mutation allows you to add a fulfillment group to an existing order. You can provide new items for it and/or move items that are in another existing group.

The operator UI does not yet implement this.

- `addOrderFulfillmentGroup` can be called internally to synchronize from an external system by setting `context.isInternalCall` to `true`
- `addOrderFulfillmentGroup` can be called through GraphQL by any user with "orders" permission for the shop that owns the order
- All tax, shipping, surcharge, and total calculations are rerun after adjusting the existing order groups and adding the new group.

## Breaking changes
None

## Testing
1. Place an order.
1. Verify that the mutation works as described.
1. Verify that you can add a new fulfillment group when authenticated as a user with "orders" role, but not when authenticated as someone else or when not authenticated.

### Specific Instructions
1. Place an order with multiple items, and note the reference (displayed) ID.
2. Go to http://localhost:3000 and log in as an admin with "orders" permission.
3. In the browser console enter `localStorage.getItem("Meteor.loginToken")`. Copy the printed token.
4. Go to http://localhost:3000/graphql-alpha in browser
5. Expand HTTP HEADERS box and add `"meteor-login-token": "JW6NV92VfuqlrhD1M1IbBovf2ELktkhGBPeMxW6y4bj"` (the token you copied) property
6. Use the `primaryShop` query to get the primary shop ID
7. Use the following query to get your order by its reference ID and the primary shop ID

```graphql
query {
  orderByReferenceId(id: "DEfA7uGPhri4rLvhZ", shopId: "cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg==") {
    _id
    totalItemQuantity
    fulfillmentGroups {
      totalItemQuantity
      items {
        nodes {
          _id
          title
          variantTitle
          price {
            amount
          }
          productConfiguration {
            productId
            productVariantId
          }
        }
      }
      selectedFulfillmentOption {
        fulfillmentMethod {
          _id
        }
      }
    }
  }
}
```

8. Then use this mutation to add another group with some items and move one item from the other group:

```graphql
mutation {
  addOrderFulfillmentGroup(
    input: {
      orderId: "cmVhY3Rpb24vb3JkZXI6Zk1ySHJpRnFQRGlod21yOEM=",
      fulfillmentGroup: {
        data: {
[...]

PR #5202, opened 2019-06-06
feat: add read-only version of orders 2.0 to break up PR's

kieckhafer:
Part of #5157   
Impact: **major**  
Type: **feature**

## Issue
The PR #5157, which refactors the Catalyst Orders panel to use all GraphQL, as well as a new design, was getting very big. This PR breaks off a read-only version of the UI, which will help break the review into smaller pieces, as well as meet the need for a read-only version of this to be available for some end to end testing we are trying to do, without waiting on the full functionality, which isn't needed for this testing.

## Breaking changes
None. The old Orders UI is still the primary UI, and you need a special URL to access the new Panel.

## Testing
1. Create an order, with multiple items and multiple payments, via the storefront
1. Break the Items into multiple fulfillment groups using the addOrderFulfillmentGroup mutation ( #5027 ) 
1. See that you can see the order in the "Classis" Orders UI by going to the orders page, clicking on the order from the Table.
1. Change the url from `/orders/`, to `/orders-2.0/`, to see the order in the new UI
1. See that all the data is correct, that the fulfillment groups are broken up

** please note translations for order statuses will not be added by default to a new shop. you can follow instructions here on how to add them, if you wish: https://github.com/reactioncommerce/reaction/pull/4981
Order_Details_for_order_reference__BpkeapiCwDpKYmJL5

### DIFF
--- a/imports/plugins/core/core/server/no-meteor/schemas/shop.graphql
+++ b/imports/plugins/core/core/server/no-meteor/schemas/shop.graphql
@@ -30,6 +30,9 @@ type Shop implements Node {
   "Returns a list of groups for this shop, as a Relay-compatible connection."
   groups(after: ConnectionCursor, before: ConnectionCursor, first: ConnectionLimitInt, last: ConnectionLimitInt, sortOrder: SortOrder = asc, sortBy: GroupSortByField = createdAt): GroupConnection
 
+  "Returns shop language"
+  language: String
+
   "Returns shop name"
   name: String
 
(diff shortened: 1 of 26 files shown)


kieckhafer:
@mikemurray and I were discussing that instead of just expanding this PR to add order update functionality, it might actually be a good idea to keep these `read-only` versions of the components, and use them for users with lesser permissions than a full admin.

If we decided to do that, would we want to keep these as separate components completely, and render a fully functional vs. `read-only` total component? Or would we want to do multiple `read-only` checks inside each component, and only have one component with the permissions that live inside it?

aldeed:
I reviewed and fixed a few things to finish this up for @kieckhafer while he's out. @kieckhafer You may want to look at my commits and make sure those changes are in your subsequent PR for this.

aldeed:
Regarding the question of how best to do read only, I think whatever makes the most sense when doing it would be fine. I would probably default to trying to do checks granularly but if it gets too complicated or messy, fall back to a higher-level entire view only page. Either way the main thing would be to share as much of the smaller specific card components as possible.

PR #5205, opened 2019-06-07
[WIP] Add Blocks API docs

mikemurray:
Impact: **minor**
Type: **docs**

## Issue

Bring docs closer to home

## Solution

Put docs right inside the package.

## Breaking changes

None.

## Testing
1. Read the docs here: https://github.com/reactioncommerce/reaction/tree/docs-mikemurray-blocks-api/imports/plugins/core/components

### DIFF
--- a/imports/plugins/core/components/README.md
+++ b/imports/plugins/core/components/README.md
@@ -1,0 +1,7 @@
+# Components Core
+
+This package provides utilities for view layout management in Reaction.
+
+## API
+
+- [Blocks API](docs/api/blocks.md)
--- a/imports/plugins/core/components/docs/api/blocks.md
+++ b/imports/plugins/core/components/docs/api/blocks.md
@@ -1,0 +1,442 @@
+---
+id: blocks-api
+title: Blocks API
+---
+
+The Blocks API is a method of component extension that allows for adding additional UI components into various regions across the admin interface. For example, you can add additional fields to the product editor to extend it's functionality without having to override the entire component or package.
+
+See [Full API docs](#api) below.
+
+## Using Reaction Blocks
+
+A small subset of blocks and regions have been registered and can be imported .....
+
+```jsx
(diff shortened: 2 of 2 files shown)


mikemurray:
@aldeed This PR is to get the ball rolling, and was part of the discussion from the last engineering meeting. I want this to be a place to experiment with the folder structure and stuff so others can see a practical example while the DR is finalized.

What I have now is based on the DR as best I could interpret it. I looked through the DR and didn't see anything mentioning a version folder structure so I didn't do that. Maybe I'm overlooking something in the document.

aldeed:
OK, yes I don't see versioning mentioned either. I remember discussing it, but maybe the branching/tagging is enough for versioning in the repos and the version folders will only some into play with the script that copies all of these docs into a single docs site. @kieckhafer @zenweasel anything to add here?

zenweasel:
Yes, our original intent was to have version as a directory structure layer. I still think that makes sense and makes the raw docs more browsable. Is there any reason **not** to do that?

aldeed:
I think we'd mainly just need to figure out what versioning docs looks like. For example, does releasing each new version mean we have to copy all docs into a new version folder? This is where I started thinking copying the way docusaurus does version folders made sense, because then at least we'd just defer to them on these decisions.

mikemurray:
Another issue is how will each plugin be versioned. If plugins can be versioned independently of the core reaction app, then it would be difficult to align all the docs for all the repos into a single docusaurus instance.

I almost feel like each plugin, especially if they're split from the main repo, should maintain its own docusaurus folder structure, as well as build its own docs into a website that's published. Then the main docs site can link to individual plugin docs.

We could then use the main docs repo for guides for developers/operators since they could span multiple plugins.

aldeed:
@mikemurray I agree that may be best if we can write some tools to make it all easy, preferably automatic on CI.

machikoyasuda:
closing for now - we're not doing this yet/too early!

PR #5212, opened 2019-06-13
Demeteorize and reorg some core startup code

### LINKED ISSUE #5210: Log out plugin name and completed message in plugin loader
Because of the purposefully synchronous nature of the plugin loader it's quite easy for a plugin to block all other plugins from loading if there is an error in the startup function. Right now it's difficult to know if a plugin is doing this as there is no logging that displays when a function has completed startup and when startup for all plugins is complete.

The solution would be to modify the variable here: https://github.com/sportsdirect/reaction/blob/b16d5800ac188152df239393d37158bea3ba81f1/imports/node-app/core/ReactionNodeApp.js#L120

to contain both the function and the name of the plugin so that you can log that each plugin has completed, and then to add logging that ALL plugins have completed loading.

### LINKED ISSUE #5219: Storefront PDP does not work on first load if Example Product hasn't been republished
## Issue Description
When loading up the Storefront PDP for the first time, an error is shown:
undefined___Reaction

This seems to stem from the Example Product not having the correct pricing data in the `Catalog` collection.

Once the product is published ( / re-published) via Catalyst, the error is no longer seen.

## Steps to Reproduce
- Start a fresh install of Reaction
- Go to the Storefront
- Click a PDP

## Possible Solution
Add price data to Example Product

aldeed:
Resolves #5210 
Resolves #5219
Impact: **minor**  
Type: **feature**

Part of epic https://github.com/reactioncommerce/reaction/issues/4996

## Changes
- Removed `Reaction.init` function and moved that code to where it was called in the main startup function
- Removed support for undocumented feature `Meteor.settings.cdnPrefix`
- Removed support for undocumented feature `Reaction.Endpoints`
- Removed support for prerender by `prerender-node` NPM package
- Updated `handlebars` package to 4.1.2 to fix snyk warning
- Moved job server startup to `imports/utils/jobs.js` (still Meteor)
- Moved accounts/groups startup code to `accounts` core plugin and de-Meteorize
- Update and de-Meteorize sample data import code. On an empty database at startup, a primary shop, a sample product, and related sample tags are still inserted as before. The `Shipping` collection no longer has sample shipping methods added. The data itself is moved from `.json` files in `/private/data` (a Meteor thing) to a single JS file `/imports/plugins/core/core/server/no-meteor/util/sampleData.js`
- Reorg the execution order of startup code based on dependency needs. Run migrations later to be sure everything they need is there.
- Partially de-Meteorize import into the `Translations` collection on startup. Some translation files are still loaded in Meteor into the `Assets` collection, but moving from `Assets` to per-shop documents in `Translations` is now done without Meteor, in an "afterShopCreate" listener.
- Set log level to "ERROR" when running Jest tests to keep the output clean

## Breaking changes
- Removed support for undocumented feature `Meteor.settings.cdnPrefix`
- Removed support for undocumented feature `Reaction.Endpoints`
- Removed support for prerender by `prerender-node` NPM package

## Testing
1. Start the app with a fresh database. Verify no errors in the startup logging.
2. Verify sample data was inserted.
3. Verify that Groups and user roles are inserted/updated for the shop as before.
4. Smoke test the operator UI
5. Configure emailing and verify that emails are sent (because job server code was moved)
6. Also try starting up against a non-empty database, such as a sample data set. Verify no errors in startup logging.


mikemurray:
**More info about tags being missing from the admin.**

The default tags show up on the product, but do not show up in the table in the `Tags` admin. If I add new tags to the product, then the new tags will show in the table.

I seem the difference between the default tags and new tags is, the default tags are missing 

- `isDeleted: false`
- `groups: []`.

I added `isDeleted: false` to one of the default tags and it now shows up in the tag admin table.

zenweasel:
I wonder if this is happening because the old Meteor code when through SimpleSchema by default and so got all the default values (like `isDeleted`) set automatically. When we load these fixtures we should probably be manually running them through SimpleSchema?

mikemurray:
@zenweasel it's definitely that, but I think @aldeed wants to avoid using SimpleSchema to fix the data, and make sure the sample data is correct. The solution here would be to add `isDeleted: false` to the sample data for tags.

zenweasel:
@mikemurray I think that's a mistake. People use this fixture loader to load their own data and this way we are encouraging them to insert data directly into the system that's not been validated.

aldeed:
@zenweasel The refactored code does not use the fixture loader. I didn't de-Meteorize it, and I wasn't planning to because I wasn't planning to support loading JSON data in this way in the pure Node API.

I believe the Importer does still clean the data and add default values for now, but it won't be available in the final refactored API app.

The reason I'm avoiding supporting `defaultValue` in the schema is because it is one of those magical things that causes confusion. It's much easier to follow what's going on when all values are actually in the object being inserted.

aldeed:
Also @mikemurray, thanks. I'll look into the issues you mentioned.

aldeed:
@mikemurray So far I've fixed the tags and nav. Migrations no longer run on a fresh DB, too. I'm seeing email templates fine, so maybe something I did fixed that issue, or maybe it's only happening for you?

I also fixed #5219

I do see the owner group issue and will try to fix that next.

aldeed:
I am seeing the missing email templates issue randomly so I'll investigate that.

aldeed:
Assigning @kieckhafer for final review and testing since @mikemurray is out for a while.

@kieckhafer All of [Mike's issues](https://github.com/reactioncommerce/reaction/pull/5212#pullrequestreview-250668044) are resolved now. I had to refactor account creation / email verification a bit, so please test those flows in addition to the things Mike mentions. These are all related to startup on a fresh database, but obviously make sure it starts up fine and runs migrations as necessary when there's already data, too.

imports/plugins/core/i18n/client/components/localizationSettings.js
```diff
@@ -152,8 +152,8 @@ class LocalizationSettings extends Component {
         {name === "language" && "|"}
         {name === "language" &&
           <Components.FlatButton
-            i18nKeyTooltip={"admin.i18nSettings.reloadTranslations"}
-            tooltip={"Reload translations asdasdasdasdasd"}
```
kieckhafer: 😬

aldeed:
@kieckhafer Migrations don't run on a fresh DB now because they shouldn't need to, but for example if your db is at 63 and you start up, you should still see it running 64 and 65.

I'll look into the missing `roles`. That might be another regression. A lot of magic stuff was happening that it's hard to keep track of.

aldeed:
@kieckhafer Roles should be inserted properly on startup now

PR #5220, opened 2019-06-18
Minor inventory performance improvments

### LINKED ISSUE #5215: simple-inventory and inventory plugin optimizations
Two performance tweaks that should make updating simple-inventory faster:
- `updateSimpleInventory` calls `recalculateReservedSimpleInventory` only when an insert happened (product that hasn't had inventory stored yet). In this situation, `recalculateReservedSimpleInventory` emits "afterInventoryUpdate" and then `updateSimpleInventory` immediately emits that same event again. `updateSimpleInventory` function can rely on it being emitted by `recalculateReservedSimpleInventory` and skip emitting it for upserts.
- In the core `inventory` plugin, there is an `afterInventoryUpdate` appEvent listener. This does `Products.find` without a projection. I believe only the `_id`, `shopId`, and `ancestors` fields are ever used, so we can add a `projection` to that query to limit the data returned.

rosshadden:
Some minor inventory performance improvements.

Fixes #5215.

### DIFF
--- a/imports/plugins/core/inventory/server/no-meteor/utils/startup.js
+++ b/imports/plugins/core/inventory/server/no-meteor/utils/startup.js
@@ -18,6 +18,10 @@ export default function startup(context) {
       ancestors: productId,
       isDeleted: { $ne: true },
       isVisible: true
+    }, {
+      _id: 1,
+      ancestors: 1,
+      shopId: 1
     }).toArray();
 
     const topVariants = variants.filter((variant) => variant.ancestors.length === 1);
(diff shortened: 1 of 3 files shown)


aldeed:
@rosshadden I'm approving the code, but making that emit conditional did cause an integration test to fail, so you should investigate that. It might be something incorrect about the test setup or expectations.

The DCO check is also failing, though I don't know why since your commits seem to be signed. We can always ignore that check if you can't resolve it.

rosshadden:
I will look into the failing integration tests. Doesn't make a lot of sense at the moment.

The DCO says:

> Commit sha: 4937a35, Author: Erik Kieckhafer, Committer: Erik Kieckhafer; The sign-off is missing.

So it's referring to a commit in `develop` already.

rosshadden:
I actually ended up having to do the opposite of what you suggested. Emitting that event without an `await` is what causes the tests to fail. The reason they didn't before is because it was always (redundantly) emitting, and awaiting. The one that runs in `recalculateReservedSimpleInventory` did not await, and thus when the code hit that path after my change it was causing tests to fail. So I just added an await to that one to satisfy tests.

I don't like that I made a change specifically to satisfy tests, especially considering it would be a tiny performance improvement to not `await` here, but I do not currently understand this code or tests enough to confidently fundamentally change the way they work.

aldeed:
@rosshadden Sounds good. I'll have to look more closely at that test. There may be a way to alter the expectations instead.

PR #5221, opened 2019-06-18
[CircleCI] Fix environment variable syntax

elhil:
`save_cache` uses special syntax for variable interpolation ([docs](https://circleci.com/docs/2.0/configuration-reference/#save_cache))

Impact: **minor**  
Type: **chore**

## Issue
Currently the `GLOBAL_CACHE_VERSION` environment variable doesn't actually change the cache version.

## Solution
Change `config.yml` to use the correct syntax.

## Breaking changes
This will invalidate any build caches in CircleCI.

## Testing
1. Merge.
2. Check for the correct cache name in the logs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
           docker_layer_caching: true
       - restore_cache:
           name: Restoring Meteor cache
-          key: reaction-{{ .Environment.GLOBAL_CACHE_VERSION }}-meteor
+          key: reaction-v2-meteor
       - run:
           name: Install Meteor
           command: |
@@ -36,7 +36,7 @@ jobs:
             fi
       - save_cache:
           name: Saving Meteor to cache
-          key: reaction-{{ .Environment.GLOBAL_CACHE_VERSION }}-meteor-{{ epoch }}
+          key: reaction-v2-meteor-{{ epoch }}
           paths:
             - ~/.meteor
       - run:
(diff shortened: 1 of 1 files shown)


focusaurus:
If I'm reading the docs correctly, this won't work for arbitrary env vars which `GLOBAL_CACHE_VERSION` is in our case I think:

> supports any environment variable exported by CircleCI or added to a specific Context—not any arbitrary environment variable

elhil:
@focusaurus ah you're correct--I was using project-level variables. Given that the old approach doesn't work either—it just creates a cache under the literal "${GLOBAL_CACHE_VERSION}-meteor"—, would it make sense to use a hardcoded prefix?

focusaurus:
There's a lot about what's going on here in the existing code/caching strategy big picture that I'm dubious about. At the very least I think it would be a good change to remove all the GLOBAL_CACHE_VERSION references and put in a hard coded thing, but thinking more broadly I think

- Why are we saving caches with `{{ epoch }}`. I do not understand the utility of this. Can we just remove these `save_cache` steps entirely?
- Why are we versioning caches manually with a global flag? What scenario would cause us to bump this version? Can we just remove the whole global version concept entirely?

elhil:
> Why are we saving caches with `{{ epoch }}`. I do not understand the utility of this. Can we just remove these `save_cache` steps entirely?

[restore_cache](https://circleci.com/docs/2.0/configuration-reference/#restore_cache) specifies a prefix, the _most recent_ match of which is used. So adding `{{ epoch }}` is just a way of ensuring key uniqueness.

> Why are we versioning caches manually with a global flag? What scenario would cause us to bump this version? Can we just remove the whole global version concept entirely?

I'm also out of my depth here. I'm not sure why we don't use `checksum` or the like.

focusaurus:
OK @elhil if you change to just a hardcoded cache version like `reaction-v2-meteor` we can merge this. Thanks!

PR #5229, opened 2019-06-20
feat: bulk simple inventory mutation

impactmass:
Resolves NA 
Impact: **major**  
Type: **feature|performance**

## Issue
When running large inventory import via an external sync system, the lack of a bulk mutation makes the import takes many hours. A bulk interface is needed.

## This PR solves it by:
- Adding a bulk inventory update mutation
- Only available on context. Not exposed on GQL at this point

## Breaking changes
N/A

## Testing
- Confirm that single update mutation (i.e `updateSimpleInventory`) works as before
- Test the bulk interface by running an external import process (see offline notes)

### DIFF
--- a/imports/plugins/included/simple-inventory/server/no-meteor/mutations/updateSimpleInventoryBulk.js
+++ b/imports/plugins/included/simple-inventory/server/no-meteor/mutations/updateSimpleInventoryBulk.js
@@ -1,0 +1,147 @@
+import SimpleSchema from "simpl-schema";
+import Random from "@reactioncommerce/random";
+import Logger from "@reactioncommerce/logger";
+import { ProductConfigurationSchema, SimpleInventoryCollectionSchema } from "../simpleSchemas";
+
+const logCtx = { name: "core-inventory" };
+
+const inputSchema = new SimpleSchema({
+  productConfiguration: ProductConfigurationSchema,
+  canBackorder: {
+    type: Boolean,
+    optional: true
+  },
+  inventoryInStock: {
+    type: SimpleSchema.Integer,
+    min: 0,
+    optional: true
+  },
+  isEnabled: {
+    type: Boolean,
+    optional: true
+  },
+  lowInventoryWarningThreshold: {
+    type: SimpleSchema.Integer,
+    min: 0,
+    optional: true
+  },
+  shopId: String
+});
+
+const updateFields = [
+  "canBackorder",
+  "inventoryInStock",
+  "isEnabled",
+  "lowInventoryWarningThreshold"
+];
+
+const defaultValues = {
+  canBackorder: false,
+  inventoryInStock: 0,
+  isEnabled: false,
+  lowInventoryWarningThreshold: 0
+};
+
+// eslint-disable-next-line require-jsdoc
+export default async function updateSimpleInventoryBulk(context, input) {
+  const { appEvents, collections } = context;
+  const { SimpleInventory } = collections;
+
+  const bulkOperations = input.updates.map((update) => {
+    const { productConfiguration, shopId } = update;
+    inputSchema.validate(update); // try/catch
+    const $set = { updatedAt: new Date() };
+    const $setOnInsert = {
+      "_id": Random.id(),
+      "createdAt": new Date(),
+      // inventoryReserved is calculated by this plugin rather than being set by
+      // users, but we need to init it to some number since this is required.
+      // Below we update this to the correct number if we inserted.
+      "inventoryReserved": 0,
+      // The upsert query below has only `productVariantId` so we need to ensure both are inserted
+      "productConfiguration.productId": productConfiguration.productId
+    };
+    updateFields.forEach((field) => {
+      const value = update[field];
+      if (value !== undefined && value !== null) {
+        $set[field] = value;
+      } else {
+        // If we are not setting the value here, then we add it to the setOnInsert.
+        // This is necessary because all fields are required by the collection schema.
+        $setOnInsert[field] = defaultValues[field];
+      }
+    });
+
+    // if (Object.getOwnPropertyNames($set).length === 1) {
+    //   throw new ReactionError("invalid-param", "You must provide at least one field to update.");
+    // }
+
+    const modifier = { $set, $setOnInsert };
+
+    SimpleInventoryCollectionSchema.validate({
+      $set,
+      $setOnInsert: {
+        ...$setOnInsert,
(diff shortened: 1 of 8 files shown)


imports/plugins/included/simple-inventory/server/no-meteor/mutations/updateSimpleInventoryBulk.js
```diff
@@ -0,0 +1,147 @@
...
+  "isEnabled",
+  "lowInventoryWarningThreshold"
+];
+
+const defaultValues = {
+  canBackorder: false,
+  inventoryInStock: 0,
+  isEnabled: false,
+  lowInventoryWarningThreshold: 0
+};
```
aldeed: Since these 3 const are the same in both files and we want to keep them the same in the future, I'd move them to a separate file and import them in both places.

imports/plugins/included/simple-inventory/server/no-meteor/mutations/updateSimpleInventoryBulk.js
```diff
@@ -0,0 +1,147 @@
...
+    const modifier = { $set, $setOnInsert };
+
+    SimpleInventoryCollectionSchema.validate({
+      $set,
+      $setOnInsert: {
+        ...$setOnInsert,
+        "productConfiguration.productVariantId": productConfiguration.productVariantId,
+        shopId
+      }
+    }, { modifier: true, upsert: true });
```
aldeed: I wonder if you could move everything from `inputSchema.validate` line to here into a third util file/function, and then call that from both the singular and the bulk mutations? Everything is identical in this section, right?

imports/plugins/included/simple-inventory/server/no-meteor/mutations/updateSimpleInventoryBulk.js
```diff
@@ -0,0 +1,147 @@
...
+      } else {
+        // If we are not setting the value here, then we add it to the setOnInsert.
+        // This is necessary because all fields are required by the collection schema.
+        $setOnInsert[field] = defaultValues[field];
+      }
+    });
+
+    // if (Object.getOwnPropertyNames($set).length === 1) {
+    //   throw new ReactionError("invalid-param", "You must provide at least one field to update.");
+    // }
```
aldeed: For this check, you could maybe keep it but just log it and `return null` instead of throwing? Then you'll need to filter out `null` from `bulkOperations` before the `bulkWrite`.

imports/plugins/included/simple-inventory/server/no-meteor/mutations/updateSimpleInventoryBulk.js
```diff
@@ -0,0 +1,147 @@
...
+    });
+  }
+
+  if (res.result.nModified > 0) {
+    const modifiedDocsOpData = bulkOperations.filter((op) => !upsertedIds.includes(op.updateOne.update.$setOnInsert._id));
+    const productConfigurations = modifiedDocsOpData.map((operationData) => ({
+      productId: operationData.updateOne.update.$setOnInsert["productConfiguration.productId"]
+    }));
+    Logger.debug({ ...logCtx, productConfigurations }, "Product configurations for afterBulkInventoryUpdate");
+    appEvents.emit("afterBulkInventoryUpdate", { productConfigurations });
```
aldeed: Even though we don't currently use it, keep this consistent with "afterInventoryUpdate" by including the user:

```
{ productConfigurations, updatedBy: context.userId }
```

I'd also add `productVariantId` to the `productConfigurations` in your map, even though we don't use that. Someone else might consume the event and need it.

imports/plugins/included/simple-inventory/server/no-meteor/mutations/updateSimpleInventoryBulk.js
```diff
@@ -0,0 +1,147 @@
...
+];
+
+const defaultValues = {
+  canBackorder: false,
+  inventoryInStock: 0,
+  isEnabled: false,
+  lowInventoryWarningThreshold: 0
+};
+
+// eslint-disable-next-line require-jsdoc
```
aldeed: When you write function docs, include a warning that we aren't currently checking permissions in here because it's not exposed through GraphQL, but permissions should be added if that ever changes.

imports/plugins/included/simple-inventory/server/no-meteor/mutations/updateSimpleInventoryBulk.js
```diff
@@ -0,0 +1,97 @@
...
+          "productConfiguration.productVariantId": productConfiguration.productVariantId,
+          shopId
+        },
+        update: modifier,
+        upsert: true
+      }
+    };
+  });
+
+  bulkOperations = bulkOperations.filter((op) => op !== null);
```
focusaurus: I read this to mean if an update doesn't validate, we ignore it and swallow the error. Seems like we should include a `invalidUpdates` property on the returned result summary object. Swallowing errors makes upstream problems harder to detect and debug.

imports/plugins/included/simple-inventory/server/no-meteor/mutations/updateSimpleInventoryBulk.js
```diff
@@ -0,0 +1,97 @@
...
+      return isFilterEqual && isUpdateEqual;
+    });
+    if (match) return true;
+    return false;
+  });
+
+  // If we inserted, set the "reserved" quantity to what it should be. We could have
+  // put this in the $setOnInsert but then we'd have to do the Orders lookup for
+  // calculating reserved every time, even when only an update happens. It's better
+  // to wait until here when we know whether we inserted.
```
focusaurus: Good comment!

imports/plugins/included/simple-inventory/server/no-meteor/mutations/updateSimpleInventoryBulk.js
aldeed: I agree, and I also think `failedOps` should have the same original `productConfiguration` that was passed in on each one. So altogether, maybe a function return object something like this:

```js
{
  failedUpdates: [
    {
      productConfiguration,
      writeError
    }
  ],
  invalidUpdates: {
    productConfiguration,
    validationError
  }
}
```

impactmass:
I've made the response object structure as recommended. I want to add a few tests before opening this back for review

impactmass:
Added a test for the bulk version based on a similar existing single-update one.
Fixed conflict & DCO
Ready for re-re-review.

PR #5230, opened 2019-06-21
Alphabetical currencies

### LINKED ISSUE #4757: [bug] Currencies non-alphabetical after enabling all via Localization
## Issue Description
The currency drop-down list is non-alphabetical after enabling all the currencies.

## Steps to Reproduce
1. Go to Localization on Admin Panel
2. Enable All Currencies
3. Refresh page
4. Press Drop Down menu for Currency on Nav Bar

### Versions
Reaction 2.0.0-rc.5

chrispotter:
Resolves #4757
Impact: *minor**  
Type: **bugfix**

## Issue
Sorts Currencies Alphabetically in Operator UI

## Solution
`this.props.currencies.sort((a, b) => a.name.localeCompare(b.name)).map(...`

## Breaking changes
None

## Testing
1. http://localhost:3000/operator/localization
2. See that currencies are in order

### DIFF
--- a/imports/plugins/core/i18n/client/components/localizationSettings.js
+++ b/imports/plugins/core/i18n/client/components/localizationSettings.js
@@ -72,7 +72,10 @@ class LocalizationSettings extends Component {
   }
 
   renderCurrencies() {
-    return this.props.currencies.map((currency, key) => (
+    console.log(this.props.currencies);
+    return this.props.currencies.sort(
+      (a, b) => a.name.localeCompare(b.name)
+    ).map((currency, key) => (
       <Components.ListItem
         actionType={"switch"}
         key={key}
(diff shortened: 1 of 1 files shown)


imports/plugins/core/i18n/client/components/localizationSettings.js
```diff
@@ -72,7 +72,7 @@ class LocalizationSettings extends Component {
   }
 
   renderCurrencies() {
-    return this.props.currencies.map((currency, key) => (
+    return this.props.currencies.sort((a, b) => a.name.localeCompare(b.name)).map((currency, key) => (
```
kieckhafer: This isn't a blocker at all, but I've been auditing `eslint warnings`, and short variable names is one of them. Can we update this to be `currencyA` and `currencyB`?

PR #5232, opened 2019-06-24
feat: add regex filter to tags GQL query

### LINKED ISSUE #5224: Catalyst Tag Table filter does not work properly
This is a ticket created from the remaining work not completed in #5218:

## Issue Description

1. Tag filtering doesn't work. The filter applies after the query, so you will still have many pages to paginate through, but sometimes the pages will have fewer results.

## Possible Solution

1. Explore using a regex search like is used in the product editor to find matches. This will admin only.

kieckhafer:
Resolves #5224   
Impact: **minor**  
Type: **feature**

## Solution
Add a regex filter to the graphql input, to send to Mongo, to filter out data before it gets to the client

## Breaking changes
None

### Possible issues with this solution
Since this re-queries each time there is new input inside the tags table, this will use the GQL tags query frequently. Will this be a heavily used feature? If not, this might be OK. If so, what are some other options on how to implement this?

## Testing
1. Load a big set of Tags into the database
1. Filter the tags in the Admin UI
1. See that all the tags filtered make it to the front of the table, and you don't need to go searching through random pages of tags to find what you need.

### DIFF
--- a/imports/plugins/core/catalog/server/no-meteor/queries/tags.js
+++ b/imports/plugins/core/catalog/server/no-meteor/queries/tags.js
@@ -6,12 +6,13 @@
  * @param {Object} context - an object containing the per-request state
  * @param {String} shopId - ID of shop to query
  * @param {Object} [params] - Additional options for the query
+ * @param {Boolean} [params.filter] - If provided, look for `filter` matching this value with regex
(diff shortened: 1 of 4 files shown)


kieckhafer:
@mikemurray updated to only allow admins, and with a regex sanitizer

PR #5234, opened 2019-06-25
Demeteorize registration for remaining API server plugins

aldeed:
Impact: **minor**  
Type: **refactor**

Part of epic #4996 

## Changes
Demeteorize plugin registration for the remaining API server plugins that didn't yet have this done. The only remaining uses of `Reaction.registerPackage` are for client-only plugins that provide `registry`. We'll convert those to use newer component registration API on the client side soon.

## Breaking changes
None

## Testing
It should be enough to verify the app starts, all tests pass, and basic happy path testing.


imports/plugins/core/address/register.js
```diff
@@ -1,33 +1,4 @@
...
-  registry: [
-    {
-      label: "Address Validation",
-      provides: ["shopSettings"],
-      container: "dashboard",
-      template: "ShopAddressValidationSettings"
-    }
-  ]
-});
+Reaction.whenAppInstanceReady(register);
```
rosshadden: I'm not a fan of this organization, as opposed to having the contents of that register file here, in this register file.

imports/node-app/index.js
```diff
@@ -15,28 +11,25 @@ if (!ROOT_URL) throw new Error("You must set ROOT_URL");
 const app = new ReactionNodeApp({
   debug: true,
   context: {
-    mutations,
-    queries,
-    rootUrl: ROOT_URL
+    rootUrl: ROOT_URL,
```
rosshadden: These were likely intentionally sorted in alphabetical order.

imports/plugins/core/email/server/no-meteor/register.js
```diff
@@ -1,5 +1,13 @@
-import mutations from "./mutations";
-import startup from "./startup";
+import { Meteor } from "meteor/meteor";
+
+// This is temporary. Mutations still import jobs, which don't
+// work outside of a Meteor environment.
+let mutations = {};
+let startup = () => {};
+if (!Meteor.isFakeMeteor) {
+  mutations = require("./mutations").default;
```
rosshadden: I think it's weird to mix import/export syntax with CommonJS requires. You can dynamically import modules with the `import` function. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#Dynamic_Imports for details.

imports/plugins/core/core/server/startup/startNodeApp.js
```diff
@@ -41,15 +37,13 @@ export default async function startNodeApp({ onAppInstanceCreated }) {
       async createUser(options) {
         return Accounts.createUser(options);
       },
-      queries: coreQueries,
-      mutations: coreMutations,
       rootUrl: ROOT_URL,
+      mutations: {},
```
rosshadden: Why are `mutations` and `queries` passed if they are empty?

willopez:
@aldeed are Ross's comments holding back this PR from being merged?

imports/plugins/core/core/server/startup/startNodeApp.js
aldeed: There's code that requires it and this was easier than refactoring that code. This is temporary and will be resolved when we finish entirely separating this into a Node app.

imports/node-app/index.js
aldeed: Good catch. I removed them, then later added back in the wrong place.

imports/plugins/core/email/server/no-meteor/register.js
aldeed: I agree but I don't think import as a function is supported yet in Node 8. It may work due to our Babel plugins, but since this is only temporary until no-Meteor conversion is done and I'd have to `await` it, I think I'll stick with this.

imports/plugins/core/address/register.js
aldeed: This is necessary temporarily to be backwards compatible while finishing removing Meteor. Soon this file will go away and only the inner `register.js` will remain, but currently this file is what our Meteor tooling loads.

aldeed: I added a comment with this explanation at the top of all these files. Thanks for calling it out as confusing.

PR #5252, opened 2019-06-29
fix: ensure translations from all packages load

aldeed:
Impact: **minor**  
Type: **bugfix**

## Issue
After one of my PRs was merged yesterday changing package registration, this caused some translations not to be loaded on the client, resulting in some operator UI text being blank or showing defaults.

## Solution
Adjust the browser i18n loading to not rely on `Packages` collection.

## Breaking changes
None

## Testing
Start the app and verify that "Products" label shows in the operator UI sidebar.


aldeed:
@willopez This needs to get into the RC

Issue #5206, opened 2019-06-10
Rename "optionTitle" to something more clear as to what it's purpose is


zenweasel:
I can't tell you how much time has been wasted and confusion caused because this field is poorly named. I don't know why it wasn't called `cartTitle` or `shortTitle` or something more clear in the first place but here we are.

aldeed:
I don't know that `cartTitle` is 100% accurate either (the `title` becomes the cart item title), but `shortTitle` makes sense to me.

Related, in `CartItem`, I think it would make more sense if:
- `title` -> `productTitle`
- `optionTitle` deprecate it because it's in `attributes`

PR #5253, opened 2019-06-29
feat: better cart and order item attributes, with labels

### LINKED ISSUE #5208: Fix CartItem attributes
## Issue Description
In `imports/plugins/core/cart/server/no-meteor/util/addCartItems.js`, the variant `title`s are used as the `attributes` values, but it should be using the `optionTitle`.

Also, `label` is `null` for both attributes. We should add a field to the variant schema to collect a label. This will move things closer to having an attributes-based catalog schema.

## Possible Solution
- Add `attributeLabel` string field to the variant schema and make it required
- Add this field to the operator UI variant edit form.
- In `imports/plugins/core/cart/server/no-meteor/util/addCartItems.js`, use `parentVariant.attributeLabel` and `chosenVariant.attributeLabel` as the `label` field values (and use `optionTitle` rather than `title` to fix the bug mentioned above).
- Change `attributes.label` to be required in `CartItem` schema.
- Migration for `Products` and `Catalog` to set `attributeLabel` to `variant` for all top-level variants and to `option` for all second-level variants.
- Migration for `CartItem` to set `label` to `variant` for length 1 `attributes` array and to `variant` and `option` for length 2 `attributes` array.

### Versions
develop

aldeed:
Resolves #5208  
Impact: **minor**  
Type: **feature**

## Issue
`CartItem` and `OrderItem` have `attributes` array, but `label` was `null` and `value` was pulled from the wrong product field. `OrderItem` always had empty array.

One underlying problem is that we had no label for variants, which is why we have null attribute labels.

A related problem is general confusion around `title` vs. `optionTitle` fields for variants.

## Solution
- Added optional `attributeLabel` string field to `Products` of type "variant"
- Added required `attributeLabel` string field to all `CatalogProductVariant`
- Added an operator UI field to the variant edit form, "Attribute label". Also generally improved the labels, placeholder text, and help text for the `attributeLabel`, `optionTitle`, and `title` fields on that form, to help avoid more confusion.
- Properly set `attributes` array on `OrderItem`s when they are created
- Properly set `attributes` array on `CartItem`s when they are created
- Migration to set `attributeLabel` to a default value in `Products` and `Catalog` collections. For top-level variants, set to "Variant". For option variants, set to "Option".
- Migration to set `label` for all `attributes` on all `CartItem` to a default value in `Cart` collection. For first attribute, set to "Variant". For second attribute, if present, set to "Option".

NOTE: I opted not to do #5206 in this PR to avoid it becoming too large.

## Breaking changes
As long as migrations run, nothing should break. If client code is using cart item `attributes` and depending on what the `value` of them currently is (the full title), it may need to be updated.

## Testing
1. Test the new and updated fields at the top of the variant edit form. Test for top-level variant and for option variant. Verify you cannot publish without an attribute label set.
2. Verify you can query GraphQL for `attributeLabel` field for any variant or option in a `CatalogItemProduct`.
3. Test the migration and verify that all data is migrated as described.

### DIFF
--- a/imports/plugins/core/core/server/Reaction/absoluteUrl.app-test.js
+++ b/imports/plugins/core/core/server/Reaction/absoluteUrl.app-test.js
@@ -6,83 +6,88 @@ import { sinon } from "meteor/practicalmeteor:sinon";
 import { composeUrl } from "/lib/core/url-common";
 import core from "./core";
 
-describe("AbsoluteUrlMixin", () => {
-  let sandbox;
+/**
+ * @return {String} A random string
+ */
+function randomString() {
+  return Math.random().toString(36);
+}
 
-  beforeEach(() => {
-    sandbox = sinon.sandbox.create();
-  });
-
-  afterEach(() => {
-    sandbox.restore();
-  });
-
-  describe("#absoluteUrl", () => {
-    let path;
-    let connectionHost;
-    let rootUrl;
-    let meteorRootUrl;
+core.onAppStartupComplete(() => {
+  describe("AbsoluteUrlMixin", () => {
+    let sandbox;
 
     beforeEach(() => {
-      // commonly used test vars
-      path = randomString();
-      connectionHost = `${randomString()}.reactioncommerce.com`;
-      rootUrl = `https://${randomString()}.reactioncommerce.com`;
-
-      // mocking Meteor
-      meteorRootUrl = composeUrl.defaultOptions.rootUrl;
-      // this is a round-about way of mocking $ROOT_URL
-      composeUrl.defaultOptions.rootUrl = rootUrl;
+      sandbox = sinon.sandbox.create();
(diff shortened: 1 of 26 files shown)


imports/plugins/core/core/server/Reaction/absoluteUrl.app-test.js
```diff
@@ -6,83 +6,88 @@ import { sinon } from "meteor/practicalmeteor:sinon";
...
       });
-    });
 
-    it("accepts options the same way composeUrl does", () => {
-      const options = {
-        secure: true,
-        replaceLocalhost: true,
-        rootUrl: "http://127.0.0.1"
-      };
+      it("accepts options the same way composeUrl does", () => {
```
willopez: @mikemurray white space, does this cause a lint error?

PR #5261, opened 2019-07-02
[WIP] refactor: remove `coreOrderItemWorkflow

kieckhafer:
Resolves #issueNumber  
Impact: **major**  
Type: **refactor**

## Issue
Items are managed by fulfillment groups. Since the group has a status on it, we don't also need a status on each item.

## Solution
Remove `coreOrderItemStatus` from the `Order.item` schema, and remove all code related to updating this status. Where applicable, change status checks to check group status.

## Testing
1. Make sure you can create an order
1. Move the order through various shipment statuses (cancel, shipped, completed, etc)
1. See that everything works as it should


kieckhafer:
@aldeed there are four places where we use the function `updateGroupStatusFromItemStatus`. [1](https://github.com/reactioncommerce/reaction/blob/ref-kieckhafer-removeCoreOrderItemWorkflow/imports/plugins/core/orders/server/no-meteor/mutations/cancelOrderItem.js#L128) [2](https://github.com/reactioncommerce/reaction/blob/ref-kieckhafer-removeCoreOrderItemWorkflow/imports/plugins/core/orders/server/no-meteor/mutations/addOrderFulfillmentGroup.js#L111) [3](https://github.com/reactioncommerce/reaction/blob/ref-kieckhafer-removeCoreOrderItemWorkflow/imports/plugins/core/orders/server/no-meteor/mutations/moveOrderItems.js#L133) [4](https://github.com/reactioncommerce/reaction/blob/ref-kieckhafer-removeCoreOrderItemWorkflow/imports/plugins/core/orders/server/no-meteor/mutations/splitOrderItem.js#L123)

This function checks the statuses of all items in a group, and if they are all the same, will update the group status to match all the item statuses.

If we no longer have item statuses, that means group statuses will need to always reflect EVERY item inside the group, so then do we need to break items into a new group if only a partial amount of items are manipulated by `cancelOrderItem` or a similar function? If 2/5 items are canceled, we'll need to create a new group with the other 3 items?

aldeed:
@kieckhafer Yes, the idea is that there would only be a group status and moving items to that group and/or creating a new group with a new status would be the way to do it now.

kieckhafer:
Closing this to re-pick up at a later time

PR #5264, opened 2019-07-03
(fix): set slugified title as permalink

### LINKED ISSUE #4735: Generated "permalink" uses ID rather than using slugified title
## Issue Description
The permalink should generated from a slugified version of the product title whenever possible (which should be always?). Currently it's using the ID instead. This is bad for SEO and readability

## Steps to Reproduce
1. Create a new product
1. Observe that the permalink is not a slugified version of the title

## Possible Solution
The `handle` field is not being set properly. Possibly this is caused by `getSlug` or similar not being available? Just a guess

### Versions
Reaction: 2.0.0-rc5

mpaktiti:
Resolves #4735 
Impact: **major**  
Type: **bugfix**

## Issue
When you create a product, the permalink (handle) is set to a random ID instead of the slugified version of the title. This is happening because:
- the code initially sets the handle to a random ID (in the meanwhile, the title is null)
- when the title is set the handle is not updated because the condition `doc.handle === doc._id` is never true (the handle is set to a random ID, not the doc ID)

Steps to reproduce:
1. Create a new product
2. Observe that the handle is not a slugified version of the title
3. Update the title -> the handle is not updated

## Solution
I removed the `doc.handle === doc._id` condition. This way when the user sets a Title for the product, the handle is updated. I also removed the initial population with a random ID because I see no use for it. 

## Breaking changes
None

## Testing
1. Log in the Operator UI and create a product
2. Both the Title and the Handle are null
3. Set a Title for the product and tab to another field
4. Notice how the Handle is updated to the title's slugified version
5. Keep changing the Title and notice how the Handle keeps getting updated each time


zenweasel:
@mpaktiti I think setting the handle temporarily is to handle the case where you would navigate out of the PDP and then click back again and the product you were editing would disappear because products are only accessible by handle so this newly created product would never be accessible. Can you double check that you haven't introduced a new bug around that scenario?

mpaktiti:
@zenweasel Steps followed:
- Go to Products
- Click Create product
- Fill no info (both title and handle are null)
- Click on Orders
- Click back on Products -> I can see the product I created earlier (2nd row in the screenshot below)

Let me know if I didn't get the steps right. 

Screenshot from 2019-07-04 15-42-59

zenweasel:
Are you able to click back into that product? If so I think we're OK

mpaktiti:
@zenweasel Yes I can step back and edit it. The URL uses the product ID. For example, for an untitled product I created the URL is `http://localhost:3000/operator/products/eMnJA25oWMGYP55zn` and I can see in Mongo that `eMnJA25oWMGYP55zn` is the product's `_id`.

zenweasel:
@mpaktiti You need to sign your commits when committing to the open source projects. Click on "details" next to the DCO check on how to rectify

zenweasel:
@manueldelreal can you give this a quick test please?

imports/plugins/core/catalog/server/methods/catalog.js
```diff
@@ -816,7 +814,7 @@ Meteor.methods({
         // Should be a call similar to the line below.
         [field]: createHandle(Reaction.getSlug(value), _id) // handle should be unique
       };
-    } else if (field === "title" && doc.handle === doc._id) {
```
mikemurray: The purpose of this is to ensure that permalink isn't updated every time someone changes a title. With the removal of the `ID` being the default permalink, this is definitely less useful; however, there should be new check added for the existence of the permalink and block updates if it's set.

The permalink changing every time the title changes will cause broken links. If someone wants to break their permalink, they can do so from the field itself, and make whatever redirects they want in their storefront if necessary.

What I'd like to see from this is:

- The permalink should be set the first time automatically based on the title. 
- The permalink doesn't update automatically with the title change if it's already set.
- The permalink is manually updateable through the field, which it is now.

mpaktiti: Hey @mikemurray , thanks for the feedback. One clarification about the first bullet point. 

When we create a product the title is null. I'll have the permalink updated automatically only the first time the title is set by the user. Until then it will be null (and after the first update it will not be updated automatically again). Sounds good or should we populate it initially with a default value?

mikemurray: The permalink could be blank until the first update of the title. Products can't be published with a null title. Not sure about a null permalink, but the null title will make sure the permalink it set.

manueldelreal:
@zenweasel @mpaktiti waiting on the conversation thread between @mikemurray and @mpaktiti to be resolved before testing this one with its final form. 

cc @zenweasel

imports/plugins/core/catalog/server/methods/catalog.js
manueldelreal: is this thread resolved? asking for a friend that needs to test this.

mpaktiti: @manueldelreal No it's not! I'll ping you once it's ready for testing.

manueldelreal:
I'm still getting the graphql error for "birthday is required" with this branch. it's con the "onSubmit" function

imports/plugins/core/catalog/server/methods/catalog.js
mpaktiti: I changed this locally and this is the behavior I see:

- Go to Products and create a new product. The title is not set yet and therefore the permalink is empty as well.

image

- Set a value for title -> permalink **is** updated automatically

image

- Update the title -> permalink **is not** updated automatically

image

- Update the permalink manually -> update successful

image

However, if I delete the permalink manually and try to Publish, the system lets me do it. Should I add the same validation we have for Title and not allow them to publish if the permalink is null?
@mikemurray @zenweasel

mpaktiti: Actually the title validation is not on Publish, it doesn't let you delete to start with:

image

Add the same for permalink?

mikemurray: I think it's fine to make the permalink required since it no-longer defaults to the product id.

mpaktiti: @mikemurray Can you please review the changes I just pushed? Behaviour is the same as I described in my previous comment, plus if you try to delete the permalink you get this error:

image

PR #5287, opened 2019-07-08
fix: typo breaking docs build (5286)

### LINKED ISSUE #5286: Deploy-docs step fails consistently during CI
## Summary
Our CI is reporting failure due to the `deploy-docs` step that happens when our `master` branch is updated. If we still rely on this step to deploy docs, let's fix this. If we don't, let's eliminate it please. I'd like Circle to report success for our `master` branch. Our CI process should give us and people looking at our projects confidence that the tests are passing and the project is maintained. 

image

```
Generating output files...
Finished running in 9.28 seconds.
Exited with code 1
```

spencern:
**Resolves** #5286 
Impact: **minor**  
Type: **bugfix**

## Issue
Fixes #5286 which was caused by this line
https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/included/product-admin/client/hocs/withVariant.js#L109 breaking the jsdoc build.

It's unclear to me why this line didn't cause our app build to fail.

### DIFF
--- a/imports/plugins/included/product-admin/client/hocs/withVariant.js
+++ b/imports/plugins/included/product-admin/client/hocs/withVariant.js
@@ -106,7 +106,7 @@ function composer(props, onData) {
   } else {
     variants = ReactionProduct.getTopVariants();
   }
-  let variantMedia: [];
+  let variantMedia = [];
 
   if (variants) {
     if (Array.isArray(variants)) {


mikemurray:
The issue is because the versions of babel react preset <v7 included https://babeljs.io/docs/en/babel-plugin-transform-flow-strip-types. Not sure what version we're on as Meteor manages that dependency.

spencern:
Thanks for digging that up @mikemurray

PR #5289, opened 2019-07-09
refactor: add isSoldOut prop to variants and options in Catalog

willopez:
Impact: minor
Type:  refactor

## Summary
Sometimes in is necessary and convenient to have inventory information at the variant and option level.

Use case 1: In product listing page, it should be possible not to render or disable sold out variants.

Use case 2. In product detail page it should be possible to not render or disable sold out variants or options.

Use case 3: When indexing data, some systems such as Elasticsearch monitor the `Catalog` collection to update the index. If inventory information is not present in the `Catalog` collection, it will not be synced to the search index.

## Testing
Test 1
1. start Reaction and verify migration 68 completed successfully.
2. Verify the `isSoldOut` props is present in the variants and options in the `Catalog` collection.
3. Verify value of `isSoldOut` prop is correct. If setting `canSellVariantWithoutInventory` is set to true, which it is by default, then `isSoldOut` should be `false`. If the setting is set to false, and inventory tracking is enabled for a variant or option, then it should be reflective of the quantity value. If `quantity > 0`, then `isSoldOut` = false, if `quantity == 0`, then `isSoldOut` should be true for that variant or option.

Test 2
1. Enable inventory tracking for an option
2. Set quantity to 0
3. Verify product option is sold out in the storefront

Test 3
1. Enable inventory tracking for a product with at least on variant and two options.
2. Set inventory to 0 for all options of a variant
3. Verify variants and options are sold out, (disabled) in storefront.


imports/plugins/core/inventory/server/no-meteor/utils/publishProductToCatalog.js
```diff
@@ -9,20 +9,47 @@ export default async function publishProductToCatalog(catalogProduct, { context,
...
+  // add inventory props to each top level Variant
+  catalogProduct.variants.forEach((variant, variantIndex) => {
+    // attempt to find this variant's inventory info
+    const foundVariantInventory = topVariantsAndOptionsInventory.find((inventoryInfo) => inventoryInfo.productConfiguration.productVariantId === variant._id);
+
+    // if inventory info was found, add to variant
+    if (foundVariantInventory) {
+      catalogProduct.variants[variantIndex].isBackorder = foundVariantInventory.inventoryInfo.isBackorder;
+      catalogProduct.variants[variantIndex].isLowQuantity = foundVariantInventory.inventoryInfo.isLowQuantity;
+      catalogProduct.variants[variantIndex].isSoldOut = foundVariantInventory.inventoryInfo.isSoldOut;
```
aldeed: I believe `variant` is by reference, so you should be able to do `variant` instead of `catalogProduct.variants[variantIndex]` here.

imports/plugins/core/inventory/server/no-meteor/utils/publishProductToCatalog.js
```diff
@@ -9,20 +9,47 @@ export default async function publishProductToCatalog(catalogProduct, { context,
...
+
+    // add inventory props to each top level option
+    variant.options.forEach((option, optionIndex) => {
+      const foundOptionInventory = topVariantsAndOptionsInventory.find((inventoryInfo) => inventoryInfo.productConfiguration.productVariantId === option._id);
+
+      // if inventory info was found, add to variant
+      if (foundOptionInventory) {
+        catalogProduct.variants[variantIndex].options[optionIndex].isBackorder = foundOptionInventory.inventoryInfo.isBackorder;
+        catalogProduct.variants[variantIndex].options[optionIndex].isLowQuantity = foundOptionInventory.inventoryInfo.isLowQuantity;
+        catalogProduct.variants[variantIndex].options[optionIndex].isSoldOut = foundOptionInventory.inventoryInfo.isSoldOut;
```
aldeed: I believe `option` is by reference, so you should be able to do `option` instead of `catalogProduct.variants[variantIndex].options[optionIndex]` here.

imports/plugins/core/inventory/server/no-meteor/utils/publishProductToCatalog.js
```diff
@@ -9,20 +9,47 @@ export default async function publishProductToCatalog(catalogProduct, { context,
...
+
+    // if inventory info was found, add to variant
+    if (foundVariantInventory) {
+      catalogProduct.variants[variantIndex].isBackorder = foundVariantInventory.inventoryInfo.isBackorder;
+      catalogProduct.variants[variantIndex].isLowQuantity = foundVariantInventory.inventoryInfo.isLowQuantity;
+      catalogProduct.variants[variantIndex].isSoldOut = foundVariantInventory.inventoryInfo.isSoldOut;
+    }
+
+    // add inventory props to each top level option
+    variant.options.forEach((option, optionIndex) => {
```
aldeed: `options` is optional field so you'll want to wrap this in `if (variant.options)`

imports/plugins/core/versions/server/util/convert68.js
```diff
@@ -0,0 +1,296 @@
...
+
+  return results;
+}
+
+/**
+ * @param {Object} catalogProduct The catalog item to transform
+ * @param {Object} collections The catalog item to transform
+ * @returns {Object} The converted item document
+ */
+export async function convertCatalogItemVariants(catalogProduct, collections) {
```
aldeed: Add a unit test for this to prove it works.

imports/plugins/core/versions/server/util/convert68.js
```diff
@@ -0,0 +1,296 @@
...
+  const { AppSettings, Products } = collections;
+
+  const shopSettings = await AppSettings.findOne({ shopId: catalogProduct.shopId });
+
+  // By default, products that do not have inventory management enabled are sellable.
+  let canSellWithoutInventory = true;
+  if (shopSettings && typeof shopSettings.canSellVariantWithoutInventory === "boolean") {
+    canSellWithoutInventory = shopSettings.canSellVariantWithoutInventory;
+  }
```
aldeed: Could `_.get` here to compress these four lines into one.

imports/plugins/core/versions/server/util/convert68.js
```diff
@@ -0,0 +1,296 @@
...
+      remainingProductConfigurations.push(inventoryResult.productConfiguration);
+    }
+  }
+
+
+  // If no inventory info was found for some of the product configs, such as
+  // if there are no plugins providing inventory info, then use default info
+  // that allows the product to be purchased always.
+  if (remainingProductConfigurations.length > 0) {
+    const inventoryInfo = canSellWithoutInventory ? DEFAULT_SELLABLE_INFO : DEFAULT_SOLD_OUT_INFO;
```
aldeed: Seeing this made me realize that there's probably a bug here. It's an existing bug, but it will be more obvious after your changes. When the `canSellVariantWithoutInventory` setting is changed, we would need to update every catalog product where at least one variant/option does not have inventory. Because if this is `true`, then `isSoldOut` needs to be `true` for such variants and `canBackorder` needs to be `false`.

This will require some thought. It doesn't seem good to need to migrate a massive catalog every time a setting switch is flipped. But I can't think of a solution without a more fundamental shift in the way we're doing this.

imports/plugins/core/inventory/server/no-meteor/utils/publishProductToCatalog.js
```diff
@@ -9,20 +9,45 @@ export default async function publishProductToCatalog(catalogProduct, { context,
...
+      variant.isSoldOut = foundVariantInventory.inventoryInfo.isSoldOut;
+    }
+
+    // add inventory props to each top level option
+    if (variant.options) {
+      variant.options.forEach((option) => {
+        const foundOptionInventory = topVariantsAndOptionsInventory.find((inventoryInfo) => inventoryInfo.productConfiguration.productVariantId === option._id);
+
+        // if inventory info was found, add to option
+        if (foundOptionInventory) {
```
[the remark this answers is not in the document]
willopez: added

imports/plugins/core/inventory/server/no-meteor/utils/startup.js
```diff
@@ -19,24 +19,65 @@ async function updateInventoryBooleansInCatalog(context, productId) {
...
+
+  // Update inventory for the parent product and all variants and options.
+  // If no inventory information is found for a variant or option, it is not mutated.
+  const updatedVariants = [];
+  catalogProduct.product.variants.forEach((variant) => {
+    // Keep variant un-mutated by default
+    let updatedVariant = variant;
+    const foundVariantInventory = variantsOptionsInventory.find((inventoryInfo) => inventoryInfo.productConfiguration.productVariantId === variant._id);
+
+    if (foundVariantInventory) {
```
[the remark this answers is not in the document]
willopez: now throwing error

imports/plugins/core/inventory/server/no-meteor/utils/startup.js
```diff
@@ -19,24 +19,65 @@ async function updateInventoryBooleansInCatalog(context, productId) {
...
+    productConfigurations: topVariantsAndTopOptions.map((option) => ({
       isSellable: !variants.some((variant) => variant.ancestors.includes(option._id)),
       productId: option.ancestors[0],
       productVariantId: option._id
     })),
-    shopId: topVariants[0].shopId
+    shopId: topVariantsAndTopOptions[0].shopId
+  });
+
+  const catalogProduct = await Catalog.findOne({ "product.productId": productId }, { "product.variants": 1 });
```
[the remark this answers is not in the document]
willopez: removed projection

imports/plugins/core/inventory/server/no-meteor/utils/startup.js
```diff
@@ -19,24 +19,65 @@ async function updateInventoryBooleansInCatalog(context, productId) {
...
+    }
+
+    const updatedOptions = [];
+    if (updatedVariant.options) {
+      updatedVariant.options.forEach((option) => {
+        // Keep option un-mutated by default
+        let updatedOption = option;
+        const foundOptionInventory = variantsOptionsInventory.find((inventoryInfo) => inventoryInfo.productConfiguration.productVariantId === option._id);
+
+        if (foundOptionInventory) {
```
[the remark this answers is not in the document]
willopez: now throwing error

imports/plugins/core/versions/server/util/convert68.js
willopez: code consolidated

imports/plugins/core/inventory/server/no-meteor/utils/publishProductToCatalog.js
```diff
@@ -9,20 +9,45 @@ export default async function publishProductToCatalog(catalogProduct, { context,
...
+  catalogProduct.isSoldOut = topVariantsAndOptionsInventory.every(({ inventoryInfo }) => inventoryInfo.isSoldOut);
+
+  // add inventory props to each top level Variant
+  catalogProduct.variants.forEach((variant) => {
+    // attempt to find this variant's inventory info
+    const foundVariantInventory = topVariantsAndOptionsInventory.find((inventoryInfo) => inventoryInfo.productConfiguration.productVariantId === variant._id);
+
+    // if inventory info was found, add to variant
+    if (foundVariantInventory) {
+      variant.isSoldOut = foundVariantInventory.inventoryInfo.isSoldOut;
```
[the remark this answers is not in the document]
willopez: now throwing error

aldeed:
I will try to finalize this later tonight or tomorrow since Will is out.

imports/plugins/core/versions/server/util/convert68.js
aldeed: Created an issue about this here: https://github.com/reactioncommerce/reaction/issues/5333

cmbirk:
Some other context - it appears this is affecting the minPrice being set as well for the sort issue here: https://github.com/sportsdirect/reaction-plugin-pricing-engine/issues/37

We're not currently checking the `isSoldOut` value but will need to once it exists throughout. We're currently setting the minPrice on the lowest ( sometimes $1 clearance ) even on sold out options which shouldn't happen

aldeed:
This should be good to merge if all checks pass. Otherwise I will fix anything that fails later tonight.

janus-reith:
I just quicky skimmed over the changes here and might have missed something, but what I don't really get here: e.g. if there is a quantity of 1 available for a variant, and this gets bought / turns to zero somehow.

How would that `isSoldOut` field be updated? As far as I got it, this would just represent the actual status after running another publish?

aldeed:
@janus-reith Situation 1, changing the quantity as a shop operator, would go through the `updateSimpleInventory` mutation, which emits `afterInventoryUpdate` event when it is done. In `startup.js` of `inventory` plugin, there is an `afterInventoryUpdate` listener that calls the inventory `publishProductToCatalog` transformation function. Essentially, we initiate a partial republish to Catalog whenever the in-stock number changes.

Similarly, situation 2, the `simple-inventory` plugin has `afterOrderCreate` and `afterOrderCancel` event listeners. They update the "reserved" and/or "in-stock" inventory values as orders are placed/processed/canceled. And after they do this, they also emit `afterInventoryUpdate`, which starts the same partial publish process as above.

janus-reith:
@aldeed Ah great, got it, thanks for the summary. Also wanted to throw in the idea of partially updating the catalog, but this has been thought about then already.

PR #5299, opened 2019-07-09
chore: update lodash to fix snyk critical error

kieckhafer:
Impact: **minor**  
Type: **chore**

## Issue
Snyk is giving us a critical error warning based on a lodash issue. This update updates lodash to the newest version, which includes their fix.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2714,6 +2714,14 @@
       "dev": true,
       "requires": {
         "lodash": "4.17.11"
+      },
+      "dependencies": {
+        "lodash": {
+          "version": "4.17.11",
+          "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.11.tgz",
+          "integrity": "sha512-cQKh8igo5QUhZ7lg38DYWAxMvjSAKG0A8wGSVimP07SIUEK2UO+arSRKbRZWtelMtN5V0Hkwh5ryOto/SshYIg==",
+          "dev": true
+        }
       }
     },
     "@snyk/dep-graph": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -2714,6 +2714,14 @@
       "dev": true,
       "requires": {
         "lodash": "4.17.11"
+      },
+      "dependencies": {
+        "lodash": {
+          "version": "4.17.11",
```
spencern: Should this have been bumped as well?

mikemurray: is that a direct dependency or a sub dependency of another package?

mikemurray: Either way, snyk seems happy, so I see this as an absolute win.

kieckhafer:
Looks like `snyk` needs to update something on their end first, saying `14.17.12` is a good version.

package-lock.json
kieckhafer: Ironically this is in a `snyk` dependency... so hopefully they'll update it soon: `@snyk/composer-lockfile-parser`

package.json
```diff
@@ -78,7 +78,7 @@
     "jquery-i18next": "^1.2.1",
     "later": "^1.2.0",
     "libphonenumber-js": "1.4.2",
-    "lodash": "^4.17.11",
+    "lodash": "^4.17.12",
```
spencern: Looks like `4.17.13` is out now as well to fix a few issues that landed in `4.17.12` since we haven't merged yet, let's go ahead and bump to that version.

kieckhafer: updated.

PR #5158, opened 2019-05-01
refactor: update Reaction Admin Orders panel

kieckhafer:
Resolves #5157   
Impact: **major**  
Type: **refactor**

All of the UI work, queries, mutations and tests have already been reviewed and merged in the PR's listed below. Therefore an in-depth test of all **_NEW_** functionality / code here **_is not_** the goal of this PR. The goal of this particular PR is to make this new Orders UI the official Orders UI ([seen in this Routing change](https://github.com/reactioncommerce/reaction/pull/5158/commits/52d7e81367df913d749ec06f527df7ea18d33de4)), and to remove the old UI, meteor methods, and related files.

Changes already implemented:

- #5240 
- #5236
- #5235
- #5226
- #5203
- #5202
- #5201
- #5200
- #5266
- #5279
- #5302
- #5352 
- #5354 
- #5405 

## Issue
Operator UI Orders panel currently 
- uses our old Meteor methods to preform necessary functions, which don't include new mutations to split fulfillment groups and move items between them
- isn't updated to easily use multiple fulfillment methods, it currently combines all products into a single group

## Solution
Update the Orders admin to use GraphQL for all data, instead of the existing Meteor methods. In addition, update the panel to preform actions against multiple fulfillment groups, instead of a single combined Order for all groups.

## Breaking changes
Meteor methods will no longer work. All orders functionality is now GQL.

## Testing
- Add orderStatusLabels to the `Shops` collection shop object. This will provide translations for all the `Chip` style order status badges in the Orders Fulfillment UI.
```
"orderStatusLabels": {
    "new" : [ 
        {
            "language" : "en",
            "label" : "Order received"
        }, 
        {
            "language" : "es",
            "label" : "Recibido"
        }
    ],
    "coreOrderWorkflow/processing" : [ 
        {
            "language" : "en",
            "label" : "Order received"
        }, 
        {
            "language" : "es",
            "label" : "Recibido"
        }
    ],
    "coreOrderWorkflow/completed" : [ 
        {
            "language" : "en",
            "label" : "Shipped"
        }, 
        {
            "language" : "es",
            "label" : "Enviado"
        }
    ],
    "coreOrderWorkflow/canceled" : [ 
        {
            "language" : "en",
            "label" : "Canceled"
        }, 
        {
            "language" : "es",
            "label" : "Cancelado"
        }
    ],
    "coreOrderWorkflow/picked" : [ 
        {
            "language" : "en",
            "label" : "Picked"
        }
    ],
[...]

PR #5302, opened 2019-07-10
refactor: fulfillment section of Orders 2.0

kieckhafer:
Fixes the fulfillment portion of #5157 
Impact: **major**  
Type: **feature|refactor**

## Issue
This is one part of a bigger ticket that is refactoring the whole Order Fulfillment UI. You will only be testing the Fulfillment section during this PR, the other pieces have already been tested.

Order_Details_for_order_reference__NLuZFydBS6oHLS4Zb

You will be testing:
1) The `Cancel order` button in the top App Bar
2) The `Cancel group` button in one or more fulfillment groups.
3) The `Update group status` button in one or more fulfillment groups.
4) The `Tracking number` field in one or more fulfillment groups.

## Breaking changes
None. this is done in the background, on a separate URL, so existing code isn't affected.

## Testing
- Add orderStatusLabels to the `Shops` collection shop object. This will provide translations for all the `Chip` style order status badges in the Orders Fulfillment UI.
```
"orderStatusLabels": {
    "new" : [ 
        {
            "language" : "en",
            "label" : "Order received"
        }, 
        {
            "language" : "es",
            "label" : "Recibido"
        }
    ],
        {
            "language" : "en",
            "label" : "Order received"
        }, 
        {
            "language" : "es",
            "label" : "Recibido"
        }
    ],
        {
            "language" : "en",
            "label" : "Shipped"
        }, 
        {
            "language" : "es",
            "label" : "Enviado"
        }
    ],
        {
            "language" : "en",
            "label" : "Canceled"
        }, 
        {
            "language" : "es",
            "label" : "Cancelado"
        }
    ],
    "coreOrderWorkflow/picked" : [ 
        {
            "language" : "en",
            "label" : "Picked"
        }
    ],
    "coreOrderWorkflow/packed" : [ 
        {
            "language" : "en",
            "label" : "packed"
        }
    ],
    "coreOrderWorkflow/shipped" : [ 
        {
            "language" : "en",
            "label" : "shipped"
        }
    ]
}
```
- Create multiple (~5) orders to perform various actions on. Some of these orders should have multiple fulfillment groups ([see this PR for instructions](https://github.com/reactioncommerce/reaction/pull/5027) on how to create multiple groups). From the Orders table, click on an order to be directed to the single order page. Once there, change the URL from `/orders/` to `/orders-beta/` to see the new UI. 

Reaction

Test the following workflows:

- [ ] Create a new user, add them to the Shop Manager Group, and give the group `Order fulfillment` permissions. Test the following with those permission.

Accounts

- [x] On an order with multiple fulfillment groups, use the `Cancel order` button in the toolbar to cancel all fulfillment groups. Do the same on an order with a single fulfillment group. See that the `Cancel` button disappears when an order has already been cancelled.

- [x] On an order with multiple fulfillment groups, use the `Cancel fulfillment group` button individually inside each group to cancel the group. Cancel all the groups this way, and see that the overall order is cancelled once all the individual groups have been cancelled. Do the same on an order with a single fulfillment group. See that the `Cancel` button disappears when an order has already been cancelled.

- [x] On an order with multiple fulfillment groups, add a tracking number on each group and see they update correctly. Then, click the tracking number to edit it, and update the number. See that it updates correctly.

- [x] On an order with multiple fulfillment groups, change the `Order status` to various stages, both in the usual order (picked -> packed -> shipped), and also in random orders. Make sure the individual group updates as it should, and no other groups update. Do the same on an order with a single fulfillment group.

- [x] Change the user permissions so they only have `Order View` permissions. See that you can still see the order, but all the buttons are gone and you cannot preform and fulfillment actions.


imports/plugins/core/orders/client/components/OrderCardFulfillmentGroupItem.js
```diff
@@ -33,33 +30,33 @@ class OrderCardFulfillmentGroupItem extends Component {
 
   render() {
```
willopez: Looks like this class can be converted to a function, we want to move towards hooks.

imports/plugins/core/orders/client/components/OrderCardFulfillmentGroupStatusButton.js
```diff
@@ -0,0 +1,147 @@
...
+import OutlinedInput from "@material-ui/core/OutlinedInput";
+import Select from "@material-ui/core/Select";
+import { i18next, Reaction } from "/client/api";
+import ConfirmButton from "/imports/client/ui/components/ConfirmButton";
+import updateOrderFulfillmentGroupMutation from "../graphql/mutations/updateOrderFulfillmentGroup";
+
+const styles = (theme) => ({
+  formControl: {
+    margin: theme.spacing.unit,
+    minWidth: 120
```
willopez: We should add a minWidth to the MUI theme, and use it across all dialogs for consistence.

imports/plugins/core/orders/client/components/OrderCardFulfillmentGroupStatusButton.js
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  // TODO: When MUI 4.x is implemented, change this to `SplitButton`
+  // instead of using a select dropdown in a popup
+  // https://material-ui.com/components/buttons/#split-button
+  render() {
+    const hasPermission = Reaction.hasPermission(["reaction-orders", "order/fulfillment"], Reaction.getUserId(), Reaction.getShopId());
+    const { classes, fulfillmentGroup } = this.props;
+    const canUpdateFulfillmentStatus = (fulfillmentGroup.status !== "coreOrderWorkflow/canceled");
+    const options = [
```
willopez: Are these options used anywhere else? or will they be used in the future in another component? if so, they should like in a `orderStatuses` helper.

imports/plugins/core/orders/client/components/OrderCardFulfillmentGroupTrackingNumber.js
```diff
@@ -0,0 +1,147 @@
...
+import Grid from "@material-ui/core/Grid";
+import Link from "@material-ui/core/Link";
+import ErrorsBlock from "@reactioncommerce/components/ErrorsBlock/v1";
+import Field from "@reactioncommerce/components/Field/v1";
+import TextInput from "@reactioncommerce/components/TextInput/v1";
+import { i18next, Reaction } from "/client/api";
+import updateOrderFulfillmentGroupMutation from "../graphql/mutations/updateOrderFulfillmentGroup";
+
+
+const PaddedField = styled(Field)`
```
willopez: let's use MUI's styling solution, we want to consolidate to use one system. Remove usage of `styled-components`

imports/plugins/core/orders/client/components/OrderCardFulfillmentGroups.js
```diff
@@ -0,0 +1,249 @@
...
+import OrderCardFulfillmentGroupTrackingNumber from "./OrderCardFulfillmentGroupTrackingNumber";
+import OrderCardFulfillmentGroupStatusButton from "./OrderCardFulfillmentGroupStatusButton";
+import OrderStatusChip from "./OrderStatusChip";
+
+const styles = (theme) => ({
+  fulfillmentGroupHeader: {
+    marginBottom: theme.spacing.unit * 4
+  },
+  verticalDivider: {
+    backgroundColor: "#e6e6e6",
```
willopez: is there a color in the theme that can be used here? Let's stay away from one off styles.

imports/plugins/core/orders/client/components/OrderCardFulfillmentGroupItem.js
kieckhafer: I have this being converted to a class in #5158 along with some other components from other PR's related that need to be converted. If really needed, I can move that over here, but would prefer to keep it where it is and do the class to function conversion in that one place.

imports/plugins/core/orders/client/components/OrderCardFulfillmentGroupStatusButton.js
kieckhafer: Currently, no, because we want to restructure all these workflows in the near future anyway, so this is the only place they are used.

imports/plugins/core/orders/client/components/OrderCardFulfillmentGroupStatusButton.js
kieckhafer: This is a min-width for that specific form field, based on the data inside the form, not for the dialog itself.

imports/plugins/core/orders/client/components/OrderCardFulfillmentGroupTrackingNumber.js
kieckhafer: Fixed

imports/plugins/core/orders/client/components/OrderCardFulfillmentGroups.js
kieckhafer: Fixed

Issue #5267, opened 2019-07-03
[ESLINT] audit rule: valid-jsdoc


kieckhafer:
As of July 4, 2019, we have 226 eslint warnings for the `valid-jsdoc` rule.

This ticket is to:
1. Fix all these `eslint` warnings. This SHOULD mean actually fixing them, but could also mean adding an `eslint-ignore`, if applicable.
1. Change the `valid-jsdoc` rule from a `warning` to an `error` so we do not introduce new warnings in the future.

PR #5309, opened 2019-07-11
chore: update js-doc throughout the app, batch 1

### LINKED ISSUE #5269: [ESLINT] audit rule: require-jsdoc
As of July 4, 2019, we have 110 eslint warnings for the `require-jsdoc` rule.

This ticket is to:
1. Fix all these `eslint` warnings. This SHOULD mean actually fixing them, but could also mean adding an `require-jsdoc`, if applicable.
1. Change the `id-length` rule from a `warning` to an `error` so we do not introduce new warnings in the future.

kieckhafer:
Resolves parts of #5269 & #5267   
Impact: **major**  
Type: **docs|chore**

## Issue
We are missing entire `jsdoc` blocks in some instances, and param / return / description lines of `jsdoc` in other instances, along with other `jsdoc` errors, all throughout Reaction. This is the biggest contributor to our `eslint` warnings.

## Solution
Add / fix `jsdoc` where necessary.
Since there are so many of these warnings, I'm going to split it up into multiple PR's to keep the file count low for easier viewing / less conflicts with open branches.

## Breaking changes
None

## Testing
1. Start the app
1. Make sure it runs
1. All changes are documentation changes, there should be no effect on the app in general

### DIFF
--- a/imports/plugins/core/shop/server/resolvers/Query/primaryShop.js
+++ b/imports/plugins/core/shop/server/resolvers/Query/primaryShop.js
@@ -1,10 +1,10 @@
 /**
- * @name "Query.primaryShop"
+ * @name Query/primaryShop
  * @method
  * @memberof Shop/GraphQL
  * @summary Gets the primary shop
- * @param {Object} parentObject - unused
- * @param {Object} args - unused
+ * @param {Object} _ - unused
+ * @param {Object} __ - unused
  * @param {Object} context - an object containing the per-request state
  * @return {Promise<Object>} The shop, based on the domain in ROOT_URL
  */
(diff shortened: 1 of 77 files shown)


imports/plugins/core/shop/server/resolvers/Query/primaryShop.js
```diff
@@ -1,10 +1,10 @@
 /**
- * @name "Query.primaryShop"
+ * @name Query/primaryShop
  * @method
  * @memberof Shop/GraphQL
  * @summary Gets the primary shop
- * @param {Object} parentObject - unused
- * @param {Object} args - unused
+ * @param {Object} _ - unused
```
bt3gl: Hi, curious what's the meaning of this change (and what does having "_" mean).

kieckhafer: @bt3gl we have all our queries / mutations set up to accept these three params: `parentResult`, `inputArgs`, and `context`. In a lot of cases, we don't need to pass in `parentResult` or any `args`, so we use `_` and `__` as placeholders for these fields.

PR #5323, opened 2019-07-15
(docs): update how to get reaction version

mpaktiti:
Resolves -
Impact: **minor**  
Type: **docs**

## Issue
Our GitHub Issue Template was asking users to check their Reaction version by running `reaction -v`. The Reaction CLI has been deprecated as of Reaction v2, therefore the template should instruct users to use `grep version package.json` instead.

## Solution
Instruct users to run `grep version package.json` in order to check their reaction version.

## Breaking changes
None

## Testing
Check the template offers the correct instructions (i.e. `grep version package.json`).


mpaktiti:
Created against wrong branch. See https://github.com/reactioncommerce/reaction/pull/5324 instead.

PR #5334, opened 2019-07-18
chore: fix consistent-return eslint warnings

### LINKED ISSUE #5272: [ESLINT] audit rule: consistent-return
We have multiple eslint warnings for the `consistent-return` rule.

This ticket is to:
1. Fix all these `eslint` warnings. This SHOULD mean actually fixing them, but could also mean adding an `eslint-disable-line`, if applicable.
1. Change the `consistent-return` rule from a `warning` to an `error` so we do not introduce new warnings in the future.

kieckhafer:
Resolves #5272   
Impact: **breaking|minor**  
Type: **chore**

## Issue
We currently have many `consistent-return` eslint warnings.

## Solution
Fix (or disable) these warnings, and then remove the `warn` override in our package.json eslint rules to make `consistent-return` an error so new violations are not introduced.

Most of the changes add `null` as a return. Some return a boolean where deemed appropriate.

## Breaking changes
`return null` has been added in a lot of places. While this shouldn't affect anything that is updated, please not that there is no longer no return / undefined return in many functions.

## Testing
1. run `eslint`
1. see there are no longer any `consistent-return` warnings
1. see that the override for this rule was removed, so in the future all issues will be errors

## Notes
- There are a few `jsdoc`s added in the PR as well, as I tried to add them if I saw missing jsdoc in a file I was updating.

### DIFF
--- a/imports/plugins/core/files/server/no-meteor/setUpFileCollections.js
+++ b/imports/plugins/core/files/server/no-meteor/setUpFileCollections.js
@@ -71,7 +71,7 @@ export default function setUpFileCollections({
       mongodb,
       name,
       async transformWrite(fileRecord) {
-        if (!transform) return;
+        if (!transform) return null;
 
         const { size, mod, format, type } = transform;
 
@@ -81,7 +81,7 @@ export default function setUpFileCollections({
         fileRecord.extension(format, { store: name });
 
         // resizing image, adding mod, setting output format
-        sharp().resize(size, size)[mod]().toFormat(format);
+        return sharp().resize(size, size)[mod]().toFormat(format);
       }
     })
   );
--- a/imports/plugins/core/files/server/no-meteor/setUpFileCollections.js
+++ b/imports/plugins/core/files/server/no-meteor/setUpFileCollections.js
@@ -81,7 +81,7 @@ export default function setUpFileCollections({
         fileRecord.extension(format, { store: name });
 
         // resizing image, adding mod, setting output format
-        return sharp().resize(size, size)[mod]().toFormat(format);
+        sharp().resize(size, size)[mod]().toFormat(format);
       }
     })
   );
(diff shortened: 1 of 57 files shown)


imports/plugins/core/files/server/no-meteor/setUpFileCollections.js
```diff
@@ -81,7 +81,7 @@ export default function setUpFileCollections({
         fileRecord.extension(format, { store: name });
 
         // resizing image, adding mod, setting output format
-        return sharp().resize(size, size)[mod]().toFormat(format);
+        sharp().resize(size, size)[mod]().toFormat(format);
```
aldeed: This needs to return a ReadWrite stream or transformation won't work. See https://github.com/reactioncommerce/reaction-file-collections#create-stores

Looking at [the package code](https://github.com/reactioncommerce/reaction-file-collections/blob/83078b95db6bfb1c36cd2d4bb80ba8ed4d45fd35/packages/file-collections-sa-base/src/StorageAdapter.js#L142-L147), it only does a truthy check, so changing the first line here to `return null` is fine.

imports/plugins/included/marketplace/client/templates/stripeConnectSignupButton/stripeConnectSignupButton.js
```diff
@@ -61,6 +61,6 @@ Template.stripeConnectSignupButton.events({
 
     const stripeConnectAuthorizeUrl = `https://connect.stripe.com/oauth/authorize?response_type=code&state=${shopId}&client_id=${clientId}&scope=read_write`;
     const autofillParams = `&stripe_user[email]=${defaultEmailAddress}&stripe_user[country]=${country}&stripe_user[phone_number]=${phoneNumber}&stripe_user[business_name]=${businessName}&stripe_user[street_address]=${streetAddress}&stripe_user[city]=${city}&stripe_user[state]=${state}&stripe_user[zip]=${zip}&stripe_user[first_name]=${firstName}&stripe_user[last_name]=${lastName}`; // eslint-disable-line max-len
-    window.open(stripeConnectAuthorizeUrl + autofillParams, "_blank");
+    return window.open(stripeConnectAuthorizeUrl + autofillParams, "_blank");
```
aldeed: This probably won't cause issues, but I'd lean more toward changing all of the above from `return Alerts.toast` to

```js
Alerts.toast(...);
return;
```

Returning the result of another function call is always prone to future errors since any change to their API will change your API without your being aware of it.

PR #5350, opened 2019-07-22
(fix) Make catalog indexes unique for product

### LINKED ISSUE #5349: Top-level product index in catalog should be unique
## Issue Description
It is critical for proper functioning of the Catalog collection that there only be on record in the Catalog for each top-level product ( `type: "simple"`). This index ensures that this is always the case. It should theoretically make queries faster since it will know that it only needs to search one match.

zenweasel:
Resolves #5349
Impact: **breaking|minor**  
Type: **bugfix**

## Issue
The system current allows for more than one Catalog record per top-level product. This causes all sort of problems and should never happen

## Solution
Creates a unique index on product fields where there should always be only one of them. I guess this got overlooked when the Catalog collection was originally created.

## Breaking changes
In order for this change to take place you must drop the existing indexes on `product._id`and `product.productId`. We currently have no system for modifying indexes in code.

## Testing
1. Drop the indexes on `Catalog.product._id, Catalog.product.productId and Catalog.product.slug`
1. Start the system
1. Verify that the new indexes are created as unique indexes.

### DIFF
--- a/imports/plugins/core/catalog/server/no-meteor/register.js
+++ b/imports/plugins/core/catalog/server/no-meteor/register.js
@@ -25,7 +25,7 @@ export default async function register(app) {
           [{ shopId: 1 }],
           [{ "product._id": 1 }, { unique: true }],
           [{ "product.productId": 1 }, { unique: true }],
-          [{ "product.slug": 1 }, { unique: true }],
+          [{ "product.slug": 1 }],
           [{ "product.tagIds": 1 }]
         ]
       }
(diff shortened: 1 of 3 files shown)


zenweasel:
@aldeed I am also not sure how to broadcast this breaking change. If you just run this it's going to throw an error. I looked at making the `createIndex` function automatically drop and replaces indexes that have been modified but thinking about it I don't think that's a good approach because dropping an index is a fully blocking operation on a very busy collection, so you probably want to have some control of it.

zenweasel:
@aldeed Also, tests are fixed.

aldeed:
@zenweasel Need signoff on your last commit, but otherwise ok to merge this.

zenweasel:
@aldeed DCO issue fixed now

PR #5352, opened 2019-07-22
feat: add refunds GraphQL query

kieckhafer:
Part of #5158 
Impact: **minor**  
Type: **feature**

A new `listRefunds` query allows you to query refunds of payments on a particular order. This was previously a meteor only method. The meteor method will be removed with #5158.

## Breaking changes
None

## Testing
1. Place an order using both the Stripe payment method and IOU method
1. Create a refund for each type of payment
1. Use the new `listRefunds` query to see the refunds

```
listRefunds(orderId: {ORDER-ID}, shopId: {SHOP-ID}) {
   _id
    amount
    ...otherFieldsYouWant
  }
```

I've added tests here to test the permission / requirements of variables, but did not add tests to check a result, as the payment method has it's own tests for this element: https://github.com/reactioncommerce/reaction/blob/feat-kieckhafer-listRefundsGQLQuery/imports/plugins/included/payments-stripe/server/no-meteor/util/stripeListRefunds.test.js#L10


kieckhafer:
@aldeed I wasn't sure if this should go into `Payments` or `Orders`. I put it in `Orders` because each specific payment method has it's own way of calculating the refunds, and this is just a list that is displayed on the `Orders` operator UI.

What do you think about it's location?

aldeed:
@kieckhafer I think I have the same thoughts as you, could go either place. Since payments are currently coupled with orders, leaving it in orders seems fine.

I'm assigning @focusaurus to do the full code review here, but my only comment is that for GraphQL, we try to omit words like `get` and `list` from the query names. That's pretty standard GQL best practice. So I'd go with just `refunds`.

Also keep in mind that something like this fits pretty well into the payment resolver. In other words, rather than (or in addition to) `Query.refunds` resolver, you could have a `Payment.refunds` resolver. This is nice because you'd already know order ID and payment ID in that resolver, plus you could avoid doing an extra query in Catalyst.

```graphql
{
  order {
    payments {
      refunds {
        amount
      }
    }
  }
}
```

Or for a combined list for all payments, it could be similarly done as `Order.refunds` resolver. There can be several resolvers that all call the same query function giving clients various ways to get the same info.

kieckhafer:
@aldeed makes sense. I took the pattern of `listRefunds` from the plugins that do it, we currently have `stripleListRefunds`. Is it OK if it's not the first word?

imports/plugins/core/orders/server/no-meteor/queries/refunds.js
```diff
@@ -0,0 +1,50 @@
+import ReactionError from "@reactioncommerce/reaction-error";
+import { getPaymentMethodConfigByName } from "/imports/plugins/core/payments/server/no-meteor/registration";
+import { getOrderQuery } from "../util/getOrderQuery";
+
+/**
+ * @name refunds
+ * @method
```
focusaurus: How/why are we using `@method` here? I see we're using it all over the place, it's just not clear from the docs what it's supposed to convey or how it renders in the doc site.

imports/plugins/core/orders/server/no-meteor/queries/refunds.js
```diff
@@ -0,0 +1,50 @@
...
+  if (!order) {
+    throw new ReactionError("not-found", "Order not found");
+  }
+
+  const paymentRefunds = [];
+
+  if (Array.isArray(order.payments)) {
+    for (const payment of order.payments) {
+      /* eslint-disable no-await-in-loop */
+      const shopRefunds = await getPaymentMethodConfigByName(payment.name).functions.listRefunds(context, payment);
```
focusaurus: Can we make this loop run in parallel by creating the promises without awaiting them then `await Promise.all` after the loop?

imports/plugins/core/orders/server/no-meteor/queries/refunds.test.js
```diff
@@ -0,0 +1,51 @@
...
+  mockContext.userHasPermission.mockReturnValueOnce(false);
+  mockContext.collections.Orders.findOne.mockReturnValueOnce(Promise.resolve(order));
+
+  await expect(refunds(mockContext, {
+    orderId: order._id,
+    shopId: order.shopId
+  })).rejects.toThrowErrorMatchingSnapshot();
+
+  expect(mockContext.userHasPermission).toHaveBeenCalledWith(["orders", "order/fulfillment", "order/view"], "SHOP_ID");
+});
```
focusaurus: Can we add a base success test case?

imports/plugins/core/orders/server/no-meteor/resolvers/Refund/createdAt.js
```diff
@@ -0,0 +1,17 @@
...
+ * @param {Object} refund - result of the parent resolver, which is a refund object in GraphQL schema format
+ * @return {String} A datetime string
+ */
+export default function createdAt(refund) {
+  // graphql-iso-data won't accept the Unix timestamp that is provided by Stripe
+  // we need to use moment to convert it to a readable format
+  const formattedDate = moment(refund.created).format();
+
+  return formattedDate;
+}
```
focusaurus: Can we add a quick unit test for this? Almost trivial but easy to test. Any concerns about stripe specifics in a generic plugin?

imports/plugins/included/marketplace/server/no-meteor/util/stripeListRefunds.js
```diff
@@ -25,11 +25,13 @@ export default async function stripeListRefunds(context, paymentMethod) {
   if (refundListResults && refundListResults.data) {
     for (const refund of refundListResults.data) {
       result.push({
-        type: refund.object,
+        _id: refund.id,
         amount: refund.amount / 100,
```
focusaurus: I know this line didn't change but are we USD-only in stripe at this point? Is it safe to do this conversion regardless of `refund.currency`?

imports/plugins/core/orders/server/no-meteor/resolvers/Refund/createdAt.js
kieckhafer: We could do it in the stripe plugin instead, however the `moment.format()` will accept other date formats as well and spit them out the same way, so although my comment and the reasoning behind this is because of stripe, this is really just making a consistent format for all dates that we get back.

imports/plugins/core/orders/server/no-meteor/queries/refunds.js
kieckhafer: I think the way I'm doing it has been the pattern we've been using with the non-meteor code. In Meteor code we were using the `Promise.await`, but in all our non-meteor code we are doing the await inside the loop. @aldeed is my assessment correct? is there a reasoning behind why we do it with `await` in the loop more recently?

imports/plugins/included/marketplace/server/no-meteor/util/stripeListRefunds.js
kieckhafer: Great question. It's probably not safe to do in all currencies: https://stripe.com/docs/currencies#zero-decimal

@spencern are we USD-only for Stripe?

imports/plugins/core/orders/server/no-meteor/queries/refunds.test.js
kieckhafer: added

imports/plugins/included/marketplace/server/no-meteor/util/stripeListRefunds.js
spencern: We should not assume USD only.

@kieckhafer will you create a ticket to audit the stripe integration to see where we have incompatibility with other currencies?

kieckhafer: Issue created: https://github.com/reactioncommerce/reaction/issues/5383

kieckhafer:
@focusaurus I added a new query here, `refundsByPaymentId`, and also added a couple resolvers so that you can query `refunds` on an order, and also `refunds` per payment, so there is some additional functionality to check since your last review.

```
query {
  orderByReferenceId(id: "ORDER-ID", shopId: "SHOP-IS") {
    _id
    payments {
      _id
      method {
        name
      }
      refunds {
        amount {
          displayAmount
        }
      }
    }
    refunds {
       amount {
        displayAmount
      }
      paymentDisplayName
    }
    shop {
      _id
    }
  }
}
```

imports/plugins/core/orders/server/no-meteor/resolvers/Order/payments.js
```diff
@@ -0,0 +1,34 @@
...
+ * @param {Object} context - an object containing the per-request state
+ * @param {Object} order order object refunds would be applied to
+ * @return {Promise<Object[]>} Promise that resolves with array of payment objects
+ */
+export default async function payments(context, order) {
+  if (Array.isArray(order.payments)) {
+    return order.payments.map(async (payment) => {
+      const xformPayment = xformOrderPayment(payment);
+
+      const refunds = await context.queries.refundsByPaymentId(context, {
```
kieckhafer: The reason I did this here, instead of inside the `xformOrderPayment` function, is I wanted to keep all the refunds xforms and data inside this orders package, instead of in the `core` package where `xformOrderPayment` is.

This could be moved into into  `xformOrderPayment` if that seems like a better place.

imports/plugins/core/orders/server/no-meteor/resolvers/Query/refundsByPaymentId.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @name Query.refunds
+ * @method
+ * @memberof Order/GraphQL
+ * @summary Get refunds applied to an order
+ * @param {Object} _ - unused
+ * @param {ConnectionArgs} args - An object of all arguments that were sent by the client
+ * @param {String} args.orderId - ID of the order
+ * @param {String} args.orderId - ID of the payment
```
focusaurus: Copypasta: `args.paymentId`

kieckhafer: 🍝

imports/plugins/core/orders/server/no-meteor/queries/refunds.js
kieckhafer: This is a habit carried over from using `meteor` methods. For certain methods, they wouldn't appear in jsdoc without this. We can probably stop using it, unless anyone has any objections / reasons why we do still use it? @machikoyasuda @mikemurray any ideas?

machikoyasuda: As long as the method's docs appear in http://api.docs.reactioncommerce.com/, we're good 👍 

Back in the old days, some Meteor methods would not appear w/o the `@method`. You can test it out locally by running `npm run docs` and checking the HTML files that get created in your personal `tmp/reaction-docs` folder to see if that method was rendered.

imports/plugins/core/orders/server/no-meteor/queries/refunds.js
aldeed: @kieckhafer Actually the pattern is to `await Promise.all` whenever parallel is fine. In many places, parallel is not fine because the order of execution matters or it's transformation chain, and in those instances we `await` in the for loop.

As I typed this it started to sound familiar, and then I remembered this: https://docs.reactioncommerce.com/docs/dev-how-do-i#loop-over-async-function-results

aldeed: In this specific case, it seems to me that parallel would be good.

imports/plugins/core/orders/server/no-meteor/resolvers/Refund/createdAt.js
```diff
@@ -0,0 +1,17 @@
...
+ * @method
+ * @memberof Order/GraphQL
+ * @summary converts the `createdAt` prop from a unix timestring to a format readable by `graphql-iso-data`
+ * @param {Object} refund - result of the parent resolver, which is a refund object in GraphQL schema format
+ * @return {String} A datetime string
+ */
+export default function createdAt(refund) {
+  // graphql-iso-data won't accept the Unix timestamp that is provided by Stripe
+  // we need to use moment to convert it to a readable format
+  const formattedDate = moment(refund.created).format();
```
aldeed: I believe `created` is MS so you can do `const formattedDate = new Date(refund.created);` without needing Moment.

imports/plugins/core/orders/server/no-meteor/resolvers/Refund/index.js
```diff
@@ -0,0 +1,10 @@
+import amount from "./amount";
+import createdAt from "./createdAt";
+
+export default {
+  _id: (node) => node._id,
+  amount,
+  createdAt,
+  paymentDisplayName: (node) => node.paymentDisplayName,
+  reason: (node) => node.reason
```
aldeed: Neither of these last two props should be necessary. If GraphQL doesn't find a resolver for it, it will just use the object property with that name, exactly what you're doing here.

imports/plugins/core/orders/server/no-meteor/resolvers/Refund/index.js
```diff
@@ -0,0 +1,10 @@
+import amount from "./amount";
+import createdAt from "./createdAt";
+
+export default {
+  _id: (node) => node._id,
```
aldeed: You need to encode this ID here. May need a new transform for "reaction/refund" namespace.

You should also add `paymentId` here and encode that (and add it to GraphQL schema).

imports/plugins/core/orders/server/no-meteor/schemas/schema.graphql
```diff
@@ -395,6 +410,27 @@ type OrderNote {
...
+  createdAt: DateTime!
+
+  "The display name of the payment refunded to"
+  paymentDisplayName: String!
+
+  "The reason for the refund"
+  reason: String
+
+  "The type of refund"
+  type: RefundType
```
aldeed: Do we use/show this anywhere? It's unclear to me what it's for so maybe we should omit it from the GraphQL for now?

imports/plugins/core/orders/server/no-meteor/resolvers/Refund/index.js
kieckhafer: Does that mean `_id` also isn't necessary?

kieckhafer: nevermind, commented before I saw your other comment below.

imports/plugins/core/orders/server/no-meteor/schemas/schema.graphql
kieckhafer: No, we don't use it, will remove.

kieckhafer:
@aldeed all comments addressed, ready for another check

imports/plugins/core/orders/server/no-meteor/queries/refunds.js
kieckhafer: A little confused about how to go about this.

I need to use the results from line 38 on line 40, where would I `await Promise.all`? If I change line 36 to a map, I still need the result from the await inside the map, I can't wait until after it to resolve the promises.

focusaurus: Here's what I think I'd do. It's not going to win any beauty pageants but I think it works in parallel and gets you the combination you need of `shopRefunds` and the corresponding `payment` in the same scope. This is from `const paymentRefunds = [];` to the end of the function body.

```js
  const paymentRefunds = [];

  if (Array.isArray(order.payments)) {
    const shopRefundsWithPaymentPromises = order.payments.map((payment) =>
      getPaymentMethodConfigByName(payment.name)
        .functions.listRefunds(context, payment)
        .then((shopRefunds) => shopRefunds, payment));
    // This should give us an arary of objects with properties
    // `shopRefunds` and the corresponding `payment`
    const shopRefundsWithPayments = await Promise.all(shopRefundsWithPaymentPromises);
    shopRefundsWithPayments.forEach(({ shopRefunds, payment }) => {
      const shopRefundsWithPaymentId = shopRefunds.map((shopRefund) => ({
        ...shopRefund,
        paymentId: payment._id,
        paymentDisplayName: payment.displayName
      }));
      paymentRefunds.push(...shopRefundsWithPaymentId);
    });
  }

  return paymentRefunds;

```

kieckhafer: Got it, this makes sense. I wasn't wrapping my head around that second loop in the `forEach`.

kieckhafer:
@aldeed comments addressed, ready for a final look

PR #5354, opened 2019-07-23
feat: add createRefund GraphQL mutation 

kieckhafer:
Part of #5158 
Impact: **minor**  
Type: **feature**

A new `createRefund` mutation allows you to create a refund for a particular payment on an order. This was previously a meteor only method. The meteor method will be removed with #5158.

## Breaking changes
None

## Testing
1. Place an order
1. Use the following mutation to create a refund on said order
1. Try to refund more than the charge was, and see that only the remaining amout was refunded

```
createRefund(
  amount: {AMOUNT}!,
  orderId: {ORDER-ID}!,
  paymentId: {PAYMENT-ID}!,
  reason: {Reason}) {
    order {
      {FIELDS-TO-RETREIVE}
    }
  }
```

`amount`, `orderId`, and `paymentId` are required. `reason` is optional. It's set up as a `String` to be used with multiple payment providers, however for `Stripe`, which is our only provider at the moment, they only accept 3 values for reason: `duplicate`, `fraudulent`, or `requested_by_customer`, so be sure to use one of those values if testing with Stripe. This limit will be built into the Operator UI if Stripe is the payment method. From their docs:

> String indicating the reason for the refund. If set, possible values are duplicate, fraudulent, and requested_by_customer. If you believe the charge to be fraudulent, specifying fraudulent as the reason will add the associated card and email to your block lists, and will also help us improve our fraud detection algorithms.

I've added tests here to test the permission / requirements of variables, but did not add tests to check a result, as the payment method has it's own tests for this element: https://github.com/reactioncommerce/reaction/blob/feat-kieckhafer-listRefundsGQLQuery/imports/plugins/included/payments-stripe/server/no-meteor/util/stripeListRefunds.test.js#L10


imports/plugins/core/orders/server/no-meteor/mutations/createRefund.js
```diff
@@ -0,0 +1,111 @@
+import SimpleSchema from "simpl-schema";
+import accounting from "accounting-js";
+import Logger from "@reactioncommerce/logger";
+import ReactionError from "@reactioncommerce/reaction-error";
+import { getPaymentMethodConfigByName } from "/imports/plugins/core/payments/server/no-meteor/registration";
+import sendOrderEmail from "../util/sendOrderEmail";
+
+const inputSchema = new SimpleSchema({
+  amount: Number,
```
focusaurus: Does it make sense to set a minimum of 0 for amount?

imports/plugins/core/orders/server/no-meteor/mutations/createRefund.test.js
```diff
@@ -0,0 +1,119 @@
...
+
+test("throws if amount isn't supplied", async () => {
+  await expect(createRefund(mockContext, {
+    amount: null,
+    orderId: fakeOrder._id,
+    paymentId: fakeOrder.payments[0]._id,
+    reason: "Customer was unsatisfied with purchase"
+  })).rejects.toThrowErrorMatchingSnapshot();
+});
+
```
focusaurus: Maybe test negative amount here.

imports/plugins/core/orders/server/no-meteor/mutations/createRefund.js
kieckhafer: Maybe not `0`, but maybe `0.01`, if it's possible to do that. I can't think of a circumstance where there is s `0` refund.

kieckhafer:
@focusaurus I've added a `min` check at `0.01`, not 0. I can't think of a scenario where there would be a refund of `0`. If you think it should be `0` instead, I can change it to that.

Also added a test for this.

Ready for another look.

imports/plugins/core/orders/server/no-meteor/mutations/createRefund.js
```diff
@@ -0,0 +1,114 @@
...
+import accounting from "accounting-js";
+import Logger from "@reactioncommerce/logger";
+import ReactionError from "@reactioncommerce/reaction-error";
+import { getPaymentMethodConfigByName } from "/imports/plugins/core/payments/server/no-meteor/registration";
+import sendOrderEmail from "../util/sendOrderEmail";
+
+const inputSchema = new SimpleSchema({
+  amount: {
+    type: Number,
+    min: 0.01
```
focusaurus: Talked with Eric D about this. Thinking `min: 0, exclusiveMin: true` is probably safer since some currencies might have small fractional positive values (Bitcoin!).

imports/plugins/core/orders/server/no-meteor/mutations/createRefund.js
```diff
@@ -0,0 +1,114 @@
...
+
+  if (modifiedCount === 0 || !updatedOrder) throw new ReactionError("server-error", "Unable to update order");
+
+  await appEvents.emit("afterOrderUpdate", {
+    order: updatedOrder,
+    updatedBy: userId
+  });
+
+  // Send email to notify customer of a refund
+  sendOrderEmail(context, order, "refunded");
```
focusaurus: I think maybe we should pass `updatedOrder` to this function. The old code passed `order` too but seems maybe a bug.

imports/plugins/core/orders/server/no-meteor/mutations/createRefund.test.js
```diff
@@ -0,0 +1,132 @@
...
+  mockContext.userHasPermission.mockReturnValueOnce(true);
+
+  await expect(createRefund(mockContext, {
+    amount: 0,
+    orderId: fakeOrder._id,
+    paymentId: "payment3",
+    reason: "Customer was unsatisfied with purchase"
+  })).rejects.toThrowErrorMatchingSnapshot();
+});
```
focusaurus: Can we add a test for a base success case? Looks like all these tests are error cases.

imports/plugins/included/marketplace/server/no-meteor/util/stripeCreateRefund.js
```diff
@@ -10,15 +10,16 @@ import formatForStripe from "./formatForStripe";
  * @param {Object} context an object containing the per-request state
  * @param {Object} paymentMethod object containing transaction ID
  * @param {Number} amount the amount to be refunded
+ * @param {String} [reason] the reason for the refund
```
focusaurus: Can we link to the stripe docs describing the allowed values for `reason` here or put in a cheat sheet?

imports/plugins/included/payments-stripe/server/no-meteor/util/stripeCreateRefund.js
```diff
@@ -10,16 +10,17 @@ import formatForStripe from "./formatForStripe";
  * @param {Object} context an object containing the per-request state
  * @param {Object} paymentMethod object containing transaction ID
  * @param {Number} amount the amount to be refunded
+ * @param {Number} [reason] the reason for the refund
```
focusaurus: (Repeat) Can we link to the stripe docs describing the allowed values for `reason` here or put in a cheat sheet?

imports/plugins/included/marketplace/server/no-meteor/util/stripeCreateRefund.js
kieckhafer: Yes... I also have an explanation in the comments in the client side code ( #5158 )

imports/plugins/included/payments-stripe/server/no-meteor/util/stripeCreateRefund.js
kieckhafer: Yes... I also have an explanation in the comments in the client side code ( #5158 )

imports/plugins/core/orders/server/no-meteor/mutations/createRefund.test.js
kieckhafer: Test added

kieckhafer:
@focusaurus all comments addressed, ready for another look

PR #5355, opened 2019-07-23
Feat: catalog partial publish / rerun transformations

nnnnat:
Impact: **minor**  
Type: **feature**

## Changes
- Transformations that happen during publication to Catalog are now centralized into `context.mutations.applyCustomPublisherTransforms`, which can be overridden by other plugins to customize the transformation process.
- Created `context.mutations.partialProductPublish`, which can be used by any plugin to request a partial re-publish (re-transformation) of any product.
- Updated inventory plugin to use the new `partialProductPublish` rather than just rerunning its own transformation. This will ensure that other transformations that depend on inventory can also rerun. For example, if prices depend on available inventory, the prices can be updated on a CatalogProduct after every inventory change.

## Breaking changes
None

## Testing
1. Verify that normal publish flow still works. Verify that tax, inventory, and price fields are set on the published catalog product.
2. `partialProductPublish` is not exposed through GraphQL at this time, but you can create a custom plugin that calls it to verify that it works as expected.


cmbirk:
@rosshadden @ticean @aldeed can we go ahead and merge this?

PR #5357, opened 2019-07-24
ci: add eslint check to only check changed files and fail when warnings are thrown

kieckhafer:
Impact: **minor**  
Type: **ci|performance**

## Issue
We currently have ~300 `eslint` warnings related to `jsdoc`.

## Solution
To incrementally fix all these issues. we will add a check to the CI process to throw an error if there are any `eslint` `errors` OR `warnings`, **only files you changed in your PR.** In addition to only checking changed files, this is different than our traditional `eslint` CI check in that it will not pass even if there are `warnings`, where our traditional check only fails against `errors`.

## Breaking changes
None

## Testing
1. Make any change (add a random comment in the file somewhere) to multiples file that is currently throwing an `eslint` warning, and make a PR with said change. Two files with `eslint` warnings:
- - `imports/plugins/included/product-variant/client/templates/products/productSettings/productSettings.js`
- - `client/modules/core/main.js`
1. see that CI does not pass until warnings is fixed

I made a few test commits with purposefully failing files, you can see the CI check running and failing when there are eslint `warnings` (not errors): https://circleci.com/gh/reactioncommerce/reaction/49620?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link


kieckhafer:
@mikemurray could you please see that this works as it should, since you know what it's supposed to do?

@aldeed / @focusaurus could you please check the bash script / config file code?

kieckhafer:
@mikemurray confirmed @focusaurus's changes still work.

Here is the latest CI run that is failing on purpose, should be good enough for testing: https://circleci.com/gh/reactioncommerce/reaction/49757?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

PR #5358, opened 2019-07-24
(fix) Add migration for index

zenweasel:
Nice

## Testing Instructions

1. Check that the index on `Catalog.product.productId` in not set to `unique`
2. Start the app with this branch
3. Observe that the app starts (on first launch it may throw a transient error)
4. Check that the index on `Catalog.product.productId` in now set to `unique`

### DIFF
--- a/imports/plugins/core/versions/server/migrations/69_make_catalog_product_unique.js
+++ b/imports/plugins/core/versions/server/migrations/69_make_catalog_product_unique.js
@@ -16,7 +16,7 @@ function handleError(error) {
       (!error.message.includes("index not found") && !error.message.includes("ns not found"))
     )
   ) {
-    Logger.warn(error, "Caught error from dropIndex calls in migration 99");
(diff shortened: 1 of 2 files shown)


imports/plugins/core/versions/server/migrations/69_make_catalog_product_unique.js
```diff
@@ -0,0 +1,37 @@
...
+ * Drop all indexes that support queries that are no longer expected
+ * to be made by any plugins, or that are already supported by other
+ * indexes.
+ */
+Migrations.add({
+  version: 69,
+  up() {
+    const { Catalog } = rawCollections;
+    Catalog.dropIndex("productId", handleError);
+    Catalog.createIndex("productId", handleError);
```
aldeed: These should be "product.productId". And for both `createIndex` calls, you're missing the options `{ unique: true, background: true }`

zenweasel:
@aldeed Requested changes made

PR #5360, opened 2019-07-24
chore: fix no-undef lint issues

kieckhafer:
Impact: **minor**  
Type: **chore**

## Testing
1. Run lint
2. See that there are no `no-undef` errors or warnings thrown
3. See that `no-undef` `warn` override has been removed from package.json

### DIFF
--- a/imports/plugins/core/core/server/methods/getOpaqueIdFromInternalId.js
+++ b/imports/plugins/core/core/server/methods/getOpaqueIdFromInternalId.js
@@ -1,3 +1,4 @@
+import { check } from "meteor/check";
 import { namespaces } from "@reactioncommerce/reaction-graphql-utils";
 import { encodeOpaqueId } from "@reactioncommerce/reaction-graphql-xforms/id";
 
--- a/package.json
+++ b/package.json
@@ -393,7 +393,6 @@
     },
     "rules": {
       "jsx-a11y/label-has-for": "off",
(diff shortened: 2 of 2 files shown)


imports/plugins/core/core/server/methods/getOpaqueIdFromInternalId.js
```diff
@@ -1,3 +1,4 @@
+import { check } from "meteor/check";
```
bt3gl: HI Erik, hat's this for?

bt3gl:
LGTM. Lint was ran, not `no-undef` warnings!

Reading about no-undef, I learned that this is a rule to disallow undeclared variables, such as a forgetting var in a for loop. What could be the real problem of having such type of issue in our code? Garbage collection? Memory / security issues?

imports/plugins/core/core/server/methods/getOpaqueIdFromInternalId.js
kieckhafer: This is a meteor helper that checks types of input to make sure the data you receive matches what you are expecting. Kind of similar to a `PropType` check in React.

Meteor magically imports a lot of the core functionality, so that's why this was not imported before, it was just being done magically in the background. We are trying to get away from that and purposefully import everything needed, which is why this error was thrown.

PR #5382, opened 2019-07-29
feat: editable tag slug

### LINKED ISSUE #5380: Allow tag slug to be editable, don't change slug when title changes
## Issue Description

Changing the title of a tag changes the slug. This can cause broken links to tag pages.

## Steps to Reproduce
1. Change the title of a tag
2. See the slug change on save

## Possible Solution

- The title sets the slug if the slug is undefined
- The title does not change the slug if the slug is defined
- The slug field is editable
- The slug field validates or slugifies its contents in a way that allows notifies the user what the resulting slug.

mikemurray:
Resolves #5380   
Impact: **major**  
Type: **feature**

## Issue

- Tag slug is not editable
- Tags slug changes ever time the `Tag name` changes

## Solution

- Make slug editable
- Default slug to a slugified `Tag name` if slug is left blank
- Notify the user if they change the slug (with an info inline alert)
- Bonus: show errors as an inline alert
- Bonus: fixed a broken save button

Reaction

Reaction

(we need to revisit error message in another PR, but for now, at least it's more human-readable than it was before)

## Breaking changes

none

### Create new
1. Create a tag, leave the slug **blank**
2. Save
3. See the slug default to the slugified tag name

### Create new
1. Create a tag, fill in the slug with normally (like you would a title, with capitalizations, punctuation etc.)
2. Save
3. See the slug is now the slugified version of your input

### Update
1. Change the `Tag name`
2. Save
3. See the slug is still the same
4. Update the slug
5. Save
6. See the slug has updated. (refresh to verify)
7. Delete the contents of the slug field
8. Save
9. See the slug is now the slugified `Tag name`

X. Try to break it.

### DIFF
--- a/imports/plugins/core/tags/client/components/TagForm.js
+++ b/imports/plugins/core/tags/client/components/TagForm.js
@@ -151,20 +151,26 @@ class TagForm extends Component {
       });
     }
 
-    const result = await mutation({
-      refetchQueries,
-      variables: {
-        input
+    this.setState({ error: null });
+    try {
+      const result = await mutation({
+        refetchQueries,
+        variables: {
+          input
+        }
+      });
+
+      if (result.data.addTag) {
+        this.props.onCreate(result.data.addTag.tag);
+      } else {
+        this.props.onUpdate(result.data.updateTag.tag);
       }
-    });
 
-    if (result.data.addTag) {
-      this.props.onCreate(result.data.addTag.tag);
-    } else {
-      this.props.onUpdate(result.data.updateTag.tag);
+      return result;
+    } catch (error) {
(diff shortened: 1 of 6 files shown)


machikoyasuda:
Manual testing looks good ✔️ 

Screen Shot 2019-07-29 at 3 08 28 PM
Screen Shot 2019-07-29 at 3 07 58 PM
Screen Shot 2019-07-29 at 3 07 25 PM
Screen Shot 2019-07-29 at 3 06 32 PM
Screen Shot 2019-07-29 at 3 07 29 PM
Screen Shot 2019-07-29 at 3 08 28 PM

PR #5391, opened 2019-07-31
feat(product table): show Action dropdown, even if no products are selected

### LINKED ISSUE #5388: Bulk Tag/Product Feature: Product table's Action dropdown should always be shown
Currently, the dropdown is only shown when there are products selected.

It should always be shown.

machikoyasuda:
Resolves #5388  
Impact: **minor**  
Type: **feature**

## Issue
In the new Bulk Product/Tag action flow, it's required that the Action dropdown menu be available to click - even if products are not selected.

## Solution
This PR changes the `renderToolbar()` method to 1) always show the dropdown, but 2) `disable` the action options if there are no products selected. 

Screen Shot 2019-07-30 at 11 13 51 PM

## Breaking changes
None

## Testing
0. http://localhost:3000/operator/products
1. Select 0 products and click the dropdown. Options should be disabled.
1. Select several products and click the dropdown. Try archiving, make-hidden-ing some things. Make sure the count is correct.

### DIFF
--- a/imports/plugins/included/product-variant/components/productGrid.js
+++ b/imports/plugins/included/product-variant/components/productGrid.js
@@ -215,78 +215,76 @@ class ProductGrid extends Component {
           </Menu>
         </Toolbar>
       );
-    // eslint-disable-next-line no-else-return
-    } else {
-      return (
-        <Toolbar>
-          <Button
-            aria-owns={bulkActionMenuAnchorEl ? "bulk-actions-menu" : undefined}
-            aria-haspopup="true"
-            onClick={this.handleShowBulkActions}
-            variant="outlined"
-          >
-            {i18next.t("admin.productTable.bulkActions.actions")}
-            <ChevronDownIcon />
-          </Button>
-          <Menu
-            id="bulk-actions-menu"
-            anchorEl={bulkActionMenuAnchorEl}
-            open={Boolean(bulkActionMenuAnchorEl)}
-            onClose={this.handleCloseBulkActions}
-          >
-            <ConfirmDialog
-              title={i18next.t("admin.productTable.bulkActions.publishTitle", { count })}
-              message={i18next.t("admin.productTable.bulkActions.publishMessage")}
-              onConfirm={this.handleBulkActionPublish}
-            >
-              {({ openDialog }) => (
-                <MenuItem disabled onClick={openDialog}>{i18next.t("admin.productTable.bulkActions.publish")}</MenuItem>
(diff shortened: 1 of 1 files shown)


imports/plugins/included/product-variant/components/productGrid.js
```diff
@@ -205,22 +203,89 @@ class ProductGrid extends Component {
...
+            onClose={this.handleCloseBulkActions}
+          >
+            <ConfirmDialog
+              title={i18next.t("admin.productTable.bulkActions.publishTitle", { count })}
+              message={i18next.t("admin.productTable.bulkActions.publishMessage")}
+              onConfirm={this.handleBulkActionPublish}
+            >
+              {({ openDialog }) => (
+                <MenuItem disabled onClick={openDialog}>{i18next.t("admin.productTable.bulkActions.publish")}</MenuItem>
+              )}
```
machikoyasuda: This implementation is so ugly.

I wanted to do something like this, but I couldn't get it to work:

```
const isAnySelected = Array.isArray(selectedProductIds) && selectedProductIds.length;
```

```
              {({ openDialog, isAnySelected }) => (
                <MenuItem { isAnySelected ? "disabled" : "" } onClick={openDialog}>{i18next.t("admin.productTable.bulkActions.publish")}</MenuItem>
              )}
```

For `disabled`, an implementation like this doesn't work either: 

```
disabled="isAnySelected"
```

Even if `disabled="false"`, or `disabled="null"`, the MenuItem will still render as `disabled.`

Any other better ideas to make this a bit cleaner??

machikoyasuda: PS. Trying the code out in this sandbox is cool and fast: https://catalyst.reactioncommerce.com/#/Base%20Components/Feedback/ConfirmDialog

imports/plugins/included/product-variant/components/productGrid.js
```diff
@@ -218,9 +215,78 @@ class ProductGrid extends Component {
           </Menu>
         </Toolbar>
       );
-    }
+    // eslint-disable-next-line no-else-return
```
kieckhafer: This doesn't need to be inside of an `else`, you can just `return` immediately after the `if`, since there isn't any other option.

machikoyasuda:
This React props error shows up in the console, but this error also already appears in `develop` branch - so I do not think it was introduced in this branch.

Screen Shot 2019-07-30 at 11 16 25 PM

PR #5396, opened 2019-07-31
perf: up data transfer limit to allow bigger sizes

kieckhafer:
Impact: **minor**  
Type: **performance**

## Issue
Large GraphQL mutations, such as updating the `NavigationTree` when the navigation has many nodes, was throwing an error because we use the default `express` setting which only allows 100kb size.

## Solution
Add a middleware which we can configure to use a larger default size. In this case, we are setting it at `5mb`, as this is the default AWS setting.

## Breaking changes
None

## Testing
1. Create a large json file that has over 5mb of data.
1. run `curl   -X POST   -H "Content-Type: application/json"  --data @{filename}.json localhost:3000/graphql-beta` and see you'll receive a `PayloadTooLargeError` error
1. Create a json file that has less than 5mb of data
1. run the same `curl` and see the `PayloadTooLargeError` isn't thrown

### DIFF
--- a/imports/node-app/core/config.js
+++ b/imports/node-app/core/config.js
@@ -1,7 +1,7 @@
 import envalid, { bool, makeValidator, str } from "envalid";
 
 const bodyParserValidator = makeValidator((value) => {
-  if (typeof value !== "number") throw new Error("Expected type number letters");
+  if (typeof value !== "number") throw new Error("Expected type number");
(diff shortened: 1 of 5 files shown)


kieckhafer:
@focusaurus added some code to check for size limit via `env` https://github.com/reactioncommerce/reaction/pull/5396/commits/9d7c05a100bc73689a8693d5de82e3ad116a34b3

PR #5403, opened 2019-08-01
ci: remove snyk as a CI task

### LINKED ISSUE #5327: Remove snyk as a blocking CI step
Snyk frequently comes up as a blocking step in our CI process.

This doesn't fully make sense, especially since the majority of the time, what Snyk flags is already in `develop` branch of Reaction, and not introduced by the PR being flagged.

This causes unneeded frustration for developers, having PR's blocked by things they have no control over.

We should remove SNYK as a blocking CI step, and use it for reference to improve the app instead. We can also look into adding different apps that might be easier to work with: Automation via Zapier, If This Than That, etc.

kieckhafer:
Resolves #5327   
Impact: **minor**  
Type: **ci**

## Issue
Snyk issues are timed randomly, with issues thrown when new bugs are reported. This could affect code we've had in the app for a while, and has nothing to do with the current PR. This causes hassle for committing code to Reaction.

## Solution
Remove the CI snyk process, and rely on the daily scanned results sent to those subscribe to snyk. In the future, we can task a scheduled `snyk` check to make sure these are being taken care of.

## Breaking changes
None

## Testing
1. See that `snyk` did not run in CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -305,46 +305,6 @@ jobs:
           command: |
             hadolint Dockerfile
 
-  snyk-security:
-    <<: *defaults
-    steps:
-      - checkout
-      - setup_remote_docker:
-          docker_layer_caching: true
-      - restore_cache:
-          keys:
-            - reaction-v2-node-modules-{{ checksum "package.json" }}-{{ checksum "package-lock.json" }}
-            - reaction-v2-node-modules-{{ .Branch }}
-            - reaction-v2-node-modules-master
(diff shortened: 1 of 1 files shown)


kieckhafer:
@spencern @aldeed I know this was discussed, but just want to confirm this was the final plan before it gets merged.

machikoyasuda:
@spencern This would be a great PR to merge today, as it's August 1st and it's blocking all new PRs from going in.

PR #5404, opened 2019-08-01
Feature: add bulk mutations to manage products and tags

### LINKED ISSUE #5384: Bulk Tag/Product Feature: Back-end bulk action API
**New API: CSV parser + Bulk add/remove Tags from Products -**

- Mutations for bulk operations for adding/removing tags from products
- Live in the `core/products` plugin

1. Upload a .csv file to browser. (doesn't need to be saved to the server)
    1. Using some npm module for .csv parsing
2. Parse CSV file for product IDs
3. Send the list of Product IDs to the server to the bulk operation mutation as the input with the tags to apply to those products
    1. API should accept multiple product ids (for not the ui only has to support uploading one file for phase 1)
    2. API accept multiple tags ids (for now the ui only has to support one for phase 1)
    3. Support adding and removing tags form the products
4. Apply tags to products using mongo bulk operations
    1. Product have a `hashtags` field
        1. Array of ids
    2. Mongo `addToSet` if there's a bulk op for that
        1. `addToSet` can accept an array
        2. Reaction bulk example: imports/plugins/core/versions/server/no-meteor/util/findAndConvertInBatches.js
        3. [https://docs.mongodb.com/manual/reference/method/db.collection.bulkWrite/#bulkwrite-write-operations-updateonemany](https://docs.mongodb.com/manual/reference/method/db.collection.bulkWrite/#bulkwrite-write-operations-updateonemany)

            bulk.find({
              _id: $in: [ /* array of product ids */]
            }).update({
              $addToSet: {
                hashtags: {
                  $each: [/* the tags */]
                }
              }
            });

willopez:
Resolves #5384  
Impact: minor
Type: feature

## Summary
This PR adds mutations to allow store operators to manage products and tags in bulk, a list of implemented mutations follows: 
```
addTagsToProducts(input: ProductTagsOperationInput):ProductTagsOperationPayload

removeTagsFromProducts(input: ProductTagsOperationInput):ProductTagsOperationPayload

```
## Breaking changes
None

## Testing
Test info for `addTagsToProducts`

**Mutation**
```
mutation addTagsToProducts($input: ProductTagsOperationInput!) {
  addTagsToProducts(input: $input) {
    foundCount
    notFoundCount
    updatedCount
    writeErrors {
      documentId
      errorMsg
    }
  }
}
```

**Sample variables**
**NOTE** Actual values for productIds and tagIds will vary depending on your Reaction instance, replace values as necessary
```
{
  "input": {
    "productIds": ["BCTMZ6HTxFSppJESk"],
    "tagIds": ["cseCBSSrJ3t8HQSNP"]
  }
}
```

Test info for `removTagsFromProducts`

**Mutation**
```
mutation removeTagsFromProducts($input: ProductTagsOperationInput!) {
  removeTagsFromProducts(input: $input) {
    foundCount
    notFoundCount
    updatedCount
    writeErrors {
      documentId
      errorMsg
    }
  }
}
```

**Sample variables**
**NOTE** Actual values for productIds and tagIds will vary depending on your Reaction instance, replace values as necessary
```
{
  "input": {
    "productIds": ["BCTMZ6HTxFSppJESk"],
    "tagIds": ["cseCBSSrJ3t8HQSNP"]
  }
}
```


willopez:
@aldeed I do like the idea applying multiple actions to a set of products. However, the design team has gone through a couple of iterations, and has gotten feedback from a client(SDI) for executing one action at a time. See image for design details. I believe that for the first implementation, executing one action(a single mutation) at a time will suffice. Further, the design would need to be updated to support multiple actions + getting feedback from clients, which would delay this feature from being released. 

In short, I believe multiple action support can be added in the future, in the mean time this will be a good start. @spencern thoughts? In

addTags

focusaurus:
The proposed bulk update mechanism seems like a bespoke/parochial thing outside of graphql norms so I would lean against it, especially if we can already do a series of mutations in the same graphql request. The more complexity and time a request needs to run the more issues we'll hit in terms of timeouts, partial applications, transactional errors, client retries, etc. If we're not cheaply queuing up work for background processing, I think we need to keep overall request workload within fairly small bounds.

Looking at your specific example, can we not already set the tags and visibility in a single mutation? Or is tag management always add/remove one tag per mutation?

aldeed:
@focusaurus We could support setting tags along with visibility, but we currently don't. However, the designs call for needing to add or remove tags without necessarily having fetched the existing tag list, so we probably need all three: add, remove, set.

@willopez I don't have a strong opinion and I believe we may someday need both varieties, so if we want to go with the simpler, more Graphql-like individual mutations for now, that works. After getting burned by performance issues on bulk updates recently, I just thought it needed some discussion with that mindset.

imports/plugins/core/product/server/no-meteor/schemas/schema.graphql
```diff
@@ -6,3 +6,68 @@ type ProductConfiguration {
   "The ProductVariant ID"
   productVariantId: ID!
 }
+
+"Operation"
+type Operation {
+  "Id of document"
+  _id: String
```
aldeed: Would we ever need to add more that an ID? Can we skip having the type and just have a `documentId` property on `WriteError`?

imports/plugins/core/product/server/no-meteor/schemas/schema.graphql
```diff
@@ -6,3 +6,68 @@ type ProductConfiguration {
...
+}
+
+"A bulk write error type"
+type WriteError {
+  "Error code"
+  code: Int
```
aldeed: These are straight from Mongo? I think we should keep such things internal.

imports/plugins/core/product/server/no-meteor/schemas/schema.graphql
```diff
@@ -6,3 +6,68 @@ type ProductConfiguration {
...
+
+  "Error message"
+  errmsg: String
```
aldeed: Change the field name to `message`

imports/plugins/core/product/server/no-meteor/schemas/schema.graphql
```diff
@@ -6,3 +6,68 @@ type ProductConfiguration {
...
+
+  "Index of of error"
+  index: Int
```
aldeed: Is this useful information? Not sure what it's for. Maybe could be removed. If not, the description has "of of" so need to fix that

imports/plugins/core/product/server/no-meteor/schemas/schema.graphql
```diff
@@ -6,3 +6,68 @@ type ProductConfiguration {
...
+
+"Input for adding tags to products in bulk"
+input ProductsBulkOperationInput {
+  "An optional string identifying the mutation call, which will be returned in the response payload"
+  clientMutationId: String
+
+  "An array of product ids to which an array of tags will be added"
+  productIds: [ID]
+
+  "An array of tag ids to add to an array of products"
```
aldeed: ids -> IDs, same on the description above

imports/plugins/core/product/server/no-meteor/schemas/schema.graphql
```diff
@@ -6,3 +6,68 @@ type ProductConfiguration {
...
+  foundCount: Int
+
+  "The number of products for which a match was not found"
+  notFoundCount: Int
+
+  "The number of products successfully updated"
+  updatedCount: Int
+
+  "An array of write errors if any were generated"
+  writeErrors: [WriteError]
```
aldeed: Overall this payload strikes me as being not too useful for the client. I think maybe you should parse the result a bit and convert it into something like this for the response payload:

```js
{
  results: [
    { documentId: "productId1", wasFound: true, wasUpdated: true, errorMessage: null },
    { documentId: "productId2", wasFound: true, wasUpdated: false, errorMessage: "something went wrong" } 
  ]
}
```

This would make it easiest for browser code to display the information in helpful way to the user.

imports/plugins/core/product/server/no-meteor/utils/executeBulkOperation.js
```diff
@@ -0,0 +1,31 @@
...
+const logCtx = { name: "core/product", file: " bulkRemoveTagsFromProducts" };
+
+/**
+ *
+ * @param {Object[]} collection Mongo collection
+ * @param {Object[]} operations bulk operations to perform
+ * @param {Int} totalProducts total number of products to operate on
+ * @return {Object} Object with information of results of bulk the operations
+ */
+export default async function (collection, operations, totalProducts) {
```
aldeed: All functions should be named

imports/plugins/core/product/server/no-meteor/utils/executeBulkOperation.js
```diff
@@ -0,0 +1,31 @@
+import Logger from "@reactioncommerce/logger";
+
+const logCtx = { name: "core/product", file: " bulkRemoveTagsFromProducts" };
```
aldeed: Might not be `bulkRemoveTagsFromProducts`

willopez: added correct file name

imports/plugins/core/product/server/no-meteor/schemas/schema.graphql
willopez: Yes, kept internal now.

imports/plugins/core/product/server/no-meteor/schemas/schema.graphql
willopez: It would be ideas to have the details you mention above for each `documentId` updated. However, the `bulkWrite` operation does not return such information for for each updated document, it only returns those details if there was a write error. https://docs.mongodb.com/manual/reference/method/db.collection.bulkWrite/index.html

imports/plugins/core/product/server/no-meteor/schemas/schema.graphql
willopez: updated

imports/plugins/core/product/server/no-meteor/schemas/schema.graphql
willopez: clean up the error messages to only include the `documentId` and `errorMsg` for write errors.

imports/plugins/core/product/server/no-meteor/schemas/schema.graphql
willopez: removed.

willopez:
@aldeed I have addressed all your comments, ready for another review.

mikemurray:
Just merged from `develop` to see if there are changes to the `changed files` check.

Issue #5285, opened 2019-07-08
update fulfillment group status update button to use split button


kieckhafer:
When Reaction is updated to use MUI 4.0, update the Orders panel fulfillment group to use the Split Button to update group status: https://material-ui.com/components/buttons/#split-button

https://github.com/reactioncommerce/reaction/tree/develop/imports/plugins/core/orders/client/components/OrderCardFulfillmenGroupStatusButton.js

PR #5405, opened 2019-08-01
refactor: refunds section of Orders 2.0

kieckhafer:
Addresses the refunds portion of #5157 
Impact: **minor**  
Type: **refactor**

## Issue
This PR is part of refactoring the Orders Operator UI to deal with multiple fulfillment groups, graphql, etc, to make it 2.0 ready. The majority of the Orders 2.0 UI has already been [tested and merged into Reaction](https://github.com/reactioncommerce/reaction/pull/5158).

This PR splits off the refunds portion of the overall work to make it easier to test in smaller chunks. This PR only addresses at the UI portion of refunds - the entirety of the refunds tab. The GraphQL queries and mutations related to refunds have already been tested and merged (#5352, #5354), and don't need to be addressed in this PR.

## Breaking changes
None

## Testing
- Place two orders (six total) of each of the following: 
- - Using only the Stripe payment method
- - Using only the example payment method
- - Using multiple payments, a combination of Stripe and Example (two or more payments)

- Navigate to the Orders Operator UI
- Click on an order, and then once on the single order page, change the URL from `/orders/` to `/orders-beta/` to see the new UI.

- Once an order is open, `Capture` all payments on the order.

- Then, navigate to the refunds tab. It should look something like this:

Order_Details_for_order_reference__Y3imj7gnML4JtX29B

At this point, do the following tests:
1. Refund the maximum amount of the single stripe payment
1. Refund the maximum amount of the single example payment method
1. refund the maximum amount of all methods in the multiple payment order (you can 

- After all those are done, confirm the input box disappears, and you can no longer create a refund.
- Confirm you can see previous refunds in the appropriate section

Then:
1. Refund a partial amount of a single stripe payment
1. Refund the partial amount of the single example payment method
1. refund the partial amount of all methods in the multiple payment order (you can 

- At this point, see that there is a line of text that tells you how much has been refunded, and how much is available to refund on each payment

Then:
1. attempt to refund more than the remaining amount on all payments
1. The UI will "let" you, but the server side will only allow the remaining amount and complete the refund

#### Adding translations for Chips:
Add orderStatusLabels to the Shops collection shop object. This will provide translations for all the Chip style order status badges in the Orders Fulfillment UI.
```
"orderStatusLabels": {
    "new" : [ 
        {
            "language" : "en",
            "label" : "Order received"
        }, 
        {
            "language" : "es",
            "label" : "Recibido"
        }
    ],
        {
            "language" : "en",
            "label" : "Order received"
        }, 
        {
            "language" : "es",
            "label" : "Recibido"
        }
    ],
        {
            "language" : "en",
            "label" : "Shipped"
        }, 
        {
            "language" : "es",
            "label" : "Enviado"
        }
    ],
        {
            "language" : "en",
            "label" : "Canceled"
        }, 
        {
            "language" : "es",
            "label" : "Cancelado"
        }
    ],
    "coreOrderWorkflow/picked" : [ 
        {
            "language" : "en",
            "label" : "Picked"
        }
    ],
    "coreOrderWorkflow/packed" : [ 
        {
            "language" : "en",
            "label" : "packed"
        }
    ],
    "coreOrderWorkflow/shipped" : [ 
        {
            "language" : "en",
            "label" : "shipped"
        }
    ]
}
```


kieckhafer:
@machikoyasuda @aldeed This is "blocked" by #5352, however I've pulled that branch into this PR so it can be tested as is, just not merged until that PR is.

imports/plugins/core/orders/client/components/OrderPreviousRefunds.js
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * @name OrderPreviousRefunds
+ * @param {Object} props Component props
+ * @returns {React.Component} returns a React component
+ */
+function OrderPreviousRefunds(props) {
+  const { classes, order } = props;
+  const { refunds: unsortedRefunds } = order;
+  // sort refunds by date, newest first, for display purposes
+  const refunds = unsortedRefunds.slice().sort((refundA, refundB) => new Date(refundB.createdAt) - new Date(refundA.createdAt));
```
aldeed: It could wait for a future PR, but I'd prefer to see this implemented through a `sort` param on the server resolver.

imports/plugins/core/orders/client/components/OrderPreviousRefunds.js
```diff
@@ -0,0 +1,98 @@
...
+                </Grid>
+              );
+            })}
+          </CardContent>
+        </Card>
+      </Grid>
+    );
+  }
+
+  return null;
```
aldeed: We don't want any empty state text?

imports/plugins/core/orders/client/components/OrderRefunds.js
```diff
@@ -0,0 +1,303 @@
+import React, { Fragment, useEffect, useRef, useState } from "react";
+import PropTypes from "prop-types";
+import { Mutation } from "react-apollo";
+import { Form } from "reacto-form";
+import withStyles from "@material-ui/core/styles/withStyles";
```
aldeed: Maybe switch to `makeStyles` and `useStyles` here since this is a function component?

imports/plugins/core/orders/client/components/OrderRefunds.js
```diff
@@ -0,0 +1,303 @@
...
+  }
+});
+
+/**
+ * @name OrderRefunds
+ * @param {Object} props Component props
+ * @returns {React.Component} returns a React component
+ */
+function OrderRefunds(props) {
+  const hasPermission = Reaction.hasPermission(["reaction-orders", "order/fulfillment"], Reaction.getUserId(), Reaction.getShopId());
```
aldeed: I would remove this and the two places where `hasPermission` are checked. We should let the server decide. But if the server does respond with a permission error, then we should show a helpful error message, e.g., "You don't have permission to do this"

imports/plugins/core/orders/client/components/OrderRefunds.js
```diff
@@ -0,0 +1,303 @@
...
+  const { classes, order } = props;
+  const { payments } = order;
+
+  // useRef
+  const inputLabel = useRef(null);
+
+  // useState
+  const [labelWidth, setLabelWidth] = useState(0);
+  const [refundReasonSelectValues, setRefundReasonSelectValues] = useState({ reason: "" });
```
aldeed: I didn't test it, but using an object here seems unnecessary to me. 

```js
const [refundReason, setRefundReason] = useState("");

setRefundReason(event.target.value);

value={refundReason}
```

imports/plugins/core/orders/client/components/OrderRefunds.js
```diff
@@ -0,0 +1,303 @@
...
+
+  // useEffect
+  // update label width when refund select is activate
+  useEffect(() => {
+    if (inputLabel && inputLabel.current) {
+      setLabelWidth(inputLabel.current.offsetWidth);
+    }
+  }, []);
+
+  const canRefund = payments.every((payment) => payment.status !== "refunded");
```
aldeed: The payment method also has a `canRefund` prop, meaning does the method even support refunds. I believe both of our built-in payment methods have this set to `true`, but you could set the example IOU method to `false` and test what happens.

You should be able to get this through GraphQL by requesting `method.canRefund` on each `payments`:

```graphql
{
  payments {
    method {
      canRefund
    }
  }
}
```

It seems like the method not supporting refunds is slightly different from the meaning of `canRefund` here. So you may want to treat them separately and maybe UI should explicitly show a "Payment method does not support refunds" message somewhere? cc @rymorgan

imports/plugins/core/orders/client/components/OrderRefunds.js
```diff
@@ -0,0 +1,303 @@
...
+      setRefundTotal(() => reducedRefundTotal);
+    }
+  };
+
+  const handleRefundAmountChange = (amount) => {
+    if (amount) handleRefundTotalUpdate();
+  };
+
+  const handleRefundReasonSelectChange = (event) => {
+    handleRefundTotalUpdate();
```
aldeed: It doesn't seem like changing the select value would have any impact on `refundTotal` value?

imports/plugins/core/orders/client/components/OrderRefunds.js
```diff
@@ -0,0 +1,303 @@
...
+        mutation({
+          variables
+        });
+      }
+    });
+  };
+
+  // When refund amounts are changed, add up amounts to display in button
+  const handleRefundTotalUpdate = () => {
+    if (this.form && this.form.state) {
```
aldeed: It might be simpler and cleaner to implement this as an `onChange` or `onChanging` prop on `Form`?

imports/plugins/core/orders/client/graphql/mutations/createRefund.js
```diff
@@ -0,0 +1,18 @@
+import gql from "graphql-tag";
+import { orderCommonFragment } from "../fragments/orderCommon";
+
+export default gql`
+  mutation createRefundMutation($amount: Float! , $orderId: ID!, $paymentId: ID!, $reason: String, $language: String! = "en") {
```
aldeed: I'm guessing `$language` was left here accidentally

imports/plugins/core/orders/client/graphql/queries/refunds.js
```diff
@@ -0,0 +1,15 @@
...
+    refunds(orderId: $orderId, shopId: $shopId, token: $token) {
+      amount {
+        amount
+        displayAmount
+      }
+      createdAt
+      paymentDisplayName
+      reason
+    }
+  }
```
aldeed: Is this query still used?

imports/plugins/core/orders/client/components/OrderPreviousRefunds.js
```diff
@@ -0,0 +1,98 @@
+import React from "react";
+import PropTypes from "prop-types";
+import moment from "moment";
+import withStyles from "@material-ui/core/styles/withStyles";
```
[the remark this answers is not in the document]
kieckhafer: 👍 We were still on MUI 3.x when I started this PR, so those weren't available at the time. Can make the update.

imports/plugins/core/orders/client/components/Order.js
```diff
@@ -35,14 +44,28 @@ function Order(props) {
         <Grid item xs={12}>
           <OrderHeader order={order} />
         </Grid>
-        <Grid container spacing={2}>
-          <Grid item xs={12}>
-            <OrderCardFulfillmentGroups order={order} />
+        <Grid className={classes.tabs} item xs={12}>
+          <Tabs value={currentTab} onChange={handleTabChange}>
+            <Tab label={i18next.t("fulfillment", "Fulfillment")} />
+            <Tab label={i18next.t("refunds", "Refunds")} />
```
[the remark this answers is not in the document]
kieckhafer: Sounds good. We might need to make a ticket with a complete overhaul of i18n then, because I think we do this style almost everywhere.

imports/plugins/core/orders/client/components/OrderPreviousRefunds.js
kieckhafer: I was debating doing this on server vs. client. In the end, I decided client because I felt like the client should take in the raw data however it came and decide how to sort it on it's own, independent.

I guess we could do it in both places, if the client wanted to re-sort it differently than the server was sending it, but was thinking one re-sort is less manipulation than two and to do it this way.

imports/plugins/core/orders/client/components/OrderPreviousRefunds.js
kieckhafer: There is no empty state card in the designs, if there are no previous refunds, it's just not there. 

Zeplin

imports/plugins/core/orders/client/components/OrderRefunds.js
kieckhafer: Correct. This is leftover from some old code, missed it while taking it out.

imports/plugins/core/orders/client/components/Order.js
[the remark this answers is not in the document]
kieckhafer: Issue created: https://github.com/reactioncommerce/reaction/issues/5416

imports/plugins/core/orders/client/components/OrderRefunds.js
kieckhafer: Yes, I think you're right. It was initially taking multiple reasons, one for each payment, but then it was combined into one reason per refund.

imports/plugins/core/orders/client/graphql/mutations/createRefund.js
kieckhafer: `language` is needed for `OrderCommon`. It was left here on purpose. Does it need to be added to the query then?

imports/plugins/core/orders/client/graphql/queries/refunds.js
kieckhafer: No it is not.

imports/plugins/core/orders/client/components/OrderRefunds.js
```diff
@@ -0,0 +1,302 @@
...
+import MenuItem from "@material-ui/core/MenuItem";
+import OutlinedInput from "@material-ui/core/OutlinedInput";
+import Select from "@material-ui/core/Select";
+import Typography from "@material-ui/core/Typography";
+import ErrorsBlock from "@reactioncommerce/components/ErrorsBlock/v1";
+import Field from "@reactioncommerce/components/Field/v1";
+import TextInput from "@reactioncommerce/components/TextInput/v1";
+import { i18next, Reaction } from "/client/api";
+import formatMoney from "/imports/utils/formatMoney";
+import ConfirmButton from "/imports/client/ui/components/ConfirmButton";
```
machikoyasuda: Use https://catalyst.reactioncommerce.com/#/Components/Feedback/ConfirmDialog instead

kieckhafer: Does `ConfirmDialog` have the button built in?

`ConfirmButton` and `ConfirmDialog` were two different components in `reaction`.

machikoyasuda: Which button? The first button is not built-in per say. You add it like this:

```
import Button from "../Button";

<ConfirmDialog
  title="Archive 24 products?"
  message="Archiving products removes them from both admin and customer views."
  onConfirm={() => { alert("Action confirmed!") }}
>
  {({ openDialog }) => (
    <Button color="primary" onClick={openDialog} variant="contained">Open Confirm Dialog</Button>
  )}
</ConfirmDialog>
```

kieckhafer: Ok... `ConfirmButton` is doing the same thing, just built in already so you didn't have to import `Button` as well.

Will update.

machikoyasuda:
We can design review this tomorrow after standup
 :) @cassytaylor @rymorgan 

Screen Shot 2019-08-05 at 1 32 24 PM

Screen Shot 2019-08-05 at 11 57 02 AM
Screen Shot 2019-08-05 at 12 00 43 PM
Screen Shot 2019-08-05 at 12 01 34 PM

kieckhafer:
@machikoyasuda yes I have been waiting to design review it until all the functionality is 100% there, but this is the last PR so it's a good time to do it. The dialog that has the dropdown in it will be changed to be the `SelectButton`, now that MUI 4.0 is being used, I just havent' gotten around to updating it yet (but do have a ticket #5285 )

imports/plugins/core/orders/client/components/OrderPreviousRefunds.js
kieckhafer: https://github.com/reactioncommerce/reaction/issues/5418

imports/plugins/core/orders/client/components/OrderRefunds.js
kieckhafer: Yeah, my `canRefund` here is "is there an amount left to refund, or has it already been fully refunded". I'll refactor that name a bit and implement a check to make sure it's even allowed.

@reactioncommerce/design I have added a few little things on top of your designs, just to provide a little contextual information where i wasn't seeing it, I can add a message if refunds aren't supported in a similar fashion, or we can discuss how to design this differently.

Order_Details_for_order_reference__JvhbMZkxhsZJqiL8f

machikoyasuda:
Other small random things:
1. "Order Details" should be "Order details"
2. "Order Summary" should be "Order summary"
3. FULFILLMENT/REFUNDS should be "Fulfillment" "Refunds"
4. Status Chips, like `completed` should be capitalized like `Completed`

kieckhafer:
@machikoyasuda Thanks!

Fixed 1 & 2.

3. is based on the default "Tab" component, it automatically capitalizes all letters, so probably something to look at when re-doing that component, because we use it in other places too (Tags I can think of off the top of my head), I think maybe products too?

4 Is a translation issue, and does show capitalized when translations are loaded. I added instructions on how to load translations in the testing part of the PR description. By default it just shows the raw data from the database.

machikoyasuda:
This could be a ticket for later but OrderStatusChip should inherit from https://catalyst.reactioncommerce.com/#/Components/Content/Chip not MUI Chip. (Just another layer of importing!) and all necessary type changes to OrderStatusChip should come from https://catalyst.reactioncommerce.com/#/Components/Content/Chip

kieckhafer:
@machikoyasuda Yeah, this PR unfortunately spanned a three week time where we went from MUI 3.x to MUI 4.x, React 16.4 to React 16.8, and Component Library to Catalyst, so there are new ways to do all these things which weren't implemented until after this PR was basically done.

#5158 is the final "switch from orders 1.0 to 2.0" ticket, so after this is merged, I am going to do any design updates from @reactioncommerce/design in that tickets, and can also do a lot of these switchovers to using Catalyst instead of MUI directly.

machikoyasuda:
@kieckhafer sounds good! that also.. makes way more sense. this way, we can create individual tickets for all the small tickets and assign them to more people, rather than 1 pr with a zillion little fixes!

imports/plugins/core/orders/client/components/OrderRefunds.js
kieckhafer: Created a ticket to revisit: https://github.com/reactioncommerce/reaction/issues/5421

imports/plugins/core/orders/client/graphql/mutations/createRefund.js
aldeed: For now I think this is fine. I didn't see it was used by the fragment.

imports/plugins/core/orders/client/components/OrderRefunds.js
kieckhafer: Updated to use `canRefund` from each payment as intended and using `isPaymentFullyRefunded` in place of the `canRefund` version I had.

imports/plugins/core/orders/client/components/OrderRefunds.js
kieckhafer: @aldeed It's kind of already doing that. I might be doing one extra step by calling `handleRefundAmountChange` on `onChanging`, which in turn calls `handleRefundTotalUpdate`. Is there something else that should be changed aside from maybe removing the extra step?

imports/plugins/core/orders/client/components/OrderRefunds.js
kieckhafer: How are we doing server-to-client alerts / error messages with GraphQL? I scanned the docs and didn't come across anything.

kieckhafer: Added using `InlineError`. Created a ticket to further explore better error messaging: https://github.com/reactioncommerce/reaction/issues/5426

kieckhafer:
@aldeed all comments either addressed and fixed, if deemed necessary for this PR, or a ticket created if not required. Ready for another check.

PR #5407, opened 2019-08-02
feat(producttable): add isvisible/hidden icons

### LINKED ISSUE #5392: Bulk Tag/Product Feature: Show isVisible/Hidden circle in Product Table
Screen Shot 2019-07-30 at 7 06 13 PM

machikoyasuda:
Resolves #5392   
Impact: **minor**  
Type: **feature**

## Issue
#5392 

## Solution
Screen Shot 2019-08-01 at 11 19 56 PM
Screen Shot 2019-08-01 at 11 22 42 PM
- use https://github.com/TeamWertarbyte/mdi-material-ui icon, like the Tag table does
- but use MUI styling, rather than Styled-Components (which is what the Tag table does)

## Breaking changes
No

## Testing
1. http://localhost:3000/operator/products - desktop
1. http://localhost:3000/operator/products - mobile

### DIFF
--- a/imports/plugins/included/product-variant/components/productGridItems.js
+++ b/imports/plugins/included/product-variant/components/productGridItems.js
@@ -2,7 +2,7 @@ import React, { Component } from "react";
 import PropTypes from "prop-types";
 import { Link } from "react-router-dom";
 import { formatPriceString, i18next } from "/client/api";
-import { TableCell, TableRow, Checkbox, IconButton, withStyles } from "@material-ui/core"
+import { TableCell, TableRow, Checkbox, IconButton, withStyles } from "@material-ui/core";
 import PencilIcon from "mdi-material-ui/Pencil";
 import CircleIcon from "mdi-material-ui/CheckboxBlankCircle";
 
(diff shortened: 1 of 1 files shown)


imports/plugins/included/product-variant/components/productGridItems.js
```diff
@@ -2,14 +2,26 @@ import React, { Component } from "react";
...
-import IconButton from "@material-ui/core/IconButton";
+import { TableCell, TableRow, Checkbox, IconButton, withStyles } from "@material-ui/core";
 import PencilIcon from "mdi-material-ui/Pencil";
+import CircleIcon from "mdi-material-ui/CheckboxBlankCircle";
+
+const styles = (theme) => ({
+  isVisible: {
+    color: theme.palette.colors.forestGreen300,
+    fontSize: theme.typography.fontSize * 1.25,
+    marginRight: theme.spacing(1)
```
kieckhafer: This is fine, but if the unit is `1` it's not required.

We should probably make a decision on whether to always use a number, even `1`, or leave it blank if it's `1`.

Doesn't matter to me either way, and will be a quick find / replace whichever way we decide.

machikoyasuda: Ah, good to know. Add it to your future Spacing docs :)

PR #5422, opened 2019-08-05
Fix issues with refund queries

aldeed:
Impact: **major**  
Type: **bugfix**

## Issue
Discovered a few issues with recently added refunds GraphQL queries. In general, these are blocking being able to check out and view the order on storefront, and being able to capture the payments on admin (using Stripe).

- There are extra permission checks on `refunds` and `refundsByPaymentIds` queries. `getOrderQuery` util function already does permission checks.
- `imports/plugins/core/orders/server/no-meteor/resolvers/Order/payments.js` is calling `refundsByPaymentId` with `order.token` for the `token` param, but that's the hashed token so it fails.
- When capturing the payment, it fails and there's a server error complaining about the ID namespace not being "reaction/order". I'm guessing somewhere there's a resolver layer that's passing an already-internal ID where an opaque ID is expected.

## Solution
- Removed the extra permission checks
- There's no reason why `refundsByPaymentId` should be looking up the order from DB again in cases where we already have it. Maybe `payments.js` should just pass the whole `order`, and update `refundsByPaymentId` to skip the permission / lookup stuff when the order is passed in.
- Remove comments from the lines of code that convert IDs before calling `captureOrderPayment` from browser

## Breaking changes
None

## Testing
- With Stripe payment set up, verify that you can place an anonymous order from `develop` branch of the example storefront, and then capture that order payment in Reaction Admin.
- Same test but with an account order

### DIFF
--- a/imports/plugins/core/orders/client/graphql/index.js
+++ b/imports/plugins/core/orders/client/graphql/index.js
@@ -20,17 +20,17 @@ export const approveOrderPayments = async ({ orderId, paymentIds, shopId }) => {
     conversionRequests.push({ namespace: "Payment", id: paymentId });
   });
 
-  // const [
-  //   opaqueOrderId,
-  //   opaqueShopId,
-  //   ...opaquePaymentIds
-  // ] = await getOpaqueIds(conversionRequests);
+  const [
+    opaqueOrderId,
+    opaqueShopId,
+    ...opaquePaymentIds
+  ] = await getOpaqueIds(conversionRequests);
 
   return approveOrderPaymentsMutate({
     input: {
-      orderId,
-      paymentIds,
-      shopId
+      orderId: opaqueOrderId,
+      paymentIds: opaquePaymentIds,
+      shopId: opaqueShopId
     }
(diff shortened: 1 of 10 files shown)


kieckhafer:
@aldeed Updated `refundsByPaymentId` to accept an order object, and skip the query of the Orders collection if it's provided: https://github.com/reactioncommerce/reaction/pull/5422/commits/fbbc149830e8d3f6f0a4e93a70bb8ded9629ccd3

aldeed:
@kieckhafer Actually I think GraphQL `refundsByPaymentId` should go back to what it was. Passing in an `order` should be an optimization for internal calls only. We wouldn't want to encourage people to pass an entire order to the server over GraphQL.

kieckhafer:
@aldeed so leave everything how it was, and just add `order` an an optional param? And then in `payments.js` pass that?

aldeed:
@kieckhafer Everything works correctly now for both anon and account orders. I can't approve this because I created it but I'm fine with merging it.

PR #5428, opened 2019-08-07
refactor: expanded product admin permissions

kieckhafer:
Impact: **major**  
Type: **feature**

## Issue
We currently use the `createProduct` as a catch-all for any product related admin action. All users should not have permission to do all things with a product.

## Solution
Granulate permissions so that different users can perform different tasks, and not have all permissions on product admin. 

- To keep backwards compatibility, we are keeping the `createProduct` permission as is, so any existing user groups with `createRefund` permission are not affected.
- We've added a new `product/admin` permission in all places where `createProduct` is used. This will provide forward capability to use this permission as a catch-all / super-admin with a better descriptive name.
- We've also added five new permissions that are added in addition to `product/admin` in places where deemed appropriate: `product/archive`, `product/clone`, `product/create`, `product/publish`, and `product/update`.

#### Thoughts for discussion
- `product/clone` and `product/create` probably _could_ be merged, if that's seen as a better fit, however I can imagine instances where a user should only be allowed to clone a variant, not create a whole new product.
- We could potentially remove `createProduct` and swap it for `product/admin`, instead of just adding `product/admin` in addition, but this would be a breaking change, and should probably be better addressed in an entire permissions overhaul, rather than this.

## Breaking changes
None. We keep our existing `createProduct` permission as is. All new permissions are In addition to `createProduct`.

## Testing
1. Invite a new user to a non-full-admin group (ex. shop manager)
1. Turn off the `createProduct` permission (called `Add product` in our UI), which is the main products permission for the entire app
Accounts

1. Turn on `Product Admin` permission in the new section
1. See that you can perform all Product related updates / deletions / etc, nothing should be different
1. Turn off `Product Admin`
1. Turn on all permissions, except, `Archive Product`. See that the `archive` button is missing from the actions dropdown
1. Turn of `Clone Product`, and see that Duplicate is missing from the dropdown
1. Turn off `Create product` and see that `+` is missing from the variant / option list

Reaction

_Note: You must have either `product/admin`, `product/update`, or `createProduct` permission in order to even see a product page. If none of those are active, you'll see a blank page / error. This is consistent with the current app._

### DIFF
--- a/imports/collections/schemas/shops.js
+++ b/imports/collections/schemas/shops.js
@@ -314,7 +314,7 @@ registerSchema("StorefrontUrls", StorefrontUrls);
  * @property {String} unitsOfMeasure.$.label default value: `Ounces`
  * @property {Boolean} unitsOfMeasure.$.default default value: `false`
  * @property {Metafield[]} metafields optional
- * @property {String[]} defaultSellerRoles default values: `["owner", "admin", "seller", "guest", "manage-users", "orders", "account/profile", "product", "createProduct", "tag", "index", "cart/completed"]`
+ * @property {String[]} defaultSellerRoles default values: `["owner", "admin", "seller", "guest", "manage-users", "orders", "account/profile", "product", "createProduct", "product/admin", tag", "index", "cart/completed"]`
  * @property {Layout[]} layout optional
  * @property {ShopTheme} theme optional
  * @property {BrandAsset[]} brandAssets optional
@@ -496,6 +496,7 @@ export const Shop = new SimpleSchema({
       "account/profile",
       "product",
       "createProduct",
+      "product/admin",
       "tag",
       "index",
       "cart/completed"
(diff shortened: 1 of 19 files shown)


aldeed:
Didn't test, but code 👍

kieckhafer:
@aldeed thanks! @bt3gl could you please do the UI testing portion of this?

bt3gl:
@kieckhafer On it

machikoyasuda:
I'm trying to test this PR locally cos Mia was having trouble.

I got up to the part where I made a new Store Manager w/ the updated Product Admin permissions... and then I tried to log out, by clicking the corner menu button and Log Out. But it immediately logged me back in. I tried opening the site in Safari, Firefox, Chrome, even _Vivaldi_, but same deal. I then manually deleted all my Session cookies. Now I'm stuck in this Zombie mode, where I cannot access the login / log out buttons at all.

Screen Shot 2019-08-09 at 5 54 03 PM

Sooo @kieckhafer What's the trick to logging into several different Admin w/ different Users? Use different browsers? Use different browsers, in Incognito mode? Manually delete session cookies?

bt3gl:
As @machikoyasuda mention, I was never able to fully test this because any time I try to turn anything on/off, I get the error below. No errors appear on logs, so I have no idea how to debug this. 
Screen Shot 2019-08-09 at 8 03 21 PM

machikoyasuda:
I tested the UI on this and it works for me 👍

PR #5431, opened 2019-08-07
Rewrite addAccountToGroup to not call through to Meteor method

### LINKED ISSUE #5336: Rewrite addAccountToGroup to not call through to Meteor method

aldeed:
Resolves #5336   
Impact: **minor**  
Type: **refactor**

## Changes
Rewrites `addAccountToGroup` GraphQL resolver to not use `context.callMeteorMethod`. Adds `addAccountToGroup` internal mutation function and calls that.

## Breaking changes
None

## Testing
Covered by integration tests, but you can manually test `addAccountToGroup` mutation. The UI still uses the Meteor method.

### DIFF
--- a/imports/plugins/core/accounts/server/no-meteor/mutations/addAccountToGroup.js
+++ b/imports/plugins/core/accounts/server/no-meteor/mutations/addAccountToGroup.js
@@ -1,0 +1,90 @@
+import ReactionError from "@reactioncommerce/reaction-error";
+import { difference } from "lodash";
+import SimpleSchema from "simpl-schema";
+import ensureRoles from "../util/ensureRoles";
+
+const inputSchema = new SimpleSchema({
+  accountId: String,
+  groupId: String
+});
+
+/**
+ * @name accounts/addAccountToGroup
+ * @memberof Mutations/Accounts
+ * @method
+ * @summary Add an account to a group
+ * @param {Object} context - GraphQL execution context
+ * @param {Object} input - Input arguments
+ * @param {String} input.accountId - The account ID
+ * @param {String} input.groupId - The group ID
+ * @return {Promise<Object>} with updated address
+ */
+export default async function addAccountToGroup(context, input) {
+  inputSchema.validate(input);
+
+  const { accountId, groupId } = input;
+  const {
+    appEvents,
+    collections: {
+      Accounts,
+      Groups,
+      users
+    },
+    user
+  } = context;
+
+  const group = await Groups.findOne({ _id: groupId });
+  if (!group) throw new ReactionError("not-found", "No group found with that ID");
+
+  const { permissions: groupPermissions = [], shopId } = group;
+
+  // An account can add another account to a group as long as the person adding
+  // has all permissions granted by that group.
+  // We can't use `userHasPermission` here because we want to make sure they
+  // have ALL the permissions rather than ANY.
+  if (difference(groupPermissions, user.roles[shopId] || []).length > 0) {
+    throw new ReactionError("access-denied", "Access Denied");
+  }
+
+  const account = await Accounts.findOne({ _id: accountId });
+  if (!account) throw new ReactionError("not-found", "No account found with that ID");
+
+  // Add all group roles to the user
+  await ensureRoles(context, groupPermissions);
+  await users.updateOne({
+    _id: account.userId
+  }, {
+    $addToSet: {
+      [`roles.${shopId}`]: {
+        $each: groupPermissions
+      }
+    }
+  });
+
+  // Save updated groups list, making sure user only belongs to one group per shop
+  const allGroupsInShop = await Groups.find({
+    shopId
+  }, {
+    projection: {
+      _id: 1
+    }
+  }).toArray();
+  const allGroupIDsInShop = allGroupsInShop.map((grp) => grp._id);
+  const newGroups = (account.groups || []).filter((grp) => allGroupIDsInShop.indexOf(grp) === -1);
+  newGroups.push(groupId);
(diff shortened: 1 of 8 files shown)


imports/plugins/core/accounts/client/containers/accountsDashboardContainer.js
```diff
@@ -2,103 +2,104 @@ import { compose, withProps } from "recompose";
...
 import { Meteor } from "meteor/meteor";
+import getOpaqueIds from "/imports/plugins/core/core/client/util/getOpaqueIds";
+import simpleGraphQLClient from "/imports/plugins/core/graphql/lib/helpers/simpleClient";
 import { Accounts, Groups } from "/lib/collections";
 import { Reaction, i18next } from "/client/api";
 import AccountsDashboard from "../components/accountsDashboard";
+import addAccountToGroupMutation from "./addAccountToGroup.graphql";
+import removeAccountFromGroupMutation from "./removeAccountFromGroup.graphql";
+
+const addAccountToGroupMutate = simpleGraphQLClient.createMutationFunction(addAccountToGroupMutation);
```
kieckhafer: Just making a note to refactor this without `simpleGraphQLClient` in the future.

imports/plugins/core/accounts/client/containers/accountsDashboardContainer.js
```diff
@@ -107,22 +108,14 @@ const composer = (props, onData) => {
...
-      }
-      return admGrps;
-    }, []);
-
-    const adminGroupIds = adminGroups.map((adminGroup) => adminGroup._id);
+    const groups = Groups.find({ shopId }).fetch();
+    const adminGroups = groups.filter((group) => group.slug !== "customer" && group.slug !== "guest");
 
+    // Find all accounts that are in any of the admin groups
+    const adminGroupIds = adminGroups.map((group) => group._id);
```
kieckhafer: Looks familiar :grin:

imports/plugins/core/accounts/server/no-meteor/mutations/addAccountToGroup.js
```diff
@@ -0,0 +1,117 @@
...
+  const group = await Groups.findOne({ _id: groupId });
+  if (!group) throw new ReactionError("not-found", "No group found with that ID");
+
+  const { permissions: groupPermissions = [], shopId } = group;
+
+  // An account can add another account to a group as long as the person adding
+  // has all permissions granted by that group.
+  // We can't use `userHasPermission` here because we want to make sure they
+  // have ALL the permissions rather than ANY.
+  if (!context.isInternalCall && difference(groupPermissions, user.roles[shopId] || []).length > 0) {
```
kieckhafer: Do we need to import`lodash` here for this?

Could we do something like `groupPermissions.filter((permission) => !user.roles[shopId].includes(permission))` instead?

I know at one point our policy was "lodash only if needed", if that's changed, then `difference` is cleaner looking.

imports/plugins/core/components/lib/hoc.js
```diff
@@ -1,8 +1,8 @@
-import _ from "lodash";
+/* eslint-disable node/no-missing-require, node/no-unpublished-require */
```
kieckhafer: These were some of the new rules added when we did the `recommended` import of `node` lint rules.

We should discuss if we want to keep these rules in our `eslint` config, I know @mikemurray has run across these lint errors a few times too.

kieckhafer:
@aldeed 

> Removes some browser-side permission checks. It's possible you'll now see buttons on Accounts page even if you don't have permission to use them. You'll get an error if you try.

Is this what we'll want to do going forward? In my recent updates, especially in the orders panel, I've done the opposite and purposely removed UI elements like buttons if a user doesn't have permission.

aldeed:
@kieckhafer I have always been using .graphql for the client gql unless it was tiny and could be `gql` template in the same file. My thought was that it's easier that way since editors support formatting, linting, etc. in `.graphql` files. That being said, I believe I now have an editor extension that does the same formatting and highlighting in `gql` tags, so it may not matter. Using a `js` file has the benefit of working without Babel or Webpack. It should probably be discussed more widely and document a decision. In the meantime, it doesn't bother me to have two different ways.

Regarding client side permissions, I do think we should have a way for people to have UI permissions, but it will eventually need to be a completely separate permission system. We're working toward the API permissions being a completely separate auth layer, so when that happens, there will be no way to find out a user's auth permissions in browser code. In the meantime it seemed easiest to just remove such checks, but maybe we should prioritize building a simple UI permissions system? It could be as simple as an array of strings (similar to roles) on an `account`, which would correspond with admin UI areas they're allowed to see.

Regarding the error you got, what permissions are on the new group? It should only show access denied if you are missing ANY permission that you're trying to grant to another person. I don't know if that's the best way to do it, but that's how it's been. The only difference is that I got rid of the logic that said "owner can do anything" because I assumed "owner" would have all the permissions they're trying to grant. Maybe that isn't true?

imports/plugins/core/accounts/server/no-meteor/mutations/addAccountToGroup.js
aldeed: No, lodash isn't needed.

imports/plugins/core/accounts/client/containers/accountsDashboardContainer.js
aldeed: Yeah I didn't want to go too far down the rabbit hole.

kieckhafer:
@aldeed I think the "Owner should be able to do anything" might be it. I am an owner trying to move the user from `Shop Manager` to `Operator`, which is a new group I made, with less permissions than the user had before.

PR #5433, opened 2019-08-07
fix partial product publish error

### LINKED ISSUE #5414: Not able to publish product
## Issue Description
When creating a new product, I am unable to publish it.

## Steps to Reproduce
1. Start Reaction in latest develop
2. Create a new product and enable inventory tracking for an option
3. Attempt to publish product

Generates the following error:
```
Error in "afterInventoryUpdate" consumer TypeError: Cannot destructure property `product` of 'undefined' or 'null'.
reaction_1  |       at Promise.asyncApply (imports/plugins/core/catalog/server/no-meteor/mutations/partialProductPublish.js:9:88)
```
Scenario 2.
1. Start Reaction on latest develop
2. Create a new product and do NOT enable inventory tracking for variant nor option
3. Attempt to publish product

Generates the following error:
```
Exception while invoking method 'catalog/publish/products' Error: Input passed into "inventoryForProductConfigurations" was invalid: inventoryInputInvalidPropertyValue[productConfigurations]
reaction_1  |     at Promise.asyncApply (imports/plugins/core/inventory/server/no-meteor/queries/inventoryForProductConfigurations.js:120:11)
```

## Possible Solution
Before calling the new partial publish function, it's necessary to ensure that inventory tracking is actually enabled. Further, do not publish inventory props when inventory tracking is NOT enable. 

Just my quick thoughts, might be the solution, might not be.

willopez:
Resolves #5414   
Impact: minor
Type: bugfix

## Issue
When attempting to publish a product and itself and variants are invisible, an ambiguous error would occur without notifying the user of the cause.

## Solution
Prevent publishing inventory information when variants are not visible.

# Testing
1. Create a new product with variants + options and leave all set to Invisible
2.  Attempt to publish
3. Observe notification informing user that the product and variants or not visible.

### DIFF
--- a/imports/plugins/core/catalog/server/no-meteor/mutations/publishProducts.js
+++ b/imports/plugins/core/catalog/server/no-meteor/mutations/publishProducts.js
@@ -39,7 +39,10 @@ export default async function publishProducts(context, productIds) {
   const success = await publishProductsToCatalog(productIds, context);
   if (!success) {
     Logger.error("Some Products could not be published to the Catalog.");
-    throw new ReactionError("server-error", "Some Products could not be published to the Catalog. Make sure your variants are visible.");
+    throw new ReactionError(
+      "server-error",
+      "Some Products could not be published to the Catalog. Make sure the parent product and its variants and options are visible."
+    );
   }
   return Catalog.find({ "product.productId": { $in: productIds } }).toArray();
 }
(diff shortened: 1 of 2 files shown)


willopez:
@kieckhafer Not able to reproduce your error. Make sure you did not already have a product in the `Catalog` collection with the same id. This is how I test this fix.

https://recordit.co/6ZekII3oJN

kieckhafer:
Your video is exactly what I did. Will try retesting, seems very unlikely I would be able to create a new product with the same ID.

kieckhafer:
@willopez  Still seeing the issue
no publish mov

PR #5434, opened 2019-08-07
ref: move filter outside of admin check

kieckhafer:
Impact: **minor**  
Type: **refactor**

## Issue
Tag filter implemented in #5232 was only accessible by admin accounts, as it was inside of a permissions check.

## Solution
As we've had some requests for all users to be able to use this, we've moved the filter ability outside the admin permission check.

@mikemurray and I initially decided to only allow admin users the permission, as we wanted to further discuss performance impacts as this hits the database frequently. If performance becomes an issue here, we can revisit.

## Breaking changes
None

## Testing
1. Login to the admin as a non-owner / admin (a `Shop Manager`).
1. See that the filter works as described here: https://github.com/reactioncommerce/reaction/pull/5232


aldeed:
Code 👍 , didn't test

PR #5436, opened 2019-08-07
refactor: add tag permissions to allow non-owner to edit

kieckhafer:
Impact: **minor**  
Type: **refactor**

## Issue
Operators that didnt' have full `admin` or `owner` permissions could not edit tags.

## Solution
Add two new permissions which will granulize tag permissions to allow other users the ability to edit tags.

## Breaking changes
None. These are new permissions that won't affect existing permissions

## Testing
1. Create a non-admin / owner (Shop Manager) account
1. See that you can edit tags if the correct permissions are turned on
Accounts


kieckhafer:
Thanks @aldeed. @machikoyasuda could you please do the ui / testing portion of this ticket?

PR #5442, opened 2019-08-12
feat: add new setting to allow / disallow custom user locale lookup

kieckhafer:
Impact: **minor**  
Type: **feature|refactor**

## Issue
When checking a users locale, we hit an external server every time the app refreshes, which can be taxing on performance, and cost.

## Solution
Add an option to not allow custom user locales, which will skip this check and use the default store locale. This can be beneficial for any shop which doesn't wish to take the performance hit of the constant external server hit, and also for shops which only have one language / currency and a locale check is not needed.

## Breaking changes
None. By default, the old settings of checking the locale are enable.

## Testing
Since we [don't have an IP address that's locale based](https://github.com/reactioncommerce/reaction/blob/refactor-kieckhafer-customUserLocaleOption/imports/plugins/core/core/server/util/geocoder.js#L113) when developing locally, the easiest way to test is to `console.log` inside of this check, and see that the log shows / doesn't show depending on the settings: https://github.com/reactioncommerce/reaction/blob/refactor-kieckhafer-customUserLocaleOption/imports/plugins/core/core/server/methods/shop/getLocale.js#L57

### DIFF
--- a/imports/plugins/core/i18n/client/components/localizationSettings.js
+++ b/imports/plugins/core/i18n/client/components/localizationSettings.js
@@ -199,6 +199,9 @@ class LocalizationSettings extends Component {
               language: {
                 type: "select",
                 options: this.props.enabledLanguages
+              },
+              allowCustomUserLocale: {
+                type: "boolean",
               }
             }}
             name="localization"
--- a/imports/plugins/core/i18n/client/components/localizationSettings.js
+++ b/imports/plugins/core/i18n/client/components/localizationSettings.js
@@ -201,7 +201,7 @@ class LocalizationSettings extends Component {
                 options: this.props.enabledLanguages
               },
               allowCustomUserLocale: {
-                type: "boolean",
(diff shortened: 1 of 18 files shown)


imports/plugins/core/i18n/client/components/localizationSettings.js
```diff
@@ -199,6 +199,9 @@ class LocalizationSettings extends Component {
               language: {
                 type: "select",
                 options: this.props.enabledLanguages
+              },
+              allowCustomUserLocale: {
+                type: "boolean",
```
rosshadden: Comma dangle. I am a fan of this, but our eslint rules are not and are failing.

kieckhafer:
Something funky is going on, I made a push of some changes, which I can see on GitHub in the branch, but for some reason they aren't showing up in this PR, and those changes fix the failing CI. So I'm closing this in favor of ___ which has my extra commit in it.

PR #5461, opened 2019-08-16
Feat bulk actions UI updates

cshepherd:
Impact: **minor**  
Type: **bugfix**

## Issue
This includes the following UI fixes for Bulk Actions:
- Navigational dead end when no products match the uploaded CSVs: Instead the UI is reset and a red (error) InlineAlert is shown saying that no products were found.
- Proper spacing (16px) between filename chips
- If CSV contains some (but not all) product ids that weren't matched, a red (error) InlineAlert is shown after the green one with the count of products not found.
- All filtered products are automatically selected

## Breaking changes
n/a

## Testing
Verify the above are properly implemented and styled

### DIFF
--- a/imports/plugins/included/product-admin/client/components/ProductTable.js
+++ b/imports/plugins/included/product-admin/client/components/ProductTable.js
@@ -136,16 +136,16 @@ function ProductTable({ onCreateProduct }) {
     } else {
       setFilteredProductIdsCount(count);
     }
-    return;
   };
 
   const iconComponents = {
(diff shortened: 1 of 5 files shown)


imports/plugins/included/product-admin/client/components/ProductTable.js
```diff
@@ -8,13 +8,17 @@ import ImportIcon from "mdi-material-ui/Download";
 import { useDropzone } from "react-dropzone";
 import { i18next } from "/client/api";
 import { Session } from "meteor/session";
+import _ from "lodash";
```
machikoyasuda: You can also do: `import { uniq } from "lodash";` if you're only using `uniq`

machikoyasuda:
@willopez you can review + merge this! :D

PR #5466, opened 2019-08-18
load account into context during surcharges

cmbirk:
Original PR here: https://github.com/sportsdirect/reaction/pull/498

Resolves(but doesn't close) sportsdirect/reaction-plugin-pricing-engine#48
Impact: critical
Type: bug

Issue
Loads the account into the context object in the surcharges afterCartUpdate event. This is needed in context to choose the correct price in getVariantPrice from the pricing engine. It was previously calculating tax based on the full priced item instead of the member price, which was causing a mismatch at checkout time of the expected pricing.

cc @rosshadden @jrphilo @kkuzemka


imports/plugins/included/surcharges/server/no-meteor/startup.js
```diff
@@ -23,6 +23,12 @@ export default function startup(context) {
     const cartSurcharges = [];
 
     // Merge surcharges from each shipping group
+    if (cart.accountId && !context.account) {
+      const { Accounts } = context.collections;
+      const account = await Accounts.findOne({ _id: cart.accountId });
+      context.account = account;
```
aldeed: The cart account may not be the same as the request account, and may have different permissions. Also, by mutating the “base” context, this value might stick around longer than you’re expecting and be used in subsequent event handler calls.

If you instead do `contextWithAccount = { ...context, account, accountId: cart.accountId };` and then pass through `contextWithAccount` instead of `context` below, that's safe, though potentially not the best fix in the long run.

cmbirk: Should be fixed in bdc57ed50

cmbirk:
@aldeed this should be good to go. Have tested that member priced items can go through the entire checkout flow

PR #5469, opened 2019-08-20
[BLOCKED] refactor: use SplitButton from Catalyst instead of ConrifmButton

### LINKED ISSUE #5421: swap ConfirmButton with Catalyst ConfirmDialog in OrderRefunds component
See https://github.com/reactioncommerce/reaction/pull/5405#discussion_r310749061

kieckhafer:
Currently [BLOCKED] as this needs Catalyst v1.5+ to be ready for use inside Reaction.

Resolves #5421  
Impact: **minor**  
Type: **refactor**

## Issue
At the time of original development, `SplitButton` was not yet available, therefore not used in places it should have been. 

## Solution
Update the UI to use `SplitButton` where needed, since it's now available in Catalyst.

## Breaking changes
None

## Testing
1. Create an Order
1. See that the `SplitButton` component is now used in the `Fulfillment Group` section of the Orders Admin
1. See that the `SplitButton` works by changing shipment status as it should

### DIFF
--- a/imports/plugins/core/orders/client/components/OrderCardFulfillmentGroupStatusButton.js
+++ b/imports/plugins/core/orders/client/components/OrderCardFulfillmentGroupStatusButton.js
@@ -2,14 +2,9 @@ import React, { Component } from "react";
 import PropTypes from "prop-types";
 import { Mutation } from "react-apollo";
 import withStyles from "@material-ui/core/styles/withStyles";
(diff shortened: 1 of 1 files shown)


kieckhafer:
@reactioncommerce/design I'd like to show you how this works in action. It's a strange workflow, having to select something from the dropdown, and then also click the button. I feel like maybe the original way, where a dialog opens, is better.

PR #5474, opened 2019-08-22
FIX: Remove cart items should return updated cart

ajporlante:
Resolves N/A
Impact: **critical**  
Type: **bugfix**

## Issue
The cart returned when items are removed from cart does not have the correct taxes.

## Solution
Wait for the `afterCartUpdate` updates to finish before returning the cart to the client.

My code changes are based on how `updateCartItemsQuantity` mutation handled returning the updated cart.

## Breaking changes
None

## Testing
Go to checkout with multiple items in cart. Enter a shipping address that will result to taxes and enter shipping method. Go back to cart page. Remove an item in the cart. See that the tax is updated in the summary.


chrispotter:
@ajporlante please sign your commit and i'll work on updating the unit-tests

aldeed:
I refactored this function in #5477 and I believe that change will also fix this issue in a different way.

ajporlante:
Closing this in favor of #5477

PR #5477, opened 2019-08-23
Refactor after-publish cart updates for speed

### LINKED ISSUE #5098: Publishing products causes performance issues
## Prerequisites

## Issue Description
When products are published in the Catalyst UI an `afterPublishProductToCatalog` app event is emitted. The core cart plugin registers a function to update all carts that have a variant of the published products. This leads to noticeable performance degradation in the Catalyst UI.

Reference: 
https://github.com/reactioncommerce/reaction/blob/develop/imports/plugins/core/cart/server/no-meteor/startup.js#L144

## Steps to Reproduce
1. Load sample test data
2. Create multiple different carts and add various different products to it
3. Republished products that have been added to carts

## Possible Solution
Batch updates to cart on start up, don't use `Promise.all`

aldeed:
Resolves #5098   
Impact: **minor**  
Type: **performance**

### After-cart-update performance improvements
There were several "afterCartUpdate" and "afterCartCreate" listeners in various plugins that were essentially transforming the cart but doing it without any coordination was chaotic and slow. Now there is instead a single cart transform function in the Cart plugin, which always runs before every cart insert or update. Plugins now register custom transforms for this using a pattern like this in `registerPlugin`:

```js
    cart: {
      transforms: [
        {
          name: "setTaxesOnCart",
          fn: setTaxesOnCart,
          priority: 30
        }
      ]
    },
```

The main cart plugin transform then runs these custom ones in priority order. This is a bit faster, uses less memory, and resulted in consolidated and cleaner code in the Cart plugin.

### After-publish performance improvements
The cart plugin has a `afterPublishProductToCatalog` listener that updates cart item properties from catalog variant properties and, if any have changed, saves the cart. We found that this could be very slow and crash with a Mongo connection error when there are thousands of carts that contain the changed variant.

Now the cart lookups and updates are batched so that memory usage and speed are optimal. This combined with the improved cart transformation flow described above appear to be ~10x speed improvement when publishing to catalog, based on local tests with 8,000 carts.

### Other
- While testing I noticed that Reaction Admin now allows setting tax info for variant options, but when publishing, only the parent variant tax info was copied. Now it copies the option info with fallback to parent variant for backwards compatibility.
- `getSurcharges` was calling `extendCommonOrder` even when there were no surcharges defined. Now it exits early if there are no surcharges, which makes publishing and cart updates faster on shops that don't use surcharges.
- Much better logging for the post-publish and cart update flows

## Breaking changes
None

## Testing
Test publishing price and/or tax code changes to a variant or option that is in one or more carts. It should work and be faster than before.

You could also test this with a variant that is in 5-10k carts to verify it no longer crashes (but I've tested this thoroughly).

Almost all code related to carts has been moved around, though is mostly unchanged. The best way to test this is with something like the following smoke test flow in the example storefront:
1. add first item to cart when not logged in
1. add additional item to cart when not logged in
1. log in / register
1. log out
1. add same item and different item to anonymous cart
1. log in
1. verify all items are merged in account cart and duplicate item qty has been increased
1. remove an item from cart
1. change qty of an item in cart
1. check out the cart


imports/plugins/core/cart/server/no-meteor/mutations/saveManyCarts.js
```diff
@@ -0,0 +1,69 @@
...
+        upsert: true
+      }
+    };
+  });
+
+  const bulkWrites = await Promise.all(bulkWritePromises);
+
+  let writeErrors;
+  try {
+    Logger.trace({ ...logCtx, bulkWrites }, "Running bulk op");
```
zenweasel: Yay, for use of `Logger.trace`

imports/plugins/core/cart/server/no-meteor/registration.js
```diff
@@ -0,0 +1,17 @@
...
+ * @param {Object} options The options object that the plugin passed to registerPackage
+ * @returns {undefined}
+ */
+export function registerPluginHandler({ cart }) {
+  if (cart) {
+    const { transforms } = cart;
+    if (Array.isArray(transforms)) {
+      cartTransforms.push(...transforms);
+      cartTransforms.sort((prev, next) => prev.priority - next.priority);
+    }
```
zenweasel: This check just silently rejects a non-array? Should it create an array or throw an error?

imports/plugins/core/cart/server/no-meteor/startup.js
```diff
@@ -29,47 +30,60 @@ function getFlatVariantsAndOptions(catalogProductVariants) {
...
+   * @return {undefined}
+   */
+  async function saveCarts() {
+    if (updatedCarts.length === 0) return;
+    await mutations.saveManyCarts(context, updatedCarts);
+    updatedCarts = [];
+  }
+
+  /**
+   * @summary Get updated prices for a single cart, and check whether there are any changes.
```
zenweasel: I think we should be adding/setting a flag here for when the cart has been modified so that the UX could display some sort of message.

imports/plugins/core/cart/server/no-meteor/util/updateCartItemsForVariantChanges.js
```diff
@@ -5,15 +5,27 @@
...
  * @param {Object} prices Various updated price info for this variant
  * @returns {Object} { didUpdate, updatedItems }
  */
-export default function updateCartItemsForVariantPriceChange(items, variantId, prices) {
+export default function updateCartItemsForVariantChanges(items, catalogProductVariant, prices) {
   let didUpdate = false;
 
   const updatedItems = items.map((item) => {
-    if (item.variantId !== variantId) return item;
+    if (item.variantId !== catalogProductVariant.variantId) return item;
```
zenweasel: Should we also be checking for stuff like weight which would affect shipping calculations?

PR #5487, opened 2019-08-28
fixes #5486

### LINKED ISSUE #5486: In tags list table individually tag are not getting unselected after selecting through checkbox
## Issue Description
In /operator/tags page tags list table in not getting unselected through checkbox after getting selected by checkbox.

## fix
replace old function
`handleToggleSelection = (key, shift, row) => {
    // update the state
      // start off with the existing state
      let selection = prevState.selection;
      // let selectedObjects = { ...prevState.selectedObjects }

      const keyIndex = selection.findIndex((element) => 'select-'+element._id === key);

      // check to see if the key exists
      if (keyIndex === -1) {
        // it does not exist so add it
        selection.push(row);
      } else {
        // it does exist so we will remove it using destructing
        selection.splice(keyIndex, 1);
      }

      return {
        selection,
        selectAll: false,
        selectedBulkAction: selection.length === 0 ? undefined : prevState.selectedBulkAction
      };
    });
  }`

### Versions
2.1.0

monarkatfactly:
Resolves #5486  
Impact: **minor**  
Type: **bugfix**

## Issue
Refer #5486 

## Solution
Changing the way findOne function was working and the way array was getting updated

## Breaking changes
none

### DIFF
--- a/imports/plugins/core/tags/client/components/TagDataTable.js
+++ b/imports/plugins/core/tags/client/components/TagDataTable.js
@@ -393,21 +393,18 @@ class TagDataTable extends Component {
     // update the state
     this.setState((prevState) => {
       // start off with the existing state
-      let selection = [...prevState.selection];
+      let selection = prevState.selection;
       // let selectedObjects = { ...prevState.selectedObjects }
 
-      const keyIndex = selection.findIndex((element) => element._id === key);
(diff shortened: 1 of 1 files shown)


monarkatfactly:
@spencern 
Due some changes I had to delete this PR, but I have created new PR https://github.com/reactioncommerce/reaction/pull/5496
In this new PR I have fixed the bug.

PR #5492, opened 2019-08-28
refactor: remove addressBookRemove meteor method and rewrite with GraphQL

### LINKED ISSUE #5338: Rewrite removeAccountAddressBookEntry to not call through to Meteor method
- Rewrite `removeAccountAddressBookEntry` GraphQL resolver to not use `context.callMeteorMethod`.

kieckhafer:
Resolves #5338  
Impact: **minor**  
Type: **feature|refactor**

## Issue
`addressBookRemove` still uses a meteor method to perform it's actions.

## Solution
Rewrite `addressBookRemove` mutation to use GraphQL instead of meteor method.
- Remove `addressBookRemove` meteor method from app

## Breaking changes
Nothing in core code. Any custom code calling the `accounts/addressBookRemove` Meteor method will break.

## Testing
1. Add an address to an account
1. Remove the address from the account
1. See that this works, with GQL now that the meteor method is removed
_This can be done through Example Storefront, with a logged in account_


kieckhafer:
@aldeed not sure if you've looked at this yet, but made some last minute changes to match some comment you made here: https://github.com/reactioncommerce/reaction/pull/5484

imports/plugins/core/accounts/server/no-meteor/mutations/removeAccountAddressBookEntry.js
```diff
@@ -0,0 +1,63 @@
...
+    returnOriginal: false
+  });
+
+  if (!updatedAccount) {
+    throw new ReactionError("server-error", "Unable to remove address from account");
+  }
+
+  await appEvents.emit("afterAccountUpdate", {
+    account: updatedAccount,
+    updatedBy: accountId
```
aldeed: `updatedBy` should be `userIdFromContext`, not the account that was updated.

kieckhafer:
@aldeed comment addressed

aldeed:
OK to merge if all green.

PR #5496, opened 2019-08-29
fixes #5486

### LINKED ISSUE #5486: In tags list table individually tag are not getting unselected after selecting through checkbox

monarkatfactly:
Resolves #5486   
Impact: **minor**  
Type: **bugfix**

## Issue
At /operator/tags route, in tags list table tag not getting unselected through checkbox after getting selected by checkbox.

## Solution
1) Change the assign of variable at here with
    `let selection = prevState.selection;`

2) Change the findOne function here with
     `const keyIndex = selection.findIndex((element) => 'select-'+element._id === key);`

3) Change the if at here condition with
    `if (keyIndex === -1) { selection.push(row); } else { selection.splice(keyIndex, 1); }`

### DIFF
--- a/imports/plugins/core/tags/client/components/TagDataTable.js
+++ b/imports/plugins/core/tags/client/components/TagDataTable.js
@@ -393,9 +393,9 @@ class TagDataTable extends Component {
     // update the state
     this.setState((prevState) => {
       // start off with the existing state
-      let selection = prevState.selection;ta
+      const { selection } = prevState;
 
-      const keyIndex = selection.findIndex((element) => 'select-'+element._id === key);
+      const keyIndex = selection.findIndex((element) => `select-${element._id}` === key);
 
       // check to see if the key exists
       if (keyIndex === -1) {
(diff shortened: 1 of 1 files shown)


kieckhafer:
@mikemurray @machikoyasuda is this addressed in Catalyst? Can we close?

machikoyasuda:
Hi @monarkatfactly - Thank you for filing this bug ticket and creating a solution for it. We're currently creating a new React component for all DataTables, including this one. We'll keep the issue open, but close this PR because we plan to use the new DataTable for Tags soon. Thank you.

Issue #5373, opened 2019-07-25
Add Integration Test for orderById GraphQL query


bt3gl:
Epic: #5310

bt3gl:
Original function is in: https://github.com/reactioncommerce/reaction/tree/master/imports/plugins/core/orders/server/no-meteor/queries

PR #5511, opened 2019-09-02
Add Integration Test for orderById GraphQL query

bt3gl:
Resolves: #5373 
Impact: **minor**
Type: **test**

This PR adds: an integration test file for [orderById query](https://github.com/reactioncommerce/reaction/blob/f1880a2b59574eb898c972819b8177c32fab479e/imports/plugins/core/orders/server/no-meteor/queries/orderById.js). 

We test whether an order is successfully returned, given an order id and a shop id. Here is the [schema](https://github.com/reactioncommerce/reaction/blob/f1880a2b59574eb898c972819b8177c32fab479e/imports/plugins/core/orders/server/no-meteor/schemas/schema.graphql) for this query. We test for auth-ed and anom user (`setLoggedInUser` will fake as if your auth token is for that user).

#### Testing Instructions:

```
npm install --save-dev jest-cli
npm run lint
npm run test
```


bt3gl:
Running the instructions for DCO is not fixing the problem. Not sure what to do.
Screen Shot 2019-09-03 at 9 56 17 PM

tests/integration/api/queries/orderById/orderById.test.js
```diff
@@ -0,0 +1,84 @@
...
+  await testApp.setLoggedInUser(mockOrdersAccount);
+  const result = await query({ id: opaqueOrderId, shopId: opaqueShopId, token: null });
+  expect(result.orderById.account._id).toBe(opaqueAccountId);
+  expect(result.orderById.shop._id).toBe(opaqueShopId);
+  expect(result.orderById.shop.name).toBe(shopName);
+});
+
+/*
+TODO: I am getting a null result with this query, so this test is failing.
+Need to find out why.
```
aldeed: I see you are passing in `token: null`. Did you try `token: tokenInfo.token` and are still getting null?

bt3gl: @aldeed Yes, I tried `tokenInfo.token` as well and still getting null?

aldeed: @bt3gl I ran this and I see the issue now. When you call `testApp.setLoggedInUser`, it persists for all tests in this same file, so it's looking for an account order again rather than an anonymous order.

You can fix by calling `await testApp.clearLoggedInUser();`, either at the top of your second test or in a `beforeEach`.

After you do that, you'll get a different error. To fix that error, add `accountId: null` in your `makeOne` call for `orderAnon`. If you don't explicitly set the field to `null`, the factory sets it to default string data "mockAccountId", which makes it think it's an account order.

Then it should pass!

bt3gl:
@aldeed This is great, thank you. The PR is updated and ready for review/merge. Mahalo!

PR #5514, opened 2019-09-03
feat: Translations without Meteor

aldeed:
Impact: **minor**  
Type: **feature**

## Issue
Reaction Admin i18next loads translations from a Translations collection, loaded by a Meteor DDP subscription. We are trying to remove the Meteor dependency from the API. 

## Solution
- The API (specifically the `i18n` core plugin) now provides translations, merged from all plugins, at the standard `/locales/resources.json` route.
- Because any plugin can add a translation namespace, the API (specifically the `i18n` core plugin) also provides a `/locales/namespaces.json` route, which returns the namespaces array. The i18next client library needs this list for initialization.
- In browser code, Reaction Admin i18next lib now loads translations in a standard way, per their docs.
- In plugins, translation files are unchanged but have been moved into the `no-meteor` folder and no longer use `loadTranslations` function. Instead, they pass translations array to `registerPlugin` within an `i18n` object.
- The API no longer loads translations from `/private` (this was a Meteor thing). Those translations have been moved into `core` plugin.
- The API no longer saves translations to the `Translations` collection on startup, no longer publishes that collection, and no longer supports "flushing" translations because they are now always automatically up to date.
- A migration removes all documents from the `Translations` collection.

## Other changes
- A more limited and standard language detection is now used by Reaction Admin. It works like this:
  - You will see UI text in whatever language is set as your browser default. For example, in Chrome, go to chrome://settings/languages and whatever language is listed first in your preferences list will be the language used.
  - To see the UI in a specific language, overriding your browser language, you can append `lng=<code>` in your query string. For example, `http://localhost:3000/operator/?lng=en`
 or `http://localhost:3000/operator/?lng=en-US` will show you English text even if your browser language is set to something else.
  - If some bit of text does not have a translation for your language, you will see it in English. (Consider submitting a pull request to add the missing translation!)
- The `shop/getLocal` Meteor method is removed.

## Breaking changes
- Any custom plugins relying on translations being in the Translations collection will need to be updated.
- Any custom plugins importing and using `loadTranslations` to load translations will need to be updated to pass them to `registerPlugin` instead. See the built-in plugins for examples. 
- Any custom plugins calling `shop/getLocale` Meteor method or using `Reaction.Locale` object will need changes. `i18next.language` should be used to get the detected/default language.

## Testing
1. Change your browser default/first preferred language to any language Reaction supports. For example, in Chrome, go to chrome://settings/languages and edit your list.
1. Go to http://localhost:3000/operator/ and verify that text is shown in that language (other than places where there is no translation for that language, which will be English)
1. Go to http://localhost:3000/operator/?lng=en (or any language other than your browser default) and verify that text shows in that language now.
1. Click through all possible paths in the Reaction Admin UI and make sure no text is missing or showing only the translation key rather than the text.


imports/plugins/included/payments-stripe/server/no-meteor/i18n/en.json
```diff
@@ -3,7 +3,7 @@
   "i18n": "en",
   "ns": "reaction-stripe",
   "translation": {
-    "reaction-payments": {
+    "reaction-stripe": {
```
kieckhafer: Same comment for all translation files in this package: Should this be `reaction-payments-stripe` or just `payments-stripe` to match the package name?

imports/plugins/included/payments-example/server/no-meteor/i18n/en.json
```diff
@@ -3,7 +3,7 @@
   "i18n": "en",
   "ns": "example-paymentmethod",
   "translation": {
-    "reaction-payments": {
+    "example-paymentmethod": {
```
kieckhafer: Same comment for all translation files in this package: Should this be `reaction-payments-example` or just `payments-example` to match the package name?

imports/plugins/included/payments-stripe/server/no-meteor/i18n/en.json
aldeed: It's supposed to match `ns` in this same file, so that's why I changed it. There were a few files that were probably copied and were not validly formatted, which didn't matter with the way we were loading before, but now caused errors.

I didn't give much thought to what the namespace _should_ be other than fixing the mismatch bug. I think the general rule would probably be to match identically the package name, but our package names are also a bit inconsistent.

@mikemurray Do you foresee any problems if I audit these namespaces and ensure they exactly match the package names?

imports/plugins/included/payments-example/server/no-meteor/i18n/en.json
aldeed: Discussion in the other comment.

imports/plugins/included/payments-stripe/server/no-meteor/i18n/en.json
mikemurray: I think it should be fine to make them match the package. As far as I remember, the name is supposed to match the package anyway, and things may have gotten out of sync with all the copy-pasting of translations around the app.

imports/plugins/core/core/server/publications/collections/shops.js
```diff
@@ -1,7 +1,4 @@
 import { Meteor } from "meteor/meteor";
-import { check } from "meteor/check";
-import Reaction from "/imports/plugins/core/core/server/Reaction";
-import ReactionError from "@reactioncommerce/reaction-error";
 import { Shops } from "/lib/collections";
 
 // We should be able to publish just the enabled languages/currencies/
```
kieckhafer: I don't think this comment is relevant anymore.

imports/plugins/core/core/server/methods/shop/updateLanguageConfiguration.js
```diff
@@ -22,12 +22,16 @@ export default function updateLanguageConfiguration(language, enabled) {
...
-  const shop = Shops.findOne({
-    _id: Reaction.getShopId()
-  });
+  const shopId = Reaction.getShopId();
+
+  const shop = Shops.findOne({ _id: shopId });
 
   const defaultLanguage = shop.language;
 
+  if (language === defaultLanguage && !enabled) {
```
kieckhafer: If I use the "all languages" toggle, the default language (English in my case) stays enabled, but I am able to turn off `English` if I toggle it individually.

When I refresh the page, `English` is still enabled, so this seems to just be a UI issue that no error is shown.

I don't think a blocker for this, but if it's an easy fix I'd tackle it now.
languages mov

imports/plugins/core/versions/server/migrations/73_remove_translations_data.js
```diff
@@ -0,0 +1,18 @@
...
+const { db } = MongoInternals.defaultRemoteCollectionDriver().mongo;
+
+const Translations = db.collection("Translations");
+
+Migrations.add({
+  version: 73,
+  async up() {
+    const { Assets } = rawCollections;
+
+    await Assets.deleteMany({ type: "i18n" });
```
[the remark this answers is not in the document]
aldeed: Good point. I can do `dropCollection` instead.

imports/plugins/core/core/server/methods/shop/updateLanguageConfiguration.js
aldeed: I'll see if it's easy to show the error.

imports/plugins/included/payments-stripe/server/no-meteor/i18n/en.json
aldeed: After reviewing all, they actually do all match the package names. So we should try to improve the consistency of the package names, but that's something I won't tackle here.

imports/plugins/core/core/server/methods/shop/updateLanguageConfiguration.js
aldeed: Fixed in https://github.com/reactioncommerce/reaction/pull/5514/commits/f9c63580b95ce21af4a75d5f41632ca14bc9ca73 and fixed for the currency list, too.

mikemurray:
@aldeed In some custom plugins I have I'm seeing this error `reaction_1  | Error: Cannot find module '../../../../core/core/server/startup/i18n'` which is noted in the breaking changes.

Should we consider a method of deprecating the old implementation rather than a hard breaking change? Maybe at the least have the file just mock the methods used and log a useful message for upgrading?

PR #5517, opened 2019-09-03
Release v2.2.1

### LINKED ISSUE #5513: Checkout hangs on shipping methods step
## Prerequisites
* [ ✓ ] Are you running the latest version?
* [ ✓ ] Are you able to consistently reproduce the issue?
* [ ✓ ] Did you search the issue queue for existing issue? [Search issues](https://github.com/reactioncommerce/reaction/issues)

## Issue Description
Running the latest release `2.2.0` and flat rate shipping configured with Operator I see an error:

`
GraphQL Error: "shipping.0.shipmentQuotes.0.method.fulfillmentTypes is not allowed by the schema"`

when I run `updateFulfillmentOptionsForGroup` GraphQL mutation.

## Steps to Reproduce

0. Anonymous user checking out
1. Proceed to shipping step, in particular choosing shipping method (`updateFulfillmentOptionsForGroup` mutation)
2. I see an infinite spinner because mutation fails with an error `shipping.0.shipmentQuotes.0.method.fulfillmentTypes`

## Possible Solution
This issues seems to be caused by explicit cart validation introduced in this commit here https://github.com/reactioncommerce/reaction/commit/7d2ef0e4e0a3ddcebbbf002429f5548b0d7b0f63

`saveCart` mutation used internally now always validates the cart according to the schema defined in https://github.com/reactioncommerce/reaction/blob/develop/imports/collections/schemas/cart.js

The `fulfillmentTypes` field does not exist in `ShippingMethod` schema, but is always persisted via Operator tools when new flat rate shipping method is added but exists in a different schema here https://github.com/reactioncommerce/reaction/blob/develop/imports/plugins/included/shipping-rates/server/no-meteor/util/methodSchema.js

The solution involves adding `fulfillmentTypes` field to the schema.

### Versions
2.2.0

willopez:
Resolves #5513 
Impact: minor
Type: patch

## Issue
This is a patch release to resolve checkout being broken due to not being able to set a shipping method.

### DIFF
--- a/imports/plugins/core/versions/server/migrations/72_add_fulfillmentType_to_shipping_methods.js
+++ b/imports/plugins/core/versions/server/migrations/72_add_fulfillmentType_to_shipping_methods.js
@@ -37,8 +37,10 @@ Migrations.add({
     await Cart.updateMany({}, {
       $unset: {
         "shipping.$[].shipmentMethod.fulfillmentTypes": "",
-        "shipping.$[].shipmentQuotes.$[].method.fulfillmentTypes": ""
+        "shipping.$[groupWithShipmentQuotes].shipmentQuotes.$[].method.fulfillmentTypes": ""
       }
+    }, {
+      arrayFilters: [{ "groupWithShipmentQuotes.shipmentQuotes": { $exists: true } }]
     });
   }
 });
--- a/imports/plugins/core/versions/server/migrations/72_add_fulfillmentType_to_shipping_methods.js
+++ b/imports/plugins/core/versions/server/migrations/72_add_fulfillmentType_to_shipping_methods.js
@@ -1,0 +1,42 @@
+import { Migrations } from "meteor/percolate:migrations";
+import rawCollections from "/imports/collections/rawCollections";
+
+Migrations.add({
+  version: 72,
+  async up() {
+    const { Cart, Shipping } = rawCollections;
+
+    await Shipping.updateMany({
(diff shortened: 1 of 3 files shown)


imports/plugins/core/versions/server/migrations/72_add_fulfillmentType_to_shipping_methods.js
```diff
@@ -0,0 +1,44 @@
...
+        "methods.$[].fulfillmentTypes": ""
+      }
+    });
+
+    await Cart.updateMany({}, {
+      $unset: {
+        "shipping.$[].shipmentMethod.fulfillmentTypes": "",
+        "shipping.$[].shipmentQuotes.$[].method.fulfillmentTypes": ""
+      }
+    });
```
aldeed: This down update will also need the arrayFilters fix.

```js
    await Cart.updateMany({}, {
      $unset: {
        "shipping.$[].shipmentMethod.fulfillmentTypes": "",
        "shipping.$[groupWithShipmentQuotes].shipmentQuotes.$[].method.fulfillmentTypes": ""
      }
    }, {
      arrayFilters: [{ "groupWithShipmentQuotes.shipmentQuotes": { $exists: true } }]
    });
```

willopez: added, thanks

PR #5520, opened 2019-09-04
refactor: remove inviteShopMember meteor method and rewrite with GraphQL

### LINKED ISSUE #5337: Rewrite inviteShopMember to not call through to Meteor method
- Rewrite `inviteShopMember` GraphQL resolver to not use `context.callMeteorMethod`.

kieckhafer:
Resolves #5337   
Impact: **minor**  
Type: **feature|refactor**

## Issue
`inviteShopMember ` still uses a meteor method to perform its actions.

## Solution
Rewrite `inviteShopMember ` mutation to use GraphQL instead of meteor method.
- Remove `inviteShopMember ` meteor method from app

## Breaking changes
Nothing in core code. Any custom code calling the `accounts/inviteShopMember ` Meteor method will break.

## Testing
1. In Reaction Admin, invite a new user (non-registered) to a group
1. See that the invite worked
1. Invite a user with an existing account, but unverified email address to a group
1. See that the invite worked
1. Invite a user with an existing account with a verified email address to a group
1. See that the invite worked


imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,159 @@
...
+  const group = await Groups.findOne({ _id: groupId });
+  if (!group) throw new ReactionError("not-found", "No gr found");
+
+  // we don't allow direct invitation of "owners", throw an error if that is the group
+  const ownerGroup = await Groups.findOne({ name: "owner" });
+  if (group._id === ownerGroup._id) {
+    throw new ReactionError("bad-request", "Cannot directly invite owner");
+  }
+
+  // An account can add another account to a group as long as the person adding
```
kieckhafer: @aldeed I modeled this after some code from `addAccountToGroup`, it seemed a little strange using `user` and `Account`, so hopefully got this correct.

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,159 @@
...
+    dataForEmail = getDataForEmail({ shop: primaryShop, currentUserName, name, url });
+
+    // Get email template and subject
+    templateName = "accounts/inviteShopMember";
+  } else { // TODO: this whole else section uses meteor
+    // There could be an existing user with an invite still pending (not activated).
+    // We create a new account only if there's no pending invite.
+    if (!invitedUser) {
+      // The user does not already exist, we need to create a new account
+      userId = MeteorAccounts.createUser({ // TODO: something with this since it's meteor
```
kieckhafer: @aldeed How do we go about adding MeteorAccounts / users outside of `Meteor`?

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,159 @@
...
+
+  // send invitation email from primary shop email
+  await context.mutations.sendEmail(context, {
+    data: dataForEmail,
+    fromShop: primaryShop,
+    templateName,
+    to: email
+  });
+
+  return Accounts.findOne({ userId });
```
kieckhafer: @aldeed the old version of this meteor method did not have any `appEvent` at the end, should there be something here?

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,165 @@
...
+  } else {
+    // TODO: this entire section uses meteor
+    // TODO: this entire section uses meteor
+    // TODO: this entire section uses meteor
+
+    // There could be an existing user with an invite still pending (not activated).
+    // We create a new account only if there's no pending invite.
+    if (!invitedUser) {
+      // The user does not already exist, we need to create a new account
+      userId = MeteorAccounts.createUser({
```
kieckhafer: @aldeed how do we go about creating `users` and `MeteorAccounts` without using `meteor`?

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
aldeed: The `MeteorAccounts.createUser` call causes `Accounts.onCreateUser` to run, and that emits "afterAccountCreate", so I think we're okay as is.

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
aldeed: For now, yes. `const userId = await context.createUser(/* same options */);`

This is one of the big sticky questions. I added `createUser` on context at some point in hopes of that being a solution, but I'm not sure. In the long run, we should be calling out to the separate identity service, asking it to create a new user. But we aren't quite there yet and currently IDP lives in this same codebase. `createUser` just calls through to Meteor's Accounts.createUser, so it's not so much as solution as just centralizing that Meteor code. Eventually we'll have to remove that Meteor dependency, too, but for this PR, `context.createUser` should work.

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,159 @@
...
+  let templateName;
+
+  if (invitedUser) {
+    userId = invitedUser._id;
+  }
+
+  // if the user already has an account, send in informative email instead of an invite email
+  if (invitedUser && isEmailVerified) {
+    // make sure user has an `Account` - this should always be the case
+    const invitedAccount = await Accounts.findOne({ _id: userId });
```
aldeed: Should be `Accounts.findOne({ userId });`. `account.userId` and `account._id` usually match but nothing says they have to if accounts were imported or something.

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,159 @@
...
+  if (!shop) throw new ReactionError("not-found", "No shop found");
+
+  const group = await Groups.findOne({ _id: groupId });
+  if (!group) throw new ReactionError("not-found", "No gr found");
+
+  // we don't allow direct invitation of "owners", throw an error if that is the group
+  const ownerGroup = await Groups.findOne({ name: "owner" });
+  if (group._id === ownerGroup._id) {
+    throw new ReactionError("bad-request", "Cannot directly invite owner");
+  }
```
aldeed: I believe you could avoid the second `findOne` here by changing this to `if (group.name === "owner")`

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    const token = Random.id();
+
+    // set token to be used for first login for the new account
+    const tokenUpdate = {
+      "services.password.reset": { token, email, when: new Date() },
+      name
+    };
+    Meteor.users.update(userId, { $set: tokenUpdate }); // TODO: something with this since it's meteor
```
aldeed: Change to `await context.collections.users.updateOne({ _id: userId }, { $set: tokenUpdate });`

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
aldeed: I'm not sure if this will work, but you could think of this as two steps:
1. Create user if doesn't already exist
2. Set user's account's group to `groupId`

So this would mean pulling the `context.mutations.addAccountToGroup` call out of the `if` and move it to after the `if`. Always do it.

Then you can maybe remove all of this group fetching and permission checking stuff as long as you remove `isInternalCall: true` from the `addAccountToGroup` call, because then all of that permission checking logic is already there?

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,159 @@
...
+  const { appEvents, collections, user: userFromContext, userHasPermission, userId: userIdFromContext } = context;
+  const { Accounts, Groups, Shops, users } = collections;
+  const {
+    email,
+    groupId,
+    name,
+    shopId
+  } = input;
+
+  if (!userHasPermission(["reaction-accounts"], shopId)) {
```
aldeed: Maybe you should make a more specific `account/invite` for this while you're in here?

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
kieckhafer: I was contemplating doing it that way, since the check was already done inside of `addAccountToGroup`, I just wasn't sure if we should remove the `isInternalCall` since technically it would be an internal call from this other function. Since you're suggesting it too, seems like it is probably the better way and I'll switch it.

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
kieckhafer: Yeah... I originally had that, but then when I added the permissions check on line 71, I was doing the extra look up anyway, so decided to move it up a bit and change this check.

But now that that second permission check will be removed, I'll revert it back to what you have commented here.

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
kieckhafer: This is confusing with all the users / accounts here. I think I do want `_id: userId` here, because at this point, `userId === user._id`, as set here: https://github.com/reactioncommerce/reaction/blob/06d7e7d7d1837e66615af710b14890c18226b501/imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js#L85

Don't `user._id` and `Accounts._id` always match, even if `Accounts._id` and `Accounts.userId` don't?

kieckhafer:
@aldeed addressed all comments, but still have questions about one, I left it "unresolved".

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,140 @@
...
+  if (invitedUser && isEmailVerified) {
+    // make sure user has an `Account` - this should always be the case
+    const invitedAccount = await Accounts.findOne({ userId });
+    if (!invitedAccount) throw new ReactionError("not-found", "User found but matching account not found");
+
+    // do not send token, as no password reset is needed
+    const url = context.getAbsoluteUrl();
+
+    // use primaryShop's data (name, address etc) in email copy sent to new shop manager
+    dataForEmail = getDataForEmail(context, { shop: primaryShop, currentUserName: invitedByName, name, url });
```
aldeed: `getDataForEmail` is async now so needs to be awaited.

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,140 @@
...
+    // set token to be used for first login for the new account
+    const tokenUpdate = {
+      "services.password.reset": { token, email, when: new Date() },
+      name
+    };
+
+    await users.updateOne({ _id: userId }, { $set: tokenUpdate }); // TODO: not sure if this is working
+
+    // use primaryShop's data (name, address etc) in email copy sent to new shop manager
+    dataForEmail = getDataForEmail(context, { shop: primaryShop, currentUserName: invitedByName, name, token });
```
aldeed: `getDataForEmail` is async now so needs to be awaited.

imports/plugins/core/accounts/server/methods/inviteShopOwner.js
```diff
@@ -57,7 +57,8 @@ export default function inviteShopOwner(options, shopData) {
   const currentUser = Meteor.user();
   const currentUserName = getCurrentUserName(currentUser);
   // uses primaryShop's data (name, address etc) in email copy sent to new merchant
-  const dataForEmail = getDataForEmail({ shop: primaryShop, currentUserName, name, token });
+  const context = Promise.await(getGraphQLContextInMeteorMethod(Reaction.getUserId()));
+  const dataForEmail = getDataForEmail(context, { shop: primaryShop, currentUserName, name, token });
```
aldeed: Wrap the `getDataForEmail` call in `Promise.await()` since that function is async now.

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,140 @@
+import _ from "lodash";
+import SimpleSchema from "simpl-schema";
+import Random from "@reactioncommerce/random";
+import ReactionError from "@reactioncommerce/reaction-error";
+import getCurrentUserName from "/imports/plugins/core/accounts/server/no-meteor/util/getCurrentUserName";
+import getDataForEmail from "/imports/plugins/core/accounts/server/util/getDataForEmail";
```
aldeed: Since it no longer has Meteor deps, you can move `getDataForEmail.js` from `server/util` to `server/no-meteor/util`.

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,140 @@
...
+
+  // send invitation email from primary shop email
+  await context.mutations.sendEmail(context, {
+    data: dataForEmail,
+    fromShop: primaryShop,
+    templateName,
+    to: email
+  });
+
+  return Accounts.findOne({ _id: userId });
```
aldeed: Do `Accounts.findOne({ userId });` same as above rather than relying on `_id` being the same as `userId`.

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,140 @@
...
+  let templateName;
+
+  if (invitedUser) {
+    userId = invitedUser._id;
+  }
+
+  // if the user already has an account, send in informative email instead of an invite email
+  if (invitedUser && isEmailVerified) {
+    // make sure user has an `Account` - this should always be the case
+    const invitedAccount = await Accounts.findOne({ userId });
```
aldeed: Since you're just checking for existence and don't need any fields back, you could improve performance by adding `{ projection: { _id: 1 } }` options argument.

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
```diff
@@ -0,0 +1,140 @@
...
+  if (primaryShop._id !== shopId) {
+    shop = await Shops.findOne({ _id: shopId });
+  }
+  if (!shop) throw new ReactionError("not-found", "No shop found");
+
+  const group = await Groups.findOne({ _id: groupId });
+  if (!group) throw new ReactionError("not-found", "No gr found");
+
+  // we don't allow direct invitation of "owners", throw an error if that is the group
+  if (group.name === "owner") {
```
aldeed: Actually I guess we usually check `group.slug` rather than `group.name`. Maybe name can be changed?

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
kieckhafer: Not sure if you saw my comment on that comment above. I think `userId` just might be a badly named `var`, because `userId` actually is just `_id` field.

https://github.com/reactioncommerce/reaction/pull/5520#discussion_r321074146

If you still think it should be changed, I will, but maybe just the `var` name should be different instead?

kieckhafer:
@aldeed comments addressed

imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js
kieckhafer: Nevermind on this, I see how I was misunderstanding what you were saying.

PR #5523, opened 2019-09-05
refactor: use moments calendar format to display time alongside date

kieckhafer:
Impact: **minor**  
Type: **style|refactor**

## Issue
The new Orders UI only shows the date of an order, but it's been requested that the time is added as it's essential when fulfilling orders.

## Solution
Use `moment`s `calendar()` formatting to add relative timestamp, instead of a date. See the docs here for formatting: https://momentjs.com/docs/#/displaying/calendar-time/

## Breaking changes
None, this is a visual change only

## Testing
1. Create an order
1. See that the timestamp is now present alongside the (now relative) order date
1. In the database, change the date of the order a few times to see past date formats (yesterday, last week, last month)

Order_Details_for_order_reference__jFXML4AW68KHDwT4E

Order_Details_for_order_reference__jFXML4AW68KHDwT4E

Order_Details_for_order_reference__jFXML4AW68KHDwT4E

### DIFF
--- a/imports/plugins/core/orders/client/components/OrderHeader.js
+++ b/imports/plugins/core/orders/client/components/OrderHeader.js
@@ -27,7 +27,7 @@ function OrderHeader(props) {
   const { classes, moment, order } = props;
   const { createdAt, displayStatus, referenceId, status } = order;
   const orderDate = (moment && moment(createdAt).calendar(null, {
-    sameElse: "DD/MM/YYYY h:mm A"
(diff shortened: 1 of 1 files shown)


samkelleher:
Just a FYI: I got some feedback today, and this was that the _absolute_ version is always preferred on this details view, from the perspective of customer service reps who always look at these screens.

machikoyasuda:
Confirmed:
Screen Shot 2019-09-10 at 2 49 15 PM

PR #5524, opened 2019-09-05
refactor: remove context.callMeteorMethod

kieckhafer:
Impact: **major**  
Type: **refactor**

## Issue
`context.callMeteorMethod` is no longer used in Reaction, and should therefore be removed

## Solution
Remove `callMeteorMethod` from the context.

## Breaking changes
None in core code, as all instances of use have already been removed. Any third-party code that uses `context.callMeteorMethod` will need to be refactored.

## Testing
1. See that the app works even though `callMeteorMethod` has been removed.


kieckhafer:
@aldeed There is one instance of `meteorContext.callMeteorMethod`, should this be removed too?

https://github.com/reactioncommerce/reaction/blob/refactor-kieckhafer-remove-callMeteorMethod/imports/plugins/core/graphql/server/getGraphQLContextInMeteorMethod.js#L53-L56

kieckhafer:
@aldeed updated, ready for another look

Issue #5529, opened 2019-09-09
Rewrite any code which calls getGraphQLContextInMeteorMethod to not use Meteor methods


kieckhafer:
- Rewrite any code which calls `getGraphQLContextInMeteorMethod` to get context directly from being a passed variable.
- This will most likely consist of rewriting all places that use `getGraphQLContextInMeteorMethod` to no longer rely on meteor.

### Where is it used

- [x] `inviteShopMember.js`
- [x] `inviteShopOwner.js`
- [x] `sendResetPasswordEmail.js`
- [x] `sendVerificationEmail.js`
- [x] `sendWelcomeEmail.js`
- [x] `catalog.js`
- [x] `publishProducts.js` - PR #5528 
- [x] `cart.js`
- [x] `getLocale.js`
- [x] `index.js`
- [x] `verifySettings.js`
- [x] `generate-sitemaps.js`

aldeed:
Done

PR #5530, opened 2019-09-10
refactor: remove getGraphQLContextInMeteorMethod from sitemaps-generator plugin

kieckhafer:
Resolves parts of #5529 
Impact: **minor**  
Type: **refactor**

## Issue
`generate-sitemaps` still use `getGraphQLContextInMeteorMethod`

## Solution
Move the `generate-sitemaps-job` into a startup function, and pass `context` from the startup function instead of using `getGraphQLContextInMeteorMethod()`

## Breaking changes
None

## Testing
1. Start the app
1. see that the `generate-sitemaps-job` ran without error on startup


kieckhafer:
Closing this, as @aldeed has made better changes in another branch that will cause these to be moot.

PR #5531, opened 2019-09-10
refactor: create verifySMTPEmailSettings GraphQL mutation, replace meteor version

kieckhafer:
Resolves parts of #5529
Impact: minor
Type: feat|refactor

## Issue
`email/verifySettings` was a meteor method which used `getGraphQLContextInMeteorMethod`

## Solution
Rewrite the meteor method to be a `verifySMTPEmailSettings` GraphQL mutation, which in turn will remove usage of `getGraphQLContextInMeteorMethod`.

The color of the status indicator is not reactive. This is not a regression, it is currently not reactive in `master`. Since there are other Meteor methods and more refactors of this code to come in the future, I've chosen to leave this as is for now.

## Breaking changes
Any custom code which used the `email/verifySettings` meteor method will need to be updated to use the GraphQL mutation

## Testing
1. Add correct email settings and see they come back as verified
1. Add incorrect email settings and see they come back as an error


kieckhafer:
@aldeed I'm not entirely sure why [graphql-lint is failing](https://circleci.com/gh/reactioncommerce/reaction/55798?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link). The message says `Every argument must have a description  arguments-have-descriptions`, but I think they all do, unless I'm not seeing something: https://github.com/reactioncommerce/reaction/pull/5531/files#diff-abfa8cc3a4ed4b0025df75047b823930

imports/plugins/included/email-smtp/server/no-meteor/mutations/verifySMTPEmailSettings.js
```diff
@@ -0,0 +1,80 @@
...
+
+  const logConfig = { ...config };
+  delete logConfig.auth;
+  Logger.debug(logConfig, "Verifying email config settings");
+
+  const transporter = nodemailer.createTransport(config);
+
+  const verified = await transporter.verify()
+    .then((result) => result)
+    .catch((error) => { throw new ReactionError(error.responseCode, error.response); });
```
aldeed: If using async/await, it's best to stick with that and try/catch instead of also using `.then` and `.catch`.

```js
let verified;
try {
  verified = await transporter.verify();
} catch (error) {
  throw new ReactionError(error.responseCode, error.response);
}

return verified;
```

imports/plugins/included/email-smtp/server/no-meteor/schemas/schema.graphql
```diff
@@ -0,0 +1,37 @@
...
+  user: String!
+}
+
+"Response payload for the verifySMTPEmailSettings mutation"
+type VerifySMTPEmailSettingsInputPayload {
+  "The same string you sent with the mutation params, for matching mutation calls with their responses"
+  clientMutationId: String
+
+  "Are the settings OK"
+  verified: Boolean
```
aldeed: Maybe do `isVerified` to follow the rule about boolean naming. It should be safe to do `Boolean!`, too. I don't think it will ever be `null` when there were no errors.

imports/plugins/included/email-smtp/server/no-meteor/schemas/schema.graphql
```diff
@@ -0,0 +1,37 @@
...
+  "The username of the SMTP email service user"
+  user: String!
+}
+
+"Response payload for the verifySMTPEmailSettings mutation"
+type VerifySMTPEmailSettingsInputPayload {
+  "The same string you sent with the mutation params, for matching mutation calls with their responses"
+  clientMutationId: String
+
+  "Are the settings OK"
```
aldeed: More descriptive description would be "True if the SMTP connection was made and authentication was successful."

imports/plugins/included/email-smtp/client/actions/settings.js
```diff
@@ -45,34 +57,42 @@ export default {
...
+      // eslint-disable-next-line promise/no-nesting
+      .catch(() => Alert({
+        title: i18next.t("mail.alerts.connectionFailed"),
+        text: i18next.t("mail.alerts.saveAnyway"),
+        type: "warning",
+        showCancelButton: true,
+        cancelButtonText: i18next.t("app.cancel"),
+        confirmButtonColor: "#DD6B55",
+        confirmButtonText: i18next.t("app.save")
+      }).then(() => save()).catch(() => true));
```
aldeed: If using async/await, it's best to stick with that and try/catch instead of also using `.then` and `.catch`.

```js
let shouldSave = true;
try {
    await client.mutate({
      mutation: verifySMTPEmailSettings,
      variables: {
        input: {
          host: settings.host,
          password: settings.password,
          port: settings.port,
          service: settings.service,
          shopId: opaqueShopId,
          user: settings.user
        }
      }
    });
} catch (error) {
  shouldSave = false;
}

if (!shouldSave) {
  try {
      await Alert({
        title: i18next.t("mail.alerts.connectionFailed"),
        text: i18next.t("mail.alerts.saveAnyway"),
        type: "warning",
        showCancelButton: true,
        cancelButtonText: i18next.t("app.cancel"),
        confirmButtonColor: "#DD6B55",
        confirmButtonText: i18next.t("app.save")
      })
  } catch (error) {
    callback();
    return;
  }
}

callback();
await save();
```

It's still a bit weird. If it doesn't work and what you have does work, then it's fine. We should eventually make a confirm Alert that returns true/false rather than throwing so we don't need so many try/catch.

kieckhafer: I was struggling with getting `try/catch` to work, and it just wasn't, which is why i'm doing the different `then...catch` in both places here. I'll try again with what you've posted here, but it was just being weird.

imports/plugins/included/email-smtp/server/no-meteor/mutations/verifySMTPEmailSettings.js
```diff
@@ -0,0 +1,89 @@
...
+  delete logConfig.auth;
+  Logger.debug(logConfig, "Verifying email config settings");
+
+  const transporter = nodemailer.createTransport(config);
+
+  // const verified = await transporter.verify()
+  //   .then((result) => result)
+  //   .catch((error) => { throw new ReactionError(error.responseCode, error.response); });
+
+  // return verified;
```
aldeed: Delete these?

kieckhafer: 👍 Was also going to try to add a `nock` test but hadn't' gotten to it yet, all other comments were addressed.

aldeed:
OK to merge, but holding off until integration test is added.

PR #5528, opened 2019-09-09
refactor: remove `catalog/publish/products` meteor method, use `publishProductsToCatalog` GQL Mutation instead

kieckhafer:
Resolves parts of #5529   
Impact: **minor**  
Type: **refactor**

## Issue
`publishContainer` and `productGridContainer` still use the `catalog/publish/products` meteor method to publish products, which in turn uses `getGraphQLContextInMeteorMethod`.

## Solution
Re-write anywhere that calls `catalog/publish/products` to use the `publishProuctsToCatalog` mutation instead, remove the `catalog/publish/products` meteor method, which will remove `getGraphQLContextInMeteorMethod` in this instance.

## Breaking changes
Any custom code that uses the `catalog/publish/products` Meteor method will need to be re-written to use the `publishProductsToCatalog` graphql mutation

## Testing
1. See the `catalog/publish/products` meteor method has been removed
1. Attempt to publish changes to a product to the catalog.
1. See you are successful

PR #5541, opened 2019-09-11
refactor: remove `catalog/publish/products` meteor method, use `publishProductsToCatalog` GQL Mutation instead

kieckhafer:
This is a replacement of #5528, which I messed up some rebasing while trying to get DCO to pass.

Resolves parts of #5529   
Impact: **minor**  
Type: **refactor**

## Issue
`publishContainer` and `productGridContainer` still use the `catalog/publish/products` meteor method to publish products, which in turn uses `getGraphQLContextInMeteorMethod`.

## Solution
Re-write anywhere that calls `catalog/publish/products` to use the `publishProuctsToCatalog` mutation instead, remove the `catalog/publish/products` meteor method, which will remove `getGraphQLContextInMeteorMethod` in this instance.

## Breaking changes
Any custom code that uses the `catalog/publish/products` Meteor method will need to be re-written to use the `publishProductsToCatalog` graphql mutation

## Testing
1. See the `catalog/publish/products` meteor method has been removed
1. Attempt to publish changes to a product to the catalog.
1. See you are successful


kieckhafer:
@mikemurray @aldeed I messed something up big time while trying to get DCO to pass on #5528, so I'm closing that PR in favor of this one. It should be the same as the last one, with your comments all addressed.

imports/plugins/core/catalog/client/components/publishControls.js
```diff
@@ -42,15 +65,28 @@ class PublishControls extends Component {
...
-          {i18next.t("productDetailEdit.publish")}
-        </Button>
+        <Mutation mutation={publishProductsToCatalog}>
+          {(mutationFunc, { data, error }) => {
+            if (error) {
+              Alerts.toast(error.message, "error");
+            }
+            if (data) {
+              Alerts.toast(i18next.t("admin.catalogProductPublishSuccess"), "success");
+            }
```
aldeed: It doesn't work to do the alerts this way because the component rerenders potentially unlimited times and every rerender pops up a new toast alert. You can see this when testing. After you change something and then click publish, every time you change another field or do anything, an additional "successfully published" alert appears.

You could move the alerts into the other functions, after the `await mutation` lines. Otherwise, `Mutation` takes `onCompleted` and `onError` callbacks, but I think I've had trouble getting those to work consistently in the past. They may have fixed since then. See https://www.apollographql.com/docs/react/v2.4/essentials/mutations/#props

You have similar code with a similar issue in the grid file, but in that case when I tested I actually never saw any toasts, even though both selected products were successfully published and their table cells updated to say "Published".

aldeed:
I'm not sure what's up with the integration tests failing on CI. Hopefully just a fluke.

imports/plugins/core/catalog/client/components/publishControls.js
kieckhafer: Updated using callbacks, they seem to be working fine.

kieckhafer:
@aldeed updated Mutation to use callbacks.

Tests seems to be passing again.

PR #5543, opened 2019-09-11
Make "Create product" button visible by default

willopez:
Impact: minor
Type:  fix

## Issue
The "Create product" button on the products table was not visible by default.

## Resolution
Make button visible

### DIFF
--- a/imports/plugins/included/product-admin/client/components/ProductTable.js
+++ b/imports/plugins/included/product-admin/client/components/ProductTable.js
@@ -148,6 +148,17 @@ function ProductTable({ onCreateProduct }) {
         setVisibility={setTagSelectorVisible}
         selectedProductIds={selectedProductIds}
       />
+      {(!isFiltered && !isTagSelectorVisible && !isFilterByFileVisible) && (
+        <Grid item sm={12} >
+          <Button
+            color="primary"
+            onClick={onCreateProduct}
+            variant="contained"
+          >
+            {i18next.t("admin.createProduct") || "Create product"}
+          </Button>
+        </Grid>
+      )}
       {isFiltered && (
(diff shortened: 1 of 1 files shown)


imports/plugins/included/product-admin/client/components/ProductTable.js
```diff
@@ -158,15 +169,6 @@ function ProductTable({ onCreateProduct }) {
...
-          {(!isFiltered && !isTagSelectorVisible && !isFilterByFileVisible) && (
-            <Button
-              color="primary"
-              onClick={onCreateProduct}
-              variant="contained"
-            >
-              {i18next.t("admin.createProduct") || "Create product"}
-            </Button>
-          )}
           {isFiltered && (
```
kieckhafer: This looks like an `isFiltered` check is nested inside of another `isFiltered` check (line 162), that doesn't' seem necessary.

willopez: That was the case before, which caused the issue. In this PR I moved the block that has the "Create product" block outside.

kieckhafer: I think that's separate from what I'm seeing here.

https://github.com/reactioncommerce/reaction/blob/a6d34916f13a7d06162a07fa46872da7a2d2b584/imports/plugins/included/product-admin/client/components/ProductTable.js#L162-L182

Line `162``is checking for `isFiltered`, and that surrounds the code up until line `182`. Then inside of that, there is another `isFiltered` check that goes from line `172` until line `180`. I think this check can be removed.

willopez: issues resolved.

PR #5549, opened 2019-09-12
Faster Jest integration tests

aldeed:
Impact: **minor**  
Type: **test**

## Issue
Jest integration tests are slow

## Solution
- Switch to `jest-mongodb` package. It does the same thing we were already doing, setting up `mongodb-memory-server`, but it does it in a way that can happen one time total rather than one time per test file.
- Run integration tests "in band" and with Jest cache, so that they can take advantage of the `jest-mongodb` package.

Running locally, integration tests are 8x faster for me.

## Breaking changes
None

## Testing
- `npm run test:integration` and confirm it's faster
- Confirm the integration test task on CircleCI is as fast or faster than before.


aldeed:
@bt3gl Did you `npm install` first to install the new `@shelf/jest-mongodb` package?

bt3gl:
@aldeed Getting another error, a memory one
Screen Shot 2019-09-16 at 3 44 13 PM

bt3gl:
After running on containers, using:
```docker-compose run --rm reaction npm install @shelf/jest-mongodb``` and then
```docker-compose run --rm reaction npm run test:integration```, all the tests passed. T
heir speed is absolutely improved!

🛳-it!

PR #5556, opened 2019-09-15
feat: add support for multilingual email

### LINKED ISSUE #5555: Feature: support multilingual emails based on account and order preferred languages
Emails should be based on account profile language or order language before shop language.
This way shop supports multilingual email sending.
Currently a shop will send emails only in the language that is set in `shop.language` field in the database. This means all emails from same shop will be in same language.

## Prerequisites
* [x ] Are you running the latest version?

jm18457:
Resolves #5555 
Impact: **major**  
Type: **feature**

## Issue
[Previous pull request](https://github.com/reactioncommerce/reaction/pull/5171), which was closed due to significant changes in reaction in this timeframe and different forked repository.

Multilingual emails by shop are not supported, because email language will always be that of `shop.language`. All users will receive email in same language.

## Solution
Added fields `profile.language` to Accounts collection and `ordererPreferredLanguage` to order collection.
Updated graphql queries and mutations to allow for optional language.
This way storefront can supply language and we support multilingual emails.

## Breaking changes
No breaking changes.

## Testing
Tests are not implemented, but code is tested locally.

### Account Profile Language
1. Run graphql mutation `setAccountProfileLanguage`.

### Order language
1. Create order either with or without `orderPreferredLanguage` field.

### Account Emails
1. Set profile language for user.
2. Do an activity which sends email with that user such as reset password, create order etc.

### DIFF
--- a/imports/plugins/core/accounts/server/no-meteor/schemas/account.graphql
+++ b/imports/plugins/core/accounts/server/no-meteor/schemas/account.graphql
@@ -310,7 +310,7 @@ type SetAccountProfileCurrencyPayload {
 "The response from the `setAccountProfileLanguage` mutation"
 type SetAccountProfileLanguagePayload {
   "The updated account"
-  account: Account
+  account: Account!
 
   "The same string you sent with the mutation params, for matching mutation calls with their responses"
   clientMutationId: String
@@ -383,7 +383,7 @@ extend type Mutation {
   setAccountProfileLanguage(
     "Mutation input"
     input: SetAccountProfileLanguageInput!
-  ): SetAccountProfileLanguagePayload
+  ): SetAccountProfileLanguagePayload!
 
   "Remove an address that exists in the `addressBook` field for an account"
   updateAccountAddressBookEntry(
--- a/imports/plugins/core/accounts/server/no-meteor/schemas/account.graphql
+++ b/imports/plugins/core/accounts/server/no-meteor/schemas/account.graphql
@@ -76,6 +76,18 @@ input SetAccountProfileCurrencyInput {
   currencyCode: String!
 }
 
+"Describes an account profile language change"
+input SetAccountProfileLanguageInput {
+ "The account ID, which defaults to the viewer account"
+ accountId: ID
+
+ "An optional string identifying the mutation call, which will be returned in the response payload"
+ clientMutationId: String
+
+ "The i18n language code value"
+ language: String!
+}
+
 "Defines a new Email and the account to which it should be added"
 input AddAccountEmailRecordInput {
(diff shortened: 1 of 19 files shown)


imports/plugins/core/accounts/server/no-meteor/schemas/account.graphql
```diff
@@ -274,6 +289,15 @@ type SetAccountProfileCurrencyPayload {
   clientMutationId: String
 }
 
+"The response from the `setAccountProfileLanguage` mutation"
+type SetAccountProfileLanguagePayload {
+  "The updated account"
+  account: Account
```
aldeed: Do `Account!` so the client knows it will never be `null`.

imports/plugins/core/accounts/server/no-meteor/schemas/account.graphql
```diff
@@ -331,6 +355,12 @@ extend type Mutation {
     input: SetAccountProfileCurrencyInput!
   ): SetAccountProfileCurrencyPayload
 
+  "Set the preferred language for an account"
+  setAccountProfileLanguage(
+    "Mutation input"
+    input: SetAccountProfileLanguageInput!
+  ): SetAccountProfileLanguagePayload
```
aldeed: Do `SetAccountProfileLanguagePayload!` so the client knows it will never be `null`.

imports/plugins/core/accounts/server/util/sendVerificationEmail.js
```diff
@@ -25,6 +25,7 @@ export default async function sendVerificationEmail({
 }) {
   // Make sure the user exists, and email is one of their addresses.
   const user = Meteor.users.findOne({ _id: userId });
+  const account = Meteor.users.findOne({ _id: userId });
```
aldeed: This should be `const account = Accounts.findOne({ userId });`

imports/plugins/core/orders/server/no-meteor/util/sendOrderEmail.js
```diff
@@ -27,13 +27,34 @@ export default async function sendOrderEmail(context, order, action) {
...
+
+/**
+ * @summary Returns language to be used for order emails.
+ *          If cart is account based and has set language
+ *          then returns that language, else order language.
+ * @param {Object} context App context
+ * @param {Object} order - The order document
+ * @returns {String} i18n language code
+ */
+async function getLanguageForOrder(context, { language, accountId }) {
```
aldeed: `language` here should be `ordererPreferredLanguage`, right?

jm18457:
I have misread your comment and rebased, instead of merging developing.
I have made changes as requested.

PR #5562, opened 2019-09-16
feat: add product and productVariant related GraphQL mutations

kieckhafer:
Resolves parts of #5529
Impact: **minor**  
Type: **feature|refactor**

## Issue
Our `product` (not `catalog`) admin server code is still fully `meteor`, including a few mutations which use `getGraphQLContextInMeteorMethod`, which we are deprecating.

## Solution
1. Add four new GraphQL mutations for product admin:
- `createProduct`
```
mutation {
  createProduct(input: { shopId: "{shopId}" }){
    product {
      _id
    }
  }
}
```

- `createProductVariant`
```
mutation {
  createProductVariant(input: { productId: "{parentProductId}", shopId: "shopId" }){
    variant {
      _id
    }
  }
}
```

- `cloneProducts`
```
mutation {
  cloneProducts(input: { productIds: ["{id1}"," {id2}"], shopId: "shopId" }){
    products {
      _id
    }
  }
}
```

- `cloneProductVariants`
```
mutation {
  cloneProductVariants(input: { variantIds: ["{id1}"," {id2}"], shopId: "shopId" }){
    variants {
      _id
    }
  }
}
```

2. Remove existing `products/createProduct`, `products/createVariant`, `products/cloneVariant` code, and all related code which is no longer needed / moved in the new mutations

3. Update client code to call new GraphQL mutations instead of meteor methods.
- This was a much bigger task than was expected. A lot of our client components use `UNSAFE_...` and other React lifecycle methods, and will need to be refactored completely in the near future, so I did my best to keep them as close to how they are as possible as to not need a complete refactor at the moment.

One regression / change here is that when you create a new product / variant, you are no longer redirect to that new variant page. The reason is when we use a GraphQL mutation, the new product / variant is created almost immediately, however, it takes 1+ seconds for the Meteor subscriptions to update and include the new product. Because of this, when redirecting, you'd land on the page of a variant that isn't yet in the Meteor subscription, and there would be an error. Because of the need to refactor the components soon anyway, I think this is a decent trade off, as opposed to delaying this PR and refactoring all of these components now. I've left a comment in the code suggesting we re-add this feature when these components are updated.

## Breaking changes
Any custom plugins which use any of the above listed meteor methods will need to be updated to use the GraphQL mutation instead

## Testing
1. Create a product using the mutation, and also in the UI (Create Product button above Product table)
1. Create a variant using the mutation, and also in the UI (+ button on product page)
1. Create an option using the mutation, and also in the UI (+ button on product variant page in options area)
1. Clone a product using the mutation, and also in the UI (check one or more products on the product table, and use the dropdown to clone, and also inside an individual product page, use the `...` menu)
1. Clone a variant using the mutation, and also in the UI (inside an individual product variant, use the `...` menu))
1. See all of the above work

You can see that all the new resolvers, `shop`, `tags`, 'slug'  and `variants` on type `Product`, and `options` and `shop` on `ProductVariant`, work with this mutation:
```
mutation {
  cloneProducts(input: { productIds: ["{productId}"], shopId: "{shopId}" }){
    products {
      _id,
     slug
      shop {
				_id
        name
      }
      tagIds,
      tags {
        edges {
          node {
            displayTitle
            heroMediaUrl
            position
            slug
          }
        }
      },
      variants {
        _id
        shop {
           _id
          }
        options {
          _id
         shop {
           _id
          }
        }
      }
    }
  }
}
```


imports/plugins/core/product/server/no-meteor/resolvers/Mutation/createProductVariant.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param {Object} args.input - mutation input object
+ * @param {String} args.input.clientMutationId - The mutation id
+ * @param {String[]} args.input.parentId - the product or variant ID which we create new variant on
+ * @param {Object} context - an object containing the per-request state
+ * @return {Promise<Object>} createProductVariant payload
+ */
+export default async function createProductVariant(_, { input }, context) {
+  const {
+    clientMutationId = null,
+    parentId
```
kieckhafer: @aldeed should we pass `shopId` here as well? I know we usually want to pass shopId, however shopId is available on the `parent` product, which is found prior to anywhere that would use it: https://github.com/reactioncommerce/reaction/pull/5562/files#diff-23f5be7d77820413c96a2327c790435bR27

imports/plugins/core/product/server/no-meteor/mutations/cloneProductVariants.js
```diff
@@ -0,0 +1,124 @@
...
+
+  if (!userHasPermission(["createProduct", "product/admin", "product/clone"], shopId)) {
+    throw new ReactionError("access-denied", "Access Denied");
+  }
+
+  // Check to make sure all variants are on the same shop
+  const count = await Products.find({ _id: { $in: variantIds }, type: "variant", shopId }).count();
+  if (count !== variantIds.length) throw new ReactionError("not-found", "One or more variants do not exist");
+
+  const newVariantIds = await Promise.all(variantIds.map(async (variantId) => {
```
kieckhafer: @aldeed I'm doing a lot of awaiting promises and in here, this works, but not sure if it's the best way to do all of this.

imports/plugins/core/product/server/no-meteor/mutations/cloneProductVariants.js
```diff
@@ -0,0 +1,125 @@
...
+          width: `${sortedVariant.width}`,
+          weight: `${sortedVariant.weight}`,
+          length: `${sortedVariant.length}`
+        });
+      }
+      delete clonedVariantObject.updatedAt;
+      delete clonedVariantObject.createdAt;
+
+      // Apply custom transformations from plugins.
+      for (const customFunc of context.getFunctionsOfType("mutateNewVariantBeforeCreate")) {
```
kieckhafer: I copied the `clone` functions as close as I could from the Meteor methods. I noticed that this `getFunctionsOfType` loop was in the `variant` clone method, but not in the `product` clone method, even the `variant` section of that method. Does it need to be added there?

imports/plugins/core/product/server/no-meteor/resolvers/Mutation/createProductVariant.js
aldeed: We've been trying to always ask for `shopId` even if we can get it because in many cases it allows the auth check to happen before any db queries, which is faster, cheaper, and can be handy if we move auth to a separate service.

aldeed: But when checking permission based on the passed in shopId, you always have to remember to include `shopId` in the database find as well. Otherwise they could pass in any shop they have access to while passing in an entity ID that belongs to a different shop.

imports/plugins/core/product/server/no-meteor/mutations/cloneProductVariants.js
aldeed: That might be an oversight. It seems like it should be in both, or possibly in neither since they are cloned from variants that were already mutated. It depends on what those functions do. Possibly there should be separate `mutateNewVariantBeforeCreate` and `mutateNewVariantAfterClone` for the most specificity.

kieckhafer:
@aldeed @mikemurray still writing tests, but since this turned into such a huge PR figured I'd tag you now if you wanted to take a preliminary look at all non test related code.

kieckhafer:
@aldeed @mikemurray integration tests are added, this is ready for an official review.

imports/plugins/core/product/server/no-meteor/schemas/product.graphql
```diff
@@ -0,0 +1,255 @@
...
+
+    "Return only results that come before this cursor. Use this with `last` to specify the number of results to return."
+    before: ConnectionCursor,
+
+    "Return at most this many results. This parameter may be used with either `after` or `offset` parameters."
+    first: ConnectionLimitInt,
+
+    "Return at most this many results. This parameter may be used with the `before` parameter."
+    last: ConnectionLimitInt,
+
```
mikemurray: Should add the `offset` field here to the schema like here: https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/catalog/server/no-meteor/schemas/schema.graphql#L421

kieckhafer: Done.

PR #5567, opened 2019-09-18
fix: add shopId argument into getMailConfig function call

### LINKED ISSUE #5566: getMailConfig in sentSMTPEmail is not passed shop id
## Prerequisites
* [ ] Did you search the issue queue for existing issue? [Search issues](https://github.com/reactioncommerce/reaction/issues)

## Issue Description
In [sendSMTPEmail](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/included/email-smtp/server/no-meteor/util/sendSMTPEmail.js#L15) `shopId` is not being passed as a second variable to function [getMailConfig](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/included/email-smtp/server/no-meteor/util/getMailConfig.js#L113) which requires shopId as parameter.

This causes that the function does not return correct mailConfig, even though mail is properly set.

## Steps to Reproduce
1. Set email provider in email settings in admin dashboard.
2. Resend an email.
3. No email will be resent.
4. If you log for shopId in above mentioned `getMailConfig` you will receive value of undefined.

## Possible Solution
Pass additional parameter `shopId` into `getMailConfig` function.

jm18457:
Resolves #5566  
Impact: **minor**  
Type: **bugfix**

## Issue
`getMailConfig` is called  without second argument `shopId`.

## Solution
Adds `shopId` as second argument to `getMailConfig` call.

## Breaking changes
None

## Testing
TBD

### DIFF
--- a/imports/plugins/included/email-smtp/server/no-meteor/util/sendSMTPEmail.js
+++ b/imports/plugins/included/email-smtp/server/no-meteor/util/sendSMTPEmail.js
@@ -1,5 +1,6 @@
 import Logger from "@reactioncommerce/logger";
 import nodemailer from "@reactioncommerce/nodemailer";
+import Reaction from "/imports/plugins/core/core/server/Reaction";
 import getMailConfig from "./getMailConfig";
 
 /**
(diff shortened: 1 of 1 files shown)


aldeed:
Thanks @jm18457. This is fixed in a different way in my upcoming PR removing the Meteor dependency from background jobs. Should be pushed later today.

PR #5574, opened 2019-09-20
Add extra field to the error message

impactmass:
Resolves: N/A
Impact: **minor**  
Type: **chore**

## Issue
Following comments on https://github.com/reactioncommerce/reaction/issues/5505, the error message from the function will be helpful if it shows the template name. That way a dev can track the missing template.


kieckhafer:
I think this looks good.

I think there could be a reason to check for `templateName` as well, maybe earlier in the function, however if the `templateName` is undefined, the error message will be `No email template undefined found for language ${language}`, so that'll give an indication you need to pass it anyway.

impactmass:
Thanks. I agree with your later point: if it's `undefined`, the error message will still shows it, and then can be fixed.

PR #5577, opened 2019-09-23
refactor: rewrite removeAccountEmail without meteor

kieckhafer:
Resolves parts of #5529
Impact: minor
Type: feature|refactor

## Issue
An app event which sends an email, `afterAddUnverifiedEmailToUser`, is being called inside a `Meteor.startup` file. In our effort to remove `getGraphQLContextInMeteorMethod`, we need to move this outside of meteor and into the `no-meteor/startup` file. 

## Solution
Move the `afterAddUnverifiedEmailToUser` appEvent call into the `no-meteor/startup` file. This also entailed writing a `removeAccountEmailRecord` GraphQL mutation. There was a pre-existing mutation schema, so the new mutation is based off this. This also entailed updating the Reaction Admin client UI to use this new mutation where applicable.

## Breaking changes
Any custom code using the `accounts/removeEmailAddress` meteor method will need to be updated to use the new GraphQL mutation.

## Testing
1. Start the app, login as an admin
1. Use the dropdown at the top right of the screen to view your profile
1. Change your email address on this page. This will call the new mutation as part of the old email address removal.


imports/plugins/core/accounts/server/no-meteor/util/sendVerificationEmail.js
```diff
@@ -55,56 +53,63 @@ export default async function sendVerificationEmail({
   const when = new Date();
   const tokenObj = { token, address, when };
```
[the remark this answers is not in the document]
kieckhafer: Some places we call this take `address` (looks like when we create a new email), and some accept `email` as the second field (looks like when we reset a password). I've set it up to accept either one, is this the best way to do it? Or should we investigate / change it so they accept the same var name?

imports/plugins/core/accounts/server/init.js
```diff
@@ -1,18 +1,13 @@
...
 import * as Collections from "/lib/collections";
 import appEvents from "/imports/node-app/core/util/appEvents";
 import Reaction from "/imports/plugins/core/core/server/Reaction";
 import sendWelcomeEmail from "/imports/plugins/core/accounts/server/util/sendWelcomeEmail";
-import sendVerificationEmail from "./util/sendVerificationEmail";
-
-appEvents.on("afterAddUnverifiedEmailToUser", ({ email, userId }) => {
-  sendVerificationEmail({ email, userId });
-});
+import generateVerificationTokenObject from "/imports/plugins/core/accounts/server/no-meteor/util/generateVerificationTokenObject";
```
kieckhafer: This will be refactored very soon in the PR moving accounts registration out of here.

imports/plugins/core/accounts/server/methods/sendResetPasswordEmail.js
```diff
@@ -47,9 +47,7 @@ async function sendResetEmail(userId, optionalEmail, language) {
   }
 
   // Create token for password reset
-  const token = Random.secret();
-  const when = new Date();
-  const tokenObj = { token, email, when };
+  const tokenObj = generateVerificationTokenObject({ email });
```
kieckhafer: This file has already been refactored in another PR that isn't merged yet. Will update that PR with this new `generateVerificatinTokenObject` once this PR is merged.

kieckhafer:
@aldeed all comment addressed. Left one question but will be a quick resolution either way.

imports/plugins/core/accounts/server/methods/sendResetPasswordEmail.js
aldeed: @kieckhafer There were conflicts when I merged that PR and I must have resolved them incorrectly. Make sure any changes from #5556 are in your non-Meteor version of the function, but otherwise OK to re-delete.

PR #5578, opened 2019-09-24
refactor: demeteorize send welcome email / account creation

kieckhafer:
Resolves parts of #5529
Impact: minor
Type: feature|refactor

## Issue
Account creation, after the Meteor `user` has been created, is currently intertwined in a Meteor.startup function. This calls the meteorized `sendWelcomeEmail`, as well as using meteor for many account creation functions.

## Solution
- Create a `createAcccount` GraphQL mutation to decouple meteor from `Account` creation
- Demeteorize `sendWelcomeEmail` util

## Breaking changes
Any app with a custom account creation flow, which customizes `Accounts.createUser`, may need to update to the new GraphQL mutation.

## Testing
1. Register for an account
1. See you get the welcome email
1. See Account is created in both `Accounts` and `users` collections


kieckhafer:
@impactmass since this is account related, want you to take a look since you're working on the auth related stuff.

imports/plugins/core/accounts/server/init.js
```diff
@@ -163,25 +143,34 @@ Meteor.startup(() => {
...
+    const context = Promise.await(getGraphQLContextInMeteorMethod(null));
+
+    context.isInternalCall = true;
+
+    Promise.await(context.mutations.createAccount(context, {
+      additionals,
+      groupId: groupToAddUser,
+      shopId,
+      tokenObj,
+      user
```
aldeed: Let's take this opportunity to flatten this input. There aren't that many actual fields we need, so having the sub-objects of `additionals` and `user` seems overly complex. I suggest:

```js
{
  shopId,
  name,
  picture,
  bio,
  username,
  userId,
  verificationToken
}
```

And update GraphQL input schema types to match that.

I think that's all the info createAccount needs. Note the following:
- Move the whole user.services block back into here, and use it to set `name`, `picture`, `bio`, and `username` if possible. Then pass just those to createAccount.
- Instead of passing the whole tokenObj, pass just `verificationToken: tokenObj.token`. I don't think it needs any more token info.
- I don't think we should support passing in a `groupId`. It would need security checks to be safe. Instead, can we replace all the groups lookup in both functions with `const group = await Groups.findOne({ slug: "customer", shopId });`? In other words, put all new accounts in the "customer" group, and they can call the other mutation to set a new group after. The only place we pass `groupId` is in `inviteShopMember`, so you can update that function to call `addAccountToGroup` after creating the user and get the same effect.

imports/plugins/core/accounts/server/no-meteor/mutations/createAccount.js
```diff
@@ -0,0 +1,123 @@
...
+    type: Object,
+    blackbox: true
+  }
+});
+
+/**
+ * @name accounts/createAccount
+ * @memberof Mutations/Accounts
+ * @summary Invite new admin users (not consumers) to secure access in the dashboard to permissions
+ * as specified in packages/roles
```
aldeed: Not a correct summary

imports/plugins/core/accounts/server/init.js
kieckhafer: There is also the `profile` object on `createUser`, which has `invited` on it at the moment. Should this be passed to `createAccount` as well?

kieckhafer: > I don't think we should support passing in a groupId. It would need security checks to be safe. Instead, can we replace all the groups lookup in both functions with const group = await Groups.findOne({ slug: "customer", shopId });? In other words, put all new accounts in the "customer" group, and they can call the other mutation to set a new group after. The only place we pass groupId is in inviteShopMember, so you can update that function to call addAccountToGroup after creating the user and get the same effect.

It looks like we might already be doing this: https://github.com/reactioncommerce/reaction/blob/refactor-kieckhafer-demeteorize-SendWelcomeEmail/imports/plugins/core/accounts/server/no-meteor/mutations/inviteShopMember.js#L125

Am I missing something, or can I just remove the group code without adding something back?

kieckhafer:
@aldeed updated based on your comments. A couple notes:

1) I didn't update to pass `userId`, but rather continue to pass `user` into the `createAccount` mutation. This is because it seems like the `user` isn't yet saved to the database at this point, it's just data waiting to be saved, so when I pass `userId` and then try to look it up in the `createAccount` mutations, it's not found.

2) I updated the `Groups` based on your comment above, but left the `guest` group. Can update if you feel we should, but I don't think we've looked deep enough into where that `guest` account is created and how that all works. If `Customer` is essentially a `guest`, then that makes sense and we can update it here.

imports/plugins/core/accounts/server/init.js
aldeed: It does look like we're doing it doubly. Seems fine to just remove groupId.

PR #5580, opened 2019-09-25
Background jobs rewrite: no Meteor dependencies!

### LINKED ISSUE #5288: De-Meteorize job control system
This is part of the [EPIC Moving the API off Meteor](#4996). This issue is about de-meteorizing the background job system. Here are some initial notes I put together.

## How background jobs currently work in reaction

- Reaction includes a plugin (`imports/plugins/included/jobcontrol`) implementing job control
- `imports/utils/jobs` implements a job server and a job queue based on `@reactioncommerce/job-queue` which has a JobQueue class
  - This adds an event handler for meteor startup so the job server starts after meteor starts
- When the job server starts, some base jobs are scheduled
  - `jobControl/removeStaleJobs` runs daily
- `@reactioncommerce/job-queue` also includes our OpenExchangRates scraper job
  - Why?
- `@reactioncommerce/job-queue` has the base API and implementation
  - The API is medium sized, at least a dozen methods
  - Do we need to port all of these and maintain compatibility? (I presume yes)
  - Can we do this in a branch and convert off meteor
    - control flow with `async`/`await` instead of fibers
    - pure mongo collections instead of meteor ones
    - js `setTimeout` instead of `Meteor.setTimeout`
    - plain `Error` instead of `Meteor.error`

- There are 12 core or included plugins that use jobs
  - Note: presumably if none of these files imported `/imports/utils/jobs`, the job server would never install the meteor startup hook and thus never start.

## core & included plugins that use jobs

```
imports/plugins/included/jobcontrol/server/jobs/cleanup.js
imports/plugins/included/jobcontrol/server/jobs/exchangerates.js
imports/plugins/included/sitemap-generator/server/jobs/generate-sitemaps-job.js
imports/plugins/core/email/client/containers/emailLogs.js
imports/plugins/core/email/client/components/emailLogs.js
imports/plugins/core/email/server/methods/retryFailed.js
imports/plugins/core/core/server/startup/collection-security.js
imports/plugins/core/core/server/publications/email.js
imports/plugins/included/sitemap-generator/server/no-meteor/mutations/generateSitemaps.js
imports/plugins/core/email/server/no-meteor/util/processEmailJobs.js
imports/plugins/core/email/server/no-meteor/mutations/sendEmail.js
imports/plugins/core/core/server/methods/shop/updateShopExternalServices.js
```

### LINKED ISSUE #5312: Emails do not work on rc 2.0.0
## Prerequisites
* [x] Are you able to consistently reproduce the issues?

## Issue Description
No emails are sent.
[shopId is sent through inner field data](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/email/server/no-meteor/mutations/sendEmail.js#L66) but destructured as it is first field [renderEmail.js](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/included/email-templates/server/mutations/renderEmail.js#L14)

## Steps to Reproduce
1. Start reaction
2. Make any event that causes email to be sent, such as reset password, order notification etc.
3. In reaction console logs you will receive error message saying example "No email template found for language en"

## Possible Solution
Do not destructure shopId and user `data.shopId`.

[changes](https://github.com/voyego/reaction/pull/3/commits/2555d6f8e7233cbb0fcaf25cb82ffc5681cc03a4)

aldeed:
Resolves #5288
Resolves #5312
Impact: **major**  
Type: **feature**

## Changes
- New included `job-queue` plugin. It's a port of `@reactioncommerce/job-queue`, without any Meteor dependencies. Although it attempts to keep the same API, the loss of fibers means that functions that used to be magically synchronous now need to be `await`ed.
- Fully demeteorizes the `sitemap-generator` plugin. Updates its client code to use GraphQL. Updates refresh period setting to use the new `shopSettings` API.
- Removed the outdated exchange rate updater job. Setting the rates some other way should still work if you have custom plugins that rely on them.
- New APIs, more details below

### New background jobs API
The background jobs API is abstracted onto `context.backgroundJobs` so that you do not need to import from outside your own plugin to use it. (It is theoretically possible to replace the included handling with your own background job solution, if you simulate the same API exactly. This isn't tested or officially supported yet.)

To register a worker, typically done in your plugin startup function:

```js
  const workerInstance = await context.backgroundJobs.addWorker({
    type: "someUniqueJobType",
    async worker(job) {
        await doTheWork(job.data); // Whatever function you create that does the task
        job.done("Done message");
        // If anything throws, it will automatically call job.fail(errorMessage), but you
        // could also call job.fail yourself to provide better failure details.
    }
  });
```

To schedule a job for that worker, either on startup or anywhere you have context:

```js
  const job = await context.backgroundJobs.scheduleJob({
    type: "someUniqueJobType",
    data: {}, // any data your worker needs to perform the work
    // Retry is optional
    retry: {
      retries: 5,
      wait: 60000,
      backoff: "exponential"
    },
    // Priority is optional
    priority: "normal",
    // Schedule is optional if you just need to run it once.
    // Set to any text that later.js can parse.
    schedule: "every day",
    // Set cancelRepeats to true if you want to cancel all other pending jobs with the same type
    cancelRepeats: true
  });
```

To cancel all jobs that match a certain filter:

```js
await context.backgroundJobs.cancelJobs({
  type: "foo",
  data: { shopId }
});
```

To get a job by ID:

```js
const job = await context.backgroundJobs.getJob(jobId);
```

### Background job cleanup
Plugins can now request cleanup of their finished/old jobs through `registerPlugin`.

```js
app.registerPlugin({
  // ...
  backgroundJobs: {
      cleanup: [
        { type: "saveImage/local", purgeAfterDays: 7 },
        { type: "saveImage/remote", purgeAfterDays: 7 }
      ]
    }
});
```

### contextAdditions
A plugin can now add properties to context using the `contextAdditions` option when calling `registerPlugin`. They are added before any `startup` functions are run.

```js
app.registerPlugin({
  // ...
  contextAdditions: {
    something: "wicked"
  }
});
```

```js
// in startup fn or anywhere you have context
console.log(context.something); // "wicked"
```

## Breaking changes
Just one: There are no longer background jobs that keep the exchange rates up to date. Stock Reaction does not use this feature any more. If you are relying on it, everything should continue to work, but you will need to devise your own way to periodically fetch exchange rates and update the `rate` field.

Despite efforts to keep the API the same, the loss of fibers means that functions that used to be magically synchronous now need to be `await`ed. In many cases existing code would just work, but to be safe, I did not remove the non-Meteor job-collection files. Custom plugins using background jobs should switch to the new `context.backgroundJobs` API sometime before the 3.0 release, at which time the old job-queue package will be removed.

## Testing
Generally need to test anything that relies on job queue. Do all of these:
- Set up email sending and do anything that causes an email to be sent, such as placing an order or registering a new account. Verify you get the email. On Email settings page, find the email row and click retry icon. Verify you get the email again.
- Upload image files
- On shop settings page, test everything about the "Sitemaps" card. You should see logging in server console as you change the refresh period or do a manual refresh. When you view the sitemap, you should see XML.


imports/plugins/included/sitemap-generator/client/components/SitemapSettings.js
```diff
@@ -0,0 +1,129 @@
...
+import { i18next } from "/client/api";
+import Logger from "/client/modules/logger";
+import generateSitemapsMutation from "../mutations/generateSitemaps";
+import updateShopSettingsMutation from "../mutations/updateShopSettings";
+import shopSettingsQuery from "../queries/shopSettings";
+
+const useStyles = makeStyles(() => ({
+  card: {
+    // Without this, the Select dropdown menu gets cut off at the bottom of the card
+    overflow: "visible"
```
kieckhafer: @catalystpod (@mikemurray @machikoyasuda @willopez @rymorgan @cassytaylor)

See @aldeed's comment here. We've noticed this getting cut off in a few places, the `Origin country` of the Product admin comes to mind as one of them.

imports/plugins/included/sitemap-generator/client/components/SitemapSettings.js
```diff
@@ -0,0 +1,129 @@
...
+          },
+          shopId
+        }
+      }
+    });
+  };
+
+  return (
+    <Card className={classes.card}>
+      <CardHeader title={i18next.t("SitemapSettings.cardTitle")} />
```
kieckhafer: Style consistency question: Should these translation keys be camelCased?

imports/plugins/included/job-queue/server/no-meteor/job-queue/util.js
```diff
@@ -0,0 +1,140 @@
...
+  return [options, callback];
+}
+
+export function splitLongArray(arr, max) {
+  if (!(arr instanceof Array) || !(max > 0)) { throw new Error("splitLongArray: bad params"); }
+  return __range__(0, Math.ceil(arr.length / max), false).map((index) => arr.slice((index * max), ((index + 1) * max)));
+}
+
+// This function soaks up num callbacks, by default returning the disjunction of Boolean results
+// or returning on first error.... Reduce function causes different reduce behavior, such as concatenation
```
kieckhafer: Only three `.` in the ellipses

imports/node-app/index.js
```diff
@@ -1,22 +1,21 @@
...
-  debug: true,
+  debug: NODE_ENV !== "production",
   context: {
+    appVersion: packageJson.version,
     mutations: {},
     queries: {},
     rootUrl: ROOT_URL
-  },
-  graphQL: {
-    graphiql: true
```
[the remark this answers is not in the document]
aldeed: No this is no longer used. It uses env variables to determine whether to serve the playground UI.

imports/plugins/core/core/server/util/fetchCurrencyRate.js
```diff
@@ -1,95 +0,0 @@
-import _ from "lodash";
```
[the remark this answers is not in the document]
aldeed: We don't use the rates anywhere so I just removed this. They were complicated to demeteorize so it didn't seem worth the time for something unused. All other rate/conversion code is still there so someone could make their own job to fetch and set rates if needed. I did put this in the PR description, but I guess I should list it as a breaking change.

imports/plugins/core/email/server/methods/retryFailed.js
```diff
@@ -1,8 +1,8 @@
 import Logger from "@reactioncommerce/logger";
+import ReactionError from "@reactioncommerce/reaction-error";
 import { check } from "meteor/check";
-import { Jobs } from "/imports/utils/jobs";
 import Reaction from "/imports/plugins/core/core/server/Reaction";
-import ReactionError from "@reactioncommerce/reaction-error";
+import getGraphQLContextInMeteorMethod from "/imports/plugins/core/graphql/server/getGraphQLContextInMeteorMethod";
```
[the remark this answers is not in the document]
aldeed: Yes we'll need to move this Meteor method to GraphQL like the others after this is merged, but I didn't want to make this PR any longer.

imports/plugins/included/sitemap-generator/client/components/SitemapSettings.js
aldeed: Also, I first tried using the Catalyst Select, but that one does not accept a `value` prop with the current value so it's kind of useless for an update form such as this. I will submit a catalyst issue about that when I get a chance.

imports/plugins/included/sitemap-generator/client/components/SitemapSettings.js
aldeed: I copied what I had done in inventory, which is exactly matching the component name. I don't remember why I did it that way, though. It doesn't matter much to me, but I will follow a consistency rule if someone wants to make one.

PR #5582, opened 2019-09-26
Feature: integrate new Catalyst DataTable in the orders view

### LINKED ISSUE #5532: Orders Admin: View all Orders in DataTable
As a Reaction Admin user, I want to be able to see all Orders from the last 6 months in a table, so that I can see the latest Orders.

Catalyst components required: DataTable

GraphQL endpoints required: Orders

willopez:
Resolves #5532 
Impact: minor  
Type: feature

## Summary
Integrate the new Catalyst DataTable into the orders view with support searching by order id, referenced id, customer name, phone number, email. In the `Orders` collection, indices for `payments.address.fullName`, `shipping.address.fullName` and `payments.address.phone` have been added.

Further, this PR also removes old Meteor dependent orders view components.

## Testing
1. Search for an order by either: order id, referenced id, customer name, phone number, email, verify the order is found.


imports/plugins/core/orders/client/components/OrderDateCell.js
```diff
@@ -0,0 +1,44 @@
...
+import PropTypes from "prop-types";
+import moment from "moment";
+
+/**
+ * @name OrderDateCell
+ * @param {Object} row A react-table row object
+ * @return {React.Component} A date component
+ */
+export default function OrderDateCell({ row }) {
+  if (!moment) return null;
```
mikemurray: Probably don't need this check

rymorgan:
- I thought the Order ID search was supposed to show matches after 4 exact match characters? It only works as an exact match. 
- I can’t test email, or phone b/c I don’t see any orders that have that info associated with them. 
- Is search by address suppose to be working? It doesn’t seem to be working.

mikemurray:
@cassytaylor 
> Column title row height is too short - should be 30px (32px)

This can be done in Catalyst, it seems too short there as well

> Not seeing an empty or loading state

This is captured in this ticket: https://github.com/reactioncommerce/catalyst/issues/110. 

> Is there any way to lock the column width based on its width on first page of the table, or to set a default width for each row (aka, not determine column width by its content?) The columns change widths on subsequent pages, and it feels jumpy

I was looking into that in Catalyst but didn't have a solution that worked. Still looking to it, especially fo the checkbox" column

mikemurray:
@cassytaylor some updates:

- Right-aligned total column
- Updated date display

image

The others I think can be in Catalyst, or additional reaction PR's which I'm going to capture in another ticket. Some of that was captured above https://github.com/reactioncommerce/reaction/pull/5582#issuecomment-537694469

janus-reith:
This contains a lot of fallback Translations - I thought these should not be used anymore.
Also some of the translations in the JSON seem to be missing or at the wrong place (e.g. `fulfillmentGroupHeader` which should be on `order` but is on `reaction-orders`).

Issue #5618, opened 2019-09-27
[demeteorization] move `shipping` package


kieckhafer:
Move package files from its' respective `no-meteor` folder into a `node-app` folder.

kieckhafer:
Closed via #5632

PR #5632, opened 2019-09-30
refactor: move `shipping` and `surcharges` from server/no-meteor to node-app

### LINKED ISSUE #5620: [demeteorization] move `surcharges` package

kieckhafer:
Resolves #5620 & #5618  
Impact: **minor**  
Type: **refactor**

This PR moves both the `shipping` and `surcharges` packages, as `surcharges` relies on `shipping`.

## Breaking changes
Any custom plugins that import files directly from the `address` package will need to be updated.

## Testing
1. Test any GraphQL mutation that has been moved in this process, such as:

```
query {
  surchargeById(surchargeId: "SURCHARGE_ID", shopId: "SHOP_ID") {
    _id
    amount
  }
}
```


kieckhafer:
@aldeed this commit - https://github.com/reactioncommerce/reaction/pull/5632/commits/75af3b3436bc4e8cf22b62bed39001f28c53cc55 - creates a new query and is more in depth than just moving files between packages, please double check that one.

imports/plugins/core/cart/server/no-meteor/queries/getCommonOrderForCartGroup.js
```diff
@@ -0,0 +1,45 @@
...
+    fulfillmentGroupId
+  } = input;
+
+  if (!cartId) {
+    throw new ReactionError("invalid-param", "You must provide a cart ID");
+  }
+
+  if (!fulfillmentGroupId) {
+    throw new ReactionError("invalid-param", "You must provide a fulfillment group ID");
+  }
```
aldeed: Neither of these checks are necessary because `inputSchema.validate` would have already thrown if these aren't set.

imports/plugins/core/cart/server/no-meteor/queries/getCommonOrderForCartGroup.js
```diff
@@ -0,0 +1,45 @@
...
+
+  const cart = await Cart.findOne({ _id: cartId });
+  const group = cart.shipping.find((grp) => grp._id === fulfillmentGroupId);
```
aldeed: After each of these you should throw if `cart` or `group`, respectively, is undefined.

imports/plugins/core/cart/server/no-meteor/resolvers/Query/getCommonOrderForCartGroup.js
```diff
@@ -0,0 +1,22 @@
...
+ * @returns {Promise<Object>|undefined} A CommonOrder object
+ */
+export default async function getCommonOrderForCartGroup(parentResult, args, context) {
+  const { cartId, fulfillmentGroupId } = args;
+
+  return context.queries.getCommonOrderForCartGroup(context, {
+    cartId: decodeCartOpaqueId(cartId),
+    fulfillmentGroupId: decodeFulfillmentGroupOpaqueId(fulfillmentGroupId)
+  });
+}
```
aldeed: Sorry for not being clear. I don't think it's a good idea to have this query in GraphQL. It should be for internal use only. Maybe this would change at some point but for now CommonOrder is more of an internal concept among plugins that doesn't need to be part of the GraphQL schema.

kieckhafer:
@aldeed comments addressed

PR #5635, opened 2019-10-01
refactor: move shop from server/no-meteor to node-app

### LINKED ISSUE #5621: [demeteorization] move `shop` package

kieckhafer:
Resolves #5621 
Impact: **minor**  
Type: **refactor**

## Breaking changes
Any custom plugins that import files directly from the `shop` package will need to be updated.

## Testing
1. Test any GraphQL mutation that has been moved in this process, such as:

```
query {
  primaryShopId() {
    _id
  }
}
```


kieckhafer:
@aldeed updated with new file in both `/imports/utils` with the alias, and also pushed to the `api-utils` repo.

imports/node-app/core-services/shop/mutations/updateShop.js
```diff
@@ -2,7 +2,6 @@ import SimpleSchema from "simpl-schema";
 import ReactionError from "@reactioncommerce/reaction-error";
 import { ShopLogoUrls, StorefrontUrls } from "/imports/collections/schemas";
```
kieckhafer: @aldeed I think I might need to do something here with these schemas.

Do all the schemas in the `shops` file need to be moved into this package? Or just the two imported here?

https://github.com/reactioncommerce/reaction/blob/refactor-kieckhafer-movePackageToNodeApp-shop/imports/collections/schemas/shops.js#L207

aldeed: Seems best to move all. I did it for you after merging in develop in https://github.com/reactioncommerce/reaction/pull/5635/commits/80633e67d8a1744a1acbf3627c62626ccce014f1

PR #5642, opened 2019-10-02
Move files methods to GraphQL and files plugin to Node app

### LINKED ISSUE #5595: [demeteorization] move `files` package

aldeed:
Resolves #5595   
Impact: **minor**  
Type: **feature**

## Changes
- Remove all Meteor methods related to media
- Create create, update, delete, archive GraphQL mutations for media records
- Update Reaction Admin to call GraphQL mutations instead of Meteor methods
- Create more specific roles for media permission checks: `media/create`, `media/update`, `media/delete`
- Remove the `toGrid` image metadata that isn't used. It's still in the GraphQL schema for backwards compatibility but is removed from the database by a migration, and will always be `1` when retrieved through GraphQL.
- Move the `files` plugin to the Node app `core-services` folder

## Breaking changes
The only breaking change is the removal of the "media/*" Meteor methods. If you have custom code that calls them, update it to call one of the new GraphQL mutations instead.

## Testing
1. Upload and remove shop logo
1. Upload, reorder, and remove product images
1. Upload, reorder, and remove product variant images


aldeed:
@kieckhafer The only way that could happen is if you are logged in as a user who has no associated account. Maybe something that got messed up in your database when you were playing with the accounts PR?

kieckhafer:
@aldeed Def. seems like that was the case. On a fully fresh app it works as described.

PR #5650, opened 2019-10-03
refactor: move simple-inventory from server/no-meteor to node-app

### LINKED ISSUE #5605: [demeteorization] move `simple-inventory` package

aldeed:
Resolves #5605 
Impact: **minor**  
Type: **feature**

## Breaking changes
Any custom plugins that import files directly from the `simple-inventory` package will need to be updated.

## Testing
It should be sufficient to log in to Reaction Admin and adjust some inventory fields, publish to catalog.


aldeed:
@kieckhafer Your change is fine for now. At some point in the splitting process it will likely cause problems and we can figure out a solution when we get there.

PR #5660, opened 2019-10-03
[WIP] Integration tests for addCartItems

### LINKED ISSUE #5645: Add Integration Test for addCartItems GraphQL mutation
Epic: #5310

chrispotter:
Signed-off-by: Chris Potter <chris@reactioncommerce.com>

Resolves #5645 
Impact: **minor**  
Type: **test**

## Testing
1. List the steps needed for testing your change in this section.
2. Assume that testers already know how to start the app, and do the basic setup tasks.
3. Be detailed enough that someone can work through it without being too granular


tests/integration/api/mutations/addCartItems/addCartItems.test.js
```diff
@@ -0,0 +1,119 @@
...
+let cartItemInputSchemaFactory;
+
+
+
+beforeAll(async () => {
+  testApp = new TestApp();
+  await testApp.start();
+  shopId = await testApp.insertPrimaryShop();
+  addCartItems = testApp.mutate(AddCartItemsMutation);
+  cartItemInputSchemaFactory = createFactoryForSchema(
```
chrispotter: @aldeed 
 i'm not sure how else to accomplish this I need the input schema for the CartItems for this test so i can mock a correct order item submission.

is this accessible in the `testApp`?

chrispotter: which comes from [this](https://github.com/reactioncommerce/reaction/blob/develop/imports/plugins/core/cart/server/no-meteor/util/addCartItems.js) file...

chrispotter:
@aldeed do you think I'm getting to granular with the tests? I had a revelation overnight that maybe I'm treating this like a unit test rather than an integration test.  
What's our integration test philosophy? Mimicking user interaction? Testing all scenarios, errors and successes?

aldeed:
@chrispotter I think it's nice to test different auth scenarios because unit tests of those may be mocking something resulting in false success. Testing "not-found" cases is not as critical. It's also good for at least one success test to be asking for all GraphQL fields back because that ensures we are hitting and testing all relevant resolvers.

tests/integration/api/mutations/addCartItems/addCartItems.test.js
aldeed: You could code the whole object and not use a factory. It's a fairly small object. It also could be fine to export the schema from the plugin for test factory purposes. We don't have a great pattern for this yet.

willopez:
taking this over will create another PR

PR #5670, opened 2019-10-03
refactor: move orders from server/no-meteor to node-app

### LINKED ISSUE #5608: [demeteorization] move `orders` package

kieckhafer:
Resolves #5608 
Impact: **minor**  
Type: **refactor**

## Breaking changes
Any custom plugins that import files directly from the `orders` package will need to be updated.

## Testing
1. See that you can create an order on from the storefront
1. See that you can fulfill an order from the Reaction Admin


kieckhafer:
@aldeed I moved a lot of Simple Schemas around in this PR, please take a close look at those to make sure they are done correctly. Both a mix of exported, and internal schemas.

Issue #5679, opened 2019-10-04
[demeteorization] Update catalog.js meteor methods to GQL


kieckhafer:
We need to update the all meteor methods in `imports/plugins/core/catalog/server/methods/catalog.js` to use GQL. Anything calling `updateCatalogProduct`, which calls `hashProduct`, and anything with an appEvent, 
 specifically.

#### Using only `appEvents`:

- [ ] products/archiveProduct [In progress: @kieckhafer] https://github.com/reactioncommerce/reaction/pull/5680

- [ ] products/deleteVariant [In progress: @kieckhafer] https://github.com/reactioncommerce/reaction/pull/5680

#### Using `updateCatalogProduct`, which uses `hashProduct`:

- [ ] products/updateProductTags

- [ ] products/removeProductTag

- [ ] products/setHandle

- [ ] products/setHandleTag

- [ ] products/updateVariantsPosition

- [ ] products/updateMetaFields

- [ ] products/removeMetaFields

#### Using both: 

- [ ] products/updateProductField  [In progress: @kieckhafer] https://github.com/reactioncommerce/reaction/pull/5686

- [ ] products/toggleVisibility  [In progress: @kieckhafer] https://github.com/reactioncommerce/reaction/pull/5686

janus-reith:
@kieckhafer It seesm one of your changes reading this introcuded an error:
The `withProductForm` HOC defines a function `handleFieldChange`and passed a non-existent function  `handleProductFieldChange `:

https://github.com/reactioncommerce/reaction/blob/10ce7bb23cc4f71fe416e3115d0d119262f174d9/imports/plugins/included/product-admin/client/hocs/withProductForm.js#L143

jackson-sandland:
Is this part of a larger effort to replace `meteor.js`? Having 2 different front end frameworks is a nightmare.

kieckhafer:
@jackson-sandland Yes, we are moving away from Meteor, striving to get meteor completely removed within the next _major_ release. You can check out the `release-3.0.0` branch to see a meteor-free `api`, which will pair with a new repo for the UI.

aldeed:
Confirmed all of these that are needed now have GQL equivalents in 3.0

PR #5686, opened 2019-10-04
create `updateProductField` GQL mutation to replace meteor methods

kieckhafer:
Resolves parts of #5679   
Impact: **minor**  
Type: **feature|refactor**

## Issue
As part of our effort to demeteorize the app, we need to convert remaining meteor methods to GQL mutations / queries.

## Solution
This PR adds the following mutation: `updateProductField`

This PR removes the ` products/updateProductField` & `products/toggleVisibility` meteor methods.

Instead of creating a second mutation to clone the `products/toggleVisibility` meteor method, I've just confirmed that `toggleVisibility` works in this new mutation.

## Breaking changes
Any custom code using any of the above meteor methods will need to be refactored to use the new GQL methods.

## Testing
1. Start the app.
1. See that you can preform the following actions:
- Update any fields on a product
- Toggle the visibility of a product

## TODO
- Update all client code that calls `` to use `updateProductField` mutation instead
- Add `hashProduct` function to new `updateProductField` mutation
- Possibly add `hashProduct` to utils


imports/plugins/included/product-admin/client/hocs/withVariantForm.js
```diff
@@ -175,19 +185,35 @@ const wrapComponent = (Comp) => (
...
+                field: fieldName,
+                shopId: opaqueShopId,
+                productId: opaqueProductId,
+                value
+              }
+            }
+          });
 
-        if (fieldName === "lowInventoryWarningThreshold") {
-          this.updateLowInventoryThresholdIfChildVariants(variant);
```
kieckhafer: @aldeed I believe this, and also the other check for `updateInventoryPolicyIfChildVariants`, is now being done in the simple-inventory plugin, with the productConfiguration, can you confirm that's the case? https://github.com/reactioncommerce/reaction/blob/refactor-kieckhafer-convertCatalogMethodsToGQL-products/updateProductField/imports/node-app/plugins/simple-inventory/mutations/updateSimpleInventory.js#L54

janus-reith:
I noticed this is still in a draft state, and currently just works for String values. (You commented that as TODO in the schema).

I could work around this by registering a custom Scalar type `IntStringFloat`for this purpouse, so value is of type: `IntStringFloat`. I think the product fields currently are either String or Float, so these should be covered.

kieckhafer:
@janus-reith I was having issues with a CustomScaler, it seems that they don't work as we wanted them too for `input` (they work for `type`). If you're able to work something up, that would be great, we are still trying to figure out the best solution for this.

They are mostly strings, but there are `bool` and arrays of objects as well, it's a lot of possibilites. One solution we were contemplating is just creating a whole `input Product` type and passing the whole product.

janus-reith:
This would be the relevant excerpt for the 3.0 meteor-free server, but also worked on 2.7 to edit fields in the product admin.

`/product/schemas/product.graphql`

```
`scalar IntStringFloatBool`

[...]

input UpdateProductFieldInput  {
[...]
value: IntStringFloatBool!
}

input UpdateProductVariantFieldInput  {
[...]
value: IntStringFloatBool!
}

```

`/product/resolvers/IntStringFloatBool.js`

```
import { createRequire } from "module";

const require = createRequire(import.meta.url); // eslint-disable-line
const { GraphQLScalarType } = require("graphql");
const { Kind } = require("graphql/language");

const checkIntStringFloatBool = (value) => {
  if (Number.isInteger(value)) {
    if (value < -0x7FFFFFFF || value > 0x7FFFFFFF) {
      throw new TypeError(`Value is integer but outside of valid range for Int Scalar (A signed 32‐bit integer): ${String(value)}`);
    }
    return value;
  }
  return String(value);
}

const description = `
This could be a String, an Integer or Float value
`;

export default new GraphQLScalarType({
  description,
  name: "IntStringFloatBool",
  serialize: checkIntStringFloatBool,
  parseValue: checkIntStringFloatBool,
  parseLiteral(ast) {

    if (Array.isArray(ast.value)) {
      throw new TypeError(`IntStringFloatBool cannot represent an array value: [${String(value)}]`);
    }

    if (ast.kind === Kind.INT) {
      return checkIntStringFloatBool(parseInt(ast.value, 10));
    }

    if (ast.kind === Kind.FLOAT) {
      return parseFloat(ast.value);
    }

    if (ast.kind === Kind.STRING) {
      return ast.value;
    }

    if (ast.kind === Kind.BOOLEAN) {
      return ast.value;
    }

    throw new Error(`IntStringFloatBool found unexpected kind in parseLiteral: ${ast.kind}`);
  }
});
```

The scalar should probably be carefully checked for plausibility. The whole product input might be a better approach, as it could also be extended from plugin code.

Tested it with the Product admin for Text Inputs, Select Fields and Price input.

janus-reith:
Okay I noticed issues with this for numbers in the parsing step at `/product/mutations/updateProductField.js`.

Technically it seems like this would leave some cases open, for example distinguishing a Number from a String that contains a something that could be a number.

kieckhafer:
I've created a new PR that takes the new mutations and points them to `release-3.0.0`: https://github.com/reactioncommerce/reaction/pull/5768

That PR does not include any `client` side updates. Those will need to be done in `reaction-admin`.

kieckhafer:
Closing this in favor of #5768, which points to the 3.0 branch

PR #5692, opened 2019-10-07
chore: clean-up tasks related to removing `appEvents` from all meteor code

### LINKED ISSUE #5691: [demeteorization] clean-up: make sure `appEvents` isn't called in Meteor code.
Make sure nothing is left calling appEvents in Meteor code. I've done a quick scan and it seems these are the remaining places `appEvents` is used inside meteor code, with the two `discounts` methods being the only ones we will probably need to do a refactor on. This does not include where it's used in the `accounts`, `catalog`, or `products` plugins, as these three currently have PR's open and they _should_ be removed in those PRs.

- [`discounts/codes/apply`](https://github.com/reactioncommerce/reaction/blob/refactor-kieckhafer-movePackageToNodeApp-cleanupTasks/imports/plugins/included/discount-codes/server/methods/methods.js#L270)`
- [`discounts/codes/remove`](https://github.com/reactioncommerce/reaction/blob/refactor-kieckhafer-movePackageToNodeApp-cleanupTasks/imports/plugins/included/discount-codes/server/methods/methods.js#L140)`
- [`updateShipmentStatus`](https://github.com/reactioncommerce/reaction/blob/refactor-kieckhafer-movePackageToNodeApp-cleanupTasks/imports/plugins/core/orders/server/util/updateShipmentStatus.js#L53) I don't believe this function is called anywhere in the app - it doesn't even seem to be attached to a meteor method anymore - so we can mostly likely just delete this file
- [`shop/createShop`](https://github.com/reactioncommerce/reaction/blob/refactor-kieckhafer-movePackageToNodeApp-cleanupTasks/imports/plugins/core/core/server/methods/shop/createShop.js#L148) I don't believe this is used anymore, we can most likely just delete this function, along with the Blaze `Template` where it's being called.
- [`workflow/pushItemWorkflow`](https://github.com/reactioncommerce/reaction/blob/refactor-kieckhafer-movePackageToNodeApp-cleanupTasks/imports/plugins/core/checkout/server/methods/workflow.js#L87) We want to remove `coreItemWorkflow` anyway, maybe this is the time? In addition, this is called from the `checkout` plugin, which I don't believe we don't use anymore, so maybe we remove this plugin?

kieckhafer:
Resolves #5691   
Impact: minor
Type: feature|style|refactor|chore

Remove or refactor all code that calls `appEvents` from meteor code.

### Refactored into mutations
- [`discounts/codes/apply`](https://github.com/reactioncommerce/reaction/blob/develop/imports/plugins/included/discount-codes/server/methods/methods.js#L270)
- [`discounts/codes/remove`](https://github.com/reactioncommerce/reaction/blob/develop/imports/plugins/included/discount-codes/server/methods/methods.js#L140)

### Removed
- [`updateShipmentStatus`](https://github.com/reactioncommerce/reaction/blob/develop/imports/plugins/core/orders/server/util/updateShipmentStatus.js#L53) I don't believe this function is called anywhere in the app - it doesn't even seem to be attached to a meteor method anymore - so we can mostly likely just delete this file
- [`workflow/pushItemWorkflow`](https://github.com/reactioncommerce/reaction/blob/develop/imports/plugins/core/checkout/server/methods/workflow.js#L87) We want to remove `coreItemWorkflow` anyway, maybe this is the time? In addition, this is called from the `checkout` plugin, which I don't believe we don't use anymore, so maybe we remove this plugin?

### TODO
- [`shop/createShop`](https://github.com/reactioncommerce/reaction/blob/develop/imports/plugins/core/core/server/methods/shop/createShop.js#L148) I don't believe this is used anymore, we can most likely just delete this function, along with the Blaze `Template` where it's being called.

## Testing
1. See that the app functions as expected
1. See that discount code applying and removing works with new mutations. We have no UI for this at the moment, so you need to do it by creating a Cart, then taking that cart data and using `GraphiQL` or another GQL interface to perform the action

```
mutation {
  applyDiscountCodeToCart(input: {
    cartId: "ENCODED_CART_ID",
    discountCode: "CODE",
    shopId: "SHOP_ID",
    token: null
  }) {
    clientMutationId
  }
}
```

A discount code is a payment, so the discount code id should be encoded with `reaction/payment`
```
mutation {
  removeDiscountCodeFROMCart(input: {
    cartId: "ENCODED_CART_ID",
    discountCodeID: "ENCODED_PAYMENT_ID",
    shopId: "SHOP_ID",
    token: null
  }) {
    clientMutationId
  }
}
```


kieckhafer:
@aldeed The following Meteor methods have been removed or refactored to no longer use `appEvents`:
`discounts/codes/apply`
`discounts/codes/remove`
`updateShipmentStatus`
`workflow/pushItemWorkflow`

I believe this only leaves `shop/createShop` as the last place where we use `appEvents` in Meteor. I forgot if we said that was being addressed in the `accounts` PR or not.

I think either way, this can be reviewed / merged to incorporate all the other changes, and we can handle `shop/createShop` separately, so I'd say this is ready for a review.

imports/node-app/plugins/discount-codes/mutations/applyDiscountCodeToCart.js
```diff
@@ -0,0 +1,134 @@
...
+    cart = await Cart.findOne({ _id: cartId });
+    if (!cart) {
+      throw new ReactionError("not-found", "Cart not found");
+    }
+
+    // TODO: figure out the correct permission check here
+    // Should it be `discounts`, or `cart`?
+    if (!userHasPermission(["owner", "admin", "discounts"], shopId)) {
+      throw new ReactionError("access-denied", "Access Denied");
+    }
```
aldeed: Since `discounts` is more for defining/editing discounts, I'd create a new role `discounts/apply` for here.

imports/node-app/plugins/discount-codes/mutations/applyDiscountCodeToCart.js
```diff
@@ -0,0 +1,134 @@
...
+  const { cartId, discountCode, shopId, token } = input;
+  const { collections, userHasPermission, userId } = context;
+  const { Cart, Discounts } = collections;
+
+  // TODO: figure out the correct permission check here
+  // Should it be `discounts`, or `cart`?
+  // How do we determine this check if the user is the cart owner?
+  if (!userHasPermission(["admin", "owner", "discounts"], shopId)) {
+    throw new ReactionError("access-denied", "Access Denied");
+  }
```
aldeed: This one seems unnecessary given that `getCart` will already limit to carts that belong to the account/token, and then you have a separate admin check below it.

imports/node-app/plugins/discount-codes/mutations/applyDiscountCodeToCart.js
```diff
@@ -0,0 +1,134 @@
...
+    if (!shopObj[item.shopId]) {
+      shopObj[item.shopId] = true;
+    }
+    return shopObj;
+  }, {});
+  const participatingShops = Object.keys(uniqueShopObj);
+
+  if (participatingShops.length > 1) {
+    throw new ReactionError("not-implemented", "discounts.multiShopError", "Discounts cannot be applied to a multi-shop cart or order");
+  }
```
aldeed: Nothing from line 63 down to here is necessary. Carts can no longer contain items from multiple shops.

imports/node-app/plugins/discount-codes/mutations/applyDiscountCodeToCart.js
```diff
@@ -0,0 +1,134 @@
...
+  // check limits
+  if (conditions) {
+    if (conditions.accountLimit) accountLimitExceeded = conditions.accountLimit <= userCount;
+    if (conditions.redemptionLimit) discountLimitExceeded = conditions.redemptionLimit <= orderCount;
+  }
+
+  // validate basic limit handling
+  if (accountLimitExceeded === true || discountLimitExceeded === true) {
+    return { i18nKeyLabel: "Code is expired", i18nKey: "discounts.codeIsExpired" };
```
aldeed: I'd prefer to throw a ReactionError and let the client translate the error to the i18n key.

imports/node-app/plugins/discount-codes/mutations/removeDiscountCodeFromCart.js
```diff
@@ -0,0 +1,67 @@
...
+  const { cartId, discountCodeId, shopId, token } = input;
+  const { collections, userHasPermission } = context;
+  const { Cart } = collections;
+
+  // TODO: figure out the correct permission check here
+  // Should it be `discounts`, or `cart`?
+  // How do we determine this check if the user is the cart owner?
+  if (!userHasPermission(["admin", "owner", "discounts"], shopId)) {
+    throw new ReactionError("access-denied", "Access Denied");
+  }
```
aldeed: Same comment as other file. This check doesn't seem necessary, and the later check can use a new `discounts/apply`. Doesn't seem necessary to have a separate `discounts/remove` role. If you can apply, you should be able to remove, right?

imports/node-app/plugins/discount-codes/mutations/removeDiscountCodeFromCart.js
```diff
@@ -0,0 +1,67 @@
...
+    cart = await Cart.findOne({ _id: cartId });
+    if (!cart) {
+      throw new ReactionError("not-found", "Cart not found");
+    }
+
+    // TODO: figure out the correct permission check here
+    // Should it be `discounts`, or `cart`?
+    if (!userHasPermission(["owner", "admin", "discounts"], shopId)) {
+      throw new ReactionError("access-denied", "Access Denied");
+    }
```
aldeed: Since you already have `shopId`, you can move this permission check above the `Cart.findOne`, but be sure to pass `shopId` into the `findOne` query.

imports/node-app/plugins/discount-codes/mutations/removeDiscountCodeFromCart.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  // Instead of directly updating cart, we remove the discount billing
+  // object from the existing cart, then pass to `saveCart`
+  // to re-run cart through all transforms and validations.
+  const updatedCartBilling = cart.billing.filter((doc) => doc._id !== discountCodeId);
+  cart.billing = updatedCartBilling;
```
aldeed: You could add a check for whether `cart.billing.length` and `updatedCartBilling.length` are the same. If so, the discount code ID was invalid so throw an error.

imports/node-app/plugins/discount-codes/resolvers/Mutation/removeDiscountCodeFromCart.js
```diff
@@ -0,0 +1,40 @@
...
+    clientMutationId = null,
+    cartId,
+    discountCodeId,
+    shopId,
+    token
+  } = input;
+
+  const updatedCartWithRemovedDiscountCode = await context.mutations.removeDiscountCodeFromCart(context, {
+    cartId: decodeCartOpaqueId(cartId),
+    discountCodeId: decodePaymentOpaqueId(discountCodeId),
```
aldeed: Even though we treat these as payments in the database, I'd make a separate namespace for it (reaction/discountCode) to keep it future-proof.

imports/node-app/plugins/discount-codes/mutations/removeDiscountCodeFromCart.js
kieckhafer: I think that makes sense. Will keep `discounts/apply`

kieckhafer:
@aldeed comments addressed.

PR #5693, opened 2019-10-07
Move accounts plugin to Node app and better split accounts from users/IDP

### LINKED ISSUE #5213: Allow plugins to register Express middleware
Some plugins need to register Express middleware, either globally or for certain routes, and have some level of control over where it is inserted in the middleware chain.

Proposed stages:

1. first
2. before-authenticate
3. authenticate
4. before-response
5. response
6. before-error
7. error

A plugin would register its middleware function along with one of these points, plus choose global, graphql, or specific route paths. Maybe something like this in `registerPlugin`:

```js
{
  expressMiddleware: [
    {
      route: "graphql", // "all" or "graphql" are special strings, else assume it's an exact route
      stage: "before-error",
      fn: middlewareFunction
    }
  ]
}
```

I propose adding this only [to the Node Express app](https://github.com/reactioncommerce/reaction/blob/master/imports/node-app/core/createApolloServer.js#L66-L75). Currently, the Node Express app is then itself added as middleware in the Meteor Connect chain when running within Meteor, but since the plan is to soon pull the API out of Meteor, I assume this should be fine.

Probably the trickiest thing to get right here is the `error` type. The general requirement is some level of control over whether error middleware comes before or after whatever built-in error middleware there is. I think some of this happens in the guts of Apollo Server, so we may not have precise control without requesting changes to that library, or we may have to initialize the server in a less standard way.

### LINKED ISSUE #5588: [demeteorization] move `account` package

aldeed:
Resolves #5588  
Resolves #5213
Impact: **minor**  
Type: **refactor**

## Changes
- Move the "account" plugin to the Node app `core-services` folder
- Move `createUser` into the account plugin. It's still relying on Meteor, but that's OK until we split the apps.
- Reaction Admin no longer creates automatic "anonymous" users in the system, and a migration deletes all anonymous users and accounts. These have not been used since 2.0.0, but the code creating them was never removed from Reaction Admin.
- Move all auth logic and middleware into the account plugin (see below)

### New Middleware Pattern
Plugins can register Express middleware using `registerPlugin`:

```js
    expressMiddleware: [
      {
        route: "graphql",
        stage: "authenticate",
        fn: tokenMiddleware
      }
    ],
```

For now, only "graphql" route is supported, and the following stages are supported in this order:
- `first`
- `before-authenticate`
- `authenticate`
- `before-response`

An `authenticate` middleware function should do something like look up the user by the Authorization header, and either set `request.user` or send a 401 response if the token is invalid. It should not require a token. This is what the built-in `account` service now does.

The `first` middleware stage can be used for loggers or anything else that needs to be first in the middleware list. `before-response` middleware will have the user available if there is one, and is called before the Apollo GraphQL middleware.

A middleware function is passed `context` and must return the Express middleware handler function, which must call `next()` or send a response.

### Auth functions
Plugins can also define auth-related functions when calling `registerPlugin`. The built-in accounts service does this.

```js
    auth: {
      accountByUserId,
      getHasPermissionFunctionForUser,
      getShopsUserHasPermissionForFunctionForUser
    },
```

These are all functions that are called on every request, when adding request-specific properties to `context`.
- `accountByUserId(context, userId)`: Must return an account for the user ID, if one can be found
- `getHasPermissionFunctionForUser(context)`: Must return a function that matches the `context.userHasPermission` signature and returns `true` or `false`. This function can assume `context.user`/`context.account` is the user/account to check.
- `getShopsUserHasPermissionForFunctionForUser(context)`: Must return a function that matches the `context.shopsUserHasPermissionFor` signature and returns an array of shop IDs. This function can assume `context.user`/`context.account` is the user/account to check.

## Testing
- Verify Reaction Admin works, especially logging in / out.
- Verify that authenticated GraphQL requests continue to work. Clicking around in Reaction Admin, such as product and order tables, is enough to verify this.


imports/node-app/core-services/account/mutations/sendResetAccountPasswordEmail.js
```diff
@@ -2,7 +2,7 @@ import _ from "lodash";
 import SimpleSchema from "simpl-schema";
 import Logger from "@reactioncommerce/logger";
 import ReactionError from "@reactioncommerce/reaction-error";
-import generateVerificationTokenObject from "../util/generateVerificationTokenObject";
+import generateVerificationTokenObject from "@reactioncommerce/api-utils/generateVerificationTokenObject.js";
```
kieckhafer: Import order should be updated

```suggestion
import SimpleSchema from "simpl-schema";
import generateVerificationTokenObject from "@reactioncommerce/api-utils/generateVerificationTokenObject.js";
import Logger from "@reactioncommerce/logger";
import ReactionError from "@reactioncommerce/reaction-error";
```

imports/node-app/core-services/account/util/sendVerificationEmail.js
```diff
@@ -1,7 +1,7 @@
 import _ from "lodash";
 import Logger from "@reactioncommerce/logger";
 import ReactionError from "@reactioncommerce/reaction-error";
-import generateVerificationTokenObject from "/imports/plugins/core/accounts/server/no-meteor/util/generateVerificationTokenObject";
+import generateVerificationTokenObject from "@reactioncommerce/api-utils/generateVerificationTokenObject.js";
```
kieckhafer: Import order should be updated

```suggestion
import _ from "lodash";
import generateVerificationTokenObject from "@reactioncommerce/api-utils/generateVerificationTokenObject.js";
import Logger from "@reactioncommerce/logger";
import ReactionError from "@reactioncommerce/reaction-error";
```

imports/node-app/core/ReactionNodeApp.js
```diff
@@ -5,23 +5,32 @@ import mongodb, { MongoClient } from "mongodb";
 import Logger from "@reactioncommerce/logger";
 import appEvents from "./util/appEvents";
 import collectionIndex from "/imports/utils/collectionIndex";
```
kieckhafer:
```suggestion
import collectionIndex from "@reactioncommerce/api-utils/collectionIndex.js";
```

imports/plugins/core/accounts/server/init.js
```diff
@@ -1,11 +1,10 @@
 import _ from "lodash";
-import Logger from "@reactioncommerce/logger";
 import ReactionError from "@reactioncommerce/reaction-error";
 import { Meteor } from "meteor/meteor";
 import { Accounts } from "meteor/accounts-base";
 import * as Collections from "/lib/collections";
 import Reaction from "/imports/plugins/core/core/server/Reaction";
-import generateVerificationTokenObject from "/imports/plugins/core/accounts/server/no-meteor/util/generateVerificationTokenObject";
+import generateVerificationTokenObject from "@reactioncommerce/api-utils/generateVerificationTokenObject.js";
```
kieckhafer: Update import order

```suggestion
import _ from "lodash";
import generateVerificationTokenObject from "@reactioncommerce/api-utils/generateVerificationTokenObject.js";
import Logger from "@reactioncommerce/logger";
import ReactionError from "@reactioncommerce/reaction-error";
import { Meteor } from "meteor/meteor";
import { Accounts } from "meteor/accounts-base";
import * as Collections from "/lib/collections";
import Reaction from "/imports/plugins/core/core/server/Reaction";
import getGraphQLContextInMeteorMethod from "/imports/plugins/core/graphql/server/getGraphQLContextInMeteorMethod";
```

imports/plugins/core/accounts/server/methods/sendResetPasswordEmail.js
```diff
@@ -5,7 +5,7 @@ import { check } from "meteor/check";
 import { Meteor } from "meteor/meteor";
```
kieckhafer: Update import order

```suggestion
import { Meteor } from "meteor/meteor";
import generateVerificationTokenObject from "@reactioncommerce/api-utils/generateVerificationTokenObject.js";
import ReactionError from "@reactioncommerce/reaction-error";
import { Shops } from "/lib/collections";
import Reaction from "/imports/plugins/core/core/server/Reaction";
import getGraphQLContextInMeteorMethod from "/imports/plugins/core/graphql/server/getGraphQLContextInMeteorMethod";
```

aldeed:
Made suggested changes

Issue #5689, opened 2019-10-07
Add Integration Test for orderByReferenceId GraphQL query


mpaktiti:
Epic: #5310

PR #5697, opened 2019-10-08
Add Integration Test for orderByReferenceId GraphQL query

mpaktiti:
Resolves: https://github.com/reactioncommerce/reaction/issues/5689
Impact: **minor**
Type: **test**

This PR adds: an integration test file for [orderByReferenceId query](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/orders/server/no-meteor/queries/orderByReferenceId.js). 

We test whether an order is successfully returned, given an `orderReferenceId` and a `shopId`. Here is the [schema](https://github.com/reactioncommerce/reaction/blob/master/imports/plugins/core/orders/server/no-meteor/schemas/schema.graphql#L74) for this query. We test for logged-in and anonymous users (`setLoggedInUser` will fake as if your auth token is for that user).

#### Testing Instructions:

```
npm install --save-dev jest-cli
npm run lint
npm run test
```


tests/integration/api/queries/orderByReferenceId/orderByReferenceId.test.js
```diff
@@ -0,0 +1,108 @@
...
+// Mock data for logged-in user test
+const mockShopId = "integ-test-shop-id";
+const opaqueShopId = "cmVhY3Rpb24vc2hvcDppbnRlZy10ZXN0LXNob3AtaWQ=";
+const shopName = "Test Shop";
+const orderId = "integ-test-order-id";
+const orderReferenceId = "123456789";
+const accountId = "integ-test-account-id";
+const opaqueAccountId = "cmVhY3Rpb24vYWNjb3VudDppbnRlZy10ZXN0LWFjY291bnQtaWQ=";
+
+const mockOrdersAccount = Factory.Accounts.makeOne({
```
chrispotter: This should be `Factory.Account` instead of `Accounts`.

mpaktiti: Fixed, thanks!

PR #5715, opened 2019-10-10
[Fix] Create and Delete navigation item mutations

trojanh:
Resolves none  
Impact: **breaking**  
Type: **bugfix**

## Issue
Creating Navigation Item from Reaction Admin UI breaks for anyone other than the owner. 

## Solution
Supplying `shopId` to `userHasPermission` fixes the above issue and the non-owner user is able to add the NavigationItem.

## Breaking changes
Steps to reproduce:

1. Login as Shop Manager or someone who has `core` permission but not as an `Owner`
2. Go to Navigation page and try adding a new Navigation Item
3. You will see the form to add the NavigationItem and after saving the item the `NavigationItem` doesn't appear in the list. 

This works fine for the Owner. The problem was the method was called with a missing parameter for 
`userHasPermission` which also expects `roleGroup` parameter which is `shopId` for us.

## Testing
1. Login as Shop Manager or someone who has `core` permission but not as an `Owner`
2. Go to Navigation page and try adding a new Navigation Item
3. You will see the form to add the NavigationItem and after saving the item the `NavigationItem` appears on the list.


trojanh:
I think `navigationTreeById` query should also have `shopId` parameter here 
https://github.com/reactioncommerce/reaction/blob/da25e657b3b3920e3187ce872842edfa19f15cea/imports/node-app/plugins/navigation/queries/navigationTreeById.js#L24

trojanh:
Thanks, @mikemurray for reviewing my PR, let me know if I could do anything to fix the failing tests.

aldeed:
@trojanh Thank you. Can you merge latest `develop` into this branch? That should fix the CI tests and allow us to be sure this doesn't break anything, but otherwise the code changes look good.

aldeed:
@trojanh Ping. See my comment above.

trojanh:
Sorry for the delay @aldeed it was Diwali so couldn't revert back, appreciate the reminder though. I have pulled the latest develop branch so it should fix the issue.

PR #5717, opened 2019-10-10
refactor: use new URL instead of url.parse

### LINKED ISSUE #5704: [3.0.0] Update url.parse usage
Change all `url.parse` to `new URL()`. See https://nodejs.org/api/url.html#url_url

kieckhafer:
Resolves #5704
Impact: **minor**  
Type: **refactor**

Change all url.parse to new URL(). See https://nodejs.org/api/url.html#url_url

## Breaking changes
None

## Testing
1. See that the api starts
1. Try the `primaryShopId` query, which uses `new URL`, see if it returns the ID


src/core-services/core/util/loadSampleData.js
```diff
@@ -44,7 +41,7 @@ export default async function loadSampleData(context) {
       throw new Error("More than one primary shop in the sample dataset");
     }
 
-    const currentDomain = rootUrl && url.parse(rootUrl).host;
+    const currentDomain = rootUrl && new URL(rootUrl).host;
```
aldeed: Most places are looking up shop by domain using `hostname` but here we set it based on `host`. One includes the port and one does not. Should the other places be changed to `host`? I'm thinking whichever one doesn't have port is probably correct for this situation.

kieckhafer: Looks like `hostname` does not have port: https://nodejs.org/api/url.html#url_url

Updated

kieckhafer:
@quophyie addressed @aldeed's comments, ready for a look.

PR #5718, opened 2019-10-10
Refactor kieckhafer remove deprecations

### LINKED ISSUE #5703: [3.0.0] Remove everything marked deprecated
Search and remove all code and GraphQL schema items that are marked as deprecated.

kieckhafer:
Resolves #5703 
Impact: **breaking**  
Type: **refactor|chore**

## Issue
We have deprecated code left over from previous releases, which we can remove on the breaking `v3.0.0` release.

## Solution
Remove deprecated code.

## Breaking changes
Any custom code that uses any of the deprecated methods will need to be updated.
- `currentQuantity` should now use `inventoryAvailableToSell`
- `Job.makeJob(root, jobDoc)` should use `new Job(root, jobDoc)` instead
- `removeAccountFromGroup` GraphQL mutation is removed. You should use `addAccountToGroup`, which will remove the account from the group they were previously in.

## Testing
1. See that the api starts


src/core-services/cart/simpleSchemas.js
```diff
@@ -829,10 +828,6 @@ export const Cart = new SimpleSchema({
     type: CartAddress,
     optional: true
   },
-  "sessionId": {
-    type: String,
-    optional: true
-  },
```
aldeed: We probably shouldn't remove this one yet, unless we already have a migration that removes it from the database. Otherwise old carts will fail validation next time we try to update them. We don't yet have a way of doing migrations in 3.0

kieckhafer:
@aldeed comments addressed.

PR #5720, opened 2019-10-14
fix: security updates

focusaurus:
Impact: **|minor**  
Type: **security**

Update deps to address fresh vulns from snyk alert email.

Changes generated via `npm audit fix` with npm v6


kieckhafer:
Seems fine... 

Do we want to commit this to the release branch instead, and then it'll get pulled back into `develop` once the release is made?

focusaurus:
For 2.7.0? I guess OK.

PR #5722, opened 2019-10-15
Various 3.0.0 fixes

aldeed:
Impact: **breaking**  
Type: **feature**

## Changes
- Updated nodemon config to use `legacyWatch`. This is slower but it ensures that automatic restarts work when running in a development Docker container on Mac
- Set CORS header to allow all origins on both translation routes
- Ignore `.meteor` folder in Git. Although 3.0.0 is no longer a Meteor app, this helps when checking out 2.x and 3.x back and forth.
- Move much of the "core" plugin code to other places, such as the node-app core or some other relevant plugin.
- New `createShop` GraphQL mutation
- Various auth-related fixes. Mostly same-version fixes, which resulted from mistakes when removing the Meteor app
- Create an Account record if one is not found when an authenticated GraphQL request comes in. All new accounts are assigned to the "customer" auth group.
- Remove the setting of `shop.layout`. NOTE: You'll need the related changes in `reaction-admin`.
- On a fresh database, we no longer automatically create the first admin user or the primary shop. There are related changes in `reaction-admin` repo that are necessary for this to work. Now, it works like this:
    - First register a new user in Reaction Admin.
    - This will log you in to Reaction Admin, which causes it to make some GraphQL requests.
    - When the first GraphQL request comes, the account for your user is created. Because no existing "owner" user is found, it assigns you to the "owner" group instead of "customer" as usual.
    - Reaction Admin then detects that there is no primary shop. It hides most of the admin UI and shows a "Create first shop" form, which uses the new `createShop` GraphQL mutation to create the primary shop.
    - At this point the rest of the admin UI appears and works as it did in 2.x.

Note: There is still some work to do on the Reaction Admin Accounts page, which will be done in a future PR. In particular, inviting new shop managers does not work.

## Breaking changes from 2.x
- The first admin user and the primary shop are no longer automatically created as necessary when the API starts.
- The `primaryShop` and `primaryShopId` GraphQL queries now return only the shop with `shopType` of "primary". Previously there was extra logic that checked for shops with `domains` array that included the ROOT_URL hostname.
- `context.queries.primaryShopId` function now takes `context` as its first argument instead of `collections`

## Testing
Use the corresponding `reaction-admin` branch and verify that things work as described, especially on a blank database.


src/core-services/core/startup.js
```diff
@@ -9,51 +7,6 @@ import upsertPackages from "./util/upsertPackages.js";
  * @returns {undefined}
  */
 export default async function startup(context) {
-  const { appEvents } = context;
-
   await loadSampleData(context);
```
kieckhafer: Mentioned in another comment, but I'm getting errors with the sample data being loaded here. I think if we just adjust the sample data itself to set some fields differently it should be OK.

src/core-services/shop/queries/index.js
```diff
@@ -1,18 +1,9 @@
-/* eslint-disable node/no-deprecated-api */
-/* TODO: revisit `url.parse` throughout Reaction */
-import url from "url";
```
kieckhafer: There are a couple instances where you are removing `url`, which I've also done in this PR here:

Just making a note to check for conflicts whichever is merged first. https://github.com/reactioncommerce/reaction/pull/5717

PR #5723, opened 2019-10-16
refactor: only update fulfillment group for which id is provided

YuuwakU:
Impact: **minor**  
Type: **refactor**

## Issue
If there are more than one fulfillment groups, this mutation will update the workflow of all the groups.

## Solution
Adding arrayFilters to the collection update to make sure only the group, for which the id is provided, gets updated. Had to remove the schema modifier validation due to this: https://github.com/aldeed/meteor-collection2/issues/387

## Breaking changes
none

### DIFF
--- a/imports/node-app/core-services/orders/mutations/updateOrderFulfillmentGroup.js
+++ b/imports/node-app/core-services/orders/mutations/updateOrderFulfillmentGroup.js
@@ -1,6 +1,5 @@
 import SimpleSchema from "simpl-schema";
 import ReactionError from "@reactioncommerce/reaction-error";
-import { Order as OrderSchema } from "../simpleSchemas.js";
 
 const inputSchema = new SimpleSchema({
   tracking: {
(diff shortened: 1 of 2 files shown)


aldeed:
Thank you for the contribution @YuuwakU

YuuwakU:
Happy to assist @aldeed

machikoyasuda:
Hi @YuuwakU - I am going to revert this PR for now because the integration test is failing: https://circleci.com/gh/reactioncommerce/reaction/59315 Could you please look into this and make a new PR?

YuuwakU:
Hey @machikoyasuda  I see that the test for updateOrderFulfillmentGroup is passing in the link you mentioned. The integration tests seem to fail due to user permissions in other tests that are not related to these changes. And the only change here is to implement arrayFilters.

YuuwakU:
@machikoyasuda also, at the time I raised this PR, there were other PRs as well that had failure in tests which was identified here: https://github.com/reactioncommerce/reaction/pull/5715#pullrequestreview-300370751 
This may help if there are any others PRs that are being flagged to be reverted.

machikoyasuda:
@YuuwakU Hi! Thanks for your swift reply! We're looking into this. We  won't revert this PR most likely. Still trying to find where the tests are failing.

YuuwakU:
@machikoyasuda from what I found out while researching regarding the integration tests timing out, is that it is due to an issue with jest using one worker in --runInBand. Here are some references that may be helpful with the issue.

- https://discuss.circleci.com/t/jest-out-of-memory-error/32263/2
- https://vgpena.github.io/jest-circleci/
- https://jestjs.io/docs/en/cli#ci

machikoyasuda:
@YuuwakU Thank you. I closed the revert PR. Adding maxWorkers=3 and removing runInBand fixed the tests so we’re all good now.

PR #5724, opened 2019-10-16
feat: authorization plugin & checkPermissions functionality

kieckhafer:
Impact: **minor**  
Type: **feature|refactor**

## Issue
Our authorization / permission related code is currently living inside of our `accounts` package. With the work being done to incorporate `ory keto` as an authorization service, we need to break out these functions into their own package.

- for #5736
- for #5737

## Solution
1. Create a core-services `simple-authorization` package, and move authorization functionality from `accounts` into this new package
1. Create a `checkPermissions` function to replace `userHasPermission` in most places. Since we were essentially duplicating all `ReactionError` code throughout the app, this update throws the error inside `checkPermissions` so there is no need to throw it everywhere.

## Testing
1. Load the app and see that it works as it should. This change should change nothing, aside from the locations of some files. Specifically, you can:
- Try to admin an order or a product from a user that isn't allowed, see it doesn't work
- Try to admin an order or product from a user that is allowed, see it works


kieckhafer:
@aldeed For tests, this will probably have to wait until your jest fixes are merged to make sure they all work.

I've moved the auth stuff into a single `authorization` plugin for now, pending further discussion on if we need a `simple-auth` as well or what. This can easily be refactored into two plugins if we go that route.

src/core-services/orders/queries/ordersByAccountId.js
```diff
@@ -42,10 +42,8 @@ export default async function ordersByAccountId(context, { accountId, orderStatu
       const shopIdsUserHasPermissionFor = shopsUserHasPermissionFor("orders");
       query.shopId = { $in: shopIdsUserHasPermissionFor };
     } else {
-      shopIds.forEach((shopId) => {
-        if (!userHasPermission(["orders", "order/fulfillment"], shopId)) {
-          throw new ReactionError("access-denied", "Access Denied");
-        }
+      shopIds.forEach(async (shopId) => {
+        await checkPermissions(["orders", "order/fulfillment"], shopId);
```
kieckhafer: Is this OK to do in a `forEach`, or does it need to be done in a `for...of` instead?

src/core-services/orders/queries/orders.js
```diff
@@ -42,10 +41,8 @@ export default async function orders(context, { filters, shopIds } = {}) {
   // If an admin wants all orders for an account, we force it to be limited to the
   // shops for which they're allowed to see orders.
   if (shopIds) {
-    shopIds.forEach((shopId) => {
-      if (!userHasPermission(["orders", "order/fulfillment"], shopId)) {
-        throw new ReactionError("access-denied", "Access Denied");
-      }
+    shopIds.forEach(async (shopId) => {
+      await checkPermissions(["orders", "order/fulfillment"], shopId);
```
kieckhafer: Is this OK to do in a `forEach`, or does it need to be done in a `for...of` instead?

src/core-services/catalog/mutations/publishProducts.js
```diff
@@ -29,10 +29,8 @@ export default async function publishProducts(context, productIds) {
 
   if (!isInternalCall) {
     const uniqueShopIds = _.uniq(products.map((product) => product.shopId));
-    uniqueShopIds.forEach((shopId) => {
-      if (!userHasPermission(["createProduct", "product/admin", "product/publish"], shopId)) {
-        throw new ReactionError("access-denied", "Access Denied");
-      }
+    uniqueShopIds.forEach(async (shopId) => {
+      await checkPermissions(["createProduct", "product/admin", "product/publish"], shopId);
```
kieckhafer: Is this OK to do in a `forEach`, or does it need to be done in a `for...of` instead?

kieckhafer:
There is one remaining test failing here, it's the test for `buildContext`. It seems that `jasmine.any(Function)` is not working as I am expecting it to - or how it was before, considering these tests passed in the past.

ek_0x656b____Projects_reaction_reaction-platform_reaction__zsh_

kieckhafer:
@quophyie @rosshadden a lot more here than we discussed, but it's all groundwork and things @aldeed mentioned wanting anyway outside of auth-pod, so it made sense to do them here.

src/core-services/settings/mutations/updateAppSettings.js
```diff
@@ -24,10 +24,10 @@ export default async function updateAppSettings(context, settingsUpdates, shopId
   }
 
   // Look up roles that are allowed to set each setting. Throw if not allowed.
-  updateKeys.forEach((field) => {
+  updateKeys.forEach(async (field) => {
     const allowedRoles = shopId ? rolesThatCanEditShopSetting(field) : rolesThatCanEditGlobalSetting(field);
-    if (allowedRoles.length === 0 || !userHasPermission(allowedRoles, shopId)) {
-      throw new ReactionError("access-denied", `You are not allowed to edit the "${field}" setting`);
+    if (allowedRoles.length === 0) {
+      await checkPermissions(allowedRoles, shopId);
```
aldeed: Use `for...of`

aldeed: Also, you need to keep the `allowedRoles.length === 0` check but also do the `checkPermissions` check always.

```js
if (allowedRoles.length === 0) {
  throw new ReactionError("access-denied", `You are not allowed to edit the "${field}" setting`);
}
await checkPermissions(allowedRoles, shopId);
```

src/core-services/shop/mutations/createShop.js
```diff
@@ -83,18 +83,16 @@ export default async function createShop(context, input) {
...
       Shop: ShopSchema
     },
-    userHasPermission,
     userId
   } = context;
 
-  if (!userHasPermission(["owner", "shop/create"], "__global_roles__")) {
-    throw new ReactionError("access-denied", "User does not have permission");
-  }
+  await checkPermissions(["owner", "shop/create"], "__global_roles__");
```
aldeed: I think you could actually just remove `"__global_roles__"` (pass no second arg) and it would default to that. That will make it easier to move to a different permission plugin.

src/core-services/authorization/i18n/index.js
```diff
@@ -0,0 +1,10 @@
+import en from "./en.json";
+
+//
+// we want all the files in individual
+// imports for easier handling by
+// automated translation software
+//
+export default {
+  translations: [...en]
+};
```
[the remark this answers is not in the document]
kieckhafer: No, just didn't know if it was required. Removed.

kieckhafer:
@aldeed all comments addressed. Having some integration testing issues, the CI issues are differnet than what I'm seeing locally too. Locally I see this on all (34) tests that import GQL:

```
Test suite failed to run

    TypeError: (0 , _module.createRequire) is not a function

      19 | const { PubSub } = require("apollo-server");
      20 |
    > 21 | const coreGraphQLSchema = importAsString("./graphql/schema.graphql");
         |                           ^
      22 |
      23 | const {
      24 |   REACTION_GRAPHQL_SUBSCRIPTIONS_ENABLED,

      at importAsString (node_modules/@reactioncommerce/api-utils/lib/importAsString.js:14:19)
      at Object.<anonymous> (src/core/ReactionAPI.js:21:27)
      at Object.<anonymous> (tests/util/TestApp.js:7:1)
```

aldeed:
@kieckhafer Can you make the same fixes as in https://github.com/reactioncommerce/reaction/pull/5715/files? (pass shopId)

kieckhafer:
@aldeed I was wondering why there was no shopId on those, but was just swapping it out 1 for 1.

src/core-services/orders/queries/ordersByAccountId.js
rosshadden: You cannot just slap an `async` on a forEach function and expect it to work how you want, sorry :(. You should use a `for-of` loop with an `await` inside the block. We also have [`for-await-of`](https://node.green/#ES2018-features-Asynchronous-Iterators-for-await-of-loops) as of NodeJS 10.x.

In general I like to use normal for/for-in/for-of loops everywhere unless there is some situation in which a forEach actually gives us something useful (which is typically just so we don't have to keep an incrementing index variable, but that's pretty rare and thus so are forEach's). Even in such cases you have to be careful, as errors can get swallowed and stack traces can be confusing.

rosshadden:
I edited the OP to include references to the issues I just created for this work.

src/core-services/orders/queries/ordersByAccountId.js
kieckhafer: 👍 All these instances have already been refactored with a `for...of`. I thought that was the case when I was refactoring the code, figured with a `forEach` it would skip past it while it was running, but just wanted to double check before making the updates.

aldeed:
@kieckhafer Nice work updating to take shopId. I had that on my 3.0 list so now I can cross it off. 😄 I do see two minor issues:
- In `navigationTreeById`, you need to include `shopId` in the `findOne` query. Otherwise they could pass in any shop ID that they have permission for, even if it isn't the shop that owns the nav tree.
- In the `createNavigationItem` mutation, `shopId` is already on `NavigationItemInput`. Rather than needing to pass it in twice, you could pass `navigationItem.shopId` to `checkPermissions`

These changes will require corresponding changes in Reaction Admin client code

kieckhafer:
@aldeed I did see that it was already provided on `NavigationItemInput`. I added it anyway just to keep the same pattern of always passing it, but can take it off it it does seem excessive.

aldeed:
@kieckhafer I think if you kept it as is you'd need to be sure to set it in the new record and therefore remove it from the input. Or at the very least do a check that both shop IDs match. Otherwise they could pass one shop ID to satisfy the permission check yet create the record with a different shop ID. That's my only concern. The exact syntax of the params doesn't matter much to me.

kieckhafer:
@aldeed reverting it and using `navigationItem.shopId` works fine, so I did that.

PR #5730, opened 2019-10-18
Update user account name

### LINKED ISSUE #4646: Add firstName and lastName to GraphQL Account type
## Work
Add GraphQL resolvers that split `account.name` and return values for `firstName` and `lastName`. Client UI often needs these separated, and it doesn't make sense for every client to write that code.

Do a bit of research on whether we should be using different field names and how we should be splitting the `name`. Some cultures do not have last names or the surname is first. Is there a software schema best practice we can follow?

Arthur236:
Resolves #4646   
Impact: **minor**  
Type: **feature**

## Issue
[#4646](https://github.com/reactioncommerce/reaction/issues/4646)

## Solution
This PR adds a mutation called `updateAccountName` that takes in a name and splits it into firstName and lastName.

It then updates the fields: `name`, `firstName` and `lastName` and can be verified by calling the `viewer` query to ensure that the details are indeed updated.

## Breaking changes
None

## Testing
1. Open up the graphql playground by navigating to `/graphql-beta`
2. Type in the `updateAccountName` mutation and pass in the input object required. See below for example: 
```
mutation updateAccountName($accountId: ID!) {
  updateAccountName(input: {
    accountId: $accountId,
    name: "Test Name Last"
  }) {
    account {
      _id
    name
    firstName
    lastName
    primaryEmailAddress
    }
  }
}
```
3. The mutation should resolve successfully after having updated the name. The expected result should be something similar to the this:
```
{
  "data": {
    "updateAccountName": {
      "account": {
        "_id": "cmVhY3Rpb24vYWNjb3VudDozWjhqTW4zQmJkS0Z5cGFpRA==",
        "name": "Test Name Last",
        "firstName": "Test",
        "lastName": "Last",
        "primaryEmailAddress": "test.user@gmail.com"
      }
    }
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,46 @@
+# v2.7.0
+
+Reaction v2.7.0 adds performance enhancements and fixes bugs.
+
+This release is being coordinated with [Reaction Platform](https://github.com/reactioncommerce/reaction-platform) and is designed to work with `v2.7.0` of [Reaction Hydra](https://github.com/reactioncommerce/reaction-hydra) and [Example Storefront](https://github.com/reactioncommerce/example-storefront).
+
+## Notable changes
+
+### More packages have continued to migrate from `no-meteor` folders to `node-app`
+
+As part of de-meteorization of the Reaction API, We have continued the migration of moving all server side code in `no-meteor` folders to their respective `node-app` folders.
+
+## Fixes
+
+- fix: pass correct userId when creating a shop ([#5694](https://github.com/reactioncommerce/reaction/pull/5694))
+- fix: filter products by file on Windows ([#5700](https://github.com/reactioncommerce/reaction/pull/5700))
+
+## Refactors
+
+- refactor: Move accounts plugin to Node app and better split accounts from users/IDP ([#5693](https://github.com/reactioncommerce/reaction/pull/5693))
+- refactor: clean-up tasks related to removing `appEvents` from all meteor code ([#5692](https://github.com/reactioncommerce/reaction/pull/5692))
+- refactor: clean-up tasks related to moving files to the node-app ([#5688](https://github.com/reactioncommerce/reaction/pull/5688))
+- refactor: simpleSchema updates for mockContext ([#5685](https://github.com/reactioncommerce/reaction/pull/5685))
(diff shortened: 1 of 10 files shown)


yashutanna:
Thanks for this Arthur! But why not have first name and last name as parameters? Not sure why it’s necessary to split the names in code on not allow the caller of the api to choose how they want to store first and last names

Arthur236:
The reason I did it this way was to align to the issue request and also I figured that a user may want to have a name that is made up of 3 or more names and it didnt make sense to me to have lets say firstName or lastName as 2 or more names in order to achieve this. So for now I think this is a good approach.

yashutanna:
I don’t think I am convinced hey. With all due respect - it makes more sense to me to allow the caller of the API to define how they want the name split up into its component parts as opposed to the code doing it for them and possibly splitting it in a way that the caller of the API may not intend

On Oct 21, 2019, at 10:15, Arthur Thungu <notifications@github.com<mailto:notifications@github.com>> wrote:


The reason I did it this way was to align to the issue request and also I figured that a user may want to have a name that is made up of 3 or more names and it didnt make sense to me to have lets say firstName or lastName as 2 or more names in order to achieve this. So for now I think this is a good approach.

Arthur236:
Makes sense. I guess we can have the firstName and lastName as the parameters then combine them to make the name

yashutanna:
That is an approach that makes more sense to me 👍🏽

On Oct 21, 2019, at 10:24, Arthur Thungu <notifications@github.com<mailto:notifications@github.com>> wrote:


Makes sense. I guess we can have the firstName and lastName as the parameters then combine them to make the name

Arthur236:
I've made the requested changes

aldeed:
@Arthur236 Thanks for your good work here. We will not be merging this because 2.x is now on fixes-only track, but I've incorporated similar changes into https://github.com/reactioncommerce/reaction/pull/6067 for the 3.0.0 release.

We opted not to do any magic related to `firstName`, `lastName`, and `name`. They are simply string fields that get set to whatever you pass in. This is the best solution due to the many different types of names in the world. Client (admin or storefront browser UI) code can collect names however makes sense, concat or not, whatever, and then just pass in the information.

Other than that, the only change is that it's a general `updateAccount` mutation that supports updating more than just names.

PR #5744, opened 2019-10-24
fix: password-reset route not working correctly

### LINKED ISSUE #5726: [v2.7.0] Link to reset password receive through reset-password email does not work
## Prerequisites

I have checked code, changelog etc. but I have not seen any mention of reset-password url being removed. 

## Issue Description

Issue links: #5247 

When user forgets password and completes form regarding forgot password, he receives email which links to page where password can be change.
Example of link: http://localhost:3000/reset-password/08udZOQl5gWJMg44DtS2BfT9cONg5Fk7hpf7guJ0wHd

When going on this link user should see a form where he can change his password, but now he goes back to sign in page.

## Steps to Reproduce
1. Have running reaction.
2. Go to: http://localhost:3000/reset-password/08udZOQl5gWJMg44DtS2BfT9cONg5Fk7hpf7guJ0wHd
3. Expect password change form, but receive regular login form.

### Versions
2.7.0

kieckhafer:
Resolves #5726 
Impact: **critical**  
Type: **bugfix**

## Issue
The reset-password route does not correctly render the right component, making it impossible for you to reset your password unless you are under certain circumstances, like an admin who is logged in.

In addition, there was an issue related to this Password reset being a GQL mutation. On this page, there is no user, so `meteor-login-token` is not set. However, we were setting that value to `null`, instead of leaving it undefined. GQL was trying to use `null` as an auth check, which was throwing `Unauthorized` errors.

By leaving the `meteor-login-token` as undefined, we are able to use this mutation without being logged in / `null`.

## Solution
- Add a URL check to see if the route is `reset-password`, and show the form if it is.
- Add a check to make sure `meteor-login-token` is a value or `undefined`, but not `null`

## Notes
This does not fix the underlying problem, which is that the `reset-password` route doesn't register as it should and instead says `not-found` for the `name` of the route. It provides a temporary fix to the issue we are seeing until we can move authentication to it's own app.

## Breaking changes
None.

## Testing
1. Register for an account
1. Use the password reset email to reset your password
1. See that you land on the password reset page after clicking the link in the email

### DIFF
--- a/imports/plugins/core/accounts/client/containers/updatePassword.js
+++ b/imports/plugins/core/accounts/client/containers/updatePassword.js
@@ -64,7 +64,10 @@ const wrapComponent = (Comp) => (
         });
         return;
       }
-      const { token } = Router.current().params;
+
+      const { path } = Router.current().route;
+      const token = path.replace("/reset-password/", "");
+
       Accounts.resetPassword(token, password, (error) => {
         if (error) {
           this.setState({
@@ -83,7 +86,7 @@ const wrapComponent = (Comp) => (
           } else if (!storefrontUrls || !storefrontUrls.storefrontLoginUrl) {
             throw new ReactionError("error-occurred", "Missing storefront URLs. Please set these properties from the shop settings panel.");
           } else {
-            window.location.href = storefrontUrls.storefrontLoginUrl;
+            Router.go(storefrontUrls.storefrontLoginUrl);
           }
         }
       });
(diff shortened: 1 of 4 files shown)


kieckhafer:
@mikemurray can you recheck this please?

imports/plugins/core/accounts/client/containers/updatePassword.js
```diff
@@ -83,7 +86,7 @@ const wrapComponent = (Comp) => (
           } else if (!storefrontUrls || !storefrontUrls.storefrontLoginUrl) {
             throw new ReactionError("error-occurred", "Missing storefront URLs. Please set these properties from the shop settings panel.");
           } else {
-            window.location.href = storefrontUrls.storefrontLoginUrl;
+            Router.go(storefrontUrls.storefrontLoginUrl);
```
mikemurray: This should still probably be `window.location.href` you're redirecting to a different site. Router.go is really only for push state within this app context.

PR #5765, opened 2019-10-29
Constant condition bugfix

preminfy:
Resolves Constant condition issue
Impact: **minor**  
Type: **bugfix**

## Issue
Condition 'keywords' is always true because it evaluates to an array. Better to check 'keywords.length' instead.

lib/api/router/metadata.js

## Solution
Changed to use keywords.length
Code snippet:
Reaction.DOM.setMetaTag({
      name: "keywords",
      content: keywords.length ? keywords.toString() : ""
    });


aldeed:
I have cherry picked the fix into `develop` branch. Please do fixes from `develop` branch in the future, but thank you for contributing.

PR #5771, opened 2019-10-30
refactor: source Stripe API keys from ENV vars

### LINKED ISSUE #5767: [GQL] Stripe settings (included/payments-stripe)
This plugin modifies the package settings for the Stripe plugin to set the API keys.

It has now been decided to move these settings out of Collection and into ENV  vars.

This can likely be handled by a single mutation either `package/update`, or `registry/update` as outlined in the https://github.com/reactioncommerce/reaction/issues/5750 issue.

### Collections

- [ ] Packages

### Autoform Fields

- [ ] settings.api_key
- [ ] settings.public.publishable_key

willopez:
Resolves #5767  
Impact: minor
Type: refactor

## Issue
The Stripe API key needs to be sourced from environment variables for security considerations.

## Solution
Source the secret Stripe API key from environment variables. Some settings were removed as they are no longer needed. 
Removed:
- publishable_key: only used by client to create Stripe tokens. [details](https://stripe.com/docs/keys)
- client_id - used by client

## Why
- In order to move towards PCI compliance
- Secrets should be stored in a secured vault, moving to ENV vars facilitates using a secrets provider
- Remove dependency on `Packages` collection as it's not needed and we are working on removing it


willopez:
Note: integration tests pass locally, they fail in Circle CI due to a known issue with in-memory Mongo, that should hopefully be fixed soon.

willopez:
@mikemurray @aldeed I have verified that I amble to checkout on this branch, so we can move forward and merge.

willopez:
@mikemurray good point, I have added a placeholder key to the `example.env` that will prevent the app from crashing.

src/plugins/payments-stripe/util/getStripeInstanceForShop.js
```diff
@@ -13,8 +14,8 @@ export default async function getStripeInstanceForShop(context, shopId) {
   const stripePkg = await getStripePackageForShop(context, shopId);
   const stripePkgSettings = (stripePkg || {}).settings || {};
 
-  const stripeApiKey = stripePkgSettings.api_key;
-  if (!stripeApiKey) {
+  const { STRIPE_API_KEY } = config;
+  if (!STRIPE_API_KEY) {
     const stripeAccessToken = (stripePkgSettings.connectAuth || {}).access_token;
```
aldeed: This part looking for `connectAuth` and the `if (stripeAccessToken)` block that follows should be removed, too. You could really remove this whole `if (!STRIPE_API_KEY)` block now because envalid is checking the ENV for you.

aldeed: That will also mean that you can avoid calling `getStripePackageForShop` above.

willopez:
@mikemurray @aldeed Removed unnecessary code and re-tested with storefront and can confirm checkout with Stripe works.

PR #5772, opened 2019-10-31
refactor: rename `simple-authorization` to `legacy-authorization`, change function footprint

kieckhafer:
Impact: **minor**  
Type: **feature|refactor**

As an interim step in our transition from our current authorization system to our new Keto based system, we have changed the function signature for our permission checks.

The new signiture looks like this, with `legacyRoles` being the array of roles previously sent into the legacy auth check:
`validatePermissions("resource", "action", { shopId, legacyRoles: [], ...}`

## Testing
You should see no changes in how the app functions with this PR. The legacy auth service `hasPermission` check has been updated to use the new function signature.

## Breaking Changes
Nothing for core code / Reaction 3.0, but this will break and third party / plugin auth checks.


src/core-services/account/mutations/addAccountEmailRecord.js
```diff
@@ -32,8 +32,12 @@ export default async function addAccountEmailRecord(context, input) {
   const user = await users.findOne({ _id: account.userId });
   if (!user) throw new ReactionError("not-found", "User not Found");
 
-  if (!context.isInternalCall && userIdFromContext !== account.userId) {
-    await checkPermissions(["reaction-accounts"], account.shopId);
+  if (!context.isInternalCall) {
+    await context.validatePermissions(`reaction:accounts:${account._id}`, "add:emails", {
+      shopId: account.shopId,
+      owner: account._id,
```
aldeed: @rosshadden @kieckhafer All `validatePermissions` calls should be using `account.userId` rather than `account._id` everywhere. We currently set them to the same string, but it's not a guarantee if people import them or something.

rosshadden: Understood, thank you!

aldeed: Referring to `owner` value. In this case `reaction:accounts:${account._id}` probably is correct because that's the account you are accessing.

rosshadden: What about `context.accountId`? We have a lot of properties that are (currently) the same thing.

kieckhafer:
@rosshadden I believe this is ready to go on my end. Have you updated the values @aldeed mentioned?

kieckhafer:
Failing tests ticket https://github.com/reactioncommerce/reaction/issues/5861

PR #5776, opened 2019-11-01
Revert "refactor: only update fulfillment group for which id is provided"

machikoyasuda:
Reverts reactioncommerce/reaction#5723


machikoyasuda:
Hi @rosshadden @bt3gl @spencern - I want to revert this PR because the integration spec is failing and causing the resulting build's integration tests to time out.

PR #5780, opened 2019-11-04
fix: integration test in circle ci

### LINKED ISSUE #5779: [2.9.0] Integration tests take 5+ minutes on CI
See current tests stats on CircleCI: `6 minutes 42 seconds`
- integration test step: https://circleci.com/gh/reactioncommerce/reaction/60378
- Full logs: https://circleci.com/api/v1.1/project/github/reactioncommerce/reaction/60378/output/103/0?file=true

```
Test Suites: 40 passed, 40 total
Tests:       118 passed, 118 total
Snapshots:   7 passed, 7 total
Time:        402.101s
```

YuuwakU:
Resolves #5779
Impact: **minor**  
Type: **performance|test**

## Issue
maxWorkers for jest integration test was set to 3 which increased the CPU and RAM consumption of CI container

## Solution
bring back --runInBand and reduce node memory limit for jest integration script to 4000 MB so that there some memory for other processes.
Refer:
- https://github.com/integrations/slack/pull/895/files
Note: Although in the reference, global.gc is invoked before each test, it wasn't required here.
- https://developer.ibm.com/articles/nodejs-memory-management-in-container-environments/

## Breaking changes
none

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ defaults: &defaults
     - DOCKER_NAMESPACE: "reactioncommerce"
     - DOCKER_NAME: "reaction"
     - GLOBAL_CACHE_VERSION: "v2"
-    - TOOL_NODE_FLAGS: "--max-old-space-size=4096"
+    - TOOL_NODE_FLAGS: "--max-old-space-size=4000"
   working_directory: ~/reaction-app
   docker:
     - image: circleci/node:8-stretch
@@ -96,7 +96,7 @@ jobs:
           name: Docker build
           command: |
             docker build \
-              --build-arg TOOL_NODE_FLAGS="--max-old-space-size=4096" \
+              --build-arg TOOL_NODE_FLAGS="--max-old-space-size=4000" \
               -t "$DOCKER_REPOSITORY:$CIRCLE_SHA1" .
             mkdir -p docker-cache
             docker save \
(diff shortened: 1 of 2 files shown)


machikoyasuda:
@rosshadden What do you think? 

I ran it locally on OS X and my stats were:

```
Test Suites: 40 passed, 40 total
Tests:       118 passed, 118 total
Snapshots:   7 passed, 7 total
Time:        308.429s
```

But it's much faster in CI. How about your machine?

YuuwakU:
> @rosshadden What do you think?
> 
> I ran it locally on OS X and my stats were:
> 
> ```
> Test Suites: 40 passed, 40 total
> Tests:       118 passed, 118 total
> Snapshots:   7 passed, 7 total
> Time:        308.429s
> ```
> 
> But it's much faster in CI. How about your machine?

@machikoyasuda one of the main reasons this happens is due to the specification of the host CPU of the docker container. For eg, you would be running the tests on macbook pro which may have an i5 or i7 mobile processor while the CI instance has a XEON server processor, which has a massive difference in single core performance. This is important as we are using --runInBand which uses a single core of the CPU.

Test results on my machine take:
```
Test Suites: 40 passed, 40 total
Tests:       118 passed, 118 total
Snapshots:   7 passed, 7 total
Time:        255.031s
```
My laptop has the intel i7 9750H mobile processor.

YuuwakU:
@rosshadden with `--maxWorkers` the tests take more time to complete and even cross the 5000 ms default jest timeout causing tests to fail locally. To not let this happen most of the tests have `jest.setTimeout(30000)` but there are some tests that don't have this setting. If you could share the error information of the tests failing while using  `--runInBand`.

aldeed:
I've been out, but to add a little context:

The `--runInBand` is necessary to get the full benefits of the `@shelf/jest-mongodb` package. It creates the MongoDB server in Jest setup code and destroys it in Jest teardown code. If you run tests in parallel, then it will do this in every process and tests will be much slower.

I'm not sure if any automatic data destruction happens between tests. I don't think any tests are creating many documents without removing them after, but that could be one thing to check, especially if it keeps some of the mongo data in memory.

So this PR looks fine to me

FYI @spencern

PR #5781, opened 2019-11-05
adding sidebar permission control

Zachary-work:
Impact: **major**  
Type: **feature**

## Issue
RegisterOperatorRoute will not be able to register a route with access control, we want to add a option for `registerOperatorRoute ` to support it.

## Solution
Adding a `permissionRequired` option in `registerOperatorRoute` to check if the current user have one of the permissions provided in `permissionRequired` array

## Example
```sh
registerOperatorRoute({
    permissionRequired: ['companies/list']
    isNavigationLink: true,
    isSetting: false,
    path: "/partners",
    layoutComponent: ContentViewExtraWideLayout,
    mainComponent: PartnerList,
    SidebarIconComponent: (props) => <BriefcaseIcon {...props} />,
    sidebarI18nLabel: "admin.dashboard.partnerLabel"
})
```

### DIFF
--- a/imports/client/ui/components/Sidebar/Sidebar.js
+++ b/imports/client/ui/components/Sidebar/Sidebar.js
@@ -18,6 +18,7 @@ import SettingsIcon from "mdi-material-ui/Settings";
 import CloseIcon from "mdi-material-ui/Close";
 import ShopLogoWithData from "../ShopLogoWithData";
 import { Translation } from "/imports/plugins/core/ui/client/components";
+import { Reaction } from "/client/api";
 
 const activeClassName = "nav-item-active";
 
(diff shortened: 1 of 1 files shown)


aldeed:
Thanks @Zachary-work. In general the plan for Reaction 3.0 and beyond is that hiding or showing UI areas will be based on a separate setting/permission rather than tied to API permissions. However, it may be OK to merge this in the 2.x stream only as a stopgap solution. @mikemurray , thoughts?

(I am changing this to target `develop`. In the future please branch from `develop` and select `develop` as the PR base for work like this.)

aldeed:
There are also some eslint errors that you will need to resolve before this can be merged: https://app.circleci.com/jobs/github/reactioncommerce/reaction/60460

aldeed:
@Zachary-work Thanks but we have decided to not merge this. 2.x is now getting fixes only, but some sort of UI hiding will be added to the new separate [reaction-admin](https://github.com/reactioncommerce/reaction-admin) project for 3.x

PR #5785, opened 2019-11-06
Add csv file type validation for all platforms

trojanh:
Resolves N.A.
Impact: **minor**  
Type: **bugfix**

## Issue
Products filter by .csv file doesn't have file validation which makes any type of files to  be uploaded causing the following error on browser console:
Screenshot 2019-11-06 at 5 43 40 PM 

## Solution
Adding following MIME type for csv validation works on all platforms
```
text/plain
text/x-csv
application/vnd.ms-excel
application/csv
application/x-csv
text/csv
text/comma-separated-values
text/x-comma-separated-values
text/tab-separated-values
```
ref: https://www.christianwood.net/csv-file-upload-validation/

## Testing
1. Go to products page 
2. Click on `Actions` dropdown and now click on `Filter by file` option
3. Upload non-csv file and see the browser console after clicking on `Filter Products`

### DIFF
--- a/imports/plugins/included/product-admin/client/components/ProductTable.js
+++ b/imports/plugins/included/product-admin/client/components/ProductTable.js
@@ -25,6 +25,18 @@ mutation createProduct($input: CreateProductInput!) {
 }
 `;
 
+const CSV_FILE_TYPES = [
+  "text/csv",
+  "text/plain",
+  "text/x-csv",
+  "application/vnd.ms-excel",
+  "application/csv",
+  "application/x-csv",
+  "text/comma-separated-values",
+  "text/x-comma-separated-values",
+  "text/tab-separated-values"
+];
+
 /**
  * ProductTable component
  * @param {Object} props Component props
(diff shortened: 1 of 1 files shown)


mikemurray:
Testing must occur on `Windows` before this PR is to be merged. We removed the validation because of an issue with windows. I hope that these new mime types do work, but again, it needs to be tested on `Windows`, not just on macOS or Linux.

spencern:
To add on to what @mikemurray mentioned here - we've removed the MIME type validation because of an issue that was reported using Windows with the previous type validation we had 

Your solution is more substantial than the initial `text/csv` type that was built in - I'm hopeful that this issue is still resolved with this PR. https://github.com/reactioncommerce/reaction/pull/5700

We need someone to test this PR on several browsers on Windows before merging this.

trojanh:
Thanks @spencern and @mikemurray, I couldn't test it too but I thought this solution will help achieve the wider goal of allowing variety of file types on different platforms. So let's wait till we have this tested by someone on different platforms.

aldeed:
Verified works with Excel CSVs and on Windows OS. I implemented the same changes in Reaction Admin 3.0 here: https://github.com/reactioncommerce/reaction-admin/pull/197

Note that 2.x is in fixes-only phase, so this is getting merged but will only get released if another release is deemed necessary at some point.

trojanh:
awesome,  thanks @aldeed , 👍

PR #5788, opened 2019-11-06
refactor: source SMTP email configuation from ENV vars instead of DB

### LINKED ISSUE #5759: [GQL] Email SMTP (included/email-smtp)
Updated email settings in the `Packages` collection.

### Collections

- [ ] Packages

### Meteor dependencies

- [ ] meteor/check
- [ ] meteor/meteor

### Methods

- [ ] email/saveSettings

### Publications

none

### File structure

```
.
├── index.js
└── methods
    ├── index.js
    └── saveSettings.js
```

willopez:
Resolves #5759 
Impact: minor
Type: refactor

## Issue
Settings for SMTP email service were being saved to the DB in plain text which is not secure. This PR refactors sourcing SMTP email configuration from the `MAIL_URL` env var, this will also help us move closer to PCI compliance.

## Solution
Source SMTP email config from ENV var.

If you're solving a UIX related issue, please attach screen-caps or gifs showing how your solution differs from the issue.

**NOTE:**
I have created a new issue to remove the old UI used to set these settings in the Admin UI. https://github.com/reactioncommerce/reaction-admin/issues/84
## Breaking changes
NA

## Testing
1. Set the `MAIL_URL` to a desired value, i.e. `MAIL_URL=https://user:pass@example.com:465`
2. Restart the API
2. Use the `verifySMTPEmailSettings` to verify the settings are correct.
```
mutation verifySMTPEmailSettings($input: VerifySMTPEmailSettingsInput!) {
  verifySMTPEmailSettings(input: $input) {
    isVerified
  }
}
```
variables:
```
{
  "input": {
    "shopId": "cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg=="
  }
}
```


src/plugins/email-smtp/util/getMailConfig.js
```diff
@@ -40,76 +39,11 @@ function getConfigFromMailUrl({ logger, mailUrl }) {
...
  * @summary get the email sending config for Nodemailer
- * @param {Object} context App context
- * @param {String} shopId The ID of the shop for which to get email sending config
  * @returns {{host: String, port: Number, secure: Boolean, auth: Object, logger: Boolean}} returns a config object
  */
-export default async function getMailConfig(context, shopId) {
-  const { collections } = context;
-  const { Packages } = collections;
+export default async function getMailConfig() {
   const { MAIL_URL, EMAIL_DEBUG } = process.env;
```
aldeed: You should update these to come from an `envalid` config file for this plugin while you're in here. `EMAIL_DEBUG` can default to `false` and `MAIL_URL` can default to `""` to make it optional. Then I would remove it from the `.env.example` file. I'm trying to keep things out of that file if possible since it isn't portable when we move plugins to NPM packages.

src/plugins/email-smtp/util/getMailConfig.js
```diff
@@ -119,21 +53,7 @@ export default async function getMailConfig(context, shopId) {
...
-  if (service && service !== "custom") return getStandardServiceConfig({ logger, password, service, user });
-
-  /**
-   * Fall back to a custom service
-   */
-  if (host && port) return getCustomServiceConfig({ host, logger, password, port, user });
-
-  /**
-   * If all else fails, try direct sending but warn that this rarely works.
+   * Try direct sending but warn that this rarely works.
```
aldeed: In the other function, we just fail if `direct` is true, so this message isn't accurate. I think the best would be:

- Remove all of this after `if (MAIL_URL)` line and just throw an error instead.
- In `sendSMTPEmail.js`, add `try/catch` around `getMailConfig()` line, and call `sendEmailFailed` in the `catch` block. No need to check `config.direct` in that file.

src/plugins/email-smtp/util/getMailConfig.js
```diff
@@ -119,21 +53,7 @@ export default async function getMailConfig(context, shopId) {
   if (MAIL_URL) return getConfigFromMailUrl({ logger, mailUrl: MAIL_URL });
```
aldeed: As far as I can tell, `getConfigFromMailUrl` does not do anything with `logger` so I see no reason to pass it in. It can be added to the returned config object before returning it from this function.

Also, since ENV should not change while the server is running, `getConfigFromMailUrl` could be called just one time in top-level code, and then the config object simply returned by this function.

src/plugins/email-smtp/util/getMailConfig.js
willopez: validation added.

src/plugins/email-smtp/util/getMailConfig.js
willopez: refactored

src/plugins/email-smtp/util/getMailConfig.js
willopez: suggestions implemented.

willopez:
@mikemurray this is ready for another review. All integration tests pass locally.

aldeed:
@mikemurray I may have resolved those welcome email issues in my other PRs related to account invites.

willopez:
@aldeed all comments resolved

PR #5789, opened 2019-11-07
Use checkPermissions in getOrderQuery

trojanh:
Resolves N.A.
Impact: **minor**  
Type: **feature|refactor**

## Issue
`getOrderQuery` is still using `userHasPermission` method to check and throw access denied error explicitly. 

## Solution
Use `checkPermissions` method to avoid throwing an error explicitly.

## Testing
Check if `orderById` Graphql query works as expected.


trojanh:
closing it since `userPermission` still makes sense here and `checkPermission` will break for anonymous users

PR #5799, opened 2019-11-11
[3.0.0] Fix account invite flow and account email data

### LINKED ISSUE #5238: Cannot setup a new account for Shop Manager
I am not sure if that is a bug or not. But I am trying to setup a new account and I cannot finish the process successfully.

So, I am not sure what is the process but I can't finish this one.

## Prerequisites

## Issue Description
Providing a clear description of the issue will enable a developer to quickly reproduce and understand the issue. Please provide a detailed description including images, animation and/or videos. For client issues, include browser console errors, for server issues please add terminal output and stack traces.

## Steps to Reproduce
1. bring the platform up using the make command.
2. Login on http://localhost:3000/operator/accounts using default admin@localhost
3. Add name and email, choose Shop manager and click on send invitation
4. I receive the email with a button link to http://localhost:3000/account/enroll/A3p2cRuJsqkJjMyAQ
5. The link sends me to the login page (I was expecting something like set password)
6. Since I do not have login, I click on Register, fill the form and I get an error user already exist with the email.
6.1 If I try to login it say ```User has no password set```

## Possible Solution
If that is not a bug I would expect some clarification on how to do it.

### Versions
```
Node: 11.14.0
NPM: 6.9.0
Meteor Node: 8.15.1
Meteor NPM: 6.9.0
Reaction CLI: 0.29.0
Docker: 18.09.2
```

### LINKED ISSUE #5763: [3.0.0] Deal with context.getAbsoluteUrl calls in email code
There are context.getAbsoluteUrl calls in the email data that now need to point to the other server.

### LINKED ISSUE #5798: [3.0.0] Fix flow for inviting a new admin user
Inviting new shop manager was broken in 3.0.0 when Reaction Admin was split off. Fix it.

aldeed:
Resolves #5798 
Resolves #5763 
Resolves #5238   
Impact: **minor**  
Type: **bugfix**

## Issue
After Reaction Admin was split off from the API in `release-3.0.0` branch
- inviting new admin users no longer works
- some account emails have broken links because they point to resources that are now on a different server.

## Solution
- Creates a new AccountInvites collection to track invited email addresses.
- No longer immediately creates a user record for the invited address
- If an `Account` document is created for an invited email address, automatically add them to the invited permission groups rather than the default group ("customer"), and then delete the AccountInvites doc.

## Breaking changes
From the perspective of the involved users, the flow for inviting admins and accepting the invitations remains the same. The user and account changes behind the scenes are different, so any custom plugins or services relying on the old way may need updates.

## Testing
Prep:
1. Create first admin user on empty db, using register form on Reaction Admin.
2. Create first shop using form shown on Reaction Admin after logging in.
3. Configure emailing for the shop in shop settings. Enter a shop email address and valid SMTP server info.

Invite New
1. On Accounts screen, click Manage Groups and then use the form to invite a new shop manager by email (real email address you can access).
1. Log out
1. Verify that you receive the invite email and it looks correct. (Some text may be missing if you didn't fill out all shop details yet.)
1. Click button in email
1. Verify you go to Reaction Admin URL. Click Register and register with the same email address that was invited.
1. Verify that you have proper access, correct permissions (`roles`) are on your new user doc, and your account has correct `groups` list. In Reaction Admin UI, you may have to refresh once or twice to see the admin UI. This is a separate issue that will be fixed later.

Invite Existing
1. Log out of Reaction Admin.
2. Register a new user with an email address that isn't in the system yet.
3. Verify you receive a welcome email that looks correct, and the new user should have "customer" group permissions.
4. Log out
5. Log in as the first owner user.
6. On Accounts screen, click Manage Groups and then use the form to invite the email address you just created an account for to become a shop manager.
7. Verify that the account now shows as a shop manager in the UI.
8. Log out, and back in with the just-promoted user. Verify you can do shop manager things.


aldeed:
@kieckhafer The groups array issue should be fixed in latest commit.

aldeed:
@kieckhafer If they are in different groups, then that sounds correct. Whatever permissions match the account group they are in would be on their `roles`. The `HK6E` ID in your screen shots is the shop ID so that remains the same.

Second user would be in "customer" group by default. If you then log in as the first user and promote user 2 to "owner" group, I would expect that the roles lists would then match for that shop ID.

kieckhafer:
Ahhh, got it, I'm so used to the old `J8...` shopId I thought this was something different.

aldeed:
@kieckhafer I *think* that is correct. It is working as programmed. The problem is that when you first register, we don't actually know which shop to give you `customer` permissions for. Right now it does no shops. It could instead put you in the `customer` group for EVERY shop. Or for primary shop. This needs more thought, but bottom line is that the `customer` group does not currently really give you any powers, so it's effectively the same to be in no group at all. I suggest you address this properly while working on the new auth system but for now it does not seem like a blocking issue.

PR #5812, opened 2019-11-19
(fix): fix security vulnerabilities

quophyie:
fix security vulnerabilties on depencencies


spencern:
@quophyie I believe this has already been resolved in https://github.com/reactioncommerce/reaction/pull/5810

quophyie:
@spencern ... That is right... @mpaktiti had asked me to look at it in her handing notes... Any this is fixed in #5810 so will close

PR #5818, opened 2019-11-20
feat: GQL products & product query

### LINKED ISSUE #5817: [GQL] Products & Product queries
Add GQL queries for a list of `Products` and for a single `Product`.

mikemurray:
Resolves #5817  
Impact: **major**  
Type: **feature**

## Issue

No queries for `products` or `product`

## Solution

Add queries for  `products` or `product`

## Breaking changes

none

## Testing
1. Query for products with various filters
2. Query for a single product

```graphql
# List of products
query Products(
  $shopIds: [ID]
  $first: ConnectionLimitInt,
  $last:  ConnectionLimitInt,
  $before: ConnectionCursor,
  $after: ConnectionCursor,
  $offset: Int
) {
  products(
    shopIds: $shopIds
    first: $first
    last: $last
    before: $before
    after: $after
    offset: $offset
  ) {
    nodes {
      _id
      title
      shop {
        _id
      }
      variants {
        _id
        title
        options {
          _id
          title
        }
      }
    }
  }
}

# Single product
query Product(
  $productId: ID!
  $shopId: ID!
) {
  product(
    productId: $productId
    shopId: $shopId
  ) {
    _id
    title
    variants {
      _id
      title
      options {
        _id
        title
      }
    }
  }
}

```

### DIFF
--- a/src/core-services/product/queries/product.js
+++ b/src/core-services/product/queries/product.js
@@ -4,7 +4,9 @@
  * @memberof GraphQL/Product
  * @summary Query the Products collection for a single product
  * @param {Object} context - an object containing the per-request state
- * @param {Object} input - ID of Shop to query against
+ * @param {Object} input - Request input
+ * @param {String} input.productId - Product ID
+ * @param {String} input.shopId - Shop ID
  * @returns {Promise<Object>} Product object Promise
  */
 export default async function product(context, input) {
--- a/src/core-services/product/queries/product.js
+++ b/src/core-services/product/queries/product.js
@@ -1,0 +1,21 @@
+/**
+ * @name product
+ * @method
+ * @memberof GraphQL/Product
+ * @summary Query the Products collection for a single product
+ * @param {Object} context - an object containing the per-request state
+ * @param {Object} input - ID of Shop to query against
+ * @returns {Promise<Object>} Product object Promise
+ */
+export default async function product(context, input) {
+  const { checkPermissions, collections } = context;
+  const { Products } = collections;
+  const { productId, shopId } = input;
+
+  checkPermissions(["owner", "admin", "createProduct"], shopId);
+
+  return Products.findOne({
+    _id: productId,
+    shopId
+  });
+}
--- a/src/core-services/product/queries/product.js
+++ b/src/core-services/product/queries/product.js
@@ -14,7 +14,7 @@ export default async function product(context, input) {
   const { Products } = collections;
   const { productId, shopId } = input;
 
-  checkPermissions(["owner", "admin", "createProduct"], shopId);
+  await checkPermissions(["owner", "admin", "createProduct"], shopId);
 
   return Products.findOne({
     _id: productId,
--- a/src/core-services/product/queries/products.js
+++ b/src/core-services/product/queries/products.js
@@ -4,7 +4,7 @@ import applyProductFilters from "../utils/applyProductFilters.js";
  * @name products
  * @method
  * @memberof GraphQL/Products
- * @summary Query the Products collection
+ * @summary Query the Products collection for a list of products
  * @param {Object} context - an object containing the per-request state
  * @param {Object} input - ID of Shop to query against
  * @returns {Promise<Object>} Products object Promise
--- a/src/core-services/product/queries/products.js
+++ b/src/core-services/product/queries/products.js
@@ -13,10 +13,10 @@ import applyProductFilters from "../utils/applyProductFilters.js";
  * @param {String} [metafieldValue] - Filter by metafield value
  * @param {Number} [priceMax] - Filter by price range maximum value
  * @param {Number} [priceMin] - Filter by price range minimum value
- * @param {String[]} [productIds] - List of product IDs to filter
+ * @param {String[]} [productIds] - List of product IDs to filter by
  * @param {String} [query] - Regex match query string
  * @param {String[]} shopIds - List of shop IDs to filter by
- * @param {String[]} [tagIds] - List of tag ids to filter
+ * @param {String[]} [tagIds] - List of tag ids to filter by
  * @returns {Promise<Object>} Products object Promise
  */
 export default async function products(context, input) {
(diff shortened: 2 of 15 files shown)


src/core-services/product/queries/product.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * @name product
+ * @method
+ * @memberof GraphQL/Product
+ * @summary Query the Products collection for a single product
+ * @param {Object} context - an object containing the per-request state
+ * @param {Object} input - ID of Shop to query against
```
kieckhafer: Update comment with all input items

src/core-services/product/queries/products.js
```diff
@@ -0,0 +1,27 @@
+import applyProductFilters from "../utils/applyProductFilters.js";
+
+/**
+ * @name products
+ * @method
+ * @memberof GraphQL/Products
+ * @summary Query the Products collection for a list of products
+ * @param {Object} context - an object containing the per-request state
+ * @param {Object} input - ID of Shop to query against
```
kieckhafer: update comment with relevant description

src/core-services/product/queries/products.js
```diff
@@ -0,0 +1,27 @@
...
+ * @returns {Promise<Object>} Products object Promise
+ */
+export default async function products(context, input) {
+  const { checkPermissions, collections } = context;
+  const { Products } = collections;
+  const { ...productFilters } = input;
+
+  // Check the permissions for all shop requested
+  await Promise.all(productFilters.shopIds.map(async (shopId) => (
```
kieckhafer: Nothing for you to do here, just making a note for myself to update this in the auth pod once this is merged.

src/core-services/product/utils/applyProductFilters.js
```diff
@@ -0,0 +1,190 @@
...
+ * @param {Object} productFilters - See filters schema above
+ * @returns {Object} Selector
+ */
+export default function applyProductFilters(context, productFilters) {
+  // if there are filter/params that don't match the schema
+  filters.validate(productFilters);
+
+  // Init default selector - Everyone can see products that fit this selector
+  let selector = {
+    ancestors: [], // Lookup top-level products
```
kieckhafer: Nothing for you to do, just making a note for myself to possibly update this in the auth updates with the new style

src/core-services/product/queries/product.js
mikemurray: updated

src/core-services/product/queries/products.js
mikemurray: updated

src/core-services/product/schemas/product.graphql
```diff
@@ -30,6 +30,9 @@ type Product {
   "Subtitle"
   pageTitle: String
 
+  "Price range"
+  price: ProductPriceRange
+
```
aldeed: This stuff should be added by extending in `src/plugins/simple-pricing/schemas/schema.graphql` rather than here. Same way it's done for catalog.

src/core-services/product/resolvers/Query/products.js
```diff
@@ -0,0 +1,54 @@
...
+    isArchived,
+    isVisible,
+    metafieldKey,
+    metafieldValue,
+    priceMin,
+    priceMax,
+    ...connectionArgs
+  } = args;
+
+  const shopIds = opaqueShopIds.map(decodeShopOpaqueId);
```
aldeed: I think the assumption here is that `shopIds` is required and must always have at least one shop ID. That would make sense because if you don't filter on shopId, they could view products for any shop, and the permission checks in the query function would be skipped. However, there is no `!` on `shopIds` field in the GraphQL schema, which means `opaqueShopIds` might be undefined/null here.

I think probably adding `!` is the best fix, as well as maybe a check in the query function to be sure there's at least one item in the array (or generally to ensure they can't access any product just by passing an empty array.

src/core-services/product/schemas/product.graphql
mikemurray: Is the standard `price` object even still used on products or is it now `pricing`?

src/core-services/product/queries/products.js
```diff
@@ -0,0 +1,37 @@
...
+ * @returns {Promise<Object>} Products object Promise
+ */
+export default async function products(context, input) {
+  const { checkPermissions, collections } = context;
+  const { Products } = collections;
+  const productFilters = input;
+
+  // Check the permissions for all shop requested
+  await Promise.all(productFilters.shopIds.map(async (shopId) => (
+    checkPermissions(["owner", "admin", "createProduct"], shopId)
```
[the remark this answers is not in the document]
kieckhafer: We already have `product/admin` if you want to use that, otherwise creating `product/read` works.

That being said, these will all be different very soon anyway so it doesn't _really_ matter, as it'll all be refactored in https://github.com/reactioncommerce/reaction/pull/5772, which we hope to have merged into `release-3.0.0` soon.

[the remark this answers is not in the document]
mikemurray: Then let's leave it alone as the check is going to change drastically anyway.

src/core-services/product/resolvers/Query/products.js
mikemurray: Updated the schema and made the `shopIds` required

src/core-services/product/schemas/product.graphql
mikemurray: Moved to `simple-pricing` schema

src/core-services/product/queries/products.js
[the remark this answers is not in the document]
spencern: @kieckhafer @rosshadden @aldeed @ticean 
This is an important conversation to have, re: standards and policies for permission checks, I agree with @mikemurray that if we're about to refactor this all anyway, this ticket may not be the best place or time to have this conversation.

Do we have any documentation started for how our team should think about adding permision checks after #5772? What about for other services?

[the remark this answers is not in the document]
kieckhafer: Working on documentation in the (upcoming) authorization plugin. We do outline the standards there, we're just in a weird transition of the `release-3.0.0` branch being worked on while we are refactoring all of this. It shouldn't be an issue in a few days / weeks once our auth branch is merged, at that time we'll have the new standards all laid out and all over the code.

PR #5821, opened 2019-11-21
refactor: update "Base Unit of Measure" to "Base Unit of Weight"

### LINKED ISSUE #3458: Change "Base Unit of Measure" to "Base Unit of Weight"
### Expected behavior
The weight UOM should be called "Base Unit of Weight" similar to how Unit of Length is called "Base Unit of Length".

### Actual behavior
It's called "Base Unit of Measure" which is probably a legacy value from when we only cared about weight.

1. Login as an admin
1. Observe that the UOM for weight is called "Base Unit of Measure"
1. Despair

Note: Changing this for real involves changing the unlying schema and writing migrations for that

### Versions
```
Node: 8.9.1
NPM: 5.5.1
Meteor Node: 8.9.3
Meteor NPM: 5.5.1
Reaction CLI: 0.24.2
Reaction: 1.6.3
Reaction branch: master
Docker: 17.12.0-ce
```

trojanh:
Resolves #3458 
Impact: **minor**  
Type: **refactor**

## Issue
The weight UOM should be called "Base Unit of Weight" similar to how the Unit of Length is called "Base Unit of Length".

## Testing
1. Test createShop mutation


aldeed:
@trojanh We're going to close this pending further discussion on what needs to be done and how. Thanks for attempting it, though.

PR #5835, opened 2019-11-28
add graphql equivalent of setUserPermissions

### LINKED ISSUE #5829: Add GraphQL equivalent of method accounts/setUserPermissions
Please add GraphQL equivalent of https://github.com/reactioncommerce/reaction-admin/blob/trunk/imports/plugins/core/accounts/server/methods/setUserPermissions.js . This will help our demeteorization efforts and also help transition to reaction-authorization service

quophyie:
Resolves: #5829   
Impact: **major**  
Type: **feature**

## Issue
 
We need to add a GraphQL equivalent of the `setUserPermissions` found in `reaction-admin` at  https://github.com/reactioncommerce/reaction-admin/blob/trunk/imports/plugins/core/accounts/server/methods/setUserPermissions.js into reaction. This will allow for adding replacing permissions (_which is equivalent of `policies` in `reaction-authorization`. A point of explanation: At the moment, in `reaction-authorization` we dont add policies per user. We apply policies to roles which users are members of_)  to an account in reaction v3.0.0. This will also help our demeteorization efforts and also help transition to reaction-authorization service. 

## Solution

- Add GraphQL equivalent of `setUserPermissions` found in `reaction-admin` to `reaction` (resolver function and mutation functions)
- Add integration and unit tests for the newly created GQL `setUserPermissions`

## Breaking changes

none

## Testing

1. Tests should pass on CI
 - `tests/integration/api/mutations/setUserPermissions`
3. unit tests are found at 
 - `src/core-services/account/resolvers/Mutation/setUserPermissions.test.js`


src/core-services/account/mutations/addOrRemoveAccountGroups.js
```diff
@@ -0,0 +1,111 @@
...
+ * @memberof Mutations/Accounts
+ * @summary Adds or removes  groups on an account
+ * @param {Object} context - GraphQL execution context
+ * @param {Object} input - Necessary input for mutation. See SimpleSchema.
+ * @param {AddOrRemoveAccountGroupsOperationType} updateGroupOperationType - the type of operation to perform.
+ * @param {Object} input.groups - groups to append to
+ * @param {String} input.accountId - optional decoded ID of account on which entry should be updated, for admins
+ * @returns {Promise<Object>} with updated account
+ */
+export default async function addOrRemoveAccountGroups(context, input, updateGroupOperationType) {
```
mikemurray: While this is still a draft, the GQL mutations must be discrete. A mutation resolver & mutation function must exist for both `add` and `remove`. Don't mix actions, or make mutations to 2 different actions with a switch. Even if it is redundant, and it does duplicate code.

quophyie: @mikemurray  ... Thanks ... will do that

quophyie:
> Can you explain what this PR is doing, and why we need it?
> 
> You have #5839 and #5842 as well which are also both calling `setUserPermissions`. But #5839 cites this as a blocking PR. If you just read the names of those PRs it seems like you did what @mikemurray said, so should this be closed? If so it should be removed as a blocker as well.

@rosshadden ... these both  #5839 and #5835 do the same thing in terms of functionality but my understanding of why we have these two is that functionality  provided by #5839 (`addUserPermissions`) is an alias for `setUserPermissions` (i.e. this PR i.e. #5835)

quophyie:
Integration tests are working locally but are currently failing on CI because of heap out of memory issues. This is being addressed in #5888 (commit 36df464#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R101a) . I have added a dependency on #5888 because when that is merged, the integration tests here will also start working

src/core-services/account/mutations/addOrRemoveAccountGroups.js
quophyie: @mikemurray  .... Done

quophyie:
I need to fix git history

rosshadden:
This has 139 commits, many of which are from me, Eric, and Will. Please take a look at the commits.

src/core-services/account/mutations/setUserPermissions.js
```diff
@@ -0,0 +1,80 @@
...
+
+  const account = await Accounts.findOne({ _id: accountId });
+
+  if (!account) throw new ReactionError("not-found", "No account found");
+
+  if (!context.isInternalCall && userIdFromContext !== accountId) {
+    await context.validatePermissions("reaction:accounts", "update", { shopId: account.shopId, legacyRoles: ["reaction-accounts"] });
+  }
+
+  await context.validatePermissions("reaction:accounts", "update", { shopId, legacyRoles: ["admin"] });
```
rosshadden: Something is off here. This is saying "make sure the person is an admin" in terms of the legacy authorization. But the check above it is checking for a `reaction-accounts` role which implies this should not be admin-only.

@kieckhafer can you look at this and determine what these few lines should be changed to?

src/core-services/account/mutations/setUserPermissions.js
```diff
@@ -0,0 +1,80 @@
...
+  const { accountId } = context;
+  const { groups, shopId } = input;
+
+
+  const account = await Accounts.findOne({ _id: accountId });
+
+  if (!account) throw new ReactionError("not-found", "No account found");
+
+  if (!context.isInternalCall && userIdFromContext !== accountId) {
+    await context.validatePermissions("reaction:accounts", "update", { shopId: account.shopId, legacyRoles: ["reaction-accounts"] });
```
rosshadden: Why is this using `account.shopId` instead of the one being passed in? But the one below uses the shop id passed in.

src/core-services/account/mutations/setUserPermissions.js
```diff
@@ -0,0 +1,80 @@
...
+    },
+    {
+      $addToSet: {
+        groups: {
+          $each: groups
+        }
+      }
+    },
+    {
+      multi: true
```
rosshadden: It is impossible for this to affect multiple documents. It is finding on `_id` which is unique.

src/core-services/account/mutations/setUserPermissions.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+  );
+
+  if (!updatedAccount) {
+    throw new ReactionError("server-error", "Unable to update account groups");
```
rosshadden: Please make this more specific. Put yourself in the shoes of somebody reading this error in the logs. What does it mean? Think about under which conditions the error gets thrown, and make those more apparent in the error message.

src/core-services/account/schemas/account.graphql
```diff
@@ -136,6 +136,19 @@ input RemoveAccountEmailRecordInput {
   email: Email!
 }
 
+"Defines groups to add given user / account to thereby giving that user certain permissions"
+input SetUserPermissionsInput {
+
+  "An optional string identifying the mutation call, which will be returned in the response payload"
+  clientMutationId: String
```
rosshadden: This allows `clientMutationId` but nothing is done with it. It is in the response schema but never returned.

src/core-services/account/mutations/setUserPermissions.js
```diff
@@ -0,0 +1,80 @@
...
+  // This is used for search, to determine if we need to re-index
+  const updatedFields = ["groups"];
+
+  await appEvents.emit("afterAccountUpdate", {
+    account: updatedAccount,
+    updatedBy: userIdFromContext,
+    updatedFields
+  });
+
+  return { account: updatedAccount };
```
rosshadden: This should just return the updated account, not wrapped in an object.

src/core-services/account/schemas/account.graphql
```diff
@@ -136,6 +136,19 @@ input RemoveAccountEmailRecordInput {
...
+
+  "the groups to add a user / account to"
+  groups: [String!]!
```
rosshadden: I just want to clarify something about this syntax in case it doesn't do exactly what you expect.

This means:
- groups must be an array
- groups must be present
- groups must only contain strings, no nulls

This does _not_ mean:
- at least one group is required

That is, `groups: []` would be acceptable. If you were trying to require at least one element in the groups array that is outside the GraphQL scope and needs to be enforced by the resolver itself.

src/core-services/account/schemas/account.graphql
```diff
@@ -349,6 +362,14 @@ type SetAccountProfileLanguagePayload {
   clientMutationId: String
 }
 
+"The response from the `setUserPermissionsMutation` mutation"
+type SetUserPermissionPayload {
```
rosshadden: Right now the return of this corresponding resolver does not match this type.

src/core-services/account/resolvers/Mutation/setUserPermissions.js
```diff
@@ -0,0 +1,21 @@
...
+ * @param {Object} context - an object containing the per-request state
+ * @param {Object} context.userId - the userId of user to add to the given group
+ * @param {Object} context.accountId - the userId of user to add to the given group
+ * @returns {Object} - object
+ */
+export default async function setUserPermissions(_, { input }, context) {
+  const { shopId } = input;
+  const decodedShopId = decodeShopOpaqueId(shopId);
+  const transformedImput = { ...input, shopId: decodedShopId };
+  return context.mutations.setUserPermissions(context, transformedImput);
```
rosshadden: Doesn't match the GQL response schema.

src/core-services/account/schemas/account.graphql
mikemurray: `clientMutationId` is an `echo`, so it needs to be returned in the mutation resolver as is, along with the other results from the mutation.

src/core-services/account/mutations/setUserPermissions.js
```diff
@@ -0,0 +1,80 @@
...
+ * adds permissions to account (user)
+ * @param {Object} context - GraphQL execution context
+ * @param {Object} input - Necessary input for mutation. See SimpleSchema.
+ * @param {Object} input.groups - groups to append to
+ * @param {String} input.accountId - optional decoded ID of account on which entry should be updated, for admins
+ * @returns {Promise<Object>} with updated account
+ */
+export default async function setUserPermissions(context, input) {
+  const itemsToValidate = { accountId: context.accountId, userId: context.userId, groups: input.groups };
+  inputSchema.validate(itemsToValidate);
```
mikemurray: This should be validating the input object that's going to be used in the `update` operation. This is currently validating the existence of the `accountId` and `userId` of the currently signed-in user, not the user that you're trying to update the groups for.

src/core-services/account/mutations/setUserPermissions.js
```diff
@@ -0,0 +1,80 @@
...
+  const { appEvents, collections, userId: userIdFromContext } = context;
+  const { Accounts } = collections;
+  const { accountId } = context;
```
mikemurray: This `accountId` is for the currently signed-in user. Shouldn't this be passed as part of the input?

src/core-services/account/mutations/setUserPermissions.js
quophyie: @rosshadden ... I had a chat with @kieckhafer this afternoon and following that chat I have a better understanding of what should be done here. we only need to check for the `reaction-accounts` role so I will change that

src/core-services/account/schemas/account.graphql
quophyie: @mikemurray @rosshadden ... done

src/core-services/account/mutations/setUserPermissions.js
quophyie: @rosshadden ... done

src/core-services/account/schemas/account.graphql
quophyie: @rosshadden ... I've updated to fix this .. done

src/core-services/account/resolvers/Mutation/setUserPermissions.js
quophyie: @rosshadden ... updated to fix this ... done

src/core-services/account/schemas/account.graphql
quophyie: @rosshadden ... thanks for the explanation ... I've updated the mutation function validation to check that there is at least one group provided in this mutation

src/core-services/account/mutations/setUserPermissions.js
quophyie: @mikemurray ...  you are right ... this have been fixed ... thanks for catching that

src/core-services/account/mutations/setUserPermissions.js
quophyie: @mikemurray ... I've updated the mutation function to update the account of the `accountId` supplied in the input. I've also updated the validations to validate the input for the `accountId` and  `groups` supplied in the input

src/core-services/account/mutations/setUserPermissions.js
quophyie: @rosshadden removed ... done

src/core-services/account/mutations/setUserPermissions.js
quophyie: This is no longer used. I've removed it.
Done

src/core-services/account/mutations/setUserPermissions.js
quophyie: @rosshadden  ... Actually.. this is hang-over from the original mutation code where this was copied from which used `update`. In any case, I will remove this because the condition here will never be true as we do a check to find out whether the account exists and we throw if it doesnt on line (https://github.com/reactioncommerce/reaction/pull/5835/files#diff-efd372419698e0c60ced0867a59f788eR34)
So when we get here, the condition will always evaluate to false and so we will never throw

quophyie: @rosshadden ... I have updated the err message to  be more specific i.e. this only gets thrown if account is not found

src/core-services/account/mutations/setUserPermissions.js
quophyie: Done

quophyie:
as per @aldeed clarification, this meant to replace permission on an account and not add to it

@aldeed clarification
>They aren’t exactly the same @George. set overwrites the whole array of permissions to a new array you provide. add adds one permission to the array. remove removes one permission from the array. Looking at Reaction Admin, I don’t really think any of these are used by UI that is still visible. The current UI only uses updateGroup to set the permissions on the group, and then accounts can be added to different groups as necessary to inherit the permissions of that group.

mikemurray:
@aldeed @quophyie @rosshadden 

I'm want to clarify what we need from these permission functions as I see come to understand from the various conversations.

- Greate account groups
- Update account groups
- Assign an account to the group
- Remove an account from a group.

This means we need methods that are around account and groups. The permissions are assigned to a `user` when they are added/removed from a group. I think some checks like `hasPermission` still relies on that for legacy roles.

### Groups
- [x] Create account group `createAccountGroup` (we have this now)
  - Create an account group with permissions
- [ ] Update account group `updateAccountGroup` #5931
  - Updates an account group

### Account + Group
- [x] Add an account to group `addAccountToGroup` (exists)
  - This adds the `account` to a group and adds the permissions (roles) to the `user`
- [ ] Remove an account from group `removeAccountFromGroup`
  - Removes a group from an `account` and removes any roles from that group from the `users`'s roles array.
  - @aldeed I think we need this.

I think work on these PR's should cease for the moment:
- add graphql equivalent of setUserPermissions #5835
- Add graphql equivalent of addUserPermissions #5839
- Add graphql equivalent of remove user permissions #5842 

Work on these one should continue:
- Add graphql equivalent of updateGroup found in reaction-admin #5931 (this is `updateAccountGroup`)
- Then create a PR for a mutation to `removeAccountFromGroup` (assuming something like this is needed/doesn't already exist)

Did I miss anything?

quophyie:
@mikemurray @rosshadden ... we are going to close this PR following the following slack conversation 

>@aldeed wrote
You can only be in one group per shop @mikemurray  so addAccountToGroup already removes you from your previous group. changeGroupForAccount might be a better name for it

>@rosshadden wrote
That one-per-shop limitation is no longer the case. Accounts can now be in any number of groups. (edited) 

>@quophyie  wrote: 
@mikemurray  @rosshadden  @aldeed .. so just to  give some  more clarity about the limitation on the one user per shop, reaction-authorization does not have that limitation `one-group-user-per-shop` as  @rosshadden has already said, however, a client of `reaction-authorization` such as `reaction`  can enforce such a limitation on the client side if that client needs that  `one-group-per-user` constaint. This brings up some questions with  `reaction` i.e.

>1. Does `reaction`  (i.e. the reaction-authorization client) need to enforce `one-group-per-user` constraint? If it does, then we will need `addAccountToGroup` and `changeGroupForAccount`  but not `removeAccountFromGroups`   as that should not be possible. If this is true, then I'm not sure why we would need `updateAccountGroup`  and  what its purpose will be i.e. does it add or remove an account from a group? . What does `updateAccountGroup` mean in this context?

>2. If `reaction` does not need to enforce `one-group-per-user  constraint` (a user can be a member of an arbitrary num of groups), then we will need `addAccountToGroups` and `removeAccountFromGroups`  but not  `changeGroupForAccount` as a user group can easily be changed either by adding or removing them from groups. In this case I'm not sure what the purpose of `updateAccountGroup`  will be because I dont know what we mean by `updateAccountGroup`  i.e. does it mean add or remove from the account group?

>3. Finally,  currently  our permission validations (i.e. calls to `context.validatePermissions`  in `3.0.0`)in `reaction`  are not using `reaction-authorization` but the old style permission checks. This is temporary measure but it means that we have to depend  roles in reaction  for permission checks 
(edited)

>@quophyie wrote
So I guess the question I'm asking is does reaction need to enforce a `one-group-per-user`  constraint? If it does, what what are the mutations that we need to expose and if doesnt, what are the mutations we need to expose for that scenario.

>@rosshadden wrote
The groups only exist for permissions. Reaction itself has no constraints around what can and cannot be done with them. We do not need to enforce that constraint. It was just how it worked with the old meteor permissions system.

>@rosshadden wrote
So # 2 in your list is what we want.

>@quophyie wrote
Thanks @rosshadden. So given what you said, we need to implement `addAccountToGroups` and `removeAccountFromGroups`  mutations but not `changeGroupForAccount` and `updateAccountGroup` I need some clarity some clarity on this because we spent some time on the `set/add/remove`  and now realise we not need them

>@quophyie wrote
@rosshadden... gotcha

>@rosshadden wrote 
@kieckhafer  can you weigh in with a sanity check here? I know we'd all like to avoid another rework situation.
:eyes:



>@kieckhafer  wrote
I believe what you last mentioned @quophyie  is correct. We DO NOT need to have the constraint, therefore we DO NOT need the "single group" mutations, which would be change and update.  (although i feel like these were the wame anyway?).
I think that means we DO need add and remove

>@quophyie wrote
Given all that has been said,  should I go ahead and close https://github.com/reactioncommerce/reaction/pull/5839, https://github.com/reactioncommerce/reaction/pull/5835,https://github.com/reactioncommerce/reaction/pull/5931 and https://github.com/reactioncommerce/reaction/pull/5842
and no longer required. I will create a ticket for removeAccountFromGroups as `addAccountToGroups`  already exists (edited) 

>@mikemurray wrote
`updateAccountGroup`  is to update the Group itself not the group for an account. The name is makes it confusing
:point_up:



>@mikemurray wrote
right now we can create a groups with `createAccountGroup` but we cannot update it. That’s why we need `updateAccountGroup`
:point_up:



>@mikemurray wrote
This PR: https://github.com/reactioncommerce/reaction/pull/5931 should stay open and be the updateAccountGroup  mutation

>@mikemurray wrote
the other 3 can be closed:
https://github.com/reactioncommerce/reaction/pull/5839
https://github.com/reactioncommerce/reaction/pull/5835
https://github.com/reactioncommerce/reaction/pull/5842

>@mikemurray wrote
`updateAccountGrou`p and `removeAccountFromGroups` Do 2 actions
`updateAccountGroup`  Updates the group document, and updates all the permissions on all the users in those groups. (#5931) seems to do this now
`removeAccountFromGroups` Will remove the group from the account, and all the roles from the matching user record, IF and only IF they don’t overlap with another group on that account

>@quophyie wrote
@mikemurray... what would we be updating the group with? will we be updating it  with new roles etc? This is slightly confusing because updateAccountGroup will mean something totally different in reaction than it means in reaction-authorization. In `reaction` it means updating name, slug, permissions etc but in `reaction-authorization` it means either adding new members or removing them from the group (role in `reaction-authorization`)

>@rosshadden wrote
I don't think we should consider what it means in `reaction-authorization` to answer that question

>@quophyie wrote
Ok

>@rosshadden wrote
In terms of Reaction it means "edit name, slug, permissions, or shopId"

>@quophyie wrote
@rosshadden  @mikemurray  Ok ... gotcha ...  I can update (i.e. title branch name) https://github.com/reactioncommerce/reaction/pull/5842 to be  `removeAccountFromGroups` it will need a minimal amount of changes to do what @mikemurray  described instead of starting afresh... is  that gonna be a problem? (edited) 

>@rosshadden wrote
The authorization service actually only cares about _id which can't change. It creates and deletes roles when Groups are created and deleted. It doesn't even care about fields like permissions.
The groups are merely an organization of users. Their significance comes from their usage in policies. We define policies that say things like "reaction:groups:order-managers can delete orders". Any user given that role can then perform that action on orders.

>@rosshadden wrote
is that gonna be a problem?
Not at all! Do what you need to do

quophyie:
Closing this PR as its no longer required... See comment here https://github.com/reactioncommerce/reaction/pull/5835#issuecomment-567463371 for details

PR #5839, opened 2019-11-29
Add graphql equivalent of addUserPermissions

### LINKED ISSUE #5828: Add GraphQL Equivalent of method accounts/addUserPermissions
Please add the GraphQL equivalent of https://github.com/reactioncommerce/reaction-admin/blob/trunk/imports/plugins/core/accounts/server/methods/addUserPermissions.js in `reaction v3.x`

quophyie:
Resolves: #5828  
Impact: **major**  
Type: **feature**

## Issue
 
We need to add a GraphQL equivalent of the `addUserPermissions` found in `reaction-admin` at  https://github.com/reactioncommerce/reaction-admin/blob/trunk/imports/plugins/core/accounts/server/methods/addUserPermissions.js into reaction. This will allow for adding permissions (_which is equivalent of `policies` in `reaction-authorization`. A point of explanation: At the moment, in `reaction-authorization` we dont add policies per user. We apply policies to roles which users are members of_)  to an account in reaction v3.0.0. This will also help our demeteorization efforts and also help transition to reaction-authorization service. 

## Solution

- Add GraphQL equivalent of `addUserPermissions` found in `reaction-admin` to `reaction` (resolver function and mutation functions)
- Add integration and unit tests for the newly created GQL `addUserPermissions`

## Breaking changes

none

## Testing

1. Tests should pass on CI
 - **tests/integration/api/mutations/addUserPermissions/addUserPermissions.test.js**
3. unit tests are found at 
 - `src/core-services/account/resolvers/Mutation/addUserPermissions.test.js`

### DIFF
--- a/src/core-services/account/mutations/addUserPermissions.js
+++ b/src/core-services/account/mutations/addUserPermissions.js
@@ -51,6 +51,8 @@ export default async function addUserPermissions(context, input) {
           $each: groups
         }
       }
+    }, {
+      returnOriginal: false
     }
   );
 
--- a/src/core-services/account/mutations/addUserPermissions.js
+++ b/src/core-services/account/mutations/addUserPermissions.js
@@ -54,10 +54,6 @@ export default async function addUserPermissions(context, input) {
     }
   );
 
-  if (!updatedAccount) {
-    throw new ReactionError("server-error", "Unable to update account groups");
-  }
-
   // Create an array which contains all fields that have changed
   // This is used for search, to determine if we need to re-index
   const updatedFields = ["groups"];
--- a/src/core-services/account/mutations/addUserPermissions.js
+++ b/src/core-services/account/mutations/addUserPermissions.js
@@ -54,6 +54,10 @@ export default async function addUserPermissions(context, input) {
     }
   );
 
+  if (!updatedAccount) {
+    throw new ReactionError("server-error", "Unable to update account groups. Account not found");
+  }
+
   // Create an array which contains all fields that have changed
   // This is used for search, to determine if we need to re-index
   const updatedFields = ["groups"];
--- a/src/core-services/account/mutations/addUserPermissions.js
+++ b/src/core-services/account/mutations/addUserPermissions.js
@@ -1,0 +1,72 @@
+import SimpleSchema from "simpl-schema";
+import ReactionError from "@reactioncommerce/reaction-error";
+
+const inputSchema = new SimpleSchema({
+  "accountId": String,
+  "groups": {
+    type: Array, // groupIds that user belongs to
+    minCount: 1
+  },
+  "groups.$": {
+    type: String
+  }
+});
+
+/**
+ * @name accounts/addUserPermissions
+ * @memberof Mutations/Accounts
+ * @summary Adds a group to an account (user) group. This addition to an account  group effectively
+ * adds permissions to account (user)
+ * @param {Object} context - GraphQL execution context
+ * @param {Object} input - Necessary input for mutation. See SimpleSchema.
+ * @param {Object} input.groups - groups to append to
+ * @param {String} input.accountId - the decoded account ID of account on which entry should be updated
+ * @returns {Promise<Object>} with updated account
+ */
+export default async function addUserPermissions(context, input) {
+  const itemsToValidate = { accountId: input.accountId, groups: input.groups };
+  inputSchema.validate(itemsToValidate);
+  const { appEvents, collections, userId: userIdFromContext } = context;
+  const { Accounts } = collections;
+  const { groups, accountId } = input;
+
+
+  const account = await Accounts.findOne({ _id: accountId });
+
+  if (!account) throw new ReactionError("not-found", "No account found");
+
+  if (!context.isInternalCall) {
+    await context.validatePermissions("reaction:accounts", "update", { shopId: account.shopId, legacyRoles: ["reaction-accounts"] });
+  }
+
+  // Update the Reaction Accounts collection with new groups info
+  // This
+  const { value: updatedAccount } = await Accounts.findOneAndUpdate(
+    {
+      _id: accountId
+    },
+    {
+      $addToSet: {
+        groups: {
+          $each: groups
+        }
+      }
+    }
+  );
+
+  if (!updatedAccount) {
+    throw new ReactionError("server-error", "Unable to update account groups");
+  }
+
+  // Create an array which contains all fields that have changed
+  // This is used for search, to determine if we need to re-index
+  const updatedFields = ["groups"];
+
+  await appEvents.emit("afterAccountUpdate", {
+    account: updatedAccount,
+    updatedBy: userIdFromContext,
+    updatedFields
+  });
+
+  return updatedAccount;
+}
--- a/src/core-services/account/mutations/index.js
+++ b/src/core-services/account/mutations/index.js
@@ -12,6 +12,7 @@ import setAccountProfileCurrency from "./setAccountProfileCurrency.js";
 import setAccountProfileLanguage from "./setAccountProfileLanguage.js";
 import updateAccountAddressBookEntry from "./updateAccountAddressBookEntry.js";
 import createAccountGroup from "./createAccountGroup.js";
+import addUserPermissions from "./addUserPermissions.js";
 
 export default {
   addressBookAdd,
@@ -27,5 +28,6 @@ export default {
   setAccountProfileCurrency,
   setAccountProfileLanguage,
   updateAccountAddressBookEntry,
-  createAccountGroup
+  createAccountGroup,
+  addUserPermissions
 };
--- a/src/core-services/account/resolvers/Mutation/addUserPermissions.js
+++ b/src/core-services/account/resolvers/Mutation/addUserPermissions.js
@@ -1,0 +1,18 @@
+import setUserPermissions from "./setUserPermissions.js";
+
+/**
+ * @name Mutation/addUserPermissions
+ * @method
+ * @memberof Accounts/GraphQL
+ * @summary resolver to add user permissions. This is an alias to {@link setUserPermissions.js}
+ * @param {Object} _ - unused
+ * @param {Object} args.input - an object of all mutation arguments that were sent by the client
+ * @param {String} args.input.groups - The groups the user is to be added to
+ * @param {Object} context - an object containing the per-request state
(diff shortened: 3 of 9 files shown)


rosshadden:
I was trying to figure out why this has 51 commits, and I found that there are quite a lot of duplicates. For example "delete unused files" occurs eight times. Same commit message, same commit body, same deleted files, but different commit hashes. Looking at the rest of the commits it looks like this happened to some or all of the rest as well.

quophyie:
> I was trying to figure out why this has 51 commits, and I found that there are quite a lot of duplicates. For example "delete unused files" occurs eight times. Same commit message, same commit body, same deleted files, but different commit hashes. Looking at the rest of the commits it looks like this happened to some or all of the rest as well.

Caused by git incorrect. I've fixed the git history

rosshadden:
Why is this blocked by #5835?

quophyie:
> Why is this blocked by #5835?

@rosshadden ... this is blocked #5835 because the functionality which is provided by this PR is an alias for the functionality in #5835 according to my understanding  from @aldeed . @aldeed can correct me if I'm wrong. That is to say that `addUserPermissions` is an alias for `setUserPermissions`. Hence the mutations which this PR depends on has been done in `#5835` and therefore, we need to merge  #5835 before we can merge this

quophyie:
Integration tests are yet to be added. Also tests are failing because at the moment depends on functionality in #5835 which has not yet been merged. Once the changes are brought into this PR, the tests should pass

quophyie:
integration tests added

quophyie:
@rosshadden @aldeed @mikemurray ... There needs to be a discussion as to whether, this  needed / should be merged as #5835 does the exact same thing as this PR. Is this still required

quophyie:
As per @aldeed clarification, this is meant to add to permissions to a user (**_note: Permissions are added by add new groups to the account groups attribute on the account_**). This means that this is not an alias for  #5835 

@aldeed clarification
>They aren’t exactly the same @George. set overwrites the whole array of permissions to a new array you provide. add adds one permission to the array. remove removes one permission from the array. Looking at Reaction Admin, I don’t really think any of these are used by UI that is still visible. The current UI only uses updateGroup to set the permissions on the group, and then accounts can be added to different groups as necessary to inherit the permissions of that group.

quophyie:
We are going to close this PR following the slack conversation below

@quophyie wrote
>@rosshadden @kieckhafer  @mikemurray  @aldeed  ... So there is this PR i.e.  https://github.com/reactioncommerce/reaction/pull/5839 which does exactly the same thing as https://github.com/reactioncommerce/reaction/pull/5835. From my understanding 5835(`setUserPermissions`)  is the same as 5839 (`addUserPermissions`) . I think I remember @Eric Dobbertin saying (addUserPermissions)  is a nicer name for (setUserPermissions)  but I stand to be corrected. In any case, Now I'm just wondering whether they should both be supported? Can we get rid of either one and mark it as a breaking change in V3.0.0? ... @rosshadden , @kieckhafer and myself spoke about it in standup but we didnt make a decision on it yet (edited) 

@rosshadden wrote
>As I said in standup my understanding on all of these issues (5842, 5839, 5835) is fuzzy. Do we have services that we intend to call these mutations? What did they do before these changes? Or is making these into GraphQL endpoints just trying to plan for future situations that are not yet reality?

@mikemurray wrote
>These were used in the UI for admins to set the user permissions. It would most likely be used in the new admin if we ever get back to it

@rosshadden wrote
>So they were internal methods, but abstracting admin into its own project requires them to be exposed through GQL mutations?

@mikemurray wrote
>yea. And so we can say that, 3.0 is full GQL coverage of everything that used to exist in 2.0, even if the admin doesn’t support it yet, there should nothing stopping you from building your own UI (community)

@aldeed wrote
>They aren’t exactly the same @quophyie . set overwrites the whole array of permissions to a new array you provide. add adds one permission to the array. remove removes one permission from the array. Looking at Reaction Admin, I don’t really think any of these are used by UI that is still visible. The current UI only uses updateGroup to set the permissions on the group, and then accounts can be added to different groups as necessary to inherit the permissions of that group.

@aldeed wrote
>So as far as I’m concerned, we don’t need GQL for any of the set/add/remove, but I think someone decided we should create them anyway because we did have Meteor methods for them.

@quophyie wrote
>@aldeed ... thanks for the clarification

@mikemurray wrote
I added a comment on GH, posting it here so it’s not lost: https://github.com/reactioncommerce/reaction/pull/5835#issuecomment-567160365
I agree with @aldeed that we don’t need the set/add/remove methods, just updateAccountGroup and we should have a removeAccountFromGroup mutation (unless theres a way to remove/override a groups on an account already)

>@aldeed wrote
You can only be in one group per shop @mikemurray  so addAccountToGroup already removes you from your previous group. changeGroupForAccount might be a better name for it

>@rosshadden wrote
That one-per-shop limitation is no longer the case. Accounts can now be in any number of groups. (edited) 

>@quophyie  wrote: 
@mikemurray  @rosshadden  @aldeed .. so just to  give some  more clarity about the limitation on the one user per shop, reaction-authorization does not have that limitation `one-group-user-per-shop` as  @rosshadden has already said, however, a client of `reaction-authorization` such as `reaction`  can enforce such a limitation on the client side if that client needs that  `one-group-per-user` constaint. This brings up some questions with  `reaction` i.e.

>1. Does `reaction`  (i.e. the reaction-authorization client) need to enforce `one-group-per-user` constraint? If it does, then we will need `addAccountToGroup` and `changeGroupForAccount`  but not `removeAccountFromGroups`   as that should not be possible. If this is true, then I'm not sure why we would need `updateAccountGroup`  and  what its purpose will be i.e. does it add or remove an account from a group? . What does `updateAccountGroup` mean in this context?

>2. If `reaction` does not need to enforce `one-group-per-user  constraint` (a user can be a member of an arbitrary num of groups), then we will need `addAccountToGroups` and `removeAccountFromGroups`  but not  `changeGroupForAccount` as a user group can easily be changed either by adding or removing them from groups. In this case I'm not sure what the purpose of `updateAccountGroup`  will be because I dont know what we mean by `updateAccountGroup`  i.e. does it mean add or remove from the account group?

>3. Finally,  currently  our permission validations (i.e. calls to `context.validatePermissions`  in `3.0.0`)in `reaction`  are not using `reaction-authorization` but the old style permission checks. This is temporary measure but it means that we have to depend  roles in reaction  for permission checks 
(edited)

>@quophyie wrote
So I guess the question I'm asking is does reaction need to enforce a `one-group-per-user`  constraint? If it does, what what are the mutations that we need to expose and if doesnt, what are the mutations we need to expose for that scenario.

>@rosshadden wrote
The groups only exist for permissions. Reaction itself has no constraints around what can and cannot be done with them. We do not need to enforce that constraint. It was just how it worked with the old meteor permissions system.

>@rosshadden wrote
So # 2 in your list is what we want.

>@quophyie wrote
Thanks @rosshadden. So given what you said, we need to implement `addAccountToGroups` and `removeAccountFromGroups`  mutations but not `changeGroupForAccount` and `updateAccountGroup` I need some clarity some clarity on this because we spent some time on the `set/add/remove`  and now realise we not need them

>@quophyie wrote
@rosshadden... gotcha

>@rosshadden wrote 
@kieckhafer  can you weigh in with a sanity check here? I know we'd all like to avoid another rework situation.
:eyes:



>@kieckhafer  wrote
I believe what you last mentioned @quophyie  is correct. We DO NOT need to have the constraint, therefore we DO NOT need the "single group" mutations, which would be change and update.  (although i feel like these were the wame anyway?).
I think that means we DO need add and remove

>@quophyie wrote
Given all that has been said,  should I go ahead and close https://github.com/reactioncommerce/reaction/pull/5839, https://github.com/reactioncommerce/reaction/pull/5835,https://github.com/reactioncommerce/reaction/pull/5931 and https://github.com/reactioncommerce/reaction/pull/5842
and no longer required. I will create a ticket for removeAccountFromGroups as `addAccountToGroups`  already exists (edited) 

>@mikemurray wrote
`updateAccountGroup`  is to update the Group itself not the group for an account. The name is makes it confusing
:point_up:



>@mikemurray wrote
right now we can create a groups with `createAccountGroup` but we cannot update it. That’s why we need `updateAccountGroup`
:point_up:



>@mikemurray wrote
This PR: https://github.com/reactioncommerce/reaction/pull/5931 should stay open and be the updateAccountGroup  mutation

>@mikemurray wrote
the other 3 can be closed:
https://github.com/reactioncommerce/reaction/pull/5839
https://github.com/reactioncommerce/reaction/pull/5835
https://github.com/reactioncommerce/reaction/pull/5842

>@mikemurray wrote
`updateAccountGrou`p and `removeAccountFromGroups` Do 2 actions
`updateAccountGroup`  Updates the group document, and updates all the permissions on all the users in those groups. (#5931) seems to do this now
`removeAccountFromGroups` Will remove the group from the account, and all the roles from the matching user record, IF and only IF they don’t overlap with another group on that account

>@quophyie wrote
@mikemurray... what would we be updating the group with? will we be updating it  with new roles etc? This is slightly confusing because updateAccountGroup will mean something totally different in reaction than it means in reaction-authorization. In `reaction` it means updating name, slug, permissions etc but in `reaction-authorization` it means either adding new members or removing them from the group (role in `reaction-authorization`)

>@rosshadden wrote
I don't think we should consider what it means in `reaction-authorization` to answer that question

>@quophyie wrote
Ok

>@rosshadden wrote
In terms of Reaction it means "edit name, slug, permissions, or shopId"

>@quophyie wrote
@rosshadden  @mikemurray  Ok ... gotcha ...  I can update (i.e. title branch name) https://github.com/reactioncommerce/reaction/pull/5842 to be  `removeAccountFromGroups` it will need a minimal amount of changes to do what @mikemurray  described instead of starting afresh... is  that gonna be a problem? (edited) 

>@rosshadden wrote
The authorization service actually only cares about _id which can't change. It creates and deletes roles when Groups are created and deleted. It doesn't even care about fields like permissions.
The groups are merely an organization of users. Their significance comes from their usage in policies. We define policies that say things like "reaction:groups:order-managers can delete orders". Any user given that role can then perform that action on orders.

>@rosshadden wrote
is that gonna be a problem?
Not at all! Do what you need to do

quophyie:
Closing this PR as its no longer required... See comment here https://github.com/reactioncommerce/reaction/pull/5839#issuecomment-567466425 for details

PR #5850, opened 2019-12-04
add graphql equivalent of create group

### LINKED ISSUE #5831: Add GraphQL equivalent of method group/createGroup
Please add GraphQL equivalent of https://github.com/reactioncommerce/reaction-admin/blob/trunk/imports/plugins/core/accounts/server/methods/group/createGroup.js. This will help our demeteorization efforts and also help transition to reaction-authorization service

quophyie:
Resolves #5831   
Impact: **major**  
Type: **feature**

## Issue

We need to add a graphQL equivalent of the `createGroup` found in `reaction-admin` at https://github.com/reactioncommerce/reaction-admin/blob/trunk/imports/plugins/core/accounts/server/methods/group/createGroup.js into reaction. This will allow for the creation of groups (called `roles` in `reaction-authorization`) that accounts can be made members

## Solution

- Add GraphQL equivalent of `createGroup` found in `reaction-admin` to `reaction` (resolver function and mutation functions)
- Add integration tests for the newly created GQL `createGroup`
- We've changed the name of the equivalent `createGroup` from `createAccountGroup` as that best describes the functionality what is being performed (see comment https://github.com/reactioncommerce/reaction/pull/5850#issuecomment-562785915)

## Breaking changes

- the `createGroup` GQL mutation has been renamed to `createAccountGroup`
- the `CreateGroupInput` GQL type has been renamed to `CreateAccountGroupInput`
- the `CreateGroupPayload` GQL type has been renamed to `CreateAccountGroupPayload`

## Testing

1. Tests should pass on CI
 - `tests/integration/api/mutations/createAccountGroup/createAccountGroup.test.js` 
3. unit tests are found at 
 - `src/core-services/account/mutations/createAccountGroup.test.js`
 - `src/core-services/account/resolvers/Mutation/createAccountGroup.test.js`


src/core/util/buildContext.js
```diff
@@ -1,5 +1,6 @@
 import Logger from "@reactioncommerce/logger";
 import ReactionError from "@reactioncommerce/reaction-error";
+import getSlug from "./getSlug.js"
```
mikemurray: `getSlug` should not be added to the context. Import it as needed in the files that need it from the `api-utils` package: https://github.com/reactioncommerce/api-utils

src/core-services/account/mutations/createGroup.js
```diff
@@ -0,0 +1,72 @@
...
+
+  // we are limiting group method actions to only users within the account managers role
+  await context.validatePermissions("reaction:account", "create", { shopId });
+
+  const defaultCustomerGroupForShop = Groups.findOne({ slug: "customer", shopId }) || {};
+
+  // TODO: Remove when we move away from legacy permission verification
+  const defaultAdminPermissions = (defaultCustomerGroupForShop.permissions || []).concat("dashboard");
+  const newGroupData = Object.assign({}, groupData, {
+    slug: context.getSlug(groupData.name), shopId
```
mikemurray: `getSlug` should be imported from the https://github.com/reactioncommerce/api-utils utils package. It should not come from context.

src/core/util/getSlug.js
```diff
@@ -0,0 +1,15 @@
+import { slugify } from "transliteration";
```
mikemurray: Remove this file. It's part of the `api-uitls` package

package.json
```diff
@@ -55,7 +55,8 @@
     "semver": "~6.3.0",
     "sharp": "~0.23.1",
     "simpl-schema": "~1.5.6",
-    "stripe": "~5.10.0"
+    "stripe": "~5.10.0",
+    "transliteration": "^2.1.7"
```
mikemurray: Remove transliteration, any function that needs it should exist in the `api-utils` package: https://github.com/reactioncommerce/api-utils

src/core-services/account/mutations/createGroup.js
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * @name group/createGroup
+ * @method
+ * @memberof Group/Methods
+ * @summary Creates a new permission group for a shop
+ * It creates permission group for a given shop with passed in roles
+ * @param {object} input - The input supplied from GraphQL
+ * @param {Object} input.groupData - info about group to create
+ * @param {String} input.groupData.name - name of the group to be created
+ * @param {String} input.groupData.description - Optional description of the group to be created
```
mikemurray: If this is optional you should wrap the variable name in brackets `[input.groupData.description]`.

src/core-services/account/mutations/index.js
```diff
@@ -11,6 +11,7 @@ import sendResetAccountPasswordEmail from "./sendResetAccountPasswordEmail.js";
 import setAccountProfileCurrency from "./setAccountProfileCurrency.js";
 import setAccountProfileLanguage from "./setAccountProfileLanguage.js";
 import updateAccountAddressBookEntry from "./updateAccountAddressBookEntry.js";
+import createGroup from "./createAccount"
```
mikemurray: I think your filename is wrong. Should it be `createGroup`?

src/core-services/account/resolvers/Group/createGroup.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param {Object} input.groupData - info about group to create
+ * @param {String} input.groupData.name - name of the group to be created
+ * @param {String} input.groupData.description - Optional description of the group to be created
+ * @param {Array} input.groupData.permissions - permissions to assign to the group being created
+ * @param {Array} input.groupData.members - members of the
+ * @param {object} context - The GraphQL context
+ * @param {String} context.shopId - id of the shop the group belongs to
+ * @returns {Object} - `object.status` of 200 on success or Error object on failure
+ */
+export default async function createdBy(_, {input}, context) {
```
mikemurray: The function name should match the filename `createGroup`.

src/core-services/account/mutations/createGroup.js
```diff
@@ -0,0 +1,72 @@
...
+ * @returns {Object} - `object.status` of 200 on success or Error object on failure
+ */
+export default async function createGroup(input, context) {
+  const { groupData, shopId } = input;
+  let _id;
+  const {
+    collections: {
+      Groups
+    }
+  } = context;
```
mikemurray: Let's avoid multi-level destructuring `const { Groups } = context.collections;` would suffice. It not only matches the style used throughout the app but is also simpler to read.

src/core-services/account/mutations/createGroup.js
```diff
@@ -0,0 +1,72 @@
...
+    // and should place the newly created group on the kakfa groups topic
+    // reaction authorization listens on the topic and creates role (group)
+    // reaction authorization
+    _id = Groups.insert(newGroupData);
+  } catch (error) {
+    Logger.error(error);
+    throw new ReactionError("invalid-parameter", "Bad request");
+  }
+
+  return { status: 200, group: Groups.findOne({ _id }) };
```
mikemurray: `return Groups.findOne({ _id })` would suffice. The response is the responsibility of Apollo. You don't have to send a status code back.

src/core-services/account/resolvers/Group/createGroup.js
```diff
@@ -0,0 +1,20 @@
...
+  return context.mutations.createGroup(input, context);
```
mikemurray: This function seems a little too light-weight. You should be accepting the `opaque shop ID` (base64 encoded) and decoding it here in the resolver. Then pass that decided shop ID to the mutation function. Please check other mutation resolvers on how this is achieved.

src/core-services/account/resolvers/Group/index.js
```diff
@@ -1,9 +1,11 @@
...
 import { encodeGroupOpaqueId } from "../../xforms/id.js";
 import createdBy from "./createdBy.js";
+import createGroup from "./createGroup.js";
 
 export default {
   _id: (node) => encodeGroupOpaqueId(node._id),
   createdBy,
-  shop: resolveShopFromShopId
+  shop: resolveShopFromShopId,
+  createGroup
```
mikemurray: `createGroup` shouldn't bein this file. This is the resolver for the `Group` object itself when it appears in a query or a mutation payload.

src/core-services/account/resolvers/Group/createGroup.js
```diff
@@ -0,0 +1,20 @@
+
```
mikemurray: If this is the mutation resolver for `createGroup` then it's not in the correct place. It should be in the `resolvers/Mutation` and exported from `resolvers/Mutation/index.js`.

mikemurray:
^To add to my above review. Your original PR comment should follow the template provided by GH (when you create a new one), and most importantly have testing instructions.

src/core/util/getSlug.js
quophyie: Done

src/core-services/account/mutations/createGroup.js
quophyie: done

src/core-services/account/resolvers/Group/createGroup.js
quophyie: done

src/core-services/account/mutations/index.js
quophyie: yeah it should be `createGroup `...  good catch

src/core-services/account/mutations/createGroup.js
```diff
@@ -0,0 +1,72 @@
...
+  } = context;
+
+  // (LEGACY) we are limiting group method actions to only users with admin roles
+  // this also include shop owners, since they have the `admin` role
+  await context.validatePermissionsLegacy(["admin"], null, { shopId });
+
+  // we are limiting group method actions to only users within the account managers role
+  await context.validatePermissions("reaction:account", "create", { shopId });
+
+  const defaultCustomerGroupForShop = Groups.findOne({ slug: "customer", shopId }) || {};
```
rosshadden: `findOne` returns a promise.

src/core-services/account/mutations/createGroup.js
```diff
@@ -0,0 +1,72 @@
...
+  // TODO: Remove when we move away from legacy permission verification
+  if (!newGroupData.permissions) {
+    newGroupData.permissions = [];
+  }
+
+  // TODO: Remove when we move away from legacy permission verification
+  newGroupData.permissions = _.uniq([...newGroupData.permissions, ...defaultAdminPermissions]);
+
+  // ensure one group type per shop
+  const groupExists = Groups.findOne({ slug: newGroupData.slug, shopId });
```
rosshadden: `findOne` returns a promise.

src/core-services/account/mutations/createGroup.js
```diff
@@ -0,0 +1,72 @@
...
+    // and should place the newly created group on the kakfa groups topic
+    // reaction authorization listens on the topic and creates role (group)
+    // reaction authorization
+    _id = Groups.insert(newGroupData);
+  } catch (error) {
+    Logger.error(error);
+    throw new ReactionError("invalid-parameter", "Bad request");
+  }
+
+  return { status: 200, group: Groups.findOne({ _id }) };
```
rosshadden: `findOne` returns a promise.

src/core-services/account/mutations/createGroup.js
```diff
@@ -0,0 +1,72 @@
...
+  if (groupExists) {
+    throw new ReactionError("conflict", "Group already exist for this shop");
+  }
+
+  try {
+    // Kafka connect mongo should be listening for insert events
+    // and should place the newly created group on the kakfa groups topic
+    // reaction authorization listens on the topic and creates role (group)
+    // reaction authorization
+    _id = Groups.insert(newGroupData);
```
rosshadden: `insert` returns a promise.

src/core-services/account/mutations/createGroup.test.js
```diff
@@ -0,0 +1,98 @@
...
+    name: "test group",
+    slug: "test group",
+    permissions: [
+      "dashboard"
+    ],
+    shopId: "test-shop-id"
+  };
+
+  mockContext.collections.Groups.insert.mockReturnValueOnce("test-group-id");
+  mockContext.collections.Groups.findOne
```
rosshadden: `insert` and `findOne` should both return a promise. It looks like you have `findOne` returning `undefined` and then later `fakeResult`. If we had to poll for variable changes I don't think any of us here would be fans of JavaScript ;).

src/core-services/account/mutations/createGroup.test.js
```diff
@@ -0,0 +1,98 @@
...
```
rosshadden: See comments below.

src/core-services/account/mutations/createGroup.js
quophyie: Done

package.json
quophyie: Done

src/core-services/account/mutations/createGroup.js
quophyie: done

src/core-services/account/resolvers/Group/createGroup.js
quophyie: @mikemurray done

src/core/util/buildContext.js
quophyie: @mikemurray  done

src/core-services/account/mutations/createGroup.js
quophyie: @mikemurray done

src/core-services/account/resolvers/Group/index.js
quophyie: @mikemurray ... I wasnt aware of this but thanks for pointing it out
Done

src/core-services/account/mutations/createGroup.js
quophyie: Done

src/core-services/account/mutations/createGroup.js
quophyie: @rosshadden Done

src/core-services/account/mutations/createGroup.js
aldeed: Also, new code should use `insertOne`, `updateOne`, `deleteOne`, etc. rather than `insert`, `update`, `remove` which are all deprecated. See driver docs here: http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html

src/core-services/account/mutations/createGroup.js
```diff
@@ -0,0 +1,72 @@
...
+
+  // TODO: Remove when we move away from legacy permission verification
+  newGroupData.permissions = _.uniq([...newGroupData.permissions, ...defaultAdminPermissions]);
+
+  // ensure one group type per shop
+  const groupExists = Groups.findOne({ slug: newGroupData.slug, shopId });
+
+  if (groupExists) {
+    throw new ReactionError("conflict", "Group already exist for this shop");
+  }
```
aldeed: I know this is from existing code, but technically it would be best to create a unique index in MongoDB on `slug` + `shopId` fields. That way it's impossible for it to happen. This existing check makes it unlikely but not impossible because two requests for the same slug could come in at virtually the exact same time. If we have the unique index, the `findOne` could be entirely removed and instead throw `Group already exists` error if there's a unique index error from Mongo from the insert below.

aldeed:
@quophyie Great start. In case you haven't already been using them, there are docs about how to do this here: https://docs.reactioncommerce.com/docs/next/graphql-create-query

I'm thinking we should take this opportunity to rename it to `createPermissionGroup` or `createAccountGroup` because "group" isn't clear at all. Thoughts @kieckhafer or @rosshadden ?

rosshadden:
> I'm thinking we should take this opportunity to rename it to `createPermissionGroup` or `createAccountGroup` because "group" isn't clear at all.

Good idea. Do you think we should rename the collection as well? This is named `createGroup` because the collection is `Groups`.
I have been seeing a future where Accounts are abstracted into its own microservice as well, and Groups along with them. So while `PermissionGroup` is slightly more expressive, I think it's worth the tradeoff to call them `AccountGroups` instead as it establishes a link with Accounts.

src/core-services/account/mutations/createGroup.js
quophyie: Done

src/core-services/account/mutations/createGroup.test.js
rosshadden: What I'm saying is the way these tests are written (using `mockReturnValueOnce`) implies the values of functions like `findOne` actually changes. So if you access it once it is `undefined` and if you access again it is `fakeResult`. This means to get the desired `fakeResult` value we would have to poll the variable until it is no longer `undefined`, and that is (very fortunately!) not how promises work.

The result of `findOne` is always a promise. In fact, it is always the _same_ promise. Before the promise resolves it is in a pending state and has no value in its internal slot, and after it resolves it is thrown into a resolved state and is given a value in its slot (which can be `undefined` if that was what it was resolved with).

Jest has great docs on working with promises in tests: https://jestjs.io/docs/en/asynchronous#promises

Also you should look at the other tests in this project to see how we have been working with them. Here is a random file pulled from this codebase: https://github.com/reactioncommerce/reaction/blob/42f24a555e56458b9a7479110430262b92195d41/imports/node-app/core-services/catalog/utils/publishProductToCatalog.test.js#L241-L257

quophyie:
@mikemurray @aldeed @rosshadden ... Sorry if I caused any confusion. This PR was originally meant to have been a draft PR / WIP. I still have some commits that I need to push up and I need to complete the integration tests

aldeed:
@quophyie Ticket to change collection name in the future sounds good. We shouldn't do it now because it's not worth a migration at this point.

quophyie:
> @quophyie Ticket to change collection name in the future sounds good. We shouldn't do it now because it's not worth a migration at this point.

@aldeed ... already done ... issue here https://github.com/reactioncommerce/reaction/issues/5863

quophyie:
> Some comments added, below and this next one about tests.
> 
> I see you have a unit test for `createGroup` and while that's certainly fine, an `integration` test is required. Unit tests are optional for the API (queries/mutations) as they're usually heavily mocked, while the integration tests are more realistic. The integration tests live in `/tests/integration` and are broken up by queries and mutations. Yours will live somewhere in the `tests/integration/api/mutations` folder.

@mike ... thanks.. this is done

src/core-services/account/mutations/createGroup.test.js
quophyie: @rosshadden ... as discussed, this is meant to return 2 different values on 2 seperate calls i.e. we are not polling as discussed. I have however, changed the return values to be promises as discussed as thats the expected return value from the calls

quophyie:
> Some comments added, below and this next one about tests.
> 
> I see you have a unit test for `createGroup` and while that's certainly fine, an `integration` test is required. Unit tests are optional for the API (queries/mutations) as they're usually heavily mocked, while the integration tests are more realistic. The integration tests live in `/tests/integration` and are broken up by queries and mutations. Yours will live somewhere in the `tests/integration/api/mutations` folder.

@mikemurray ... I have added the integration tests as requested

quophyie:
The tests are failing but I need to rebase again onto `v.3.0.0` as the  failing tests have been fixed in an updated `v.3.0.0`. However, my rebase seems to be failing so I'm working to fix that

quophyie:
Integration tests are working locally but are currently failing on CI because of heap out of memory issues. This is being addressed in https://github.com/reactioncommerce/reaction/pull/5888/commits/36df4641ab54dcae50ebcae6f60ca3902176e5d5

src/core-services/account/mutations/createGroup.js
quophyie: @aldeed ... with regards to the snippet below
>I know this is from existing code, but technically it would be best to create a unique index in MongoDB on slug + shopId fields

Were you referring to a compound index or two single indexes? ... I have created a compound index but If you clarify it, I will make the necessary change

src/core-services/account/mutations/createAccountGroup.js
```diff
@@ -0,0 +1,74 @@
...
+
+  // we are limiting group method actions to only users within the account managers role
+  await context.validatePermissions("reaction:account", "create", { shopId, legacyRoles: ["admin"] });
+
+  const defaultCustomerGroupForShop = await Groups.findOne({ slug: "customer", shopId }) || {};
+
+  // TODO: Remove when we move away from legacy permission verification
+  const defaultAdminPermissions = (defaultCustomerGroupForShop.permissions || []).concat("dashboard");
+
+  const nowDate = new Date(Date.now());
```
rosshadden: `new Date()` creates a "now" date by default.

src/core-services/account/mutations/createAccountGroup.js
```diff
@@ -0,0 +1,74 @@
...
+  // TODO: Remove when we move away from legacy permission verification
+  if (!newGroupData.permissions) {
+    newGroupData.permissions = [];
+  }
+
+  // TODO: Remove when we move away from legacy permission verification
+  newGroupData.permissions = _.uniq([...newGroupData.permissions, ...defaultAdminPermissions]);
+
+  // ensure one group type per shop
+  const groupExists = await Groups.findOne({ slug: newGroupData.slug, shopId });
```
rosshadden: I think this should be an upsert instead of a find/insert. It would be one operation, and there should be nothing wrong with overwriting data in this case. We would probably want to rename this mutation to `upsertAccountGroup`.

Does anyone have a problem with this? @mikemurray @aldeed 
We may want to do something like this across the codebase, not just here, to save DB operations.

src/core-services/account/mutations/createAccountGroup.js
```diff
@@ -0,0 +1,74 @@
...
+
+  if (groupExists) {
+    throw new ReactionError("conflict", "Group already exist for this shop");
+  }
+
+  try {
+    // Kafka connect mongo should be listening for insert events
```
rosshadden: Small nitpick, but please make this a block comment instead of a lot of line comments.

src/core-services/account/mutations/createAccountGroup.js
rosshadden: Let's just leave it for now, but I am a performance nut and would like to revisit this later, vetting other endpoints we have as well.

src/core-services/account/resolvers/Mutation/createAccountGroup.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {Array} [input.group.members] - members of the
+ * @param {object} context - The GraphQL context
+ * @param {String} context.shopId - id of the shop the group belongs to
+ * @returns {Object} - `object.status` of 200 on success or Error object on failure
+ */
+export default async function createAccountGroup(_, { input }, context) {
+  const { shopId } = input;
+  const decodedShopId = decodeShopOpaqueId(shopId);
+
+  const transformedInput = Object.assign({}, input, { shopId: decodedShopId });
```
rosshadden: You can do this to be (arguably) more concise:

```javascript
const transformedInput = { ...input, shopId: decodedShopId };
```

src/core-services/account/resolvers/Mutation/createAccountGroup.test.js
```diff
@@ -0,0 +1,48 @@
...
+      ],
+      shopId: endcodedShopId
+    }
+  };
+
+
+  mockContext.mutations.createAccountGroup.mockReturnValueOnce(Promise.resolve(fakeCreatedGroupResult));
+
+  const result = await createAccountGroup(null, { input }, mockContext);
+  const resolverMutationFunctionInternalInput = Object.assign({}, input, { shopId: "test-shop-id" });
```
rosshadden: Same as above:

```javascript
const resolverMutationFunctionInternalInput = { ...input, shopId: "test-shop-id" };
```

src/core-services/account/mutations/createAccountGroup.test.js
```diff
@@ -0,0 +1,92 @@
...
+      "dashboard"
+    ],
+    shopId: "test-shop-id"
+  };
+
+  const insertOneRes = {
+    ops: [fakeResult]
+  };
+  mockContext.collections.Groups.insertOne.mockReturnValueOnce(insertOneRes);
+  mockContext.collections.Groups.findOne
```
rosshadden: These mongo methods both return promises.

src/core-services/account/mutations/createAccountGroup.js
aldeed: @rosshadden Since the expectation when calling the mutation is that a group would be created, it's helpful to the client to actually throw an error if that group already exists, versus just updating it silently.

But re performance, I did leave a comment about this code on an earlier commit saying that we could use a dual key unique index on slug + shopId, and then catch the Mongo error below. That would eliminate this findOne.

quophyie: left as is

src/core-services/account/mutations/createAccountGroup.js
quophyie: done

src/core-services/account/mutations/createAccountGroup.js
quophyie: @rosshadden ... done

src/core-services/account/resolvers/Mutation/createAccountGroup.test.js
quophyie: @rosshadden  ... done

src/core-services/account/mutations/createAccountGroup.test.js
quophyie: @rosshadden ... done... oversight on my part

quophyie: done ... oversight on my part

src/core-services/account/resolvers/Mutation/createAccountGroup.test.js
quophyie: done

src/core-services/account/resolvers/Mutation/createAccountGroup.js
quophyie: @rosshadden ... done

src/core-services/account/mutations/createAccountGroup.js
quophyie: @rosshadden ... done

src/core-services/account/mutations/createAccountGroup.js
quophyie: @rosshadden ... done

src/core-services/account/mutations/createAccountGroup.js
quophyie: @rosshadden ... as requested ... leaving as is

PR #5852, opened 2019-12-04
Don't use Packages for tax settings, and tax rates GQL fixes

aldeed:
Impact: **minor**  
Type: **feature|bugfix**

## Issue
- Tax settings were stored in the `Packages` collection, which we're aiming to remove in 3.0
- New tax rates mutations were using `TaxRates` collection name while tax calculation code was still using `Taxes`.
- Some tax rates fields were listed as required in GQL schema but should not be. Also many of the descriptions were incorrect.

## Solution
- Move all tax settings to newer shop settings API
- Updated all `TaxRates` collection names to `Taxes`
- Fixed TaxRates GQL schema issues

## Breaking changes
At the moment there is not an automatic migration for tax settings, so you will need to re-set your primary and fallback tax service selections, as well as the default tax code for new variants. We may add a migration for all Packages collection things in the future.

## Testing
- Verify you can set and read all three tax settings using the shop settings GraphQL mutation and query. Setting requires admin permission but reading is open to anyone since they do not seem like sensitive data.
- Verify you can create, update, and delete tax rates. (3.0 may not be stable enough to test the storefront / order placement part of this yet, but it should be closer to working than it was before.)

### DIFF
--- a/src/core-services/taxes/i18n/en.json
+++ b/src/core-services/taxes/i18n/en.json
@@ -13,18 +13,28 @@
           "taxesTitle": "Taxes",
           "taxesDescription": "Tax configuration"
         },
-        "taxGrid": {
-          "taxCode": "Item Code",
-          "rate": "Rate",
+        "taxFormFields": {
           "country": "Country",
+          "postal": "Postal code",
+          "rate": "Rate",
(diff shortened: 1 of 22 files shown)


aldeed:
Added permission checks @mikemurray

PR #5862, opened 2019-12-07
refactor: fix failing integration tests in release-3.0.0

### LINKED ISSUE #5861: failing 3.0 integration tests
There are a few order related integration tests failing on https://github.com/reactioncommerce/reaction/pull/5772

They don't seem related to any of the work done there, and seem related to work that was pulled in separately.

We should merge #5772 into `release-3.0.0` to get those 200 files changes merged, and then figure out what's up with these tests, since it's not going to a released / default branch anyway.

https://github.com/reactioncommerce/reaction/blob/release-3.0.0/tests/integration/api/queries/orderByReferenceId/orderByReferenceId.test.js#L81-L86

kieckhafer:
Resolves #5861 
Impact: **minor**  
Type: **test**

## Issue
There were a few failing integration tests in #5772 

## Solution
Update the tests to use the new auth style

## Breaking changes
None

## Testing
1. See all tests pass

### DIFF
--- a/src/core-services/orders/queries/ordersByAccountId.js
+++ b/src/core-services/orders/queries/ordersByAccountId.js
@@ -16,12 +16,17 @@ export default async function ordersByAccountId(context, { accountId, orderStatu
   const { collections } = context;
   const { Orders } = collections;
 
-  if (!accountId) {
-    throw new ReactionError("invalid-param", "You must provide accountId arguments");
+  if (!accountId) throw new ReactionError("invalid-param", "You must provide accountId arguments");
+
+  // Validate user has permission to view orders for all shopIds
+  if (!shopIds) throw new ReactionError("invalid-param", "You must provide ShopId(s)");
+  for (const shopId of shopIds) {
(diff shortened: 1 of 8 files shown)


src/core-services/orders/queries/ordersByAccountId.js
```diff
@@ -16,12 +16,17 @@ export default async function ordersByAccountId(context, { accountId, orderStatu
   const { collections } = context;
   const { Orders } = collections;
 
-  if (!accountId) {
-    throw new ReactionError("invalid-param", "You must provide accountId arguments");
+  if (!accountId) throw new ReactionError("invalid-param", "You must provide accountId arguments");
+
+  // Validate user has permission to view orders for all shopIds
+  if (!shopIds) throw new ReactionError("invalid-param", "You must provide ShopId(s)");
```
kieckhafer: I moved this check higher up in the function.

In the past, we did not require shopIds, it was an optional field, but with the 3.0 changes, we have been requiring shopIds to be passed everywhere. Because of this, I can check this higher up in the function, and also remove what was line 36, and add it to line 29 here instead, because we will always have a shopId available.

Does that all make sense? Or should I still do the check that was line 36?

mikemurray: I think its fine to do the `shopId` check early since its always required.

PR #5890, opened 2019-12-11
test: add create/update/delete navigation item integration tests

### LINKED ISSUE #5886: add createNavigationItem mutation integration test
Add missing mutation integration test for creating navigation items

mikemurray:
Resolves #5886   
Impact: **major**  
Type: **bugfix|test**

## Issue

Missing tests for:
- `createNavigationItem`
- `updateNavigationItem`
- `deleteNavigationItem`
- incorrect id field passed to mongo update/query in `updateNavigationItem` mutation

## Solution

- Add the missing tests
- Fix a bug found in the `updateNavigationItem` mutation that prevented updates from taking place

## Breaking changes

none

## Testing

1. Tests should pass on CI


tests/integration/api/mutations/navigationItem/deleteNavigationItem.test.js
```diff
@@ -0,0 +1,113 @@
...
+  testApp = new TestApp();
+  await testApp.start();
+  await testApp.insertPrimaryShop({ _id: internalShopId, name: shopName });
+  await testApp.createUserAndAccount(mockAdminAccount);
+  await testApp.setLoggedInUser(mockAdminAccount);
+
+  createNavigationItem = testApp.mutate(CreateNavigationItemMutation);
+  deleteNavigationItem = testApp.mutate(DeleteNavigationItemMutation);
+
+  const { createNavigationItem: createResult } = await createNavigationItem({
```
trojanh: @mikemurray I think we should try to keep tests as isolated as possible from other APIs as if `CreateNavigationItemMutation` fails it will lead to failure of unrelated `deleteNavigationItem` and `updateNavigationItem` tests. 

So for every `CreateNavigationItemMutation` input/payload update, we will need to make sure all dependent tests are updates. 

I might be missing some point here as we are not creating data using factory like other tests, so please feel free to shed some light.

mikemurray: The create/update/delete mutations are all part of the navigation API so if the create is broken, then the delete and update especially could be affected.

We have no rules that state that integration tests must use factories, nor if they can or cannot use queries/mutations from other parts of the API where available. We are trying to verify things work together in full app integration, not as individual units. That being said, I would avoid going out of the boundaries of the plugin you're testing (unless it's necessary for the test). At that point, factories and mocks may prove useful.

trojanh: understood @mikemurray , really appreciate the quick explanation. Also, this is an "integration test" 😄 .  This will help me write better tests in the future.

Issue #4284, opened 2018-05-30
Catalog integration test to use Factory mocks


nnnnat:
## Issue Description
The Catalog GQL integration test has an entire folder of hand-coded mock objects. These mocks need to be replaced with the new Factory test util.

trojanh:
@aldeed is this relevant? can I work on this?

aldeed:
@trojanh Still some stuff to do here, yes. Specifically:

- In `tests/mocks` folder, there are four files, which are imported into `tests/integration/api/queries/catalogItemProduct` and `tests/integration/api/queries/catalogItems` tests. It would be good to move these closer to where they are used, maybe in `tests/integration/api/queries/catalogItems` since that is more the pattern lately. `mockShop` and `mockTags` files can possibly be deleted entirely and contents moved to the files that import them. Goal would be to delete the `tests/mocks` folder.
- Within those mock files, particularly `mockCatalogItems.js` and `mockCatalogProducts.js`, the mocks are much larger than they need to be. You can combine factory functions such as `Factory.CatalogItem`, `Factory.CatalogProduct`, `Factory.CatalogProductVariant`, and `Factory.CatalogProductOption` to create the objects, which will auto-fill most of the field values. You can provide override values for those that need to be something specific. Search for `Factory.CatalogProduct` in other tests for examples.
- After above changes, some of the test expected objects or snapshots may need to be updated because the field values are different, etc. This is fine to do as long as the test is still testing what it says it is testing.

trojanh:
great, thanks @aldeed , I will create a PR for this soon. 

One question: `catalogItemProduct` and `catalogItems` query both do not take shopId as input, any reason to not include it or do we plan to do it soon?

aldeed:
@trojanh `catalogItems` takes `shopIds` array, which seems like the same thing. `catalogItemProduct` probably does not take `shopId` because we usually include it in order to do permission checks (most permissions are shop-specific), but `catalogItemProduct` has no permission checks because it's meant to be public. It's not a bad idea to add `shopId` everywhere for consistency, but I don't know if everyone will agree. You could propose it in a separate issue if you want.

trojanh:
thanks @aldeed for the clarification. 

[...]

PR #5892, opened 2019-12-11
Eliminate factory mocks for catalogItemProduct

trojanh:
Resolves https://github.com/reactioncommerce/reaction/issues/4284
Impact: **minor**  
Type: **test|chore**

## Issue
Use Factory mocks for Catalog integration and removes `tests/mocks`

## Testing
1. Make sure tests pass


willopez:
@trojanh one of your commits is not signed off, please follow instructions here: https://github.com/reactioncommerce/reaction/pull/5892/checks?check_run_id=343813454

willopez:
@trojanh the DCO check is still failing, this is the specific commit that is missing the sign off: https://github.com/reactioncommerce/reaction/pull/5892/commits/7b162ca858d20447abab6afa28f6b4393fdb2c75

trojanh:
I have fixed that commit @willopez , thanks for helping me with this.

trojanh:
@willopez I have updated the test cases to use Factory mocks but one test is failing as offset doesn't seem to be working as expected i.e no matter what offset value I use it returns the same number of items, so next and previous pages are not present as expected earlier.

You can check the test case [here](https://github.com/trojanh/reaction/blob/eliminate-factory-mocks/tests/integration/api/queries/catalogItems/catalogItems.test.js#L208) 

Let me know if this should be behaving like this or do I need to update my code somewhere.

willopez:
@trojanh Let's revert the updates you made to the `catalogItems.test.js` for the featured sort test case. That particular query will be re-written soon, and the existing test case will work for now. This will allow us to move forward with this PR.

trojanh:
@willopez are you suggesting I keep using the mock folder for that particular test case, while we update the query to fix the underlying issue? 

If you can confirm there's an issue with the `catalogItems` for featured products query, I could fix that along with this test case. 

If this PR needs to be shipped without the fix, maybe I can comment or skip that particular test case with an explanation and we do it in another PR. 

Let me know your thoughts because the original intention of this PR was to eliminate mocks folder.

willopez:
@trojanh Let's break the `catalogItems` test case into its own PR and move forward with the rest in this PR. There is no bug AFAK with the sort by featured query, what I meant is that it will be re-written to be more performant in the future. So, the test is failing due to the use of the new mock factories, just not sure what is causing the issue. So, if you can debug and find out why that is, that would be great.

trojanh:
@willopez done, will create another PR for `catalogItems` tests and debug the issue. Thanks

PR #5899, opened 2019-12-11
fix: tags and nav query permissions

kieckhafer:
Impact: **minor**  
Type: **bugfix**

## Issue
Because of the removal of "customer" and "anonymous" roles for accounts, there are a few places where read-only permissions were not functioning as expected.

## Solution
Add an "all" group to mock the old `customer` role and allow for navigation and tags to be read by anyone.

## Breaking changes
None

## Testing
1. See that tags and navigation queries work as expected


kieckhafer:
Branched off of wrong branch. Closing for a correct PR.

PR #5900, opened 2019-12-11
fix: comment out auth checks that aren't valid until new permissions are finished

kieckhafer:
Impact: **minor**  
Type: **bugfix**

## Issue
Because of the removal of "customer" and "anonymous" roles for accounts, there are a few places where read-only permissions were not functioning as expected.

## Solution
Add an "all" group to mock the old `customer` role and allow for navigation and tags to be read by anyone.

## Breaking changes
None

## Testing
1. See that tags and navigation queries work as expected


rosshadden:
I don't like it but this seems okay because it's legacy.

kieckhafer:
I agree and want to revisit it ASAP but don't think it's a blocker for the `alpha` and because it's `legacy`, so considering these were additional checks that weren't there before, that's why I made the decision to do it.

PR #5910, opened 2019-12-12
feat: add updateProductVariantPrices mutation to API

kieckhafer:
Impact: **minor**  
Type: **feature**

## Issue
There is no way to update price of a variant via the API

## Solution
Add a new mutation, `updateProductVariantPrices`, to the `simple-pricing` package.

## Breaking changes
None

## Testing
1. See that you can update `price` via the `updateProductVariantPrices` mutation
```
mutation {
  updateProductVariantPrices(input: {
shopId: "{encoded-shop-id}"
    variantId: "{encoded-product-id}",
    prices: {
      price: 19.99
      compareAtPrice: 29.99
    }
  }) {
    clientMutationId
  }
}
```


src/plugins/simple-pricing/mutations/updateProductVariantPrices.js
```diff
@@ -0,0 +1,57 @@
...
+    {
+      _id: variantId,
+      shopId
+    },
+    {
+      $set: updateDocument
+    }
+  );
+
+  const updatedProduct = Products.findOne({ _id: variantId, shopId });
```
aldeed: @kieckhafer If you use `findOneAndUpdate` pattern that we've used elsewhere then all three of these db commands are combined into one.

src/plugins/simple-pricing/mutations/updateProductVariantPrices.js
```diff
@@ -0,0 +1,60 @@
...
+  const { prices, variantId, shopId } = input;
+
+  // Check that user has permission to create product
+  await context.validatePermissions(`reaction:products:${variantId}`, "update:prices", {
+    shopId,
+    legacyRoles: ["createProduct", "product/admin", "product/update"]
+  });
+
+  const product = await Products.findOne({ _id: variantId, shopId });
+  if (!product) throw new ReactionError("not-found", "Product variant not found");
```
mikemurray: I wonder if we should also check `type` to make sure it's a variant?

src/plugins/simple-pricing/mutations/updateProductVariantPrices.js
```diff
@@ -0,0 +1,60 @@
...
+    optional: true
+  },
+  price: {
+    type: Number,
+    optional: true
+  }
+});
+
+/**
+ * @method updateProductVariantPrice
```
mikemurray: should be `updateProductVariantPrices`

kieckhafer:
@mikemurray @aldeed all comments addressed

src/plugins/simple-pricing/mutations/updateProductVariantPrices.js
```diff
@@ -35,24 +44,13 @@ export default async function updateProductVariantPrices(context, input) {
...
+  const { value: updatedProduct } = await Products.findOneAndUpdate(
+    { _id: variantId, shopId, type: "variant" },
+    { $set: { ...prices }},
+    { returnOriginal: false }
   );
 
-  const updatedProduct = Products.findOne({ _id: variantId, shopId });
+  if (!updatedProduct) throw new ReactionError("error-occurred", "Unable to update variant prices");
 
   appEvents.emit("afterVariantUpdate", { productId: variantId, product: updatedProduct });
```
aldeed: @kieckhafer It looks like the correct way to emit this as is currently done in reaction admin is

```
appEvents.emit("afterVariantUpdate", { _id, field, value });
```

Where `_id` is the variant ID and it is called once per field that was changed.

I also noticed that in `src/core-services/catalog/startup.js`, it is looking for `productId`, which we never even emit, and using it as the variant ID (not product). BUT it actually just uses it to GET the product ID. So...

Can you?
- Change this to emit once per changed field (price and/or compareAtPrice)
- Emit with this signature: `{ _id: variantId, productId: updatedProduct.ancestors[0], variant: updatedProduct, field, value }`
- Change the event consumer in `src/core-services/catalog/startup.js` to this:

```js
appEvents.on("afterVariantUpdate", async ({ productId }) => {
    hashRelatedProduct(productId, collections).catch((error) => {
      Logger.error("Error in afterVariantUpdate", error);
    });
  });
```

kieckhafer: @aldeed I think this takes care of it: https://github.com/reactioncommerce/reaction/pull/5910/commits/92a296609daf0eeeaf8630c1e5764c4c5749bcbf

Not sure if looping over the object like that is the best way to do it though.

kieckhafer:
I updated this to point to `release-3.0.0` instead of the alpha branch. This was branched off alpha, so it'll need to wait to be merged until alpha is released and merged back into `release-3.0.0`

aldeed:
Approved code. Didn't test

PR #5928, opened 2019-12-17
tests: Use Factory for catalogItems test

trojanh:
Resolves https://github.com/reactioncommerce/reaction/issues/4284
Impact: **minor**  
Type: **test|chore**

## Issue
Use Factory mocks for catalogItems integration and removes `tests/mocks`

## Testing
1. Make sure tests pass


trojanh:
I debugged this issue a bit and following is the flow 
1. [catalogItems.js#L41 => catalogItemsAggregate](https://github.com//reactioncommerce/reaction/blob/0f9fef20fb50087b51fb74c23e8608f4e7a246ac/src/core-services/catalog/resolvers/Query/catalogItems.js#L41)
2. [catalogItemsAggregate.js#L38 => arrayJoinPlusRemainingQuery](https://github.com//reactioncommerce/reaction/blob/0f9fef20fb50087b51fb74c23e8608f4e7a246ac/src/core-services/catalog/queries/catalogItemsAggregate.js#L38)
3. [arrayJoinPlusRemainingQuery.js#L154 => arrayJoinQuery ](https://github.com/reactioncommerce/api-utils/blob/trunk/lib/arrayJoinPlusRemainingQuery.js#L154)

I think at this step itself `connectionArgs` starts ignoring `offset` because of which the final response doesn't skip the offset pages. 

@aldeed @willopez I will be debugging and finding a solution for this tomorrow but if you could provide me some early leads that would be great.

trojanh:
@aldeed will adding `applyOffsetPaginationToMongoCursor` method before this line 
[arrayJoinPlusRemainingQuery.js#L238](https://github.com/reactioncommerce/api-utils/blob/trunk/lib/arrayJoinPlusRemainingQuery.js#L238) to calculate and adjust cursor for `offset` be fine? or there are chances of breaking?

We need to write some more integration tests to cover `offset`, `before` and `after` for `arrayJoinPlusRemainingQuery` and `getPaginatedResponse` methods.

aldeed:
@trojanh It seems to me that that test was written incorrectly and should have been failing this whole time. `arrayJoinPlusRemainingQuery` supports cursor-based pagination but was never updated to support `offset` pagination. I'm sure it's possible to do so, but such a change should be within the `api-utils` repo.

I suggest:
- Comment out that test with a note about why
- Get remaining tests working and add any others you think are missing (`mockShop` file is still imported in another file)
- We'll merge this PR
- Create an issue in https://github.com/reactioncommerce/api-utils requesting that `arrayJoinPlusRemainingQuery` be updated to handle the `offset` connection arg.
- Optionally take that issue and try to do it if you want. I'd start by creating a simple unit test for it in `api-utils`, and then modify `arrayJoinPlusRemainingQuery` to get that test passing. Hopefully then publishing that change and updating version of `api-utils` pkg in Reaction API will be all that's necessary to get the offset featured order pagination working.

trojanh:
great! thanks @aldeed , I'll make the changes to this PR and try to fix this issue in  `api-utils`.

trojanh:
I created the issue for `offset` here https://github.com/reactioncommerce/reaction/issues/6514 and also mentioned it in the [code comment here](https://github.com/reactioncommerce/reaction/pull/5928/files#diff-98a07a740516f1e479472f1ac2064103R191)

> Get remaining tests working and add any others you think are missing (mockShop file is still imported in another file)

I couldn't find any other file using `mockShop`, could you help me with the file you found?

Please review this PR and let me know if there's still anything left.

kieckhafer:
@trojanh accidently closed this when we changed some branches, I will re-open in a new PR.

kieckhafer:
Actually @trojanh can you please make a new PR for this, you can just copy/paste all the same text here, just make it to the `trunk` branch. When we deleted the `release-3.0.0` branch, it closed all PR's that were open, and won't let us update to use `trunk`.

Sorry for the confusion here.

trojanh:
sure @kieckhafer no problem, have raised a new PR 
https://github.com/reactioncommerce/reaction/pull/6095

PR #5972, opened 2019-12-19
refactor: use globalRoles if shopId is not provided

kieckhafer:
Impact: **minor**  
Type: **feature**

## Issue
If shopId is not provided in a permission check, the check fails no matter what.

## Solution
Add a check to check against `__global_roles__` if `shopId` isn't provided.

## Testing
1. Perform an action using a mutation where `shopId` is not provided, such as `updateAppSettings.js`
1. See that permissions pass if the user has global roles


kieckhafer:
This functionality was deemed unnecessary.

PR #5974, opened 2019-12-19
chore: adds integration test for updateShopSettings mutation

### LINKED ISSUE #5971: Add integration test for updateShopSettings mutation
## Issue Description
There is no integration test for `updateShopSettings` mutation

willopez:
Signed-off-by: Will Lopez <will.lopez77@gmail.com>

Resolves #5971  
Impact: minor
Type: chore | test

## Issue
Adds integration test for `updateShopSettings` mutation

### DIFF
--- a/tests/integration/api/mutations/updateShopSettings/updateShopSettings.test.js
+++ b/tests/integration/api/mutations/updateShopSettings/updateShopSettings.test.js
@@ -14,6 +14,8 @@ let testApp;
 let shopSettingsMutation;
 
 const mockAdminAccount = Factory.Account.makeOne({
+  _id: "345",
+  shopId,
   roles: {
     [shopId]: ["admin"]
   }
@@ -52,7 +54,7 @@ test("an anonymous user cannot update shop settings", async () => {
 
 test("shop settings can be updated by an admin user", async () => {
   let result;
-  testApp.setLoggedInUser(mockAdminAccount);
+  await testApp.setLoggedInUser(mockAdminAccount);
 
   try {
     result = await shopSettingsMutation({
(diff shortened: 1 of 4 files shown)


kieckhafer:
It seems like the issue you're seeing with permissions is that the `user` is not being set correctly on the `context` / `mockContext`. If you log `userId` [right before the if check here](https://github.com/reactioncommerce/reaction/blob/chore-integration-test-updateShopSettings/src/core/util/buildContext.js#L27), you'll see that `userId` is `null`, which means that it'll use the `else`, which automatically throws a permission error if there is no user.

I don't think this is an issue with the permissions itself, just an issue with setting the current user when testing.

It does look like you are [setting the user correctly in the test](https://github.com/reactioncommerce/reaction/blob/chore-integration-test-updateShopSettings/tests/integration/api/mutations/updateShopSettings/updateShopSettings.test.js#L55:L55), so perhaps it gets reset somewhere along the way?

tests/integration/api/mutations/updateShopSettings/updateShopSettings.test.js
```diff
@@ -0,0 +1,74 @@
...
+          canSellVariantWithoutInventory: false
+        }
+      }
+    });
+  } catch (error) {
+    expect(error).toBeUndefined();
+    return;
+  }
+
+  expect(result.updateShopSettings.shopSettings.canSellVariantWithoutInventory).toEqual(false);
```
mikemurray, line 83: I think there needs to be some verification of what the value was before setting it to `false`.

willopez: verification added.

PR #5976, opened 2019-12-19
test: full cart/checkout integration test

### LINKED ISSUE #5518: Write a full checkout flow integration test
Add a full checkout flow GraphQL integration test. It should create some initial catalog data, and then do all of the same GraphQL calls that the example storefront does, starting from add to cart, including each step of checkout, and ending with `placeOrder`.

This will help ensure that changes we make are not breaking the storefront checkout flow.

---

Tests cases needed

- createCart (with items)
- addCartItems (add more items)
- reconcileCarts (login flow)
- removeCartItems
- setEmailOnAnonymousCart
- setShippingAddressOnCart
- updateCartItemsQuantity
- updateFulfillmentOptionsForGroup
- selectFulfillmentOptionForGroup
- availablePaymentMethods
- placeOrder

mikemurray:
Resolves #5518 
Impact: **critical**  
Type: **bugfix|test**

## Issue

Need to test various mutations for cart/checkout

## Solution

- Add a test suite that runs though a full cart/checkout flow
- Fixes a bug with shipping price calculations

Test Cases
- [x] As a singed in user from the beginning, add items to cart and checkout
- [x] As an anonymous user, add items to cart and checkout
- [x] An an anonymous user, add items to cart, sign in, reconcile carts, and checkout

## Breaking changes

none

## Testing
1. Tests should pass


tests/integration/api/mutations/checkout/anonymousCheckout.test.js
```diff
@@ -0,0 +1,501 @@
...
+const SetEmailOnAnonymousCart = importAsString("./SetEmailOnAnonymousCartMutation.graphql");
+
+
+jest.setTimeout(300000);
+
+const encodeProductOpaqueId = encodeOpaqueId("reaction/product");
+
+const internalShopId = "123";
+const internalProductId = "999";
+const opaqueProductId = "cmVhY3Rpb24vcHJvZHVjdDo5OTk="; // reaction/product:999
```
willopez: why not use `const opaqueProductId = encodeOpaqueId("reaction/product", 999);`?

tests/integration/api/mutations/checkout/anonymousCheckout.test.js
```diff
@@ -0,0 +1,501 @@
...
+const internalTagIds = ["923", "924"];
+const opaqueShopId = "cmVhY3Rpb24vc2hvcDoxMjM="; // reaction/shop:123
```
willopez: Same as above, use the encoding utility

tests/integration/api/mutations/checkout/anonymousCheckout.test.js
```diff
@@ -0,0 +1,501 @@
...
+  placeOrder = testApp.mutate(PlaceOrderMutation);
+  publishProducts = testApp.mutate(PublishProductToCatalogMutation);
+  removeCartItems = testApp.mutate(RemoveCartItemsMutation);
+  setEmailOnAnonymousCart = testApp.mutate(SetEmailOnAnonymousCart);
+  selectFulfillmentOptionForGroup = testApp.mutate(SelectFulfillmentOptionForGroupMutation);
+  setShippingAddressOnCart = testApp.mutate(SetShippingAddressOnCartMutation);
+  updateCartItemsQuantity = testApp.mutate(UpdateCartItemsQuantityMutation);
+  updateFulfillmentOptionsForGroup = testApp.mutate(UpdateFulfillmentOptionsForGroupMutation);
+
+  await testApp.insertPrimaryShop({
```
willopez: seeing this error when running the tests
```
 Error in "afterOrderCreate" consumer Error: No email template orders/new found for language mockLanguage
```
this is due to the fact that the email templates do not exist in the database. They are created when the `createShop` mutation is executed. For tests such as this one that need email templates created, use the `createShop` mutation instead of in inserting the shop directly into the db. Further, to set other settings, use `updateOne` after the shop is created. This issue is present in all three tests.


only admins can create shops, so you will need a mock admin account 
```
const mockAdminAccount = Factory.Account.makeOne({
  _id: "mockAdminAccountId",
  roles: {
    [internalShopId]: ["admin"]
  },
  shopId: internalShopId
});
```

```
  await testApp.createUserAndAccount(mockAdminAccount);
  await testApp.setLoggedInUser(mockAdminAccount);
 await testApp.context.mutations.createShop({ ...testApp.context, isInternalCall: true }, {
    name: shopName,
    shopId: internalShopId
  });

  // Set other settings
  await testApp.collections.Shops.updateOne(
    { _id: internalShopId },
    { $set: { availablePaymentMethods: ["iou_example"] } },
    { $set: { allowGuestCheckout: true } }
  );
```

for a complete reference checkout this test: https://github.com/reactioncommerce/reaction/blob/release-3.0.0/tests/integration/api/mutations/inviteShopMember/inviteShopMember.test.js#L49

tests/integration/api/mutations/checkout/anonymousCheckout.test.js
mikemurray: Didn't encode since this style was used for pretty much all of the other tests. If this is something we want to change, all those other tests should be on the radar for an update at some point.

willopez: we have both styles, for all the tests I have written I use the encoding utility. Not a hill to die on, but I believe it makes it clear how the value was computed if we use the util function.

mikemurray: Updated to use the util

tests/integration/api/mutations/checkout/anonymousCheckout.test.js
mikemurray: Updated to use the util

PR #5986, opened 2019-12-21
Revert "Fix cart total calculation when shipping includes handling fee"

kieckhafer:
Reverts reactioncommerce/reaction#5985

The solution provided here, while working, is not the correct place to make these changes.


manueldelreal:
Files reverted to their original state

PR #5987, opened 2019-12-21
fix: remove extra `handling` charge from fulfillment total

kieckhafer:
Impact: **major**  
Type: **bugfix**

## Issue
Total price mismatch between submitted payment and expected total when shipping includes a handling fee.

## Solution
Remove extra `handling` charge addition in `addInvoiceToGroup`

## Breaking changes
No breaking changes

## Testing
1. Set a shipping method that includes a handling fee.
2. Complete a checkout using this fulfillment method.
3. Payment must match the total for the order.

### DIFF
--- a/imports/node-app/core-services/orders/util/addInvoiceToGroup.js
+++ b/imports/node-app/core-services/orders/util/addInvoiceToGroup.js
@@ -27,7 +27,6 @@ export default function addInvoiceToGroup({
 
   // Fulfillment
   const shippingTotal = group.shipmentMethod.rate || 0;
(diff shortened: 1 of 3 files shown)


kieckhafer:
@mikemurray this seems to be the appropriate fix for the price mismatch issue we were debugging. @aldeed mentioned he'd look into refactoring the names and confirming the calculations in the 3.0 branch, but you should probably update #5976 to include this change instead of [this one](https://github.com/reactioncommerce/reaction/pull/5976/commits/b25ed022b610533a510293ac9ce075bfed26cae9).

PR #5991, opened 2019-12-30
fix: fix default public password reset URL

aldeed:
Impact: **minor**  
Type: **bugfix**

## Issue
Default `REACTION_IDENTITY_PUBLIC_PASSWORD_RESET_URL` env variable value used when `NODE_ENV` is `development` doesn't match the actual Reaction Identity URL.

## Solution
Updated URL to match Reaction Identity.

NOTE: In production environments, you will need to update the URL manually. The path part should be `/account/reset-password/TOKEN`

## Breaking changes
None

## Testing
Test this with https://github.com/reactioncommerce/example-storefront/pull/637. All instructions are in that pull request.


willopez (approved):
LGTM

PR #6000, opened 2020-01-06
refactor: pass cart down to calculateOrderTaxes fn

### LINKED ISSUE #5583: Pass `cart` to `calculateOrderTaxes` functions
## Issue
Some custom `calculateOrderTaxes` functions need more information off the `cart`. They currently receive only the CommonOrder, which is a transformed version of each cart fulfillment group. There is no way to pass down any custom fields that have been added to `cart`.

## Solution
Pass `cart` all the way down and then into `calculateOrderTaxes`. When it's an order, `cart` will be `null`.

aldeed:
Resolves #5583   
Impact: **minor**  
Type: **refactor**

## Changes
When calling `calculateOrderTaxes` function for a fulfillment group that's in a cart, the parent `cart` object is passed in now. When called for an order, it's `null`. This is helpful for some tax plugins that may be storing additional information on the cart itself.

## Breaking changes
None

## Testing
Using logging or Node debugger, verify that `cart` is set on the `calculateOrderTaxes` function input object. It is called whenever you add or remove items from your cart or otherwise modify it.

### DIFF
--- a/src/core-services/taxes/mutations/getFulfillmentGroupTaxes.js
+++ b/src/core-services/taxes/mutations/getFulfillmentGroupTaxes.js
@@ -7,7 +7,8 @@ import { TaxServiceResult } from "../simpleSchemas.js";
  * @summary Returns all taxes that apply to a provided order, delegating to a more specific
  *   tax calculation service for the actual calculations.
  * @param {Object} context App context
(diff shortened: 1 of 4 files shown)


willopez (approved):
Verified

PR #6001, opened 2020-01-06
Warn about non-unique names of functionsByType functions

### LINKED ISSUE #5464: Require all functions with a `functionsOfType` to have a unique name
Right now the pattern seems to be if registering a function under `publishProductToCatalog` call your function `publishProductToCatalog`. This is really the opposite of what you should do, and probably almost always incorrect since your plugin *doesn't* actually publish anything to catalog it just mutates it. 

The driver here from me is that it makes this impossible to know what function added by what plugin is causing a problem.

Since we can't programmatically require GOOD names we should at least require UNIQUE names and this would be relatively easy to enforce and force people to come up with a better name

aldeed:
Resolves #5464   
Impact: **minor**  
Type: **refactor**

## Issue
Logging and stack traces are confusing when all functions of a particular type have the same name, as has been the pattern. For example, all plugins that register a "startup" function also name that function "startup".

## Solution
#5464 suggested making unique names mandatory but I'm only logging a warning on startup. I've updated all built-in services and plugins so that all function names are unique and the warning does not print on startup.

## Breaking changes
None

## Testing
Verify no warnings about `functionsByType` name are shown in the API startup logs. Temporarily change one fn name to be the same as another of that type, and verify that the warning now shows in the logs.

### DIFF
--- a/src/core-services/account/index.js
+++ b/src/core-services/account/index.js
@@ -2,7 +2,7 @@ import i18n from "./i18n/index.js";
 import mutations from "./mutations/index.js";
 import policies from "./policies.json";
 import queries from "./queries/index.js";
-import { registerPluginHandler } from "./registration.js";
+import { registerPluginHandlerForAccounts } from "./registration.js";
 import resolvers from "./resolvers/index.js";
 import schemas from "./schemas/index.js";
 import startup from "./startup.js";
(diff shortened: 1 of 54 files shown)


willopez (approved):
LGTM

PR #6008, opened 2020-01-07
Fix user roles and owner permission checks

### LINKED ISSUE #6002: [3.0.0] addAccountToGroup throws an error after account creation due to missing roles object
## Issue Description
The new reaction-identity Meteor app does not take care of roles anymore, therefore new users in the `users` collection will not have an object `roles`.
Then in the `reaction-api` node server`createAccount` checks that the user to be created has no default groups, and will search for the Id of the customer group by slug `customer`.
It instructs `addAccountToGroup` to put the User to the customer group.

`addAccountToGroup` will try to fetch existing rules:
`const newAccountUserRoles = new Set(accountUser.roles[shopId] || []);`

This will fail, as for new users there is no `accountUser.roles` anymore.
As the `addAccountToGroup` calls are inside of an try catch block, `createAccount` will still finish and send an email.

## Steps to Reproduce
1. Create an account for the `users` collection in the identity app
2. When redirected to the storefront and the first GraphQL call is made, notice how the `reaction-api` container throws an error `TypeError: Cannot read property 'J8Bhq3uTtdgwZx3rz' of undefined`

## Possible Solution
There are multiple things here:
- First of all, if still needed at all, `accountUser.roles` should be checked to exist in `addAccountToGroup` to fix the current error
- This would result in the `customer` group being added to the user, which I read is considered deprecated.
- Should customer be removed? Stop creating it in `createAuthGroupsForShop` and in `createAccount` just leave an empty `groups` array as is, and don't try adding a customer group.

- Maybe something else to consider when keto is in place: Should `addAccountToGroup` try to add Meteor user roles at all with `newAccountUserRoles`? They should already be redundant for new customers, but might still be needed for admin users in the meantime(TODO, need to check: Are they? Or are they already assigned when a user becomes admin/owner?) Backwards-compatibility for existing accounts might be another consideration, although a migration probably is preferable here.

### Versions
latest release-3.0.0 branch

aldeed:
Resolves #6002   
Impact: **minor**  
Type: **bugfix**

## Issue
- `addAccountToGroup` mutation doesn't properly handle `user` doc that has no `roles` property
- check for global ownership in `createAccount` isn't awaited and is therefore always truthy
- When `account.shopId` is null, which is usually the case now in 3.0.0 release, an account was not initially added to any groups.

## Solution
- `addAccountToGroup` mutation will not fail if `user.roles` array is missing
- check for global ownership in `createAccount` is awaited so only the first user created will end up in the "owner" group
- When `account.shopId` is null and a primary shop exists, the account is associated with the correct group for the primary shop

Additionally removed `addRolesToGroups` support for `registerPlugin` config, which is no longer needed. The roles list is currently stored in the reaction-admin app for all plugins. This will be implemented in a better way with the new authorization service.

## Breaking changes
If your plugin passes `addRolesToGroups` option to `registerPlugin`, it's now ignored.

## Testing
I suggest just code review plus testing the basic user reg/login/logout flows in conjunction with testing https://github.com/reactioncommerce/reaction-admin/pull/171


mikemurray (approved):
👍

PR #6020, opened 2020-01-09
Better internal context design

### LINKED ISSUE #5808: [3.0.0] Better context auth switching
In 2.x API, sometimes `context.isInternalCall = true` is used to skip authorization checks in individual functions. There are a few problems with this approach:

- It has to be checked explicitly in each function (`if (!context.isInternalCall)`). Some don't, which means they then cannot be called from other internal code.
- There is some slight chance that `isInternalCall` might be stuck to `true` in an unexpected way, which could result in security issues. This is unlikely since the context object is rebuilt on every request, but it would be best to avoid the potential issue entirely.
- We pass it like `{ ...context, isInternalCall: true }`, which is a little clumsy and non-intuitive

For 3.0, I propose the following better way of changing the auth context, which also allows easily simulating any user or no user.

- `context.getContextForUser(userId)`: returns `context` with `user`, `userId`, `account`, `accountId`, `userHasPermission`, and `checkPermissions` all adjusted as if `userId` had authenticated
- `context.getContextForUser(null)`: returns `context` with `user`, `userId`, `account`, `accountId`, `userHasPermission`, and `checkPermissions` all adjusted as if an unauthenticated request was made
- `context.getInternalContext()`: returns `context` with `user`, `userId`, `account`, `accountId`, `userHasPermission`, and `checkPermissions` all adjusted as if the user has all possible permissions

Before, when you want to call a function internally and bypass its permission checks:

```js
await context.mutations.createNavigationItem({ ...context, isInternalCall: true }, { /* ... */ });
```
Now, when you want to call a function internally and bypass its permission checks:

```js
await context.mutations.createNavigationItem(context.getInternalContext(), { /* ... */ });
```

aldeed:
Resolves #5808   
Impact: **breaking**  
Type: **feature**

## Issue
See #5808 

## Solution
This implements just the bare minimum suggested in #5808, to make the existing "internal calls" feature less error prone. After new authorization is fully implemented, it may be a better approach to authorize service-to-service calls rather than skipping authorization entirely. This will certainly be true for any calls that come from another microservice. But for now we have a need of calling internal functions while skipping their auth checks, and this way of doing it is better than the piecemeal way we did it in 2.x.

Before:

```js
context.mutations.someFunction({ ...context, isInternalCall: true }, ...);
```

After:

```js
context.mutations.someFunction(context.getInternalContext(), ...);
```

The primary benefit here is that this will now work correctly and securely for all queries and mutations without needing each function to individually implement the feature.

## Breaking changes
Plugins that call internal functions with `isInternal: true` on context will need to be updated.

## Testing
Code review in conjunction with the existing integration tests should be sufficient, but you could step through some code to verify what is happening.


kieckhafer (approved):
Looks good.

Makes sense where / why the 3 places of `isInteralCall` still exist.

👍

PR #6021, opened 2020-01-10
refactor: remove legacyRoles from authorization

kieckhafer:
Impact: **major**  
Type: **refactor|chore**

## Issue
As a way to keep legacy roles / permissions working while we updated our authorization service, we had added a `legacyRoles` array to our `userHasPermission` and `validatePermission` checks to pass in the legacyRoles and use the legacy system. This was known to be a temporary solution, and as we move forward, we wanted to remove it.

## Solution
Remove the `legacyRoles` array from the `userHasPermission` and `validatePermission` checks, and instead re-write the legacy `hasPermission` function to be able to use a variation of our new keto-based resources structure: `<organization>:<system>:<entity>:<id>`.

To use the keto structure, we take the first three pieces of our namespacing -<organization>:<system>:<entity> - and attach the action to the end, to create a new role. For example:

The old validatePermissions would use legacyRoles to check:
```
await context.validatePermissions(
      `reaction:legacy:orders:${order._id}`,
      "move:item",
      { shopId: order.shopId, legacyRoles: ["orders", "order/fulfillment"] }
    );
```

The new validatePermissions looks like this:
```
await context.validatePermissions(
      `reaction:legacy:orders:${order._id}`,
      "move:item",
      { shopId: order.shopId }
    );
```
And will pass this string to our existing auth service: `reaction:legacy:orders/move:item`.

The existing roles of `orders` and `order/fulfillment` are replaced by `reaction:legacy:orders/move:item`

## Breaking changes
- Anywhere using `legacyRoles` for validation checks will no longer work. This should not be an issue as this was introduced in 3.0.0, which is still in alpha.

## Testing
1. This replaces ALL permission checks throughout the app, so click around, perform as many actions as you can think of: create orders, fulfill orders, update payments, update shop settings, etc, etc, and make sure they all work as expected / fail as expected based on the current user and the action trying to be taken.
1. Integration tests should prove that it works, but it would be good to manually click around as well.


kieckhafer:
@mikemurray @aldeed @rosshadden Since this touches an obscene amount of files, just wanted as many eyes on this as possible.

Possibly left to do is creating a migration to assign new roles to old users based on what they had before. @aldeed is this something that needs to be done, or are we saying 3.0 starts a new?

rosshadden (approved):
:+1:. And +1 to the question you asked @aldeed. I had been assuming we would make a migration but it is definitely reasonable that we may be able to start fresh.

aldeed:
@kieckhafer @rosshadden A few things I can think of at the moment:
- https://github.com/reactioncommerce/reaction-identity/pull/14/files will need to be changed somehow. Probably just global shop:create is needed there for the first user? (We'll also need similar code in Identity for keto auth for the first user, but that's a separate issue.)
- I vote to use this opportunity to remove the `"owners" should always have access` logic. As long as we give the first user `shop:create` and anything else they might need, we don't need the concept of an "owner" permission since "owner" is more of a role. This might expose a bunch of faulty permission checks that were previously hidden, but that's a good thing we can work through before the final release.
- What do you think about adding a just-in-time migration for this? In the legacy-authorization `hasPermission` function, you can check whether `roles` contains any old roles. Then use an old-to-new map to build a new roles list, and do a `users.updateOne` to "upgrade" that user. Then check the new list.

mikemurray (approved):
👍

kieckhafer:
@aldeed opened a ticket for your comments, this PR is too big, and already causing many conflicts with every PR, to leave open for too long: https://github.com/reactioncommerce/reaction/issues/6030

Issue #6011, opened 2020-01-08
create  createNavigationTree mutation and an integration test for it.


willopez:
### Issue Summary
In order to support switching between multiple navigation trees in the admin UI, there needs to be a mutation to create navigation trees: `createNavigationTree`, which does not exist at the moment.

trojanh:
@willopez is this same as [createDefaultNavigationTreeForShop](https://github.com//reactioncommerce/reaction/blob/d587c859a8809b3df80088f16376d7039e0e8928/src/plugins/navigation/util/createDefaultNavigationTreeForShop.js#L6) with the ability to allow the some of the inputs in `createNavigationTree` mutation variable ?. 

If yes, what all fields are we planning to take from as input in this mutation?

aldeed:
@trojanh I think the only part of `createDefaultNavigationTreeForShop` that would be the same is this:

```js
// Create the tree
  const navigationTreeId = Random.id();
  await NavigationTrees.insertOne({
    _id: navigationTreeId,
    draftItems: items,
    hasUnpublishedChanges: true,
    items: [],
    name,
    shopId
  });
```

I would start from `updateNavigationTree` mutation function, but without `navigationTreeId` param because it will be randomly generated. Remove some things where they don't make sense, such as checking for an existing record. Instead of `NavigationTrees.updateOne`, do the `insertOne` above. 

The rest will be pretty similar to an update. In the gql schema unfortunately the existing `input NavigationTreeInput` won't be suitable because `name` should be required. I'm not sure the extra layer is needed anyway. I suggest adding this gql:

In `extend type Mutation` block:

```graphql
"Create a new navigation tree"
  createNavigationTree(
    "Mutation input"
    input: CreateNavigationTreeInput!
  ): CreateNavigationTreePayload!
```

And the new types:

```graphql
"Input for the `createNavigationTree` mutation"
input CreateNavigationTreeInput {
  "An optional string identifying the mutation call, which will be returned in the response payload"
  clientMutationId: String

  "The draft navigation items that make up this tree"
  draftItems: [NavigationTreeItemInput]

  "The name of the tree, for operator display purposes"
  name: String!

  "The ID of the shop this navigation tree belongs to"
  shopId: ID!
}

"Response payload for the `createNavigationTree` mutation"
type CreateNavigationTreePayload {
  "The same string you sent with the mutation params, for matching mutation calls with their responses"
  clientMutationId: String

  "The created navigation tree"
[...]

PR #6023, opened 2020-01-13
Add createNavigationTree mutation & its tests

trojanh:

Resolves https://github.com/reactioncommerce/reaction/issues/6011
Impact: **major**  
Type: **feature|test**

## Issue
In order to support switching between multiple navigation trees in the admin UI, there needs to be a mutation to create navigation trees: createNavigationTree, which does not exist at the moment.


src/plugins/navigation/schemas/schema.graphql
```diff
@@ -345,6 +360,18 @@ input NavigationTreeItemInput {
   navigationItemId: ID!
 }
 
+"Input for the `createNavigationTree` mutation"
+input CreateNavigationTreeInput {
+  "An optional string identifying the mutation call, which will be returned in the response payload"
+  clientMutationId: String
+
+  "The field updates to apply"
+  navigationTree: NavigationTreeInput!
```
aldeed: You can't reuse `NavigationTreeInput` because we need `name` to be required for an insert. You can put `name` and `draftItems` directly in this input as I did in https://github.com/reactioncommerce/reaction/issues/6011#issuecomment-572713305

src/plugins/navigation/schemas/schema.graphql
```diff
@@ -345,6 +360,18 @@ input NavigationTreeItemInput {
...
+
+  "The ID of the shop navigation item belongs to"
```
aldeed: "item" should be "tree" in this description

src/plugins/navigation/mutations/createNavigationTree.js
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  // Validate the navigation tree with the defaults set
+  NavigationTreeSchema.validate(navigationTreeData);
+  const { draftItems } = navigationTreeData;
+
+  await context.validatePermissions("reaction:navigationTrees", "create", { shopId, legacyRoles: ["core"] });
+
+  if (draftItems) {
+    decodeNavigationTreeItemIds(draftItems);
```
aldeed: @trojanh I see you copied this from how it's done in the update, but it's incorrect. All ID decoding should happen in the GraphQL resolvers. This function should assume that all IDs are internal. Would you mind fixing it here and also for the `updateNavigationTree` mutation, by moving `decodeNavigationTreeItemIds` lines to the resolvers?

src/plugins/navigation/mutations/createNavigationTree.js
```diff
@@ -0,0 +1,60 @@
...
+    shopId,
+    _id: Random.id(),
+    draftItems: setDefaultsForNavigationTreeItems(navigationTree.draftItems, visibilityDefaults)
+  };
+
+  // Validate the navigation tree with the defaults set
+  NavigationTreeSchema.validate(navigationTreeData);
+  const { draftItems } = navigationTreeData;
+
+  await context.validatePermissions("reaction:navigationTrees", "create", { shopId, legacyRoles: ["core"] });
```
aldeed: @kieckhafer What should this permission name be now? No `legacyRoles` needed, right?

aldeed (changes requested):
@trojanh Looks pretty good. Just a few minor adjustments needed.

src/plugins/navigation/mutations/createNavigationTree.js
kieckhafer: Correct, no legacy roles needed, and this permission [is already registered in a policy](https://github.com/reactioncommerce/reaction/blob/release-3.0.0/src/plugins/navigation/policies.json#L30:L30) so no other changes needed aside from the following:

```suggestion
  await context.validatePermissions("reaction:legacy:navigationTrees", "create", { shopId });
```

src/plugins/navigation/mutations/createNavigationTree.js
trojanh: sure @aldeed I was skeptical about this part but since it was used in `updateNavigationTree` I overlooked it. Thanks for correcting. 

I will fix this.

src/plugins/navigation/mutations/createNavigationTree.js
trojanh: thanks, @aldeed, and @kieckhafer for helping me with this, `roles` and `permission` are a bit tricky to understand at least for me. 

I would like to help you guys with the documentation if it's not already documented.

kieckhafer: 👍 We do have some documentation written up internally at the moment, it will be publicized with the official 3.0 launch but we are still working through some of the functionality so didn't want to publish it publicly yet knowing it's going to change. 

But we agree, it's tricky!

trojanh: awesome, I am very excited about the v3 launch and hope that adoption will be more and complains will be far less. More importantly, the developer experience will be much better now.

HarisSpahijaPon: @kieckhafer can you not invite some contributors for writing documentation on their own parts of code? A bit like how component library have documentation next to the component in the folder hierarchy.

aldeed (approved):

PR #6027, opened 2020-01-14
feat: change graphql route

rosshadden:
Impact: **minor**  
Type: **feature**

## Issue
We had our default graphql route set to `/graphql-beta` as we were phasing it in. Now that it is officially our only recommended API entrypoint we should remove the `-beta` suffix.

## Solution
This changes the default graphql route to be `/graphql`, updates all docs within the project, and adds a redirect for `/graphql-*` to point to the new path.

## Breaking changes
Technically none, as this adds a redirect for legacy endpoints. All the same, this does change the default graphql endpoint so projects should ultimately update to the new path on their own time.

## Testing
1. Verify that `/graphql` works for graphql requests.
2. Verify that `GET /graphql` loads the Apollo playground.
3. Verify that `/graphql-beta`, `/graphql-alpha`, and `/graphql-anything` all redirect to `/graphql`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ dc up -d mongo
 docker run -e NODE_ENV=production -e MONGO_URL=mongodb://mongo:27017/reaction -e ROOT_URL=http://localhost:3000 -p 3000:3000 --network streams.reaction.localhost -it test-api
 ```
 
-Use an external GraphQL client to test http://localhost:3000/graphql-beta. GraphQL Playground isn't served on GET requests because it's in production mode.
+Use an external GraphQL client to test http://localhost:3000/graphql. GraphQL Playground isn't served on GET requests because it's in production mode.
 
 # Get involved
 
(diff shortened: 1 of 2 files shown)


aldeed (changes requested):
@rosshadden This isn't from your change, but it turns out the `res.redirect` doesn't actually work. Neither Apollo Client nor GraphQL Playground seem to follow redirects for GQL requests. `/graphql-beta` does load GraphQL Playground in a browser after redirect, but requests to `/graphql-beta` fail and storefront won't load with `/graphql-beta` as the route.

I tried changing

```
res.redirect(path);
```

to

```
req.url = path;
app.handle(req, res);
```

That works for me. Please change and verify. Other than that, this looks good to merge.

rosshadden:
Doing a rewrite works for me too. Great catch!

aldeed (approved):

PR #6029, opened 2020-01-14
feat: update to Node 12.14.1

### LINKED ISSUE #5770: [3.0.0] Update to Node 12 LTS version
1. Make new prod and dev images in https://github.com/reactioncommerce/docker-base for the LTS (or higher) Node 12 version.
2. Update `.nvmrc`, "engines" in `package.json` file, and the image tags in `Dockerfile` and `docker-compose.yml` all to the same matching Node 12 LTS version.
3. Verify everything still works. I think they moved JSON import to be behind a flag so probably need to add that flag to our Node commands (see https://github.com/nodejs/node/pull/29754).

aldeed:
Resolves #5770   
Impact: **minor**  
Type: **feature**

## Changes
The project now runs on Node 12.14.1 both for development and for future production builds. This is an [LTS version](https://nodejs.org/en/about/releases/).

## Breaking changes
This version of Node requires the new `--experimental-json-modules` flag to run because we `import` JSON files from many ES modules. This is handled already in the `package.json` scripts but if you are doing something non-standard, you may need to add this flag yourself.

## Testing
In general you should test 3 variations:

Case 1:
`docker-compose up -d` with the `docker-compose.override.yml` file should just work

Case 2:

```sh
docker-compose down
docker build . -t test-api
docker-compose up -d mongo
docker run --env-file ./.env -p 3000:3000 --network streams.reaction.localhost -it test-api:latest
```

The `docker build` command should not have any errors and the API should start in production mode without any errors.

Case 3:
To test `docker-compose up -d` without the `docker-compose.override.yml` file, you will need to temporarily change the `api.image` value in `docker-compose.yml` to `reactioncommerce/reaction:feat-aldeed-node-12.14.1`


willopez (approved):
Tested all scenarios and they all work as intended.

PR #6032, opened 2020-01-15
fix: check if the url is already absolute

mpaktiti:
Signed-off-by: Maria Paktiti <maria.paktiti@gmail.com>

Resolves https://github.com/reactioncommerce/catalog-publisher/issues/253
Impact: **minor**  
Type: **bugfix**

## Issue
The image URLs coming from the Catalog Publisher, are already absolute. When this function is called to "absolutify" it again, the result looks like this:
```
{
  "data": {
    "catalogItemProduct": {
      "product": {
        "title": "Timberlands",
        "primaryImage": {
          "URLs": {
            "large": "http://reaction.api.reaction.localhost:3000/http://reaction.api.reaction.localhost:3000/assets/files/Media/rx9fN8A7qJtnAQr4L/large/adult-beautiful-beautiful-girl-1054249.jpg"
          }
        },
```

## Solution
Add a check to see if the `path` already is absolute. If not add the rootUrl, otherwise return the path. 

## Breaking changes
None, it's an additional check.

## Testing
Run the tests (I added one for this case), or send a GraphQL request for a product that has an absolute url already. The request should look like that:
```
{
  catalogItemProduct(slugOrId: "product3") {
    _id
    createdAt
    product {
      title
      primaryImage {
        URLs {
          large
        }
      }
    }
  }
}
```

### DIFF
--- a/src/core/util/getAbsoluteUrl.js
+++ b/src/core/util/getAbsoluteUrl.js
@@ -6,15 +6,16 @@
  * @returns {String} Full URL
  */
 export default function getAbsoluteUrl(rootUrl, path = "") {
+  // If the path already contains the rootUrl
+  // don't add it again
+  const regExp = /^https?:\/\/|^\/\//i;
+  if (regExp.exec(path) !== null) {
+    return path;
+  }
+
   let pathNoSlash = path;
   if (path.startsWith("/")) {
     pathNoSlash = path.slice(1);
-  }
-
-  // If the path already contains the rootUrl
-  // don't add it again
-  if (rootUrl === path.slice(0, rootUrl.length)) {
-    return pathNoSlash;
   }
 
   return `${rootUrl}${pathNoSlash}`;
--- a/src/core/util/getAbsoluteUrl.js
+++ b/src/core/util/getAbsoluteUrl.js
@@ -11,5 +11,11 @@ export default function getAbsoluteUrl(rootUrl, path = "") {
     pathNoSlash = path.slice(1);
   }
 
+  // If the path already contains the rootUrl
+  // don't add it again
+  if (rootUrl === path.slice(0, rootUrl.length)) {
+    return pathNoSlash;
+  }
+
   return `${rootUrl}${pathNoSlash}`;
 }
--- a/src/core/util/getAbsoluteUrl.js
+++ b/src/core/util/getAbsoluteUrl.js
@@ -6,10 +6,9 @@
  * @returns {String} Full URL
  */
 export default function getAbsoluteUrl(rootUrl, path = "") {
-  // If the path already contains the rootUrl
-  // don't add it again
+  // Check if the path is already absolute
   const regExp = /^https?:\/\/|^\/\//i;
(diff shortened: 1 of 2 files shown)


focusaurus (changes requested):
Thanks for adding tests. I'm thinking we should handle arbitrary URLs like cloudinary and other 3rd party services too. So I think any absolute URL starting with http or https we should ignore, and probably just to be fancy, we should ignore protocol-relative URLs (starting with "//"). Here's a screenshot of a regex that should work and some unit tests via regex101.com

screen-2020-01-15-09:40:54-url-regex-101

focusaurus (approved):
I think the `.test()` regex method is the best one for this case so you can swap that in but otherwise LGTM. Let's see if Eric D feels OK about this change.

aldeed (approved):
@mpaktiti I'm ok with this change, but I'll point out that it would be best for performance if the `files` service never even registers its `xformCatalogProductMedia` function when it's known that all image paths will already be absolute. That function does nothing else. Perhaps it could register it conditionally based on an ENV?

Either way the change in this PR is also good, but one point, which is nitpicky but this fn does run A LOT: RegExp can be slow so `path.toLowerCase().startsWith("http")` may be faster. You could run some speed tests on typical URLs to see.

Or even `path.slice(0, 4).toLowerCase() === "http"`

focusaurus:
I think making this all or nothing would be a pretty big constraint. It might be fine to require totally uniform handling of product media but I like that fact that it can be data-driven one media at a time which would facilitate a gradual migration to a 3rd party service like cloudinary for example.

If we want to optimize for performance I'd probably still want to be a touch more precise in the check so pseudo code

- slice off first 7 chars
- lowercase them
- check `startsWith("https://")` (first presuming it would be more common)
- check `startsWith("http://")`
  - Basically I don't want someone filing a bug that `"http.jpg"` isn't working as an image
- check `startsWith("//")` (probably a rare edge case)

aldeed:
@focusaurus Good points

src/core/util/getAbsoluteUrl.js
```diff
@@ -6,6 +6,12 @@
  * @returns {String} Full URL
  */
 export default function getAbsoluteUrl(rootUrl, path = "") {
+  // Check if the path is already absolute
+  const pathStart = path.slice(0, 7).toLowerCase();
+  if (pathStart.startsWith("https://") || pathStart.startsWith("http") || pathStart.startsWith("//")) {
```
aldeed: This should be `.slice(0, 8)`. Otherwise `https://` becomes `https:/` and will never pass the first check. I think the only reason your test is passing is because of `pathStart.startsWith("http")`, which should be `pathStart.startsWith("http://")` to prevent the `http.jpg` case that @focusaurus mentioned.

I suggest adding a test with `http.jpg` relative URL and make sure it is properly converted to absolute.

aldeed (changes requested):

aldeed (approved):

PR #6035, opened 2020-01-16
refactor: remove restriction that limits account to one group per shop

### LINKED ISSUE #5959: Remove one-user-per-group-per-shop constraint from addAccountToGroup mutation
## Issue Description
We need to remove the `one-group-per-user-per-shop` constraint found in `addAccountToGroup` found in https://github.com/reactioncommerce/reaction/blob/release-3.0.0/src/core-services/account/mutations/addAccountToGroup.js . The `one-group-per-user-per-shop` constraint  enforces that a user / account can only belong to one group per shop. This is a hangover from using `meteor` in `2.x`. Version `3.x` does not have such a constraint as we will eventually use  `reaction-authorization` which does not have such a constraint

## Possible Solution
remove the `one-group-per-user-per-shop` constraint found in `addAccountToGroup` mutation

kieckhafer:
Resolves #5959 
Impact: **minor**  
Type: **feature**

## Issue
Accounts are only allowed to belong to one Group per shop. This will impede how we envision groups working in the future.

## Solution
- Update the existing `addAccountToGroup` mutation to no longer remove the account from the previous group when the account is added to a new group.
- Create a new `removeAccountFromGroup` mutation that will remove users from a group.
- Update `canAddAccountToGroup` helper to use `reaction:legacy:groups/manage:accounts` permission as the main check. Previously, it was comparing the `roles` of the account doing the removing, with the `roles` of the account being removed, and they had to be equal. This didn't make sense in the past because that means anyone in your group could remove you from the group. A "customer" could remove a "customer" from the group since they shared permissions. Additionally, with these changes, it didn't make sense because a user can be in multiple groups, meaning User A could have 1000 permissions, User B could only have 1 permission, which User A and B shared, and yet User A would be denied access because the `roles` didn't match exactly.

## Breaking changes
None

## Testing
1. Create two users: an admin and a customer
2. As an admin, add the customer account to the two non-admin groups (guest, customer) from the same shop (this will allow for permission testing later on)
1. See that the `Groups` array on the `account` has multiple IDs in it
1. See that the `user.roles` is updated to include all the permissions these two groups offer, with no duplicates (This will be refactored in #6031 but is needed now).
1. Log in as the customer, and attempt to add your own account to an admin account. See that you are restricted.

### DIFF
--- a/src/core-services/account/mutations/addAccountToGroup.js
+++ b/src/core-services/account/mutations/addAccountToGroup.js
@@ -48,6 +48,9 @@ export default async function addAccountToGroup(context, input) {
   const accountUser = await users.findOne({ _id: account.userId });
   if (!accountUser) throw new ReactionError("not-found", "No user found with that ID");
 
+  const groupToAdd = account.groups.includes(groupId);
+  if (groupToAdd) throw new ReactionError("group-found", "Account is already in this group");
+
   // Get existing roles from a user
   const newAccountUserRoles = new Set((accountUser.roles || {})[shopId] || []);
 
--- a/src/core-services/account/mutations/addAccountToGroup.js
+++ b/src/core-services/account/mutations/addAccountToGroup.js
@@ -60,8 +60,7 @@ export default async function addAccountToGroup(context, input) {
   const uniqueRoles = _.uniq(newRoles);
 
   // Add all group roles to the user. Make sure this stays in this order.
-  const newAccountUserRolesArray = [...newAccountUserRoles];
-  await ensureRoles(context, newAccountUserRolesArray);
+  await ensureRoles(context, newRoles);
   await users.updateOne({
     _id: account.userId
   }, {
@@ -72,7 +71,7 @@ export default async function addAccountToGroup(context, input) {
   // TODO(pod-auth): this will likely be removed in #6031, where we no longer get roles from user.roles
 
   // Add new group to Account
-  const accountGroups = (Array.isArray(account.groups) && account.groups) || [];
+  const accountGroups = Array.isArray(account.groups) ? account.groups : [];
   accountGroups.push(groupId);
 
   await Accounts.updateOne({ _id: accountId }, { $set: { groups: accountGroups } });
--- a/src/core-services/account/mutations/addAccountToGroup.js
+++ b/src/core-services/account/mutations/addAccountToGroup.js
@@ -1,3 +1,4 @@
+import _ from "lodash";
 import ReactionError from "@reactioncommerce/reaction-error";
 import SimpleSchema from "simpl-schema";
 import canAddAccountToGroup from "../util/canAddAccountToGroup.js";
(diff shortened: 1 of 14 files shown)


kieckhafer:
@aldeed and @mikemurray, I'm working on an integration test, and updating the existing test, but wanted to get the review on this going ASAP as it's the last remaining mutation that's needed for this weeks release. Hoping everything looks good and then we can merge as soon as the tests are ready.

src/core-services/account/mutations/addAccountToGroup.js
```diff
@@ -47,58 +48,38 @@ export default async function addAccountToGroup(context, input) {
...
-    newAccountUserRoles.add(newRole);
-  }
+  // Merge existing roles on user and roles from new group
+  // TODO(pod-auth): this will likely be removed in #6031, where we no longer get roles from user.roles
+  const newRoles = [...newAccountUserRoles, ...groupPermissions];
+  const uniqueRoles = _.uniq(newRoles);
 
+  // Add all group roles to the user. Make sure this stays in this order.
   const newAccountUserRolesArray = [...newAccountUserRoles];
   await ensureRoles(context, newAccountUserRolesArray);
```
aldeed: Seems like `const newAccountUserRolesArray =` line above should be deleted and `newRoles` should be passed into `ensureRoles` instead of `newAccountUserRolesArray`.

src/core-services/account/mutations/addAccountToGroup.js
```diff
@@ -47,58 +48,38 @@ export default async function addAccountToGroup(context, input) {
...
   }, {
     $set: {
-      [`roles.${shopId}`]: newAccountUserRolesArray
+      [`roles.${shopId}`]: uniqueRoles
     }
   });
+  // TODO(pod-auth): this will likely be removed in #6031, where we no longer get roles from user.roles
+
+  // Add new group to Account
+  const accountGroups = (Array.isArray(account.groups) && account.groups) || [];
```
aldeed: This would be clearer written as ternary

src/core-services/account/mutations/removeAccountFromGroup.js
```diff
@@ -0,0 +1,88 @@
...
+
+  const allGroupsUserBelongsTo = await Groups.find({ _id: { $in: account.groups } }).toArray();
+
+  // Get all other groups user belongs to, and all permissions from these groups, and flatten into single array
+  const remainingGroupsUserBelongsTo = allGroupsUserBelongsTo.filter((group) => group.shopId === shopId && group._id !== groupId);
+  const remainingRolesToGiveUser = _.uniq(remainingGroupsUserBelongsTo.map((group) => group.permissions).flat());
+
+  // update user to only have roles from other groups they belong
+  // TODO(pod-auth): this will likely be removed in #6031, where we no longer get roles from user.roles
+  await ensureRoles(context, remainingRolesToGiveUser || []);
```
aldeed: ` || []` unnecessary because we know for sure it's an array

src/core-services/account/mutations/removeAccountFromGroup.js
```diff
@@ -0,0 +1,88 @@
...
+  await users.updateOne({
+    _id: account.userId
+  }, {
+    $set: {
+      [`roles.${shopId}`]: remainingRolesToGiveUser || []
```
aldeed: ` || []` unnecessary because we know for sure it's an array

src/core-services/account/mutations/removeAccountFromGroup.js
```diff
@@ -0,0 +1,88 @@
...
+  };
+
+  await appEvents.emit("afterAccountUpdate", {
+    account: updatedAccount,
+    updatedBy: userId,
+    updatedFields: ["groups"]
+  });
+
+  // Return the group the account was added to
+  return Groups.findOne({ _id: groupId });
```
aldeed: You could use a `.find` on `allGroupsUserBelongsTo` to avoid another database call here. Also comment should say "removed from".

src/core-services/account/schemas/group.graphql
```diff
@@ -185,6 +197,16 @@ type UpdateAccountGroupPayload {
   group: Group
 }
 
+"The response from the `removeAccountFromGroup` mutation"
+type RemoveAccountFromGroupPayload {
+  "The same string you sent with the mutation params, for matching mutation calls with their responses"
+  clientMutationId: String
+
+  "The removed group"
+  group: Group
```
aldeed: Can add `!` to `Group` because we'll always return it if an error wasn't thrown.

aldeed (dismissed):
A few minor things

aldeed (approved):

PR #6046, opened 2020-01-21
Port displayPrice resolver from 3.0.0 branch

chrispotter:
Signed-off-by: Chris Potter <chris@reactioncommerce.com>

Impact: **minor**  
Type: **feature**

## Issue
Ports working resolver and test from 3.0.0 to 2.9.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,8 +40,8 @@ jobs:
           paths:
             - ~/.meteor
       - run:
-          name: Meteor NPM Install
-          command: meteor npm install
+          name: NPM Install
+          command: npm install
       - run:
           name: Build dist bundle
           command: meteor build --directory ../build/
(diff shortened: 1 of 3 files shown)


chrispotter:
not sure what's up with the tests here, any suggestions @aldeed or @focusaurus?

focusaurus:
Hmm, the `sharp` npm module is notoriously difficult to reliably build. The errors are not introduced by your change though. Let my try circleci with ssh and see if I can figure out what's going on.

kieckhafer:
@chrispotter i was running into the same issue in #6049 , I updated `sharp` in the `package.json` to the latest release which came out a bout a week ago, and that seemed to fix it. We are still seeing  what looks like some failures in the `meteor npm install` step, although it still passed CI so not 100% sure what the deal is. Try updating `sharp` to `^0.24.0` and see if that helps.

focusaurus:
Note as I understand our current requirements, this is not technically required in 2.x for us. I think it's pretty harmless and OK to merge nonetheless, but if anyone has reason to leave this in 3.x only that's fine too AFAIK.

aldeed:
@focusaurus @chrispotter Yeah I don't see why we'd need this in 2.x. Is it fixing any issue? Seems like it's only related to new catalog streaming that is 3.0 only. If so, I would vote to not merge since 2.x is supposed to be getting urgent fixes only.

chrispotter:
this was to deploy for testing purposes against an internal environment running 2.9, but since we've decided to upgrade this environment to 3.0.0-beta I think it's better to close. 👍

PR #6047, opened 2020-01-21
Fix running integration tests in Docker

aldeed:
Impact: **minor**  
Type: **feature**

## Issue
The `npm run test:integration` commands do not work in Docker.

## Solution
After trying various ways to get the `jest-mongodb`/`mongo-memory-server` packages to work in our Alpine Docker containers, it was clear that it isn't worth the effort. Separately, we had decided to move toward running integration tests against a full MongoDB server in a Docker image, to better reflect how things work when the application is actually running.

So now as of this PR:
- `jest-mongodb` package is removed along with related Jest config code (JEST_MONGO=1)
- Integration tests (TestApp) uses `MONGO_URL` env variable to run tests. It still generates a new database with a random name for each test file, to prevent one test file impacting another. Now it additionally drops that database after the tests run.
- Because every file gets a fresh database and it is dropped when the tests finish, I removed all code that was deleting specific collections, to make the tests run faster.
- Updated CircleCI config to spin up a MongoDB container with port 27017 mapped to the container that runs integration tests
- Added steps for running integration tests to readme

With tests working in Docker, there were some legit failures that we hadn't been seeing. So then to fix those:
- Using an ENV variable check, we no longer register the `files` plugin when running integration tests. This avoids some issues with its use of MongoDB change streams.
- Fixed a few places in the code and in tests that didn't work when `files` plugin is disabled. (Checks to ensure `Media` collection exists, etc.)
- Updated the three checkout integration tests to use `createShop` GraphQL mutation rather than calling it directly. Also moved all of the duplicated code for these tests into a single `checkoutTestsCommon.js` file.
- A few other fixes to `await` event emission and such, to cut down on the number of errors printed when running tests

## Breaking changes
None

## Testing

```sh
docker-compose run --rm api npm run test:integration # Test all mutations and queries
docker-compose run --rm api npm run test:integration:query # OR test queries only
docker-compose run --rm api npm run test:integration:mutation # OR test mutations only
docker-compose run --rm api npm run test:integration:file:watch -- <filename> # OR test one file
```


janus-reith:
Somehow I get an error about this:

```
if (!process.env.REACTION_DISABLE_FILES_PLUGIN) {
    const registerFilesPlugin = import("./core-services/files/index.js");
    await registerFilesPlugin(app);
 }
```

```
09:36:55.307Z ERROR Reaction: registerFilesPlugin is not a function
  TypeError: registerFilesPlugin is not a function
      at registerPlugins (file:///usr/local/src/app/src/registerPlugins.js:96:11)
      at async runApp (file:///usr/local/src/app/src/index.js:26:3)

```
when `REACTION_DISABLE_FILES_PLUGIN` is not set.
The import should probably stay on top.

Also, I still had same errors about mongo timeouts during my tests, need to investigate why they happened.

aldeed:
@janus-reith Thanks, actually I may not have tested that after adding it. Unfortunately I don't think the import can stay on top because then it loads the code that was causing test failures. I'll see if I can find a solution.

aldeed:
@janus-reith @willopez The "registerFilesPlugin is not a function" error on startup is fixed now. I just had my ESM dynamic import syntax wrong.

janus-reith:
Interesting, I've never seen such a default destructure - Really need to catch up more on ES Modules soon.

aldeed:
@janus-reith https://nodejs.org/dist/latest-v12.x/docs/api/esm.html is the most accurate/primary resource I've found on the subject, but even it does not mention exactly how to do dynamic imports of default.

willopez (approved):
LGTM

willopez (approved):
LGTM

PR #6053, opened 2020-01-24
[3.0.0] Various product GQL fixes

### LINKED ISSUE #6034: [3.0.0] AccountInvites are case-sensitive
## Issue Description
Currently, the Account Invites are case-sensitive - Is this the intended beahviour?
e.g. when inviting a user as Test.User@domain.com, if the user then signs up as test.user@domain.com the Invite would stay active and the user is signed up with the customer group.

## Steps to Reproduce
1. Invite some user, e.g. Test.User@domain.com as shop manager
2. Recieve invite info(no customized link here), sign up with any letter in a different case.
3. Notice that you are unable to use shop manager privileges, like using the admin.
(4). Check you user group in the db, check AccountTables and verify that the original invite to Test.User@domain.com which would be deleted after usage is still present.

## Possible Solution
Maybe make both creating the invites and the check for invites on creating accounts transform the email to lowercase?

### Versions
latest release-3.0.0

### LINKED ISSUE #6044: Add default sort for products API as createdAt ascending
Submitting this here per @nnnnat 

## Issue Description
For the admin when listing Products (not consumer-side Catalog records), the list is more or less random. The more intuitive way is to list them by `createdAt` ascending so that the last created records show up first. So this should be added as the default sort when making a `getProducts` API call.

### Versions
3.0.0

aldeed:
Resolves #6034
Resolves #6044 
Impact: **minor**  
Type: **bugfix**

## Fixes
- Refactored `createProduct`, `createProductVariant`, `updateProduct`, and `updateProductVariant` mutations, fixing and optimizing various things
- The arguments to "afterVariantUpdate" event were not being consistently emitted and used. Now they are. Also, it's now called once with `fields` rather than many times with `field`.
- Removed unneeded stuff (label, defaultValue, etc.) from product and variant SimpleSchemas
- Fix GQL/input schemas for `updateProduct` and `updateProductVariant` so that you can't update things like `_id` and `ancestors` that would cause errors or be insecure.
- Fix handling of email addresses when inviting emails that don't already have an account. The case in which an admin enters the invite email no longer has to match the case in which the invitee enters it when registering. #6034
- Update `products` GQL query to have a default sort of `createdAt` descending. Ensure that we're always setting `createdAt` and `updatedAt` properly everywhere. #6044 

## Breaking changes
The refactor of the "afterVariantUpdate" emits is technically breaking, but it was previously broken, so now it's double broken, which equals fixed.

## Testing
- Review code of the refactored mutations and test them
- Test inviting with different email case (see steps in #6034)
- Test various sortOrder and sortBy params for `products` GQL query


mikemurray (approved):
👍

Issue #6055, opened 2020-01-27
[3.0.0] Implement proper account subscription to "new order" emails


aldeed:
https://github.com/reactioncommerce/reaction/commit/2098d97095527091104597d47bdccc4cfcd76089 removes code that would automatically send emails to certain "admin" roles when a new order came in. With changes to authorization in 3.0, this was no longer a simple task, and it wasn't the correct way to determine recipients in the first place.

This issue is to re-implement email notifications to non-customer accounts based on some way for them to "subscribe" to receiving them (and unsubscribe).

PR #6056, opened 2020-01-27
Remove most remaining uses of users collection

aldeed:
Impact: **minor**  
Type: **feature**

## Issue
The "users" collection is the domain of the identity service, which is no longer part of the API in 3.0. But it is still being used in several places in API code.

## Solution
I've removed most uses of the "users" collection where possible. For a couple mutations in which the identity service currently calls the API, I've moved the users-specific things to live in the "authentication" plugin, which is also specific to the default identity service.

Big picture, this allows you to swap out the Reaction Identity service for some other compatible identity service by creating your own custom plugin to replace the default "authentication" plugin. (In theory. We haven't yet tested any in particular.)

## Breaking changes
Because we have no other solution ready, we've removed the sending of emails to "admins" when a new order comes in. Issue https://github.com/reactioncommerce/reaction/issues/6055 is tracking re-implementation of this in a better way.

`addAccountEmailRecord` and `removeAccountEmailRecord` still modify the Account emails list but no longer modify the User emails list. Changing User email list can be implemented on Reaction Identity if needed.

## Testing
Verify that the affected mutations/flows work.
- Email verification doesn't currently work anyway, so can't test that one yet
- Test reset password by clicking the button from the login screen
- Verify `addAccountEmailRecord` and `removeAccountEmailRecord` GQL mutations still modify the Account emails list properly.


kieckhafer (changes requested):
I'm seeing an odd error message when using the `addAccountEmailRecord` and `removeAccountEmailRecord` email mutations.

When I `addAccountEmailRecord`, the email gets added to my `account`, however I get an error message that says `"No unverified email found"`.

The same thing happens when I `removeAccountEmailRecord`, the email is removed, but I get the same error message.

When checking in the collection, `verified = false`, so this error message doesn't seem to make sense to be thrown here.

In addition, when I add an email, the new email gets a `provides: default` added onto the object... in fact every new email I add has `provides:default` attached to it, so If I add 10 emails, I have 10 emails that all `provides: default`

Robo_3T_-_1_3

I'd expect only one email address to have `provides: default` attached to it, and I would also expect that to be the email I signed up with, not the newly added one, unless I explicitly changed my default.

aldeed:
@kieckhafer Regarding `provides:default`, I don't think I changed that and I wasn't sure what made sense there.

How would you propose that we handle it? Do you think we should just add them without any `provides` value, and then maybe make a different mutation for `setDefaultAccountEmail`, which could be used to switch which one is default?

What about the case where there isn't already a default one and I call `addAccountEmailRecord`? Should it then be assumed to be default? Or make them explicitly call `setDefaultAccountEmail` after adding it?

Maybe we should prevent removal of the default in the first place? You have to make another one default before removing that one?

I fixed the "No unverified email found" error by just removing the email validation calls. Those were looking at "users" whereas this is doing "accounts". We can add a validation flow for account emails in a separate PR later.

kieckhafer:
@aldeed I think you've mentioned a combo of what's probably the best flow here.

1. Ideally, when you sign up with your email address, that address is set to `provides: default`
1. When you add a new email, if there is already an email that provides default, then the new one doesn't get this field. If there is not already an email that `provides: default`, then set it for this new email
1. You cannot delete the default email. If you want to delete your email address, it needs to be the non-default one, or you need to make another one default first
1. There probably should be a new `setAccountDefaultEmail` mutation to do this

I created a PR (into your `feat-remove-users-collection` branch) which takes care of #'s 2, 3, and 4: https://github.com/reactioncommerce/reaction/pull/6062

I think this PR will be good to go once #6062 is merged into it, so if you want to take a look at my work there, merge it in here if it looks good, then this one should be mergeable.

kieckhafer (approved):

kieckhafer:
Approved, pending getting the changes from #6062 merged into this branch.

PR #6040, opened 2020-01-18
Fix add missing permissions check to 

willopez:
Resolves #6038  
Impact: minor 
Type: fix

## Issue
Permissions checks were missing in `addTagsToProducts` and `removeTagsFromProducts` mutations. This permitted anyone to execute these mutations, which is incorrect.

## Solution
Added permission checks and also added the `shopId` param, which was also missing.

## Breaking changes
None

## Mutations 
Mutation below are adding/removing products to the product with id `2djnCs2n4wRfac22C` from the sample data set.

```
mutation addTagsToProducts{
  addTagsToProducts(input: {
   shopId: "cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg==",
    productIds: [
      "cmVhY3Rpb24vcHJvZHVjdDoyZGpuQ3MybjR3UmZhYzIyQw==",
      "cmVhY3Rpb24vcHJvZHVjdDo2OWNoa3Vnejg0OEszQXNZSw=="
    ],
    tagIds: ["cmVhY3Rpb24vdGFnOjRqZXZ0QkdYTHoyQkY4TGY4",
    "cmVhY3Rpb24vdGFnOjU5UFBoWEV0WTZIdDZFQllX"]
  }){
    updatedCount
  }
}

mutation removeTagsFromProducts{
  removeTagsFromProducts(input: {
   shopId: "cmVhY3Rpb24vc2hvcDpKOEJocTN1VHRkZ3daeDNyeg==",
    productIds: [
      "cmVhY3Rpb24vcHJvZHVjdDoyZGpuQ3MybjR3UmZhYzIyQw==",
      "cmVhY3Rpb24vcHJvZHVjdDo2OWNoa3Vnejg0OEszQXNZSw=="
    ],
    tagIds: ["cmVhY3Rpb24vdGFnOjRqZXZ0QkdYTHoyQkY4TGY4",
    "cmVhY3Rpb24vdGFnOjU5UFBoWEV0WTZIdDZFQllX"]
  }){
    updatedCount
  }
}
```

## Testing
1. Start API
2. Execute mutations above without an authorization token
3. Verify that a permissions error is thrown 
4. Add authorization token to the requests
5. Verify tag are add or removed correctly

PR #6057, opened 2020-01-27
release v3.0.0-beta.2

kieckhafer:
# v3.0.0-beta.2

This is the second beta release of our new Reaction project.

*Reaction releases will no longer be coordinated across all projects - we'll release each project, independently, as needed. This means version numbers will no longer be in sync. The newest versions of each project will work together. This change has two exceptions: we will release all projects in coordination for a `beta` release, and all projects in coordination for the official `v3.0.0` release.*

### Permissions are checked against groups, not users

Permissions are now checked by querying a `group` that user belongs to, and checking against `permissions` that the `group` has. In the past, each `user` had `roles` (renamed permissions) attached to it, and the roles were checked. This will help keep permissions in sync via groups, instead of individually updating all users every time a permission changes or a user is added to a group.

## Fixes

- fix: add missing permissions check to [#6040](https://github.com/reactioncommerce/reaction/pull/6040)

## Refactor

- refactor: get roles / permissions directly from group(s), not user object [#6031](https://github.com/reactioncommerce/reaction/pull/6031)

## Docs

- docs: update links to use trunk branch of docs [#6050](https://github.com/reactioncommerce/reaction/pull/6050)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,27 @@
+# v3.0.0-beta.2
+
+This is the second beta release of our new Reaction project.
+
+*Reaction releases will no longer be coordinated across all projects - we'll release each project, independently, as needed. This means version numbers will no longer be in sync. The newest versions of each project will work together. This change has two exceptions: we will release all projects in coordination for a `beta` release, and all projects in coordination for the official `v3.0.0` release.*
+
+## Notable changes
+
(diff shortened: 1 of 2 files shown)


rosshadden (approved):

manueldelreal (approved):
:shipit:

PR #6060, opened 2020-01-28
Fix problems with accounts queries

### LINKED ISSUE #5906: Account Groups is not returned in account GQL query
## Issue Description
When running the GQL below, the `groups` attribute on the `account` is `null` even when the account has groups  

## Steps to Reproduce

1. You must be logged in to run this query i.e. use `meteor-login-token` http header to provide the logged in token

2. Run the query
```
query {
  account(id: "SOME_OPAQUE_ACCOUNT_ID") {
    _id
    groups {
      nodes {
        _id
        description
        name
        permissions
      }
    }
  }
}
```

### LINKED ISSUE #6006: [3.0.0] Delete administrators queries

### LINKED ISSUE #6017: [3.0.0] Accounts query has duplicate field shopId
The `accounts` query includes the field `shopId` twice with different description:

https://github.com/reactioncommerce/reaction/blob/8ed0165a7ab2c07ddfa394317d12c34ca57c37ee/src/core-services/account/schemas/account.graphql#L444
https://github.com/reactioncommerce/reaction/blob/8ed0165a7ab2c07ddfa394317d12c34ca57c37ee/src/core-services/account/schemas/account.graphql#L447

### Versions
latest 3.0

aldeed:
Resolves #6006 
Resolves #5906 
Resolves #6017   
Impact: **minor**  
Type: **feature|bugfix**

## Changes
- Rather than `accounts` GQL query returning only non-admin accounts and requiring `shopId` param for a filter, it now returns all accounts and accepts an optional `groupIds` array for filtering. You can't filter by `shopId` because this no longer makes sense; accounts do not belong to a particular shop in 3.0.
- The above change automatically fixed #6017 due to removing both `shopId` params
- `Account.groups` was always `null` when requested through GraphQL. Now it's correct.
- Removed any code where we were using `account.shopId`, which isn't usually set anymore
- I also updated ESLint pkg version in order to get support for `.cjs` extension on `.eslintrc` file, and then changed that extension. This silences a warning that was showing when running lint command.

## Breaking changes
The `accounts` GQL query takes different params and returns different results as described above.

## Testing
Run GQL queries to verify the changes. I've updated the relevant integration tests, too.


kieckhafer (approved):
Looks good overall.

Came across one issue that seems to maybe stem from the `api-utils`, not this PR. [I've created a ticket for that issue](https://github.com/reactioncommerce/api-utils/issues/15), and also [made a change in this PR](https://github.com/reactioncommerce/reaction/pull/6060/commits/240649f1c120ecc6799fed9bad8c0cd13d90e310) to fix the issue for now.

PR #6063, opened 2020-01-29
refactor: remove "owner" and other default permissions / groups

### LINKED ISSUE #6030: Revisit owner role and migrations for permissions
This comment needs to be addressed, but we wanted to merge the [original ticket ](https://github.com/reactioncommerce/reaction/pull/6021)since it was so big and not have it handing around: 

https://github.com/reactioncommerce/reaction/pull/6021#issuecomment-574306926

@kieckhafer @rosshadden A few things I can think of at the moment:
- https://github.com/reactioncommerce/reaction-identity/pull/14/files will need to be changed somehow. Probably just global shop:create is needed there for the first user? (We'll also need similar code in Identity for keto auth for the first user, but that's a separate issue.)
- I vote to use this opportunity to remove the `"owners" should always have access` logic. As long as we give the first user `shop:create` and anything else they might need, we don't need the concept of an "owner" permission since "owner" is more of a role. This might expose a bunch of faulty permission checks that were previously hidden, but that's a good thing we can work through before the final release.
- What do you think about adding a just-in-time migration for this? In the legacy-authorization `hasPermission` function, you can check whether `roles` contains any old roles. Then use an old-to-new map to build a new roles list, and do a `users.updateOne` to "upgrade" that user. Then check the new list.

kieckhafer:
Resolves #6030   
Impact: **breaking|major**  
Type: **feature|refactor**

## Issue
We'd like to move away from the all encompassing "owner" role.

## Solution
- remove the code which adds users to an "owner" role when creating an account 
- remove code which adds user to "customer" role when creating a non-owner account, users no longer are required to be in a group
- add code which adds first use to `system-manager` and `accounts-manager` _**GLOBAL**_ groups, and also creates those groups if they do not already exists
- removes default creation of `guest` and `customer` groups
- remove restriction of inviting user to owner group (meaning we can have more than one owner)
- remove `isOwner` logic from `canAddAccountToGroup`, and leave only policy validation
- remove `owner` "override" in `hasPermissions`
- remove any code related to `owner` permissions that should no longer be there

## Breaking changes
- `owner` is no longer a permission. Since this permission was basically a key to do anything in the app, if you had custom code that was passing `owner` for permission checks, you'll need to update to make sure the user has more specific permissions

## Testing
- make sure all places where "owner" was being used as the permission still work with more specific permissions


kieckhafer:
@aldeed overall I think this is good for a look. Two places I wasn't 100% sure what to do:

1. I don't think anyone is going to have anonymous anymore, can we remove this? https://github.com/reactioncommerce/reaction/blob/refactor-removeOwner/src/core-services/cart/mutations/reconcileCarts.js#L48:L48
1. I think you mentioned in another PR we might not even use this anymore? Can this be removed? https://github.com/reactioncommerce/reaction/blob/refactor-removeOwner/src/plugins/notifications/util/sendNewOrderNotifications.js#L57:L57

aldeed:
@kieckhafer Both of those code areas will be gone if you merge in `release-3.0.0`. I had removed them in https://github.com/reactioncommerce/reaction/pull/6056, which is merged now.

src/core-services/account/mutations/createAccount.js
```diff
@@ -76,42 +77,89 @@ export default async function createAccount(context, input) {
...
-  // The identity provider service gives the first created user the global "owner" role. When we
-  // create an account for this user, they should be assigned to the "owner" group.
-  if (authUserId === userId) {
-    const isGlobalOwner = await context.userHasPermission("reaction:legacy:shops", "owner", { shopId }); // TODO(pod-auth): update this permissions check
-    if (isGlobalOwner) groupSlug = "owner";
+  let groups = [];
+
+  // if this is the first user created overall, add them to the
+  // `system-manager` and `accounts-manager` groups
+  const appAccounts = await Accounts.find({}).toArray();
```
aldeed: This would definitely be a very slow query and probably crash after a certain number of accounts. I've found the fastest way is:

```js
const anyAccount = await Accounts.findOne();
if (!anyAccount) {
}
```

There is also `find().count()` but we don't need an accurate count in this case and supposedly count can sometimes be very slow, too.

src/core-services/account/mutations/createAccount.js
```diff
@@ -76,42 +77,89 @@ export default async function createAccount(context, input) {
...
+  if (appAccounts.length === 0) {
+    // get IDs of `system-manager` and `account-manager` groups
+    const systemManagerGroup = await Groups.findOne({ slug: "system-manager" });
```
aldeed: On this and the accounts manager one, you could add `{ projection: { _id: 1 } }` as the second arg to optimize a bit.

src/core-services/account/mutations/createAccount.js
```diff
@@ -76,42 +77,89 @@ export default async function createAccount(context, input) {
...
+          "reaction:legacy:accounts/update:language",
+          "reaction:legacy:accounts/read",
+          "reaction:legacy:accounts/read:admin-accounts",
+          "reaction:legacy:accounts/create",
+          "reaction:legacy:accounts/read",
+          "reaction:legacy:accounts/read:admin-accounts",
+          "reaction:legacy:shops/create"
+        ],
+        shopId: null
+      });
```
aldeed: Given the amount of code here and the fact that we perhaps will need to do this from other places, I'd create two util functions, `createSystemManagerGroup` and `createAccountsManagerGroup`, and call them here.

aldeed: Or perhaps `ensureSystemManagerGroup`, which would also do the check for existence and return the ID either way.

src/core-services/account/mutations/createAccount.js
```diff
@@ -76,42 +77,89 @@ export default async function createAccount(context, input) {
...
+    }
+    groups.push(systemManagerGroupId);
+
+    const accountsManagerGroup = await Groups.findOne({ slug: "system-manager" });
```
aldeed: Should be "accounts-manager" slug.

src/core-services/account/mutations/createAccount.js
```diff
@@ -76,42 +77,89 @@ export default async function createAccount(context, input) {
...
+      });
+    }
+    groups.push(accountsManagerGroupId);
   }
 
-  // If we didn't already upgrade them to the "owner" group, see if they're been invited to any groups
-  if (groupSlug === "customer") {
+  // if this isn't the first account, check to see if this user was invited to a group
+  let invites;
+  if (appAccounts) {
```
aldeed: Just use `else`?

src/core-services/account/util/canAddAccountToGroup.js
```diff
@@ -17,18 +15,9 @@ export default async function canAddAccountToGroup(context, group) {
...
-  // are able to add any user to any group, regardless of other permissions.
-  const ownerGroup = await Groups.findOne({ name: "owner", shopId });
-  const isOwnerAccount = (
-    !!ownerGroup && contextUserAccount && contextUserAccount.groups.includes(ownerGroup._id)) ||
-    userHasPermission("reaction:legacy:shops", "owner", { shopId }); // TODO(pod-auth): update this to figure out what to do with "owner"
-
-  const hasPermissionToAddAccountToGroup = await isOwnerAccount || await userHasPermission("reaction:legacy:groups", "manage:accounts", { shopId });
+  // An account can add another account to a group as long as the account adding
+  // has `reaction:legacy:groups/manage:accounts` permissions
+  const hasPermissionToAddAccountToGroup = await userHasPermission("reaction:legacy:groups", "manage:accounts", { shopId });
```
aldeed: At this point, should we just delete this fn and pass these same args to `validationPermissions` in the two fns that call this one? More of the standard pattern.

src/core-services/account/util/defaultRoles.js
```diff
@@ -1,27 +1,4 @@
...
-  "guest", // legacy role
-  "index", // legacy role
-  "product", // legacy role (unused)
-  "tag", // legacy role (unused)
-  "reaction:legacy:products/read",
-  "reaction:legacy:tags/read"
-];
-
 export const defaultShopManagerRoles = [
-  ...defaultCustomerRoles,
```
aldeed, line 24: I'd say delete `reaction:legacy:shops/owner` and `owner` from the `defaultOwnerRoles` list. Just accounts mgmt plus shop create for now.

aldeed (changes requested):
@kieckhafer Looks like you hit everything in the list here. I left a few suggestions, and I'll do some more testing before we merge this.

kieckhafer:
@aldeed addressed all your comments, and made a few more changes regarding global permission in `permissionsByUserId`

aldeed (approved):

PR #6064, opened 2020-01-30
Implement `canSeeAdminUI` check

trojanh:

Resolves https://github.com/reactioncommerce/reaction-admin/issues/192 
Impact: **breaking**  
Type: **feature**

## Issue
Currently the admin UI always shows but for non-admins, data will be missing and some things will be broken due to API authorization checks.

## Solution
Add a global setting canSeeAdminUI in the API, and in Reaction Admin:

Add a way to toggle it (boolean) from settings
Hide the UI when it's false. Show a nice message, perhaps linking to the storefront.

## Breaking changes
none

## Testing
1. Open GraphQL playground
2. Check if `canSeeAdminUI` field is visible in `GlobalSettings` query and `updateGlobalSettings` mutation.


trojanh:
@aldeed let me know if this suffices the needs of the issue.

aldeed:
@trojanh I'm sorry I was not thinking straight when I told you how to implement this. `canSeeAdminUI` will need to be a per-account setting, so the global settings API won't work.

I think we should maybe figure out a little broader solution here that can also work for hiding and showing individual pages and menus of the admin UI. I will think about it a bit and then put more details in the issue.

trojanh:
sure @aldeed , I was wondering `globalSettings` were meant for all the accounts and wasn't tied specifically to some users to toggle `canSeeAdmin` for an account, so this answers it all. Thanks

Issue #6061, opened 2020-01-28
Update Babel and rename babel.config.js to babel.config.cjs


aldeed:
The following is logged whenever Babel is involved (running tests mostly):

```
(node:39890) Warning: require() of ES modules is not supported.require() of babel.config.js from node_modules/@babel/core/lib/config/files/configuration.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which de
fines all .js files in that package scope as ES modules.Instead rename babel.config.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from package.json.
```

Babel 7.7.0 adds support for `.cjs` extension: https://github.com/babel/babel/blob/master/CHANGELOG.md#v770-2019-11-05

So the tasks:
- Update Babel and all related packages to 7.7.0, or as high as we can go without breaking things
- Rename `babel.config.js` -> `babel.config.cjs` to avoid this warning

PR #6065, opened 2020-01-30
chore: update Babel

trojanh:

Resolves https://github.com/reactioncommerce/reaction/issues/6061
Impact: **minor**  
Type: **chore**

## Issue
The following is logged whenever Babel is involved (running tests mostly):
```
(node:39890) Warning: require() of ES modules is not supported.require() of babel.config.js from node_modules/@babel/core/lib/config/files/configuration.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which de
fines all .js files in that package scope as ES modules.Instead rename babel.config.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from package.json.
```

## Solution
Babel 7.7.0 adds support for .cjs extension: https://github.com/babel/babel/blob/master/CHANGELOG.md#v770-2019-11-05

So the tasks:

Update Babel and all related packages to 7.7.0, or as high as we can go without breaking things
Rename babel.config.js -> babel.config.cjs to avoid this warning

## Breaking changes
none

## Testing
Run the tests and verify if the warning is gone,


aldeed (approved):

aldeed:
Thanks for doing this one @trojanh !

PR #6072, opened 2020-02-01
fix: fix adding shop creator to owner group

aldeed:
Impact: **minor**  
Type: **bugfix**

## Issue
After a shop is created, shop account groups are created and the account that created it is added as a shop owner, so that they can then access things about that shop. This was being done with incorrect permissions and failing, thus resulting in not being able to do anything with a shop (query, mutate, whatever) after creating it.

## Solution
Updated to call the account group mutations with internal context so they don't fail.

## Breaking changes
None

## Testing
1. Start platform on a clean database
2. Go to http://localhost:4080 (admin), and register a new user.
3. Enter a shop name and create a shop.
4. Go to other pages and verify that tables load, updating settings works, etc.

### DIFF
--- a/src/core-services/shop/mutations/createShop.js
+++ b/src/core-services/shop/mutations/createShop.js
@@ -125,10 +125,14 @@ export default async function createShop(context, input) {
 
   try {
     // Create account groups for the new shop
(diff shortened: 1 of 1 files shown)


kieckhafer (approved):
👍

PR #6075, opened 2020-02-02
fix: fix order email data

### LINKED ISSUE #5849: OrderEmail missing fields
##  Prerequisites
* [ x ] Are you running the latest version?
* [ x ] Are you able to consistently reproduce the issue?
* [ x ] Did you search the issue queue for existing issue? 

## Issue Description
When creating an order from the Storefront as a customer I receive an email without the fullname of the customer.

## Steps to Reproduce
1. Add item to cart
2. Complete checkout procedure
3. Open email in mailbox

## Possible Solution
I think there might be two options to solve this:

1. (Most preferred solution). Currently you can create a plugin with the function 'getDataForOrderEmail'. When it is registered, it will be at the bottom of the registered functions. When the email is generated, it will select the first registered function of 'getDataForOrderEmail'. I would like to have the option to register the function 'getDataForOrderEmail' at the top of the list or have the core select the last registered function.

2. Extend the function 'getDataForOrderEmail' with custom fields so I can create a more complete and custom email message.  

### Versions
```2.9.0```

aldeed:
Resolves #5849   
Impact: **minor**  
Type: **bugfix**

Order email template data fixes

## Issue
- If a plugin registered an additional function of type `getDataForOrderEmail`, it was ignored because the one in the `orders` plugin always took precedence.
- The default `getDataForOrderEmail` function in the `orders` plugin does not include `account` data when the order is linked with an account.
- The default `getDataForOrderEmail` function in the `orders` plugin does not include `fullName` field with any of the address objects.
- The default order email templates do not show the name line for billing or shipping address
- The default order email templates show some text that says "undefined" if you haven't added a shop address yet.
- The default order email templates do not properly show the variant title for each item. They show only the product title.

## Solution
- If a plugin registers an additional function of type `getDataForOrderEmail`, the object that it returns is combined with the default object. There can still be some uncertainty about which plugin's function will run first, but as long as you return an object with unique keys, you can rely on them in your email templates.
- The default `getDataForOrderEmail` function in the `orders` plugin now includes `account` data when the order is linked with an account. For example, something like this will work for an email subject line:
    ```
    Your order is confirmed, {{#if account.profile.firstName}}{{account.profile.firstName}}{{else if account.profile.name}}{{account.profile.name}}{{else}}{{order.email}}{{/if}} - {{order.referenceId}}
    ```
- The default `getDataForOrderEmail` function in the `orders` plugin now includes `fullName` field with all of the address objects.
- The default order email templates now show the name line for billing and shipping address
- The default order email templates no longer contain "undefined" text if you haven't added a shop address yet.
- The default order email templates now properly show the variant title for each item.

## Breaking changes
None

## Testing
1. Use a fresh database and create a new shop so that the default email templates get used.
2. Set up emailing, ensure your shop does not have any addresses listed, and then place an order while logged in to the storefront.
3. Verify everything looks good in the new order email you receive.

### DIFF
--- a/src/core-services/orders/util/getDataForOrderEmail.js
+++ b/src/core-services/orders/util/getDataForOrderEmail.js
@@ -32,10 +32,16 @@ function formatDateForEmail(date) {
  */
 export default async function getDataForOrderEmailDefault(context, { order }) {
   const { collections, getAbsoluteUrl } = context;
-  const { Shops } = collections;
+  const { Accounts, Shops } = collections;
 
   // Get Shop information
   const shop = await Shops.findOne({ _id: order.shopId });
+
+  // Get Account information
+  let account = null;
+  if (order.accountId) {
+    account = await Accounts.findOne({ _id: order.accountId });
+  }
 
   // TODO need to make this fully support multiple fulfillment groups. Now it's just collapsing into one
   const amount = order.shipping.reduce((sum, group) => sum + group.invoice.total, 0);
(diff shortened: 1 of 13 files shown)


willopez (approved):
LGTM

PR #6076, opened 2020-02-02
feat: implement origin tax sourcing

### LINKED ISSUE #5166: Taxation Location field is not used

aldeed:
Resolves #5166  
Impact: **minor**  
Type: **feature**

## Changes
- It's now possible to set `sourcing` to "origin" with the `createTaxRate` and `updateTaxRate` mutations. Previously they always used "destination" as the sourcing.
- The tax-rates plugin respects origin sourcing when determining which tax jurisdictions apply to an order. This uses `order.originAddress`, which currently is always set to the first address in the shop address book. If the shop has no addresses, no origin-type tax rates will be applied.
- Allow `udateShop` to set/update the `addressName` field

## Breaking changes
None, except that if you have origin-type tax rates defined, they will now begin getting applied to orders. The only way you could have these defined is by creating them directly in the database or through your own custom process.

The `createTaxRate` and `updateTaxRate` mutations are backwards compatible because the `sourcing` field is optional and defaults to "destination" as before.

## Testing
1. Use `updateShop` mutation to set `addressBook` to an array with a single address in it.
2. Use `createTaxRate` and/or `updateTaxRate` mutations to create at least one tax rate rule where the address filter is based on "origin" sourcing. Make sure this will match the shop address you added in the previous step.
3. Place an order and ensure that your tax rate gets applied to it.

### DIFF
--- a/src/core-services/shop/mutations/updateShop.js
+++ b/src/core-services/shop/mutations/updateShop.js
@@ -56,7 +56,6 @@ const inputSchema = new SimpleSchema({
 });
 
 const complexSettings = [
-  "addressBook",
   "emails",
   "shopLogoUrls",
   "storefrontUrls"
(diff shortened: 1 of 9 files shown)


willopez (approved):
LGTM

PR #6082, opened 2020-02-04
refactor: allow all users with token to view order

### LINKED ISSUE #5882: Cannot view anonymous order when logged in.
Checkout as an anonymous customer, you get an order with an ID and token so you can visit the order complete page `/order/:id/:token` in your storefront.

When you visit this page not logged in, you see the order details, as the token is in the URL.

Now if you login, the order is not found.

It appears that once logged in, the token value is ignored and it will attempt to find the order from your account. The order resolver, when given a order ID and token, should use these to locate the order, and ignore who they are logged in with.  If it's presented with only an Order ID and no token, then it should try to find the order from the current account.

kieckhafer:
Resolves #5882   
Impact: **minor**  
Type: **refactor**

## Issue

> Checkout as an anonymous customer, you get an order with an ID and token so you can visit the order complete page /order/:id/:token in your storefront.
> 
> When you visit this page not logged in, you see the order details, as the token is in the URL.
> 
> Now if you login, the order is not found.
> 
> It appears that once logged in, the token value is ignored and it will attempt to find the order from your account. The order resolver, when given a order ID and token, should use these to locate the order, and ignore who they are logged in with. If it's presented with only an Order ID and no token, then it should try to find the order from the current account.

## Solution
Refactor check order to allow all users who have the order `token` to view the order.

## Breaking changes
None

## Testing
1. Create an order as a logged out user using `example-storefront`
1. See the order confirmation page, and copy the link to that page
1. Login as any user
1. Paste the link, see you can still see the order with the token URL

### DIFF
--- a/src/core-services/orders/util/getOrderQuery.js
+++ b/src/core-services/orders/util/getOrderQuery.js
@@ -1,46 +1,37 @@
 import hashToken from "@reactioncommerce/api-utils/hashToken.js";
-import ReactionError from "@reactioncommerce/reaction-error";
 
 /**
  * @name getOrderQuery
  * @method
  * @memberof Order/helpers
- * @summary Creates Order mongo selector based on user permissions
+ * @summary Queries for an order and returns it if user has correct permissions
  * @param {Object} context An object containing the per-request state
- * @param {Object} selector A mongo selector
+ * @param {Object} selector Order ID or Reference ID to query
  * @param {String} shopId Shop ID of the order
  * @param {String} token An anonymous order token, required if the order was placed without being logged in
- * @returns {Object} A mongo selector
+ * @returns {Object} An order object
  */
 export async function getOrderQuery(context, selector, shopId, token) {
-  const { accountId: contextAccountId } = context;
-  const newSelector = { ...selector, shopId };
+  const { collections } = context;
+
+  const order = await collections.Orders.findOne(selector);
 
   // If you have the hashed token, you don't need to pass a permission check
-  if (token) {
-    newSelector["anonymousAccessTokens.hashedToken"] = hashToken(token);
-    return newSelector;
+  if (token && order.anonymousAccessTokens.some((accessToken) => accessToken.hashedToken === hashToken(token))) {
+    return order;
   }
 
   // if you don't have the hashed token,
   // you must either have `reaction:legacy:orders/read` permissions,
   // or this must be your own order
-  const userHasPermission = await context.userHasPermission(
+  await context.validatePermissions(
     "reaction:legacy:orders",
     "read",
-    { shopId }
+    {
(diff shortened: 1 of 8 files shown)


src/core-services/orders/util/getOrderQuery.js
```diff
@@ -5,35 +5,38 @@ import ReactionError from "@reactioncommerce/reaction-error";
...
+  const { collections } = context;
+
+  const order = await collections.Orders.findOne(selector);
+
+  if (!order) {
+    throw new ReactionError("not-found", "Order not found");
+  }
+
+  // If you have the hashed token, you don't need to pass a permission check
+  if (token && order.anonymousAccessTokens.some((accessToken) => accessToken.hashedToken === hashToken(token))) {
```
aldeed, line 25: Do `const hashedToken = hashToken(token);` above this line and then use `hashedToken` to compare in `some` fn. Otherwise we could be hashing it dozens of times if they have lots of tokens.

aldeed: Also `anonymousAccessTokens` is optional so do `(order.anonymousAccessTokens || [])` or somehow guard against it being undefined.

src/core-services/orders/util/getOrderQuery.js
```diff
@@ -5,35 +5,38 @@ import ReactionError from "@reactioncommerce/reaction-error";
...
+  // if you don't have the hashed token,
+  // you must either have `reaction:legacy:orders/read` permissions,
+  // or this must be your own order
+  await context.validatePermissions(
     "reaction:legacy:orders",
     "read",
-    { shopId }
+    {
+      shopId,
+      owner: order.accountId
```
aldeed, line 37: Everywhere else `owner` is being passed the user ID. So you'll need to look up the account first and then do `owner: account.userId`

src/core-services/orders/util/getOrderQuery.js
```diff
@@ -5,35 +5,38 @@ import ReactionError from "@reactioncommerce/reaction-error";
...
- * @returns {Object} A mongo selector
+ * @returns {Object} An order object
  */
 export async function getOrderQuery(context, selector, shopId, token) {
-  const { accountId: contextAccountId } = context;
-  const newSelector = { ...selector, shopId };
-  const userHasPermission = await context.userHasPermission(
+  const { collections } = context;
+
+  const order = await collections.Orders.findOne(selector);
```
aldeed, line 18: You are going to need to keep the `const newSelector = { ...selector, shopId };` line. Otherwise I could pass a `shopId` where I have permissions and an order ID that belongs to a different shop, and I'd still be able to see it.

aldeed (changes requested):

PR #6085, opened 2020-02-05
fix: find catalog product regardless of visibility

### LINKED ISSUE #6077: xFormCartItems and xFormOrderItems should not rely on Catalog
## Issue Description
xFormCartItems is currently used to query a cart in the storefront. If that catalog Item can't be found as it has `isVisible: false`, `isDeleted: true `or is actually just not present in the db, the query will throw an error. A user would by now have no way of removing that defunct cart Item.

The xFormOrderItems is used in the admin to display orders, and behaves similar. If the catalog item is not present, the query wil fail and the admin crashes.

IMHO both functions should not have to rely on the catalog item at all, as all the relevant data is present in the respective order or cart item.

## Steps to Reproduce
Cart Item:
1. Put an item in the cart
2. Set it to invisible in the admin
3. Reload the strorefront, the cart badge will display items but the cart is shown as empty.

Order Item:
1. Complete an order
2. Set the item used for the order to invisible in the admin
3. Go to the orders admin page, click that order, page will crash.

## Possible Solution
The catalog Item is only used to get the correct media for the item and should not be relevant for the basic functionality of these queries. All potential updates to that item or additional data from outside of that object should not be a requirement.

Options could be:

- For now, simply omit the media if not present and return the item data without it.
- Not preferred but simple to plug in without messing up existing data: Directly search the Media collection for matching entries like it is done for the product admin.
- Put the media, or at lest the primaryImage directly on the cart/order item.

### Versions
release-3.0.0

aldeed:
Resolves #6077 
Impact: **minor**  
Type: **bugfix**

## Issue
Cart and order transformations fail when the catalog item related to a cart or order item has since been deleted or hidden.

## Solution
Remove deleted/hidden checks from the queries. Look up by product ID only. That way it will only fail if the document has been hard deleted, which can only be done outside of the official GraphQL API and should never be done outside of a development environment as a rule.

## Breaking changes
None

## Testing
See #6077 for reproduction steps and verify fixed.

### DIFF
--- a/src/core-services/cart/xforms/xformCartItems.js
+++ b/src/core-services/cart/xforms/xformCartItems.js
@@ -25,11 +25,10 @@ async function xformCatalogProductMedia(mediaItem, context) {
 /**
  * @param {Object} context - an object containing the per-request state
  * @param {Object[]} catalogItems Array of CatalogItem docs from the db
- * @param {Object[]} products Array of Product docs from the db
  * @param {Object} cartItem CartItem
  * @returns {Object} Same object with GraphQL-only props added
  */
-async function xformCartItem(context, catalogItems, products, cartItem) {
+async function xformCartItem(context, catalogItems, cartItem) {
   const { productId, variantId } = cartItem;
 
   const catalogItem = catalogItems.find((cItem) => cItem.product.productId === productId);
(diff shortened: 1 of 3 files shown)


src/core-services/cart/xforms/xformCartItems.js
```diff
@@ -76,26 +75,13 @@ async function xformCartItem(context, catalogItems, products, cartItem) {
...
-    "product.isDeleted": { $ne: true },
-    "isDeleted": { $ne: true }
-  }).toArray();
+  const catalogItems = await Catalog.find({ "product.productId": { $in: productIds } }).toArray();
 
-  const products = await Products.find({
-    ancestors: {
-      $in: productIds
-    }
-  }).toArray();
```
aldeed, line 96: `products` was no longer being used by anything so we can eliminate a db lookup here

src/core-services/orders/xforms/xformOrderItems.js
```diff
@@ -79,19 +79,18 @@ async function xformOrderItem(context, item, catalogItems) {
...
-    "product.isDeleted": { $ne: true },
-    "isDeleted": { $ne: true }
-  }).toArray();
+  const catalogItems = await Catalog.find({ "product.productId": { $in: productIds } }).toArray();
+
+  const xformedItems = await Promise.all(items.map((item) => xformOrderItem(context, item, catalogItems)));
+
+  for (const mutateItems of getFunctionsOfType("xformOrderItems")) {
+    await mutateItems(context, xformedItems); // eslint-disable-line no-await-in-loop
+  }
```
aldeed, line 93: Added this `xformOrderItems` functionsByType block to make this identical to `xformCartItems` fn.

kieckhafer (changes requested):
I am still seeing the same issues as described in #6077

- When I make an item in my cart invisible, I still see the badge with the number in the cart, but I can't see anything in the cart. I also can't add anything new in the cart.
- When i make an item in an order invisible, the order review page crashes when I try to view it.

kieckhafer (changes requested):
After further testing it seems that if the `product` is made `visible` / `invisible`, then this PR fixes the issue, and I can still see the item in the cart and I can see an order correctly, even after it's invisible.

However, if the `variant` or `option` that was ordered is made `invisible`, I still see the same errors.

aldeed:
@kieckhafer Oh, right! Let me see if there's a quick solution to that

PR #6096, opened 2020-02-07
feat: move legacy permissions into npm package

kieckhafer:
Impact: **major|minor** (minor code changes, major impact as we move plugins to npm)
Type: **feature**

This PR is the first of many which will move our internal plugins to be externally installed npm packages.

This PR removes the `legacy-authorization` plugin code from the API, and installs the `@reactioncommerce/plugin-simple-authorization` to perform the same functions.

## Breaking changes
None

## Testing
The authorization code is the same as it was before, it's just being installed from a different place, so you should see nothing different in the behavior of this PR and the `v3.0.0` release it was created from.
1. Start the API
1. Create a user
1. Create a shop
1. Click around and see that you have permission to do what you should be able to , as the first user
1. create a second user
1. see that you can't perform actions a "regular" user shouldn't be able to

Once this PR is merged, we should do a `3.1` release alongside [this PR ](https://github.com/reactioncommerce/api-migrations/pull/4)in `api-migrations`


mikemurray (approved):
👍

kieckhafer:
The license issue being called out by SNYK has been looked at, and we are OK to merge with the error being thrown.

PR #6100, opened 2020-02-12
feat: The Federation

### LINKED ISSUE #6099: Add Reaction to Federation
# Add Reaction 3.x to Federated Gateway
Use Apollo's [Federation](https://github.com/apollographql/apollo-server/tree/master/packages/apollo-federation) package to build a federated schema for the reaction api service.

mikemurray:
Resolves #6099 
Impact: **major**
Type: **feature**

## Issue

The Reaction API must be a participant in the (Apollo) federation gateway

## Solution

Follow steps as outlined by the Apollo docs to make Reaction a participant, which lead to...
- Build a federated schema
- Remove subscriptions (not compatible)
- Remove schema stitching (not compatible, and federation supersedes it)
- Fix all double-quotes in block comments and escaped double-quotes in all loaded `.graphql` files

## Breaking changes

- ~~GraphQL subscriptions will not work. It is not compatible with apollo-server 2.x. It may make a re-appearance in apollo-server 3.0, whenever that's released.~~

- ~~Schema stitching has been removed. It is not compatible with federation.~~

## Backwards compatibility over breaking changes

Reaction API will run like it always has with `GQL subscriptions` and `schema stitching` enabled by default.

If you want Apollo Federation support, then you will sacrifice the above two features. You'll have to enable federation and disable subscriptions to get the full federation support in the federated gateway.

```
REACTION_APOLLO_FEDERATION_ENABLED=true
REACTION_GRAPHQL_SUBSCRIPTIONS_ENABLED=false
```

> q: If subscriptions aren't supported by Apollo federation, then why do I need to disable them. Why don't you do it in code?

> a: Subscriptions aren't supported yet, but might be supported in `apollo-server@3.0`. Best to leave the flag as a simple yes/no without automatic overrides based on some other variable.

## Testing
**auth token might not yet be supported for requests through the gateway**

1. Ensure all tests pass, as you do
2. Start the API like normal and run your favorite queries/mutations
3. Stop the API and update your env vars like so
```
REACTION_APOLLO_FEDERATION_ENABLED=true
REACTION_GRAPHQL_SUBSCRIPTIONS_ENABLED=true
```
4. See this error when starting up `Error: Subscriptions are not supported with Apollo Federation. Set `REACTION_GRAPHQL_SUBSCRIPTIONS_ENABLED=false` to disable subscriptions.`
5. Update your env vars to this:
```
REACTION_APOLLO_FEDERATION_ENABLED=true
REACTION_GRAPHQL_SUBSCRIPTIONS_ENABLED=false
```
6. The app should start, test your favorite queries and mutation with the gateway.


mikemurray:
Put back into WIP to see if there are options to make this not a breaking change.

kieckhafer (approved):
This works as it's expected to.

API & Gateway work when the correct `env` vars are set.
API works as expected when no `env`s are set
API gives you a warning when `env`s are set incorrectly

One "issue", which is expected here, is that Authenticated queries and mutations do not work through the Gateway. They continue to work through the API directly even when Gateway is enabled, so I see this as OK to merge. @rosshadden thoughts on that?

rosshadden:
It's okay to merge. We will have to figure out authorization but it's not a blocker for merging this as it will not affect our existing workflow.

rosshadden (approved):
It's okay to merge. We will have to figure out authorization but it's not a blocker for merging this as it will not affect our existing workflow.

rosshadden:
I changed the "**breaking**" in the OP to match latest changes.

PR #6104, opened 2020-02-14
feat: add updateProductsVisibility mutation

### LINKED ISSUE #6102: add mutation to update multiple products visibility
## Prerequisites

## Issue Description
There is no mutation to update certain properties such as `isVisible` in bulk. Currently, in the products table admin we are sending on request per product when updating a product's visibility, which is inefficient and will not scale.

## Possible Solution
Add a `updateProductsVisibility([productIds], shopId, isVisible)` muation that will bulk update a collections of product's visibility

willopez:
Resolves #6102 
Impact: **minor**
Type: **feature**

## Summary
Adds `updateProductsVisibility` mutation that will update multiple products by setting their visibility property.

```
mutation{
  updateProductsVisibility(input: 
  {
    productIds: ["357"],
    shopId: "shopId",
    isVisible: false
  }){
    updatedcount
  }
}
```

## Breaking changes
None

## Testing
1. Use the mutation above to set the `isVisible` property on multiple products.


src/core-services/product/mutations/updateProductsVisibility.js
```diff
@@ -0,0 +1,46 @@
...
+ * @method updateProductsVisibility
+ * @summary Updates a set a products visibility
+ * @param {Object} context -  an object containing the per-request state
+ * @param {Object} input - Input arguments for the bulk operation
+ * @param {String[]} input.productIds - an array of Product IDs
+ * @param {String} input.shopId - the shop's id
+ * @param {String[]} input.isVisible - the desired visibility
+ * @return {Boolean} A boolean flag that indicates whether the operation completed successfully or not.
+ */
+export default async function addTagsToProducts(context, input) {
```
mikemurray: function name doesn't match file name

src/core-services/product/mutations/updateProductsVisibility.js
```diff
@@ -0,0 +1,46 @@
...
+  const totalProducts = productIds.length;
+  let success = false;
+
+  for (const _id of productIds) {
+    // TODO(pod-auth): figure out a better way to loop through this
+    // eslint-disable-next-line no-await-in-loop
+    await context.validatePermissions(`reaction:legacy:products:${_id}`, "update", { shopId });
+  }
+
+  // // Generate update statements
```
mikemurray: double `//`

mikemurray (changes requested):

src/core-services/product/mutations/updateProductsVisibility.js
```diff
@@ -0,0 +1,46 @@
...
+        }
+      }
+    }
+  }));
+
+  const results = await executeBulkOperation(Products, operations, totalProducts);
+
+  success = totalProducts === results.updatedCount;
+
+  return success;
```
mikemurray: This causes "success" that doesn't actually reflect whether or not the operation was successful. The updated count only counts things that could be updated. If you set the visibility to `false` and it's already `false` on one or more products, then `success` will `false` because not all of the products were updated.

I think success is determined whether or not the bulk operation itself ran. Perhaps reporting on how may documents were updated would work better.

mikemurray (dismissed):

mikemurray:
@willopez This needs an integration test as well.

src/core-services/product/mutations/updateProductsVisibility.js
willopez: I agree, now returning `updatedCount` to the client.

willopez:
ready for another review

mikemurray (approved):
👍

PR #6105, opened 2020-02-19
refactor: move system information plugin to npm

kieckhafer:
Impact: **minor**
Type: **refactor**

This PR removes the `system-info` plugin code from the API, and installs the `@reactioncommerce/plugin-system-information` to perform the same functions.

## Breaking changes
None

## Testing
The system information code is the same as it was before, it's just being installed from a different place, so you should see nothing different in the behavior of this PR and the `v3.1.0` release it was created from.
1. Start the API
1. Create a user
1. Create a shop
1. Use the `systemInformation` GraphQL query to query system information
```
{
systemInformation(shopId: "SHOP_ID") {
  plugins {
    version
    name
  }
  apiVersion
  mongoVersion {
    version
  }
}
}
```
1. See system information


package.json
```diff
@@ -25,9 +25,10 @@
     "@reactioncommerce/db-version-check": "^1.0.0",
     "@reactioncommerce/file-collections": "~0.7.0",
     "@reactioncommerce/file-collections-sa-gridfs": "~0.1.0",
-    "@reactioncommerce/plugin-simple-authorization": "~1.0.0",
     "@reactioncommerce/logger": "~1.1.2",
     "@reactioncommerce/nodemailer": "~5.0.5",
+    "@reactioncommerce/plugin-simple-authorization": "~1.0.0",
+    "@reactioncommerce/plugin-system-information": "^1.0.0",
```
willopez: is there any reason this particular plugin uses the `^` for its version. I believe the `~` should be used so that only patch versions are updated automatically.

willopez (changes requested):
Everything looks good, just a minor comment.

package.json
kieckhafer: No reason, when I `npm install`ed this is what it automatically did, I just didn't update. Changed now.

kieckhafer:
@willopez  comment addressed

willopez (approved):
LGTM 👍

PR #6108, opened 2020-02-20
refactor: replace internal authentication plugin with NPM installed plugin-authentication

kieckhafer:
Impact: **minor**
Type: **refactor**

This PR removes the `authentication` plugin code from the API, and installs the `@reactioncommerce/plugin-authentication` to perform the same functions.

## Breaking changes
None

## Testing
The authentication code is the same as it was before, it's just being installed from a different place, so you should see nothing different in the behavior of this PR and the `v3.1.0` release it was created from.
1. Start the API
1. Create a user
1. Create a shop
1. See that you can reset your password. The mutations here are meant to be internally called, so you need to reset through Identity.


focusaurus (approved):
Code changes look good. If you want me to test locally mention me again and I'll allocate 30m to do it, but I haven't had a full system running locally recently so it's not super fast for me to test.

mikemurray (approved):
👍

PR #6110, opened 2020-02-24
chore: update `plugin-simple-authorization` package

kieckhafer:
Update `plugin-simple-authorization` package. New version includes migrations which will fix group related permission issues.


manueldelreal (dismissed):
:shipit: tested locally with an older dataset

aldeed (dismissed):

manueldelreal (approved):

PR #6118, opened 2020-02-28
refactor: move navigation plugin into npm

### LINKED ISSUE #6127: move navigation plugin to its own npm package, and install npm version Reaction
New repo: https://github.com/reactioncommerce/plugin-navigation  

Convert the navigation plugin. This includes moving all related files into a new repository, releasing that repo as an `npm` package, and then updating this repo, `reaction`, to use the `npm` version of the plugin.

Part of [this epic](https://github.com/reactioncommerce/reaction/issues/6107) to move all plugins into their own repositories and release as `npm` packages.

kieckhafer:
Resolves #6127 

This PR removes the internal `navigation` plugin, and installs the new `plugin-navigation` as a replacement.

Running this query will show you the `defaultNavigationTreeId`, which has moved into the new package. If this query works, you should be good to go:

```
{
  primaryShop {
    defaultNavigationTreeId
  }
}
```


willopez (dismissed):
LGTM

tests/util/factory.js
```diff
@@ -1,3 +1,9 @@
+import {
+  NavigationItem,
+  NavigationTree,
+  NavigationTreeItem
+} from "@reactioncommerce/plugin-navigation/src/simpleSchemas.js";
```
kieckhafer, line 5: @aldeed this works, just wondering if this is how we want to import simpleSchemas from outside packages, or if there was something else in mind?

kieckhafer:
@aldeed does the failing GraphQL issue mean we need to move that part of the `schema` into the `navigation` plugin itself, or is there another way for that to work?

tests/util/factory.js
aldeed: @kieckhafer This would be fine. I had kind of hoped that when we got to this we'd start moving the integration tests out of this repo and into their respective package repos, but we have some work to do yet before that can happen.

aldeed:
@kieckhafer Yes I think the correct solution in this case is to cut the following out of the shop schema file, and add it to the nav package schema file within `extend type Shop { }`.

```graphql
"The default navigation tree for this shop"
  defaultNavigationTree(
    "Navigation tree language"
    language: String!,

    "Whether to include secondary navigation items"
    shouldIncludeSecondary: Boolean = false
  ): NavigationTree

  "The ID of the shop's default navigation tree"
  defaultNavigationTreeId: String
```

aldeed:
And move all related resolvers into that plugin package, too.

willopez (approved):
confirmed it still works

kieckhafer:
@spencern this is good to go, outside of the `snyk` issue. Once you figure out the deal there, can you please re-run snyk and merge this if it's good (or ping me and I can do it)

aldeed (approved):

package.json
```diff
@@ -28,6 +28,7 @@
     "@reactioncommerce/logger": "~1.1.2",
     "@reactioncommerce/nodemailer": "~5.0.5",
     "@reactioncommerce/plugin-authentication": "~1.0.0",
+    "@reactioncommerce/plugin-navigation": "~1.0.2",
```
spencern, line 31: This is the plugin that is getting flagged by snyk's license check as GPLv3. We've kept the GPLv3 license that this code was originally licensed under after migrating it to it's own package. We'll need to merge this code anyway and then add an ignore to snyk for this particular package.

spencern (approved):
Override approval of license failure.

PR #6149, opened 2020-03-13
fix: variant clone produces null and undefined values

### LINKED ISSUE #6148: Cloning a product variant generates error that breaks the UI
Prerequisites

## Issue Description
When a product variant is cloned, the UI breaks with the following error coming from the API:

```
1: {message: "Float cannot represent non numeric value: "null"", locations: [{line: 67, column: 3}],…}
message: "Float cannot represent non numeric value: "null""
locations: [{line: 67, column: 3}]
0: {line: 67, column: 3}
path: ["product", "variants", 1, "options", 0, "length"]
0: "product"
1: "variants"
2: 1
3: "options"
4: 0
5: "length"
extensions: {code: "INTERNAL_SERVER_ERROR",…}
code: "INTERNAL_SERVER_ERROR"
exception: {stacktrace: ["TypeError: Float cannot represent non numeric value: "null"",…]}
stacktrace: ["TypeError: Float cannot represent non numeric value: "null"",…]
0: "TypeError: Float cannot represent non numeric value: "null""
1: "    at GraphQLScalarType.serializeFloat [as serialize] (/usr/local/src/app/node_modules/graphql/type/scalars.js:96:11)
```

## Steps to Reproduce
1. Create a new product and variant
2. Click on the "Duplicate" link
3. Reload page and observe page load with no data, see image below for details

Screen Shot 2020-03-13 at 3 17 05 PM

## Possible Solution
Looks like the price is not being cloned correctly

mikemurray:
Resolves #6148 
Impact: **critical**
Type: **bugfix**

## Issue

Cloning a variant with that has missing values for title, optionTitle, length, width, height, and weight, would produce `null` values in the database. This breaks the API.

## Solution

Ensure the values valid before inserting them into the database.

## Breaking changes

none

## Testing
1. Use the reaction admin
2. Clone a variant with options, but don't fill out the fields `title, optionTitle, length, width, height, and weight`
3. Ensure everything clones correctly
4. Now fill out the missing values and clone a variant that has options
5. See everything clones correctly

### DIFF
--- a/src/core-services/product/mutations/cloneProductVariants.js
+++ b/src/core-services/product/mutations/cloneProductVariants.js
@@ -75,29 +75,44 @@ export default async function cloneProductVariants(context, input) {
     await Promise.all(sortedVariants.map(async (sortedVariant) => {
       const originalVariantId = sortedVariant._id;
       let type = "child";
-      const clonedVariantObject = {};
+      let clonedVariantObject;
       if (variantId === sortedVariant._id) {
         type = "parent";
-        Object.assign(clonedVariantObject, sortedVariant, {
+        clonedVariantObject = {
+          ...sortedVariant,
           _id: variantNewId,
-          title: `${sortedVariant.title} - copy`,
-          optionTitle: `${sortedVariant.optionTitle} - copy`
-        });
+          title: `${sortedVariant.title || "Untitled"} - copy`,
+          optionTitle: `${sortedVariant.optionTitle || "Untitled"} - copy`
+        };
       } else {
         const parentIndex = sortedVariant.ancestors.indexOf(variantId);
         const ancestorsClone = sortedVariant.ancestors.slice(0);
         // if variantId exists in ancestors, we override it by new _id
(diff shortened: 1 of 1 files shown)


willopez (changes requested):
Cloning variants now works correctly. However, it's not possible to set the price on a variant. When I set the price and click "Save changes" the prices gets cleared out to an empty string.

mikemurray:
@willopez I don't think this PR is the ultimate cause of these issues. In the Reaction Admin, I'm seeing missing `price/compareAtPrice` from the product GQL fragment.

I'm also noticing that the price field for variants with options is an `object` and not `float` to represent the price range. This is a bigger problem because a field in GQL is changing between two different types.

mikemurray:
@willopez Added a ticket about the price issue: https://github.com/reactioncommerce/reaction/issues/6152

willopez:
@mikemurray as discussed, we'll handle that issue in a separate PR.

willopez (approved):
LGTM

PR #6150, opened 2020-03-15
feat: support passing product data with createProduct and createProductVariant

### LINKED ISSUE #6103: CreateProductInput & CreateProductVariantInput needs product data support
## Prerequisites

## Issue Description
The new `createProduct` and `createProductVariant` mutations let users create product information via the GraphQL API. From the UI perspective these mutations work as expected, click the "create a product" button, it calls these mutations, creates an empty product in the DB then the user can add product data from the UI using the `updateProduct` or `updateProductVariant` mutation. From an API perspective this creates an the need for an extra request if I'm trying to import product data via an external service/lambda.  

## Steps to Reproduce
See [CreateProductInput](http://apidocs.v3.reactioncommerce.io/#definition-CreateProductInput) & [CreateProductVariantInput](http://apidocs.v3.reactioncommerce.io/#definition-CreateProductVariantInput) type definitions and see there is no way to provide product data to the mutation call.

## Possible Solution
Allow an optional product data object as part of the mutation's input types.

### Versions
3.x

aldeed:
Resolves #6103 
Type: **feature**

## Changes
- [x] Add optional ProductInput to `createProduct` mutation
- [x] Add a `createProduct` mutation param to allow skipping auto-creation of the first variant
- [x] Add optional ProductVariantInput to `createProductVariant` mutation

## Breaking changes
None

## Testing
1. Pass `product` param to `createProduct` mutation with all available fields set, and verify that they have correct values in the response and the database.
1. Pass `variant` param to `createProductVariant` mutation with all available fields set, and verify that they have correct values in the response and the database.
1. Ensure that `updateProduct` and `updateProductVariant` mutations still work correctly for updating the same fields.
1. Pass `shouldCreateFirstVariant: false` when calling the `createProduct` mutation, and verify that the `variants` array is empty in the response, and that no variant was created in the database.


mikemurray (commented):
@aldeed I can create variants with a `price` but not `compareAtPrice`. I thought prices should be set through `updateVariantPrices` mutation rather than through the main variant mutations. Not sure if we should remove the `prices` field, add `compareAtPrice`, or just leave it as is.

Otherwise, things seem to work fine.

brent-hoover:
@aldeed I think we should allow (optional) passing of `_id` when creating a product via the API. This would allow import processes that want to use a natural key to specify this. If not specified then it could use the typical random string.

aldeed:
@zenweasel Added setting _id on creation for products and variants

@mikemurray This is an existing issue so I cannot remove `price` without it being a breaking change to the `updateProductVariant` mutation. Instead, I added `compareAtPrice` to keep it consistent, but marked both of those fields as deprecated with instructions to use `updateProductVariantPrices` instead.

@manueldelreal This should be ready for final testing now.

manueldelreal (approved):
:shipit:

PR #6155, opened 2020-03-20
feat: reaction as library

rosshadden:
Impact: **major**
Type: **feature|refactor**

## Issue

Reaction is currently unusable as a library. You can only really launch it as-is or fork it (or similar).

## Solution

There are multiple changes toward allowing use of Reaction as a library.

1. The `ReactionAPI` class, and all related code from `src/core` has been moved to a `@reactioncommerce/api-core` package, where it is now named `ReactionAPICore`. Similarly, the `TestApp` class is moved to that package and is now named `ReactionTestAPICore`.
2. The `registerPlugins` code file is removed in favor of a `plugins.json` file that is more robust in terms of allowing automatic generation as well as referencing either packages or relative paths to plugins.
3. While the above would be sufficient if all plugins had already been moved to be NPM packages, this is not yet the case. So to support extending the default Reaction without forking it, until all built-in plugins are in NPM, it's also possible to load a `plugins.json` file from a non-standard location and transform the plugin list before they are registered. See "Extend Reaction API with default plugins"

### Extend Reaction API with default plugins

The basic idea is to "install" the `reaction` repo as if it is a package (referenced as `reaction-api` in the code below), and then transform the plugins list as necessary for your installation.

```js
import { ReactionAPICore } from "@reactioncommerce/api-core";
import pkg from "reaction-api/package.json";
import authorizationPlugin from "@reactioncommerce/reaction-plugin-authorization";

const api = new ReactionAPICore({
  // See "ReactionAPICore Configuration" section. Many options are also configured
  // by environment variables. See "Supported Environment Variables" section.
  serveStaticPaths: ["public"],
  version: pkg.version
});

async function run() {
  // Load the default list of built-in plugins but then transform it to add/remove
  const plugins = await ReactionAPICore.importPluginsJSONFile("reaction-api/plugins.json", {
    transformPlugins(pluginList) {
      pluginList.authorization = authorizationPlugin;

      // you could also delete pluginList.something here

      return pluginList;
    }
  });

  await api.registerPlugins(plugins);
  await api.start();
}

run().catch((error) => {
  Logger.error(error);
  process.exit(1);
});
```

## Breaking changes

Since Reaction was not previously a library there are no outward-facing breaking changes here. It does drastically change how plugins are imported though, and this affects this repo and anybody that may be using it for their own projects. This pretty much comes down forks/copies, because if they are just using it as-is they would have nothing break.

## Testing Phase 1 (`api-core` not yet released)

CI checks for integration tests won't pass yet, but it can be tested manually

### Prep
1. Clone https://github.com/reactioncommerce/api-core
2. Check out `trunk` branch on `api-core` and `npm install` on the host machine.
3. Review all `api-core` code. There is no PR for this package because it's new, but 99% of it is directly copied from the `src/core` folder with just some renaming.
4. Do a local link to the `api-core` package as described [here](https://docs.reactioncommerce.com/docs/devs-update-custom-plugin#how-to-link-in-a-local-npm-package-for-a-plugin-while-working-on-it)

### Run API
1. Prep steps above
2. `make`
3. `make dev-reaction`
4. Check API logs to verify it started without errors
5. A few smoke tests to be sure things generally continue to work.

### Run Integration Tests
1. Prep steps above
2. `docker-compose run api npm run test:integration` and verify all tests pass

## Testing Phase 2 (after `api-core` is released)

1. Repeat phase 1 tests, but without linking to the local `api-core` package
2. All CI checks should now pass.
3. For bonus points you can install reaction as a library. See instructions above.


aldeed:
Took this over. Updated the description and testing instructions. Ready for phase 1 of testing.

rosshadden:
Now that the usage is a bit different from how it originally was when I added the `transformPlugins` option, I think it would be more straightforward to make the second parameter of `importPluginsJSONFile` a function instead of an object. `transformPlugins` was originally passed through as a Reaction option but that is not true anymore.

aldeed:
@rosshadden I made that change and merged in latest trunk

aldeed (approved):

aldeed:
I discovered that the federation support was added after I moved those files to `api-core`, so not merging this yet. I'll update `api-core`.

aldeed:
Nevermind, federation support is there.

PR #6159, opened 2020-03-25
fix: adds new pricing field to product variants in order to correctly display prices

### LINKED ISSUE #6152: GQL: Variant can be both a float or an object(price range)
## Issue Description

The price field for `Variants` can be either a `float` or an. `object` e.g. `{ range, min, max }`. This is bad as the type is changing for a field that is labeled as a `Float` in the GQL schema.

## Possible Solution

The data must have a consistent form, regardless of if it's a variant, option or variant with options.

In the `simple-pricing` plugin
- The `Product` and `Variant` schema extensions should have the field `pricing: ProductPricingInfo` field, and properly resolve that field as best a possible.
- We are then, discouraging the use of `price` and `compareAtPrice` and instead urge users to use `pricing` instead.
- The mutations for `updateVariantPrice` should still work

willopez:
Resolves #6152
Impact: **minor**
Type: **bugfix**

**NOTE** test with PR: https://github.com/reactioncommerce/reaction-admin/pull/246
## Issue

## Solution
The `price` field on variants was not flexible enough to accommodate a variant having options with their own prices. A new field `pricing` has been added to the `Product` and `ProductVariant` types. This new field has a transformer that determines the correct value to return depending on whether a variant is a parent variant or not.

## Breaking changes
None.

**NOTE** The field `price` has been deprecated in favor of `pricing` both on products and variants.

## Testing
1. Test with Admin PR: https://github.com/reactioncommerce/reaction-admin/pull/246 and follow testing  instructions there.


src/plugins/simple-pricing/xforms/xformProductVariantPricing.js
```diff
@@ -0,0 +1,25 @@
...
+      compareAtPrice: null,
+      displayPrice: getDisplayPrice(price.min, price.max),
+      maxPrice: price.max,
+      minPrice: price.min,
+      price: null
+    };
+  } else {
+    pricing = {
+      compareAtPrice: node.compareAtPrice || 0,
+      displayPrice: null,
```
mikemurray: `displayPrice` here could be just the single formatted price.

src/plugins/simple-pricing/xforms/xformProductVariantPrice.js
```diff
@@ -0,0 +1,15 @@
...
+  let { price } = node;
+  // If a variant has options, then its price will become an object
+  // that includes the following props: min, max and range. This change
+  // in the underlying data structure will break the API as the variant
+  // field `price` will attempt to return an `Object` for a field that previously
+  // expected a `Float`. This field resolver mitigates this problem by returning
+  // a `Float` as expected. This will prevent the API and UI from breaking. Further,
+  // this field, is now deprecated in favor of the new `pricing` field.
+  if (typeof price === "object") {
+    price = price.min;
```
mikemurray: I think this should be `null` when it's an object because `min/max` price can be misleading. `null` should be fine since the schema doesn't make this field required.

src/plugins/simple-pricing/xforms/xformProductVariantPrice.js
```diff
@@ -0,0 +1,15 @@
+export default (node) => {
+  let { price } = node;
```
mikemurray, line 4: We should add `Logger.warn` here to notify users that this field is deprecated?

src/plugins/simple-pricing/resolvers/Product.js
```diff
@@ -0,0 +1,5 @@
+import xformProductPricing from "../xforms/xformProductPricing.js";
+
+export default {
+  pricing: (node) => xformProductPricing(node)
```
mikemurray: Can we add a resolver for `price` that...
- Returns the `price` as is with no modification
- but, `Logger.warn` with a deprecation notice?

src/plugins/simple-pricing/resolvers/ProductVariant.js
```diff
@@ -0,0 +1,7 @@
+import xformProductVariantPricing from "../xforms/xformProductVariantPricing.js";
+import xformProductVariantPrice from "../xforms/xformProductVariantPrice.js";
+
+export default {
+  price: (node) => xformProductVariantPrice(node),
```
mikemurray, line 7: Can we add a resolver for `compareAtPrice` that...

- Returns the `compareAtPrice` as is with no modification
- but, Logger.warn with a deprecation notice?

mikemurray (changes requested):

src/plugins/simple-pricing/xforms/xformProductVariantPrice.js
willopez: updated

src/plugins/simple-pricing/resolvers/Product.js
willopez: updated

src/plugins/simple-pricing/resolvers/ProductVariant.js
willopez: updated

src/plugins/simple-pricing/xforms/xformProductPricing.js
```diff
@@ -0,0 +1,11 @@
+export default (node) => {
+  const { price } = node;
+
+  const pricing = {
+    displayPrice: price.range,
```
mikemurray: The display price needs to contain the currency symbol, kinda how you used `getDisplayPrice` but with the currency symbol passed in as an option. I'd default to the shop's default currency since products don't have a currency field.

src/plugins/simple-pricing/xforms/xformProductVariantPricing.js
```diff
@@ -0,0 +1,26 @@
...
+      compareAtPrice: null,
+      displayPrice: getDisplayPrice(price.min, price.max),
+      maxPrice: price.max,
+      minPrice: price.min,
+      price: null
+    };
+  } else {
+    pricing = {
+      compareAtPrice: node.compareAtPrice || 0,
+      displayPrice: accounting.formatMoney(price),
```
mikemurray: Display price needs the currency symbol, I believe you can pass in the currency info as the second param.

src/plugins/simple-pricing/xforms/xformProductVariantPricing.js
```diff
@@ -0,0 +1,26 @@
+import accounting from "accounting-js";
+import getDisplayPrice from "../util/getDisplayPrice.js";
+
+export default (node) => {
+  let pricing = {};
+  const { price } = node;
+  if (typeof price === "object") {
+    pricing = {
+      compareAtPrice: null,
+      displayPrice: getDisplayPrice(price.min, price.max),
```
mikemurray: I think you need to pass in some currency info as the 3rd param to get currency symbols in the string.

mikemurray (dismissed):
I'm seeing an issue where the data in the database is saved incorrectly for Variants with Options. The price `min` and `max` are being saved as objects instead of as floats. I'm also seeing incorrect values for the price ranges for the Product, and top-level Variant, likely due to this bug. Something may be wrong with the `updateVariantPrices` mutation or app events thereafter.

image

src/plugins/simple-pricing/xforms/xformProductPricing.js
willopez: will create a new issue to add the currency symbol.

willopez:
Currency symbol issue: https://github.com/reactioncommerce/reaction/issues/6162

mikemurray (approved):
👍

Issue #5481, opened 2019-08-27
Spike: Investigate transitioning from Simple-Schema to Mongo Schemas


zenweasel:
Now that Mongo schemas have been out for a while, we should look into them and see how they could be leveraged to add a better level of data integrity/performance to using Simple-Schema.

janus-reith:
I wondered the same thing when I found out that `attachSchema` is something Meteor-based that will become obsolete.

While the whole stack is tied tightly to mongoDB and there don't seem to be alternatives yet, I still wonder if it would be possible to keep the Schemas database-agnostic and simply enhance Simple-Schema to make use of Mongo Schemas where available.

Maybe Simple-Schema could somehow pass through data were applicable and thereby still give performance benefits.

And while we are at it: Moving a reference of the schema definitions to each `registry.js` would really help keep things more structured, as collections and their indices are defined there aswell. Currently all of them still are in `/imports/collections/schemas`, while `attachSchema` happens in /lib/collections and for some collections in the `/lib` or `/client` folder of plugins.

focusaurus:
Getting mongodb schemas going and enforced in the db: :+1: 

JSONSchema seems like the most pragmatic approach. Hopefully we can find a good implementation of validation on npm. I've used a few in the past and the quality was so-so.

One stack to consider would be migrating our SimpleSchema code to `joi` and finding an npm package that could convert/downsample that to JSONSchema for mongo.

aldeed:
I've used `joi` a bit and it's nice for what it is, but may not be best for the needs here. I recently tried [superstruct](https://github.com/ianstormtaylor/superstruct) and there are a few things I didn't like but overall it probably does not get much simpler, with the added benefit of being fast and everything being strings and scalars, which means it might be the easiest library to write a JSONSchema transform for.

janus-reith:
I think it could be great to find a solution where we could also derive GraphQL types from, so Schemas don't need to be written twice.

janus-reith:
Hey, has there been any further discussion on this and did someone maybe come up with a new idea?

aldeed:
I'm not aware of any further discussion. The last I heard the thought was that a single plugin could try implementing mongo schemas for the collections it owns as an example and a trial, and then evaluate based on that. If someone wants to do a PR for that at some point, that could get things moving.

janus-reith:
[...]

PR #6163, opened 2020-03-28
Add support for mongo collections validation schemas

set321go:
Signed-off-by: Alex Edwards <alex.edwards@data-realm.com>

This change is pulled from a spike i did for #5481, it changes the method used to create collections on application startup and allows for a validator to be used. 

I have not added it but this change could be bubbled out to the options object allowing general properties to be set on collections. In addition using the collectionOptions should allow the registration of indexes to be done in the request.

## Testing
I've run the tests. I don't know enough about mongo do know what impacts changing the method used here would have.


aldeed (changes requested):
I see there are two additional validation options listed here: http://mongodb.github.io/node-mongodb-native/3.1/api/Db.html#createCollection

I don't think we should accept and pass through the entire `collectionOptions` object because we should be aware of what exactly we are allowing plugins to do. But I do think we should pass through all of the available `validation*` options because they are very useful for transitioning to this type of validation. We can additionally accept `schema` as you do, as a convenience alternative. So like this:

```js
const collectionOptions = {};
if (collectionConfig.schema) {
  collectionOptions.validator = {
    $jsonSchema: collectionConfig.schema
  };
} else if (collectionConfig.validator) {
  collectionOptions.validator = collectionConfig.validator;
}

if (collectionConfig.validationLevel) {
  collectionOptions.validationLevel = collectionConfig.validationLevel;
}
if (collectionConfig.validationAction) {
  collectionOptions.validationAction = collectionConfig.validationAction;
}
```

After that change, this looks good to merge.

aldeed:
This PR is replaced by https://github.com/reactioncommerce/api-core/pull/3

PR #6165, opened 2020-03-30
refactor: filter out variant media from product media query

willopez:
Impact: **minor**
Type: **bugfix**

## Issue
The top level product media resolver was returning media from variants. This causes all media in child variants and options to be rendered on the parent top level product.

## Solution
Filter out variant media from product media resolved.

## Breaking changes
NONE

```
query product($productId: ID!, $shopId: ID!, $shouldIncludeVariantMedia: Boolean) {
  product(productId: $productId, shopId: $shopId, shouldIncludeVariantMedia: $shouldIncludeVariantMedia) {
    media(shouldIncludeVariantMedia: $shouldIncludeVariantMedia {
    _id
    URLs {
      small
    }
  }
  }
}
```
Variables:
```
{
  "productId": "myProductId",
  "shopId": "myShopId",
  "shouldIncludeVariantMedia": false
}
```

## Testing
1. Execute product media query above and verify that only media for the top level product is returned.


aldeed:
@willopez I think this was an intentional design decision, but I don't remember for sure. IDK which way people expect, but there might be people relying on this behavior so this could be considered a breaking change. @spencern or @zenweasel what do you think?

mikemurray:
@willopez, @aldeed is correct. It "was" intentional, though not properly displayed in the Admin UI. 

The media bubbles up to the top. So If I upload an image at the `Option` level, then the parent `Variant` inherits the option media, and the top-level `Product` then inherits all the `Variant` and `Option` media.

This is so you can organize the top-level product images if you want, and so you don't have to upload images to every single level. You could just upload only option media, and leave the other levels blank.

This made sense when the products were displayed on the storefront from the products collection. Now that we get the info from the Catalog, we can still have that functionality there, while doing something different with Products.

---

I can see @willopez intent here because...

When displaying the product images, the bubble-up effect is useful. When uploading images in the Admin UI, having child media in the uploader, especially at the product level, is overwhelming.

Perhaps adding a param the `getProductMedia` called like `shouldIncludeVariantMedia` could be an option to keep the old functionality while allowing the new.

Otherwise, we can always filter out the media in the UI and display a better user interface for managing direct media dependencies, while also allowing the organization of all product media at the top-level.

willopez:
Adding a param `shouldIncludeVariantMedia` seems good to me.

spencern:
Sounds good to me 👍

brent-hoover:
@aldeed We don't currently rely on this behavior because we don't use Product Media at all but I think for building an admin screen it would be good to have the flexibility that the `shouldIncludeVariantMedia` parameter would provide

aldeed:
@willopez `shouldIncludeVariantMedia` sounds good to me as long as it defaults to `true` when not set so that current behavior doesn't break. I think it should be an argument on the `Product.media` field rather than on the query, so that you can use it anywhere that a `Product` type is returned.

```js
query product($productId: ID!, $shopId: ID!) {
  product(productId: $productId, shopId: $shopId) {
    media(shouldIncludeVariantMedia: false) {
      _id
      URLs {
        small
      }
    }
  }
}
```

willopez:
@mikemurray @aldeed @spencern `shouldIncludeVariantMedia` argument added.

src/core-services/product/schemas/product.graphql
```diff
@@ -605,6 +608,9 @@ extend type Query {
 
     "Shop ID"
     shopId: ID!
+
+    "Should variant media be included in the top level product?"
+    shouldIncludeVariantMedia: Boolean
```
aldeed: I don't think this one is needed if you have the media one.

src/core-services/product/utils/getProductMedia.js
```diff
@@ -3,18 +3,26 @@
...
-export default async function getProductMedia(product, context) {
+export default async function getProductMedia(product, { shouldIncludeVariantMedia = true }, context) {
   const { Media } = context.collections;
   const { _id: productId, shopId } = product;
+
   if (!Media) return [];
 
+  let includeVariantMedia = {};
+  if (!shouldIncludeVariantMedia) {
+    includeVariantMedia = { "metadata.variantId": { $eq: null } };
```
aldeed: Just `"metadata.variantId": null` should work

src/core-services/product/schemas/product.graphql
```diff
@@ -19,7 +19,10 @@ type Product {
   isVisible: Boolean!
 
   "All media for a product"
-  media: [ImageInfo]
+  media(
+    "Determines whether variant media should be included in the product or not"
+    shouldIncludeVariantMedia: Boolean
```
aldeed: You can do `shouldIncludeVariantMedia: Boolean = true` here, too, so that it's clear in the schema explorer that the default is true.

aldeed (dismissed):
A few comments

src/core-services/product/schemas/product.graphql
willopez: updated

src/core-services/product/schemas/product.graphql
willopez: updated

src/core-services/product/utils/getProductMedia.js
willopez: updated

mikemurray (approved):
👍

PR #6183, opened 2020-04-01
feat: deprecate bin/token script

aldeed:
Impact: **minor**
Type: **feature**

## Changes
The `bin/token` script now prints a message instead of installing packages and running the `getAuthToken.js` script.

> Please globally install the @reactioncommerce/hydra-token package:
> 
> https://www.npmjs.com/package/@reactioncommerce/hydra-token
> 
> Then run 'hydra-token get' command instead of 'bin/token'

## Breaking changes
None

## Testing
Run `bin/token 123` and verify that you see the message.


focusaurus (dismissed):

focusaurus:
Does this count as breaking? Should we run hydra-token get via npx so it's still "here's your token, but also here's the new way?"

aldeed:
@focusaurus I would not consider this breaking, but your idea is good anyway. How's this? https://github.com/reactioncommerce/reaction/pull/6183/commits/15a2884c8c7be63de8adc88cf07bcb82ddd1934f

aldeed:
@focusaurus Can you approve? It removed your previous approval when I pushed, and I can't merge without an approval.

focusaurus (approved):

PR #6189, opened 2020-04-07
refactor: move accounts plugin to npm

kieckhafer:
Resolves #5693 

This PR removes the internal account plugin, and installs the new api-plugin-accounts as a replacement.

To test this, start a store and create an account from the UI.

This PR also adds the file `src/mockTypes.graphql`. This is a temporary file and is being used to appease graphql linting while we move these packages out of reaction and into their own plugins. Because the types are no longer always registered internally in reaction. and instead in each plugin, we are sometimes seeing errors in linting that we won't see once these packges are `peerDependency`'d with each other outside of the api.


focusaurus (dismissed):
I didn't test locally but looks OK.

aldeed (dismissed):
👍 App starts and `viewer` query works

aldeed (approved):

PR #6195, opened 2020-04-14
(feature) Surcharge and AppliedSurcharge available on context

brent-hoover:

Resolves: None
Impact: **minor**
Type: **feature**

## Issue
In order to modify the behavior of Surcharges you need to have access to the schemas but they are not available on context.

## Solution
This PR adds them to the context. Also moves/renames one file to match the naming schema

## Breaking changes
None

## Testing
1. Ensure everything behaves as normal


aldeed:
We are moving all plugins out to NPM packages with code in separate repos, so when that is done (hopefully later this week), I will manually copy your changes over to the new surcharges plugin.

brent-hoover:
@aldeed Great. Just a note that it has changes in both Cart and Surcharges

aldeed:
Moved these changes to here: https://github.com/reactioncommerce/api-plugin-surcharges/pull/3

Note that `AppliedSurcharge` schema was actually duplicated in both carts and orders plugins, so I moved it to surcharges plugin and will remove from those. After all three of those plugin PRs are released and then pulled into here, this will be "fixed"

PR #6213, opened 2020-04-18
updated surcharge schema

### LINKED ISSUE #6106: Surcharge schema doesn't include all fields
## Prerequisites

## Issue Description
The `Surcharge` schema (SimpleSchema) in `surcharges` plugin doesn't have all fields. This is because we're running the validation before we add some fields to it. There's no actual bug here, but it leads to confusion for those looking at the schema to see what fields to expect.

Additionally, the GraphQL `Surcharge` schema is missing some, but not all, of the same fields.

## Solution
Add the following fields to the `Surcharge` SimpleSchema in the `surcharges` plugin (verify no others are missing):

```js
{
  _id: String,
  shopId: String,
  createdAt: Date,
  updatedAt: {
    type: Date,
    optional: true
  }
}
```

Add the following fields to the `Surcharge` GraphQL schema:

```gql
"The date and time at which this surcharge was created"
createdAt: DateTime!

"The date and time at which this surcharge was last updated"
updatedAt: DateTime
```

Then adjust the validation code in `createSurcharge` and `updateSurcharge` mutations:

1. In the input schemas, change `surcharge: surchargeSchema` to `surcharge: { type: Object, blackbox: true }`
2. Right before the insert or update, after all fields are added, do `surchargeSchema.validate(surcharge)` or `surchargeSchema.validate(modifier, { modifier: true })`

### Versions
3.1.0

robwgreenjr:
Resolves #6106 
Type: refactor

Updated surcharge schema and then set surcharge validations for create and update surcharge

### DIFF
--- a/src/plugins/surcharges/mutations/createSurcharge.js
+++ b/src/plugins/surcharges/mutations/createSurcharge.js
@@ -4,8 +4,11 @@ import ReactionError from "@reactioncommerce/reaction-error";
 import surchargeSchema from "../util/surchargeSchema.js";
 
 const inputSchema = new SimpleSchema({
-  surcharge: surchargeSchema,
-  shopId: String
+  shopId: String,
+  surcharge: {
+    type: Object,
+    blackbox: true
+  },
 });
 
 /**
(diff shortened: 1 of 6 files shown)


aldeed:
@robwgreenjr Thank you for fixing this! We are moving all plugins out to NPM packages with code in separate repos, so when that is done (hopefully later this week), I will manually copy your changes over to the new surcharges plugin.

aldeed:
Moved these changes to here: https://github.com/reactioncommerce/api-plugin-surcharges/pull/3

PR #6216, opened 2020-04-20
refactor: move discounts plugin to NPM package

### LINKED ISSUE #6170: move discounts plugin to its own npm package, and install npm version Reaction
New repo: https://github.com/reactioncommerce/api-plugin-discounts  

Convert the `discounts` plugin. This includes moving all related files into a new repository, releasing that repo as an `npm` package, and then updating this repo, `reaction`, to use the `npm` version of the plugin.

Part of [this epic](https://github.com/reactioncommerce/reaction/issues/6107) to move all plugins into their own repositories and release as `npm` packages.

aldeed:
Resolves #6170 

Removes the internal discounts plugin, and installs the new api-plugin-discounts as a replacement.


kieckhafer (approved):
👍

PR #6229, opened 2020-04-24
chore: remove unused npm packages

kieckhafer:
Many packages we used to need to install in the core api are no longer needed, as they are installed in each plugin instead. This PR removed the unneeded npm installations.

Removed packages:
- `@reactioncommerce/reaction-error`
- `accounting-js`
- `envalid`
- `graphql`
- `handlebars`
- `later`
- `ramda`
- `simple-schema`

## Testing
- See the app starts as expected
- Running `npm lint` should tell you if any plugins are missing


kieckhafer:
@aldeed a few questions about other packages:

- `graphql` isn't used, but feels weird for me to remove, so just want to double check
- `lodash` is only used in the `.reaction` folder, do we still need it?
- `@reactioncommerce/nodemailer` isn't "used", but is mocked in a test, do we still need it in the `package.json` (i'm thinking no)? https://github.com/reactioncommerce/reaction/blob/chore-kieckhafer-removeUnusedNPMPackages/tests/integration/api/mutations/verifySMTPEmailSettings/verifySMTPEmailSettings.test.js#L8:L8

aldeed (changes requested):
@kieckhafer 

- I discovered the other day that `graphql` package is similar to `react` in that it can cause errors if there are two different versions installed. So I'm now thinking maybe `graphql` should be switched to be a peer dependency in the `api-core` package, rather then a regular dep. So if we do that, then `graphql` should remain installed in this repo.
- lodash is used in `jsdoc` folder, but there is no point to building jsdoc docs anymore since this repo has very little code. So I suggest delete `.reaction/jsdoc` folder, search for and delete anything else related to `jsdoc`, and then also get rid of `lodash`.
- I think you're correct that mocked package doesn't need to be listed, but not sure either. If you remove it and tests still pass, then I'd say we're good.

Also, it looks to me like `faker`, `graphql-tools`, and `node-fetch` can be moved to devDependencies because only test files import them, and `sharp` is unused and can be removed. Only `semver` and the `@reactioncommerce` packages are needed in `dependencies`

kieckhafer:
@aldeed 

- looks like graphql already is a peerDependency in `api-core`, so I'll leave installed here. https://github.com/reactioncommerce/api-core/blob/trunk/package.json#L100
- `sharp` is used in the Dockerfile, which is why I left it. Is it used in a different way which doesn't need it to be installed? I think because it's in the `node_modules` we still need to install (but maybe as a devDependency?) https://github.com/reactioncommerce/reaction/blob/chore-kieckhafer-removeUnusedNPMPackages/Dockerfile#L8:L8

aldeed:
@kieckhafer node_modules/sharp should continue to exist in the production image as long as any of the plugin packages depend on it. I'd say go ahead and remove it entirely, and after this is merged and we get a new Docker image build for trunk, we'll be able to run and verify that I'm correct. If I'm not, easy to add back before the next release.

aldeed (dismissed):

aldeed:
Removed the direct dependency on `sharp`. It didn't cause the package-lock to change so I think we're okay.

aldeed (approved):

PR #6238, opened 2020-04-29
chore: add package link and unlink scripts

aldeed:
Impact: **minor**
Type: **chore**

## Issue
It's difficult to properly link a work-in-progress plugin (or other npm package) into the API running in a Docker container.

## Solution
Adds two new scripts in `bin`:

**Run `docker-compose up -d` before running any of these commands, and wait until the API has fully started.**

- Enter `bin/package-link <package-name>` to link locally cloned package repo into the container and `npm link` it. This assumes that the package lives in `../api-plugins`, which will be true if it's a plugin package and you used the `make clone-api-plugins` command to clone it.
- Alternatively enter `bin/package-link <package-name> <path-to-package-repo>` for any package that is not located in  `../api-plugins`
- When you're done testing, run `bin/package-unlink <package-name>`

### Some additional points
- It should work to link multiple plugins at once (one at a time)
- If you do not unlink before you stop the `api` service, you will likely get into a loop where the API won't ever start. If this happens, delete the node_modules volume with something like `docker volume rm reaction_api_node_modules`, and then `docker-compose up -d` should work.
- When you make changes to the package repo code, run the same `bin/package-link` command again. You do NOT have to (and must not) manually restart the `api` container. The Reaction API will magically restart after the linking finishes.

### Example workflow
I cloned [reaction-development-platform](https://github.com/reactioncommerce/reaction-development-platform) and ran `make`. API is now running the latest production.

Now I want to make a change in the `api-plugin-shops` plugin.

First I can clone all plugins by running this in the `reaction-development-platform` directory:

```sh
make clone-api-plugins
```

Now I need to switch the API to run in development mode:

```sh
make dev-reaction
```

Now let's move into the API directory:

```sh
cd reaction
```

Then I'll view the logs to make sure the API fully starts:

```sh
docker-compose logs -f api
```

When I see that it's running, I'll CTRL+C to stop.

Now I can open `<reaction-development-platform directory>/api-plugins/api-plugin-shops` in my editor and make some changes.

When I'm ready to link in those changes to test them, I run this in the API directory:

```sh
bin/package-link api-plugin-shops
```

The API will restart when it's done and should reflect the local code changes.

I then make more changes and again run:

```sh
bin/package-link api-plugin-shops
```

When everything is working properly I run:

```sh
bin/package-unlink api-plugin-shops
```

This reverts the link. Then I can `docker-compose down` if I want to, or link in other packages.

## Breaking changes
None

## Testing
Try it out!

PR #6251, opened 2020-05-20
chore: fix package link script issues

### LINKED ISSUE #6248: bin/package-unlink unlinks all packages instead of the one that was specified
Prerequisites

## Issue Description
Running `bin/package-unlink some-plugin` (works for all plugins) seems to unlink *all* plugins and not just `some-plugin`. This is the case regardless of whether the plugin was linked with `bin/package-link` beforehand or not.

## Steps to Reproduce
1. Link a plugin with `bin/package-link some-plugin`, then link another plugin with `bin/package-link other-plugin`.
2. Open a shell on your API container with `docker exec -it reaction_api_1 bash`.
3. In your container's shell, go to `/usr/local/src/app/node_modules/@reactioncommerce`.
4. Still in your container's shell, run `ls -la`. Notice you have two symlinks in place: one for `some-plugin` and another for `other-plugin`.
5. On your host machine, run `bin/package-unlink some-plugin` to unlink `some-plugin`.
6. Back into your container's shell, run `ls-ls`. Notice that both `some-plugin` and `other-plugin` were unlinked.

## Possible Solution
No idea. Is this even on Reaction's side or is it an NPM issue?

### LINKED ISSUE #6250: Linked packages don't get updated with new changes
Prerequisites

## Issue Description
Linking a package with `bin/package-link` works fine. However, linking this same package a second time, once changes were made locally, doesn't update the one in the container. It doesn't matter whether you unlink the package before re-linking it. The behavior is the same.

## Steps to Reproduce
1. Link a package with `bin/package-link`. For the sake of having something to compare, include a `Logger.info("Hello World")` in a query resolver, for example.
2. Call this query, and notice the `Hello World` being printed in the server logs.
3. Remove the `Logger.info` line from the local copy of the package.
4. Unlink this same package with `bin/package-unlink`, then re-link it with `bin/package-link`.
5. After the API has restarted, call the same query again and notice the `Hello World` being printed, even though the `Logger.info` call was removed.

## Possible Solution
Maybe this is due to some NPM cache or something. There's probably a `--no-cache` flag we can use when linking/unlinking packages.

aldeed:
Resolves #6250
Impact: **minor**
Type: **chore**

- npm install official packages after unlink
- don't unlink unless it is truly linked
- require package namespace for both commands

## Breaking changes
The `package-link` and `package-unlink` commands now need the full package name, including `@reactioncommerce/` org prefix.

## Testing
Try it

### DIFF
--- a/bin/package-link
+++ b/bin/package-link
@@ -20,12 +20,16 @@ container_copied_packages_path="/home/node/copied-packages"
 container_destination_path="${container_copied_packages_path}/${package_name}"
 
 if [[ -z "${package_path}" ]]; then
-  package_path="../api-plugins/${package_name}"
+  package_name_without_org="${package_name/#@reactioncommerce\/}"
+  package_path="../api-plugins/${package_name_without_org}"
+  full_package_path="$(cd ${package_path} && pwd)"
+  echo "\nUsing local package path ${full_package_path}"
+  echo "If this is not correct, specify the correct path as the second argument.\n"
 fi
 
 echo "Copying ${package_name} package code into container..."
-docker-compose exec api sh -c "mkdir -p ${container_copied_packages_path}"
-docker cp "${package_path}" "${container_id}:${container_destination_path}"
+docker-compose exec api sh -c "mkdir -p ${container_destination_path}"
+docker cp "${package_path}/." "${container_id}:${container_destination_path}"
 
 # Then npm link into this project
 echo "Linking package into API..."
@@ -33,4 +37,5 @@ docker-compose exec api sh -c "npm config set prefix /home/node/npm && cd /usr/l
 
 # Fool nodemon into thinking something has changed so that it restarts.
 # Touch first file found in /src with .js extension
+echo "Restarting API..."
 docker-compose exec api sh -c "touch -c $(ls ./src/*.js | head -n1)"
--- a/bin/package-unlink
+++ b/bin/package-unlink
@@ -14,11 +14,28 @@ IFS="$(printf "\n\t")"
 # ---- End unofficial bash strict mode boilerplate
 
 package_name=$1
+package_path_in_container="./node_modules/${package_name}"
+container_copied_packages_path="/home/node/copied-packages"
+container_destination_path="${container_copied_packages_path}/${package_name}"
 
-# Then npm link into this project
-echo "Unlinking package from API..."
-docker-compose exec api sh -c "cd /usr/local/src/app && npm unlink "${package_name}""
+# We start by attempting to confirm that a package with the given name is actually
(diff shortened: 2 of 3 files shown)


aldeed:
I'd like to also figure out a way to maintain the other links when unlinking one, but I think we'd have to create a file to track the links and re-create them all in the unlink script after unlinking. In other words, essentially write our own version of the NPM RFC proposal.

A bigger concern is that I am seeing that often, after I link a second package, the code from the first package is no longer executing, even though the symlink is still there and the symlinked files have the changes. It's like there's a different version of the package also installed somewhere in the `node_modules` tree, and the `import()` is finding that one instead. So far I haven't figured out what's going on.

aldeed:
My "bigger concern" comment is captured in #6250, and I think I actually fixed it by changing `docker cp` command in this PR. So I believe the only remaining issue is that `unlink` unlinks all packages and then you have to relink any you still want linked. Most likely will not figure out a solution for that in this PR.

focusaurus:
OK I think I encountered some of what is being discussed in #6250 today so I'm going to test this branch and see if it helps. One thing I'd like to confirm about this tooling, is it:

- use BOTH package-link and a docker volume mount

or is this package-link workflow designed to work without an extra docker volume mount for the api-plugin you're working on?

kieckhafer:
> OK I think I encountered some of what is being discussed in #6250 today so I'm going to test this branch and see if it helps. One thing I'd like to confirm about this tooling, is it:
> 
> 
> 
> - use BOTH package-link and a docker volume mount
> 
> 
> 
> or is this package-link workflow designed to work without an extra docker volume mount for the api-plugin you're working on?

When testing originally, I did not have to add the additional volume mount. I followed the example workflow here and was able to link and unlink: https://github.com/reactioncommerce/reaction/pull/6238 I only did a single time, so did not run into the issues of #6250

kieckhafer:
I’m going to hold off on the 3.8.0 release until this is merged. The only big change in that release is the original PR which this current PR is fixing, so makes sense to wait. cc @mdelreal

loan-laux:
I gave it another try with @aldeed's latest changes and didn't run into the issues mentioned in #6250 anymore. Been using `bin/package-link` for the whole day and it seems very stable. On my end, seems like this is good to go.

kieckhafer:
I've also tested this out and it seems to be good. @focusaurus have you had a chance to try it out?

focusaurus (approved):

focusaurus:
Yeah I've used this a few times. I'm still not 100% clear on the workflow and the various levels of technical challenges and solutions that are in play here, but it seems to work OK for me.

PR #6271, opened 2020-08-02
chore: add placholder email env var

ghost:
Signed-off-by: Josh Habdas <jhabdas@protonmail.com>

Resolves: No issue created. If you need a backing issue let me know.
Impact: **minor**
Type: **chore**

## Issue
[Docs reference](https://docs.reactioncommerce.com/docs/email-admin#configuring-your-email-provider) setting an environment variable called `MAIL_URL` for sending emails. However, there's no placeholder and given the broader project refactoring going on it's not clear at first where to set the mail URL.

## Solution
Add the placeholder URL mentioned in the docs.

## Breaking changes
n/a

## Testing
1. Attempting to create an account for a shop owner from Admin without email configured.
2. Observe no email is received by the recipient.
3. Update the `MAIL_URL` with a valid value and try again.
4. Observe transactional emails are working and status updates in Admin > Settings > Email.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,4 @@
 MONGO_URL=mongodb://mongo.reaction.localhost:27017/reaction
 ROOT_URL=http://localhost:3000
 STRIPE_API_KEY=YOUR_PRIVATE_STRIPE_API_KEY
+MAIL_URL=smtp://username:password@example-mail-host.com:465
--- a/package-lock.json
+++ b/package-lock.json
@@ -1784,9 +1784,9 @@
       }
     },
     "@reactioncommerce/api-plugin-authorization-simple": {
-      "version": "1.2.1",
-      "resolved": "https://registry.npmjs.org/@reactioncommerce/api-plugin-authorization-simple/-/api-plugin-authorization-simple-1.2.1.tgz",
-      "integrity": "sha512-L9vOQlOn+RG7My+qyyFAHuv36OcNApsiwFK6PS8+BZqh0rFHfD9gDwuDio1y/aeZn6HDd8zImDBWL2L+gzwubw==",
+      "version": "1.3.0",
+      "resolved": "https://artifactory.rsglab.com:443/artifactory/api/npm/npm/@reactioncommerce/api-plugin-authorization-simple/-/api-plugin-authorization-simple-1.3.0.tgz",
+      "integrity": "sha1-JVGhesfbKLyewRsjjpl3IwQvx/s=",
       "requires": {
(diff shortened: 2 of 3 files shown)


kieckhafer:
Thanks for the PR @balibebas 

Looking at this a bit, the [docs you link to](https://docs.reactioncommerce.com/docs/email-admin#configuring-your-email-provider) do seem to point to the location where the ENV variable should be added (`<PROJECT_ROOT>/reaction/.env`). We also call our `MAIL_URL` in the `Environmental Variables` section of the docs: https://docs.reactioncommerce.com/docs/environment-variables

I'd be inclined to leave the `env` file as is, and not fill it up with placeholders. @rosshadden thoughts?

rosshadden:
I agree with @kieckhafer. Consider this excerpt from the environment variables docs he links above:

> `MAIL_URL`
> No default. Set this in .env if you want emails to be sent while you are developing or testing.

If we were to provide a default placeholder setting our code would have to be changed to check to make sure it's not the default placeholder value before sending, which then tightly couples it to the default placeholder used and would not be very clean.

With that being said, I emphasize with and appreciate the frustration in not knowing what you don't know, and feeling like you are expected to read every bit of docs ever. Something I think could be a great compromise is to add an empty placeholder:

```
MAIL_URL=
```

This makes it clear something is needed if you want email support, while not tightly coupling any code with default configuration.
What do you two think?

Edit: Or a commented out default value, so the format is communicated.

kieckhafer:
Ross's suggestions of a blank or commented out ENV are OK with me.

We could also add a line at the top of the file directing people to the ENV docs:

`# For examples and information on available ENV variables, see our docs: https://docs.reactioncommerce.com/docs/developer-environment-variables`

That would allow us to update just the docs, and not the `env` file with example / placeholder values in the future.

ghost:
Thanks to both of you for having a look.

I'm still getting up to speed and wasn't aware of the [Enviornment Variables](https://docs.reactioncommerce.com/docs/environment-variables) document until Erik pointed it out. Had I seen that I likely wouldn't have opened this. As Ross suggested, I didn't know what I didn't know.

What threw me was seeing `MAIL_URL` in the version history of the `.env.examples` file while I was reviewing the source (of thruth). That and the fact there _is_ a placeholder for Stripe in the current file - which left me wondering if the documentation I was looking at might have been outdated (which I eventually figured out was not after some FUD). Knowing what I know now I feel the change is too small to warrant a doc update or potential change in anyone's workflow.

Thanks again fellas. :smile:

PR #6289, opened 2020-09-24
Node14 esmodule fix

kieckhafer:
update filename suffix to be cjs


kieckhafer:
closing for funky DCO stuff. #6291 will replace.

PR #6293, opened 2020-09-25
release v3.10.0

kieckhafer:
# v3.10.0

Reaction v3.10.0 adds minor features and performance enhancements, and contains no breaking changes since v3.9.0

### Update node from v12.x to v14.x

Update Reaction to current stable node for general maintenance.

Please note that node's ES Modules support is stable, but still experimental. While we are not aware of any specific breaking changes or issues, some CommonJS/ES Modules refactoring may be necessary in some installations.

## Features

feat: update to nodejs v14 ([#6292](https://github.com/reactioncommerce/reaction/pull/6292))


CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+# v3.10.0
+
+Reaction v3.10.0 adds minor features and performance enhancements, and contains no breaking changes since v3.9.0
```
focusaurus, line 3: I think either in this heading or under notable changes we should mention that node's ES Modules support is stable but still experimental and while we are not aware of any specific breaking changes or issues, some CommonJS/ES Modules refactoring may be necessary in some installations.

package.json
```diff
@@ -148,5 +151,6 @@
       "types-are-capitalized",
       "types-have-descriptions"
     ]
-  }
+  },
+  "snyk": true
```
focusaurus: Do we have a solid convention here? I myself am skeptical of snyk install-time scripting and patching so I usually just try to get the depdendencies themselves updated but don't let snyk mess with this file.

focusaurus (changes requested):

CHANGELOG.md
kieckhafer: What would need to be changed, if we supported it in its experimental form before, do any changes here make it so you can or can't use it any differently than 3.9?

I'm fine calling it out, but just wondering if nothing changes officially, do we need to call it out?

package.json
kieckhafer: We don't have precedent for this.

This is the PR it was introduced with, the language made it seem required to keep the snyk vulnerability testing going, which is why I went with it: https://github.com/reactioncommerce/reaction/pull/6283#issue-488946841

CHANGELOG.md
focusaurus: I think I'm just nervous that this may be one of those "supposed to be compatible" upgrades but in the real world does cause breakages. This is mostly anxiety on my part. Going to mark this resolved so things can progress.

kieckhafer: That's fair. It doesn't hurt to add extra info, was just wondering. I'll add it.

focusaurus (approved):
:shipit:

manueldelreal (approved):
:shipit:

PR #6296, opened 2020-10-12
Update @reactioncommerce dependencies

loan-laux:
Signed-off-by: Loan Laux <loan@outgrow.io>

Impact: **minor**
Type: **chore**

## Issue
`@reactioncommerce/api-plugin-accounts` and `@reactioncommerce/api-plugin-authorization-simple` are outdated.

## Solution
Update the packages to their latest versions.

## Breaking changes
None.


package.json
```diff
@@ -21,47 +21,47 @@
   },
   "dependencies": {
     "@reactioncommerce/api-core": "~1.4.0",
-    "@reactioncommerce/api-plugin-accounts": "~1.4.4",
-    "@reactioncommerce/api-plugin-address-validation-test": "~1.0.0",
+    "@reactioncommerce/api-plugin-accounts": "^1.5.0",
```
mpaktiti: Can you change this to use a tilde `~` range like the rest?

package.json
```diff
@@ -21,47 +21,47 @@
...
     "@reactioncommerce/api-plugin-address-validation": "~1.3.1",
+    "@reactioncommerce/api-plugin-address-validation-test": "~1.0.0",
     "@reactioncommerce/api-plugin-authentication": "~1.2.0",
-    "@reactioncommerce/api-plugin-authorization-simple": "~1.3.0",
+    "@reactioncommerce/api-plugin-authorization-simple": "^1.3.1",
```
mpaktiti: Same as above, we should use a tilde range like the rest of the dependencies.

mpaktiti (changes requested):
Thank you for this @loan-laux! I left a couple of comments regarding the semver ranges.

loan-laux:
Thanks for catching that @mpaktiti. I'm used to using `^` on other projects so I didn't think of checking what you were using in the `package.json`. Should be all good now!

mpaktiti (approved):

loan-laux:
Thanks @mpaktiti. Would be amazing if we could get a minor release for this one!

PR #6312, opened 2020-12-09
chore: update api-plugin-accounts package

mikemurray:
Resolves n/a
Impact: **minor**
Type: **chore**

## Issue

Migrations were broken due to an old version of the `api-plugin-accounts` package

## Solution

Update account package

## Breaking changes

none

## Testing
1. Migrations should run
2. ...and the app should run!


manueldelreal (approved):
:shipit:

PR #6326, opened 2021-05-06
feat: using account-js for auth

Akarshit:
BREAKING CHANGE: hydra and idenity are no longer shipped by default

Signed-off-by: Akarshit Wal <akarshitwal@gmail.com>

## Testing
1. Pull in the storefront branch corresponding to accounts-js
2. Use the reaction-dev-platform to start the containers
3. Try to login on the example-storefront and make sure the new API for auth is working.


Akarshit:
The basic problem is that we are not running `rs.initiate()` because of which the mongo server is stuck at secondary. I couldn't find a easy way to do this. Will look again next week.

PR #6330, opened 2021-05-28
Mark `@reactioncommerce` dependencies for auto minor updates

### LINKED ISSUE #6328: Let plugins feat update happen automatically.
## Current System:
Right now all the plugins in `package.json` have `~` before package version, which mean they only get patch updates.

### Cons:
1. Once we do a minor upgrade of a plugin, any patch update after that doesn't get fetched by reaction (because the patch update is only on the latest version)
2. Makes the release process hard since someone has to look for the latest version of plugin and make a PR into reaction.
3. Make the process of merging features and getting them deployed long because of the above reason.

### Pros:
1. Since the release is QAed, we are confident that the release works.
2. If someone messes up something in a plugin, the main product won't be harmed.

## Proposal:

Let's replace `~` with `^` and allow minor upgrades to plugins. This will help deliver features directly to the users faster.
But it would come with more responsibility of making sure that plugins are not broken while merging PRs.

In case they are broken by a PR, we can do a patch release to fix the issue and hopefully minify the damage.

Please let me know what people think about this in the comments.

loan-laux:
Signed-off-by: Loan Laux <loan@outgrow.io>

Resolves #6328
Impact: **critical**
Type: **chore**

Please see #6328 for initial discussion. This PR would allow auto minor updates on all `@reactioncommerce` production dependencies.


Akarshit:
I will keep this PR sitting for a while. Basically I am waiting for the release of v4.0 and just get more discussion/validation around this.

loan-laux:
Sure @Akarshit. That's what I intended this one for!

Akarshit:
@loan-laux I got the green flag to do this as part of `v4.0`

Akarshit:
@loan-laux I see that the tests are failing in this one. They shouldn't be but I suppose there is some change in a plugin that wasn't scrutinized properly causing this. 
Would it be possible to update the tests along with this PR so that we have a working branch.

loan-laux:
@Akarshit I've fixed the query tests, but the mutation tests keep failing. It's weird because they complain about the shop type being required as part of `CreateShopInput`, but:

1. We pass a `type` value in these tests anyways
2. The `type` is not required in the GraphQL schema

So I'm a bit puzzled here. Don't have more time today to dive in but I'll come back to it soon. Let me know if you have any thoughts on what's happening here. I'm sure it's something simple I'm missing.

loan-laux:
I see that there's a step in the CI/CD pipeline that restores cache for `node_modules` from a previous build... Is that the case? And if so, can we disable that and make sure that the dependencies are re-installed every time? Pretty sure these `createShop` tests are failing because the CI doesn't have the latest version of the `api-plugin-shops` package.

Akarshit:
I will get this fixed today.

Akarshit:
Looks like reactioncommerce/api-plugin-shops#30 should fix this.

loan-laux:
@Akarshit I re-ran the build and we're still getting the same error. Pretty sure there's a caching issue with CircleCI, on top of that problem you've fixed with the shop plugin. Is there any way to force it to reinstall the dependencies at every build?

Akarshit:
I think we need to remove the cache so that all the PRs are tested with the latest version of plugin. That change should be a part of this PR.

Akarshit (dismissed):

Akarshit (approved):

PR #6334, opened 2021-06-07
feat: doing semantic releases

Akarshit:
Right now the process is to to release manually. With this PR we plan to shift to having semantic release of reaction, so that whenever a new PR is merged in, we have a docker image on dockerhub and also a new release tagged on github.
Still the new release should only be publicly used as a part of reaction-development-platform upgrade. That way we will get a chance to QA the before making a new release in reaction-development-platform.

## Testing
I have tested this with `--dry-run` flag and it works for the semantic release part. 
For the pushing image to dockerhub, I have tested and the image is being pushed with the correct name.
Though we would know the final end to end run only after this is merged and the CI is run on trunk.


loan-laux:
Excited about this! 🎉

MohanNarayana (approved):
👍

Akarshit:
🤞

PR #6336, opened 2021-06-10
fix: update docker-compose on release

### LINKED ISSUE #6329: Push image with version to dockerhub on every merge
## Current Process
All PRs get merged to trunk, when we want to make a release, we create a new branch, build image from that branch, tag it and push it to docker.

### Cons:
1. This is heavy process for releasing.
2. This slows down the release, since we have to do manually.

### Pros:
1. This let's us choose when we want to release

## Proposal
Introduce sematic-release type releasing. So on every merge, a new version is tagged and pushed to dockerhub. Note, this won't make the new version to  be available to the public for immediate consumption. This will just have the images on dockerhub.
The actual "release" only happens when we upgrade the version in reaction-development-platform. So once we are ready to do the release, we only create a PR in `reaction-development-platform` and QA that. If everything looks well we can release that.
So even if we merge something that break the software, it will still pass through QA before being released.

Akarshit:
Resolves #6329 
Impact: **major**
Type: **chore**

## Issue
The current release process through semantic release was not updating the docker-compose.yml and package.json and thus the "release" on github was always stuck with the same image version on dockerhub.

## Solution
This PR updates the `docker-compose.yml` along with `package.json` with the correct version number for the tagged release on github.
## Breaking changes
None

## Testing
The PR has been tested locally and was pushing the right versions. This has also been tested with the `--dry-run` flag and everything looked fine. But with such PRs there is always a risk of something going wrong with permissions and thus failing. The plan to address those is to fix the issue and create a next patch release.


rondlite:
that should work 👍

MohanNarayana (approved):

Akarshit:
I hope this works 🤞

PR #6342, opened 2021-08-04
fix: update api-plugin-files plugin

Akarshit:
This PR updates the `api-plugin-files` so that we can have the latest `sharp` in production images.


MohanNarayana (approved):
Looks good

PR #6348, opened 2021-08-18
feat: update Stripe SCA payments plugin

### LINKED ISSUE #5304: Refactor the Stripe Card Payment Plugin to use their Payment Intents API and some general considerations
## Issue Description
Currently the Stripe Form executes a call to stripe from the client to create a token:
[/reaction-component-library/src/components/StripePaymentInput/v1/StripePaymentInput.js](https://github.com/reactioncommerce/reaction-component-library/blob/d3ca15bf3c586c8070900d5ddaa3ef088d27be72/package/src/components/StripePaymentInput/v1/StripePaymentInput.js#L78)

These kind of integrations will stop working reliably in the EU from September 14 2019 due to the new [Strong Customer Authentication](https://stripe.com/docs/strong-customer-authentication) rule coming into effect.

The current Token approach is considered outdated, while the new [Payment Intents AP](https://stripe.com/docs/payments/payment-intents/quickstart)I meets the requirement of the SCA.

We should therefore refactor the Stripe payments to this new API.
From what I got so far that kind of integration works similar to the way I did the Paypal integration. The token can be created on the backend and transported to the client using a graphql mutation.

What could be tricky: With Paypal I could trigger such a mutation upon clicking the button, therefore I could always be sure to have the latest total and trigger the checkout from there.

Filling the stripe credit card  form and generating a token doesn't exclude the possibility of changes to the card happening afterwards. While the Payment Intents API takes care of such scenarios by providing an endpoint to update the `amount` of a token, I'd prefer not to make the backend logic too complex just for this provider, or at least see if it makes sense to settle for a generic solution for payment providers with updated cart totals similar to the existing authorize/capture functions on context?

While every payment plugin could just use the cardUpdate appEvents for this currently, a separated function might help to structure things in an easier way?

delagroove:
Signed-off-by: Alvaro Bueno <alvaro.bueno@mailchimp.com>

Resolves #5304
Impact: **major**
Type: **feature**

## Issue

Stripe needs work with SCA for european users.

## Solution

Integrate stripe SCA by default as a payment system

## Breaking changes

None

## Testing

1. set private key in .env
2. set public key in storefront .env
3. activate payment in admin > settings > payment
4. activate a shipment method
5. checkout in storefront and go to step 3 (payment)


jest.config.cjs
```diff
@@ -26,6 +26,7 @@ const externalNodeModules = [
   "@reactioncommerce/api-plugin-orders",
   "@reactioncommerce/api-plugin-payments-example",
   "@reactioncommerce/api-plugin-payments-stripe",
+  "@reactioncommerce/api-plugin-payments-stripe-scas",
```
Akarshit: typo here

PR #6377, opened 2021-10-14
fix: add check for org name/package name format

MohanNarayana:
Signed-off-by: Mohan Narayana <mohan.narayana@mailchimp.com>

Resolves #6377 
Impact: **minor**
Type: **bugfix**

## Issue

When linking a package without providing the account name (the @accountname/...) will end in an error when the API restarts.
the command does not provide a informative error message when this happens

## Solution

Validate the input format and check if package is valid.

## Testing

Test values:
Invalid - @reaXioncommerce/api-plugin-example
               @reactioncommerce/not-a-plugin
               @reactioncommerceapi-plugin-example
               @reactioncommerce/
               /api-plugin-example

Valid - @reactioncommerce/api-plugin-example

### DIFF
--- a/bin/package-link
+++ b/bin/package-link
@@ -19,6 +19,12 @@ yalc_packages_file_name="yalc-packages"
 container_id="$(docker-compose ps -q api)"
 container_copied_packages_path="/home/node/copied-packages"
 
+org_name='@reactioncommerce'
+if [[ $package_name != $org_name* ]]; then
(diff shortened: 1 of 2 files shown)


Akarshit:
@MohanNarayana I think will stop people from being able to link their internal plugins. We should get on a call and figure out a better way to implement this.

delagroove (dismissed):

delagroove (approved):

PR #6393, opened 2021-12-15
feat: add dotenv package to allow .env files natively

brent-hoover:
Impact: **minor**
Type: **feature**

## Issue

Currently the root app does not respect .env files if running directly, only through docker. This means bootstrapped package cannot use them.

## Solution

Add the dotenv package

## Breaking changes

None

## Testing

Everything should operate as before


delagroove (dismissed):

brent-hoover:
Closing. Will reopen new PR if necessary

PR #6415, opened 2022-03-01
feat: add api-plugin-prometheus-exporter

MohanNarayana:
Signed-off-by: Mohan Narayana <mohan.narayana@mailchimp.com>

You need this for api-plugin-prometheus-exporter


brent-hoover:
@MohanNarayana I don't think this should be included as part the default Reaction build. This should be an optional plugin

MohanNarayana:
Yes, this is just to put the current progress somewhere. The final plugin will be designed to be able to add it optionally

brent-hoover:
Closing as we aren't adding this to default set

PR #6430, opened 2022-04-27
chore: bump eslint versions to be the same as eslint-config

brent-hoover:
Versions were bumped in the eslint package but not bumped here so running `npm install` gives you a dependency error


Akarshit (approved):

brent-hoover:
Work completed in another PR

Issue #6389, opened 2021-12-07
Restructure repo into Monorepo


brent-hoover:
Restructure this repo so that all packages (including library packages like Logger and Error) are all contained in this repo and published upon merge of PR.

This will do a lot to make this easier to manage, understand, and maintain.

The main challenge here is to incorporate the existing workflow of using Semantic Release along with auto-publishing packages.

**Definition of Done:**

1. All Reaction packages (including lib plugins like Logger and Error) live inside a folder called "packages"
1. Whenever a PR is merged into a package a new version gets published via semantic release exactly like it does now
1. Whenever a plugin is updated, a new version of reaction is released semantically unless the package bump is breaking (e.g. a MAJOR change under semantic release)
1. The build time (to run reaction) doesn't increase(at-least significantly).

**Background info:**

1. An analysis of the different options available is located [here](https://monorepo.tools/)
2. Some feedback from @aldeed 

```
I agree that a mono repo can be nice for this case, and I also think you want to avoid incrementing and publishing all packages when only some have changed. I’ve struggled with Lerna enough to not love it as a solution.

I’d probably lean toward a bash or node script custom for what needs to happen, with each package folder managing its own scripts.
I am not sure exactly what it would be on CircleCI, but on GitHub Workflows I have had success with this config:
jobs:
  project-changed:
   runs-on: ubuntu-latest
   outputs:
    didChange: ${{ steps.projectDiff.outputs.count > 0 }}
   steps:
   - uses: actions/checkout@v2
   - name: Determine whether package files changed
     id: projectDiff
     uses: technote-space/get-diff-action@v6
     with:
      PATTERNS: |
      api-plugin-whatever/**

  publish:
   runs-on: ubuntu-latest
   needs: project-changed
   if: ${{ needs.project-changed.outputs.didChange == 'true' }}

You would need a separate workflow file for each package, but the “publish” job (or any job that you make dependent on the “project-changed” check) would not run. So that would limit it to only running checks and publishing packages with changed files.

That is using this: https://github.com/technote-space/get-diff-action

Then the actual `publish` job would do something similar to what each repo does for the publish step right now, using semantic-release to determine the version and publish it.
As for being able to figure out which commits go with which packages for the semantic version bump, have you looked at this? https://www.npmjs.com/package/semantic-release-monorepo
```

Akarshit:
I got the KT for this from Alvaro. Have started working on this.

Akarshit:
[...]

PR #6440, opened 2022-08-25
feat: monorepo

vanpho93:
Resolves: #6389
Impact: minor
Type: feature

## Issue

Too many repos.

## Solution

Using changeset and npm workplace to convert things into monorepo

If you're solving a UIX related issue, please attach screen-caps or gifs showing how your solution differs from the issue.


.changeset/config.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@2.1.1/schema.json",
+  "changelog": ["@changesets/changelog-github", { "repo": "vanpho93/reaction-monorepo" }],
```
aldeed: Need to change to correct `repo`

.circleci/config.yml
```diff
@@ -53,7 +59,9 @@ jobs:
             - reaction-v6-node-modules-{{ .Branch }}
       - run:
           name: Lint JavaScript
-          command: npm run lint
+          command: |
+            npm run build:packages
+            npm run lint
```
aldeed: Is building necessary to run eslint and tests? Usually it is not.

.circleci/config.yml
```diff
@@ -125,10 +149,13 @@ jobs:
           keys:
             - reaction-v6-node-modules-{{ checksum "package.json" }}-{{ checksum "package-lock.json" }}
             - reaction-v6-node-modules-{{ .Branch }}
+      - run:
+          name: Authenticate with registry
+          command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc
```
aldeed: This will overwrite the contents of `.npmrc`. You might want to do `> ~/.npmrc` to avoid losing the contents of the repo file, and/or do `>>` instead of `>` to append to the existing file contents.

.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Install Dependencies
+        run: npm install
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }} # Ensure to have this set up under GitHub secrets
+      - name: Create Release Pull Request or Publish to npm
+        uses: changesets/action@v1
+        with:
+          publish: npm run release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # Ensure to have this set up under GitHub secrets
```
aldeed: You don't set `GITHUB_TOKEN` in secrets. Instead, you add `permissions` object on this workflow job, and then `GITHUB_TOKEN` will be automatically available with those permissions. Docs: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

package.json
```diff
@@ -1,51 +1,21 @@
 {
-  "name": "reaction-api",
-  "version": "4.2.0",
+  "name": "reaction",
+  "version": "1.0.0",
```
aldeed: Need to decide what version this should be. Or omit the version property.

package.json
```diff
@@ -1,51 +1,21 @@
...
-          ],
-          "message": "chore(release): v${nextRelease.version} [skip ci]"
-        }
-      ]
-    ]
+  "main": "index.js",
+  "scripts": {
+    "start:dev": "npm run start:dev -w=apps/reaction",
+    "start:meteor-blaze-app": "npm run start -w=apps/meteor-blaze-app",
+    "build:packages": "npm run build -w=packages/file-collections -w=packages/file-collections-sa-base -w=packages/file-collections-sa-gridfs -w=packages/logger -w=packages/random",
```
aldeed: I think it would be better to do `npm run build --if-present --ws` so that it will build any new packages automatically.

package.json
```diff
@@ -1,51 +1,21 @@
...
+    "test": "npm run test --if-present --ws",
+    "test:integration:query": "npm run test:integration:query --workspace=apps/reaction",
+    "test:integration:mutation": "npm run test:integration:mutation --workspace=apps/reaction",
+    "changeset": "changeset",
+    "release": "changeset publish",
+    "lint": "eslint . --resolve-plugins-relative-to .",
+    "lint:docker": "npx --quiet --package @reactioncommerce/ci-scripts@1.6.2 lint-dockerfiles --workspace=apps/reaction",
+    "lint:gql": "npm run lint:gql -w=apps/reaction -w=packages/api-core -w=packages/api-plugin-surcharges",
```
aldeed: I think it would be better to do `npm run lint:gql --if-present --ws` so that it will lint GraphQL for all packages that have it set up.

aldeed (changes requested):
Looks pretty good! I left some inline comments, and you can delete .circleci and .github folders from every package/app folder.

.github/workflows/release.yml
vanpho93: We don't manually set GITHUB_TOKEN in secrets, this is just get the value only.

apps/reaction/tests/integration/api/mutations/addTag/addTag.test.js
```diff
@@ -154,14 +154,11 @@ describe("authorized user", () => {
   });
   test("cannot create tag with existing slug", async () => {
     const logLevel = Logger.level();
-    const duplicateTagSlug = Factory.Tag.makeOne({
```
brent-hoover, line 157: Confused why this change is here?

apps/reaction/tests/integration/api/mutations/updateTag/updateTag.test.js
```diff
@@ -182,7 +182,7 @@ describe("authorized user", () => {
       slug: "tag-slug"
     });
     Logger.level("FATAL");
-    expect.assertions(1);
+    // expect.assertions(1);
```
brent-hoover: Why is this expect removed?

packages/api-core/package.json
```diff
@@ -50,7 +51,7 @@
   "dependencies": {
     "@apollo/federation": "^0.15.0",
     "@graphql-tools/merge": "^8.0.1",
-    "@reactioncommerce/api-utils": "^1.12.0",
```
brent-hoover: Is this actually what we want here?

packages/api-plugin-files/package.json
```diff
@@ -21,24 +22,22 @@
...
   "dependencies": {
     "@reactioncommerce/api-utils": "^1.16.7",
-    "@reactioncommerce/file-collections": "~0.9.3",
+    "@reactioncommerce/file-collections": "^0.9.3",
     "@reactioncommerce/file-collections-sa-gridfs": "~0.1.5",
     "@reactioncommerce/logger": "~1.1.4",
     "@reactioncommerce/random": "~1.0.2",
     "@reactioncommerce/reaction-error": "~1.0.1",
-    "envalid": "~5.0.0",
+    "envalid": "^7.3.1",
```
brent-hoover: We should be keeping the version of envalid consistent at 6.x.x

packages/api-plugin-inventory-simple/src/index.js
```diff
@@ -45,6 +45,7 @@ export default async function register(app) {
     },
     mutations,
     policies,
-    queries
+    queries,
+    additionalFields: { test: 4 }
```
brent-hoover: I think this is leftover test code?

packages/api-plugin-pricing-simple/src/index.js
```diff
@@ -45,6 +45,7 @@ export default async function register(app) {
     },
     simpleSchemas: {
       PriceRange
-    }
+    },
+    additionalFields: { test: 4 }
```
brent-hoover: More test code?

packages/api-utils/package.json
```diff
@@ -70,8 +71,7 @@
     "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-react-hooks": "^3.0.0",
     "graphql": "^15.0.0",
-    "jest": "^26.0.1",
-    "semantic-release": "^17.0.4"
+    "jest": "^26.0.1"
```
brent-hoover: Are we keeping the version of this consistent?

brent-hoover (commented):
Just have a few questions

apps/reaction/tests/integration/api/mutations/addTag/addTag.test.js
vanpho93: Fixed by adding shopId. This change caused by an unique index changed from `slug: 1` into `slug: 1, shopId: 1`.

packages/api-utils/package.json
vanpho93: There are many inconsistent dependencies, I am not sure if we could solve them all in this PR.

.circleci/config.yml
vanpho93: image
Without `build:packages` there will be an error. Maybe there's a way around to fix this.

packages/api-utils/package.json
brent-hoover: Yeah, but we're touching this one in this PR

brent-hoover: Sorry, misread the diff. Disregard the comment. But we should do a push to make all common dependencies consistent.

tedraykov:
The Circle CI run doesn't seem to be completely configured. We don't want to run linting and unit testing on every package on every commit, because it is redundant. Most CI/CD tools provide some sort of path filtering that run jobs only for the changed directories. 

Another concern I have is if we want to release a new docker image for `reaction`, we must be sure the integration tests that are running before the `docker-build-push` are using the newly published api-plugins.
Let's say we update the `orders` plugin and we run `npm run changeset` to generate the changesets. The changeset bot will generate a PR that bumps the `orders` plugin version and also update that version in the `reaction` api. When we merge this `release` PR, the circle ci and the github actions runs will start in parallel. A problem that might arise is when circle ci tries to build the `reaction` app that depends on the new `orders` plugin version, that might not be published by the github action job.

.github/workflows/release.yml
aldeed: I was referring to the comment, which can be removed. And also I didn't see a `permissions` section in the yaml at the time, but maybe the default permissions are fine.

packages/api-utils/package.json
vanpho93: I updated jest and eslint. i think we can do more later.

vanpho93:
> The Circle CI run doesn't seem to be completely configured. We don't want to run linting and unit testing on every package on every commit, because it is redundant. Most CI/CD tools provide some sort of path filtering that run jobs only for the changed directories.

This's an improvement but we can do it later, there is a (tool)[https://turborepo.org/] that @aldeed suggested which could help.

> Another concern I have is if we want to release a new docker image for reaction, we must be sure the integration tests that are running before the docker-build-push are using the newly published api-plugins.

Integration tests run outside of docker.
There are still problem when a plugin upgrade trigger reaction app upgrade. In that case, `changeset` will update only package.json but not package-lock.json. I am trying to fix this.

@tedraykov @zenweasel

aldeed (changes requested):
I'm not sure about CircleCI, but for GitHub Actions, there is a good action plugin, `technote-space/get-diff-action`, that you can use to check folder diffs before running additional per-project checks.

```yaml
jobs:
  project-changed:
    runs-on: ubuntu-latest
    outputs:
      didChange: ${{ steps.projectDiff.outputs.count > 0 }}
    steps:
      - uses: actions/checkout@v3
      - name: Find changed files
        id: projectDiff
        uses: technote-space/get-diff-action@v6
        with:
          # List files and/or folders that should trigger remaining jobs when changed
          PATTERNS: |
            packages/package/**

  next-job:
    runs-on: ubuntu-latest
    # Skip if no project files have changed
    needs: project-changed
```

Note that it usually isn't correct to use the CI `paths` feature because that only checks the most recent commit, whereas you would want to run jobs if _any_ commit included in a PR has modified files in that folder.

My other concerns are:
- I don't know if shrinkwrap works with all of this. `api-plugin-authentication` is the only package that uses shrinkwrap. It's being used to pin a subdependency to an older version. You should remove the shrinkwrap file and instead use the newer [overrides](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) feature.
- To gain full advantages of the monorepo, you should migrate npm locks to the root package-lock and remove package-lock files from all packages.
- Did you test whether developing locally with multiple layers of package dependencies is working? For example, if running the app, which depends on `file-collections-sa-gridfs`, which in turn depends on `file-collections-sa-base`, can you edit `file-collections-sa-base` with hot reloading?

aldeed (changes requested):
A few more suggestions...

In the root `package.json`, add:

```json
"engines": {
    "node": ">=14.18.1",
    "npm": ">=7"
  },
  "engineStrict": true,
```

To support NPM v8.6+, run the following command and commit ([explanation](https://github.com/npm/cli/issues/4664#issuecomment-1086096726)):

```txt
npm config set legacy-peer-deps=true --location=project
```

(For this repo, there's a conflicting `graphql` version for which peer warnings were previously ignored. Eventually you could update packages to make everything happy and then migrate to strict peer deps.)

When I run `npm run start:dev`, there is a package error `Cannot find module 'json-bigint'`, which is coming from a dependency of `tus-node-server`. You can fix it by updating `tus-node-server` to `"~0.7.1"` in `packages/file-collections/package.json` and then re-run `npm install` in repo root. It seems like there are not any breaking changes with this update, but you should re-test file uploads to be sure. You may also need to update to a newer `tus-js-client`, which might have breaking changes.

In the new section of the readme, you are missing two steps for being able to run the app:
- Copy `apps/reaction/.env.example` to `apps/reaction/.env`.
- Start MongoDB. Maybe you can add a `docker-compose.yml` file in the repo root that just starts mongo, so they can run `docker-compose up -d` before `npm run start:dev`? It would be similar to the `docker-compose.yml` file in `apps/reaction`, but without the app service. You would also have to change the MONGO_URL to `mongodb://localhost:27017/reaction` in the `.env.example` file.

brent-hoover:
If we add a docker-compose with Mongo, we could use the one from the CLI and the CLI could grab it from here

README.md
```diff
@@ -60,6 +60,44 @@ When make completes, three services will be running on localhost:
 
 Go to the complete [installation instructions](https://mailchimp.com/developer/open-commerce/guides/quick-start/#access-the-dashboard-playground-and-storefront) to see how to set up your store
 
+# API development
+
+## What you need
+* NPM >= 7 (support workspaces)
+* Docker & docker-compose
+* MongoDB
```
brent-hoover: This is incorrect, you don't need Mongo installed if you are using the docker-compose and most people are better of not installing it locally so you can have different instances of it

.github/workflows/release.yml
```diff
@@ -12,21 +12,30 @@ jobs:
...
+
       - name: Use Node.js 14.x
         uses: actions/setup-node@v2
         with:
           node-version: 14
 
-      - name: Install npm 8.1.0
-        run: npm i -g npm@8.1.0
+      - name: Install pnpm
+        run: npm i -g pnpm@latest
```
brent-hoover, line 31: I think we had agreed to pin npm at a particular version? There has been problems with it pulling in a breaking version of npm

packages/api-plugin-sample-data/package.json
```diff
@@ -10,6 +10,7 @@
   "homepage": "https://github.com/reactioncommerce/reaction",
   "url": "https://github.com/reactioncommerce/reaction.git",
   "email": "hello-open-commerce@mailchimp.com",
+  "private": true,
```
brent-hoover, line 13: I think this disables publishing to npm, which is something we do want?

packages/api-core/src/importPluginsJSONFile.js
```diff
@@ -41,7 +41,11 @@ export default async function importPluginsJSONFile(pluginsFile, transformPlugin
       Logger.debug({ pluginPath, pluginRefs });
       throw new Error(`Plugin "${name}" is not set to a string`);
     } else if (/[a-zA-Z@]/.test(pluginPath[0])) {
-      ({ default: plugin } = await import(pluginPath));
+      try {
```
brent-hoover, line 44: What's the reasoning for this?

brent-hoover (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Install Dependencies
+        run: pnpm install --ignore-scripts
+
+      - name: Create Release Pull Request or Publish to npm
+        uses: changesets/action@v1
+        with:
+          publish: pnpm release
```
tedraykov, line 39: I'm skeptical this would work as intended. First of all there is no pnpm command called release, also there are no scripts called `release` in the plugins. When we merge this in trunk, are we sure what part of the repository will be published?

Dockerfile
```diff
@@ -1,5 +1,36 @@
 # Dockerfile for production builds
-FROM reactioncommerce/node-prod:14.18.1-v1
+# syntax=docker/dockerfile:1.4
+
+FROM node:14.18.1-alpine as deps
```
tedraykov: This docker file is better suited for the app folder itself, not In the root of the repo. If you want to build a container image with the root context but use the `apps/reaction/Dockerfile` you can specify this in the `docker build` command like so:
```
docker build -t ... -f ./apps/reaction/Dockerfile .
```

tedraykov (changes requested):

.github/workflows/release.yml
vanpho93: Actually release will be called on circleci on `pnpm publish -r`.

Issue #6471, opened 2022-09-13
As a user, I should be able to checkout with multiple fulfillment types in a single order using the new set of plugins


brent-hoover:
The user should be able to checkout with multiple fulfillment types in a single order using the new set of plugins. The proposed fulfillment types are "shipping" and "pickup" (although "pickup" would be a minimal shell implementation).

Issue #6472, opened 2022-09-13
As an admin I want to be able to migrate my data to the new format required for the new FFT plugins


brent-hoover:
This should be implemented using the current Migration system

PR #6480, opened 2022-09-20
Fulfillment feature

sujithvn:

Resolves #6471 
Impact: **breaking**
Type: **feature**

## Issue

## Solution

## Breaking changes

## Testing

To perform review and testing of the related code changes, we are adding details related to this feature implementation which spans across multiple plugins (each in its own folders in the packages folder in monorepo).

###### Existing plugin changes
api-plugin-products 
api-plugin-catalogs 
api-plugin-carts
api-plugin-orders 

###### Plugins to be removed
api-plugin-shipments
api-plugin-shipments-flat-rate

###### New Plugins to be added
api-plugin-fulfillment
api-plugin-fulfillment-type-shipping 
api-plugin-fulfillment-method-shipping-flat-rate
api-plugin-fulfillment-method-shipping-ups
api-plugin-fulfillment-type-pickup
api-plugin-fulfillment-method-pickup-store

##### CARTS

##### CATALOGS

##### PRODUCTS

##### ORDERS

##### ORDERS (specific to placeOrder refactor and introduction of validateOrder end-point)

##### API-PLUGIN-FULFILLMENT
This is the base plugin, basic features listed below. 
Mutations
* createFulfillmentType - this is for internal call only from ff-type plugins (NEW)
* createFulfillmentMethod  - this is for internal call only from ff-method plugins (NEW)
* updateFulfillmentType - public mutation which allows to update certain specific fields of the ff-type (NEW)
* updateFulfillmentMethod - public mutation which allows to update certain specific fields of the ff-type (NEW)
* updateFulfillmentOptionsForGroup - Existing code/functionality from the old shipment plugin. This updates the cart with quote details from all available ff-methods for the provided ff-type
* selectFulfillmentOptionForGroup - Existing code/functionality from the old shipment plugin. This updates the cart with one of the selected ff-method from the available quote presented by above mutation.

Queries
* getFulfillmentMethodsWithQuotes - Existing code/functionality from the old shipment plugin. Calls the quote providing functions by collecting them via functionByType. Calls are made only for the specific methods of the provided ff-type.
* getFulfillmentType - returns the ff-type matching the provided ID (NEW)
* getFulfillmentTypes - returns all the ff-types (NEW) 
* getFulfillmentMethods - returns all the ff-methods for the specified ff-type (NEW)

Others
* preStartup - extends all impacted schemas mostly with dynamic values for 'allowedValues' (NEW)
* startup - inserts the default 'undecided' entry to the new Fulfillment collection (NEW)
* registration - collects all the "registeredFulfillmentType" from ff-type plugins and adds allRegisteredFulfillmentTypes to context (contextAddition) (NEW)
* util folder - copied over from shipment plugin (Existing code)

##### API-PLUGIN-FULFILLMENT-TYPE-SHIPPING
* Registers the 'shipping' as a ff-type via registeredFulfillmentTypes: ["shipping"]
* Inserts the default entry for shipping ff-type in Fulfillment collection

##### API-PLUGIN-FULFILLMENT-TYPE-PICKUP
* Registers the 'pickup' as a ff-type via registeredFulfillmentTypes: ["pickup"]
* Inserts the default entry for pickup ff-type in Fulfillment collection

##### API-PLUGIN-FULFILLMENT-METHOD-SHIPPING-FLAT-RATE
* Mostly copied over from api-plugin-shipment-flat-rate by making certain fixes to adhere to new structure and new DB collections. Fixes are similar to the details provided for the other 2 ff-method plugins below.
* The mutations & queries present were left as it is and it works slightly differently compared to the new mutations/queries that were introduced in api-plugin-fulfillment (which is used by remaining ff-methods)
* Shipping restrictions feature is copied over from existing code. This feature is not added to other ff-methods.

##### API-PLUGIN-FULFILLMENT-METHOD-SHIPPING-UPS
*  getFulfillmentMethodsWithQuotesShippingUPS - returns the quote or equivalent details for the method when called from base ff plugin
* preStartup - extends the union of "methodAdditionalData" with data structure specific to UPS
* util/calculateUPSRate - dummy function to simulate api providing UPS specific info while returning quotes.
* util/ validateOrderMethodsups - dummy function to simulate UPS specific validations done. Called by prepareOrder.js 

##### API-PLUGIN-FULFILLMENT-METHOD-PICKUP-STORE
*  getFulfillmentMethodsWithQuotesPickupStore - returns the quote or equivalent details for the method when called from base ff plugin
* preStartup - extends the union of "methodAdditionalData" with data structure specific to Store
* util/collectStoreDetails - dummy function to simulate api providing Store specific info while returning quotes.
* util/ validateOrderMethodsstore - dummy function to simulate Store specific validations done. Called by prepareOrder.js 

### Test cases
With the above changes, we should be able to test the following cases:

1. The current code base has 2 ff-types and 3 ff-methods implemented. We could add more by following the same template (pref type-pickup, method-store). New ff-type should register itself similar to registration.js and insert the base entry in Fulfillment collection. New ff-method should have the functionalities that provide quotes/equivalent with additional data if any, validation results (refer method-store-ups).
2. Use "createCart" endpoint to create a new cart and add an item. If the "selectedFulfillmentType" is provided in the input, a fulfillment group of that ff-type will be created. Else the item will be placed in "undecided" group.
3. More items could be added to the cart using the "addCartItems" endpoint. Here also we could optionally include the "selectedFulfillmentType".  All items belonging to one ff-type will be grouped together.
4. We can use endpoint "setFulfillmentTypeForItems" to re-assign the items in "undecided" group to any of the selected ff-group.
5. Use "updateFulfillmentOptionsForGroup" to update the cart with quote details from all available ff-methods for the provided ff-type and then subsequently use "selectFulfillmentOptionForGroup" to select one of them as the selected ff-method for the ff-group.
6. Place the order as usual by providing the "selectedFulfillmentMethodId" for each of the ff-group created. Order should be placed successfully
7. validateOrder endpoint could be used to verify the correctness & completeness of the input to placeOrder prior to actually placing the order.

#### Pending items

The below pending items will be introduced via separate PR/tickets

* Documentation providing details of the Fulfillment features for a UI team
* i18n feature
* Data migration
* Feature to split order resulting in multiple groups for same ff-type
* Enable default ff-type for products


apps/reaction/tests/integration/api/mutations/checkout/checkoutTestsCommon.js
```diff
@@ -186,7 +188,8 @@ beforeAll(async () => {
 
   // Add shipping methods
   mockShippingMethod.shopId = internalShopId;
-  await testApp.collections.Shipping.insertOne(mockShippingMethod);
+  // await testApp.collections.Shipping.insertOne(mockShippingMethod);
```
brent-hoover: Please don't ever comment out code. That's what source control is for

apps/reaction/tests/integration/api/mutations/shippingRates/createFlatRateFulfillmentMethod.test.js
```diff
@@ -28,7 +28,7 @@ const adminGroup = Factory.Group.makeOne({
   _id: "adminGroup",
   createdBy: null,
   name: "admin",
-  permissions: ["reaction:legacy:shippingMethods/create"],
+  permissions: ["reaction:legacy:fulfillmentMethods/create"],
```
brent-hoover, line 31: Make sure you handle these changing permissions in your migrations.

packages/api-plugin-carts/src/mutations/createCart.js
```diff
@@ -4,6 +4,8 @@ import ReactionError from "@reactioncommerce/reaction-error";
 import Logger from "@reactioncommerce/logger";
 import addCartItems from "../util/addCartItems.js";
 
+const cartVersion = "v2"; // Update this for each major change of cart object, v2 => Fulfillment changes
```
brent-hoover: I would remove the "v" so we could do numeric evaluation on it if we wanted to

brent-hoover: Also this feels like it should live somewhere else like in a separate file or something

packages/api-plugin-carts/src/mutations/transformAndValidateCart.js
```diff
@@ -15,7 +15,7 @@ const logCtx = { name: "cart", file: "transformAndValidateCart" };
  */
 export default async function transformAndValidateCart(context, cart) {
   const { simpleSchemas: { Cart: cartSchema } } = context;
-  updateCartFulfillmentGroups(context, cart);
+  await updateCartFulfillmentGroups(context, cart);
```
brent-hoover, line 18: I think you should check because I think transforms expect to get promises but I am not sure

packages/api-plugin-carts/src/simpleSchemas.js
```diff
@@ -571,9 +582,9 @@ const Shipment = new SimpleSchema({
     optional: true
   },
   "type": {
-    type: String,
-    allowedValues: ["shipping"],
-    defaultValue: "shipping"
+    type: String
+    // allowedValues: ["undecided"] // extended with dynamic values in fulfillment plugin startup
```
brent-hoover, line 586: Why is this code commented?

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,124 @@
...
+
+    // This is a new optional field that UI can pass in case the user selects fulfillment type
+    // for each item in the product details page instead of waiting till checkout
+    let { selectedFulfillmentType } = item;
+
     // Do not re-allocate the item if it is already in the group. Otherwise difficult for other code
     // to create and manage fulfillment groups
-    const itemAlreadyInTheGroup = currentGroups.find(({ itemIds }) => itemIds && itemIds.includes(item._id));
-    if (itemAlreadyInTheGroup) return;
+    // Commenting out the below check since the item by default (if no fulfillment type is provided) will be assigned to 'undecided' group
```
brent-hoover: These comments probably belong more in a PR comment than in the actual code. Once this PR is merged the context is lost.

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,124 @@
...
-    } else if (!group.itemIds) {
-      // If there is a compatible group but it has no items array, add one with just this item in it
-      group.itemIds = [item._id];
-    } else if (!group.itemIds.includes(item._id)) {
-      // If there is a compatible group with an items array but it is missing this item, add this item ID to the array
-      group.itemIds.push(item._id);
-    }
-  });
+    // check if the selectedFulfillmentType is an 'enabled' fulfillmentType, if not set is 'undecided'
+    /* eslint-disable no-await-in-loop */
```
brent-hoover, line 113: Is this eslint-disable still relevant?

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,124 @@
...
+    const enabledFulfillmentTypeObjs = await Fulfillment.find({ "shopId": item.shopId, "provider.enabled": true }).toArray();
+    /* eslint-enable no-await-in-loop */
```
brent-hoover, line 115: And this one?

packages/api-plugin-carts/src/xforms/xformCartCheckout.js
```diff
@@ -14,6 +14,7 @@ function xformCartFulfillmentGroup(fulfillmentGroup, cart) {
       displayName: option.method.label || option.method.name,
       group: option.method.group || null,
       name: option.method.name,
+      methodAdditionalData: option.method.methodAdditionalData || { gqlType: "emptyData", emptyData: false },
```
brent-hoover, line 17: Can you explain this `gqlType` code here?

packages/api-plugin-carts/src/xforms/xformCartCheckout.js
```diff
@@ -64,8 +66,8 @@ function xformCartFulfillmentGroup(fulfillmentGroup, cart) {
     selectedFulfillmentOption,
     shippingAddress: fulfillmentGroup.address,
     shopId: fulfillmentGroup.shopId,
-    // For now, this is always shipping. Revisit when adding download, pickup, etc. types
-    type: "shipping"
+    // For now, this is always shipping. Revisit when adding download, pickup, etc. types - done.
```
brent-hoover: Please remove this no longer relevant comment

packages/api-plugin-fulfillment-method-pickup-store/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Summary
+
+This plugin Implements Store pickup as a fulfillment method under the type pickup.
+
+## Included in this example plugin
+ß
+### `src/`
+
+The `src` folder contains all the plugin files. An `index.js` with a bear-bones `registerPlugin` is included.
```
brent-hoover: This is the template from the example plugin, please update with relevant information about this plugin

packages/api-plugin-fulfillment-method-pickup-store/src/util/validateOrderMethodsstore.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * @summary Sample dummy function validating the data requirements for this metod
+ * @param {Object} context - Context object
+ * @param {Object} commonOrder - Current order which provide available details to perform validation
+ * @param {Object[]} validationResults - Validation results collected till now
+ * @returns {Object[]} validationResults - with the validation details populated
+ */
+export default function validateOrderMethodsstore(context, commonOrder, validationResults = []) {
```
brent-hoover, line 8: function violates naming convention, must be camel-case

packages/api-plugin-fulfillment-method-shipping-flat-rate/README.md
```diff
@@ -0,0 +1,70 @@
...
+[![CircleCI](https://circleci.com/gh/reactioncommerce/api-plugin-fulfillment-method-shipping-flat-rate.svg?style=svg)](https://circleci.com/gh/reactioncommerce/api-plugin-fulfillment-method-shipping-flat-rate)
+[![semantic-release](https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg)](https://github.com/semantic-release/semantic-release)
+
+## Summary
+
+This plugin provides the functionalities of a sample fulfillment method flat-rate under the fulfillment type of shipping.
+
+## Included in this example plugin
+
+### `src/`
```
brent-hoover, line 13: Again this is the example plugin templates, please replace with relevant data for this plugin

packages/api-plugin-fulfillment-method-shipping-ups/README.md
```diff
@@ -0,0 +1,70 @@
...
+## Summary
+
+This plugin Implements Shipping-UPS as a fulfillment method under the type shipping.
+
+## Included in this example plugin
+
+### `src/`
+
+The `src` folder contains all the plugin files. An `index.js` with a bear-bones `registerPlugin` is included.
+
```
brent-hoover, line 16: Replace boilerplate with actual documentation

packages/api-plugin-fulfillment-method-shipping-ups/README.md
```diff
@@ -0,0 +1,70 @@
+# api-plugin-fulfillment-method-shipping-ups
```
brent-hoover, line 1: I am uncomfortable with calling this `shipping-ups` since it doesn't actually implement UPS. I would rather it be called something like "dynamic-rate"

packages/api-plugin-fulfillment-method-shipping-ups/src/getFulfillmentMethodsWithQuotesShippingUPS.js
```diff
@@ -0,0 +1,77 @@
...
+  const currentMethodInfo = { packageName };
+
+  if (retrialTargets.length > 0) {
+    const isNotAmongFailedRequests = retrialTargets.every((target) => target.packageName !== packageName);
+    if (isNotAmongFailedRequests) {
+      return previousQueryResults;
+    }
+  }
+
+  // TODO This has to be a flag in Fulfillment collection
```
brent-hoover: this much be copy pasta? Is this relevant to this code at all?

packages/api-plugin-fulfillment-method-shipping-ups/src/util/sampleTest.test.js
```diff
@@ -0,0 +1,3 @@
+test("sample test", async () => {
```
brent-hoover, line 1: Remove this

packages/api-plugin-fulfillment/src/preStartup.js
```diff
@@ -0,0 +1,21 @@
...
+ * @summary Called before startup to extend schemas
+ * @param {Object} context Startup context
+ * @returns {undefined}
+ */
+export default async function fulfillmentPreStartup(context) {
+  const allFulfillmentTypesArray = context.allRegisteredFulfillmentTypes?.registeredFulfillmentTypes;
+
+  if (!allFulfillmentTypesArray || allFulfillmentTypesArray.length === 0) {
+    Logger.warn(logCtx, "No fulfillment types available");
+    throw new ReactionError("invalid", "No fulfillment types available");
```
brent-hoover: Is "invalid" one of the error types? I think it's `invalid-parameter`

brent-hoover (changes requested):
Mostly changes around style. Will test more tomorrow

packages/api-plugin-fulfillment/src/preStartup.js
sujithvn: Changed to `invalid-parameter` and other relevant error types

packages/api-plugin-fulfillment-method-shipping-ups/src/getFulfillmentMethodsWithQuotesShippingUPS.js
sujithvn: Removed the commented code. Added validation for shipping-UPS which is based on the entry `"provider.enabled": true` in Fulfillment collection instead of checking appSettings.

sujithvn: Updated the other two ff-method functions also
```
getFulfillmentMethodsWithQuotesShippingFlatRate
getFulfillmentMethodsWithQuotesPickupStore
```

packages/api-plugin-carts/src/mutations/createCart.js
sujithvn: Done, moved this to /src/cartVersion.js with value 2

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
sujithvn: Updated the comments in code

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
sujithvn: Linter throws error
`Unexpected await inside a loop. eslint[no-await-in-loop](https://eslint.org/docs/rules/no-await-in-loop)`

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
sujithvn: Linter throws error
`Unexpected await inside a loop. eslint[no-await-in-loop](https://eslint.org/docs/rules/no-await-in-loop)`

packages/api-plugin-carts/src/xforms/xformCartCheckout.js
sujithvn: done

packages/api-plugin-fulfillment-method-shipping-ups/README.md
sujithvn: Done

packages/api-plugin-fulfillment-method-shipping-flat-rate/README.md
sujithvn: Done

packages/api-plugin-fulfillment-method-pickup-store/README.md
sujithvn: Done

apps/reaction/tests/integration/api/mutations/checkout/checkoutTestsCommon.js
sujithvn: Removed the commented code

packages/api-plugin-carts/src/mutations/transformAndValidateCart.js
sujithvn: New code within `updateCartFulfillmentGroups` makes async calls. Hence changed this function into async and included await in the calling function.

packages/api-plugin-carts/src/simpleSchemas.js
sujithvn: I moved the schemaExtend code from preStartup into the simpleSchema.js file and this function now gets called from preStartup. Advantage is that the function can now be called from the factory.js used in integration testing and now we need not require the above commented code. Also, unit-test will fail with the above code. 
We can discuss further over a call.

packages/api-plugin-fulfillment-method-shipping-ups/README.md
sujithvn: As discussed, if we are planning to expand this to a more functional (and less dummy) plugin, then we could try out ups as the first option and retain the name.
If that is not the plan, we can change the name as you have suggested.

packages/api-plugin-orders/src/xforms/xformOrderFulfillmentGroupSelectedOption.js
```diff
@@ -1,18 +1,24 @@
...
+ * @param {Object} node The group
  * @returns {Object} Transformed fulfillment option
  */
-export default function xformOrderFulfillmentGroupSelectedOption(fulfillmentOption) {
+export default function xformOrderFulfillmentGroupSelectedOption(fulfillmentOption, node) {
+  let fulfillmentType = "";
+  if (node && node.type) {
+    fulfillmentType = node.type;
+  }
+
```
vanpho93, line 12:
```suggestion
  const fulfillmentType = (node && node.type) ? node.type : "";
```

packages/api-plugin-fulfillment-type-shipping/.gitignore
```diff
@@ -0,0 +1,61 @@
+# Logs
+logs
```
vanpho93, line 2: Should remove all new .gitignore files.

packages/api-plugin-products/src/simpleSchemas.js
```diff
@@ -385,7 +385,10 @@ export const Product = new SimpleSchema({
   "supportedFulfillmentTypes": {
     type: Array
   },
-  "supportedFulfillmentTypes.$": String,
+  "supportedFulfillmentTypes.$": {
+    type: String
+    // allowedValues: ["undecided"] // This is extended in fulfillment plugin with dynamic values for allowedValues
```
tedraykov, line 390: I assume this is leftover after you moved the schema extension to the fulfillment pre-startup and it can be removed.

packages/api-plugin-products/src/mutations/createProduct.js
```diff
@@ -59,7 +59,7 @@ export default async function createProduct(context, input) {
     isVisible: false,
     shopId,
     shouldAppearInSitemap: true,
-    supportedFulfillmentTypes: ["shipping"],
+    supportedFulfillmentTypes: initialProductData.supportedFulfillmentTypes || [],
```
tedraykov, line 62: The spread operator on line 69 will override this value anyway so we either don't need it at all or we want it after the spread operator.

packages/api-plugin-catalogs/src/simpleSchemas.js
```diff
@@ -468,10 +468,13 @@ export const CatalogProduct = new SimpleSchema({
   },
   "supportedFulfillmentTypes": {
     type: Array,
-    label: "Supported fulfillment types",
-    defaultValue: ["shipping"]
+    label: "Supported fulfillment types"
+    // defaultValue: ["shipping"]
```
tedraykov, line 472: Can we remove these comments as well?

packages/api-plugin-catalogs/src/simpleSchemas.js
```diff
@@ -468,10 +468,13 @@ export const CatalogProduct = new SimpleSchema({
...
+  },
+  "supportedFulfillmentTypes.$": {
+    type: String
+    // allowedValues: ["undecided"] // This is extended in fulfillment plugin with dynamic values for allowedValues
```
tedraykov, line 476: Same as above ^

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,118 @@
...
+    // This is a new optional field that UI can pass in case the user selects fulfillment type
+    // for each item in the product details page instead of waiting till checkout
+    let { selectedFulfillmentType } = item;
 
     if (!supportedFulfillmentTypes || supportedFulfillmentTypes.length === 0) {
-      supportedFulfillmentTypes = ["shipping"];
+      if (!cart.cartVersion) {
+        supportedFulfillmentTypes = ["shipping"]; // we use 'shipping' as default fulfillment type for old v1 carts (prior to fulfillment types)
+      } else {
+        throw new ReactionError("not-found", "Product does not have any supported FulfillmentTypes");
```
tedraykov, line 95: Setting the supportedFulfillmentType on a product and on a catalog is not mandatory. This causes the `supportedFulfillmentTypes` to be an empty array by default ([packages/api-plugin-products/src/mutations/createProduct.js](https://github.com/reactioncommerce/reaction/pull/6480/files#diff-e962279b1a681b2c929f0a1ffdf4e54b0184659a794704843c657d6c95ea35c9))

We always will hit this error when adding to cart products that didn't had explicit `supportedFulfillmentTypes` set. We either have to make the `CreateProductInput.supportedFulfillmentTypes` required, or provide a default list of types. Otherwise, the system can get into an invalid state.

The invariant of the product catalog is that it should always have at least one supported ff type. If it doesn't, it cannot be fulfilled by any method and therefore it shouldn't be visible in the first place.

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,118 @@
...
 
+/**
+ * @summary Check if the provided fulfillment type is present in any of the groups and adds if not
+ * @param {Object[]} currentGroups The current cart fulfillment groups array
+ * @param {String} fulfillmentType Specific fulfillment type to be checked
+ * @param {Object} item The item (product) object
+ * @returns {undefined}
+ */
+function checkAndAddToGroup(currentGroups, fulfillmentType, item) {
+  let removeFromGroups = determineRemoveFromGroups(currentGroups, item._id, item.shopId);
```
tedraykov, line 44: This doesn't have any side effects. The `removeFromGroups` is not used anywhere and the cart object is not mutated.

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,118 @@
...
+}
+
 /**
  * @summary Updates the `shipping` property on a `cart`
  * @param {Object} context App context
  * @param {Object} cart The cart, to be mutated
  * @returns {undefined}
  */
-export default function updateCartFulfillmentGroups(context, cart) {
+export default async function updateCartFulfillmentGroups(context, cart) {
```
tedraykov, line 76: It's quite odd to me to have this transformation in the carts plugin. To me it's more suitable for the fulfillment plugin to register a `cartTransform` and mutate the cart shipping there. We are not building a microservices system, but I think we should avoid microservices anti-patterns nonetheless.  Ideally, no plugin should call a collection defined in another plugin because this increases the coupling between plugins. In the case of this function, we call the fulfillment collection which indicates this logic is more suitable for the fulfillment plugin. This is opinion-based comment :)

packages/api-plugin-carts/src/simpleSchemas.js
```diff
@@ -733,6 +746,19 @@ export const CartItem = new SimpleSchema({
...
+    allowedValues: ["undecided"], // extended with dynamic values in fulfillment plugin startup
+    optional: true
+  },
+  "supportedFulfillmentTypes": {
+    type: Array,
+    optional: true
+  },
+  "supportedFulfillmentTypes.$": {
+    type: String,
+    allowedValues: ["undecided"] // extended with dynamic values in fulfillment plugin startup
```
tedraykov, line 760: I don't understand the `"undecided"` fulfillment type and how it differs from having the ff type just be `null` but I bet you had a lot of discussions about it.

Also another thing. This might be a dumb question but if a product doesn't include an `undecided` as a supported ff type, would that break any workflows in the system?

packages/api-plugin-carts/src/resolvers/Mutation/setFulfillmentTypeForItems.js
```diff
@@ -0,0 +1,36 @@
...
+ * @param {Object} args.input - an object of all mutation arguments that were sent by the client
+ * @param {String} args.input.cartId - The opaque ID of the cart to add the items to.
+ * @param {String} args.input.cartToken - The anonymous access cartToken that was returned from `createCart`.
+ * @param {String} args.input.fulfillmentType - The fulfillment type to be set fo all items.
+ * @param {String} [args.input.items] - An array of cart items to be assigned to the fulfillment type.
+ * @param {String} args.input.clientMutationId - An optional string identifying the mutation call
+ * @param {Object} context - an object containing the per-request state
+ * @returns {Promise<Object>} setFulfillmentTypeForItemsPayload
+ */
+export default async function setFulfillmentTypeForItems(parentResult, { input }, context) {
```
tedraykov, line 18: It's good to have core business logic covered by unit tests.

This mutation might belong to the fulfillment plugin instead but the domain boundaries here are not so well defined and it's hard to determine. This is an opinion-based comment I felt like sharing.

packages/api-plugin-fulfillment-method-pickup-store/README.md
```diff
@@ -0,0 +1,70 @@
+# api-plugin-fulfillment-method-pickup-store
+
+[![npm (scoped)](https://img.shields.io/npm/v/@reactioncommerce/api-plugin-fulfillment-method-pickup-store.svg)](https://www.npmjs.com/package/@reactioncommerce/api-plugin-fulfillment-method-pickup-store)
+[![CircleCI](https://circleci.com/gh/reactioncommerce/api-plugin-fulfillment-method-pickup-store.svg?style=svg)](https://circleci.com/gh/reactioncommerce/api-plugin-fulfillment-method-pickup-store)
+[![semantic-release](https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg)](https://github.com/semantic-release/semantic-release)
```
tedraykov, line 5: We don't use semantic release anymore so we can exclude this.

packages/api-plugin-fulfillment/src/schemas/schema.graphql
```diff
@@ -0,0 +1,516 @@
...
+  name: String!
+
+  "The common-name for this method, to group all variants of the same method"
+  fulfillmentMethod: String
+
+  "A fixed price to charge for fulfillment costs when this fulfillment method is selected for an order"
+  rate: Float!
+
+  # "The shop to which this fulfillment method belongs"
+  # shop: Shop!
```
tedraykov, line 81: Don't leave commented-out code.

packages/api-plugin-fulfillment/src/schemas/schema.graphql
```diff
@@ -0,0 +1,516 @@
...
+  "The updated Group"
+  group: updateFulfillmentTypeGroup!
+
+  "The same string you sent with the mutation params, for matching mutation calls with their responses"
+  clientMutationId: String
+}
+
+
+
+# "Provider info input"
```
tedraykov, line 212: Also here ^

packages/api-plugin-fulfillment/src/schemas/schema.graphql
```diff
@@ -0,0 +1,516 @@
...
+  "Rate"
+  rate: Int!
+
+  "Flag defining enabled/disabled status"
+  enabled: Boolean!
+
+  "FulfillmentMethod"
+  fulfillmentMethod: String
+}
+# "The updated group infofrom updateFulfillmentType"
```
tedraykov, line 255: As well as here ^

tedraykov (changes requested):
I have reviewed the PR partially. It's quite a daunting task to review within days something that's been developed for months. Eventually will get to it :)

apps/reaction/tests/integration/api/mutations/shippingRates/createFlatRateFulfillmentMethod.test.js
sujithvn: Included this in the data migration PR #6472

apps/reaction/plugins.json
```diff
@@ -14,26 +14,30 @@
   "accounts": "@reactioncommerce/api-plugin-accounts",
   "authentication": "@reactioncommerce/api-plugin-authentication",
   "authorization": "@reactioncommerce/api-plugin-authorization-simple",
-  "products": "@reactioncommerce/api-plugin-products",
-  "catalogs": "@reactioncommerce/api-plugin-catalogs",
+  "products": "../../packages/api-plugin-products/index.js",
```
vanpho93, line 17: should use `@reactioncommerce/...` for all plugins including new ones.

packages/api-plugin-fulfillment-method-pickup-store/src/index.js
```diff
@@ -0,0 +1,36 @@
...
+    },
+    simpleSchemas: {
+      MethodStoreData
+    },
+    functionsByType: {
+      preStartup: [fulfillmentMethodPickupStorePreStartup],
+      startup: [fulfillmentMethodPickupStoreStartup],
+      validateOrderMethods: [validateOrderMethodsstore],
+      getFulfillmentMethodsWithQuotes: [getFulfillmentMethodsWithQuotesPickupStore],
+      getFulfillmentMethodsWithQuotespickup: [getFulfillmentMethodsWithQuotesPickupStore]
```
vanpho93, line 33: Just my opinion on how to passing function with key in register function.
This may look better

```suggestion
      getFulfillmentMethodsWithQuotes: {
        key: "pickup",
        handler: getFulfillmentMethodsWithQuotesPickupStore,
      }
```

or 

```
```suggestion
      getFulfillmentMethodsWithQuotes: [{
        key: "pickup",
        handler: getFulfillmentMethodsWithQuotesPickupStore,
      }]
```
This an example on promotion: https://github.com/reactioncommerce/reaction/blob/4820b35b6d6c0f9f2951f1831d0d831b4ba590f9/packages/api-plugin-promotions-coupons/src/index.js#L20

packages/api-plugin-fulfillment-type-pickup/src/startup.js
```diff
@@ -0,0 +1,35 @@
...
+  context.appEvents.on("afterShopCreate", async (payload) => {
+    const { shop } = payload;
+    const shopId = shop._id;
+
+    // We do not have validatePermissions in context during this startup stage, hence commenting below
+    // await context.validatePermissions("reaction:legacy:fulfillmentTypes", "read", { shopId });
+
+    const pickupRecord = await Fulfillment.findOne({ fulfillmentType: "pickup", shopId });
+    if (!pickupRecord) {
+      const groupInfo = {
```
vanpho93, line 22: I saw this code duplicated on other fulfillment types. could this be done in a declarative way via register function? I saw the same pattern for fulfillment methods too.
cc: @zenweasel

packages/api-plugin-fulfillment-method-pickup-store/src/startup.js
```diff
@@ -0,0 +1,39 @@
...
+    const { shop } = payload;
+    const shopId = shop._id;
+    const method = {};
+
+    // We do not have validatePermissions in context during this startup stage, hence commenting below
+    // await context.validatePermissions("reaction:legacy:fulfillmentTypes", "read", { shopId });
+
+    const pickupRecord = await Fulfillment.findOne({ fulfillmentType: "pickup", shopId });
+    if (!pickupRecord) throw new ReactionError("server-error", "Unable to create fulfillment method Pickup-Store without defined type");
+
```
vanpho93, line 23: can we use

```
const method = {
  name: "store",
  label: "Pickup from Store",
  ...
}
```

instead of `method.name = "store";`

packages/api-plugin-fulfillment-method-pickup-store/src/getFulfillmentMethodsWithQuotesPickupStore.js
```diff
@@ -0,0 +1,78 @@
...
+      const updatedMethod = collectStoreDetails(method, commonOrder);
+
+      rates.push({
+        carrier,
+        handlingPrice: updatedMethod.handling,
+        method: updatedMethod,
+        rate: updatedMethod.rate,
+        shippingPrice: updatedMethod.rate + updatedMethod.handling,
+        shopId: doc.shopId
+      });
```
vanpho93, line 60: this can be done in a declarative way too?

each method will provide a function that returns new rates.
the root plugin will collect all of rates to return so we won't have `previousQueryResults`

packages/api-plugin-fulfillment/src/schemas/schema.graphql
sujithvn: Removed in new PR

packages/api-plugin-fulfillment/src/schemas/schema.graphql
sujithvn: Removed in new PR

packages/api-plugin-fulfillment/src/schemas/schema.graphql
sujithvn: Removed in new PR

packages/api-plugin-fulfillment-type-pickup/src/sampleTest.test.js
```diff
@@ -0,0 +1,3 @@
+test("sample test", async () => {
```
sujithvn, line 1: To be removed

packages/api-plugin-fulfillment-method-pickup-store/README.md
sujithvn: Done

packages/api-plugin-fulfillment-method-pickup-store/src/getFulfillmentMethodsWithQuotesPickupStore.js
sujithvn: We are collecting rates from multiple methods under same ff-type. We also give each method more than one chance to return the rates (since they could be fetched from external systems). So if initial fetch attempt fails, they are added into retrialTargets

packages/api-plugin-fulfillment-method-pickup-store/src/startup.js
sujithvn: Done

packages/api-plugin-fulfillment-method-pickup-store/src/util/validateOrderMethodsstore.js
sujithvn: Fixed by converting the ff-type to Title case before concatenating

packages/api-plugin-fulfillment-method-pickup-store/src/startup.js
sujithvn: Done in PR [01 02 fulfillment type pickup](https://github.com/reactioncommerce/reaction/pull/6577/files#top)
#6577

packages/api-plugin-products/src/mutations/createProduct.js
sujithvn: Correct.

I am handling this separately in a separate PR
https://github.com/reactioncommerce/reaction/pull/6543/files#diff-e962279b1a681b2c929f0a1ffdf4e54b0184659a794704843c657d6c95ea35c9

So I have retained this change as it is so that I could avoid a merge conflict when I merge the above PR

packages/api-plugin-products/src/simpleSchemas.js
sujithvn: Yes, removed.

packages/api-plugin-catalogs/src/simpleSchemas.js
sujithvn: Done

packages/api-plugin-catalogs/src/simpleSchemas.js
sujithvn: done

packages/api-plugin-carts/src/resolvers/Mutation/setFulfillmentTypeForItems.js
sujithvn: Yes, adding test cases.

As you mentioned this could go in FFT plugin. But considering that this is mutation is specific for the cart alone, I thought of adding it here.

packages/api-plugin-carts/src/simpleSchemas.js
sujithvn: 'undecided' is used for default fulfillment group. When items are added to the cart and no specific ff-group/type is selected yet, we add those entries to 'undecided' group. Later during checkout or earlier, user can use the setFulfillmentTypeForItem endpoint to set the ff-group from undecided to any valid entries.

Products are not expected to have 'undecided' as a supported ff-type. In current scenario, all products have shipping as a ff-type. Undecided is mostly used in handling the cart items which are not decided on the ff-type.

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
sujithvn: We use `removeFromGroups` inside the `checkAndAddToGroup` func. If the item is present in more that one Fulfillment group, we remove it from all(for backward compatibility).

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
sujithvn: Thanks for your comment and it is reasonable.

As you mentioned elsewhere, there is an overlapping of boundaries here. We are very much in Cart but the cart includes fulfillment groups which are based on the fulfillment plugin. The transformations are happening only in the cart and we are referring to the Fulfillment plugin only to verify the current status (active or not) of the fulfillment type selected by user. Example, considering an old cart which had a valid ff-type earlier and the user is updating the cart on a future date when that ff-type has gone inactive.

sujithvn: @zenweasel CC

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
sujithvn: We have a separate PR (https://github.com/reactioncommerce/reaction/pull/6543/) which helps the admin to set base ff-types and these ff-types would get added to all new products along with (any) user provided ff-types.

packages/api-plugin-orders/src/xforms/xformOrderFulfillmentGroupSelectedOption.js
sujithvn: Thanks, implemented via separate PR

packages/api-plugin-fulfillment-method-shipping-ups/src/util/sampleTest.test.js
sujithvn: removed and replaced with valid test cases

packages/api-plugin-fulfillment-type-pickup/src/sampleTest.test.js
sujithvn: removed

packages/api-plugin-fulfillment-method-pickup-store/src/index.js
sujithvn: Thanks @vanpho93  ... we were thinking on how we could fix it. Your suggestion is a very elegant and made it easy to implement

packages/api-plugin-fulfillment-type-shipping/.gitignore
sujithvn: Yes, will be removing them in the split PRs (#6570, #6579, #6577)

packages/api-plugin-fulfillment-method-pickup-store/src/index.js
sujithvn: will be implementing them in the split PRs (https://github.com/reactioncommerce/reaction/pull/6570, https://github.com/reactioncommerce/reaction/pull/6579, https://github.com/reactioncommerce/reaction/pull/6577)

sujithvn:
**This PR is replaced by [PR#6480](https://github.com/reactioncommerce/reaction/pull/6840)**

PR #6538, opened 2022-09-28
feat: promotions framework v2

### LINKED ISSUE #6442: As an admin I want to be able to award a promotion based on attributes of the cart
### Possible criteria

- Total merchandise value of the cart
- Total merchandise value filtered by an attribute of a product:
    - Brand
    - Class
    - Tag

vanpho93:
Resolves https://github.com/reactioncommerce/reaction/issues/6442 https://github.com/reactioncommerce/reaction/issues/6460 https://github.com/reactioncommerce/reaction/issues/6461 https://github.com/reactioncommerce/reaction/issues/6464
Impact: major
Type: feature

## Issue

The PR stakes out the framework for promotions, and for offer triggers which allow you you create rules

## Solution

This PR uses event listeners to listen for cart update events, evaluates the cart against the active promotions and triggers the emits an event which listeners to handle the actions can listen for.

## Breaking changes

This PR has no breaking changes yet as it is currently all new code.


packages/api-plugin-promotions/src/startup.js
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  await context.mutations.saveCart(context, enhancedCart, "promotions");
+}
+
+/**
+ * @summary Perform various scaffolding tasks on startup
+ * @param {Object} context - The application context
+ * @returns {Promise<void>} undefined
```
brent-hoover, line 6: I don't almost any of this code should sit in a startup file, but in separate files. I would never expect this stuff to be here

packages/api-plugin-promotions-offers/src/startup.js
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * @summary handle cart events
+ * @param {Object} context - The per request application context
+ * @returns {void}
+ */
+export default function startupOffers(context) {
+  const { promotionContext } = context;
+
+  promotionContext.registerEnhancer(enhancers);
+  promotionContext.registerTrigger("offers", handlers.offerTriggerHandler);
```
brent-hoover: Both actions and triggers should accept a third parameter of a schema

packages/api-plugin-promotions/src/preStartup.js
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * @summary apply all schema extensions to the Promotions schema
+ * @param {Object} context - The application context
+ * @returns {undefined} undefined
+ */
+function extendSchemas(context) {
+  const {
+    promotions: { schemaExtensions },
+    simpleSchemas: { Promotions }
+  } = context;
```
brent-hoover: Style nitpick. We do this on one line in like 1000 places in the app so would like to keep it consistent.

brent-hoover (changes requested):
A couple of style notes

packages/api-plugin-promotions/src/preStartup.js
brent-hoover: This is not actually resolved. Please fix

packages/api-plugin-promotions/src/preStartup.js
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * @summary extend the cart schema
+ * @param {Object} context - The application context
+ * @returns {Object} the extended schema
+ */
+function extendCartSchema(context) {
+  const {
+    simpleSchemas: { Cart, Promotion }
+  } = context; // we get this here rather than importing it to get the extended version
```
brent-hoover: Again, let's not introduce another idiom here. Please make this one line. This makes it harder rather than easier to read.

packages/api-plugin-promotions/src/utils/canBeApplied.test.js
```diff
@@ -0,0 +1,60 @@
+import canBeApplied from "./canBeApplied.js";
+
+const promotion = {
+  _id: "test id",
+  actions: [{ actionKey: "test" }],
+  triggers: [{ triggerKey: "test", triggerParameters: { name: "test trigger" } }],
+  stackAbility: "none"
+};
+
+test("should return true when the cart don't have promotion are applied", () => {
```
brent-hoover: Not sure who wrote this but I can't understand the description on this test

packages/api-plugin-promotions/src/utils/canBeApplied.test.js
```diff
@@ -0,0 +1,60 @@
...
+  };
+  const secondPromotion = {
+    ...promotion,
+    _id: "promotion 2",
+    stackAbility: "none"
+  };
+  expect(canBeApplied(cart.appliedPromotions, secondPromotion)).toBe(false);
+});
+
+test("should return true when the promotions have stack ability", () => {
```
brent-hoover: Test description should read something like "when stackability is set to all"

packages/api-plugin-sample-data/src/loaders/loadImages.js
```diff
@@ -93,25 +90,30 @@ async function storeFromAttachedBuffer(fileRecord) {
...
 /**
  * @summary loads Images for the products
  * @param {Object} context - The application context
  * @param {String} shopId - The shop id
  * @returns {Promise<boolean>} true if success
  */
 export default async function loadImages(context, shopId) {
-  const { collections: { Media } } = context;
-  const { mutations: { publishProducts } } = context;
+  const {
```
brent-hoover: Please fix whatever robot is forcing this convention.

brent-hoover (changes requested):
Still a couple of comments/small changes

packages/api-plugin-promotions-coupons/src/utils/isPromotionExpired.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * @summary check if promotion is expired
+ * @param {Object} promotion - The promotion to check
+ * @returns {Boolean} - Whether the promotion is expired
+ */
+export default function isPromotionExpired(promotion) {
+  const { endDate } = promotion;
+  const now = Date.now();
+  return endDate && endDate < now;
```
sujithvn, line 9: I believe it is expected that there could be promotions which does NOT have an expiry and those promotions will not have an endDate field.

packages/api-plugin-promotions-coupons/src/triggers/couponsTriggerHandler.js
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * @summary Trigger handler for coupon
+ * @param {Object} context - The application context
+ * @param {Object} enhancedCart - The cart to apply promotions to
+ * @param {Object} trigger - The parameters to pass to the trigger
+ * @returns {Boolean} - Whether the promotion can be applied to the cart
+ */
+export async function couponTriggerHandler(context, enhancedCart, { triggerParameters }) {
+  return triggerParameters.couponCode === "code";
```
sujithvn: Just for clarity - 
Is this a temp/dummy function that compares the couponCode to the hard-coded value of "code". Or is this expected to be always compared against 'code'?

packages/api-plugin-promotions-coupons/src/simpleSchemas.js
```diff
@@ -0,0 +1,8 @@
+import SimpleSchema from "simpl-schema";
+
+export const CouponTriggerParameters = new SimpleSchema({
+  name: String,
+  couponCode: {
+    type: String
```
sujithvn, line 6: `couponCode: String ` 
Not sure why specifically used a longhand definition

packages/api-plugin-promotions-coupons/src/schemas/schema.graphql
```diff
@@ -0,0 +1,12 @@
+input ApplyCouponToCartInput {
+  cartId: String!
+  couponCode: String!
+}
+
+type ApplyCouponToCartPayload {
```
sujithvn, line 17: Do we need to add the description for each input/type to show up in GraphQL documentation?

packages/api-plugin-promotions-offers/src/triggers/offerTriggerHandler.test.js
```diff
@@ -0,0 +1,43 @@
+import mockContext from "@reactioncommerce/api-utils/tests/mockContext.js";
+import merchandiseTotal from "../enhancers/merchandiseTotal.js";
+import { offerTriggerHandler } from "./offerTriggerHandler.js";
+
+const pluginPromotion = {
+  operators: {}
+};
+
+const triggerParameters = {
+  name: "50% off your entire order when you spend more then $200",
```
sujithvn, line 10: minor typo
then -> than

packages/api-plugin-promotions/src/utils/isPromotionExpired.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * @summary check if promotion is expired
+ * @param {Object} promotion - The promotion to check
+ * @returns {Boolean} - Whether the promotion is expired
+ */
+export default function isPromotionExpired(promotion) {
```
sujithvn, line 6: We have a duplicate function in 
api-plugin-promotions-coupons/src/utils/isPromotionExpired.js

packages/api-plugin-promotions/src/simpleSchemas.js
```diff
@@ -0,0 +1,77 @@
...
+  actionParameters: {
+    type: Object,
+    blackbox: true
+  }
+});
+
+export const Trigger = new SimpleSchema({
+  triggerKey: {
+    type: String,
+    allowedValues: []
```
sujithvn, line 17: Sorry, I am not clear why we have an empty array for allowedValues

packages/api-plugin-promotions/src/qualifiers/stackable.js
```diff
@@ -0,0 +1,27 @@
...
+ * @summary does promotion meet stackability requirements
+ * @param {Object} context - The application context
+ * @param {Array<Object>} appliedPromotions - The promotions already applied
+ * @param {Object} promotion - The promotions we are trying to apply
+ * @return {{reason: string, qualifies: boolean}} - If it qualifies and if it doesn't why not
+ */
+export default function stackable(context, appliedPromotions, promotion) {
+  if (appliedPromotions[0].stackAbility === "none" || promotion.stackAbility === "none") {
+    Logger.info(logCtx, "Cart disqualified from promotion because stack ability is none");
+    return { qualifies: false, reason: "Cart disqualified from promotion because stack ability is none" };
```
sujithvn, line 24: Just my personal view point-

Is the current "reason" message really required/adding value, since it is evident that false means promotions are not stackable.

"reason" would make sense if we are passing explicit/separate messages like:

if (appliedPromotions[0].stackAbility === "none") return { qualifies: false, reason: "A non stackable promotion is already applied to the cart and new promotions cannot be applied" };

if (promotion.stackAbility === "none") return { qualifies: false, reason: "The applied promotion does not support stackability" };

sujithvn (dismissed):
Just few clarifications and minor suggestions. I did not see any issues.

packages/api-plugin-promotions/src/simpleSchemas.js
brent-hoover: Because we concatenate them later

packages/api-plugin-promotions/src/qualifiers/stackable.js
brent-hoover: No, incorrect. This might apply to stackability but there are other systems that may apply qualifiers that need to supply a reason that need to show up in the cart.

brent-hoover (dismissed):

packages/api-plugin-promotions/src/qualifiers/stackable.js
brent-hoover: Also, even stackability is going to have several types which will have separate messages

packages/api-plugin-promotions-coupons/src/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @summary Import and call this function to add this plugin to your API.
+ * @param {Object} app The ReactionAPI instance
+ * @returns {undefined}
+ */
+export default async function register(app) {
+  await app.registerPlugin({
+    label: pkg.label,
+    name: pkg.name,
```
tedraykov, line 18: I would suggest following the naming style of other plugins. If we fetch the plugin name from the `package.json` we would have the `@reactioncommerce` org prefix in the logs, which differs from other plugins.
Screenshot 2022-10-18 at 15 58 29

packages/api-plugin-promotions-coupons/src/utils/isPromotionExpired.js
vanpho93: yes, that's correct.

packages/api-plugin-promotions/src/utils/isPromotionExpired.js
vanpho93: I think it's acceptable since it's a very small function and just 2 plugins use it.

packages/api-plugin-promotions-coupons/src/simpleSchemas.js
vanpho93: I think using `couponCode` is clearer than `code`. `code` can be anything. It's debatable.

packages/api-plugin-promotions-coupons/src/schemas/schema.graphql
vanpho93: Added

tedraykov (commented):
I'm sorry, but I don't seem to get the whole picture. There are no mutations to create any type of promotion and I can't go through a scenario of applying a discount. I tried to make sense of the framework but without going through an e2e use case I can't give a confident review. Is there technical documentation on how the promotions are designed to work? Namely,  I don't get what is the difference between an offer and a discount and what exactly actions and triggers are doing.

packages/api-plugin-promotions-coupons/src/index.js
zenweasel: @tedraykov this is precisely why I asked you to go through it with Brian. I suggest you do that you do that as soon as possible.

packages/api-plugin-promotions-coupons/src/schemas/schema.graphql
```diff
@@ -0,0 +1,21 @@
+"Input for the applyCouponToCart mutation"
+input ApplyCouponToCartInput {
+  "The ID of the Cart"
+  cartId: String!
```
tedraykov: Whenever we have an operation over a cart, we have to cover the anonymous cart use case. That is why we need to pass the anonymous cart token in the input. 
The existing discount codes plugin has it in the input:
https://github.com/reactioncommerce/reaction/blob/871d46c2e070d5d80621dc1cdfaae2398e42d99f/packages/api-plugin-discounts-codes/src/schemas/schema.graphql#L137

https://github.com/reactioncommerce/reaction/blob/trunk/packages/api-plugin-discounts-codes/src/util/getCart.js
Identical check need to be done in this plugin.

packages/api-plugin-promotions/src/handlers/applyPromotions.js
```diff
@@ -0,0 +1,92 @@
...
+const logCtx = {
+  name,
+  version,
+  file: "applyImplicitPromotions.js"
+};
+
+/**
+ * @summary get all implicit promotions
+ * @param {Object} context - The application context
+ * @returns {Array<Object>} - An array of promotions
```
tedraykov:
```suggestion
 * @returns {Promise<Array<Object>>} - An array of promotions
```

packages/api-plugin-promotions/src/handlers/applyPromotions.js
```diff
@@ -0,0 +1,92 @@
...
+  Logger.info({ ...logCtx, applicablePromotions: promotions.length }, "Fetched applicable promotions");
+  return promotions;
+}
+
+/**
+ * @summary apply promotions to a cart
+ * @param {Object} context - The application context
+ * @param {Object} cart - The cart to apply promotions to
+ * @param {Object} explicitPromotion - The explicit promotion to apply
+ * @returns {Object} - The cart with promotions applied
```
tedraykov:
```suggestion
 * @returns {Promise<Object>} - The cart with promotions applied
```

packages/api-plugin-promotions/src/handlers/applyPromotions.test.js
```diff
@@ -0,0 +1,69 @@
+import mockContext from "@reactioncommerce/api-utils/tests/mockContext.js";
+import applyImplicitPromotions from "./applyPromotions.js";
+
+const testTrigger = jest.fn().mockReturnValue(Promise.resolve(true));
```
tedraykov, line 4: Whenever we mock an async function it's more concise to use `mockResolvedValues`. Also in multiple places in the unit test, we are returning sync value although the method is supposed to return a Promise. It's good to review the tests and replace `mockReturnValue` with `mockResolvedValue` especially if the returned value is not wrapped in a `Promise.resolve(value)` because we might be skipping an `await` somewhere an a bug might not be catched by tests.

tedraykov (changes requested):
This looks like a pretty good foundation for the promotions. It would be interesting to see how the actions will be defined.

packages/api-plugin-promotions-coupons/src/schemas/schema.graphql
vanpho93: Updated, can you check again?

packages/api-plugin-promotions-coupons/src/mutations/applyCouponToCart.test.js
```diff
@@ -13,40 +18,55 @@ test("should call applyExplicitPromotionToCart mutation", async () => {
...
   mockContext.collections.Cart = {
-    findOne: jest.fn().mockReturnValueOnce(cart)
+    findOne: jest.fn().mockResolvedValueOnce(cart)
   };
   mockContext.collections.Promotions = {
-    findOne: jest.fn().mockReturnValueOnce(promotion)
+    findOne: jest.fn().mockResolvedValueOnce(promotion)
   };
-  mockContext.mutations.applyExplicitPromotionToCart = jest.fn().mockName("applyExplicitPromotionToCart").mockReturnValueOnce(Promise.resolve(cart));
+  mockContext.mutations.applyExplicitPromotionToCart = jest.fn().mockName("applyExplicitPromotionToCart").mockResolvedValueOnce(Promise.resolve(cart));
```
tedraykov: This will return a Promise of a Promise. Remove the `Promise.resolve` wrapper.

packages/api-plugin-promotions/src/index.js
```diff
@@ -0,0 +1,45 @@
...
+ * @returns {undefined}
+ */
+export default async function register(app) {
+  await app.registerPlugin({
+    label: pkg.label,
+    name: pkg.name,
+    version: pkg.version,
+    collections: {
+      Promotions: {
+        name: "Promotions"
```
tedraykov: The promotions, especially the implicit ones will be queried a lot and won't be updated often. That's why we should add indexes to some promotions' attributes. For example in the `getImplicitPromotions`, we are querying `enabled`, `type`, `startDate`, `endDate`. Also, each promotion implementation plugin should define it's own indexes, provided they are not already indexed in the root plugin.
 
The `coupons` plugin should index the `triggers.triggerKey`, `triggers.triggerParameters.couponCode` attributes.

tedraykov (changes requested):

packages/api-plugin-promotions/src/index.js
tedraykov: I'm not sure if `api-core` allows to add indexes to a collection that is already defined in another plugin.

vanpho93: Updated.

tedraykov (approved):
You don't seem to have addressed my comment here but I assume it's not relevant.
https://github.com/reactioncommerce/reaction/pull/6538#discussion_r998182826

Otherwise, it's good to go for me.

brent-hoover:
> You don't seem to have addressed my comment here but I assume it's not relevant. [#6538 (comment)](https://github.com/reactioncommerce/reaction/pull/6538#discussion_r998182826)
> 
> Otherwise, it's good to go for me.

@tedraykov I don't really see anything in that comment that is actionable since it's about your not understanding how it works? If you need more clarity on anything please work with @vanpho93 

@vanpho93 I think in the future it would be helpful to provide a little more context similar to what @sujithvn has been doing on his PR's so people who are not up to speed understand the context. Especially for someone like Ted who is doing a lot of context-switching.

tedraykov:
I was referring to the plugin names in the logging here

brent-hoover:
> I was referring to the plugin names in the logging here

Got it, my mistake.

PR #6543, opened 2022-10-04
Set default ff-type for shop to be used in CreateProduct

### LINKED ISSUE #6474: As an admin I should be able to define what available fulfillment types should be added by default to new products
You should be able to check a box in the admin on any/all/none of the fulfillment types and those would automatically be added to new products as they are created. THIS WILL NOT AFFECT EXISTING PRODUCTS AT ALL

sujithvn:

Resolves #6474 
Impact: **major**
Type: **feature**

## Issue

Admin should be able to define what available fulfillment types should be added by default to new products(#6474)
User should be able to check a box in the admin on any/all/none of the fulfillment types and those would automatically be added to new products as they are created. THIS SHOULD NOT AFFECT EXISTING PRODUCTS AT ALL

## Solution

Current available FF-types can be retrieved using the query `getFulfillmentTypes` and presented to user. User can select one/all/none of the ff-types and use the `updateShopSettings` mutation to update the same.

While creating a new Product, the `supportedFulfillmentTypes` in the input (if provided) will be added to the default ff-types from the shop settings and used for creating new product.

## Breaking changes
None. If there are no ff-types added then no entries are added to product.  User input will be retained as it is.

## Testing
1. use the `updateShopSettings` mutation to update the default ff-types for shop.
2. new products created will include the default ff-type from shop settings 
3. use the `updateShopSettings` mutation to remove any of the default ff-types for shop.
4. new products created will contain only user input ff-types

### DIFF
--- a/packages/api-plugin-fulfillment/src/index.js
+++ b/packages/api-plugin-fulfillment/src/index.js
@@ -56,17 +56,16 @@ export default async function register(app) {
       MethodEmptyData
     },
     shopSettingsConfig: {
-      defaultFulfillmentTypesForShop: {
+      baseFulfillmentTypesForShop: {
         permissionsThatCanEdit: ["reaction:legacy:fulfillmentTypes/update:settings"],
         simpleSchema: {
           type: {
-            "defaultFulfillmentTypesForShop": {
+            "baseFulfillmentTypesForShop": {
               type: Array,
               optional: true
             },
-            "defaultFulfillmentTypesForShop.$": {
-              type: String,
-              optional: true
+            "baseFulfillmentTypesForShop.$": {
+              type: String
             }
           }
         }
--- a/packages/api-plugin-fulfillment/src/index.js
+++ b/packages/api-plugin-fulfillment/src/index.js
@@ -54,6 +54,23 @@ export default async function register(app) {
     },
     simpleSchemas: {
       MethodEmptyData
+    },
+    shopSettingsConfig: {
+      defaultFulfillmentTypesForShop: {
+        permissionsThatCanEdit: ["reaction:legacy:fulfillmentTypes/update:settings"],
+        simpleSchema: {
+          type: {
+            "defaultFulfillmentTypesForShop": {
+              type: Array,
+              optional: true
+            },
+            "defaultFulfillmentTypesForShop.$": {
+              type: String,
+              optional: true
+            }
+          }
+        }
+      }
     }
   });
 }
--- a/packages/api-plugin-products/src/mutations/createProduct.js
+++ b/packages/api-plugin-products/src/mutations/createProduct.js
@@ -49,10 +49,12 @@ export default async function createProduct(context, input) {
     throw new ReactionError("invalid-param", "Creating a deleted product is not allowed");
   }
 
-  const { defaultFulfillmentTypesForShop } = await context.queries.appSettings(context, shopId);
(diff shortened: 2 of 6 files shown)


brent-hoover (approved):

brent-hoover:
@sujithvn Please correct your branches on this and other PR's

packages/api-plugin-products/src/mutations/createProduct.js
```diff
@@ -49,6 +49,13 @@ export default async function createProduct(context, input) {
     throw new ReactionError("invalid-param", "Creating a deleted product is not allowed");
   }
 
+  const { defaultFulfillmentTypesForShop } = await context.queries.appSettings(context, shopId);
+  const { supportedFulfillmentTypes } = initialProductData;
+  let combinedSupportedFulfillmentTypes = [];
+  if (defaultFulfillmentTypesForShop) combinedSupportedFulfillmentTypes = [...defaultFulfillmentTypesForShop];
+  if (supportedFulfillmentTypes) combinedSupportedFulfillmentTypes = [...combinedSupportedFulfillmentTypes, ...supportedFulfillmentTypes];
+  combinedSupportedFulfillmentTypes = [...new Set(combinedSupportedFulfillmentTypes)];
```
aldeed, line 59: This seems wrong. If a product has `supportedFulfillmentTypes` supplied when creating it, wouldn't you want to keep that as is? I would think it would just be:

```js
supportedFulfillmentTypes: initialProductData.supportedFulfillmentTypes || defaultFulfillmentTypesForShop || []
```

If the intention is to _force_ every produce to support these, then another prefix like `base` or `global` might be a better name than `default` for the setting.

packages/api-plugin-fulfillment/src/index.js
```diff
@@ -54,6 +54,23 @@ export default async function register(app) {
...
+        permissionsThatCanEdit: ["reaction:legacy:fulfillmentTypes/update:settings"],
+        simpleSchema: {
+          type: {
+            "defaultFulfillmentTypesForShop": {
+              type: Array,
+              optional: true
+            },
+            "defaultFulfillmentTypesForShop.$": {
+              type: String,
+              optional: true
```
aldeed: Optional on the item schema would mean a sparse array (with nulls allowed) so `"defaultFulfillmentTypesForShop.$": String` is probably better.

aldeed (changes requested):

packages/api-plugin-products/src/mutations/createProduct.js
sujithvn: @aldeed 
Yes, my intention was to add the user provided `supportedFulfillmentTypes` to the `base` entries defined for the Shop. But either of them could be undefined. That is the reason I did so.
As you mentioned, `base` would be a better name than `default` in the shopSettings.

packages/api-plugin-fulfillment/src/index.js
sujithvn: Ok, removing the `optional`

aldeed (approved):

vanpho93 (approved):

aldeed (approved):

sujithvn:
This PR is replaced by the commit `df5ed5e2f8ef9509490c2e90af20f9b80f439773` on  branch `00-fulfillment-base`

PR #6544, opened 2022-10-04
Datamigration for new Fulfillment type feature

### LINKED ISSUE #6472: As an admin I want to be able to migrate my data to the new format required for the new FFT plugins

sujithvn:
Resolves #6472 
Impact: **breaking**
Type: **feature**

# This PR is replaced by #6633

## Issue

Migrate data to the new format required for the new FFT plugins

## Solution

We have 3 stages in a single migration;
* Copies over the existing `Shipments` & `FlatRateFulfillmentRestrictions` collections to `Fulfillment` & `FulfillmentRestrictions` collections respectively
* Updates the `Fulfillment` collection with the newly introduced fields (`fulfillmentType` & `fulfillmentMethod`). Since the existing user base had only Shipping/Flatrate as the fulfillment type/method till date, we would be using those values to update collection
* update the `Groups` collection with the new permissions

## Breaking changes
The 2nd stage above handles the breaking change where we would be using the possible existing values as the values or newly introduced fields.

## Testing
In an existing installation, perform the data-migration. Update the installation to use the new version having fulfillment types. All activities in admin/shop (related to Products/Orders/FF-types/Catalogs/Carts) should continue to work without issues

To test migration steps, install api-migrations package and include the below entry in the `migrator.config.js` file.

```
export default {
  tracks: [
    {
      namespace: "fulfillment",
      package: "../reaction/packages/api-plugin-fulfillment",
      version: 2
    }
  ]
};

```

Run the command:
`MONGO_URL=mongodb://localhost:27017/reaction npx migrator migrate`


packages/api-plugin-fulfillment/migrations/4.js
```diff
@@ -0,0 +1,44 @@
...
+    slug: { $in: affectedGroups }
+  }, {
+    $addToSet: { permissions: { $each: newShopPermissions } }
+  });
+
+  progress(100);
+}
+
+export default {
+  down: "impossible",
```
aldeed: Seems like down migration would actually be possible with similar code except `$pullAll` rather than `$addToSet`.

packages/api-plugin-fulfillment/migrations/3.js
```diff
@@ -0,0 +1,56 @@
...
+  operations.push(ffTypeUpdate);
+  operations.push(ffMethodUpdate);
+
+  const bulkWriteResp = await db.collection(COLL_DEST).bulkWrite(operations);
+  if (bulkWriteResp.writeErrors && bulkWriteResp.writeErrors.length) throw new Error("Error while updating Fulfillment collection");
+  progress(100);
+}
+
+export default {
+  down: "impossible", // We are not tracking the updated documents, hence cannot revert
```
aldeed: It seems like migrating down might not really be necessary (old code would work either way). If so, you could use `"unnecessary"` rather than `"impossible"` here, which would then not block rollback.

aldeed (changes requested):

packages/api-plugin-fulfillment/migrations/4.js
sujithvn: Yes, I am adding the down migration

packages/api-plugin-fulfillment/migrations/3.js
sujithvn: Thanks for the suggestion, that approach is correct. I am making this change

packages/api-plugin-fulfillment/migrations/4.js
sujithvn: @aldeed @zenweasel 
**Note:**
Additionally, I removed the below entry since it would be better to include this entry in the PR #6543 
`"reaction:legacy:fulfillmentTypes/update:settings",`

Although this would require a separate migration script for only this entry, the change actually belong to that PR.

packages/api-plugin-fulfillment/migrations/4.js
```diff
@@ -0,0 +1,64 @@
...
+ * @param {Object} context Migration context
+ * @param {Object} context.db MongoDB `Db` instance
+ * @param {Function} context.progress A function to report progress, takes percent
+ *   number as argument.
+ * @return {undefined}
+ */
+async function up({ db, progress }) {
+  progress(0);
+
+  await db.collection("Groups").updateMany({
```
brent-hoover: This just adds them to shop manager and owner which is fine if they used the Out of the Box permissions and made no changes, but what it actually should do is add it to any group that has the old equivalent permissions?

pnpm-lock.yaml
```diff
@@ -245,7 +245,7 @@ importers:
       '@reactioncommerce/logger': link:../../packages/logger
       '@reactioncommerce/nodemailer': 5.0.5
       '@reactioncommerce/random': link:../../packages/random
-      '@snyk/protect': 1.1017.0
+      '@snyk/protect': 1.1020.0
```
brent-hoover, line 248: You should remove this

packages/api-plugin-fulfillment/migrations/2.js
```diff
@@ -0,0 +1,47 @@
+const COLL_FF_SOURCE = "Shipping";
+const COLL_FF_DEST = "Fulfillment";
+const COLL_FFR_SOURCE = "FlatRateFulfillmentRestrictions";
+const COLL_FFR_DEST = "FulfillmentRestrictions";
+
```
brent-hoover, line 24: What is the reasoning behind breaking this into three different migrations? This is definitely not the pattern we should follow.

brent-hoover (changes requested):
A few changes needed here

packages/api-plugin-fulfillment/migrations/2.js
sujithvn: I was keeping it as 3 different stages within the migration just for easy understanding
1. Copy over the old collection as new ones (down: removes these)
2. Update the new collections (down: unnecessary)
3. Update the other (existing) collections to add rules (down: removes these added rules)

If we have to add all these steps into a single migration, we can combine them.
@zenweasel

packages/api-plugin-fulfillment/migrations/4.js
sujithvn: @zenweasel 

When we say "_any group that has the old equivalent permissions_", does it mean any group which would have the same entries in the array as of shop-manager and owner? 

If the user has made a minor add/remove of any entry, then how do we decide whether that groups has to be considered or not?

pnpm-lock.yaml
sujithvn: Done in new PR

packages/api-plugin-fulfillment/migrations/2.js
sujithvn: As discussed, I will be combining all the migrations into one. This will be implemented in the new replacement PR

packages/api-plugin-fulfillment/migrations/3.js
sujithvn: Done

packages/api-plugin-fulfillment/migrations/4.js
sujithvn: As discussed, I would be re-introducing the removed entry back to this PR as it is preferred to have a single migration

sujithvn: Done

packages/api-plugin-fulfillment/migrations/4.js
sujithvn: @zenweasel 
As discussed, I have made the code change to update 'any' group that has the permissions. 
As per the change, we are checking each group for any of the  old permissions (related to shipping) and inserting the corresponding individual FFT permissions (using a mapper) in all of the impacted groups.

packages/api-plugin-fulfillment/migrations/2.js
sujithvn: I have implemented here also and will be copied over to the new replacement PR

sujithvn:
This PR is replaced by 6633
https://github.com/reactioncommerce/reaction/pull/6633

PR #6546, opened 2022-10-06
feat: pushing tags and releases

### LINKED ISSUE #6540: When a new version of the app is created, create a Github release

vanpho93:
Resolves #6540
Impact: minor
Type: feature

## Issue

Missing releases and tags in Github

## Solution

When there is new commit on trunk:
* Add release / tag for reaction app
* Add tags for other packages

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,8 +35,6 @@ jobs:
 
       - name: Create Release Pull Request
         uses: changesets/action@v1
-        with:
-          version: pnpm changeset version
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,9 +33,9 @@ jobs:
       - name: Install Dependencies
         run: pnpm install --ignore-scripts
 
-      - name: Create Release Pull Request or Publish to npm
+      - name: Create Release Pull Request
         uses: changesets/action@v1
         with:
-          publish: pnpm release
+          version: pnpm changeset version
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,7 +20,7 @@ jobs:
           cache-name: cache-node-modules
         with:
           key: reaction-v6-node-modules-${{ hashFiles('package.json') }}-${{ hashFiles('pnpm-lock.yaml') }}
-          path: '**/node_modules'
+          path: "**/node_modules"
 
       - name: Use Node.js 14.x
         uses: actions/setup-node@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/tagging-and-release.yml
```diff
@@ -0,0 +1,78 @@
+name: Tagging And Release
```
tedraykov, line 1: In my opinion, there is no need to manually create tags and do GitHub releases as Changesets' action is doing that out of the box.
https://github.com/changesets/action#inputs by default the `createGithubReleases` is set to true.

.github/workflows/release.yml
```diff
@@ -33,9 +33,22 @@ jobs:
       - name: Install Dependencies
         run: pnpm install --ignore-scripts
 
-      - name: Create Release Pull Request or Publish to npm
+      - name: Create Release Pull Request
         uses: changesets/action@v1
         with:
-          publish: pnpm release
+          version: pnpm changeset version
```
tedraykov: This is not doing anything as the version input is set to `changeset version` by default anyway.

.github/workflows/release.yml
```diff
@@ -33,9 +33,22 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Update pnpm-lock.yaml
```
tedraykov, line 41: I don't quite get why we need to manually push the modified package lock here. Can you elaborate on that?

From what I understand the workflow is the following:
- We create a PR with some changes and we run `changeset`.
- When we merge the PR to trunk, this workflow should be triggered and the `changesets/action` should recognise it's a new feature commit and will run the `version` script and will then create the `Version Packages` PR (without publishing yet)
- Next, we check if everything is okay with the version PR and we merge it.
- This workflow should trigger again, but this time the `changesets/action` will detect the commit is `Version Package` and will run the `publish` script instead of the `version` one. If we leave the `publish` script empty here, changesets action will internally run `npm publish` and `changeset tag`. After all packages are published, it will push the github tags and will create github releases from them.

.github/workflows/release.yml
tedraykov: Do we want to release the npm packages from the github action or from circle ci? If the latter, we can use the gh action just for releasing the github releases by using `publish: pnpm changeset tag`

.github/workflows/tagging-and-release.yml
vanpho93: We only create GitHub release for reaction app so cannot use `createGithubReleases` from changesets.

.github/workflows/release.yml
vanpho93: Since we update version of subpackages, we the pnpm-lock file will change if we bump up versions in apps/reaction/package.json.

Like this trunk inside pnpm-lock.yaml

```
  apps/reaction:
    specifiers:
      '@reactioncommerce/api-core': ^2.0.0
      '@reactioncommerce/api-plugin-accounts': ^2.0.0
      '@reactioncommerce/api-plugin-address-validation': ^1.3.1
```

.github/workflows/release.yml
vanpho93: > This is not doing anything as the version input is set to changeset version by default anyway.

Actually we have to call it manually. Can you suggest the code change here so I can try out to see if it works?

`pnpm changeset tag` only do the tagging in local, you need to push tags too.

vanpho93: Sorry @tedraykov , we can remove 

```
        with:
          version: pnpm changeset version
```

.github/workflows/release.yml
tedraykov: Okay, I agree this is needed. My only note is that changesets won't bump versions of the dependencies in the `app/reaction` unless we are doing a major release because we are using ^ for all plugins and in turn, the lock file won't change.

.github/workflows/release.yml
tedraykov: `pnpm changeset tag` only creates the tags, but if you add it to the `changesets/action`, it will also push the tags to the remote.

Here is an example:
https://github.com/tedraykov/reaction/actions/runs/3202693539/jobs/5231938200#step:7:69

tedraykov (approved):

PR #6534, opened 2022-09-26
Product/Tag Date Filtering

dcrdev:
Impact: **minor**
Type: **feature**

## Issue

This PR adds basic date filtering to products and tags via several different operands eg. after, before, between.

At the moment this is light touch, only servicing products and tags and could be improved upon by constraining these date filter types with graphql interfaces, however I wasn't sure whether that's an existing pattern, nor where they would live.

I wasn't aware of #6527 when I began this, so this wasn't specifically drafted with that in mind, but rather precipitated by our own internal requirements.

Issue #6527, opened 2022-09-26
Data Filtering/Search


brent-hoover:
Establishing a system-wide pattern for doing searching and filtering, eventually with the ability to save filters. Mostly focused on Products/Orders/Customers

chethan27:
Hi zenweasel , are you planning to implement this ?  if yes, I am happy to start working no this.

PR #6549, opened 2022-10-07
Product/Tag Date Filtering

dcrdev:
Impact: **minor**
Type: **feature**

## Issue

This PR adds basic date filtering to products and tags via several different operands eg. after, before, between.

At the moment this is light touch, only servicing products and tags and could be improved upon by constraining these date filter types with graphql interfaces, however I wasn't sure whether that's an existing pattern, nor where they would live.

I wasn't aware of #6527 when I began this, so this wasn't specifically drafted with that in mind, but rather precipitated by our own internal requirements.

Copied from: #6534

### DIFF
--- a/.changeset/chilly-geckos-deliver.md
+++ b/.changeset/chilly-geckos-deliver.md
@@ -1,0 +1,6 @@
+---
+"@reactioncommerce/api-plugin-products": minor
+"@reactioncommerce/api-plugin-tags": minor
+---
+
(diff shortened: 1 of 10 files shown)


dcrdev:
@zenweasel as requested

Still doesn't look like CI is running

vanpho93:
There's sth weird in CircleCi

image

dcrdev:
Closing in favor of https://github.com/reactioncommerce/reaction/pull/6550

PR #6552, opened 2022-10-12
feat: promotion framework with coupons

### LINKED ISSUE #6443: Promotions - Milestone 2
Coupons

Scenario:
Get $10 off when you use coupon code OCTOBER

vanpho93:
Resolves #6443
Impact: major
Type: feature

## Issue

Currently, there is no explicit promotion in the framework so we need to check if it works.

## Solution

Adding coupon as new type of promotion to see how it works with current promotion framework


packages/api-plugin-promotions-coupons/src/triggers/couponsTriggerHandler.js
```diff
@@ -0,0 +1,18 @@
+import { CouponTriggerParameters } from "../simpleSchemas.js";
+
+/**
+ * @summary a no-op function for testing of promotions
```
brent-hoover: Copy-pasta Summary

packages/api-plugin-sample-data/src/loaders/loadPromotions.js
```diff
@@ -35,7 +35,33 @@ const OrderPromotion = {
   stackAbility: "none"
 };
 
-const promotions = [OrderPromotion];
+const CouponPromotion = {
+  _id: "couponPromotion",
+  type: "explicit",
+  label: "15 percent off your entire order when you spend more then $100",
```
brent-hoover: So this description is confusing. Coupons don't usually have a separate condition and are these being enforced?

packages/api-plugin-promotions-coupons/src/mutations/applyCouponToCart.js
```diff
@@ -0,0 +1,53 @@
...
+
+  const { collections: { Cart, Promotions } } = context;
+  const { cartId, couponCode } = input;
+
+  const cart = await Cart.findOne({ _id: cartId });
+  if (!cart) {
+    throw new ReactionError("not-found", "Cart not found");
+  }
+
+  const now = new Date();
```
brent-hoover, line 31: One thing to think about here (and everywhere we are doing date/time comparisons) is that this is going to be server time which is generally going to be GMT which means that a coupon would expire 8 hours earlier than it should which is going to make people angry. We probably need to have the user set a timezone to use and use that everywhere.

brent-hoover (changes requested):
Looking really good. Just a few minor comments

packages/api-plugin-promotions-coupons/src/mutations/applyCouponToCart.js
vanpho93: I think we should have only one strict end time for promotions, but showing that time on UI should change regard user timezones. If we do the comparisons base on user timezones, they may change timezone setting to be able to use the expired promotion?

brent-hoover: By user I mean our customer, the person creating the promotion, not the consumer. Should have been more clear.

vanpho93: It seems to me this is a common front end task.
They can get the current timezone add it to date time input like this `+10:00` inside `new Date("2022-10-14T08:12:39.584+10:00")`.

brent-hoover: I think you still don't understand what I am saying

PR #6556, opened 2022-10-13
fix: corrected the validation condition

### LINKED ISSUE #6507: Surcharge resolver treats amount zero als falsy value
This Resolver checks `node.amount` : https://github.com/reactioncommerce/api-plugin-surcharges/blob/43707f9ae1f2272123a8100d9f72ef1c43c01bc4/src/resolvers/Surcharge/index.js#L8
While the intention here is to check if node.amount is defined, which would be the case for a value `0`, it will instead translate to `false`. 

## Steps to Reproduce
1. Create a surcharge with amount 0.
2. Try to query it

## Possible Solution
Switch to e.g. `typeof node.amount !== "undefined"` instead of `node.amount`.

### Versions
latest

shwetha263:
if the amount was zero then node.amount would return false, hence changed the check to check for undefined.

Resolves #6507
Impact: **minor**
Type: **bugfix**

## Issue

Conditional check was incorrect for the resolver. if the amount is 0 then the condition was resolved to false. iif the amount is undefined the condition has to be evaluated to false else for "0" amount also the condition must be evaluated as true.

## Solution

Checking if the amount is not undefined and not empty.

## Breaking changes
None 

## Testing
Tested the condition.

More detail for what each of these sections should include are available in our [Contributing Docs](CONTRIBUTING.md). This project uses [semantic-release](https://semantic-release.gitbook.io/semantic-release/), please use their [commit message format.](https://semantic-release.gitbook.io/semantic-release/#commit-message-format).


brent-hoover:
@shwetha263 Thanks for the fix. When we fix an issue like this I would really like to see a test covering it.

vanpho93 (dismissed):

shwetha263:
@vanpho93 @zenweasel : I added test case for create surcharges with amount 0 and undefined. 
Let me know if this is ok .

brent-hoover:
@shwetha263 Great, thanks so much. Please remember to add a changeset as well to bump the package version. https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md

apps/reaction/tests/integration/api/mutations/surcharges/createSurcharge.test.js
```diff
@@ -122,3 +122,55 @@ test("an authorized user can create a surcharge", async () => {
...
+          messagesByLanguage: surchargeMessagesByLanguage,
+          type: "surcharge",
+          attributes: surchargeAttributes,
+          destination: surchargeDestination
+        }
+      }
+    });
+  } catch (error) {
+    expect(error).toBeUndefined();
+  }
```
vanpho93, line 145: I think this would be enough.

```suggestion
  const  result = await createSurcharge({
      createSurchargeInput: {
        shopId: opaqueShopId,
        surcharge: {
          amount: 0,
          messagesByLanguage: surchargeMessagesByLanguage,
          type: "surcharge",
          attributes: surchargeAttributes,
          destination: surchargeDestination
        }
      }
    });

```

apps/reaction/tests/integration/api/mutations/surcharges/createSurcharge.test.js
```diff
@@ -122,3 +122,55 @@ test("an authorized user can create a surcharge", async () => {
...
+          destination: surchargeDestination
+        }
+      }
+    });
+  } catch (error) {
+    //expect error to be thrown
+    expect(error);
+  }
+  
+});
```
vanpho93: I am not sure about this, should be cannot or can?

```suggestion
test("an authorized user cannot create a surcharge with amount undefined", async () => {
  await testApp.setLoggedInUser(mockAdminAccount);

  try {
    await createSurcharge({
      createSurchargeInput: {
        shopId: opaqueShopId,
        surcharge: {
          amount: undefined,
          messagesByLanguage: surchargeMessagesByLanguage,
          type: "surcharge",
          attributes: surchargeAttributes,
          destination: surchargeDestination
        }
      }
    });
  } catch (error) {
    //expect error to be thrown
    expect(error);
  }
});
```

brent-hoover:
@shwetha263 also linting is failing now

apps/reaction/tests/integration/api/mutations/surcharges/createSurcharge.test.js
shwetha263: It is cannot . corrected the test case name

shwetha263:
> @shwetha263 also linting is failing now
@zenweasel 
Have fixed it

brent-hoover (dismissed):

brent-hoover:
@shwetha263 Changeset being the last thing we need to bump the package https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md

shwetha263:
> @shwetha263 Changeset being the last thing we need to bump the package 
@zenweasel  
https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md

Added the changeset with the summary. Let me know if i need to add more details.

.changeset/surcharge-resolver-amount-updated.md
```diff
@@ -0,0 +1,6 @@
+---
+"reaction": major
+"@reactioncommerce/api-plugin-surcharges": major
```
vanpho93: @zenweasel
Do you think these are correct bumps?

brent-hoover: No, a bug fix would just be a patch change.

brent-hoover:
@shwetha263 Thanks so much for this. Can you correct the changeset to just be a patch change? The [semver standard](https://semver.org/) would define a backwards compatible bug fix as just a patch change. Thanks again

shwetha263:
Hi @zenweasel 
I have updated the change set.

brent-hoover (approved):

PR #6560, opened 2022-10-14
fix: unable to create github release

vanpho93:
Impact: minor
Type: bugfix

## Issue

Unable to create GitHub release because the CHANGELOG file is too big.

## Solution

Move old changes into RESERVED_CHANGELOG

### DIFF
--- a/apps/reaction/CHANGELOG.md
+++ b/apps/reaction/CHANGELOG.md
@@ -6,12 +6,12 @@
 
 - updated CircleCI config to also tag image with latest. ([ec745c7](https://github.com/reactioncommerce/reaction/commit/ec745c709d00856b822239dbfdf494c2c7d1b872))
 
-## 4.1.8(https://github.com/reactioncommerce/reaction/compare/v4.1.7...v4.1.8) (2022-03-02)
 
+## 4.1.8
 
 ### Bug Fixes
 
-* update sharp to match files ([f4ad49e](https://github.com/reactioncommerce/reaction/commit/f4ad49e9a07a49e1805013b3ec2f503361a80242))
+- update sharp to match files ([f4ad49e](https://github.com/reactioncommerce/reaction/commit/f4ad49e9a07a49e1805013b3ec2f503361a80242))
 
 ## [4.1.7](https://github.com/reactioncommerce/reaction/compare/v4.1.6...v4.1.7) (2022-02-14)
 
--- a/apps/reaction/CHANGELOG.md
+++ b/apps/reaction/CHANGELOG.md
@@ -1,11 +1,12 @@
-# [4.2.0](https://github.com/reactioncommerce/reaction/compare/v4.1.8...v4.2.0) (2022-03-26)
+# reaction
 
+## 4.2.0
 
-### Features
+### Minor Changes
 
-* updated CircleCI config to also tag image with latest. ([ec745c7](https://github.com/reactioncommerce/reaction/commit/ec745c709d00856b822239dbfdf494c2c7d1b872))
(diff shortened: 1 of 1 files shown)


brent-hoover (dismissed):

tedraykov (changes requested):
Actually, the problem can be solved in a better way.

The changeset action releases the API with such a big changeset because it reads the `CHANGELOG.md` and tries to parse the latest version. The action expects a different format of the changelog file than the one we have.

Let's say we have the following `CHANGELOG.md` file at the beginning:
```markdown
# [4.2.0](https://github.com/reactioncommerce/reaction/compare/v4.1.8...v4.2.0) (2022-03-26)

### Features

* updated CircleCI config to also tag image with latest. ([ec745c7](https://github.com/reactioncommerce/reaction/commit/ec745c709d00856b822239dbfdf494c2c7d1b872))

## [4.1.8](https://github.com/reactioncommerce/reaction/compare/v4.1.7...v4.1.8) (2022-03-02)


### Bug Fixes

* update sharp to match files ([f4ad49e](https://github.com/reactioncommerce/reaction/commit/f4ad49e9a07a49e1805013b3ec2f503361a80242))

## [4.1.7](https://github.com/reactioncommerce/reaction/compare/v4.1.6...v4.1.7) (2022-02-14)
... [truncated]
```

We make changes to the code and run `changeset` to create the changeset. When we push the changes, the changeset actions run `changeset version` and creates the `Version Packages` PR. What `changeset version` does is it bumps the versions, remove the intermediary changelog file and prepend the latest changelogs to the  `CHANGELOG.md` file in each changed package. The problem is how it actually prepends it to the `apps/reaction/CHANGELOG.md`. This is how it looks like after the `changeset version` script is ran:
```markdown
# [4.2.0](https://github.com/reactioncommerce/reaction/compare/v4.1.8...v4.2.0) (2022-03-26)

## 4.3.0

### Minor Changes

- This is a new feature

### Features

- updated CircleCI config to also tag image with latest. ([ec745c7](https://github.com/reactioncommerce/reaction/commit/ec745c709d00856b822239dbfdf494c2c7d1b872))

## [4.1.8](https://github.com/reactioncommerce/reaction/compare/v4.1.7...v4.1.8) (2022-03-02)
... [truncated]
```

This format is broken and the `Create github Release` script doesn't know what to do with it. The way to fix it is to refactor the `CHANGELOG.md` file in the reaction app.
This is the desired format:
```markdown
# reaction

## 4.2.0

### Minor Changes

- updated CircleCI config to also tag image with latest. ([ec745c7](https://github.com/reactioncommerce/reaction/commit/ec745c709d00856b822239dbfdf494c2c7d1b872))

## 4.1.8(https://github.com/reactioncommerce/reaction/compare/v4.1.7...v4.1.8) (2022-03-02)

### Bug Fixes
...
```
As you can see only the last version should comply with the `changeset` format because it's the one that the script is analyzing. We can leave the rest of the versions in the file as is.

tedraykov:
The same fix needs to be done for `@reactioncommerce/file-collections`, `@reactioncommerce/file-collections-sa-gridfs`, and `@reactioncommerce/file-collections-sa-base`

vanpho93:
> The same fix needs to be done for `@reactioncommerce/file-collections`, `@reactioncommerce/file-collections-sa-gridfs`, and `@reactioncommerce/file-collections-sa-base`

No need to do that, I will just remove created releases for these.

tedraykov:
> > The same fix needs to be done for `@reactioncommerce/file-collections`, `@reactioncommerce/file-collections-sa-gridfs`, and `@reactioncommerce/file-collections-sa-base`
> 
> No need to do that, I will just remove created releases for these.

Actually, the problem with these packages is that they had `CHANGELOG.md` files as well prior to the monorepo and are having the same issue of dumping everything from the `CHANGELOG.md` into the GitHub release. That's why I mentioned them.

tedraykov (approved):

PR #6563, opened 2022-10-14
fix: awaited recursive createHandle call #6496

### LINKED ISSUE #6496: Recursive createHandle call should be awaited
## Issue Description
`createHandle` is a recursive function to determine an available product handle.
It is async due to the database queries which need to be awaited.
The recursive call should be awaited aswell:
https://github.com/reactioncommerce/api-plugin-products/blob/cfe03ad3064d431018d1e30750c4ea16f1e775dc/src/utils/createHandle.js#L69

## Steps to Reproduce (untested)
1. Have a product with slug in your db, and another product with the same slug + "-copy" at the end
2. Try adding a new product with that slug

## Possible Solution
Await `createHandle`.

### Versions
trunk

smriti0710:
Resolves https://github.com/reactioncommerce/reaction/issues/6496
Impact: **minor**
Type: **bugfix**

## Issue

createHandle is a recursive function to determine an available product handle.
It is async due to the database queries which need to be awaited.
The recursive call should be awaited aswell

## Solution

Await createHandle.


delagroove:
@smriti0710 you need to sign-off the commit in order for us to review it/merge it

https://github.com/reactioncommerce/reaction/pull/6563/checks?check_run_id=8897369421

smriti0710:
@delagroove signed off the commit.

brent-hoover:
@smriti0710 you need to [add a changeset](https://github.com/changesets/changesets/blob/master/docs/adding-a-changeset.md) to bump the package. A bug fix would be a "patch" level change

smriti0710:
Please ignore this PR. Created a new one with the fix : https://github.com/reactioncommerce/reaction/pull/6565

Issue #6468, opened 2022-09-06
As an admin I want order discounts to be pro-rated across items in the cart


brent-hoover:
For a $10 discount where there are two items of $50 a piece, each item would be discounted $5

sushmitha-malae:
@zenweasel I would like to work on this, could you please assign this to me?

brent-hoover:
This work is already part of a much larger project that is already underway. So it doesn't make sense as a one off ticket. It would be better to focus on the tickets that have the `hacktoberfest` tag.

brendan-shea-pm:
@zenweasel Does the user have other options on how to calculate the discount or will this be the standard? Asking because if there are other options I'll want to mock that up in the UI, otherwise I think this would just be a backend issue

brent-hoover:
No, no other options

brent-hoover:
Completed

Issue #6462, opened 2022-09-06
As a customer I want to be able to extend the ways that discount calculations can be done via a plug in


brent-hoover:

brent-hoover:
Moving this to Discounts Epic/Sprint

brent-hoover:
Completed

Issue #6459, opened 2022-09-06
As an admin I want to control how a discount is calculated


brent-hoover:
Discounts can be calculated by:

- Percentage off
- Fixed amount off
- Flat amount (set it to a exact amount)

brent-hoover:
Completed

Issue #6453, opened 2022-09-06
As an admin I want to be able to exclude certain items from eligibility for promotions


brent-hoover:
Some items may not be eligible to apply against things like Merchandise Total

zenweasel:
Some items may not be eligible to apply against things like Merchandise Total

brent-hoover:
Applied loyalty rewards WOULD lower eligible amount

brent-hoover:
Completed

PR #6575, opened 2022-10-19
feat: promotions discounts v2

### LINKED ISSUE #6458: As an admin I want to be able to apply a discount to an order via a trigger
example : $10 off $50 or more

trigger qualifier: Cart Value ( i.e $50 )

This discount would have exclusion criteria. So we can apply the discount to items that meet criteria:

- Brand
- Class
- Tag

vanpho93:
Resolves #6458 #6462 #6459 #6453 #6468
Impact: major
Type: feature

## Issue

Missing discounts in the new promotion framework

## Solution

Adding api-plugin-discounts


apps/reaction/tests/integration/api/mutations/discountCodes/discountCodes.test.js
```diff
@@ -61,7 +61,7 @@ beforeAll(async () => {
 // test file gets its own test database.
 afterAll(() => testApp.stop());
 
-test("user can add a discount code", async () => {
+test.skip("user can add a discount code", async () => {
```
brent-hoover: rather than skip, if these are no longer relevant, let's remove

packages/api-plugin-promotions-discounts/src/discountTypes/order/applyOrderDiscountToCart.js
```diff
@@ -0,0 +1,86 @@
...
+ * @summary Splits a discount across all cart items
+ * @param {Number} totalDiscount - The total discount to split
+ * @param {Array<Object>} cartItems - The cart items to split the discount across
+ * @returns {void} undefined
+ */
+export function splitDiscountForCartItems(totalDiscount, cartItems) {
+  const totalItemPrice = cartItems.reduce((acc, item) => acc + item.subtotal.amount, 0);
+  const discountForEachItems = cartItems.map((item) => {
+    const discount = (item.subtotal.amount / totalItemPrice) * totalDiscount;
+    return { _id: item._id, amount: Number(accounting.toFixed(discount, 2)) };
```
brent-hoover: Need to handle amounts that don't divide evenly, e.g. $10 over 3 items = 3.33 * 3 = 9.99 !== $10

packages/api-plugin-promotions-discounts/src/discountTypes/shipping/getGroupDisountTotal.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable no-unused-vars */
+
+/**
+ * @summary Get the group discount total for a order
+ * @param {Object} context - The application context
+ * @param {Object} params.commonOrder - The order to get the group discount total for
+ * @returns {Number} The total discount amount for the order
+ */
+export default function getGroupDiscountTotal(context, { commonOrder }) {
```
brent-hoover: Can you explain what this does?

vanpho93: Temporarily  removed.

brent-hoover (changes requested):
Need to update the sample data to reflect these changes

brent-hoover:
Applied discounts is not coming through on the order.

Also I think should include at least one top level integration test that tests the functionality end to end:

1. Create a promotion (using the mutation preferably)
2. Create a cart (there is existing test stuff you can reuse)
3. Validate that the promotions and discounts are applied
4. Place an order
5. Validate that the promotions and discounts are carried through to the order

I think this will really help us as we make tweaks to this to ensure we don't break it

apps/reaction/tests/integration/api/mutations/checkout/promotionCheckout.test.js
```diff
@@ -0,0 +1,268 @@
...
+        cartToken,
+        fulfillmentGroupId: opaqueFulfillmentGroupId
+      }
+    });
+
+    const option = result.updateFulfillmentOptionsForGroup.cart.checkout.fulfillmentGroups[0].availableFulfillmentOptions[0];
+    opaqueFulfillmentMethodId = option.fulfillmentMethod._id;
+  });
+
+  test("select the `Standard mockMethod` fulfillment option", async () => {
```
brent-hoover, line 184: These are set up like separate tests but in fact there is only one test, which should describe what it's testing, e.g. "place an order with discount and get the correct values"

packages/api-plugin-discounts/package.json
```diff
@@ -34,5 +34,10 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "scripts": {
+    "test": "jest --passWithNoTests",
```
brent-hoover: We should remove this `passWithNoTests` because it should not pass with no tests

packages/api-plugin-discounts/src/util/setDiscountsOnCart.test.js
```diff
@@ -0,0 +1,59 @@
+import setDiscountsOnCart from "./setDiscountsOnCart.js";
+
+jest.mock("../queries/getDiscountsTotalForCart.js", () => jest.fn().mockReturnValue({ total: 10 }));
+
+test("should set discounts on cart when discountCalculationMethods doesn't existd", async () => {
```
brent-hoover: typo in test description `exist`

packages/api-plugin-promotions-discounts/src/actions/discountAction.js
```diff
@@ -0,0 +1,100 @@
...
+ * @returns {Promise<void>} undefined
+ */
+export async function discountActionHandler(context, cart, params) {
+  const { discountType } = params.actionParameters;
+
+  Logger.info({ params, cartId: cart._id, ...logCtx }, "applying discount to cart");
+
+  const { cart: updatedCart } = await functionMap[discountType](context, params, cart);
+
+  Logger.info(logCtx, "Completed applying Discount to Cart");
```
brent-hoover: Should include enough info in each logging line to ascertain what cart, etc. it's about. Rarely are you going to read logs line by line but searching through them with something like Kibana

packages/api-plugin-promotions-discounts/src/discountTypes/item/applyItemDiscountToCart.test.js
```diff
@@ -0,0 +1,136 @@
...
+  expect(itemDiscount).toEqual({
+    promotionId: "promotion1",
+    discountType: "test",
+    discountCalculationType: "test",
+    discountValue: 10,
+    dateApplied: expect.any(Date)
+  });
+});
+
+test("should return cart with applied discount when parameters not include rule", async () => {
```
brent-hoover: typo in test description

packages/api-plugin-promotions-discounts/src/methods/index.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {Number} discountValue - The discount value
+ * @param {Number} price - The price to calculate the discount for
+ * @returns {Number} The discount amount
+ */
+function percentage(discountValue, price) {
+  return price * (1 - discountValue / 100);
+}
+
+/**
+ * @summary Calculates the discount amount for the fixed discount type
```
brent-hoover: flat not fixed in JSDoc

brent-hoover (approved):
A few typos and things but it looks good

traykov1 (commented):
For some reason when i diff the working tree of the base `feat/promotions` branch with this PR's branch, it says that it deletes the defaultRoles for the promotions.
```
"reaction:legacy:promotions/create",
"reaction:legacy:promotions/read",
"reaction:legacy:promotions/update"
  ```

**Comment by brent-hoover** [comment]
I think Brian just hasn't rebased against the feature branch as that just got merged today

File: packages/api-plugin-promotions-discounts/src/discountTypes/order/applyOrderDiscountToCart.js [hunk 10]
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * @summary Get the discount amount for a discount item
+ * @param {Object} context - The application context
+ * @param {Array<Object>} items - The cart to calculate the discount for
+ * @param {Object} discount - The discount to calculate the discount amount for
+ * @returns {Number} - The discount amount
+ */
+export function getCartDiscountAmount(context, items, discount) {
+  const merchandiseTotal = getTotalEligibleItemsAmount(items);
```
tedraykov: Why do we calculate the merchandise total here, when the enhancer is doing the same?
Another question is if we calculate it here, why do we need the enhancer?

tedraykov (commented):
- In the `api-plugin-promotions` graphql schema the `PromotionFilter` input should be suffixed by `Input` to match the style of all graphql inputs. 

- The `createPromotion` mutation should use `PromotionInput` simple schema for the input. Also the following code in the same mutation:
```
  const { triggerKey } = promotions.triggers[0];
  const trigger = promotions.triggers.find((tr) => tr.triggerKey === triggerKey);
  promotion.triggerType = trigger.type;
```
Can't it be simplified to:
```
promotion.triggerType = promotions.triggers[0].type;
```

tedraykov:
Regarding the `success` field in the `PromotionCreateUpdatePayload` I would suggest we create and accept an ARD before we start introducing new schema styles.

packages/api-plugin-discounts/src/util/setDiscountsOnCart.js
```diff
@@ -7,6 +7,9 @@ import getDiscountsTotalForCart from "../queries/getDiscountsTotalForCart.js";
  * @returns {undefined}
  */
 export default async function setDiscountsOnCart(context, cart) {
-  const { total } = await getDiscountsTotalForCart(context, cart);
-  cart.discount = total;
+  // check if promotion discounts are enabled
```
tedraykov, line 10: Don't we want to remove this plugin altogether?

tedraykov:
Overall, the plugin looks excellent with the sample data. I still don't have a settled mindmap of the whole framework but it seems like it's really extendable.
In this PR I don't see being addressed what happens with existing carts with applied promotions when existing promotion parameters are changed but I assume it will be addressed separately.

packages/api-plugin-discounts/src/util/setDiscountsOnCart.js
vanpho93: I think adding this code is safer. Someone may still install both dependencies.

packages/api-plugin-promotions-discounts/src/discountTypes/order/applyOrderDiscountToCart.js
vanpho93: this variable should be rename to `totalEligibleItemsAmount`, this total is different, it counts inclusion and exclusion and used for discount value calculation.

Enhancer can do sth more dynamic like total of item quantity. Enhancer is at general level of the framework. this `totalEligibleItemsAmount` is in detail of discount plugin.

packages/api-plugin-discounts/src/util/setDiscountsOnCart.js
brent-hoover: > * In the `api-plugin-promotions` graphql schema the `PromotionFilter` input should be suffixed by `Input` to match the style of all graphql inputs.
> * The `createPromotion` mutation should use `PromotionInput` simple schema for the input. Also the following code in the same mutation:
> 
> ```
>   const { triggerKey } = promotions.triggers[0];
>   const trigger = promotions.triggers.find((tr) => tr.triggerKey === triggerKey);
>   promotion.triggerType = trigger.type;
> ```
> 
> Can't it be simplified to:
> 
> ```
> promotion.triggerType = promotions.triggers[0].type;
> ```

No, because it may be an array with multiple triggers at some point

zenweasel:
> Regarding the `success` field in the `PromotionCreateUpdatePayload` I would suggest we create and accept an ARD before we start introducing new schema styles.

These comments are about code not in this PR

brent-hoover:
> Regarding the `success` field in the `PromotionCreateUpdatePayload` I would suggest we create and accept an ARD before we start introducing new schema styles.

I agree about putting together an (an ADR is what I am assuming you meant) but I don't think we want to introduce new legacy code here

brent-hoover:
> In this PR I don't see being addressed what happens with existing carts with applied promotions when existing promotion parameters are changed but I assume it will be addressed separately.

Yes, this is just discounts. This is being addressed [here](https://github.com/reactioncommerce/reaction/issues/6612)

packages/api-plugin-discounts/src/util/setDiscountsOnCart.js
brent-hoover: > Don't we want to remove this plugin altogether?

The plugin should be removed but I would like to retain backwards compatibility if possible

vanpho93:
> * In the `api-plugin-promotions` graphql schema the `PromotionFilter` input should be suffixed by `Input` to match the style of all graphql inputs.
> * The `createPromotion` mutation should use `PromotionInput` simple schema for the input. Also the following code in the same mutation:
> 
> ```
>   const { triggerKey } = promotions.triggers[0];
>   const trigger = promotions.triggers.find((tr) => tr.triggerKey === triggerKey);
>   promotion.triggerType = trigger.type;
> ```
> 
> Can't it be simplified to:
> 
> ```
> promotion.triggerType = promotions.triggers[0].type;
> ```

@zenweasel  Do you have opinion on this, it seems like the API part. I think it should address these in another PR.

brent-hoover:
Yes, all the comments about the api should be dealt in a different pr

PR #6576, opened 2022-10-19
fix: updateSurcharge returns the entire document from db

### LINKED ISSUE #6508: updateSurcharge should return the full item from db
## Issue Description
The `updateSurcharge `Mutation will only return the updated fields instead of returning the full item from db including the changes here: https://github.com/reactioncommerce/api-plugin-surcharges/blob/98cb4ca48fcd53edbfd36dedbe6986bfe8bf795c/src/mutations/updateSurcharge.js#L40

The mutation will fail to return properly, if mandatory fields like `createdAt` are not part of the update.

## Steps to Reproduce
1. Create a surcharge
2. Use updateSurcharge to update e.g. the amount and query all Surcharge fields in the return
3. Observe that `Cannot return null for non-nullable field Surcharge.createdAt.` is thrown.

## Possible Solution
Use `findOneAndUpdate` with `returnOriginal: false` would probably be a better fit here, to return the full upldated item

### Versions
trunk

sushmitha-malae:
Signed-off-by: Sushmitha Malae <malaesushmitha@gmail.com>

Resolves #6508 
Impact: **minor**
Type: **bugfix**

## Issue

The updateSurcharge Mutation will only return the updated fields instead of returning the full item from db

## Solution

Returning the updated full document instead of the partially updated fields.

## Breaking changes

None

## Testing

Modified the existing utc, to test the change.

More detail for what each of these sections should include are available in our [Contributing Docs](CONTRIBUTING.md). This project uses [semantic-release](https://semantic-release.gitbook.io/semantic-release/), please use their [commit message format.](https://semantic-release.gitbook.io/semantic-release/#commit-message-format).


sushmitha-malae:
@delagroove Please review the changes, Thanks!

delagroove:
I'm a bit busy today but if @vannguyenn @vanpho93 @sujithvn can go ahead and review that would work as well!

sushmitha-malae:
@zenweasel do you have some time for the review?

brent-hoover (approved):

PR #6578, opened 2022-10-19
fix: allow restrictions on nested properties

### LINKED ISSUE #6501: Attribute checks in shipping restrictions don't work for nested item properties (e.g. parcel.*)
## Issue Description
The checks for shipping restrictions based on an item's properties only work top for level properties such as price. An attribute deny rule that uses the property "weight", as mentioned in the documentation (https://docs.reactioncommerce.com/docs/concepts-fulfillment), does not work, since the property "weight" is nested in "parcel".

## Steps to Reproduce
1. Create a product with the parcel.weight attribute set
2. Create a shipping method
3. Create a shipping restriction with the attribute "weight" which would exclude the shipping method for the product if it was applied (as explained in the documentation)
4. See that the shipping method is available for the product in checkout process

## Possible Solution
Access item properties with `_.at(item, attribute.property)` in attributeDenyCheck.js. Change the property in the documentation to "parcel.weight"

### Versions
api-plugin-shipments-flat-rate@1.0.4
reaction@3.11.0

range123:
Resolves https://github.com/reactioncommerce/reaction/issues/6501
Impact: **minor**
Type: **bugfix**

## Issue
The checks for shipping restrictions based on an item's properties only work top for level properties in the item. It doesn't work for nested properties such as `price.amount` or `subtotal.amount`.

Steps to reproduce:

1. Create a product, create a variant with the price attribute set.
2. Create a shipping method.
4.  Create a "deny" shipping restriction with the attribute `price.amount` which would exclude the shipping method for the product if it was applied.
5. See that the shipping method is available for the cart in checkout process even when it shouldn't be (based on restriction `price` value).

## Solution

Can be solved by using lodash `_.get()` for getting the nested property from the item, instead of using the default javascript getter. This would make sure the restrictions work with both nested and non-nested attributes.

## Testing

1. Create a product, create a variant with the weight and price attribute set.
2. Create a shipping method.
3. Create "deny" shipping restriction with the attribute `weight` which would exclude the shipping method for the product if it was applied.
4.  Create another "deny" shipping restriction with the attribute `price.amount` which would exclude the shipping method for the product if it was applied.
5. See that the shipping method is available/unavailable for the cart in checkout process (based on the restrictions).

### DIFF
--- a/.changeset/modern-poems-smash.md
+++ b/.changeset/modern-poems-smash.md
@@ -1,0 +1,5 @@
+---
+"@reactioncommerce/api-plugin-shipments-flat-rate": patch
+---
+
+allow nested properties to be accessible for shipment restrictions.
--- a/packages/api-plugin-shipments-flat-rate/package.json
+++ b/packages/api-plugin-shipments-flat-rate/package.json
@@ -30,7 +30,8 @@
     "@reactioncommerce/logger": "^1.1.3",
     "@reactioncommerce/random": "^1.0.2",
     "@reactioncommerce/reaction-error": "^1.0.1",
-    "simpl-schema": "^1.12.0"
+    "simpl-schema": "^1.12.0",
+    "lodash": "^4.17.15"
   },
   "devDependencies": {
     "@babel/core": "^7.7.7",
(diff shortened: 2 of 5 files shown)


brent-hoover:
@range123 I would like to see a test case for this

range123:
Added tests! @zenweasel

brent-hoover (approved):

PR #6570, opened 2022-10-18
01 fulfillment base

sujithvn:
Signed-off-by: Sujith mail.sujithvn@gmail.com

---
# This PR has been replaced by PR #6610 
---

Resolves #6471 
Impact: **breaking**
Type: **feature**

## Issue

In the current system, the only way for order fulfillment is via shipping which is hardcoded into the system. We need to introduce the flexibility for users to add their own fulfillment types via plugins.

## Solution

We are creating a base fulfillment plugin which would enable other fulfillment types (like shipping, pickup, digital) to be introduced via plugins. Each of the newly introduced fulfillment type plugin would need to have the specific fulfillment methods also to be added as separate plugins. Example, fulfillment type 'pickup' could have fulfillment methods like 'store pickup' and 'curb-side pickup'.

## PR Details in order

```
feat/fulfillment-types
 | 
 | __ 01 fulfillment base <--- This PR #6570 
 |        | __ 01-01 fulfillment-type-shipping       #6579
 |        | __ 01-02 fulfillment-type-pickup. #6577 
 |__02 fulfillment-type-impacted-plugins. [carts, products, orders, catalogs] #6582 
 |__03 placeOrder refactor #6583 

```
## Existing PRs based on the #6480 (to be updated to point to #6570) 
We shall change the base branch of the 3 existing PRs (i18n #6545, dataMigration #6544, setDefaultFFtype #6543) to  point to 01 fulfillment base

##### API-PLUGIN-FULFILLMENT
This is the base plugin, basic features listed below. 
Mutations
* createFulfillmentType - this is for internal call only from ff-type plugins (NEW)
* createFulfillmentMethod  - this is for internal call only from ff-method plugins (NEW)
* updateFulfillmentType - public mutation which allows to update certain specific fields of the ff-type (NEW)
* updateFulfillmentMethod - public mutation which allows to update certain specific fields of the ff-type (NEW)
* updateFulfillmentOptionsForGroup - Existing code/functionality from the old shipment plugin. This updates the cart with quote details from all available ff-methods for the provided ff-type
* selectFulfillmentOptionForGroup - Existing code/functionality from the old shipment plugin. This updates the cart with one of the selected ff-method from the available quote presented by above mutation.

Queries
* getFulfillmentMethodsWithQuotes - Existing code/functionality from the old shipment plugin. Calls the quote providing functions by collecting them via functionByType. Calls are made only for the specific methods of the provided ff-type.
[...]

PR #6545, opened 2022-10-05
Initial i18n for Fulfillment

sujithvn:
Signed-off-by: Sujith <mail.sujithvn@gmail.com>

Resolves #6473 
Impact: **minor**
Type: **feature**

## Issue

There are no localised versions of customer or admin facing messages.

## Solution

Created i18n config files for the newly introduced plugins. 

The required files for two of the below mentioned plugins were completely copied over from old plugin. 
`api-plugin-fulfillment-type-shipping` (copied over from `api-plugin-shipments`)
`api-plugin-fulfillment-method-shipping-flat-rate` (copied over from `api-plugin-shipments-flat-rate`)

Added new config files for remaining 4 plugins below
`api-plugin-fulfillment`
`api-plugin-fulfillment-type-pickup`
`api-plugin-fulfillment-method-pickup-store`
`api-plugin-fulfillment-method-shipping-ups`

## Breaking changes
None

## Testing
Need to co-ordinate with the Front-end team to understand if there are any additional entries required in the config-files for the new plugins.

PR #6582, opened 2022-10-20
02 fulfillment type impacted plugins

sujithvn:
# This PR is replaced by https://github.com/reactioncommerce/reaction/pull/6615

Resolves #6471 
Impact: **breaking**
Type: **feature**

## Issue

## Solution

This PR is 4th entry (02 fulfillment-type-impacted-plugins)

## PR Details in order

```
feat/fulfillment-types
 | 
 | __ 01 fulfillment base #6570 
 |        | __ 01-01 fulfillment-type-shipping       #6579
 |        | __ 01-02 fulfillment-type-pickup.    #6577 
 |__02 fulfillment-type-impacted-plugins. [carts, products, orders, catalogs]        <--- This PR #6582 
 |__03 placeOrder refactor #6583 

```
## Existing PRs based on the #6480 (to be updated to point to #6570) 
We shall change the base branch of the 3 existing PRs (i18n #6545, dataMigration #6544, setDefaultFFtype #6543) to  point to 01 fulfillment base

##### CARTS

##### CATALOGS

##### PRODUCTS

##### ORDERS

##### Integration Tests updated
* Updated the integration tests to match the changes in plugins.

## Breaking changes

## Testing


brent-hoover:
@sujithvn Broken build on this as well? Is this expected?

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,124 @@
...
  * @param {Object} context App context
  * @param {Object} cart The cart, to be mutated
  * @returns {undefined}
  */
-export default function updateCartFulfillmentGroups(context, cart) {
+export default async function updateCartFulfillmentGroups(context, cart) {
   // Every time the cart is updated, create any missing fulfillment groups as necessary.
   // We need one group per type per shop, containing only the items from that shop.
   // Also make sure that every item is assigned to a fulfillment group.
+  const { collections } = context;
```
brent-hoover, line 80: Combine these destructuring statements

pnpm-lock.yaml
```diff
@@ -245,7 +245,7 @@ importers:
       '@reactioncommerce/logger': link:../../packages/logger
       '@reactioncommerce/nodemailer': 5.0.5
       '@reactioncommerce/random': link:../../packages/random
-      '@snyk/protect': 1.1032.0
```
brent-hoover, line 248: You should remove this (Snyk), it's going to just keep causing problems

packages/api-plugin-carts/src/mutations/transformAndValidateCart.js
```diff
@@ -15,7 +15,7 @@ const logCtx = { name: "cart", file: "transformAndValidateCart" };
  */
 export default async function transformAndValidateCart(context, cart) {
   const { simpleSchemas: { Cart: cartSchema } } = context;
-  updateCartFulfillmentGroups(context, cart);
+  await updateCartFulfillmentGroups(context, cart);
```
brent-hoover, line 18: A little concerned about this. @aldeed was there a reason we didn't await these transforms?

brent-hoover (changes requested):
A couple more changes

packages/api-plugin-carts/src/mutations/transformAndValidateCart.js
sujithvn: Earlier we did not have any async calls in the `updateCartFulfillmentGroups`. Now we have added a new async call to check if the fulfillment-type selected is currently enabled/disabled. That is why I added a new await here now.

pnpm-lock.yaml
sujithvn: removed

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
sujithvn: done

sujithvn:
This PR is replaced by https://github.com/reactioncommerce/reaction/pull/6615

PR #6583, opened 2022-10-20
03 placeOrder refactor and validateOrder

sujithvn:
# This PR is replaced by #6616 
 
Resolves #6471 
Impact: **breaking**
Type: **feature**

## Issue

## Solution

This PR is 5th entry (03 placeOrder refactor  and validateOrder)

## PR Details in order

```
feat/fulfillment-types
 | 
 | __ 01 fulfillment base #6570 
 |        | __ 01-01 fulfillment-type-shipping       #6579
 |        | __ 01-02 fulfillment-type-pickup.    #6577 
 |__02 fulfillment-type-impacted-plugins. [carts, products, orders, catalogs].   #6582 
 |__03 placeOrder refactor         <--- This PR   #6583 

```
## Existing PRs based on the #6480 (to be updated to point to #6570) 
We shall change the base branch of the 3 existing PRs (i18n #6545, dataMigration #6544, setDefaultFFtype #6543) to  point to 01 fulfillment base

### Details (specific to placeOrder refactor and introduction of validateOrder end-point)

## Breaking changes

## Testing


packages/api-plugin-orders/src/mutations/placeOrder.js
```diff
@@ -112,173 +8,10 @@ async function createPayments({
...
-    // See https://eslint.org/docs/rules/no-await-in-loop#when-not-to-use-it
-    for (const transformCustomOrderFieldsFunc of transformCustomOrderFieldsFuncs) {
-      customFields = await transformCustomOrderFieldsFunc({ context, customFields, order }); // eslint-disable-line no-await-in-loop
-    }
-    order.customFields = customFields;
-  } else {
-    order.customFields = customFieldsFromClient;
-  }
+  const { appEvents, collections, userId } = context;
+  const { Orders } = collections;
```
brent-hoover: These two destructuring statements can be combined

packages/api-plugin-orders/src/queries/validateOrder.js
```diff
@@ -0,0 +1,16 @@
+import prepareOrder from "../util/orderValidators/prepareOrder.js";
+
+/**
+ * @name validateOrder
+ * @method
+ * @memberof Order/NoMeteorQueries
```
brent-hoover: Remove the NoMeteorQueries

packages/api-plugin-orders/src/util/orderValidators/getFinalFulfillmentGroups.js
```diff
@@ -0,0 +1,147 @@
...
+  // Add more props to each fulfillment group, and validate/build the items in each group
+
+  const finalFulfillmentGroups = await Promise.all(fulfillmentGroups.map(async (inputGroup) => {
+    const { data, items, selectedFulfillmentMethodId, shopId, totalPrice: expectedGroupTotal, type } = inputGroup;
+    const group = {
+      _id: Random.id(),
+      address: data ? data.shippingAddress : null,
+      shopId,
+      type,
+      workflow: { status: "new", workflow: ["new"] }
```
brent-hoover, line 52: Shouldn't this workflow be based on the fulfillment group?

brent-hoover (changes requested):

packages/api-plugin-orders/src/mutations/placeOrder.js
sujithvn: Done

packages/api-plugin-orders/src/queries/validateOrder.js
sujithvn: Done

packages/api-plugin-orders/src/util/orderValidators/getFinalFulfillmentGroups.js
sujithvn: @zenweasel I am not clear on your comment. Could we discuss?

brent-hoover: As discussed previously, the workflow of the order should be defined by the FFT. Since the workflow for a shipped order is very different for a download which is very different for a store pickup

brent-hoover: Has this comment been dealt with?

sujithvn: Not yet, I shall update it today

sujithvn:
This PR is replaced by 6616
https://github.com/reactioncommerce/reaction/pull/6616

PR #6588, opened 2022-10-24
feat: upgrade mongodb to 4.4

### LINKED ISSUE #6383: Update to MongoDB 5
MongoDB 4.2 will reach EOD soon. We should update to MongoDB 5 (if there are even breaking changes at all).

vanpho93:
Signed-off-by: Brian Nguyen [vanpho02@gmail.com](mailto:vanpho02@gmail.com)

Resolves https://github.com/reactioncommerce/reaction/issues/6383
Type: feature

## Issue

MongoDB 4.2 will reach EOD soon. We should update to MongoDB 5.

## Solution

Bump mongodb driver version from v3.6.2 to v4.4.1.
Modify collection object to support v3 queries

## Breaking changes

https://github.com/mongodb/node-mongodb-native/blob/4.1/docs/CHANGES_4.0.0.md

Some changes resolved by this PR:

returnOriginal
{ result: { ok: 1, n: 1 } }

## Testing

Run e2e test.

More detail for what each of these sections should include are available in our [Contributing Docs](https://docs.reactioncommerce.com/reaction-docs/trunk/contributing-to-reaction)

### DIFF
--- a/packages/file-collections-sa-gridfs/package.json
+++ b/packages/file-collections-sa-gridfs/package.json
@@ -68,8 +68,7 @@
   "dependencies": {
     "@babel/runtime-corejs2": "^7.14.8",
     "@reactioncommerce/file-collections-sa-base": "^0.2.2",
-    "debug": "^4.3.2",
-    "gridfs-stream": "^1.1.1"
+    "debug": "^4.3.2"
   },
   "devDependencies": {
     "@babel/cli": "^7.10.5",
--- a/packages/file-collections-sa-gridfs/package.json
+++ b/packages/file-collections-sa-gridfs/package.json
@@ -87,7 +87,7 @@
     "babel-core": "^7.0.0-bridge.0"
   },
   "peerDependencies": {
-    "mongodb": "4.4.1"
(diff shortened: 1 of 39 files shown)


brent-hoover:
Should this be a minor change rather than patch? It's not a bug fix

packages/file-collections-sa-gridfs/package.json
```diff
@@ -87,6 +86,9 @@
     "@babel/preset-env": "^7.10.4",
     "babel-core": "^7.0.0-bridge.0"
   },
+  "peerDependencies": {
+    "mongodb": "4.4.1"
```
aldeed: I find it best to be loose with peer dependency versions because now that npm fails on mismatched peer deps (I assume pnpm does too), you would need to continually update this range to prevent troubles for consumers of this package. So maybe `"mongodb": ">= 4.4.1 < 5"`

aldeed (dismissed):
One suggestion, but LGTM

brent-hoover (dismissed):

brent-hoover (approved):

PR #6609, opened 2022-11-01
feat: add function to calculate eligible total based on rules

brent-hoover:


brent-hoover:
Implemented elsewhere

PR #6610, opened 2022-11-01
00 fulfillment base

sujithvn:

Resolves #6471 
Impact: **breaking**
Type: **feature**

## Issue

## Solution

## PR Details in order

```
feat/fulfillment-types
 | 
 | __ 00 fulfillment base <--- This PR #6610 
 |        | __ 01-01 fulfillment-type-shipping       #6614 (old #6579)
 |        | __ 01-02 fulfillment-type-pickup. #6613 (old #6577) 
 |__02 fulfillment-type-impacted-plugins. [carts, products, orders, catalogs] #6615 (old #6582) 
 |__03 placeOrder refactor #6616 (old #6583)

```
We shall change the base branch of the 3 existing PRs (i18n #6545, dataMigration #6633, setDefaultFFtype #6543) to  point to 01 fulfillment base

##### API-PLUGIN-FULFILLMENT
This is the base plugin, basic features listed below. 
Mutations
* createFulfillmentType - this is for internal call only from ff-type plugins (NEW)
* createFulfillmentMethod  - this is for internal call only from ff-method plugins (NEW)
* updateFulfillmentType - public mutation which allows to update certain specific fields of the ff-type (NEW)
* updateFulfillmentMethod - public mutation which allows to update certain specific fields of the ff-type (NEW)
* updateFulfillmentOptionsForGroup - Existing code/functionality from the old shipment plugin. This updates the cart with quote details from all available ff-methods for the provided ff-type
* selectFulfillmentOptionForGroup - Existing code/functionality from the old shipment plugin. This updates the cart with one of the selected ff-method from the available quote presented by above mutation.

Queries
* fulfillmentType - returns the ff-type matching the provided ID (NEW)
* fulfillmentTypes - returns all the ff-types (NEW) 
* fulfillmentMethods - returns all the ff-methods for the specified ff-type (NEW)
* fulfillmentMethodsWithQuotes - Calls the quote providing functions by collecting them via functionByType. Calls are made only for the specific methods of the provided ff-type.
* getFulfillmentMethodsWithQuotes - Existing code/functionality from the old shipment plugin. Deprecated and calls fulfillmentMethodsWithQuotes internally

Others
* preStartup - extends all impacted schemas mostly with dynamic values for 'allowedValues' (NEW)
* startup - inserts the default 'undecided' entry to the new Fulfillment collection (NEW)
* registration - collects all the "registeredFulfillmentType" from ff-type plugins and adds allRegisteredFulfillmentTypes to context (contextAddition) (NEW)
* util folder - copied over from shipment plugin (Existing code)

## Breaking changes

## Testing


packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.test.js
```diff
@@ -0,0 +1,72 @@
...
+      shopId: "SHOP_ID",
+      cost: 99,
+      fulfillmentMethod: "flatRates",
+      displayMessageMethod: "Sample display message"
+    },
+    fulfillmentTypeId: "fulfillment123",
+    shopId: "SHOP_ID"
+  };
+
+  await expect(createFulfillmentMethodMutation(mockContext, fulfillmentMethodInput)).rejects.toThrowErrorMatchingSnapshot();
```
sujithvn: @aldeed @zenweasel @tedraykov @vanpho93 
Brent has the below suggestion in the old PR [link](https://github.com/reactioncommerce/reaction/pull/6570#discussion_r1004018632) 

> Just my opinion but I think using snapshots are fine for frontend testing but for backend PR's they make it unnecessarily brittle. Open to second opinions on that.

I shall go ahead and make this change unless someone thinks otherwise.

packages/api-plugin-fulfillment/src/schemas/schema.graphql
```diff
@@ -0,0 +1,414 @@
...
+  fulfillmentGroupId: ID!
+
+  "The updated group name"
+  name: String!
+
+  "The updated group label"
+  label: String!
+
+  "Flag defining enabled/disabled status"
+  enabled: Boolean
```
sujithvn: Make this field required. (Refer https://github.com/reactioncommerce/reaction/pull/6570#discussion_r1004039840)

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.test.js
aldeed: I often use the `to*InlineSnaphot` variants with good success. Basically just saves you from needing to type out the objects. No strong preference, though

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.js
```diff
@@ -0,0 +1,63 @@
...
+    _id: fulfillmentTypeId
+  }, {
+    $addToSet: {
+      methods: method
+    }
+  });
+
+  if (matchedCount === 0) throw new ReactionError("server-error", "Unable to create fulfillment method");
+
+  return { group: method };
```
aldeed: Why does this use the property name "group"? Wouldn't "method" be best?

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.js
```diff
@@ -0,0 +1,63 @@
...
+  await context.validatePermissions("reaction:legacy:fulfillmentMethods", "create", { shopId });
+
+  const ffTypeRecord = await Fulfillment.findOne({ _id: fulfillmentTypeId, shopId });
+  if (!ffTypeRecord) throw new ReactionError("server-error", "Unable to create fulfillment method without defined type");
+
+  const ffTypeMethodRecord = await Fulfillment.findOne({
+    _id: fulfillmentTypeId,
+    shopId,
+    methods: { $elemMatch: { fulfillmentMethod: method.fulfillmentMethod } }
+  });
```
aldeed: This does another database call but you already just retrieved this record from the db. So you can do this check in code by using `ffTypeRecord.methods.find()` and save a db call.

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.js
```diff
@@ -0,0 +1,63 @@
...
+ * @param {Object} input - Input
+ * @param {Number} cost - optional cost
+ * @param {String[]} fulfillmentTypes - fulfillment type retained for backward compatibility
+ * @param {String} group - Group to which fulfillment method belong
+ * @param {Number} handling - handling charges
+ * @param {Boolean} enabled - status of fulfillment method
+ * @param {String} label - displayed on the UI
+ * @param {String} fulfillmentMethod - used by application, not user editable
+ * @param {String} displayMessageMethod - used to display additional message on UI specific to ff-method
+ * @param {Number} rate - ratefor themethod
```
aldeed: These field names should clarify the full path to the property. Example, `input.fulfillmentTypeId` or `input.method.cost`

packages/api-plugin-fulfillment/src/simpleSchemas.js
```diff
@@ -0,0 +1,169 @@
...
+ * @property {String} displayMessageType - used to display additional message on UI specific to ff-type
+ * @property {FulfillmentMethodSchema[]} methods - array of ff-methods under this ff-type
+ */
+export const FulfillmentTypeSchema = new SimpleSchema({
+  "name": String,
+  "shopId": String,
+  "provider": {
+    type: ProviderSchema
+  },
+  "fulfillmentType": String,
```
aldeed, line 135: Maybe something like `baseType` would be clearer than `fulfillmentType` since you're already calling each document in this collection a "fulfillment type"?

packages/api-plugin-fulfillment/src/index.js
```diff
@@ -0,0 +1,57 @@
...
+    version: pkg.version,
+    collections: {
+      FulfillmentRestrictions: {
+        name: "FulfillmentRestrictions",
+        indexes: [
+          [{ methodIds: 1 }]
+        ]
+      },
+      Fulfillment: {
+        name: "Fulfillment",
```
aldeed, line 34: Any reason not to name this collection `FulfillmentTypes` everywhere? The schema is named that and there are arguments such as `fulfillmentTypeId`, so it would make sense for the collection name to match.

packages/api-plugin-fulfillment/src/mutations/createFulfillmentType.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+export default async function createFulfillmentType(context, input) {
+  const cleanedInput = FulfillmentTypeSchema.clean(input);
+  if (!cleanedInput.provider) cleanedInput.provider = {};
+  cleanedInput.provider.name = cleanedInput.name;
+
+  if (cleanedInput.method) {
+    cleanedInput.method._id = Random.id();
+    cleanedInput.method.fulfillmentTypes = [cleanedInput.fulfillmentType];
+  }
```
aldeed, line 26: This `if` block seems unnecessary because the input has been cleaned and `method` isn't allowed, which means it will never exist.

packages/api-plugin-fulfillment/src/simpleSchemas.js
```diff
@@ -0,0 +1,169 @@
...
+ * @type {SimpleSchema}
+ * @summary Defines Schema for Fulfillment Type
+ * @property {String} name - name of the fulfillment type, user editable
+ * @property {String} shopId - Shop ID
+ * @property {ProviderSchema} provider - Provider details
+ * @property {String} fulfillmentType - fulfillmentType, not user editable
+ * @property {String} displayMessageType - used to display additional message on UI specific to ff-type
+ * @property {FulfillmentMethodSchema[]} methods - array of ff-methods under this ff-type
+ */
+export const FulfillmentTypeSchema = new SimpleSchema({
```
aldeed, line 129: All new collections are supposed to have `createdAt` and `updatedAt` properties. You'd need to add in the SimpleSchema, in GraphQL, and also set them properly in the create/update mutations. If you look at newer code in other plugins you should be able to find examples of this.

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentMethod.js
```diff
@@ -0,0 +1,68 @@
...
+  // Find the matching fulfillmentMethod object and use those values to over-write
+  const currentFulfillmentMethod = (ffTypeMethodRecord.methods || []).find((meth) => meth._id === methodId);
+  if (!currentFulfillmentMethod) throw new ReactionError("server-error", "Fulfillment Method does not exist");
+  const updatedMethod = {
+    ...method,
+    _id: methodId,
+    name: currentFulfillmentMethod.name,
+    group: currentFulfillmentMethod.group,
+    fulfillmentMethod: currentFulfillmentMethod.fulfillmentMethod,
+    fulfillmentType: [ffTypeMethodRecord.fulfillmentType]
```
aldeed: This could be a fragile way to avoid users overwriting unintended props. Instead of using `method: FulfillmentMethodSchema` on the input schema, you could use `method: UserEditableMethodFields`, and then create `UserEditableMethodFields` schema as a subset of `FulfillmentMethodSchema`. That way, after you have called `inputSchema.clean(input)` above, it will be impossible for users to overwrite any fields that they should not. Which means that it would be safe to do:

```js
const updatedMethod = { ...currentFulfillmentMethod, ...method, updatedAt: new Date() };
```

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentMethod.js
```diff
@@ -0,0 +1,68 @@
...
+    "methods._id": methodId,
+    shopId
+  }, {
+    $set: {
+      "methods.$": updatedMethod
+    }
+  });
+  if (matchedCount === 0) throw new ReactionError("not-found", "Fulfillment type to be updated not found");
+
+  return { group: updatedMethod };
```
aldeed: I think `group` should be `method`

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentType.js
```diff
@@ -0,0 +1,60 @@
...
+ * @param {Boolean} input.enabled - status of ff-type
+ * @param {String} input.label - label of ff-type
+ * @param {String} input.displayMessageType - optional display message for UI
+ * @returns {Promise<Object>} An object with the updated type
+ */
+export default async function updateFulfillmentType(context, input) {
+  const cleanedInput = inputSchema.clean(input); // add default values and such
+  inputSchema.validate(cleanedInput);
+
+  const { fulfillmentGroupId, shopId, name, enabled, label, displayMessageType } = cleanedInput;
```
aldeed: `fulfillmentGroupId` should be `fulfillmentTypeId`

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentType.js
```diff
@@ -0,0 +1,60 @@
...
+  const { collections: { Fulfillment } } = context;
+
+  if (!shopId) throw new ReactionError("invalid-parameter", "Shop ID to be updated not provided");
+  if (!fulfillmentGroupId) throw new ReactionError("invalid-parameter", "FulfillmentType ID to be updated not provided");
+  if (!name) throw new ReactionError("invalid-parameter", "FulfillmentType Name to be updated not provided");
```
aldeed: None of these error throwing lines should be necessary since `inputSchema.validate(cleanedInput)` would have already thrown an error if these were missing.

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentType.js
```diff
@@ -0,0 +1,60 @@
...
+  name: String,
+  enabled: {
+    type: Boolean,
+    defaultValue: true
+  },
+  label: String,
+  displayMessageType: {
+    type: String,
+    optional: true
+  }
```
aldeed, line 22: Usually for an `update*` mutation, all of these would be optional except `fulfillmentGroupId` and `shopId`, and if they are not passed in they would not be updated.

packages/api-plugin-fulfillment/src/resolvers/Mutation/updateFulfillmentMethod.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {String} [args.input.clientMutationId] - An optional string identifying the mutation call
+ * @param {Object} context - an object containing the per-request state
+ * @returns {Promise<Object>} updateFulfillmentMethodPayload
+ */
+export default async function updateFulfillmentMethod(parentResult, { input }, context) {
+  const { shopId: opaqueShopId, clientMutationId = null, fulfillmentTypeId: opaqueFulfillmentTypeId, methodId: opaqueMethodId, method } = input.groupInfo;
+
+  const methodId = decodeFulfillmentMethodOpaqueId(opaqueMethodId);
+  const fulfillmentTypeId = decodeFulfillmentGroupOpaqueId(opaqueFulfillmentTypeId);
+  const shopId = decodeShopOpaqueId(opaqueShopId);
```
aldeed: Since this is a new mutation, you don't need to use opaque IDs and you don't need to accept and return `clientMutationId`

packages/api-plugin-fulfillment/src/resolvers/Mutation/updateFulfillmentType.js
```diff
@@ -0,0 +1,34 @@
...
+ * @param {String} [args.input.clientMutationId] - An optional string identifying the mutation call
+ * @param {Object} context - an object containing the per-request state
+ * @returns {Promise<Object>} updateFulfillmentTypePayload
+ */
+export default async function updateFulfillmentType(parentResult, { input }, context) {
+  const { groupInfo, clientMutationId = null } = input;
+  const { shopId: opaqueShopId, fulfillmentGroupId: opaqueFulfillmentGroupId } = groupInfo;
+
+  const fulfillmentGroupId = decodeFulfillmentGroupOpaqueId(opaqueFulfillmentGroupId);
+  const shopId = decodeShopOpaqueId(opaqueShopId);
```
aldeed: Since this is a new mutation, you don't need to use opaque IDs and you don't need to accept and return `clientMutationId`

packages/api-plugin-fulfillment/src/resolvers/Query/getFulfillmentMethods.js
```diff
@@ -0,0 +1,32 @@
...
+ * @param {String} args.shopId - Shop ID to get records for
+ * @param {Object} context - an object containing the per-request state
+ * @param {Object} info Info about the GraphQL request
+ * @returns {Promise<Object>} Fulfillment methods
+ */
+export default async function getFulfillmentMethods(_, args, context, info) {
+  const { shopId: opaqueShopId, fulfillmentTypeId: opaqueFulfillmentId, ...connectionArgs } = args;
+
+  const shopId = decodeShopOpaqueId(opaqueShopId);
+  const fulfillmentTypeId = decodeFulfillmentGroupOpaqueId(opaqueFulfillmentId);
```
aldeed: Since this is a new query, you don't need to use opaque IDs.

packages/api-plugin-fulfillment/src/resolvers/Query/getFulfillmentType.js
```diff
@@ -0,0 +1,28 @@
...
+ * @returns {Promise<Object>} Fulfillment type
+ */
+export default async function getFulfillmentType(_, args, context) {
+  const {
+    fulfillmentTypeId: opaqueTypeId,
+    shopId: opaqueShopId
+  } = args;
+
+  const fulfillmentTypeId = decodeFulfillmentGroupOpaqueId(opaqueTypeId);
+  const shopId = decodeShopOpaqueId(opaqueShopId);
```
aldeed: Since this is a new query, you don't need to use opaque IDs.

packages/api-plugin-fulfillment/src/resolvers/Query/getFulfillmentTypes.js
```diff
@@ -0,0 +1,31 @@
...
+ * @param {Object} args - an object of all arguments that were sent by the client
+ * @param {String} args.shopId - Shop ID to get records for
+ * @param {Object} context - an object containing the per-request state
+ * @param {Object} info Info about the GraphQL request
+ * @returns {Promise<Object>} Fulfillment types under the Shop
+ */
+export default async function getFulfillmentTypes(_, args, context, info) {
+  const { shopId: opaqueShopId, ...connectionArgs } = args;
+
+  const shopId = decodeShopOpaqueId(opaqueShopId);
```
aldeed: Since this is a new query, you don't need to use opaque IDs.

aldeed (changes requested):
I left quite a few inline comments, but the overall plan seems good I think. At one point there was a decision and style guide entry that all queries should _not_ start with the "get" prefix (which I think is also what GraphQL folks recommend). If you're still following that, then all of the queries should be renamed to remove `get` prefix, both internal queries and GraphQL resolvers/schema.

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.test.js
sujithvn: As we are not including new Jest specific entries, did a comparison against locally declared variable.

packages/api-plugin-fulfillment/src/resolvers/Query/getFulfillmentTypes.js
sujithvn: Removed opaqueShopId, now directly using shopId

packages/api-plugin-fulfillment/src/resolvers/Query/getFulfillmentType.js
sujithvn: Removed opaqueShopId, now directly using shopId
Removed opaqueTypeId, now directly using fulfillmentTypeId

packages/api-plugin-fulfillment/src/resolvers/Query/getFulfillmentMethods.js
sujithvn: Removed opaqueShopId, now directly using shopId
Removed opaqueTypeId, now directly using fulfillmentTypeId

packages/api-plugin-fulfillment/src/resolvers/Mutation/updateFulfillmentType.js
sujithvn: Removed opaqueShopId, now directly using shopId
Removed opaqueFulfillmentGroupId, now directly using fulfillmentGroupId
Removed clientMutationId

packages/api-plugin-fulfillment/src/resolvers/Mutation/updateFulfillmentMethod.js
sujithvn: Removed opaqueShopId, now directly using shopId
Removed opaqueFulfillmentTypeId, now directly using fulfillmentTypeId
Removed opaqueMethodId, now directly using methodId
Removed clientMutationId

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentType.js
sujithvn: Removed

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.js
sujithvn: Updated to use `.find()`

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.js
sujithvn: Yes, I have updated

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.js
sujithvn: Added

packages/api-plugin-fulfillment/src/index.js
sujithvn: I vaguely remember there was an initial discussion with @zenweasel on this where I changed the name from Fulfillments to Fulfillment.

What I remember is that a collection which had the FulfillmentTypes and also its FulfillmentMethods, was called Fulfillment. Not sure if that is a strong enough reason. 

If we have to make this change, I guess we have to update most of the mutations/queries/resolvers.

packages/api-plugin-fulfillment/src/mutations/createFulfillmentType.js
sujithvn: -updated-

@aldeed 
Thanks for pointing out the issue.

The implementation was expected to add ff-method also (if details provided) along with adding a new ff-type. But as you pointed out, it had a bug and would not have worked s expected. 

Now after thinking about this for sometime, I have made the following changes.

I removed the `method` property from the `FulfillmentType` input. Now this mutation will only serve the purpose of creating the fulfillment-type and will not add any ff-method under it. For adding ff-methods, we should be using the mutation `createFulfillmentMethod` (as it makes sense to add a new ff-method only by a new plugin implementation).

The problem that  I see here is that the user may be expecting to add the ff-method also by adding it into the input data (since the FulfillmentTypeSchema allows it as optional), but we are deleting it and adding only ff-type.

@zenweasel - please let me know if the changes are okay.

packages/api-plugin-fulfillment/src/schemas/schema.graphql
```diff
@@ -0,0 +1,414 @@
+"Allowed fulfillment types, extended by fulfillment-type plugins"
+enum FulfillmentType {
+  "Default fulfillment type when none is decided by user"
+  undecided
+}
+
+"Default empty object for additionalData"
+type emptyData {
```
tedraykov: We are using PascalCase for all types and inputs in the GraphQL schema files across the project. I think we should do too here.

packages/api-plugin-fulfillment/src/simpleSchemas.js
sujithvn: @aldeed @zenweasel 
I have made the changes to SimpleSchema to add the createdAt & updatedAt for the ff-type at root level of the doc.
I have not added in GraphQL since the createFFType mutation is not public and the updateFFType we populate the updatedAt in the code instead of user-input

Question:
We have `methods` array under each ff-type. Do we need to introduce the createdAt & updatedAt for each ff-method as well?

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentMethod.js
sujithvn: Implemented except for updatedAt field.
Awaiting response on the comment [above](https://github.com/reactioncommerce/reaction/pull/6610#discussion_r1018187273)

packages/api-plugin-fulfillment/src/schemas/schema.graphql
sujithvn: Updated

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentType.js
sujithvn: updated

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentMethod.js
sujithvn: Updated in mutation & resolver

tedraykov:
Sujith, you said you have addressed the comments, but you haven't pushed the changes to the branch.

packages/api-plugin-fulfillment/src/preStartup.js
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * @summary Called before startup to extend schemas
+ * @param {Object} context Startup context
+ * @returns {undefined}
+ */
+export default async function fulfillmentPreStartup(context) {
+  const allFulfillmentTypesArray = context.fulfillment?.registeredFulfillmentTypes;
+
+  if (!allFulfillmentTypesArray || allFulfillmentTypesArray.length === 0) {
```
tedraykov: This will never be true because this plugin defined the `undecided` fulfillmentType so no need for the check.

tedraykov (changes requested):
Looks good apart from the PascalCase comment. I have some comments regarding the simple schema extensions, but the 4th PR with the affected plugins is a more suitable place for that.

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentType.js
sujithvn: Updated to make all fields except `fulfillmentTypeId` and `shopId` optional.
The code will check for each optional field if there is input data and then accordingly create the `updateObject` to be passed to `updateOne`

packages/api-plugin-fulfillment/src/preStartup.js
sujithvn: Thank you. Removed.
Will be pushing the changes soon

sujithvn:
> 

@aldeed 
Reg: remove 'get' prefix for queries
Yes, I remember @zenweasel  had mentioned this to me once, but I missed. I shall update the newly introduced queries but `getFulfillmentMethodsWithQuotes` is an existing query from old shipment plugin and I guess I should be retaining it as such with the get prefix.

brent-hoover:
You should create the query under the new compliant name and create an alias for the old with a depreciation marker

packages/api-plugin-fulfillment/src/index.js
brent-hoover: I don't have a strong opinion here, I think the relevant decider is "element of least surprise" which I _think_ might lean towards `Fulfillment` since it covers everything about Fulfillment, not just fulfillment types similar to the way "Shipping" did

packages/api-plugin-fulfillment/src/mutations/createFulfillmentType.js
brent-hoover: The description of the changes makes sense to me. I have not reviewed the code.

packages/api-plugin-fulfillment/src/simpleSchemas.js
brent-hoover: > We have methods array under each ff-type. Do we need to introduce the createdAt & updatedAt for each ff-method as well?

No. These fields are at the collection level only

packages/api-plugin-fulfillment/src/simpleSchemas.js
sujithvn: Hi @aldeed 
Generally I think the field name `baseType` inside the collection should be fine.

But considering that we have the collection as `Fulfillment` with the 2 major components as `fulfillmentType` and `fulfillmentMethod` with a logical association (in the names), I feel like `baseType` could create a slight confusion.

Also, this fix would require changes in many files across multiple plugins which increases the probability of breaking something.

This is just what I thought about it (not a very strong opinion) and please let me know if you feel strongly about `baseType` and I can make the changes

packages/api-plugin-fulfillment/src/mutations/createFulfillmentType.js
sujithvn: I would be pushing the code shortly along with all the other changes. I just wanted a common understanding on the way we deal with user provided input.

packages/api-plugin-fulfillment/src/simpleSchemas.js
sujithvn: @zenweasel Thanks for clarifying.

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentMethod.js
sujithvn: As confirmed by Brent in the above comment, the updatedAt field is only at doc root and not required for each element of the array. So the change is complete. Code will be pushed soon

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentType.js
```diff
@@ -0,0 +1,60 @@
...
+      name,
+      "provider.enabled": enabled,
+      "provider.name": name,
+      "provider.label": label,
+      displayMessageType
+    }
+  });
+  if (matchedCount === 0) throw new ReactionError("not-found", "Fulfillment type to update not found");
+
+  return { group: cleanedInput };
```
tedraykov: Related to Eric's comment but `group` has been mentioned multiple times in the updateFilfillmentType mutation. In the input schema, the resolver, as well as the graphql schema definition. Rename it to `fulfillmentType` everywhere.

packages/api-plugin-fulfillment/src/schemas/schema.graphql
```diff
@@ -0,0 +1,414 @@
...
+  rate: Float!
+}
+
+"Input needed to select a fulfillment option for a single fulfillment group on a cart"
+input updateFulfillmentTypeInputInfo {
+  "The shop to which this group belongs"
+  shopId: ID!
+
+  "The group to update"
+  fulfillmentGroupId: ID!
```
tedraykov: Am i getting things wrong or this should be `fulfillmentTypeId` rather than `group`. If so, also address the comments of the input.

tedraykov:
Could we define indexes on the `Fulfillment` collection. From what I found, the attributes we are searching by are `shopId`, `fulfillmentType`, `provider.enabled`, and `methods._id`.

packages/api-plugin-fulfillment/src/xforms/id.js
```diff
@@ -0,0 +1,18 @@
...
+  Cart: "reaction/cart",
+  FulfillmentGroup: "reaction/fulfillmentGroup",
+  FulfillmentMethod: "reaction/fulfillmentMethod"
+};
+
+export const encodeShopOpaqueId = encodeOpaqueId(namespaces.Shop);
+export const encodeFulfillmentMethodOpaqueId = encodeOpaqueId(namespaces.FulfillmentMethod);
+export const encodeFulfillmentGroupOpaqueId = encodeOpaqueId(namespaces.FulfillmentGroup);
+
+export const decodeShopOpaqueId = decodeOpaqueIdForNamespace(namespaces.Shop);
```
vanpho93, line 15: don't know if we still need these decode functions?

packages/api-plugin-fulfillment/src/startup.js
```diff
@@ -0,0 +1,29 @@
...
+        shopId,
+        provider: {
+          enabled: true,
+          label: "Undecided",
+          name: "undecided"
+        },
+        fulfillmentType: "undecided"
+      };
+      await context.mutations.createFulfillmentType(context.getInternalContext(), groupInfo);
+    }
```
vanpho93: Should returns early so it makes it easier to read.

ref: https://gomakethings.com/the-early-return-pattern-in-javascript/

```suggestion
    if (undecidedRecord) return;
    const groupInfo = {
      name: "Undecided Group",
      shopId,
      provider: {
        enabled: true,
        label: "Undecided",
        name: "undecided"
      },
      fulfillmentType: "undecided"
    };
    await context.mutations.createFulfillmentType(context.getInternalContext(), groupInfo);
```

packages/api-plugin-fulfillment/src/queries/fulfillmentMethodsWithQuotes.js
```diff
@@ -0,0 +1,51 @@
...
+  const commonOrderExtended = await extendCommonOrder(context, commonOrder);
+
+  const fulfillmentTypeInGroup = commonOrder.fulfillmentType;
+  if (!fulfillmentTypeInGroup) throw new ReactionError("not-found", "Fulfillment type not found in commonOrder");
+
+  const allFuncsArray = context.getFunctionsOfType("fulfillmentMethodsWithQuotes");
+  const ffTypeFuncObjects = allFuncsArray.filter((func) => fulfillmentTypeInGroup === func.key);
+  const funcs = ffTypeFuncObjects.map((func) => func.handler);
+
+  if (!funcs || !Array.isArray(funcs) || !funcs.length) throw new ReactionError("not-found", `No methods for Fulfillment type ${fulfillmentTypeInGroup}`);
```
vanpho93: Using lodash can shorten this function

```suggestion
  if (_.isEmpty(funcs)) throw new ReactionError("not-found", `No methods for Fulfillment type ${fulfillmentTypeInGroup}`);
```

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.test.js
sujithvn: Marking resolved as changes were pushed to branch

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.js
sujithvn: Marking resolved as changes were pushed to branch

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.js
sujithvn: Marking resolved as changes were pushed to branch

packages/api-plugin-fulfillment/src/mutations/createFulfillmentType.js
sujithvn: Changes are pushed to branch

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentMethod.js
sujithvn: Changes are pushed to branch

packages/api-plugin-fulfillment/src/xforms/id.js
sujithvn: @vanpho93  @zenweasel 
I have removed the opaqueId stuff from the new endpoints introduced by this plugin, but retained them for the old endpoints like `updateFulfillmentOptionsForGroup` & `selectFulfillmentOptionForGroup` thinking it may be required by the current storeFront implementations. The above mentioned xforms/id.js is used by those old endpoints.
Should we be removing it from them also?

packages/api-plugin-fulfillment/src/startup.js
sujithvn: Thank you. Updated. Will be pushed soon

packages/api-plugin-fulfillment/src/queries/fulfillmentMethodsWithQuotes.js
sujithvn: Thank you. Updated. Will be pushed soon

packages/api-plugin-fulfillment/src/queries/fulfillmentMethodsWithQuotes.js
```diff
@@ -0,0 +1,51 @@
...
+ * @method
+ * @summary Just gets rates, without updating anything
+ * @param {Object} commonOrder - details about the purchase a user wants to make.
+ * @param {Object} context - Context
+ * @returns {Array} return updated rates in cart
+ * @private
+ */
+export default async function fulfillmentMethodsWithQuotes(commonOrder, context) {
+  const rates = [];
+  const retrialTargets = [];
```
vanpho93, line 19: Rates should be immutable. Each func currently push new things into `rates` and `retrialTargets`.
This can be improve by do sth like this

```javascript
const allRates = await Promise.all(funcs.map(async (func) => {
  const firstResult = func().catch(error => null);
  if (firstResult) return result;
  // retry here
  return func().catch(error => null);
});

const rates = _.compact(allRates).flattern();
```

packages/api-plugin-fulfillment/src/schemas/schema.graphql
sujithvn: Updated the schema elements and the comments
'Groups' are now retained only for the specific field which tracks the group (ground/free...) from the old plugin.
Update will be pushed soon

packages/api-plugin-fulfillment/src/mutations/updateFulfillmentType.js
sujithvn: Updated group to fulfillmentType/Method everywhere applicable (basically the new endpoints to create/update fulfillment type/method).
The old plugin endpoints `selectFulfillmentOptionForGroup` and `updateFulfillmentOptionsForGroup` has group retained  since it is dealing with ff-groups in cart.
Update will be pushed soon

packages/api-plugin-fulfillment/src/schemas/schema.graphql
sujithvn: @zenweasel 
**Update**: 
I had made this field as 'required' as per our earlier discussion. But now, I have reverted it back to 'optional'. Reason being, this is part of EDIT/UPDATE input and we are making all editable fields optional. 
Meanwhile, in the create input we have this field as 'required'.

packages/api-plugin-fulfillment/src/xforms/id.js
brent-hoover: No, keep them for now but pleases mark the code as deprecated

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.js
```diff
@@ -0,0 +1,62 @@
...
+  const cleanedInput = inputSchema.clean(input); // add default values and such
+  inputSchema.validate(cleanedInput);
+
+  const { method: inputMethod, fulfillmentTypeId, shopId } = cleanedInput;
+  const { collections: { Fulfillment } } = context;
+  const method = { ...inputMethod };
+
+  await context.validatePermissions("reaction:legacy:fulfillmentMethods", "create", { shopId });
+
+  const ffTypeRecord = await Fulfillment.findOne({ _id: fulfillmentTypeId, shopId });
```
vanpho93: We don't have consistent naming here.

I think it makes sense to change to sth like

```
const fulfillmentType = await FulfillmentType.findOne({ _id: fulfillmentTypeId, shopId });
```

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.js
```diff
@@ -0,0 +1,62 @@
...
+  if (!ffTypeRecord) throw new ReactionError("server-error", "Unable to create fulfillment method without defined type");
+
+  let ffTypeMethodRecord;
+  if (ffTypeRecord.methods && Array.isArray(ffTypeRecord.methods)) {
+    ffTypeMethodRecord = ffTypeRecord.methods.find((currMethod) => currMethod.fulfillmentMethod === method.fulfillmentMethod);
+  }
```
vanpho93:
```suggestion
  const ffTypeMethodRecord = ffTypeRecord.methods?.find((currMethod) => currMethod.fulfillmentMethod === method.fulfillmentMethod);
  }
```

aldeed (approved):

packages/api-plugin-fulfillment/src/xforms/id.js
sujithvn: Added the check for opaqueId before calling decode
This update is dependent on #6744

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.js
sujithvn: Updated

packages/api-plugin-fulfillment/src/mutations/createFulfillmentMethod.js
sujithvn: updated

packages/api-plugin-fulfillment/src/queries/fulfillmentMethodsWithQuotes.js
sujithvn: This will be updated after the initial merge to feat/fulfillment-type branch
Issue for tracking [here](https://app.zenhub.com/workspaces/moc-main-board-62fd8e881686130027eafc31/issues/gh/reactioncommerce/reaction/6821)

PR #6577, opened 2022-10-19
01 02 fulfillment type pickup

sujithvn:
# This PR is replaced by #6613 

Signed-off-by: Sujith mail.sujithvn@gmail.com

Resolves #6471 
Impact: **breaking**
Type: **feature**

## Issue

In the current system, the only way for order fulfillment is via shipping which is hardcoded into the system. We need to introduce the flexibility for users to add their own fulfillment types via plugins.

## Solution

We are creating a base fulfillment plugin which would enable other fulfillment types (like shipping, pickup, digital) to be introduced via plugins. Each of the newly introduced fulfillment type plugin would need to have the specific fulfillment methods also to be added as separate plugins. Example, fulfillment type 'pickup' could have fulfillment methods like 'store pickup' and 'curb-side pickup'.

This PR is 3rd entry for 01-02 fulfillment-type-pickup

## PR Details in order

```
feat/fulfillment-types
 | 
 | __ 01 fulfillment base #6570 
 |        | __ 01-01 fulfillment-type-shipping       #6579
 |        | __ 01-02 fulfillment-type-pickup.       <--- This PR #6577 
 |__02 fulfillment-type-impacted-plugins. [carts, products, orders, catalogs] #6582 
 |__03 placeOrder refactor #6583 

```
## Existing PRs based on the #6480 (to be updated to point to #6570) 
We shall change the base branch of the 3 existing PRs (i18n #6545, dataMigration #6544, setDefaultFFtype #6543) to  point to 01 fulfillment base

### API-PLUGIN-FULFILLMENT-TYPE-PICKUP
* Registers the 'pickup' as a ff-type via registeredFulfillmentTypes: ["pickup"]
* Inserts the default entry for pickup ff-type in Fulfillment collection

### API-PLUGIN-FULFILLMENT-METHOD-PICKUP-STORE
*  getFulfillmentMethodsWithQuotesPickupStore - returns the quote or equivalent details for the method when called from base ff plugin
* preStartup - extends the union of "methodAdditionalData" with data structure specific to Store
* startup - Inserts the required ff-method entry into Fulfillment collection
* util/checkAndCreateFulfillmentMethod - confirms existing ff-type entry and adds a new ff-method under it.
* util/collectStoreDetails - dummy function to simulate api providing Store specific info while returning quotes.
* util/ validateOrderMethodsstore - dummy function to simulate Store specific validations done. Called by prepareOrder.js 

## Breaking changes

[...]

PR #6613, opened 2022-11-03
New fulfillment type pickup

sujithvn:

### This PR replaces #6577 

Resolves #6471 
Impact: **breaking**
Type: **feature**

## Issue

## Solution

This PR is 3rd entry for new-fulfillment-type-pickup

## PR Details in order

```
feat/fulfillment-types
 | 
 |        | __ new-fulfillment-type-shipping       #6614 
 |        | __ new-fulfillment-type-pickup.       <--- This PR #6613
 |__ new-fulfillment-impacted-plugins. [carts, products, orders, catalogs]   #6615 
 |__ new-placeOrder refactor #6616 

```
We shall change the base branch of the 3 existing PRs (i18n #6545, dataMigration #6633 , setDefaultFFtype #6543) to  point to 00 fulfillment base

### API-PLUGIN-FULFILLMENT-TYPE-PICKUP
* Registers the 'pickup' as a ff-type via registeredFulfillmentTypes: ["pickup"]
* Inserts the default entry for pickup ff-type in Fulfillment collection

### API-PLUGIN-FULFILLMENT-METHOD-PICKUP-STORE
*  getFulfillmentMethodsWithQuotesPickupStore - returns the quote or equivalent details for the method when called from base ff plugin
* preStartup - extends the union of "methodAdditionalData" with data structure specific to Store
* util/checkAndCreateFulfillmentMethod - confirms existing ff-type entry and adds a new ff-method under it.
* util/collectStoreDetails - dummy function to simulate api providing Store specific info while returning quotes.
* util/ validateOrderMethodsstore - dummy function to simulate Store specific validations done. Called by prepareOrder.js 

## Breaking changes

## Testing


packages/api-plugin-fulfillment-method-pickup-store/src/getFulfillmentMethodsWithQuotesPickupStore.js
```diff
@@ -0,0 +1,75 @@
...
+  const currentMethodInfo = { packageName };
+
+  if (retrialTargets.length > 0) {
+    const isNotAmongFailedRequests = retrialTargets.every((target) => target.packageName !== packageName);
+    if (isNotAmongFailedRequests) {
+      return previousQueryResults;
+    }
+  }
+
+  const pickupDocs = await Fulfillment.find({
```
tedraykov: Why are we searching for multiple fulfillment types here? Based on the `createFulfillmentType` mutation, we are allowed to have only one fulfillment entry in the database with a specific fulfillment type, in this case, the `pickup` type.

tedraykov: The naming `Docs` here is not really descriptive. It was previously used in the shipping-flat-rate plugin, but looks like it was picked because of naming negligence. I think it's better not to adopt this naming from the previous plugin but rather use something more unambiguous like `pickupFulfillmentType` because this is what the variable actually stores.

tedraykov: In that case a `findOne` is better suited.

packages/api-plugin-fulfillment-method-pickup-store/src/util/validateOrderMethodsstore.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @summary Sample dummy function validating the data requirements for this metod
+ * @param {Object} context - Context object
+ * @param {Object} commonOrder - Current order which provide available details to perform validation
+ * @param {Object[]} validationResults - Validation results collected till now
+ * @returns {Object[]} validationResults - with the validation details populated
+ */
+export default function validateOrderMethodsstore(context, commonOrder, validationResults = []) {
```
tedraykov: Capitalize the `store` in the function name, as well as in the file name.
```suggestion
export default function validateOrderMethodsStore(context, commonOrder, validationResults = []) {
```

packages/api-plugin-fulfillment-method-pickup-store/src/getFulfillmentMethodsWithQuotesPickupStore.js
sujithvn: I am not sure if I got your question correct.

In the `Fulfillment` collection we have one entry for each `fulfillmentType`. Each of this entry could have multiple `fulfillmentMethods`. So here, assuming we are looking for 'pickup' type, we are getting the document for 'pickup' from DB and returning the 'rates' for all the methods under 'pickup' (example 'store', 'kerbside' etc)

packages/api-plugin-fulfillment-method-pickup-store/src/startup.js
```diff
@@ -0,0 +1,21 @@
...
+ * @param {Object} context Startup context
+ * @returns {undefined}
+ */
+export default async function fulfillmentMethodPickupStoreStartup(context) {
+  context.appEvents.on("afterShopCreate", async (payload) => {
+    const { shop } = payload;
+    const shopId = shop._id;
+
+    // We do not have validatePermissions in context during this startup stage, hence commenting below
+    // await context.validatePermissions("reaction:legacy:fulfillmentTypes", "read", { shopId });
```
tedraykov: Why do we have this comment left here if we are not going to validate the permissions?

packages/api-plugin-fulfillment-method-pickup-store/src/getFulfillmentMethodsWithQuotesPickupStore.js
sujithvn: Yes, I shall update it

tedraykov (changes requested):

packages/api-plugin-fulfillment-method-pickup-store/src/getFulfillmentMethodsWithQuotesPickupStore.js
tedraykov: That's right, but we are not fetching the fulfillment methods here, but rather the parent fulfillment type. So it's always going to be one result.

tedraykov: The iteration over all fulfillment methods is happening later on line 48.

sujithvn: Yes, I shall change to findOne as you suggested below

packages/api-plugin-fulfillment-method-pickup-store/src/fulfillmentMethodsWithQuotesPickupStore.js
```diff
@@ -0,0 +1,75 @@
...
+ * @param {Object} commonOrder - details about the purchase a user wants to make.
+ * @param {Array} [previousQueryResults] - an array of shipping rates and
+ * info about failed calls to the APIs of some shipping methods providers
+ * @returns {Array} - an array that contains two arrays: the first array will
+ * be an updated list of shipping rates, and the second will contain info for
+ * retrying this specific package if any errors occurred while retrieving the
+ * shipping rates.
+ * @private
+ */
+export default async function fulfillmentMethodsWithQuotesPickupStore(context, commonOrder, previousQueryResults = []) {
```
vanpho93, line 21: One comment from base: https://github.com/reactioncommerce/reaction/pull/6610#discussion_r1022283027

packages/api-plugin-fulfillment-type-pickup/src/checkAndCreateFulfillmentType.js
```diff
@@ -0,0 +1,25 @@
...
+      provider: {
+        enabled: true,
+        label: "Pickup",
+        name: "pickup"
+      },
+      fulfillmentType: "pickup"
+    };
+    await context.mutations.createFulfillmentType(context.getInternalContext(), groupInfo);
+  }
+  return true;
```
vanpho93: short return first makes it looks better

```suggestion
  if (pickupRecord) return true;
  
    const groupInfo = {
      name: "Pickup Provider",
      shopId,
      provider: {
        enabled: true,
        label: "Pickup",
        name: "pickup"
      },
      fulfillmentType: "pickup"
    };
    await context.mutations.createFulfillmentType(context.getInternalContext(), groupInfo);
  return true;
```

packages/api-plugin-fulfillment-method-pickup-store/src/preStartup.js
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  SelectedFulfillmentOption.extend({
+    methodAdditionalData: {
+      type: SelectedFulfillmentOption.getDefinition(
+        "methodAdditionalData",
+        ["type"]
+      ).type[0].type.extend(MethodStoreData)
+    }
+  });
```
aldeed, line 26: Slightly better architecture might be to pass these in the `registerPlugin` object:

```js
{
  fulfillmentProviders: [
    {
      type: "pickup",
      name: "store",
      methodAdditionalDataSchema: MethodStoreData
    }
  ]
}
```

Then use that information in the cart plugin or wherever `ShippingMethod`, etc. schemas are defined.

In general I prefer the pattern of providing information from more specific plugins to less specific plugins, and let the less specific plugins extend their own schemas based on that information. Otherwise the less specific plugin may be confused about how its schemas have been altered.

aldeed (commented):
One pattern suggestion but otherwise LGTM

packages/api-plugin-fulfillment-method-pickup-store/src/preStartup.js
sujithvn: This will be done after the merge of this branch to feat/fulfillment-type
issue for tracking [here](https://app.zenhub.com/workspaces/moc-main-board-62fd8e881686130027eafc31/issues/gh/reactioncommerce/reaction/6822)

packages/api-plugin-fulfillment-method-pickup-store/src/fulfillmentMethodsWithQuotesPickupStore.js
sujithvn: This will be updated after the initial merge to feat/fulfillment-type branch
Issue for tracking [here](https://app.zenhub.com/workspaces/moc-main-board-62fd8e881686130027eafc31/issues/gh/reactioncommerce/reaction/6821)

PR #6614, opened 2022-11-03
New fulfillment type shipping

sujithvn:

Resolves #6471 
Impact: **breaking**
Type: **feature**

## Issue

## Solution

This PR is for 2nd entry  new-fulfillment-type-shipping

## PR Details in order

```
feat/fulfillment-types
 | 
 |        | __ new-fulfillment-type-shipping       <--- This PR #6614
 |__ new-fulfillment-impacted-plugins. [carts, products, orders, catalogs]  #6615 
 |__ new-placeOrder refactor validateOrder #6616 

```
## Existing PRs originally based on the #6480 (to be updated to point to #6610) 
We shall change the base branch of the 3 existing PRs (i18n #6545, dataMigration #6633 , setDefaultFFtype #6543) to  point to 00 fulfillment base

### API-PLUGIN-FULFILLMENT-TYPE-SHIPPING
* Registers the 'shipping' as a ff-type via registeredFulfillmentTypes: ["shipping"]
* Inserts the default entry for shipping ff-type in Fulfillment collection

##### API-PLUGIN-FULFILLMENT-METHOD-SHIPPING-FLAT-RATE
* Mostly copied over from api-plugin-shipment-flat-rate by making certain fixes to adhere to new structure and new DB collections. Fixes are similar to the details provided for the other 2 ff-method plugins below.
* The mutations & queries present were left as it is and it works slightly differently compared to the new mutations/queries that were introduced in api-plugin-fulfillment (which is used by remaining ff-methods)
* Shipping restrictions feature is copied over from existing code. This feature is not added to other ff-methods.

##### API-PLUGIN-FULFILLMENT-METHOD-SHIPPING-DYNAMIC-RATE
*  getFulfillmentMethodsWithQuotesShippingDynamicRate - returns the quote or equivalent details for the method when called from base ff plugin
* preStartup - extends the union of "methodAdditionalData" with data structure specific to UPS
* util/checkAndCreateFulfillmentMethod - confirms existing ff-type entry and adds a new ff-method under it.
* util/calculateUPSRate - dummy function to simulate api providing UPS specific info while returning quotes.
* util/ validateOrderMethodsDynamicRate - dummy function to simulate UPS specific validations done. Called by prepareOrder.js 

## Breaking changes

## Testing


packages/api-plugin-fulfillment-type-shipping/src/startup.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+export default async function fulfillmentTypeShippingStartup(context) {
+  context.appEvents.on("afterShopCreate", async (payload) => {
+    const { shop } = payload;
+    const shopId = shop._id;
+
+    // We do not have validatePermissions in context during this startup stage, hence commenting below
+    // await context.validatePermissions("reaction:legacy:fulfillmentTypes", "read", { shopId });
+
+    const insertSuccess = await checkAndCreateFulfillmentType(context, shopId);
```
tedraykov, line 15: When we create an entity, isn't it implied that we check the prerequisite conditions anyway? The `checkAnd` part of the function name seems a bit overexplicit.

packages/api-plugin-fulfillment-method-shipping-dynamic-rate/src/getFulfillmentMethodsWithQuotesShippingDynamicRate.js
```diff
@@ -0,0 +1,70 @@
...
+    "provider.enabled": true
+  }).toArray();
+  if (!shippingRateDocs || !shippingRateDocs.length) {
+    return [rates, retrialTargets];
+  }
+  const initialNumOfRates = rates.length;
+
+  shippingRateDocs.map(async (doc) => {
+    const carrier = doc.provider.label;
+    const currentPluginMethods = doc.methods.filter((method) => ((method.fulfillmentMethod === (fulfillmentMethodName)) && (method.enabled)));
```
tedraykov: Why is the reasoning behind persisting the fulfillment method for a dynamic method? If the rates for are method are dynamically calculated on every `fulfillmentMethodQuotes` by calling an external API I don't see much point in having a method with predefined rates here. It can be dynamically constructed on every get quotes request. The only reasoning I can think of is i we allow the method to be enabled or not. But in that case we should check that before pushing the newly calculated rate. The same goes of all other method plugins.

packages/api-plugin-fulfillment-method-shipping-dynamic-rate/src/fulfillmentMethodsWithQuotesShippingDynamicRate.js
```diff
@@ -0,0 +1,69 @@
...
+    "shopId": commonOrder.shopId,
+    "fulfillmentType": fulfillmentTypeName,
+    "provider.enabled": true
+  }).toArray();
+  if (!shippingRateDocs || !shippingRateDocs.length) {
+    return [rates, retrialTargets];
+  }
+  const initialNumOfRates = rates.length;
+
+  shippingRateDocs.map(async (doc) => {
```
vanpho93:
```suggestion
  shippingRateDocs.forEach(async (doc) => {
```

packages/api-plugin-fulfillment-type-shipping/src/startup.js
sujithvn: Inside the `checkAndCreateFulfillmentType` we have the call to the original `createFulfillmentType` mutation.

Although the mutation call is explicitly making the call to the one defined in context, I thought I could make this a bit explicit to avoid any quick confusion.

packages/api-plugin-fulfillment-method-shipping-dynamic-rate/src/getFulfillmentMethodsWithQuotesShippingDynamicRate.js
sujithvn: Not sure if I completely understood your point. The idea is that we are showing a placeholder function here which could be replaced by an actual API invoking function to collect the dynamic rate. The placeholder function is a dummy and currently returning some hard-coded values. It is expected to collect the actual dynamic quote/rate each time. Now the call to `fulfillmentMethodQuotes` will be placed to all ff-methods defined under the specific fulfillment-type (like all shipping or all pickup) and this is collected and presented to the user to select one of them as the  preferred method via the `selectFulfillmentOptionForGroup` call.

We can have a quick call if I got your question wrong.

packages/api-plugin-fulfillment-method-shipping-dynamic-rate/src/fulfillmentMethodsWithQuotesShippingDynamicRate.js
sujithvn: Updated, will be pushed soon

aldeed (commented):
Similar comments as in the pickup PR and my comments about "type" vs. "method" vs. "provider" in another PR.

PR #6615, opened 2022-11-03
New fulfillment impacted plugins

sujithvn:

Resolves #6471 
Impact: **breaking**
Type: **feature**

## Issue

## Solution

This PR is 4th entry (new-fulfillment-impacted-plugins)

## PR Details in order

```
feat/fulfillment-types
 | 
 |        | __ new-fulfillment-type-shipping      #6614
 |__ new-fulfillment-impacted-plugins. [carts, products, orders, catalogs]  <--- This PR #6615 
 |__ new-placeOrder refactor validateOrder #6616 

```
We shall change the base branch of the 3 existing PRs (i18n #6545, dataMigration #6544, setDefaultFFtype #6543) to  point to 00 fulfillment base

##### CARTS

##### CATALOGS

##### PRODUCTS

##### ORDERS

##### Integration Tests updated
* Updated the integration tests to match the changes in plugins.

## Breaking changes

## Testing


apps/reaction/package.json
```diff
@@ -20,51 +20,55 @@
     "url": "https://github.com/reactioncommerce/reaction/issues"
   },
   "dependencies": {
-    "@reactioncommerce/api-core": "2.0.1",
```
vanpho93: This seems wrong. New dep versions are downgrades from the old ones?

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,123 @@
...
+ * @summary Figures out from which fulfillment groups the item should be removed
  * @param {Object[]} currentGroups The current cart fulfillment groups array
- * @param {String[]} supportedFulfillmentTypes Array of fulfillment types supported by the item
+ * @param {String} itemId Id of the item to be removed
+ * @param {String} shopId The ID of the shop that owns the item (product)
+ * @returns {Object[]|null} The groups array or null if no viable group
+ */
+function determineRemoveFromGroups(currentGroups, itemId, shopId) {
+  let removeFromGroups = currentGroups.map((group) => {
+    if (group.itemIds && group.itemIds.includes(itemId) && shopId === group.shopId) {
```
vanpho93: This can work too.

```suggestion
    if (group.itemIds?.includes?.(itemId) && shopId === group.shopId) {
```

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,123 @@
...
+      return group;
+    }
+    return null;
+  });
+  removeFromGroups = removeFromGroups.filter((group) => !!group); // remove nulls
+  // If the item is present in more that one Fulfillment group, we remove it from all(for backward compatibility)
+  return removeFromGroups || null;
```
vanpho93: `|| null` is redundant. 

```suggestion
  return _.compact(removeFromGroups);
```

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,123 @@
...
+ * @returns {undefined}
+ */
+function checkAndAddToGroup(currentGroups, fulfillmentType, item) {
+  let removeFromGroups = determineRemoveFromGroups(currentGroups, item._id, item.shopId);
+  if (removeFromGroups && removeFromGroups.length > 0) {
+    removeFromGroups = (removeFromGroups || []).map((group) => {
+      group.itemIds = (group.itemIds || []).filter((itemId) => item._id !== itemId);
+      return group;
+    });
+  }
```
vanpho93: I think there are many redundant null checks.
There is no need to reassign `removeFromGroups`, for each is enough.
```suggestion
  const removeFromGroups = determineRemoveFromGroups(currentGroups, item._id, item.shopId);
removeFromGroups.forEach((group) => {
  group.itemIds = _.pull(group.itemIds, "itemId");
});
```

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,123 @@
...
+      shopId: item.shopId,
+      type: fulfillmentType
+    });
+  } else if (!addToGroup.itemIds) {
+    // If there is a compatible group but it has no items array, add one with just this item in it
+    addToGroup.itemIds = [item._id];
+  } else if (!addToGroup.itemIds.includes(item._id)) {
+    // If there is a compatible group with an items array but it is missing this item, add this item ID to the array
+    addToGroup.itemIds.push(item._id);
+  }
```
vanpho93:
```suggestion
addToGroup.itemIds = _.uniq([...addToGroup.itemIds || [], item._id]);
```

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,123 @@
...
+        supportedFulfillmentTypes = ["shipping"]; // we use 'shipping' as default fulfillment type for old v1 carts (prior to fulfillment types)
+      } else {
+        throw new ReactionError("not-found", "Product does not have any supported FulfillmentTypes");
+      }
     }
 
-    // Out of the current groups, returns the one that this item should be in by default, if it isn't
-    // already in a group
-    const group = determineInitialGroupForItem(currentGroups, supportedFulfillmentTypes, item.shopId);
+    if (selectedFulfillmentType && !supportedFulfillmentTypes.includes(selectedFulfillmentType)) {
```
vanpho93: Should create a separate function to avoid too many if else and long function.

```js
const normalizeSelectedFulfillmentType = (selectedFulfillmentType) {
  if (_.isNil(selectedFulfillmentType)) {
    const hasOnlyOneSupportedFulfillmentType = supportedFulfillmentTypes.length === 1;
    if (hasOnlyOneSupportedFulfillmentType) return _.first(supportedFulfillmentTypes);
   return "undecided";
  }
  if(!supportedFulfillmentTypes.includes(selectedFulfillmentType)) {
    throw new ReactionError("not-found", "Selected fulfillmentType is not supported by the Product");
  }
  return selectedFulfillmentType;
}
```

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,123 @@
...
+      } else {
+        selectedFulfillmentType = "undecided";
+      }
     }
-  });
+    // check if the selectedFulfillmentType is an 'enabled' fulfillmentType, if not set is 'undecided'
+    /* eslint-disable no-await-in-loop */
+    const enabledFulfillmentTypeObjs = await Fulfillment.find({ "shopId": item.shopId, "provider.enabled": true }).toArray();
+    /* eslint-enable no-await-in-loop */
+    let enabledFulfillmentTypes = (enabledFulfillmentTypeObjs || []).map((ffType) => ffType.fulfillmentType);
```
vanpho93: Try to avoid `let`.

```suggestion
    const enabledFulfillmentTypes = enabledFulfillmentTypeObjs.map((ffType) => ffType.fulfillmentType).filter((type) => _.isString(type));
```

packages/api-plugin-carts/src/schemas/cart.graphql
```diff
@@ -340,6 +352,24 @@ input AddCartItemsInput {
...
+  cartToken: String
+
+  "An optional string identifying the mutation call, which will be returned in the response payload"
+  clientMutationId: String
+
+  "Fulfillment Type to be assigned to the items"
+  fulfillmentType: String!
+
+  "Array of item IDs for which selectedFulfillmentType has to be updated"
+  items: [String]!
```
aldeed: `itemIds` would be a clearer name

packages/api-plugin-carts/src/simpleSchemas.js
```diff
@@ -225,6 +226,16 @@ const ShippingMethod = new SimpleSchema({
...
+  "fulfillmentMethod": {
+    type: String,
+    optional: true,
+    label: "Method name identifier Internal"
+  },
+  "displayMessageMethod": {
+    type: String,
+    optional: true,
+    label: "Display message to show in UI for this Method"
+  },
```
aldeed, line 238: I believe `fulfillmentMethod` is sort of a subtype, corresponding with the providing package (e.g. flat rate vs dynamic rate, but both are "shipping" type). The problem is that the project already uses the term "method" for individual methods, so it seems confusing. I think some other ecommerce platforms use `fulfillmentProvider`, and that makes more sense to me. @zenweasel

packages/api-plugin-carts/src/simpleSchemas.js
```diff
@@ -733,6 +744,19 @@ export const CartItem = new SimpleSchema({
     blackbox: true
   },
   "updatedAt": Date,
+  "selectedFulfillmentType": {
+    type: String,
+    allowedValues: ["undecided"], // extended with dynamic values in fulfillment plugin startup
```
aldeed, line 749: Another option would be to have fulfillment plugins include an array in their plugin registration options:

```js
{
  fulfillmentTypes: ["shipping"]
}
```

Then use `registerPluginHandler` in this cart plugin to read all of the `plugin.fulfillmentTypes` and add them all to `allowedValues` in the schema. That way all cart schema changes will happen in the cart plugin, which makes the most sense to me.

Something similar is how payment methods, address validation methods, etc. are registered from individual plugins.

packages/api-plugin-carts/src/util/updateCartFulfillmentGroups.js
```diff
@@ -1,61 +1,123 @@
...
+      if (supportedFulfillmentTypes.length === 1) {
+        ([selectedFulfillmentType] = supportedFulfillmentTypes);
+      } else {
+        selectedFulfillmentType = "undecided";
+      }
     }
-  });
+    // check if the selectedFulfillmentType is an 'enabled' fulfillmentType, if not set is 'undecided'
+    /* eslint-disable no-await-in-loop */
+    const enabledFulfillmentTypeObjs = await Fulfillment.find({ "shopId": item.shopId, "provider.enabled": true }).toArray();
```
aldeed: It seems like you could add `fulfillmentType: selectedFulfillmentType` to the query object and use `findOne` instead of `find().toArray()`. Then you don't need to do any of the array mapping and checking in code and the db operation will be faster.

aldeed (changes requested):
See a few suggestions, but overall this seems on the right track. I am not sure I like the "undecided" group. What if someone registers a fulfillment type plugin with the name "undecided"? If this ability is necessary, you could make the `group.type` field optional and have `null` mean undecided. But it seems like the best approach would be to have a required shop default and use that.

PR #6616, opened 2022-11-03
New place order refactor validate order

sujithvn:

Resolves #6471 
Impact: **breaking**
Type: **feature**

## Issue

There is no way to verify if the order data in the UI is ready/complete and free of errors.

## Solution

Along with the changes, we are creating a new API query end-point validateOrder which provides the option for frontend to validate the correctness and completeness of the order before actually placing the order.

Also, we are refactoring the placeOrder mutation and cutting down the functional layers

This PR is 5th entry (new-placeOrder refactor validateOrder)

## PR Details in order

```
feat/fulfillment-types
 | 
 |        | __ new-fulfillment-type-shipping      #6614
 |__ new-fulfillment-impacted-plugins. [carts, products, orders, catalogs] #6615 
 |__ new-placeOrder refactor validateOrder  <--- This PR. #6616

```
We shall change the base branch of the 3 existing PRs (i18n #6545, dataMigration #6544, setDefaultFFtype #6543) to  point to 00 fulfillment base

### Details (specific to placeOrder refactor and introduction of validateOrder end-point)

## Breaking changes

## Testing


packages/api-plugin-orders/src/schemas/schema.graphql
```diff
@@ -112,6 +112,12 @@ extend type Query {
     "A valid anonymous access token for this order. Required if the order is not linked with an account."
     token: String
   ): [Refund]
+
+  "Validate the order for correctness & completeness before placing the order"
+  validateOrder(
+    "input with order and payment objects similar to the one used for placeOrder"
+    input: validateOrderInput!
+  ): validateOrderPayload!
```
aldeed: `ValidateOrderInput` and `ValidateOrderPayload` should begin with uppercase letters by convention.

packages/api-plugin-orders/src/util/addShipmentMethodToGroup.js
```diff
@@ -42,16 +42,18 @@ export default async function addShipmentMethodToGroup(context, {
...
   const selectedFulfillmentMethod = rates.find((rate) => selectedFulfillmentMethodId === rate.method._id);
   if (!selectedFulfillmentMethod) {
+    const eventData = { field: "selectedFulfillmentMethodId", value: selectedFulfillmentMethodId };
     throw new ReactionError("invalid", "The selected fulfillment method is no longer available." +
-      " Fetch updated fulfillment options and try creating the order again with a valid method.");
+      " Fetch updated fulfillment options and try creating the order again with a valid method.", eventData);
   }
 
-  group.shipmentMethod = {
+  const output = {
```
aldeed, line 56: If this will no longer be mutating the group object, maybe it should be in a new function with a different name, such as `getShipmentMethodForGroup`?

packages/api-plugin-orders/src/util/orderValidators/createPayments.js
```diff
@@ -0,0 +1,110 @@
...
+  accountId,
+  billingAddress,
+  context,
+  currencyCode,
+  email,
+  orderTotal,
+  paymentsInput,
+  shippingAddress,
+  shop,
+  flag
```
aldeed: `mode` might be a more standard name than `flag` for this use case

aldeed: Also consider throwing an error if flag/mode is not an expected value

packages/api-plugin-orders/src/util/orderValidators/getDiscounts.js
```diff
@@ -0,0 +1,19 @@
...
+  let discounts = [];
+  let discountTotal = 0;
+  if (cart) {
+    const discountsResult = await context.queries.getDiscountsTotalForCart(context, cart);
+    if (discountsResult) {
+      ({ discounts } = discountsResult);
+      discountTotal = discountsResult.total;
+    }
+  }
+  return { discounts, discountTotal };
```
aldeed: It seems like the only thing this function does is rename `total` to `discountTotal` and return some defaults. Maybe `getDiscountsTotalForCart` could be modified to return this way, and then `getDiscounts` would not be needed?

packages/api-plugin-orders/src/util/orderValidators/validateInitialOrderData.js
```diff
@@ -0,0 +1,33 @@
...
+  const { collections: { Cart }, userId } = context;
+
+  if (!shopId) throw new ReactionError("invalid-param", "ShopID not found in order data", { field: "ShopId", value: shopId });
+
+  const shop = await context.queries.shopById(context, shopId);
+  if (!shop) throw new ReactionError("not-found", "Shop not found while trying to validate order data", { field: "ShopId", value: shopId });
+
+  let cart;
+  if (cartId) {
+    cart = await Cart.findOne({ _id: cartId });
```
aldeed: Should this use `queries.getCartById` to avoid accessing another plugin's collection?

aldeed (commented):
Aside from the few inline suggestions, `prepareOrder` is a good example of a fn where returning errors rather than throwing them would be a lot cleaner. You wouldn't need to try/catch and formatErrors everywhere. I'm not sure whether you want to tackle that level of refactoring across all those functions now, but it would be good to do eventually.

packages/api-plugin-orders/src/util/orderValidators/prepareOrder.js
```diff
@@ -0,0 +1,316 @@
...
+    inputSchema.validate(cleanedInput);
+  } else {
+    try {
+      inputSchema.validate(cleanedInput);
+    } catch (err) {
+      const validationErrors = formatErrors(err);
+      validationResults.push(...validationErrors);
+      return { errors: validationResults, success: false };
+    }
+  }
```
vanpho93: This way looks a bit better. Same for `validateInitialOrderData` and other steps.

```suggestion
    try {
      inputSchema.validate(cleanedInput);
    } catch (err) {
      if (flag === "createOrderObject") {
        throw err;
      }    
      const validationErrors = formatErrors(err);
      validationResults.push(...validationErrors);
      return { errors: validationResults, success: false };
    }
  }
```

packages/api-plugin-orders/src/util/orderValidators/getCustomFields.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+test("should return transformed fields if there are functions defined", async () => {
+  mockContext.getFunctionsOfType = jest.fn().mockReturnValueOnce([
+    jest.fn().mockName("transformCustomOrderFields").mockReturnValueOnce(Promise.resolve({ customField2: "customValue2" }))
+  ]);
+
+  const orderInput = { orderId: "order123" };
+  const customFieldsFromClient = { customField1: "customValue1" };
+  const result = await getCustomFields(mockContext, customFieldsFromClient, orderInput);
+  expect(result).toEqual({ customField2: "customValue2" });
```
vanpho93: Is that expected behavior. I thought response should be `{ customField2: "customValue2",  customField1: "customValue1" };

With current behavior, new function from `transformCustomOrderFields` will override all the precedent functions?

packages/api-plugin-orders/src/util/orderValidators/getFinalFulfillmentGroups.js
```diff
@@ -0,0 +1,148 @@
...
+      workflow: { status: "new", workflow: ["new"] }
+    };
+
+    // Build the final order item objects. As part of this, we look up the variant in the system and make sure that
+    // the price is what the caller expects it to be.
+    if (items) {
+      group.items = await Promise.all(items.map((inputItem) => buildOrderItem(context, { currencyCode, inputItem, cart })));
+    } else {
+      group.items = [];
+    }
```
vanpho93: This may work and look a bit better?
```suggestion
    group.items = await Promise.all((items || []).map((inputItem) => buildOrderItem(context, { currencyCode, inputItem, cart })));
```

packages/api-plugin-orders/src/util/orderValidators/getFinalFulfillmentGroups.js
```diff
@@ -0,0 +1,148 @@
...
+
+    if (Array.isArray(additionalItems) && additionalItems.length) {
+      group.items.push(...additionalItems);
+    }
```
vanpho93:
```suggestion
  group.items.push(...(additionalItems || []));
```

packages/api-plugin-orders/src/util/orderValidators/prepareOrder.js
```diff
@@ -0,0 +1,316 @@
...
+ * @name prepareOrder
+ * @method
+ * @memberof Order/NoMeteorQueries
+ * @summary Validates if the input order details is valid and ready for order processing
+ * @param {Object} context - an object containing the per-request state
+ * @param {Object} input - order details, refer inputSchema
+ * @param {String} flag - flag which define if the call is from placeOrder or validateOrder
+ * @returns {Promise<Object>} output - order, token and validation results
+ */
+export default async function prepareOrder(context, input, flag) {
```
vanpho93: This function is quite long with many duplications. This should be cleaned up.

PR #6633, opened 2022-11-07
New data migration fulfillment

### LINKED ISSUE #6472: As an admin I want to be able to migrate my data to the new format required for the new FFT plugins

sujithvn:
Resolves #6472 
Impact: **breaking**
Type: **feature**

## Issue

Migrate data to the new format required for the new FFT plugins

## Solution

We have 3 stages in a single migration;
* Copies over the existing `Shipments` & `FlatRateFulfillmentRestrictions` collections to `Fulfillment` & `FulfillmentRestrictions` collections respectively
* Updates the `Fulfillment` collection with the newly introduced fields (`fulfillmentType` & `fulfillmentMethod`). Since the existing user base had only Shipping/Flatrate as the fulfillment type/method till date, we would be using those values to update collection
* update the `Groups` collection with the new permissions

## Breaking changes
The 2nd stage above handles the breaking change where we would be using the possible existing values as the values or newly introduced fields.

## Testing
In an existing installation, perform the data-migration. Update the installation to use the new version having fulfillment types. All activities in admin/shop (related to Products/Orders/FF-types/Catalogs/Carts) should continue to work without issues

To test migration steps, install api-migrations package and include the below entry in the `migrator.config.js` file.

```
export default {
  tracks: [
    {
      namespace: "fulfillment",
      package: "../reaction/packages/api-plugin-fulfillment",
      version: 2
    }
  ]
};

```

Run the command:
`MONGO_URL=mongodb://localhost:27017/reaction npx migrator migrate`


aldeed (approved):

packages/api-plugin-fulfillment/migrations/2.js
```diff
@@ -0,0 +1,177 @@
...
+          "eachMethod.fulfillmentMethod": { $exists: false }
+        }
+      ]
+    }
+  };
+
+  operations.push(ffTypeUpdate);
+  operations.push(ffMethodUpdate);
+
+  const bulkWriteResp = await db.collection(COLL_DEST).bulkWrite(operations);
```
brent-hoover, line 111: This will probably work but it's not really what the bulkWrite API is really for an would probably be better as written as a single query. Bulkwrite is when you are doing a large amount of atomic operations, not one large one.

packages/api-plugin-fulfillment/migrations/2.js
```diff
@@ -0,0 +1,177 @@
...
+        if (currentPerms.includes(oldPerm)) {
+          permsToAdd.push(mapperSetFFMethodsRestricts[oldPerm]);
+          if (mapperSetFFTypes[oldPerm]) {
+            permsToAdd.push(mapperSetFFTypes[oldPerm]);
+          }
+        }
+      });
+    }
+    if (permsToAdd.length) {
+      permsToAdd.push("reaction:legacy:fulfillmentTypes/update:settings"); // add this setting to groups deailing with ff-types
```
brent-hoover, line 160: typo in comment: dealing

brent-hoover (changes requested):
Just suggesting using regular operations rather than bulk operations

vanpho93 (approved):

Issue #6641, opened 2022-11-08
Add promotion type as an attribute of the promotion record


brent-hoover:

brent-hoover:
@brendan-shea-pm should promotion type be a required field?

brent-hoover:
The promotion types defined in the [Kinetic ticket](https://app.zenhub.com/workspaces/moc-main-board-62fd8e881686130027eafc31/issues/reactioncommerce/kinetic/104):

> Cart Discount (Discount: Order)
Item Discount (Discount: Item)
Shipping Promotion (Discount: Shipping)

brent-hoover:
Completed

PR #6644, opened 2022-11-09
Feat/promotions add promotion type

brent-hoover:
Resolves [#6641](https://github.com/reactioncommerce/reaction/issues/6641)
Impact: **minor**
Type: **feature**

## Issue

For stackability purposes and for quick creation purposes we need to have "promotionTypes"

## Solution

We define promotion types and allow plugin authors to add their own promotionTypes which can indicate any combination for triggers and actions for creation purposes

## Breaking changes

None

### DIFF
--- a/packages/api-plugin-promotions-offers/src/triggers/offerTriggerHandler.js
+++ b/packages/api-plugin-promotions-offers/src/triggers/offerTriggerHandler.js
@@ -49,5 +49,6 @@ export async function offerTriggerHandler(context, enhancedCart, { triggerParame
 export default {
   key: "offers",
   handler: offerTriggerHandler,
-  paramSchema: OfferTriggerParameters
+  paramSchema: OfferTriggerParameters,
+  triggerType: "implicit"
 };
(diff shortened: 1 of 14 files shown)


vanpho93 (approved):

packages/api-plugin-promotions-offers/src/triggers/offerTriggerHandler.js
```diff
@@ -49,5 +49,6 @@ export async function offerTriggerHandler(context, enhancedCart, { triggerParame
 export default {
   key: "offers",
   handler: offerTriggerHandler,
-  paramSchema: OfferTriggerParameters
+  paramSchema: OfferTriggerParameters,
+  triggerType: "implicit"
```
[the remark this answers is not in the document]
zenweasel: I strongly disagree. It's really a type at all, plus using "type" always paints you into a corner when you another type (like we do here).

That said, I forgot to change the name in the queries for records

brent-hoover: Disregard that last comment, I thought it was on a different bit of code

brent-hoover: Did not mean to delete your comment. Will make that change

PR #6645, opened 2022-11-10
feat/promotions create auto increment function for generating auto-increment Ids

brent-hoover:
Resolves https://github.com/reactioncommerce/kinetic/issues/107
Impact: **minor**
Type: **feature**

## Issue
Unlike most SQL databases Mongo does not have a built in function for creating auto-incrementing IDs

## Solution

Using `findOneAndUpdate` and `$inc` is a fully atomic operation that is not subject to race-conditions and thus works as an equivalent. Please read that again for commenting, nerds.

## Breaking changes

None


packages/api-plugin-sequences/src/simpleSchemas.js
```diff
@@ -0,0 +1,7 @@
+import SimpleSchema from "simpl-schema";
+
+export const Sequence = new SimpleSchema({
+  shopId: String,
+  entity: String,
+  value: SimpleSchema.Integer
```
aldeed: Consider adding `min: 0` validation

aldeed (commented):
A couple general comments:
- A couple Jest tests might be nice to prove it works
- Readme could show a usage example
- There is not much code here. Should this be built into core and available always on context similar to `getAbsoluteUrl`?

packages/api-plugin-sequences/src/simpleSchemas.js
brent-hoover: @aldeed Could not think of what to write tests for since it relies to heavily on the db which would need to mocked

packages/api-plugin-promotions/src/index.js
```diff
@@ -56,6 +57,11 @@ export default async function register(app) {
       qualifiers,
       promotionTypes
     },
+    Sequences: [
```
vanpho93: I think using key makes it more flexible, we may need to create id for sub-element in the future? also each key can have custom startIndex

`{ key: "Promotions", startIndex: 1000 }`

packages/api-plugin-sequences/README.md
```diff
@@ -0,0 +1,66 @@
+# api-plugin-settings
```
vanpho93:
```suggestion
# api-plugin-sequences
```

packages/api-plugin-sequences/src/mutations/incrementSequence.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * @summary returns an auto-incrementing integer id for a specific entity
+ * @param {Object} context - The application context
+ * @param {String} shopId - The shop ID
+ * @param {String} entity - The entity (normally a collection) that you are tracking the ID for
+ * @return {Promise<Number>} - The auto-incrementing ID to use
+ */
+export default async function incrementSequence(context, shopId, entity) {
+  const { collections: { Sequences } } = context;
+  const { value: { value } } = await Sequences.findOneAndUpdate(
```
vanpho93, line 10:
```suggestion
  const { sequence: { value } } = await Sequences.findOneAndUpdate(
```

packages/api-plugin-sequences/src/registration.js
```diff
@@ -0,0 +1,12 @@
+export const Sequences = [];
```
vanpho93: Optional naming suggestion. `Sequences` looks like a collection name to me.

```suggestion
export const sequenceConfigs = [];
```

packages/api-plugin-sequences/src/startup.js
```diff
@@ -0,0 +1,43 @@
...
+import config from "./config.js";
+
+const { SEQUENCE_INITIAL_VALUES } = config;
+
+/**
+ * @summary create new sequences if necessary
+ * @param {Object} context - The application context
+ * @return {Promise<void>} undefined
+ */
+export default async function startupSequences(context) {
```
vanpho93, line 12: How about remove this and add check into `incrementSequence`. ex: if findOneAndUpdate returns null then create sequence? It seems to be a problem if we create new shop?

brent-hoover: No, that is still susceptible to a race condition which is precisely why I didn't do that

packages/api-plugin-promotions/src/index.js
brent-hoover: That's why I called it entity, and start indexes are defined through env vars rather than hard coding it here

packages/api-plugin-sequences/src/registration.js
brent-hoover: Changed

packages/api-plugin-sequences/src/mutations/incrementSequence.js
brent-hoover: I don't think this works? Can you explain this change?

vanpho93: My bad.

packages/api-plugin-sequences/README.md
brent-hoover: fixed

packages/api-plugin-promotions/src/index.js
```diff
@@ -32,10 +32,11 @@ export default async function register(app) {
       Promotions: {
         name: "Promotions",
         indexes: [
-          [{ shopId: 1, type: 1, enable: 1, startDate: 1, endDate: 1 }, { name: "c2__shopId__type__enable__startDate_endDate" }],
+          [{ shopId: 1, type: 1, enable: 1, startDate: 1, endDate: 1 }, { name: "shopId__type__enable__startDate_endDate" }],
```
vanpho93: This was my fault but can help change it to `enabled`.

```suggestion
          [{ shopId: 1, type: 1, enabled: 1, startDate: 1, endDate: 1 }, { name: "shopId__type__enabled__startDate_endDate" }],
```

vanpho93 (approved):

PR #6656, opened 2022-11-16
feat: add promotion date before/after filter

vannguyenn:
## Issue
Add support filtering promotions startDate and endDate with a before/after day. This is helpful when we want to know active/upcoming/past Promotions.

### DIFF
--- a/packages/api-plugin-promotions/src/queries/promotions.js
+++ b/packages/api-plugin-promotions/src/queries/promotions.js
@@ -6,9 +6,7 @@
  * @return {Promise<Promotions>} - A list of promotions
  */
 export default async function promotions(context, shopId, filter) {
-  const {
-    collections: { Promotions }
-  } = context;
+  const { collections: { Promotions } } = context;
 
   const selector = { shopId };
 
--- a/packages/api-plugin-promotions/src/queries/promotions.js
+++ b/packages/api-plugin-promotions/src/queries/promotions.js
@@ -6,14 +6,40 @@
  * @return {Promise<Promotions>} - A list of promotions
  */
 export default async function promotions(context, shopId, filter) {
-  const { enabled } = filter;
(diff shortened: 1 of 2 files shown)


packages/api-plugin-promotions/src/queries/promotions.js
```diff
@@ -6,14 +6,40 @@
  * @return {Promise<Promotions>} - A list of promotions
  */
 export default async function promotions(context, shopId, filter) {
-  const { enabled } = filter;
-  const { collections: { Promotions } } = context;
+  const {
+    collections: { Promotions }
```
brent-hoover: This should be one line like it is everywhere else we do this.

packages/api-plugin-promotions/src/schemas/schema.graphql
```diff
@@ -120,15 +119,24 @@ type PromotionConnection {
   totalCount: Int!
 }
 
+input PromotionDateOperators {
+  "The value must be equal to the given value"
+  eq: Date
+
+  "The value must be greater than the given value"
```
brent-hoover: Shouldn't this be less than?

packages/api-plugin-promotions/src/schemas/schema.graphql
```diff
@@ -228,52 +235,43 @@ input PromotionQueryInput {
 }
 
 extend type Mutation {
-  createPromotion(
-    input: PromotionCreateInput
-  ): PromotionUpdateCreatePayload
+  createPromotion(input: PromotionCreateInput): PromotionUpdateCreatePayload
```
brent-hoover: To keep this consistent this should stay with the multiline format

brent-hoover (changes requested):
Just a couple of style nitpicks and an incorrect description

PR #6662, opened 2022-11-17
feat: add state field and archivePromotion endpoint

### LINKED ISSUE #6655: Promotion records should have a "state" field that relates to the current status of the promotion
The currently defined states would:

created - The default status on creation
active - When the current promotion is active
completed - When the promotion has completed it's run but has not been archived
archived - When the user wishes to be able to "hide" the promotion, they can place it in the archived state

brent-hoover:
Resolves #6655
Impact: **minor**
Type: **feature**

Adds the "state" field and the `archivePromotion` endpoint. Also includes a "watcher" that marks records either "active" or "completed" so that shop TZ specific math can be done here. Also a utility to get times based on shop timezone with in-memory caching to avoid constantly hitting the Shops collection.


packages/api-plugin-promotions/src/resolvers/Mutation/archivePromotion.js
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @method archivePromotion
+ * @summary Mark a promotion as archived
+ * @param {Object} _ - unused
+ * @param {Object} args - The input arguments
+ * @param {Object} args.input - the promotionId of the promotion to archive
+ * @param {Object} context - an object containing the per-request state
+ * @return {Promise<Object>} archiveProduct payload
+ */
+export default async function updatePromotion(_, { input }, context) {
```
vanpho93: function name or filename is incorrect?

packages/api-plugin-promotions/src/resolvers/Mutation/duplicatePromotion.js
```diff
@@ -8,6 +8,6 @@
 export default async function duplicatePromotion(_, { input }, context) {
   const { promotionId, shopId } = input;
   await context.validatePermissions("reaction:legacy:promotions", "create", { shopId });
-  const duplicatePromotionResults = await context.mutations.duplicatePromotion(context, promotionId);
+  const duplicatePromotionResults = await context.mutations.duplicatePromotion(context, { shopId, promotionId });
```
vanpho93, line 11: Should update on `duplicatePromotion` mutation file too.

packages/api-plugin-promotions/src/watchers/setPromotionState.js
```diff
@@ -0,0 +1,75 @@
...
+async function markCompleted(context) {
+  const { collections: { Promotions } } = context;
+  const shopTimes = await getCurrentShopTime(context);
+  let totalUpdated = 0;
+  for (const shop of Object.keys(shopTimes)) {
+    const shopTime = shopTimes[shop];
+    // eslint-disable-next-line no-await-in-loop
+    const { modifiedCount } = await Promotions.updateMany({
+      shopId: shop,
+      state: "active",
```
vanpho93, line 57: is this state correct?

packages/api-plugin-promotions/src/watchers/setPromotionState.js
```diff
@@ -0,0 +1,75 @@
...
+async function markActive(context) {
+  const { collections: { Promotions } } = context;
+  const shopTimes = await getCurrentShopTime(context);
+  let totalUpdated = 0;
+  for (const shop of Object.keys(shopTimes)) {
+    const shopTime = shopTimes[shop];
+    // eslint-disable-next-line no-await-in-loop
+    const { modifiedCount } = await Promotions.updateMany({
+      shopId: shop,
+      state: "created",
```
vanpho93, line 30: is this state correct?

packages/api-plugin-promotions/src/watchers/setPromotionState.js
brent-hoover: Yes, why do you think it's not?

packages/api-plugin-promotions/src/watchers/setPromotionState.js
brent-hoover: I think you might be looking at the filter part of the statement and thinking it's the $set part?

vanpho93 (approved):

PR #6664, opened 2022-11-18
feat: upgrade apollo to v3

### LINKED ISSUE #6659: Upgrade to Apollo 3
We are actually upgrading to Apollo 4, but their docs suggest going to 3 first before going to 4. So this ticket covers that first step

vanpho93:
Resolves #6659
Impact: **major**
Type: **feature**

## Issue

Current Apollo version does not support GraphQL subscription

## Solution

Upgrade to Apollo V3

### DIFF
--- a/packages/api-core/src/ReactionAPICore.js
+++ b/packages/api-core/src/ReactionAPICore.js
@@ -1,4 +1,3 @@
-import { createServer } from "http";
 import { createRequire } from "module";
 import diehard from "diehard";
 import express from "express";
@@ -19,7 +18,8 @@ import importPluginsJSONFile from "./importPluginsJSONFile.js";
 import coreResolvers from "./graphql/resolvers/index.js";
 
 const require = createRequire(import.meta.url); // eslint-disable-line
-const { PubSub } = require("apollo-server");
+const { PubSub } = require("graphql-subscriptions");
+const { RedisPubSub } = require("graphql-redis-subscriptions");
 
 const coreGraphQLSchema = importAsString("./graphql/schema.graphql");
 const coreGraphQLSubscriptionSchema = importAsString("./graphql/subscription.graphql");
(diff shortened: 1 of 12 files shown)


packages/api-core/src/createApolloServer.js
```diff
@@ -161,11 +185,13 @@ export default function createApolloServer(options = {}) {
     app.handle(req, res);
   });
 
+  await apolloServer.start();
   apolloServer.applyMiddleware({ app, cors: true, path });
 
   return {
     apolloServer,
     expressApp: app,
+    appHttpServer,
```
brent-hoover: Can you help me understand how appHttpServer differs from expressApp?

vanpho93: `const httpServer = require("http").createServer();`

so httpServer is from nodejs default package `http`. We use `httpServer` to init `wsServer` which will be used to init Apollo server

```
const wsServer = new WebSocketServer({ server: httpServer, path: DEFAULT_GRAPHQL_PATH });
useServer({ schema, context: { ...contextFromOptions } }, wsServer);
```
Just change some variable names.

packages/api-core/src/createApolloServer.js
```diff
@@ -84,6 +92,10 @@ export default function createApolloServer(options = {}) {
     subscriptions = {
       path
     };
+
+    const wsServer = new WebSocketServer({ server: httpServer, path: DEFAULT_GRAPHQL_PATH });
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    wsServerCleanup = useServer({ schema, context: { ...contextFromOptions } }, wsServer);
```
aldeed: I believe you need to grab the `context(session)` function being passed to `new ApolloServer`, define it above here instead, and pass in that function rather than `context: { ...contextFromOptions }`. In that function there is `if (connection) return context;`, but you might need to make it smarter than that. In general, confirm that auth checks work, or document that subscriptions cannot require auth. Also, it might make sense to move the `contextFuncs` loop to before the `if (connection) return context;` (or generally to make sure it runs for WS).

See also the "ws server usage with custom context value" section [here](https://github.com/enisdenjo/graphql-ws)

packages/api-core/src/ReactionAPICore.js
```diff
@@ -107,6 +107,31 @@ const startOptionsSchema = new SimpleSchema({
...
 
+const getPubSub = () => {
+  const {
+    REDIS_PUB_SUB_URL,
+    REDIS_PUB_SUB_HOST,
+    REDIS_PUB_SUB_PORT,
+    REDIS_PUB_SUB_USERNAME,
+    REDIS_PUB_SUB_PASSWORD,
+    REDIS_PUB_SUB_DB
+  } = config;
```
aldeed, line 118: I would add some info-level logging before each of these returns so that it's clear in the logs which type of pubsub is being used.

aldeed (commented):
This looks like the correct approach, but I left a couple suggestions

packages/api-core/src/createApolloServer.js
aldeed: I noticed after reviewing this that some of the context changes I propose are already done in https://github.com/reactioncommerce/reaction/pull/6671/files. So just pull in those changes and add some tests to confirm.

vanpho93: Updated.

aldeed (approved):
LGTM now

Issue #6667, opened 2022-11-22
Fix incomplete Logger entry


brent-hoover:
https://github.com/reactioncommerce/reaction/blob/51882bb9e49bf4b845673a9907502fe16df1f68f/packages/api-plugin-promotions/src/startup.js#L31

brent-hoover:
Closed, complete

PR #6670, opened 2022-11-23
Use bull queue for jobs to replace api-plugin-job-queue

brent-hoover:
Resolves: #6679 #6667 

This replaces the current api-plugin-job-queue with a plugin backed by [Bull](https://www.npmjs.com/package/bull)

The only existing use of the api-plugin-job-queue was sending email which has been reworked to now use this new plugin. Also removed the ancient fork of nodemailer and started using the standard package. Added a "watcher" for promotions that are ready to made active and marked them active. Emitted an event which will be used to kick off processing existing carts in a subsequence PR.


packages/api-plugin-bull-queue/src/api/addJob.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @summary add a job to a named queue
+ * @param {Object} context - The application context
+ * @param {String} queueName - The queue to add the job to
+ * @param {Object} jobData - Data the job uses to process
+ * @param {Object} options - options for the add job function
+ * @return {Promise<Object>|{Boolean}} - The job instance or false
+ */
+export default function addJob(context, queueName, jobData, options = defaultConfig) {
+  Logger.info({ queueName, ...logCtx }, "Added job to queue");
```
aldeed: "Added" seems wrong here since it's before "Adding"

packages/api-plugin-bull-queue/src/api/createQueue.js
```diff
@@ -0,0 +1,53 @@
...
+ * @param {Function} processorFn - The processor function to use for jobs in the queue
+ * @return {Object|Boolean} - An instance of a BullMQ queue
+ */
+export default function createQueue(context, queueName, options = defaultOptions, processorFn) {
+  if (typeof queueName !== "string" || typeof options !== "object" || typeof processorFn !== "function") {
+    Logger.error(logCtx, "Invalid parameters provided to create queue");
+    return false;
+  }
+  Logger.info({ queueName, ...logCtx }, "Creating queue");
+  if (!options.url) options.url = REDIS_SERVER;
```
aldeed: Ideally you wouldn't mutate the options object since it came from the caller. You could remove this line and instead do:

```js
const newQueue = new Queue(queueName, options.url ?? REDIS_SERVER, options);
```

packages/api-plugin-bull-queue/src/api/createQueue.js
```diff
@@ -0,0 +1,53 @@
...
+  const newQueue = new Queue(queueName, options.url, options);
+  context.bullQueue.jobQueues[queueName] = newQueue;
+  newQueue.process((job) => processorFn(job.data));
```
aldeed: While passing only the job data simplifies the API for your consumers, it means they won't be able to examine the job or call its methods. See properties and methods [here](https://api.docs.bullmq.io/classes/Job.html). Truthfully, I rarely have used most job methods, but certain of them such as `updateProgress` and `retry` sometimes come in handy. A happy medium might be to pass `job` as the second arg, which most people could ignore unless they need to use it.

apps/reaction/package.json
```diff
@@ -71,6 +71,7 @@
     "@reactioncommerce/nodemailer": "5.0.5",
     "@reactioncommerce/random": "1.0.2",
     "@snyk/protect": "latest",
+    "bull": "4.10.1",
```
aldeed: I totally forgot that this package is still maintained. My recent experience is actually with using the `bullmq` package, which has a slightly different API. Seems like either could work, and I'm not sure what the current feature comparison is.

packages/api-plugin-bull-queue/src/api/scheduleJob.js
```diff
@@ -0,0 +1,35 @@
...
+ * @return {Boolean} - true if success
+ */
+export default async function scheduleJob(context, queueName, jobData, schedule) {
+  if (typeof jobData !== "object" || typeof schedule !== "string") {
+    Logger.error(logCtx, "Invalid parameters supplied to scheduleJob");
+    return false;
+  }
+  if (context.bullQueue.jobQueues[queueName]) {
+    const thisQueue = context.bullQueue.jobQueues[queueName];
+    await thisQueue.add(jobData, { repeat: { cron: schedule } });
```
aldeed: You might need either some auto-remove logic here or a `descheduleJob` function. Try calling this twice in a row, with 2 different schedules, which would simulate someone changes the cron config and restarting the server. I don't think it dedupes, so you end up with 2 jobs existing and both scheduled.

packages/api-plugin-email-smtp/src/util/sendSMTPEmail.js
```diff
@@ -11,11 +11,10 @@ import { SMTPConfig } from "../config.js";
  * @returns {undefined} Calls one of the callbacks with a return
  */
 export default async function sendSMTPEmail(context, { job, sendEmailCompleted, sendEmailFailed }) {
-  const { to, shopId, ...otherEmailFields } = job.data;
-
+  const { to, shopId, ...otherEmailFields } = job;
   const transport = nodemailer.createTransport(SMTPConfig);
 
-  transport.sendMail({ to, shopId, ...otherEmailFields }, (error) => {
+  await transport.sendMail({ to, shopId, ...otherEmailFields }, (error) => {
```
aldeed: I'm pretty sure you need to remove the callback argument for this to return a promise, so the `await` would not be necessary. Or you can keep the await and use try/catch instead of the callback.

packages/api-plugin-email/src/startup.js
```diff
@@ -1,12 +1,15 @@
...
 /**
  * @summary Called on startup
  * @param {Object} context Startup context
  * @param {Object} context.collections Map of MongoDB collections
  * @returns {undefined}
  */
-export default function emailStartup(context) {
-  processEmailJobs(context);
+export default async function emailStartup(context) {
+  if (!REACTION_WORKERS_ENABLED) return;
```
aldeed: I would handle these `REACTION_WORKERS_ENABLED` slightly differently. What if you don't want this instance working jobs, but you do still need to schedule emails to be sent, etc.? You'll need the queue to exist to add jobs to it.

You could:
- Make `processorFn` arg optional
- Don't pass `processorFn` when `!REACTION_WORKERS_ENABLED`
- In `createQueue`, skip the `.process` call when `processorFn` isn't provided

I'm not sure which way I lean, but you might also consider respecting `REACTION_WORKERS_ENABLED` env directly in the bull plugin. That way each consumer plugin wouldn't have to remember to separately respect it.

packages/api-plugin-promotions/src/watchers/setPromotionState.js
```diff
@@ -46,30 +54,50 @@ async function markActive(context) {
...
+    return new Promise(async (resolve, reject) => {
+      try {
+        const totalMadeActive = await markActive(context);
+        const totalMarkedCompleted = await markCompleted(context);
+        Logger.info({ ...logCtx, totalMarkedCompleted, totalMadeActive }, "Scanned promotions for changing state");
+        resolve({ totalMarkedCompleted, totalMadeActive });
+      } catch (error) {
+        reject(error);
+      }
+    });
```
aldeed: It doesn't look like the `new Promise` wrapper is necessary here. This would be identical:

```js
async function markPromotionStates() {
  const totalMadeActive = await markActive(context);
  const totalMarkedCompleted = await markCompleted(context);
  Logger.info({ ...logCtx, totalMarkedCompleted, totalMadeActive }, "Scanned promotions for changing state");
  return { totalMarkedCompleted, totalMadeActive };
}
```

apps/reaction/package.json
brent-hoover: I chose this one because their comment seems to imply that the typescript version was not as complete and well-tested. And since the last release was just 2 weeks ago it seems just as well maintained. "safe and battle tested" seemed like what I want

> If you want to start using the next major version of Bull written entirely in Typescript you are welcome to the new repo [here](https://github.com/taskforcesh/bullmq). Otherwise you are very welcome to still use Bull, which is a safe, battle tested codebase.

packages/api-plugin-bull-queue/src/api/createQueue.js
brent-hoover: Yes, that's a good idea

packages/api-plugin-email/src/startup.js
brent-hoover: So I am basically taking this approach
> Don't pass processorFn when !REACTION_WORKERS_ENABLED

In createQueue so that you can add jobs to your queue, but no processor is attached.

packages/api-plugin-bull-queue/src/api/scheduleJob.js
brent-hoover: This feels like something developers would need to do, how would we know that they didn't intend to schedule the job twice with a different schedule? How would the API know it's the same job since we don't keep any global with that info, just queues?

The docs do explicitly say it does not dedupe when the schedule changes and that's the intended behavior.

brent-hoover: Ok, I took a slightly different tact but I think it's still less than optimal. If you add a job with an existing name, it will remove the existing job before adding a new one. I could not figure out a way to determine if the job you were adding was identical to the job that already exists so you could skip both the remove and the add. The thing that is suboptimal is that every time you restart the cluster you have a bunch of servers all removing and re-adding the job all at once with maybe some unpredictable results. Whereas before adding the same job over and over with the same schedule was a no-op. SO I guess I am not completely convinced I made it better

packages/api-plugin-bull-queue/src/shutdown.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @name shutdown
+ * @summary Called on shutdown
+ * @param {Object} context App context
+ * @returns {undefined}
+ */
+export default function bullQueueShutdown(context) {
+  Logger.info("Shutting down bull queue jobs server");
+  return new Promise((resolve, reject) => {
```
vanpho93: I don't see the point of wrapping a Promise here? Can it be a async function.

packages/api-plugin-bull-queue/src/api/scheduleJob.js
```diff
@@ -0,0 +1,43 @@
...
+    return false;
+  }
+  if (context.bullQueue.jobQueues[queueName]) {
+    const thisQueue = context.bullQueue.jobQueues[queueName];
+    const repeatableJobs = await thisQueue.getRepeatableJobs();
+    const jobToRemove = repeatableJobs.find((jbName) => jobName === jbName.name);
+    if (jobToRemove) {
+      await thisQueue.removeRepeatable(jobToRemove);
+      Logger.info({ queueName, jobName, ...logCtx }, "Removed repeatable job");
+    }
```
vanpho93: this can have race condition when many instances start at the same time? This may be weird but can we use a queue to register scheduled jobs.

vanpho93: Actually, `Bull is smart enough not to add the same repeatable job if the repeat options are the same.` so this is good.

packages/api-plugin-bull-queue/src/api/scheduleJob.js
aldeed: Yes, it does seem as though there is no great silver bullet solution. You could add a function options argument for it, such as `replaceExisting: true`, and either default that to `true` or `false` as seems best to you.

packages/api-plugin-bull-queue/src/api/createQueue.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * @summary create a new named instance of the BullMQ Queue
+ * @param {Object} context - The application context
+ * @param {String} queueName - The name of the queue to create, this name is used elsewhere to reference the queue
+ * @param {Object} options - Any additional options to pass to the instance
+ * @param {Function} processorFn - The processor function to use for jobs in the queue
+ * @return {Object|Boolean} - An instance of a BullMQ queue
+ */
+export default function createQueue(context, queueName, options = defaultOptions, processorFn) {
+  if (typeof queueName !== "string" || typeof options !== "object" || typeof processorFn !== "function") {
```
aldeed, line 42: Should the `typeof processorFn !== "function"` check be dependent on whether `REACTION_WORKERS_ENABLED`? I could see arguments for requiring it always, but it truthfully wouldn't be needed if this instance won't run workers.

aldeed (approved):
I didn't test it, but based on reviewing the code patterns this seems like as good a starting point as any. More usage will help determine if changes are needed. It only needs to be as good as the previous solution to be worth merging, and that's not too hard.

vanpho93 (changes requested):
Should update [this file](https://github.com/reactioncommerce/reaction/blob/ef310ac1e18d1a698646e9babe2f5b4dc18946af/docker-compose.circleci.yml) and adding env to this part after rebasing.

```
      - run:
          name: Create .env file
          command: |
            touch .env
            echo "MONGO_URL=mongodb://mongo.reaction.localhost:27017/reaction?&replicaSet=rs0" >> .env
            echo "ROOT_URL=http://localhost:3000" >> .env
            echo "STORE_URL=http://localhost:4000" >> .env
            echo "STRIPE_API_KEY=YOUR_PRIVATE_STRIPE_API_KEY" >> .env
```

brent-hoover:
> Should update [this file](https://github.com/reactioncommerce/reaction/blob/ef310ac1e18d1a698646e9babe2f5b4dc18946af/docker-compose.circleci.yml) and adding env to this part after rebasing.
> 
> ```
>       - run:
>           name: Create .env file
>           command: |
>             touch .env
>             echo "MONGO_URL=mongodb://mongo.reaction.localhost:27017/reaction?&replicaSet=rs0" >> .env
>             echo "ROOT_URL=http://localhost:3000" >> .env
>             echo "STORE_URL=http://localhost:4000" >> .env
>             echo "STRIPE_API_KEY=YOUR_PRIVATE_STRIPE_API_KEY" >> .env
> ```

What exactly am I rebasing? This PR is up to date with trunk

```
[brent@ferret: 15:13:21 ~/Projects/merchstack/monorepo-reaction  (feat/promotions $) - [2023]
 »» git merge release-5
Already up to date.
[brent@ferret: 15:13:30 ~/Projects/merchstack/monorepo-reaction  (feat/promotions $) - [2024]
 »» git co use-bull-queue 
Switched to branch 'use-bull-queue'
Your branch is up to date with 'origin/use-bull-queue'.
[brent@ferret: 15:14:04 ~/Projects/merchstack/monorepo-reaction  (use-bull-queue $) - [2025]
 »» git merge feat/promotions
Already up to date.
```

vanpho93 (approved):

PR #6671, opened 2022-11-28
feat: add cart updated subscription

### LINKED ISSUE #6437: As a developer, I want to enable cart subscription so I know when my cart has been modified
Per our discussions with @aldeed, need to look at upgrading to Apollo 3 and having some sort of persistent backend like Redis or Kafka.

I'd like to see a complete implementation plan including how it would be folded into the development environment.

vanpho93:
Resolves #6437
Impact: **major**
Type: **feature**

## Issue

Unable to notify client when cart changed.

## Solution

Add GraphQL Subscription to cart

### DIFF
--- a/packages/api-plugin-carts/src/schemas/cart.graphql
+++ b/packages/api-plugin-carts/src/schemas/cart.graphql
@@ -419,6 +419,17 @@ input SetEmailOnAnonymousCartInput {
   email: String!
 }
 
+input CartUpdatedInput {
+  "The cart ID"
+  cartId: ID!,
+
+  "The cart account ID"
+  accountId: ID
+
+  "The cart anonymous token"
+  cartToken: String
+}
+
 ####################
 # Payloads
 # These types are used as return values for mutation calls
@@ -594,4 +605,14 @@ extend type Mutation {
     "Mutation input"
     input: UpdateCartItemsQuantityInput!
   ): UpdateCartItemsQuantityPayload!
+}
+
+####################
+ # Subscriptions
+ ####################
+extend type Subscription {
+  "Subscribe to changes to cart"
+  cartUpdated(
+    input: CartUpdatedInput!
+  ): Cart!
 }
--- a/packages/api-plugin-carts/src/util/publishCartUpdatedEvent.js
+++ b/packages/api-plugin-carts/src/util/publishCartUpdatedEvent.js
@@ -1,0 +1,11 @@
+/**
+ * @summary Publishes a cart updated event
+ * @param {Object} context - The application context
+ * @param {Object} cart - The cart that was updated
+ * @param {Boolean} params.publishUpdatedEvent - Whether to prevent publishing the event
+ * @returns {void} - undefined
+ */
+export default function publishCartUpdatedEvent(context, cart, { publishUpdatedEvent = undefined }) {
+  if (!context.app.hasSubscriptionsEnabled || !publishUpdatedEvent) return;
+  context.pubSub.publish("CART_UPDATED", { cartUpdated: cart });
+}
--- a/packages/api-plugin-carts/src/util/publishCartUpdatedEvent.test.js
+++ b/packages/api-plugin-carts/src/util/publishCartUpdatedEvent.test.js
@@ -1,0 +1,29 @@
+import publishCartUpdatedEvent from "./publishCartUpdatedEvent.js";
+
+const context = {
+  app: {
+    hasSubscriptionsEnabled: true
+  },
+  pubSub: {
+    publish: jest.fn()
+  }
+};
+
+test("shouldn't publish event when subscription is disabled", async () => {
+  context.app.hasSubscriptionsEnabled = false;
+  publishCartUpdatedEvent(context, {}, { publishUpdatedEvent: true });
+  expect(context.pubSub.publish).not.toHaveBeenCalled();
+});
+
+test("shouldn't publish event when publishUpdatedEvent arg is undefined", async () => {
+  context.app.hasSubscriptionsEnabled = true;
+  publishCartUpdatedEvent(context, {}, { publishUpdatedEvent: undefined });
+  expect(context.pubSub.publish).not.toHaveBeenCalled();
+});
+
+test("should publish cart updated event", async () => {
+  context.app.hasSubscriptionsEnabled = true;
+  const cart = { _id: "cartId" };
+  publishCartUpdatedEvent(context, cart, { publishUpdatedEvent: true });
+  expect(context.pubSub.publish).toHaveBeenCalledWith("CART_UPDATED", { cartUpdated: cart });
+});
--- a/packages/api-plugin-carts/package.json
+++ b/packages/api-plugin-carts/package.json
@@ -31,6 +31,7 @@
     "@reactioncommerce/random": "^1.0.2",
     "@reactioncommerce/reaction-error": "^1.0.1",
     "accounting-js": "^1.1.1",
+    "graphql-subscriptions": "^2.0.0",
     "lodash": "^4.17.15",
     "simpl-schema": "^1.12.0"
   },
--- a/packages/api-plugin-carts/src/resolvers/Subscription/cartUpdated.js
+++ b/packages/api-plugin-carts/src/resolvers/Subscription/cartUpdated.js
@@ -1,0 +1,49 @@
+import { withFilter } from "graphql-subscriptions";
+import ReactionError from "@reactioncommerce/reaction-error";
+import hashToken from "@reactioncommerce/api-utils/hashToken.js";
+
+/**
+ * @summary Filters the subscription result
+ * @param {Object} payload - The subscription payload
+ * @param {Object} variables - The subscription variables
+ * @returns {Boolean} - Whether the subscription result should be sent to the client
+ */
+export function filter(payload, variables) {
+  const { cartUpdated: cart } = payload;
+  const { input: { cartId, accountId, cartToken } } = variables;
+
+  if (!cart) return false;
+  if (cart._id !== cartId) return false;
+  if (accountId) return cart.accountId === accountId;
+  return cart.anonymousAccessToken === hashToken(cartToken);
+}
+
+/**
+ * @summary Publishes the updated cart to the client
+ * @param {Object} _ unused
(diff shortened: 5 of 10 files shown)


packages/api-plugin-carts/src/resolvers/Subscription/accountCartUpdate.test.js
```diff
@@ -0,0 +1,86 @@
...
+  expect(filter(payload, variables)).toBe(false);
+});
+
+test("filter return true when input cartId and payload cart _id are equal", () => {
+  const payload = { accountCartUpdate: { _id: "cartId" } };
+  const variables = { input: { cartId: "cartId" } };
+  expect(filter(payload, variables)).toBe(true);
+});
+
+test("accountCartUpdate throws access denied error when context is'nt contain userId", async () => {
```
brent-hoover: several super nit-picky comment but it is "isn't" not "is'nt"

packages/api-plugin-carts/src/resolvers/Subscription/accountCartUpdate.test.js
```diff
@@ -0,0 +1,86 @@
...
+  const context = { ...mockContext };
+  const input = {};
+  context.userId = undefined;
+  context.collections.Cart.findOne.mockReturnValueOnce(Promise.resolve({}));
+  await expect(accountCartUpdate.subscribe(null, { input }, context)).rejects.toThrowErrorMatchingSnapshot();
+});
+
+test("accountCartUpdate throws No account found error when userId is not have account", async () => {
```
brent-hoover: "does not"

packages/api-plugin-carts/src/resolvers/Subscription/accountCartUpdate.test.js
```diff
@@ -0,0 +1,86 @@
...
+  const context = { ...mockContext };
+  const input = { cartId: "cartId" };
+  context.collections.Accounts.findOne.mockReturnValueOnce(Promise.resolve(null));
+  context.userId = "noAccountUserId";
+  await expect(accountCartUpdate.subscribe(null, { input }, context)).rejects.toThrowErrorMatchingSnapshot();
+});
+
+test("accountCartUpdate throws validation error when input is'nt contain cartId", async () => {
```
brent-hoover: doesn't not is'nt

packages/api-plugin-carts/src/resolvers/Subscription/anonymousCartUpdate.test.js
```diff
@@ -0,0 +1,72 @@
...
+  withFilter: jest.fn()
+}));
+
+mockContext.pubSub = {
+  asyncIterator: jest.fn().mockName("pubSub.asyncIterator")
+};
+
+beforeEach(() => jest.resetAllMocks());
+
+test("filter returns false when payload does't contain cart", () => {
```
brent-hoover: doesn't

packages/api-plugin-carts/src/resolvers/Subscription/anonymousCartUpdate.test.js
```diff
@@ -0,0 +1,72 @@
...
+  expect(filter(payload, variables)).toBe(false);
+});
+
+test("filter return true when input cartId and payload cart _id are equal", () => {
+  const payload = { anonymousCartUpdate: { _id: "cartId" } };
+  const variables = { input: { cartId: "cartId" } };
+  expect(filter(payload, variables)).toBe(true);
+});
+
+test("anonymousCartUpdate throws validate error when input is'nt contain cartId", async () => {
```
brent-hoover: doesn't

packages/api-plugin-carts/src/util/publishCartUpdatedEvent.test.js
```diff
@@ -0,0 +1,36 @@
...
+const context = {
+  app: {
+    hasSubscriptionsEnabled: true
+  },
+  pubSub: {
+    publish: jest.fn()
+  }
+};
+
+test("should'nt publish event when subscription is disabled", async () => {
```
brent-hoover: shouldn't

brent-hoover (approved):
Looks ok to me except a bunch of spelling errors in test descriptions. Would want @aldeed to look at it as well

packages/api-plugin-carts/src/resolvers/Subscription/accountCartUpdate.test.js
vanpho93: Yeah, I blame Copilot for this.

packages/api-core/src/util/buildSubscriptionContext.js
```diff
@@ -0,0 +1,21 @@
...
+ * @param {Object} args - The Ws server args
+ * @return {Object} The subscription context object
+ */
+export default async function buildSubscriptionContext(contextFromOptions, ctx, msg, args) {
+  const context = { ...ctx, ...contextFromOptions };
+
+  const { connectionParams } = ctx;
+  if (!connectionParams.Authorization) return context;
+
+  const user = await context.queries.getUserFromAuthToken(connectionParams.Authorization);
```
aldeed: This breaks encapsulation of the authentication package. Essentially, there should be no requirement that any authentication plugin is installed, and you should be able to swap auth plugins.

I think you could use the same pattern that is already used:
- In `api-plugin-authentication`, register a second function in the `functionsByType.graphQLContext` array. It should be some combination of `getUserFromAuthToken` and `accountsGraphQL.context` that does not require `req`.
- In `api-plugin-authentication`, don't expose `getUserFromAuthToken` on `queries` object.
- Ideally this change will allow you to pass the same `context` function to both `new ApolloServer` and `useServer`. You might need to adjust certain function calls based on `req` vs. `connectionParams`, but ideally WS should run nearly the same context-creation code as non-WS.

packages/api-plugin-carts/src/resolvers/Subscription/anonymousCartUpdate.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+/**
+ * @summary Publishes the updated cart to the client
+ * @param {Object} _ unused
+ * @param {Object} args - The arguments passed to the subscription
+ * @param {Object} context - The application context
+ * @returns {Promise<Object>} the filtered subscription result
+ */
+async function anonymousCartUpdate(_, args, context) {
```
aldeed: Since this is the subscribe function/query, which is related to an event, I think using a past-tense convention similar to the event name makes sense. In other words, I would add a "d" to make this `anonymousCartUpdated`, and change the file name and schema to match. Same for `accountCartUpdated`.

packages/api-plugin-carts/src/util/publishCartUpdatedEvent.js
```diff
@@ -0,0 +1,15 @@
...
+ * @param {Object} cart - The cart that was updated
+ * @param {Boolean} params.publishUpdatedEvent - Whether to prevent publishing the event
+ * @returns {void} - undefined
+ */
+export default function publishCartUpdatedEvent(context, cart, { publishUpdatedEvent = undefined }) {
+  if (!context.app.hasSubscriptionsEnabled || !publishUpdatedEvent) return;
+  if (cart.accountId) {
+    context.pubSub.publish("ACCOUNT_CART_UPDATED", { accountCartUpdate: cart });
+  } else {
+    context.pubSub.publish("ANONYMOUS_CART_UPDATED", { anonymousCartUpdate: cart });
```
aldeed: It seems fine for both of these to pass `{ cart }` rather than separate names `accountCartUpdate` and `anonymousCartUpdate`. The subscriptions could also then share the same `filter` function if you want.

aldeed: You might even consider simplifying this to emit `CART_UPDATED` always, and then filter on `cart.accountId` in each subscription resolver instead. There are probably pros and cons to each approach, performance vs. consumer complexity.

packages/api-plugin-carts/src/schemas/cart.graphql
```diff
@@ -595,3 +610,18 @@ extend type Mutation {
...
   ): UpdateCartItemsQuantityPayload!
 }
+
+####################
+ # Subscriptions
+ ####################
+extend type Subscription {
+  "Subscribe to changes to a account cart"
+  accountCartUpdate(
+    input: AccountCartUpdateInput!
```
aldeed: Generally the current pattern is to use `input` for mutations but queries just have arguments at the top level. Since these are more like queries than mutations, I would lean toward removing the extra `input` layer. They would match the arguments for `anonymousCartByCartId` and `accountCartByAccountId` queries.

aldeed (changes requested):
This will be nice to have available! I left some suggestions, mostly about patterns and naming.

packages/api-core/src/util/buildSubscriptionContext.js
vanpho93: Updated on v3 PR: https://github.com/reactioncommerce/reaction/blob/f9509ebba25c5df5ff4589647ded75f973b28a3f/packages/api-plugin-authentication/src/index.js#L21

packages/api-plugin-carts/src/util/publishCartUpdatedEvent.js
vanpho93: Updated.

aldeed: @vanpho93 I still think this should be simplified further to:

```js
context.pubSub.publish("CART_UPDATED", { cart });
```

Otherwise it is unnecessary complexity for the event consumer to check for both `anonymousCartUpdate` and `accountCartUpdate` when they often will not care what type of cart it is. If they do care, they can check it with `if (cart.accountId)` in the consumer function.

For example, if you make this change, then your account cart subscription filter can just be:

```js
export function filter(payload, variables) {
  const { cart } = payload;
  const { accountId, shopId } = variables;
  return cart.accountId === accountId && cart.shopId === shopId;
}
```

aldeed (changes requested):
See my comment response about using `cart` as the event payload property. Otherwise looking good.

vanpho93:
@aldeed Updated, can you take another look?

aldeed (approved):

PR #6676, opened 2022-11-30
feat: upgrade to apollo server v4

### LINKED ISSUE #6660: Upgrade to Apollo 4 from Apollo 3
We are on 2 and plan to move to 3, but 3 is not being deprecated so we need to move to 4. This ticket is for the work on moving from 3 to 4.

This work is linked to Promotion because we require Cart subscriptions and subscriptions require the latest version of Apollo

vanpho93:
Resolves #6660
Impact: **major**
Type: **feature**

## Issue

Apollo v3 is deprecated

## Solution

Upgrade to Apollo v4


apps/reaction/tests/util/makeRemoteExecutableSchema.js
```diff
@@ -0,0 +1,29 @@
+import { print } from "graphql";
```
brent-hoover: So @aldeed probably knows this stuff better than I do, but one of the reasons we didn't do Executable Schemas in the past was because it is incompatible with Federated Gateways (which we may or may not do in the future). Is this no longer so?

brent-hoover (approved):
Just one question. For anybody who wrote code that used the underlying express server, is that code going to be broken at all?

apps/reaction/tests/util/makeRemoteExecutableSchema.js
aldeed: I don't know if it's true, so I think this needs testing. If there isn't already an integration test for federation, it might be a good time to add one. Whether or not federation is part of the standard offerings, it should be fairly easy to continue supporting it for those who want to opt in.

aldeed (commented):
I don't know v4 that well but this looks correct based on my reading of docs and what little I do know. I think if it works and federation works and integration tests pass, that should be good.

apps/reaction/tests/util/makeRemoteExecutableSchema.js
vanpho93: This is only for testing on `remoteGraphQL.test.js`, does it affect federation support?

vanpho93: This works with our federation but we need to disable subscription. Apollo Federation does not support yet. I think we can merge this PR and then create another ticket to add test for federation.

aldeed (approved):

PR #6684, opened 2022-12-02
Filter Feature

### LINKED ISSUE #6528: SPIKE: Doing Filtering/Search across Products/Customers/Orders
Initial investigation into what it would take to provide filtering/search functionality and provide a consistent pattern/interface across all Data objects where it makes sense, but for now focused on Products/Customers/Orders.

The definition of done for this ticket is that we provide:
* A specific, detailed proposal for how this would be done
* Preferably a simple proof-on-concept of some subset of the functionality (e.g. just Orders without saving)

Basically broken down into two groups of functionality:

# Filtering
* Be able to filter for records (with pagination) based on any field and a value or subvalue (e.g. a partial match) including NOT. e.g. Orders where date_created > Sep 1 and NOT status = shipped
* Be able to combine filters in AND/OR pairs with (debatable how complex we want these groupings to be)
* Be able to sort this filtered result by any field
* Be able to name and save these filters and use them by name
* Be able to filter search results based on a list of IDs (Orders, Products)
* Be able to filter from product attributes such as Brand
* Be able to see and filter on slug in the search results
* Be able to filter (include or exclude) by Tags
* Be able to sort and filter a list of IDs entered via CSV

sujithvn:

Resolves #6528 
Impact: **major**
Type: **feature**

## Issue
As mentioned in the ticket, requirement is to provide filtering/search functionality and provide a consistent pattern/interface across all Data objects where it makes sense, but for now focused on Products/Customers/Orders.

## Solution
The format of the  input data (conditions) to perform the filter is described in the fle `packages/api-core/src/graphql/schema.graphql`
A new generic function to handle the incoming conditions is defined in `packages/api-utils/lib/generateFilterQuery.js`

For each of the collection where filter feature has to be enabled, we basically need two new files `packages/api-plugin-xxx/src/queries/filterSearchxxx.js` and the corresponding resolver function `packages/api-plugin-xxx/src/resolvers/Query/filterSearchxxx.js`. Also the corresponding `schema.graphql` also has to be updated to add the new query endpoint.

We have implemented this currently in Orders, Products, Accounts/Customers collections and the same set of changes could be used as template for other collections.

## Breaking changes
No breaking change expected as all the files and changes are new.

## Testing
Unit test file for the `generateFilterQuery.js` updated.

### DIFF
--- a/packages/api-core/src/graphql/schema.graphql
+++ b/packages/api-core/src/graphql/schema.graphql
@@ -44,7 +44,7 @@ enum LogOpTypes{
 }
 
 "Relational Operator Types for filtering"
-enum RelOpTypes{
+enum RelationalOperatorTypes{
   "Begins With used with String types"
   beginsWith
 
@@ -91,7 +91,7 @@ enum FilterLevels {
   TWO
 }
 
-"Single Condition for filterSearch"
+"Single Condition for filterSearch, use exactly one of the optional input value type"
 input SingleConditionInput {
   "Value to filter if it is Boolean input"
   boolValue: Boolean
@@ -118,10 +118,10 @@ input SingleConditionInput {
 	key : String!
 
   "Logical NOT operator to negate the condition"
-  logNOT: Boolean
+  logicalNOT: Boolean
 
   "Relational Operator to join the key and value"  
-	relOper: RelOpTypes!
+	relationalOperator: RelationalOperatorTypes!
 
   "Value to filter if it is String Array input"
   stringArrayValue: [String]
@@ -130,31 +130,22 @@ input SingleConditionInput {
 	stringValue: String
 }
 
-"Filter search with Three levels of input"
-input FilterThreeLevelInput {
-  "Array of conditions where all have to be true ($and)"
-  all: [FilterTwoLevelInput]
-
-  "Array of conditions where any have to be true ($or)"
-  any: [FilterTwoLevelInput]
-}
-
-"Filter search with Two levels of input"
-input FilterTwoLevelInput {
-  "Array of conditions where all have to be true ($and)"
-  all: [FilterOneLevelInput]
-
-  "Array of conditions where any have to be true ($or)"
-  any: [FilterOneLevelInput]
-}
-
-"Filter search with One level of input"
-input FilterOneLevelInput {
-  "Array of conditions where all have to be true ($and)"
+"Filter search with One level of conditions (use either 'any' or 'all' not both)"
+input ConditionsArray {
+  "Array of single-conditions"
   all: [SingleConditionInput]
 
-  "Array of conditions where any have to be true ($or)"
+  "Array of single-conditions"
   any: [SingleConditionInput]
+}
+
+"Filter search with nested conditions of input (use either 'any' or 'all' not both)"
+input FilterConditionsInput {
+  "Array holding Nested conditions (use either 'any' or 'all' not both)"
+  all: [ConditionsArray]
+
+  "Array holding Nested conditions (use either 'any' or 'all' not both)"
+  any: [ConditionsArray]
 }
 
 
--- a/packages/api-core/src/graphql/schema.graphql
+++ b/packages/api-core/src/graphql/schema.graphql
@@ -34,16 +34,7 @@ enum MassUnit {
   oz
 }
 
-"Logical Operator Types for filtering"
-enum LogOpTypes{
-  "AND"
-  AND
-
-  "OR"
-  OR
-}
-
-"Relational Operator Types for filtering"
+"Relational Operator Types used in filtering inside a single condition"
 enum RelationalOperatorTypes{
   "Begins With used with String types"
   beginsWith
@@ -79,25 +70,13 @@ enum RelationalOperatorTypes{
   regex
 }
 
-"Flag to specify the number of levels used in Filter Search"
-enum FilterLevels {
-  "Use Filter with 1 level"
-  ONE
-
-  "Use Filter with 3 levels"
-  THREE
-
-  "Use Filter with 2 levels"
-  TWO
-}
-
 "Single Condition for filterSearch, use exactly one of the optional input value type"
 input SingleConditionInput {
   "Value to filter if it is Boolean input"
   boolValue: Boolean
 
   "Flag to set if the regex is case insensitive"
-	caseSensitive: Boolean
+  caseSensitive: Boolean
 
   "Value to filter if it is Date input"
   dateValue: DateTime
@@ -120,8 +99,8 @@ input SingleConditionInput {
   "Logical NOT operator to negate the condition"
   logicalNOT: Boolean
 
-  "Relational Operator to join the key and value"  
-	relationalOperator: RelationalOperatorTypes!
+  "Relational Operator to join the key and value"
+  relationalOperator: RelationalOperatorTypes!
 
   "Value to filter if it is String Array input"
   stringArrayValue: [String]
--- a/packages/api-core/src/graphql/schema.graphql
+++ b/packages/api-core/src/graphql/schema.graphql
@@ -36,13 +36,13 @@ enum MassUnit {
 
 "Relational Operator Types used in filtering inside a single condition"
 enum RelationalOperatorTypes{
-  "Begins With used with String types"
+  "Begins With used with String types to filter based on the beginning of the string"
   beginsWith
 
-  "Ends With used with String types"
+  "Ends With used with String types to filter based on the end of the string"
   endsWith
 
-  "Equal"
+  "Equal to"
   eq
 
   "Greater Than"
@@ -51,7 +51,7 @@ enum RelationalOperatorTypes{
   "Greater Than or Equal"
   gte
 
-  "In"
+  "In used with Array types to filter based on the array containing the value"
   in
 
   "Less Than"
@@ -60,20 +60,20 @@ enum RelationalOperatorTypes{
   "Less Than or Equal"
   lte
 
-  "Not Equal"
+  "Not Equal to"
   ne
 
-  "Not In"
+  "Not In used with Array types to filter based on the array not containing the value"
   nin
 
-  "Regex"
+  "Regex used with String types to filter based on the regex pattern"
   regex
 }
 
-"Single Condition for filterSearch, use exactly one of the optional input value type"
+"Single Condition for filter, use exactly one of the optional input value type"
 input SingleConditionInput {
   "Value to filter if it is Boolean input"
-  boolValue: Boolean
+  booleanValue: Boolean
 
   "Flag to set if the regex is case insensitive"
   caseSensitive: Boolean
@@ -87,17 +87,17 @@ input SingleConditionInput {
   "Value to filter if it is Float input"
   floatValue: Float
 
-  "Value to filter if it is Int Array input" 
-  intArrayValue: [Int]
+  "Value to filter if it is Integer Array input" 
+  integerArrayValue: [Int]
 
-  "Value to filter if it is Int input"
-  intValue: Int
+  "Value to filter if it is Integer input"
+  integerValue: Int
 
   "Field name"
-	key : String!
+  key: String!
 
   "Logical NOT operator to negate the condition"
-  logicalNOT: Boolean
+  logicalNot: Boolean
 
   "Relational Operator to join the key and value"
   relationalOperator: RelationalOperatorTypes!
@@ -106,10 +106,10 @@ input SingleConditionInput {
   stringArrayValue: [String]
 
   "Value to filter if it is String input"
-	stringValue: String
+  stringValue: String
 }
 
-"Filter search with One level of conditions (use either 'any' or 'all' not both)"
+"Filter with One level of conditions (use either 'any' or 'all' not both)"
 input ConditionsArray {
   "Array of single-conditions"
   all: [SingleConditionInput]
@@ -118,7 +118,7 @@ input ConditionsArray {
   any: [SingleConditionInput]
 }
 
-"Filter search with nested conditions of input (use either 'any' or 'all' not both)"
+"Filter with nested conditions of input (use either 'any' or 'all' not both)"
 input FilterConditionsInput {
   "Array holding Nested conditions (use either 'any' or 'all' not both)"
   all: [ConditionsArray]
--- a/packages/api-core/src/graphql/schema.graphql
+++ b/packages/api-core/src/graphql/schema.graphql
@@ -34,6 +34,121 @@ enum MassUnit {
   oz
 }
 
+"Logical Operator Types for filtering"
+enum LogOpTypes{
+  "AND"
+  AND
+
+  "OR"
+  OR
+}
+
+"Relational Operator Types for filtering"
+enum RelOpTypes{
+  "Equal"
+  eq
+
+  "Not Equal"
+  ne
+
+  "Greater Than"
+  gt
+
+  "Greater Than or Equal"
+  gte
+
+  "Less Than"
+  lt
+
+  "Less Than or Equal"
+  lte
+
+  "Begins With used with String types"
+  beginsWith
+
+  "Ends With used with String types"
+  endsWith
+
+  "In"
+  in
+
+  "Not In"
+  nin
+
+  "Regex"
+  regex
+}
+
+"Flag to specify the number of levels used in Filter Search"
+enum FilterLevels {
+  "Use Filter with 1 level"
+  ONE
+
+  "Use Filter with 2 levels"
+  TWO
+
+  "Use Filter with 3 levels"
+  THREE
+}
+
+"Single Condition for filterSearch"
+input SingleConditionInput {
+  "Field name"
+	key : String!
+
+  "Value to filter if it is String input"
+	stringValue: String
+
+  "Value to filter if it is Int input"
+  intValue: Int
+
+  "Value to filter if it is Float input"
+  floatValue: Float
+
+  "Value to filter if it is Boolean input"
+  boolValue: Boolean
+
+  "Value to filter if it is Date input"
+  dateValue: DateTime
+
+  "Value to filter if it is String Array input"
+  stringArrayValue: [String]
+
+  "Value to filter if it is Int Array input" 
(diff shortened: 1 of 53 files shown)


packages/api-core/src/graphql/schema.graphql
```diff
@@ -34,6 +34,130 @@ enum MassUnit {
   oz
 }
 
+"Logical Operator Types for filtering"
+enum LogOpTypes{
+  "AND"
```
brent-hoover: This could use a really well written description/explanation here. GraphQL interfaces should be self-documenting whenever possible.

packages/api-core/src/graphql/schema.graphql
```diff
@@ -34,6 +34,130 @@ enum MassUnit {
...
+  TWO
+}
+
+"Single Condition for filterSearch"
+input SingleConditionInput {
+  "Value to filter if it is Boolean input"
+  boolValue: Boolean
+
+  "Flag to set if the regex is case insensitive"
+	caseSensitive: Boolean
```
brent-hoover: Indentation is inconsistent

packages/api-core/src/graphql/schema.graphql
```diff
@@ -34,6 +34,130 @@ enum MassUnit {
...
+  intValue: Int
+
+  "Field name"
+	key : String!
+
+  "Logical NOT operator to negate the condition"
+  logNOT: Boolean
+
+  "Relational Operator to join the key and value"  
+	relOper: RelOpTypes!
```
brent-hoover: More inconsistent indentation

brent-hoover (changes requested):
This needs some really good documentation and examples before I think anyone is going to be able to figure out how to use this. Probably emphasis on the examples

packages/api-core/src/graphql/schema.graphql
sujithvn: Removed LogTypes as it was not used. Meanwhile, updated the comments with more details

packages/api-core/src/graphql/schema.graphql
sujithvn: There was tab instead of space, updated.

packages/api-core/src/graphql/schema.graphql
sujithvn: There was tab instead of space, updated.

packages/api-core/src/graphql/schema.graphql
```diff
@@ -34,6 +34,100 @@ enum MassUnit {
...
+  floatValue: Float
+
+  "Value to filter if it is Int Array input" 
+  intArrayValue: [Int]
+
+  "Value to filter if it is Int input"
+  intValue: Int
+
+  "Field name"
+	key : String!
```
aldeed: Remove space before colon

packages/api-core/src/graphql/schema.graphql
```diff
@@ -34,6 +34,100 @@ enum MassUnit {
...
+  any: [SingleConditionInput]
+}
+
+"Filter search with nested conditions of input (use either 'any' or 'all' not both)"
+input FilterConditionsInput {
+  "Array holding Nested conditions (use either 'any' or 'all' not both)"
+  all: [ConditionsArray]
+
+  "Array holding Nested conditions (use either 'any' or 'all' not both)"
+  any: [ConditionsArray]
```
aldeed, line 127: This looks like it forces you to have 2 levels, but often you would want to have only 1 level. I believe unions are still not supported for `input` types, but you could do something like this:

```graphql
input FilterConditionsInput {
  "Array of single-conditions"
  all: [SingleConditionInput]

  "Array of single-conditions"
  any: [SingleConditionInput]

  "Array holding Nested conditions (use either 'any' or 'all' not both)"
  allNested: [ConditionsArray]

  "Array holding Nested conditions (use either 'any' or 'all' not both)"
  anyNested: [ConditionsArray]
}
```

I don't love it but it's all I can think of right now.

aldeed (commented):
I didn't test it but it looks like a nice flexible pattern in general.

packages/api-core/src/graphql/schema.graphql
brent-hoover: We actually considered that approach and decided that the forced nesting was the less confusing of the two options.

aldeed (dismissed):

vanpho93 (changes requested):
plugin orders will only works with the latest api-utils after this?
If that, in plugin orders package.json, we should change "@reactioncommerce/api-utils": "^1.16.5" to "1.17.0". The same with accounts / product.

packages/api-core/src/graphql/schema.graphql
```diff
@@ -34,6 +34,100 @@ enum MassUnit {
...
+  "Equal"
+  eq
+
+  "Greater Than"
+  gt
+
+  "Greater Than or Equal"
+  gte
+
+  "In"
```
brent-hoover: These could still use more complete explanations of what they are. Not things like equals but not everyone knows what "Not In" means

packages/api-core/src/graphql/schema.graphql
```diff
@@ -34,6 +34,100 @@ enum MassUnit {
...
+  nin
+
+  "Regex"
+  regex
+}
+
+"Single Condition for filterSearch, use exactly one of the optional input value type"
+input SingleConditionInput {
+  "Value to filter if it is Boolean input"
+  boolValue: Boolean
```
brent-hoover: Do not abbreviate in camel case

packages/api-core/src/graphql/schema.graphql
```diff
@@ -34,6 +34,100 @@ enum MassUnit {
...
+  dateValue: DateTime
+
+  "Value to filter if it is Float Array input" 
+  floatArrayValue: [Float]
+
+  "Value to filter if it is Float input"
+  floatValue: Float
+
+  "Value to filter if it is Int Array input" 
+  intArrayValue: [Int]
```
brent-hoover: Do not abbreviate in camel-case

packages/api-core/src/graphql/schema.graphql
```diff
@@ -34,6 +34,100 @@ enum MassUnit {
...
+
+  "Value to filter if it is Int input"
+  intValue: Int
+
+  "Field name"
+	key : String!
+
+  "Logical NOT operator to negate the condition"
+  logicalNOT: Boolean
```
brent-hoover: I can't think of any reason why you would be capitalize all of NOT

packages/api-plugin-accounts/src/queries/filterSearchAccounts.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+export default async function filterSearchAccounts(context, conditions, shopId) {
+  const { collections: { Accounts } } = context;
+
+  if (!shopId) {
+    throw new Error("shopId is required");
+  }
+  await context.validatePermissions("reaction:legacy:accounts", "read", { shopId });
+
+  const { filterQuery } = generateFilterQuery(context, "Account", conditions, shopId);
```
brent-hoover: The fact that this is `Account` and not `Accounts` seems worrisome

packages/api-plugin-accounts/src/queries/filterSearchCustomers.js
```diff
@@ -0,0 +1,25 @@
...
+  const { collections: { Accounts } } = context;
+
+  if (!shopId) {
+    throw new Error("shopId is required");
+  }
+  await context.validatePermissions("reaction:legacy:accounts", "read", { shopId });
+
+  const { filterQuery } = generateFilterQuery(context, "Account", conditions, shopId);
+
+  filterQuery.groups = { $in: [null, []] }; // filter out non-customer accounts
```
brent-hoover: Are we sure that this is a completely correct way to retrieve only customers?

packages/api-plugin-accounts/src/resolvers/Query/filterSearchAccounts.js
```diff
@@ -0,0 +1,31 @@
...
+ * @summary Query for a list of accounts
+ * @param {Object} _ - unused
+ * @param {Object} args - an object of all arguments that were sent by the client
+ * @param {String} args.shopId - id of shop to query
+ * @param {Object} args.conditions - object containing the filter conditions
+ * @param {Object} context - an object containing the per-request state
+ * @param {Object} info Info about the GraphQL request
+ * @returns {Promise<Object>} Accounts
+ */
+export default async function filterSearchAccounts(_, args, context, info) {
```
brent-hoover: As I have mentioned a couple of times, not sure why we keep on insisting to calling this `filterSearch` when filter would suffice. If we add search functionality this seems like it would be confusing. Why not just call it filter?

packages/api-plugin-accounts/src/resolvers/Query/filterSearchAccounts.js
```diff
@@ -0,0 +1,31 @@
+import getPaginatedResponse from "@reactioncommerce/api-utils/graphql/getPaginatedResponse.js";
+import wasFieldRequested from "@reactioncommerce/api-utils/graphql/wasFieldRequested.js";
+
+/**
+ * @name Query/accounts
```
brent-hoover: Is there a reason this has to be a separate endpoint rather than functionality added on to the existing queries?

packages/api-plugin-products/README.md
```diff
@@ -8,6 +8,159 @@
 
 Products plugin for the [Reaction API](https://github.com/reactioncommerce/reaction)
 
+### Example on how to use Filter Conditions
+
+We have a query endpoint defined in this plugin which allows us to query products collection based on the input GraphQL conditions object. This query endpoint is defined as `filterSearchProducts.js` and it calls the `generateFilterQuery` function from the `api-utils` plugin to generate the MongoDB filter query.
```
brent-hoover: These instructions are just what I was looking for

packages/api-utils/docs/generateFilterQuery.md
```diff
@@ -0,0 +1,149 @@
...
+* Finally there are two more optional fields `caseSensitive` and `logicalNOT`. The `caseSensitive` is a boolean flag to set if the regex is case sensitive. The `logicalNOT` is a boolean flag to set if the condition is to be negated.
+
+FilterConditionsInput format below (from GraphQL Schemas). __Example__ follows the format below:
+
+
+```graphql
+
+"Filter search with nested conditions of input (use either 'any' or 'all' not both)"
+input FilterConditionsInput {
+  "Array holding Nested conditions (use either 'any' or 'all' not both)"
```
brent-hoover: This seems like just asking for this to be outdated by repeating the graphql format in the doc plus I don't see how it adds a lot of value? I would just remove it

packages/api-utils/package.json
```diff
@@ -51,7 +51,8 @@
     "graphql-relay": "^0.9.0",
     "lodash": "^4.17.15",
     "ramda": "^0.28.0",
-    "transliteration": "^2.1.9"
+    "transliteration": "^2.1.9",
+    "simpl-schema": "^1.12.0"
```
brent-hoover, line 55: Can we try this using the latest version of SimpleSchema, i.e. 3.x?

brent-hoover (changes requested):
I think it still needs a little polish

packages/api-core/src/graphql/schema.graphql
sujithvn: Done

packages/api-core/src/graphql/schema.graphql
sujithvn: Updated

packages/api-core/src/graphql/schema.graphql
sujithvn: Updated

packages/api-core/src/graphql/schema.graphql
sujithvn: Updated to logicalNot

packages/api-utils/docs/generateFilterQuery.md
sujithvn: Removed

packages/api-plugin-accounts/src/resolvers/Query/filterSearchAccounts.js
sujithvn: Renamed all the instances
filterSearchAccounts -> filterAccounts
filterSearchCustomers -> filterCustomers
filterSearchOrders -> filterOrders
filterSearchProducts -> filterProducts

packages/api-plugin-accounts/src/queries/filterSearchAccounts.js
sujithvn: It is 'Account' in SimpleSchema.
```
  Account: SimpleSchema {
    pick: [Function: pickOrOmit],
    omit: [Function: pickOrOmit],
    _constructorOptions: { humanizeAutoLabels: true, requiredByDefault: true },
    _cleanOptions: {
      autoConvert: true,
      extendAutoValueContext: {},
      ...
```

packages/api-plugin-accounts/src/queries/filterSearchCustomers.js
sujithvn: The existing query endpoint uses the same approach. Hence I re-used it
`packages/api-plugin-accounts/src/queries/customers.js`

packages/api-plugin-accounts/src/resolvers/Query/filterSearchAccounts.js
sujithvn: The existing query endpoint is using a fixed input structure like so
` const { groupIds, notInAnyGroups } = input;`

The new filter has an entirely different input format to handle multiple conditions. So I assume it will break any existing implementation of the query if we change it.

packages/api-core/src/graphql/schema.graphql
sujithvn: Added description for in, nin, regex, beginsWith, endWith

packages/api-plugin-products/README.md
sujithvn: Ok, I assume we are good with these updated readme

packages/api-plugin-accounts/src/resolvers/Query/filterSearchAccounts.js
brent-hoover: Can we not extend it while not breaking the existing?

sujithvn: I was thinking that we keep the implementation of the new 'filter' feature as a standard/similar implementation across all the selected collections. Something like a common format including the name of the query endpoint.

All the existing end-points (for accounts/customers/orders/products) accept input in different formats and we would have to make different changes in each of these to accept both the old & new format of input.

Note: I have not checked, but we may again hit the earlier problem of graphql not flexible in allowing different input formats.

Please let me know if we want to make these changes for all the collections or selected ones.

brent-hoover: I can see the logic of having consistent naming of the endpoint across collections but I also think it's confusing to have two endpoints that do basically the same thing but differently. What do you suggest to resolve this?

packages/api-utils/package.json
sujithvn: Are you referring to use 3.x only in this plugin (api-utils) while keeping 1.12 in others or make the change across all?

packages/api-plugin-accounts/src/resolvers/Query/filterSearchAccounts.js
sujithvn: Would you be okay to use 'filterXYZ' as the go-forward endpoint across all collections and mark the old ones as 'deprecated'? If so, we can add the new filter to all collections where there is something similar currently in place.

Else, as you mentioned initially, we have to update the existing query end-points.  We can start & test with a simple one like accounts (to confirm graphql does not create issues) and then implement that for the remaining 3 (customers/orders/products). In this case also, we may have to implement this update to other collections where there is something similar currently in place.

brent-hoover: I think I'm fine with the "create new and deprecate" old approach

sujithvn: Ok, I shall find other collections that are using 'filter' like query and add new filter along with the promotions-filter ticket

brent-hoover: No, let's just stick with the three for now

packages/api-utils/package.json
brent-hoover: I was thinking anywhere we are adding new, we should add the current version

brent-hoover: Since this is causing problems let's just go with the standard version of SimpleSchema and look at going to 3 separately

sujithvn: Reverted simpl-schema to 1.12

sujithvn:
> plugin orders will only works with the latest api-utils after this? If that, in plugin orders package.json, we should change "@reactioncommerce/api-utils": "^1.16.5" to "1.17.0". The same with accounts / product.

@vanpho93 
As discussed, updated the api-util version to ~1.17.1 in accounts/products/orders.

vanpho93 (approved):

packages/api-plugin-accounts/src/resolvers/Query/filterAccounts.js
```diff
@@ -0,0 +1,31 @@
...
+ * @name Query/accounts
+ * @method
+ * @memberof Accounts/Query
+ * @summary Query for a list of accounts
+ * @param {Object} _ - unused
+ * @param {Object} args - an object of all arguments that were sent by the client
+ * @param {String} args.shopId - id of shop to query
+ * @param {Object} args.conditions - object containing the filter conditions
+ * @param {Object} context - an object containing the per-request state
+ * @param {Object} info Info about the GraphQL request
```
brent-hoover, line 14: missing - in JSDoc

brent-hoover (approved):

PR #6688, opened 2022-12-06
feat: implement cart messages

### LINKED ISSUE #6642: Add informational warnings to the cart whenever we either remove a promotion or a promotion can't be applied
When we remove a previously applied promotion, or we attempt to apply a qualifying promotion but can't because of things like stackability, we need to add warnings to the cart that can be displayed to the customer. These would include things like:

1. **The severity of the message,**: info, warn, error (I am thinking we might be able to leverage these messages for other things like price change message)
1. **Requires Read acknowledgement:** There may be cases where we want to make sure a customer has read the message and acknowledged it
1. **What happened**: Promotion was removed, etc
1.  **Why it happened**: An explanation for why the promotion was removed
1. **Read**: A boolean that this was displayed,, and if it requires read acknowledgement, that it was acknowledged

These would be manifest by an array of message records on the cart. These would not carry over into the order.

vanpho93:
Resolves #6642
Impact: **major**
Type: **feature**

## Issue

No messages for users when an applied promotion not qualify anymore

## Solution

Add messages field to cart

### DIFF
--- a/packages/api-plugin-carts/src/mutations/acknowledgeCartMessage.js
+++ b/packages/api-plugin-carts/src/mutations/acknowledgeCartMessage.js
@@ -24,7 +24,7 @@ export default async function acknowledgeCartMessage(context, { cartId, messageI
   } else {
     // Anonymous cart
     if (!cartToken) {
-      throw new ReactionError("not-found", "Cart not found");
+      throw new ReactionError("invalid-params", "Cart token not provided");
     }
 
     selector = { _id: cartId, anonymousAccessToken: hashToken(cartToken) };
@@ -45,12 +45,15 @@ export default async function acknowledgeCartMessage(context, { cartId, messageI
     throw new ReactionError("invalid-param", "Message does not require acknowledgement");
   }
 
-  message.acknowledged = true;
+  const { value } = await Cart.findOneAndUpdate(
+    { "_id": cart._id, "messages._id": messageId },
+    { $set: { "messages.$.acknowledged": true } },
+    { returnDocument: "after" }
+  );
 
-  const { result } = await Cart.updateOne({ _id: cart._id }, { $set: { messages: cartMessages } });
-  if (result.n !== 1) {
+  if (!value) {
     throw new ReactionError("server-error", "Unable to update cart");
   }
 
-  return { cart };
+  return { cart: value };
 }
--- a/packages/api-plugin-carts/src/mutations/acknowledgeCartMessage.js
+++ b/packages/api-plugin-carts/src/mutations/acknowledgeCartMessage.js
@@ -1,0 +1,56 @@
+import _ from "lodash";
+import hashToken from "@reactioncommerce/api-utils/hashToken.js";
+import ReactionError from "@reactioncommerce/reaction-error";
+
+/**
+ * @name acknowledgeCartMessage
+ * @method
+ * @summary Query the Cart collection for a cart with the provided accountId and shopId
+ * @param {Object} context -  an object containing the per-request state
+ * @param {Object} params - request parameters
+ * @param {String} [params.accountId] - An account ID
+ * @param {String} [params.shopId] - A shop ID
+ * @param {String} [params.messageId] - A cart message ID
+ * @returns {Promise<Object>|undefined} A Cart document, if one is found
+ */
+export default async function acknowledgeCartMessage(context, { cartId, messageId, cartToken } = {}) {
+  const { collections, accountId } = context;
+  const { Cart } = collections;
+
+  let selector;
+  if (accountId) {
+    // Account cart
+    selector = { _id: cartId, accountId };
+  } else {
+    // Anonymous cart
+    if (!cartToken) {
+      throw new ReactionError("not-found", "Cart not found");
+    }
+
+    selector = { _id: cartId, anonymousAccessToken: hashToken(cartToken) };
+  }
+
+  const cart = await Cart.findOne(selector);
+  if (!cart) {
+    throw new ReactionError("not-found", "Cart not found");
+  }
+
+  const cartMessages = cart.messages || [];
+  const message = _.find(cartMessages, { _id: messageId });
+  if (!message) {
+    throw new ReactionError("not-found", "Message not found");
+  }
+
+  if (!message.requiresReadAcknowledgement) {
+    throw new ReactionError("invalid-param", "Message does not require acknowledgement");
+  }
+
+  message.acknowledged = true;
+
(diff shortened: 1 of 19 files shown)


packages/api-plugin-carts/src/mutations/acknowledgeCartMessage.js
```diff
@@ -0,0 +1,56 @@
+import _ from "lodash";
+import hashToken from "@reactioncommerce/api-utils/hashToken.js";
+import ReactionError from "@reactioncommerce/reaction-error";
+
+/**
+ * @name acknowledgeCartMessage
+ * @method
+ * @summary Query the Cart collection for a cart with the provided accountId and shopId
```
brent-hoover: I don't think this JSDoc is completely accurate. Some copy-pasta maybe

packages/api-plugin-carts/src/mutations/acknowledgeCartMessage.js
```diff
@@ -0,0 +1,56 @@
...
+  const { Cart } = collections;
+
+  let selector;
+  if (accountId) {
+    // Account cart
+    selector = { _id: cartId, accountId };
+  } else {
+    // Anonymous cart
+    if (!cartToken) {
+      throw new ReactionError("not-found", "Cart not found");
```
brent-hoover: Should this maybe be "Cart token not provided"?

packages/api-plugin-carts/src/mutations/acknowledgeCartMessage.js
```diff
@@ -0,0 +1,56 @@
...
+    throw new ReactionError("not-found", "Message not found");
+  }
+
+  if (!message.requiresReadAcknowledgement) {
+    throw new ReactionError("invalid-param", "Message does not require acknowledgement");
+  }
+
+  message.acknowledged = true;
+
+  const { result } = await Cart.updateOne({ _id: cart._id }, { $set: { messages: cartMessages } });
```
brent-hoover: I think this pattern of take an array, modify it and then rewrite the entire away is a bit of an anti-pattern because of the possibility for a race condition (acknowledging another message on the same cart a split second before) could result in data corruption. I think there are more atomic array operations you can use that don't touch other elements.

packages/api-plugin-promotions-discounts/src/discountTypes/item/applyItemDiscountToCart.js
```diff
@@ -85,6 +85,7 @@ export default async function applyItemDiscountToCart(context, params, cart) {
   }
 
   const affected = discountedItems.length > 0;
+  const reason = !affected ? "No items were discounted" : undefined;
```
brent-hoover, line 88: I'm a little confused by this, does this mean this will add a message every time a cart can't be discounted via offers or is this just for coupons or ???

packages/api-plugin-promotions/src/handlers/applyPromotions.js
```diff
@@ -75,23 +124,61 @@ export default async function applyPromotions(context, cart) {
...
       }
-      affected && appliedPromotions.push(promotion);
+
+      if (affected) {
+        appliedPromotions.push(promotion);
+        continue;
+      }
+
+      if (canAddToCartMessages(promotion)) {
+        cartMessages.push(createCartMessage({
```
brent-hoover, line 159: I don't understand why we're pushing a message when nothing was affected?

brent-hoover (changes requested):
Just a couple changes and some questions

packages/api-plugin-promotions/src/handlers/applyPromotions.js
vanpho93: The current in-calculation cart is not affected, but the old cart will be, the promotion that applied in the old cart is no longer works anymore so it will be removed, that's why we put a message.

packages/api-plugin-promotions-discounts/src/discountTypes/item/applyItemDiscountToCart.js
vanpho93: We only add message when the promotion (no matter it's offer or coupon) get removed from a cart, except when user maunally remove it, this will be implemented in coupon later.

packages/api-plugin-carts/src/resolvers/Mutation/acknowledgeCartMessage.js
```diff
@@ -0,0 +1,27 @@
...
+export default async function acknowledgeCartMessage(parentResult, { input }, context) {
+  const { cartId, messageId, clientMutationId = null, cartToken } = input;
+
+  const { cart } = await context.mutations.acknowledgeCartMessage(context, {
+    cartId: decodeCartOpaqueId(cartId),
+    messageId,
+    cartToken
+  });
+
+  return {
```
brent-hoover, line 29: We should probably emit the `afterCartUpdate` message here

brent-hoover (changes requested):
Just one more small change I missed

packages/api-plugin-promotions/src/handlers/applyPromotions.js
```diff
@@ -29,13 +30,31 @@ async function getImplicitPromotions(context, shopId) {
     shopId,
     enabled: true,
     triggerType: "implicit",
-    startDate: { $lt: now },
-    endDate: { $gt: now }
+    startDate: { $lt: now }
```
brent-hoover, line 33: Why did we remove startDate there?

vanpho93: We remove the endDate actually.

pnpm-lock.yaml
```diff
@@ -255,7 +255,7 @@ importers:
       '@reactioncommerce/logger': link:../../packages/logger
       '@reactioncommerce/nodemailer': 5.0.5
       '@reactioncommerce/random': link:../../packages/random
-      '@snyk/protect': 1.1061.0
```
brent-hoover, line 258: Generally I think you should avoid pulling in these Snyk updates so that you can avoid unnecessary merge conflicts uypstream

packages/api-plugin-promotions/src/handlers/applyPromotions.js
brent-hoover: Yes, you're right. Why did we remove the endDate?

vanpho93: Because we want to add message for expired promotion.

brent-hoover (approved):

PR #6691, opened 2022-12-07
Support backslash in getSlug util

tedraykov:
Signed-off-by: tedraykov <tedraykov@gmail.com>

Impact: **minor**
Type: **bugfix**

## Issue

When we save a slug that contains a `/`, the API automatically converts it to `-`.
For example, the tag slug `men/jacket` turns into `men-jacket`

## Solution

Extend the slugify function to support `/` charecter.

## Breaking changes

None

## Testing

- Create a new tag with a slug, that contains a `/`
- Validate that the saved result contains the `/`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@
 #
 # See comment in docker-compose.dev.yml if you want to run for development.
 
-version: "3.4"
+version: "3.9"
 
 networks:
   reaction:
--- a/packages/api-utils/lib/getSlug.js
+++ b/packages/api-utils/lib/getSlug.js
@@ -13,5 +13,5 @@ const { slugify } = require("transliteration");
  * @returns {String} slugified string
  */
 export default function getSlug(slugString) {
-  return (typeof slugString === "string" && slugify(slugString)) || "";
+  return (typeof slugString === "string" && slugify(slugString, { allowedChars: "a-zA-Z0-9-/" })) || "";
 }
--- a/.changeset/happy-snails-study.md
+++ b/.changeset/happy-snails-study.md
@@ -1,0 +1,5 @@
+---
+"@reactioncommerce/api-utils": patch
+---
+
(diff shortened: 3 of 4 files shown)


brent-hoover (approved):

docker-compose.yml
```diff
@@ -4,7 +4,7 @@
 #
 # See comment in docker-compose.dev.yml if you want to run for development.
 
-version: "3.4"
+version: "3.9"
```
brent-hoover, line 7: Why does this PR include a change to the docker file?

vanpho93 (approved):

sujithvn (approved):
Good to go
Could add an additional test with string "new men/jacket"  to get "new-men/jacket"

docker-compose.yml
tedraykov: The syntax we use for the network declaration in the docker compose file is not supported in 3.4

```
networks:
  reaction:
    name: reaction.localhost
    external: true
```

I had errors in the IDE so I decided to bump it. This does not bring breaking behavior from what I've observed. I had all my docker-compose files to version 3.9 locally for months and haven't encountered any problems.

packages/api-utils/lib/getSlug.js
```diff
@@ -13,5 +13,5 @@ const { slugify } = require("transliteration");
  * @returns {String} slugified string
  */
 export default function getSlug(slugString) {
-  return (typeof slugString === "string" && slugify(slugString)) || "";
+  return (typeof slugString === "string" && slugify(slugString, { allowedChars: "a-zA-Z0-9-/" })) || "";
```
brent-hoover, line 16: @tedraykov 

Actually coming back to this I am concerned about this change being a global one and not specific to tags and may result in some unexpected behavior. I wonder if we can add a second argument of allowedChars, so that the default behavior remains the same.

PR #6693, opened 2022-12-08
feat: add archive promotion filter and permissions

vannguyenn:
- Allow filtering promotions by state

### DIFF
--- a/packages/api-plugin-promotions/src/queries/promotions.js
+++ b/packages/api-plugin-promotions/src/queries/promotions.js
@@ -18,13 +18,7 @@ export default async function promotions(context, shopId, filter) {
     }
 
     if (state) {
-      const allowed =
-        state === "archived"
-          ? await context.userHasPermission("reaction:legacy:promotions", "read:archived", { shopId })
-          : true;
-      if (allowed) {
-        selector.state = { $eq: state };
-      }
+      selector.state = { $eq: state };
     }
 
     if (startDate && startDate.eq) {
--- a/packages/api-plugin-promotions/src/queries/promotions.js
+++ b/packages/api-plugin-promotions/src/queries/promotions.js
@@ -31,16 +31,8 @@ export default async function promotions(context, shopId, filter) {
       selector.startDate = { $eq: startDate.eq };
     }
 
-    if (startDate && startDate.beforeInclusive) {
-      selector.startDate = { ...selector.startDate, $lte: startDate.beforeInclusive };
-    }
-
     if (startDate && startDate.before) {
       selector.startDate = { ...selector.startDate, $lt: startDate.before };
-    }
-
-    if (startDate && startDate.afterInclusive) {
-      selector.startDate = { ...selector.startDate, $gte: startDate.afterInclusive };
     }
 
     if (startDate && startDate.after) {
@@ -51,16 +43,8 @@ export default async function promotions(context, shopId, filter) {
       selector.endDate = { $eq: endDate.eq };
     }
 
-    if (endDate && endDate.beforeInclusive) {
-      selector.endDate = { ...selector.endDate, $lte: endDate.beforeInclusive };
-    }
-
     if (endDate && endDate.before) {
       selector.endDate = { ...selector.endDate, $lt: endDate.before };
-    }
-
-    if (endDate && endDate.afterInclusive) {
-      selector.endDate = { ...selector.endDate, $gte: endDate.afterInclusive };
     }
 
     if (endDate && endDate.after) {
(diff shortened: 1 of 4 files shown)


packages/api-plugin-promotions/src/resolvers/Mutation/archivePromotion.js
```diff
@@ -10,7 +10,8 @@
  */
 export default async function archivePromotion(_, { input }, context) {
   const { promotionId, shopId } = input;
-  await context.validatePermissions("reaction:legacy:promotions", "update", { shopId });
+  await context.validatePermissions("reaction:legacy:promotions", "archive", { shopId });
```
brent-hoover: What is the reasoning behind wanting to have a separate permission besides "update"?

vannguyenn: That's the requirements from this ticket https://app.zenhub.com/workspaces/moc-main-board-62fd8e881686130027eafc31/issues/reactioncommerce/kinetic/111. Brendan said we need a separate permission for archived promotions

brent-hoover: I see statements regarding permissions to view archived but nothing about the ability to archive?

brent-hoover: That being said I think it makes sense to add the additional permission, wondering if we should add "unarchive" functionality

brent-hoover: If we add permissions these need to be added to the default roles in this PR

brent-hoover: https://github.com/reactioncommerce/reaction/blob/5eb585913b0c0ec79682355849fb6168c537ba1c/packages/api-plugin-authorization-simple/src/util/defaultRoles.js#L21

vannguyenn: Yeah, I thought we should have separate permissions for archive(view/update), but seems like we only need additional permission for read archived promotions. I will update the PR. Thanks

packages/api-plugin-promotions/src/queries/promotions.js
```diff
@@ -8,21 +8,27 @@
...
+    const { enabled, startDate, endDate, state } = filter;
     // because enabled could be false we need to check for undefined
     if (typeof enabled !== "undefined") {
       selector.enabled = enabled;
     }
+
+    if (state || (state === "archived" && await context.userHasPermission("reaction:legacy:promotions", "read:archived", { shopId }))) {
+      selector.state = { $eq: state };
+    }
+
```
vanpho93: if state === "archived" the permission check will not be reached because it pass the first check.
Or can we just throw an error?

```suggestion
    if (state ) {
      const canUpdate = state !== "archived" || await context.userHasPermission("reaction:legacy:promotions", "read:archived", { shopId }));
      if (canUpdate) selector.state = { $eq: state };
    }
```

packages/api-plugin-promotions/src/resolvers/Mutation/archivePromotion.js
brent-hoover: @vannguyenn If you read my comments I do agree we should have a separate permission for "archive". Do you disagree?

packages/api-plugin-promotions/src/resolvers/Mutation/archivePromotion.js
```diff
@@ -11,6 +11,7 @@
 export default async function archivePromotion(_, { input }, context) {
   const { promotionId, shopId } = input;
   await context.validatePermissions("reaction:legacy:promotions", "update", { shopId });
+
```
brent-hoover: Is there a reason to add these blank lines?

packages/api-plugin-promotions/src/resolvers/Mutation/archivePromotion.js
brent-hoover: Actually I am clarifying with @brendan-shea-pm

vannguyenn: Oh, I thought you said that we only need the additional view archived permission based on the statements in the ticket. I will revert my changes. Thanks

vanpho93 (approved):

vannguyenn:
Hi @zenweasel, I removed all archive permissions. Please help to review it again. Thanks

brent-hoover (approved):

PR #6695, opened 2022-12-09
feat: allow for second parameter on getSlug function for allowed char…

brent-hoover:

Impact: **minor**
Type: **bugfix**

## Issue

getSlug in Tags needs to incorporate slashes however we should not change the global behavior which affects many other items

## Solution

Provide a second parameter which allows you to specify allowedChar

## Breaking changes

None

## Testing
You should be able to create a tag with a slash in the name and have it retain the slash

### DIFF
--- a/packages/api-utils/lib/getSlug.js
+++ b/packages/api-utils/lib/getSlug.js
@@ -10,8 +10,9 @@ const { slugify } = require("transliteration");
  * @see https://www.npmjs.com/package/transliteration
  * @memberof Utils
  * @param {String} slugString - string to slugify
+ * @param {String|Boolean} allowedChars - specify extra characters that are not removed by slugify
  * @returns {String} slugified string
  */
-export default function getSlug(slugString) {
-  return (typeof slugString === "string" && slugify(slugString, { allowedChars: "a-zA-Z0-9-/" })) || "";
+export default function getSlug(slugString, allowedChars = false) {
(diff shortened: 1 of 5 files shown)


packages/api-utils/lib/getSlug.js
```diff
@@ -4,14 +4,17 @@ const require = createRequire(import.meta.url);
 
 const { slugify } = require("transliteration");
 
+const standardSlug = "a-zA-Z0-9-";
```
tedraykov: The default `allowedChars` based on the transliteration documentation is `a-zA-Z0-9-_.~'`

packages/api-utils/lib/getSlug.js
```diff
@@ -4,14 +4,17 @@ const require = createRequire(import.meta.url);
...
+
 /**
  * @name getSlug
  * @summary Return a slugified string using "slugify" from transliteration
  * @see https://www.npmjs.com/package/transliteration
  * @memberof Utils
  * @param {String} slugString - string to slugify
+ * @param {String|Boolean} allowedChars - specify extra characters that are not removed by slugify
```
tedraykov: The official JSDocs syntax for optional parameter with default value:

```suggestion
 * @param {String} [allowedChars=a-zA-Z0-9-_.~'] - specify extra characters that are not removed by slugify
```

tedraykov (changes requested):

tedraykov (approved):

vanpho93 (approved):

PR #6696, opened 2022-12-09
feat: add price type for product

### LINKED ISSUE #6638: Create "price type" attribute that can be used for Promotions
Basically this is used to include or exclude items that are on sale or on clearance.

Stock price types are:

full - (not MSRP, but not discounted)
clearance - Permanently marked down to move
sale - Temporarily on sale (e.g. Black Friday or Mother's Day sale) but return to full price

vanpho93:
Resolves #6638
Impact: **major**
Type: **feature**

Basically this is used to include or exclude items that are on sale or on clearance.

Stock price types are:

full - (not MSRP, but not discounted)
clearance - Permanently marked down to move
sale - Temporarily on sale (e.g. Black Friday or Mother's Day sale) but return to full price

### DIFF
--- a/packages/api-plugin-pricing-simple/src/schemas/schema.graphql
+++ b/packages/api-plugin-pricing-simple/src/schemas/schema.graphql
@@ -80,6 +80,17 @@ type ProductPriceRange {
   range: String
 }
 
+enum PriceType {
+  "The full price of the product"
+  full
+
+  "The price that was permanently marked down to move"
+  clearance
+
+  "Temporarily on sale (e.g. Black Friday or Mother's Day sale) but return to full price"
+  sale
+}
+
 extend type CatalogProduct {
   "Price and related information, per currency"
   pricing: [ProductPricingInfo]!
@@ -88,6 +99,9 @@ extend type CatalogProduct {
 extend type CatalogProductVariant {
   "Price and related information, per currency"
   pricing: [ProductPricingInfo]!
+
+  "The type of price for this variant"
+  priceType: PriceType
 }
 
 extend type Product {
@@ -107,6 +121,14 @@ extend type ProductVariant {
 
   "Pricing information"
   pricing: ProductPricingInfo!
+
+  "The type of price for this variant"
+  priceType: PriceType
+}
+
+extend type CartItem {
+  "The price type of the product"
+  priceType: PriceType
 }
 
 extend input ProductVariantInput {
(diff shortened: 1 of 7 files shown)


packages/api-plugin-pricing-simple/src/schemas/schema.graphql
```diff
@@ -107,6 +107,9 @@ extend type ProductVariant {
 
   "Pricing information"
   pricing: ProductPricingInfo!
+
+  "The price"
+  priceType: String
```
brent-hoover: The comment says "the price" but it is not the price

pnpm-lock.yaml
```diff
@@ -255,7 +255,7 @@ importers:
       '@reactioncommerce/logger': link:../../packages/logger
       '@reactioncommerce/nodemailer': 5.0.5
       '@reactioncommerce/random': link:../../packages/random
-      '@snyk/protect': 1.1061.0
+      '@snyk/protect': 1.1065.0
       graphql: 14.7.0
```
brent-hoover: It would probably be a good habit to get into to not include these Snyk updates as they will catch you with merge conflicts at some point

packages/api-plugin-pricing-simple/src/util/publishProductToCatalog.js
```diff
@@ -13,7 +13,8 @@ function getPricingObject(doc, priceInfo) {
     displayPrice: priceInfo.range,
     maxPrice: priceInfo.max,
     minPrice: priceInfo.min,
-    price: typeof doc.price === "number" ? doc.price : null
+    price: typeof doc.price === "number" ? doc.price : null,
+    priceType: doc.priceType
```
brent-hoover, line 17: This mutation would cause the priceType to be published to the catalog, but I don't see where it would be added to the cart. I think (but not sure) that the mutation is [here](https://github.com/reactioncommerce/reaction/blob/f796e252ef1f8a72ce5b8ea566a6d4172455cd67/packages/api-plugin-carts/src/xforms/xformCartItems.js#L6)

brent-hoover (changes requested):

packages/api-plugin-pricing-simple/src/util/publishProductToCatalog.js
vanpho93: Added,

```
transforms: [
        {
          name: "addPriceTypeToCartItems",
          fn: addPriceTypeToCartItems,
          priority: 10
        }
      ]
   ```

brent-hoover (approved):

Issue #6612, opened 2022-11-02
When a promotion becomes active, all existing carts need to be checked if the promotion retroactively applies


brent-hoover:
Will have to think about how we would do this since there is no real trigger action happening there. It would just be a matter of finding all carts and firing the cart updated event. But it should also probably be metered to avoid some sort of event storm

We should also do the inverse, when a promotion expires, it should be removed from all carts

brent-hoover:
Initial thoughts on this go something like:

1. Cron job runs once an hour, determines if there are any enabled promotions that are not set to active, sets those to active. Does the inverse for active jobs whose time has expired. It records each of those.
2. For the second case: For existing carts where `appliedPromotions` exists and is not empty, a cart change is triggered which should recalculate the applicable promotions. If a promotion is removed, a message is added to the cart to explain why which can be shown to the customer whenever they access the cart again (if they do).
3. For newly active promotions, the same process is applied, a recalc is triggered (now that we are not using events, not sure exactly how we do that). Since this can be a huge number (10k is a very real example, with longer timeouts the number get easily get to 50k) We need to figure out how we spread that work out via a job system so that we don't slam the system all at once. I think one logical way would be work backwards from most recent carts, which are the most likely to not be abandoned carts, and also prioritize non-anonymous carts. 

So something like:

1. Registered carts first (which never expire)
2. Carts within the last 48 hours
3. then 7 days
4. then 30 days
5. then the rest.

Probably some sort of cap where if we have more than X we break it down into a couple of different batches. Each batch could be scheduled as a job that does 1000 records with a slight delay between to avoid swamping and to spread the work out over multiple workers. Hopefully this can run quickly enough that all carts are updated within an hour as the target.

brent-hoover:
Closed, completed

PR #6709, opened 2022-12-14
Process existing carts when promotion is active

brent-hoover:
Resolves: #6612 

## Problem:

When a promotion becomes active or completed, existing carts are not updated

## Solution:

Existing watcher checks promotions if any are ready to become active. When they are it emits an event and the handlePromotionChangedState creates jobs of 500 carts a piece to check each cart if it's promotion state has changed and if it has, to reprocess it.


packages/api-plugin-promotions/src/handlers/handlePromotionChangedState.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+export default async function handlePromotionChangedState(context) {
+  Logger.info(logCtx, "Reprocessing all existing carts because promotion has changed state");
+  const { bullQueue } = context;
+  const cartsCursor = await getCarts(context);
+  const carts = [];
+  let totalCarts = 0;
+  cartsCursor.forEach((cart) => {
+    carts.push(cart._id); // we don't push the whole cart because it can't be completely serialized
+    if (carts.length >= 500) {
```
vanpho93, line 40: Can use _.trunk

packages/api-plugin-promotions/src/utils/checkCartForPromotionChange.js
```diff
@@ -0,0 +1,92 @@
...
+};
+
+
+/**
+ * @summary check if the cart promotion state has changed
+ * @param {Object} context - The application context
+ * @param {Object} Cart - The carts collection from the context
+ * @param {String} cartId - The id of the cart to check
+ * @return {Promise<{reason: null, updated: boolean, cart: *}|{reason: string, updated: boolean, cart: *}>} - Whether its changed and how
+ */
```
vanpho93, line 34: This way is shorter but it does not have reason. If reason is important so your way is better

```js
function normalizePromotions(promotions) {
  _.chain(promotions).sortBy("_id").map(promotion => _.omit(promotion, "updatedAt")).value();
}

return _.isEqual(cart.appliedPromotions, updatedCart.appliedPromotions);
```

packages/api-plugin-promotions/src/handlers/handlePromotionChangedState.js
brent-hoover: Is that a lodash function because I don't see it in the docs

vanpho93: It's here: https://lodash.com/docs/4.17.15#chunk

brent-hoover: so chunk not trunk

brent-hoover: This is a mongo cursor not a regular array so I don't think that will work. the forEach is a cursor method

packages/api-plugin-promotions/src/utils/checkCartForPromotionChange.js
brent-hoover: This is better. I will do it this way. Reason was only in there because I had so many different conditional branches

vanpho93 (approved):

Issue #6438, opened 2022-07-27
Create endpoint that allows introspection in fields available for surcharges/restrictions


brent-hoover:
Both our surcharges and restrictions capabilities allow you to apply "rules" which are `field` `operator` (like eq) `value`. When creating these in Kinetic we need to know which fields are available for use in these operators. Probably could use the SimpleSchema instances to introspect what fields are there and what value types they are. Probably good to investigate exactly how these plugins are implemented to see to exactly what things you can operate on (is it just product or is it anything in the cart)

PR #6731, opened 2022-12-25
Introspect Schema endpoint

### LINKED ISSUE #6529: As a developer I want to be able to query for fields I can filter by on any Object
Several places we allow you to apply various things (shipping restrictions, surcharges). These filters are built on attributes of the Product or Other elements like Order address. We need a way to introspect what attributes are available on any data attribute and what type that attribute is. I believe you could do this by simple creating an endpoint that would allow you to query any Simple Schema object.

I should be able to query the entire object, or by using JSON path syntax like `$.account.address`

Related to this[ this ticket ](https://github.com/reactioncommerce/kinetic/issues/64)in Kinetic

This code could live in the `api-plugin-simple-schema` api plugin. It should also have permissions assigned to it as it should not be available to consumers

sujithvn:

Resolves #6529 #6438 
Impact: **major**
Type: **feature**

## Issue

There is a requirement to introspect any given schema to analyse the fields available and its type with proper permission validation.

## Solution

We have developed a new query end-point [introspectSchema] which accepts the schemaName and shopId and returns the schema details as a JSONObject.

## Breaking changes

All the changes are new and does not impact any existing setup. But since this change will need additional permissions, it will have a mandatory migration to be done to start using this endpoint.

## Testing
Unit test cases currently failing due to issues of the latest simpleSchema 3.4.0 with Jest.


brent-hoover (approved):
This looks good to me. I wonder if we shouldn't merge this into release-5 though? Would also like Brian to look at it.

brent-hoover:
@sujithvn Because this requires a mandatory migration I am changing the root branch to `release-5` as merging it into trunk would cause productions services to stop

packages/api-plugin-simple-schema/src/queries/introspectSchema.js
```diff
@@ -0,0 +1,75 @@
...
+ */
+export default async function introspectSchema(context, { shopId, schemaName } = {}) {
+  const validationString = "reaction:legacy:simpleSchema";
+  inputSchema.validate({ shopId, schemaName });
+
+  // Check for global schema permissions, if not found, check for schema specific permissions
+  try {
+    await context.validatePermissions(validationString, "introspect:*", { shopId });
+  } catch (error) {
+    await context.validatePermissions(validationString, `introspect:${schemaName}`, { shopId }); // if this throws, its valid access denied
```
aldeed: The first argument is the `resource` and when necessary, that can include a fourth piece that targets a specific resource by ID or name.

For example, we do this:

```js
await context.validatePermissions(`reaction:legacy:accounts:${accountId}`, "add:emails", {
  owner: account.userId
});
```

So it makes more sense to me to use `` `reaction:legacy:simpleSchema:${schemaName}` `` as the resource name and `"introspect"` as the action.

The only problem is that in `api-plugin-authorization-simple`, I don't think we ever properly coded `hasPermission` to work with that specific level of permissions. The other auth plugin supported it but not the simple auth.

The way it _should_ work is that you would not need the try/catch but would only do

```js
await context.validatePermissions(`reaction:legacy:simpleSchema:${schemaName}`, "introspect", { shopId });
```

And then if the schema were `cart` for example, the user would be allowed if they had _either_ `reaction:legacy:simpleSchema:cart` permission _or_ `reaction:legacy:simpleSchema` permission. If their permission does not have the specific `:cart` ending, then it means "all".

However, right now I think the simple auth plugin ignores the `:cart` piece and would give "all" access to everyone. I believe we only intended it to work that way as a transition but I guess it got stuck that way.

So... what you have can work as a workaround. Or you could take this opportunity to "fix" the simple auth permission checking.

packages/api-plugin-simple-schema/src/schemas/schema.graphql
```diff
@@ -0,0 +1,19 @@
...
+  schemaName: String!
+
+  "The JSONObject schema for the schema"
+  schema: JSONObject
+}
+
+extend type Query {
+  "Query the fields of a schema"
+  introspectSchema(
+    "The shop ID of the schema"
```
aldeed: Schemas are not shop specific, so this would be "The shop ID to use for checking permissions". But I think it would be better to use global permissions for this, which means you could remove the `shopId` input.

packages/api-plugin-simple-schema/src/queries/introspectSchema.js
```diff
@@ -0,0 +1,75 @@
...
+ * @summary Extract the Schema of the required collection from context, convert to Json and return
+ * @param {Object} context - an object containing the per-request state
+ * @param {Object} args - an object of all arguments that were sent by the client
+ * @param {String} [args.shopId] - shopID
+ * @param {String} [args.schemaName] - schemaName to query
+ * @returns {Promise<Object>} JSON Schema object Promise
+ */
+export default async function introspectSchema(context, { shopId, schemaName } = {}) {
+  const validationString = "reaction:legacy:simpleSchema";
+  inputSchema.validate({ shopId, schemaName });
```
aldeed, line 62: Since it's impossible to have different schemas for different shops, I would remove the `shopId` input and use global permissions. We use global permissions for "accounts" auth as an example.

packages/api-plugin-simple-schema/src/queries/introspectSchema.js
```diff
@@ -0,0 +1,75 @@
...
+  schemaName: String
+});
+
+/**
+ * @name updateJsonSchema
+ * @summary Update the JSON Schema to add additional path field for each object
+ * @param {Object} jsonSchema - JSON Schema object
+ * @returns {Object} JSON Schema object with path field added
+ */
+function updateJsonSchema(jsonSchema) {
```
aldeed: Since the `currentPath` array is in top-level code and this mutates that array, I worry that this may have unexpected results when run multiple times. It would be better to use an argument with a default value:

```
function updateJsonSchema(jsonSchema, currentPath = ["$"]) {
}
```

And then pass it in wherever you call it recursively:

```js
updateJsonSchema(currentObject[key], currentPath); // recursive call
```

That way it will be reset every time this is called from the query function.

packages/api-plugin-simple-schema/src/queries/introspectSchema.js
```diff
@@ -0,0 +1,75 @@
...
+
+const currentPath = ["$"];
+const inputSchema = new SimpleSchema({
+  shopId: String,
+  schemaName: String
+});
+
+/**
+ * @name updateJsonSchema
+ * @summary Update the JSON Schema to add additional path field for each object
```
aldeed: Maybe this could explain _why_ we add this additional field.

aldeed (changes requested):
There are a lot of additional changes in the diff due to pointing at the release branch. Those changes should be merged first or this PR should be rebased off the release branch.

I did my best to review the introspection files, though. See my inline comments.

packages/api-plugin-simple-schema/src/util/checkAndSeedDb.js
```diff
@@ -0,0 +1,65 @@
...
+  * @param {Object} context An object with request-specific state
+  * @param {Function} dataloaderFactory dataloader factory
+  * @param {Function} convertToDataloaderResult function to convert data to array
+  * @returns {Array} converted result
+  */
+export default async function checkAndSeedDb(context) {
+  const { collections: { roles: Roles, Groups } } = context;
+  const allGroups = await Groups.find({}).toArray();
+
+  for (let index = 0; index < allGroups.length; index += 1) {
```
aldeed: Any reason not to use `for...of` here?

packages/api-plugin-simple-schema/src/util/dbVersionCheck.js
```diff
@@ -0,0 +1,54 @@
...
+ * @returns {undefined}
+ */
+export default async function dbVersionCheck(context) {
+  const setToExpectedIfMissing = async () => {
+    // Check if atleast one of the simpleSchema introspect permissions exist in the roles collection
+    const allRoles = await context.collections.roles.find({}).toArray();
+    let rolesFound = false;
+    if (allRoles && Array.isArray(allRoles) && allRoles.length) {
+      rolesFound = allRoles.some((role) => simpleSchemaIntrospectPermissions.includes(role.name));
+    }
```
aldeed: The above could be simplified and optimized to

```js
const existingRole = await context.collections.roles.findOne({ name: { $in: simpleSchemaIntrospectPermissions } });
const rolesFound = !!existingRole;
```

sujithvn: Thank you, updated as recommended

packages/api-plugin-simple-schema/src/util/checkAndSeedDb.js
sujithvn: Updated to use for...of

packages/api-plugin-simple-schema/src/queries/introspectSchema.js
sujithvn: Added additional info as below
```
* @summary Update the JSON Schema to add additional path field for each object. This is as per the requirement from
* the FrontEnd team as it would reduce the work to calculate this property on the client side.
 ```

packages/api-plugin-simple-schema/src/queries/introspectSchema.js
sujithvn: @aldeed cc: @zenweasel 
I checked the code in `hasPermissions` and as you mentioned it is ignoring the 4th part of the resource string which could be used for the resource specifier like ID. Meanwhile, there are few functions which are passing this 4th field to validatePermissions (like `addressBookAdd` and few others), but looks like this is getting discarded.  So if we introduce the validation for this 4th field, I am assuming that, these functions will suddenly start to fail with 'Access' issues.

So I think it would be better to deal with this entire auth plugin stuff separately in a new ticket instead of including this in introspectSchema PR. Please let me know if you disagree.

packages/api-plugin-simple-schema/src/schemas/schema.graphql
sujithvn: @aldeed @zenweasel 

As per an earlier discussion with Brent, we decided to give access to some common schemas (like Cart, Product, Order, Account, Promotion) for users in "shop manager" or "owner" groups. 
The migration script and the checkAndSeed.js reflects this plan. As per my understanding, this by default makes the permissions/roles to be added under a specific shopId. Meanwhile the global_roles (userPermission.__global_roles__) are picked from the global groups (groups without a shopId, namely "accounts manager" and "system-manager"). 

So if we remove the shopId param in the introspectSchema query, then we have to re-visit the earlier plan. At least we have to include these roles (selected schema access roles) in the global-groups in migration/checkAndSeed code.

Please let me know if I should go ahead with those changes as well.

packages/api-plugin-simple-schema/src/queries/introspectSchema.js
sujithvn: Copy-pasting the same response from above

As per an earlier discussion with Brent, we decided to give access to some common schemas (like Cart, Product, Order, Account, Promotion) for users in "shop manager" or "owner" groups. 
The migration script and the checkAndSeed.js reflects this plan. As per my understanding, this by default makes the permissions/roles to be added under a specific shopId. Meanwhile the global_roles (userPermission.__global_roles__) are picked from the global groups (groups without a shopId, namely "accounts manager" and "system-manager"). 

So if we remove the shopId param in the introspectSchema query, then we have to re-visit the earlier plan. At least we have to include these roles (selected schema access roles) in the global-groups in migration/checkAndSeed code.

Please let me know if I should go ahead with those changes as well.

packages/api-plugin-simple-schema/src/queries/introspectSchema.js
sujithvn: Thank you. Made the change and will push soon

packages/api-plugin-simple-schema/src/queries/introspectSchema.js
aldeed: @sujithvn I could be wrong, but I don't think they would start to fail because the changes should be additive. I mean that if a user has `1:2:3` permission, that should still be allowed for `validatePermissions("1:2:3:4")`. `1:2:3` can still mean `1:2:3:*` as it does now, but we would _additionally_ allow any user who has `1:2:3:4` listed.

If you want to do that in a later PR, which I agree would be good, then I think you'll just need to change `checkAndSeedDb` to add only a single permission, `reaction:legacy:simpleSchema/introspect`, which would mean "can introspect any schema".

packages/api-plugin-simple-schema/src/schemas/schema.graphql
aldeed: @sujithvn Thanks for the explanation. That does change the situation slightly. But what if there is an accounts manager who also needs to introspect schemas? This person might not belong to any shops. To support every case, I think `shopId` argument should be kept but should be optional. Along with also changing the description to "The shop ID to use for checking permissions" as I mentioned.

You could consider giving this access to `"system-manager"` by default, too.

Another option would be to make a new default global group "schema viewer", which any user could be added to. Just depends on exactly how all of this will be used over time.

packages/api-plugin-simple-schema/src/queries/introspectSchema.js
sujithvn: @aldeed 
what you said make sense. With the current update, we will extract both `1:2:3` and also `1:2:3:4` (if provided) versions of the permission strings  and  check it against the global & shop level permission (if shopId provided)

Also the permission `reaction:legacy:simpleSchema/introspect` is added in seed script & migration for shop-managet, owner & system-manager

packages/api-plugin-simple-schema/src/schemas/schema.graphql
sujithvn: @aldeed 
updated as you have suggested. 
- shopId is optional
- provided default permission to `system-manager` also

Did not go for the option of new default group schema-viewer in the interest of time to release this

packages/api-plugin-simple-schema/src/queries/introspectSchema.js
sujithvn: same update as above comment

aldeed (approved):
I didn't actually test it, but from a code review perspective this LGTM now!

vanpho93 (approved):
Works on my local.

brent-hoover (approved):
Excited to see this finally merged

PR #6733, opened 2022-12-27
feat: improve docker release flow

### LINKED ISSUE #6714: Create two step process for Docker release flow
To ensure that we have a docker image that starts properly we need to release packages before we release the docker image, so instead of our current 1 step process we would have a process like:

1. Current Step 1 that versions all packages and releases the app and publishes to npm and creates a PR for the docker release
2. A PR that when merged will release the docker file to Docker hub

vanpho93:
Resolves #6714
Impact: **major**
Type: **feature**

## Issue

To ensure that we have a docker image that starts properly we need to release packages before we release the docker image, so instead of our current 1 step process we would have a process like:

Current Step 1 that versions all packages and releases the app and publishes to npm and creates a PR for the docker release
A PR that when merged will release the docker file to Docker hub

## Solution

Create second step for Docker image release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,6 +102,9 @@ jobs:
         command: mongod --oplogSize 128 --replSet rs0 --storageEngine=wiredTiger
         ports:
           - "27017:27017"
+      - image: redis
+        ports:
+          - "6379:6379"
     steps:
       - checkout
       - restore_cache:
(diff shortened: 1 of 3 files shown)


brent-hoover:
@tedraykov Can you please take a look at this as well?

brent-hoover (approved):

tedraykov:
Looks good to me!

tedraykov (approved):

PR #6734, opened 2022-12-29
feat: create standard coupon

### LINKED ISSUE #6723: As an operator I want to create a Standard coupon
As an operator I want to create a standard coupon so that it will trigger my promotion. This is more of a back end ticket to capture the need. Promotions with a Standard coupon will be triggered by a code that is the same code for all customers eg: NRTHFACE5, whereas Promotions with single use codes will be triggered by many different longer, more complex codes that can only be used once eg: uvXDcvdew2xP

The attributes of the standard coupon should include

- [ ] User generated name

- [ ] System generated Coupon ID

- [ ] User generated Coupon Code 

- [ ] Promotion ID (to associate it with the promotion it is a part of)

- [ ] In Store use flag (indicates whether a coupon code can also be used in store, for V.1 this will just be a flag, no other functionality needed around in store redemption)

- [ ] Expiration date (should be inherited from the promotion End Date for V.1, future functionality could include custom end dates such as 30 days from date x)

Reference: Promotions / Detail Standard Coupon Trigger https://www.figma.com/file/vWot2cXXgjcsLpMq4SAX90/Kinetic?node-id=2020%3A54523&t=DX1SCEHurQYZHXY8-0

CreateCoupon

vanpho93:
Resolves #6723
Impact: **major**
Type: **feature**

## Solution
Creating the mutation allows the admin can create a standard coupon.

### DIFF
--- a/packages/api-plugin-promotions-coupons/src/mutations/createStandardCoupon.test.js
+++ b/packages/api-plugin-promotions-coupons/src/mutations/createStandardCoupon.test.js
@@ -11,6 +11,34 @@ test("throws if validation check fails", async () => {
   }
 });
 
+test("throws error when coupon code already created", async () => {
+  const input = { code: "CODE", shopId: "123", promotionId: "123", canUseInStore: true };
+  const coupon = { _id: "123", code: "CODE", promotionId: "promotionId" };
+  const promotion = { _id: "promotionId" };
+  mockContext.collections = {
+    Promotions: {
+      findOne: jest.fn().mockResolvedValueOnce(Promise.resolve(promotion)),
+      find: jest.fn().mockReturnValue({
+        toArray: jest.fn().mockResolvedValueOnce(Promise.resolve([promotion]))
+      })
+    },
+    Coupons: {
+      findOne: jest.fn().mockResolvedValueOnce(Promise.resolve(coupon)),
+      find: jest.fn().mockReturnValue({
+        toArray: jest.fn().mockResolvedValueOnce(Promise.resolve([coupon]))
+      }),
+      // eslint-disable-next-line id-length
+      insertOne: jest.fn().mockResolvedValueOnce(Promise.resolve({ insertedId: "123", result: { n: 1 } }))
+    }
+  };
+
+  try {
+    await createStandardCoupon(mockContext, input);
+  } catch (error) {
+    expect(error.message).toEqual("Coupon code already created");
+  }
+});
+
 test("throws error when promotion does not exist", async () => {
   const input = { code: "CODE", shopId: "123", promotionId: "123", canUseInStore: true };
   mockContext.collections = {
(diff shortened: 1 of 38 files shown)


brent-hoover:
Broken unit tests and broken DCO.

Also, I think we should be moving coupon work into a separate feature branch so we can test the existing functionality separately.

packages/api-plugin-promotions-coupons/src/mutations/createStandardCoupon.test.js
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  try {
+    await createStandardCoupon(mockContext, input);
+  } catch (error) {
+    expect(error.message).toEqual("A coupon code CODE already exists in this promotion window");
+  }
+});
+
+test("should insert a new coupon and results created results", async () => {
```
brent-hoover: Not sure what "results created results" means

packages/api-plugin-promotions-coupons/src/mutations/createStandardCoupon.test.js
```diff
@@ -0,0 +1,95 @@
...
+      code: "CODE",
+      createdAt: jasmine.any(Date),
+      expirationDate: now,
+      maxUsageTimes: 0,
+      maxUsageTimesPerUser: 0,
+      promotionId: "123",
+      shopId: "123",
+      updatedAt: jasmine.any(Date),
+      usedCount: 0
+    }
```
brent-hoover, line 124: Wouldn't hurt to add `calledOnce` expects for findOne and insertOne

pnpm-lock.yaml
```diff
@@ -256,7 +256,7 @@ importers:
       '@reactioncommerce/file-collections-sa-gridfs': link:../../packages/file-collections-sa-gridfs
       '@reactioncommerce/logger': link:../../packages/logger
       '@reactioncommerce/random': link:../../packages/random
-      '@snyk/protect': 1.1071.0
+      '@snyk/protect': 1.1080.0
```
brent-hoover: Should probably exclude this?

brent-hoover (dismissed):

tedraykov (changes requested):
Why is the coupon referred to as Standard Coupon? Does that imply there are nonstandard coupons? If the collection name is just `Coupon` do we need to overstate that it is a standard one?

Apart from that the PR looks good.

vanpho93:
> Why is the coupon referred to as Standard Coupon? Does that imply there are nonstandard coupons? If the collection name is just `Coupon` do we need to overstate that it is a standard one?

@tedraykov a standard coupon does not require userId so it can be applied by anyone.

tedraykov:
This can't be implied from the naming alone, but I haven't seen the full specs for the coupons so I'll take it as is.

tedraykov (dismissed):

PR #6732, opened 2022-12-26
feat: add migration for promotion permissions

vanpho93:
Resolves #6722
Impact: **minor**
Type: **feature**

## Issue
Missing new permissions for the promotion in database.

## Solution

Add migration for Promotion permissions

PR #6738, opened 2023-01-01
Filter for Promotions

### LINKED ISSUE #6689: Add filtering for Promotions collection like was done for Customer/Orders

sujithvn:

Resolves #6689 
Impact: **minor**
Type: **feature**

## Issue

Add filtering for Promotions collection like was done for Customer/Orders (Ticket #6689)

## Solution

Added new query endpoint `filterPromotions` which will provide the requested feature.
To get a list of available fields and their type, the UI could use the new query end-point `introspectSchema` (currently in PR status).

## Breaking changes

None.
But for passing the validate permissions, we need an entry ["reaction:legacy:promotions", "read"] which is handled in the PR #6732 

## Testing

Unit test for the common query function is already included in the Filter PR (6684). This PR is only adding the new end-point in the same format

### DIFF
--- a/packages/api-plugin-promotions/src/schemas/schema.graphql
+++ b/packages/api-plugin-promotions/src/schemas/schema.graphql
@@ -62,6 +62,30 @@ enum PromotionState {
   archived
 }
 
+"The fields by which you are allowed to sort any query that returns an `PromotionConnection`"
+enum PromotionSortByField {
+  "Promotion ID"
+  _id
+
+  "What type of promotion is this"
+  PromotionType
+
+  "What type of trigger this promotion uses"
+  TriggerType
+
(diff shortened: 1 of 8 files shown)


packages/api-plugin-promotions/src/schemas/schema.graphql
```diff
@@ -315,4 +339,34 @@ extend type Query {
 
     sortOrder: String
   ): PromotionConnection!
+
+    "Query to get a filtered list of Accounts"
```
brent-hoover, line 343: Copy pasta

brent-hoover (approved):
One copy-pasta in the GraphQL description but otherwise it looks good

vanpho93 (approved):

PR #6740, opened 2023-01-04
Upgrade Node to 18

### LINKED ISSUE #6396: Move to Node 18
Node 18 is LTS. Let's try to get on top of this. Hopefully not many breaking changes if any.

sujithvn:
Resolves #6396 
Impact: **major**
Type: **feature|bugfix**

## Issue

Current version is 14. Upgrade to 18.10.0

## Solution

Primarily addressed the changes related to 'assert json' for json imports. All json imports are updated to have have `assert { type: "json" }`.
Additional config updates in babel, eslint and docker files (please refer the files changed)
Jest had to be upgraded from v25.5.4 to v29.4.3 to include fixes related to `workerIdleMemoryLimit`
Jasmine2 had to be explicitly installed while performing Jest upgrade

## Breaking changes
Node required version upgraded from 14 to 18.10.0

## Testing
Unit tests and Integration:query tests passed locally. 
Integration:mutation failed in a Mac M1 Pro 32GB system, but passed in a Linux system with similar config.
All test passed in CircleCI after upgrade to LARGE configuration 
Also successfully placed order using old storefront with updated version.


package.json
```diff
@@ -3,15 +3,15 @@
...
+    "node": ">=18.10.0",
     "npm": ">=7"
   },
   "engineStrict": true,
   "scripts": {
     "start:dev": "npm run start:dev -w apps/reaction",
     "start:meteor-blaze-app": "npm run start -w=apps/meteor-blaze-app",
     "build:packages": "pnpm -r run build",
-    "test": "pnpm -r run test",
+    "test": "pnpm -r run test --forceExit",
```
brent-hoover: Is this the right place to put this parameter?

brent-hoover (changes requested):
* Base branch needs to be release-5
* needs changesets for all plugins
* let's make the require/import thing consistent across all plugins

Otherwise seems ok

tedraykov:
> 

I think the changeset we should add to each plugin should be `major` because a plugin without the `assert json` part won't work with reaction 5 and vice versa.

What do you mean by require/import being consistent? Do we have plugins that use require?

brent-hoover:
Yes, we have many. It was the other method we used for reading in the json files

package.json
sujithvn: I believe the parameter was correctly passed to jest as we could see in the earlier successful unit-tests (ex test-unit (79061)).

But thinking about it now, this was one of the option suggested when we had the unit-test stalling with jest v25. Now after the upgrade, the tests are passing without this param. Hence I have removed the same and all tests passed.

sujithvn:
@zenweasel @tedraykov 

Regarding the changeset, I selected all plpugins for Major bump, but was not sure if we could assign Major bump for Reaction4. I assume this could change Reaction to v5 with the merge of this branch to release5 branch and create confusion. Hence I have not yet applied the changeset. I shall discuss tomorrow morning.

Regarding the require to import change, my understanding is that we should be making the change from require -> import only for .json file imports (by including the assert). I found only two files which has this condition and have updated that. (Currently eslint failed, but will fix it along with changeset changes). Rest of the 'require' calls are for few other external modules and files like config.cjs. I assume we need not update those.

aldeed:
External requires can be changed to import only if that package publishes an ESM entry point. It's becoming increasingly common, but still a lot of packages don't. I'd say that's a separate effort, but worth keeping an eye on the release notes of packages as you update them, in case they now include ESM.

apps/reaction/src/checkNodeVersion.cjs
```diff
@@ -4,8 +4,9 @@
 // this check to work on older Node versions that
 // do not support ES modules.
 
+import * as packageJson from "./package.json" assert { type: "json" };
```
aldeed: Some of these have `* as` and some files don't. I think it isn't necessary for JSON files, but not sure.

aldeed (approved):

apps/reaction/src/checkNodeVersion.cjs
sujithvn: @aldeed 
Yes, you are correct. The above ones did not throw any error while testing, but I got error when I made updates to CLI/Admin and used them together. I shall revert those two file updates [[checkNodeVersion.cjs & setupJestTests.js].

I also need to resolve the pnpm-lock.yaml conflict after changing the base branch

cc @zenweasel

vanpho93 (approved):

apps/reaction/src/checkNodeVersion.cjs
sujithvn: Changed only `setupJestTests.js`
`checkNodeVersion.cjs` restored back to the original 'require'

tedraykov (approved):

sujithvn:
> * Base branch needs to be release-5
> * needs changesets for all plugins
> * let's make the require/import thing consistent across all plugins
> 
> Otherwise seems ok

* Base changed to release-5
* Changeset added
* import/require - discussed here https://github.com/reactioncommerce/reaction/pull/6740#issuecomment-1452496117

brent-hoover (approved):

PR #6748, opened 2023-01-11
feat: prerelease

### LINKED ISSUE #6746: Prerelease for packages
To unblock people from trying packages that have been waiting for release.

Approach:
Using [changeset prerelease](https://github.com/changesets/changesets/blob/main/docs/prereleases.md), create version PR on a prerelease branch. When we merge that PR, it will release packages under @next on npm.

vanpho93:

Resolves #6746
Impact: **major**
Type: **feature**

## Issue

To unblock people from trying packages that have been waiting for release.

## Solution

Using [changeset prerelease](https://github.com/changesets/changesets/blob/main/docs/prereleases.md), create version PR on a prerelease branch. When we merge that PR, it will release packages under https://github.com/next on npm.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -10,7 +10,7 @@ on:
 permissions: {}
 jobs:
   version:
-    if: github.event_name == 'workflow_dispatch'
+    if: github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/prerelease'
     permissions:
       contents: write # to create release (changesets/action)
       pull-requests: write # to create pull request (changesets/action)
@@ -42,7 +42,7 @@ jobs:
 
       - name: Enter prerelease mode
         run: pnpm changeset pre enter next || true
-      
+
       - name: Check should create release
         id: should-create-release
         run: |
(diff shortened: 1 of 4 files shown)


brent-hoover:
Concerned about this `bump up packages versions as `1.0.0-next.{number}`. Is that going to be confusing if a plugin is already at 2.x.x?

vanpho93:
> Concerned about this `bump up packages versions as `1.0.0-next.{number}`. Is that going to be confusing if a plugin is already at 2.x.x?

@zenweasel If it's already 2.x.x and we have a major upgrade, the package version will be `3.0.0-next.{number}`

brent-hoover (approved):
This looks good to me but I would like if @tedraykov could take a look at it.

aldeed (commented):
Overall approach looks good to me

tedraykov (approved):
The code seems good. I reproduced the prerelease workflow on a fork and seems to work as expected. The only concern is if we would be able to publish to the @reactioncommerce npmjs org from GitHub action, rather than from a Circle CI as in the normal releases as we don't have the API keys configured in GitHub.

PR #6758, opened 2023-01-13
feat: add metafields as fact on get eligibleitems function

### LINKED ISSUE #6640: All filterable/excludable attributes should be included in the cart by default
Functions that add this extra data to the cart when you add a product

Attributes:
- Tag
- Brand
- Price

vanpho93:
Resolves #6640
Impact: **major**
Type: **feature**

## Usage
Example metafields on the cart item:
```json
...
"metafields" : [
    {
        "key" : "name",
        "value" : "Test name",
        "description": "Test"
    }
]
...
```

Define the inclusion/exclusion rules:
```json
...
"inclusionRules" : {
    "conditions" : {
        "all" : [
            {
                "fact" : "keyValueArray",
                "operator" : "equal",
                "params": {
                    "fieldName": "metafields" => the metafields field in the cart item
                    "inputField": "key",
                    "inputValue": "value",
                    "outputField": "value"
                 },
                "value" : "Test name"
            }
        ]
    }
}
....
```

### DIFF
--- a/packages/api-plugin-carts/src/util/addCartItems.js
+++ b/packages/api-plugin-carts/src/util/addCartItems.js
@@ -110,7 +110,7 @@ export default async function addCartItems(context, currentItems, inputItems, op
       attributes,
       compareAtPrice: null,
       isTaxable: chosenVariant.isTaxable || false,
-      metafields,
+      metafields: metafields || catalogProduct.metafields,
       optionTitle: chosenVariant.optionTitle,
       parcel: chosenVariant.parcel,
       // This one will be kept updated by event handler watching for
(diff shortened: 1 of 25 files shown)


brent-hoover:
I see a few problems with this. Firstly that the fact is called "keyValueArray". Does this just assume it's the cart? What if it's the customer? Secondly what if there is a second key-value-array on the product (as there is for some customers who have extended the product)?

Thirdly, will this work for Kinetic @vannguyenn? There are extra values required to access these values.

What advantage does this approach afford over just flattening out the metafields on the cart items?

vanpho93:
@zenweasel 

> Firstly that the fact is called "keyValueArray". Does this just assume it's the cart?

This only for checking item inclusion / exclusion, so this fact only applies to cart items.

> Secondly what if there is a second key-value-array on the product (as there is for some customers who have extended the product)?

=> They just need to change `"fieldName": "metafields"` to their fields.

> What advantage does this approach afford over just flattening out the metafields on the cart items?
Flattening still works but it will pollute the schema, can be conflicts between fields too.

brent-hoover (approved):

PR #6761, opened 2023-01-13
fix: change start/end dates to DateTime rather than Date

brent-hoover:
Promotion Dates should have always been DateTime so users can set a promo to start at a particular time.


vanpho93:
image
There are some failed tests related to this change.

sujithvn (approved):

PR #6771, opened 2023-02-02
feat: add additional coupon validation

### LINKED ISSUE #6724: Add additional coupon code validation and fields
The backend needs to enforce some elements of Coupon codes

1. Redemption is not case-sensitive
2. Once a promotion is in window the coupon code cannot be edited
3. An operator is prevented from saving a promotion with a coupon code that is a duplicate of a coupon code used in a different promotion during an overlapping promotion window . In other words, an operator could have 2 different promotions with a coupon code "WELCOME" as long as the promotion Start and End dates do not overlap.
4. If an operator tries to Save Changes on a promotion that contains a duplicate coupon code and that promotion overlaps the window with another promotion, they should receive an API error "A promotion with this coupon code is already set to be active during part of this promotion window. Please either adjust your coupon code or your Promotion Start and End Dates

Additional fields"
1. "isUsableInStore": Boolean that currently has no consequences on the backend

vanpho93:

Resolves #6724
Impact: **major**
Type: **featuree**

## Solution

* Add new validations
* Add update coupon mutation

### DIFF
--- a/packages/api-plugin-promotions-coupons/src/mutations/updateStandardCoupon.js
+++ b/packages/api-plugin-promotions-coupons/src/mutations/updateStandardCoupon.js
@@ -49,7 +49,7 @@ export default async function updateStandardCoupon(context, input) {
 
   const now = new Date();
   if (promotion.startDate <= now) {
-    throw new ReactionError("invalid-params", "This coupon cannot be edited because the promotion is on the window time");
+    throw new ReactionError("invalid-params", "Cannot update a coupon for a promotion that has already started");
   }
 
   if (input.code && coupon.code !== input.code) {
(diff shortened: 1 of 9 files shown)


packages/api-plugin-promotions-coupons/src/mutations/updateStandardCoupon.js
```diff
@@ -0,0 +1,83 @@
...
+
+  const coupon = await Coupons.findOne({ _id: couponId, shopId });
+  if (!coupon) throw new ReactionError("not-found", "Coupon not found");
+
+  const promotion = await Promotions.findOne({ _id: coupon.promotionId, shopId });
+  if (!promotion) throw new ReactionError("not-found", "Promotion not found");
+
+  const now = new Date();
+  if (promotion.startDate <= now) {
+    throw new ReactionError("invalid-params", "This coupon cannot be edited because the promotion is on the window time");
```
brent-hoover: We need to improve the wording of this since I can't tell what this means

brent-hoover (approved):
Need to fix the wording of the one error but otherwise this looks good

PR #6777, opened 2023-02-03
fix: update Promotion schema

vannguyenn:
Resolves https://github.com/reactioncommerce/kinetic/issues/172
Impact: **minor**
Type: **bugfix**

## Issue

Promotions's description field should be optional

### DIFF
--- a/packages/api-plugin-promotions/src/simpleSchemas.js
+++ b/packages/api-plugin-promotions/src/simpleSchemas.js
@@ -81,7 +81,8 @@ export const Promotion = new SimpleSchema({
     type: String
   },
   "description": {
-    type: String
+    type: String,
+    optional: true
   },
   "enabled": {
     type: Boolean,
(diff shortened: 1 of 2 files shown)


memery2023 (approved):

vanpho93 (approved):

packages/api-plugin-promotions/src/simpleSchemas.js
```diff
@@ -81,7 +81,8 @@ export const Promotion = new SimpleSchema({
     type: String
   },
   "description": {
-    type: String
+    type: String,
+    optional: true
```
brent-hoover, line 85: If it's going to be optional here, shouldn't it also be optional in the [GraphQL schema](https://github.com/reactioncommerce/reaction/blob/e73f2c972894201f3a3d77626a43f4a873381422/packages/api-plugin-promotions/src/schemas/schema.graphql#L113)?

brent-hoover (changes requested):

brent-hoover (approved):

PR #6781, opened 2023-02-07
feat: use timezone in shop settings for time calculations not server time

### LINKED ISSUE #6603: Promotions should use timezone in shop settings for time calculations not server time
Time zone is set here in the shop settings

image

I think this probably only involves changing `const now = new Date()` to something like `const now = new Date().toLocaleString("en-US", { timeZone: "America/Chicago" });` where "America/Chicago" would be the timezone from ShopSettings. And probably want to define that function globally and maybe cache the timezone.

vanpho93:

Resolves #6603
Impact: **major**
Type: **feature**

## Issue

Currently we using server time for promotion calculation

## Solution

Change to timezone in shop settings

### DIFF
--- a/.changeset/smooth-bags-sleep.md
+++ b/.changeset/smooth-bags-sleep.md
@@ -1,0 +1,5 @@
+---
+"@reactioncommerce/api-utils": patch
+---
+
+feat: using real shop timezone on insetPrimaryShop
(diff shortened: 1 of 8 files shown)


vanpho93:
FE transformed input dates into UTC before sending to server so we don't need to convert before using.

PR #6784, opened 2023-02-08
feat: create migration for old discount

### LINKED ISSUE #6690: Create migration for old discount codes to new promotions-discount
This will allow us to completely remove the old discount code code. This work can't be done until we do Milestone 2 (Coupons) though

vanpho93:
Resolves #6690
Impact: **major**
Type: **feature**

## Issue

Old discount codes will not work in new promotion framework

## Solution

Create migration for carts / discount codes to new promotions / coupons

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "engineStrict": true,
   "scripts": {
-    "start:dev": "pnpm --filter=reaction run start:dev",
+    "start:dev": "npm run start:dev -w apps/reaction",
     "start:meteor-blaze-app": "npm run start -w=apps/meteor-blaze-app",
     "build:packages": "pnpm -r run build",
     "test": "pnpm -r run test",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "engineStrict": true,
   "scripts": {
-    "start:dev": "npm run start:dev -w apps/reaction",
+    "start:dev": "pnpm --filter=reaction run start:dev",
     "start:meteor-blaze-app": "npm run start -w=apps/meteor-blaze-app",
     "build:packages": "pnpm -r run build",
     "test": "pnpm -r run test",
--- a/packages/api-plugin-promotions-coupons/migrations/2.js
+++ b/packages/api-plugin-promotions-coupons/migrations/2.js
@@ -1,0 +1,194 @@
+/* eslint-disable no-await-in-loop */
+import Random from "@reactioncommerce/random";
+import getCurrentShopTime from "./getCurrentShopTime.js";
+
+/**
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   },
   "engineStrict": true,
   "scripts": {
-    "start:dev": "npm run start:dev -w apps/reaction",
+    "start:dev": "pnpm --filter=reaction run start:dev",
```
brent-hoover: Generally we should not just throw in extra changes unrelated to solving a particular problem. We should try to keep PR's as focused as possible. Not blocking on this one though.

packages/api-plugin-promotions-coupons/migrations/2.js
```diff
@@ -0,0 +1,194 @@
...
+import getCurrentShopTime from "./getCurrentShopTime.js";
+
+/**
+ * @summary returns an auto-incrementing integer id for a specific entity
+ * @param {Object} db - The db instance
+ * @param {String} shopId - The shop ID
+ * @param {String} entity - The entity (normally a collection) that you are tracking the ID for
+ * @return {Promise<Number>} - The auto-incrementing ID to use
+ */
+async function incrementSequence(db, shopId, entity) {
```
brent-hoover, line 12: We have a function defined in the code already for this, we should not duplicate it here.

packages/api-plugin-promotions-coupons/migrations/2.js
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * @summary Performs migration down from previous data version
+ * @param {Object} context Migration context
+ * @param {Object} context.db MongoDB `Db` instance
+ * @param {Function} context.progress A function to report progress, takes percent
+ *   number as argument.
+ * @return {undefined}
+ */
+async function down({ progress }) {
```
brent-hoover: For this to be a proper down, it should reverse the above migration

packages/api-plugin-promotions-coupons/migrations/migrationsNamespace.js
```diff
@@ -0,0 +1 @@
+export const migrationsNamespace = "promotion-coupons";
```
brent-hoover, line 1: So I am not seeing the check for version but we need to do that, but ONLY when they have existing discounts created.

brent-hoover (changes requested):

packages/api-plugin-promotions-coupons/migrations/2.js
vanpho93: The defined function is in another plugin, and we only have access to ({ db, progress }) so it's not easy to do.

brent-hoover (approved):

PR #6802, opened 2023-02-16
feat: shipping discount

### LINKED ISSUE #6767: As an operator I want Shipping Discount promotions to stack unless they are actioned on the same Shipping Method
By default an operator will want promos of the Shipping promotion type to stack with each other unless the stacking promos have their action set for the same Shipping Method

### LINKED ISSUE #6773: Promotions: Add shipping discounts
This needs to be added as a new Promotion Type "Shipping Discount"

Add the ability to add a discount (either as a flat amount of as a percentage) to a particular shipping method.

Extra parameters:
* I should be able to include/exclude products from this discount
* I should be able to discount a particular shipping method e.g. UPS Blue, but not UPS Ground
* I should be able to specify the method match parameter optionally as regex so that minor changes don't break the match

The work to complete this ticket should also include any changes to the API to make this work.

vanpho93:
Resolves #6773 #6767
Impact: **major**
Type: **feature**

Adding shipping discount with an option to prevent using 2 shipping discounts for one shipment.

Loom: https://www.loom.com/share/0ea5076ca3004ae9b1be6b3880a32e86

### DIFF
--- a/packages/api-plugin-promotions-discounts/src/discountTypes/shipping/applyShippingDiscountToCart.js
+++ b/packages/api-plugin-promotions-discounts/src/discountTypes/shipping/applyShippingDiscountToCart.js
@@ -3,7 +3,6 @@ import { createRequire } from "module";
 import _ from "lodash";
 import Logger from "@reactioncommerce/logger";
 import recalculateShippingDiscount from "../../utils/recalculateShippingDiscount.js";
-import getTotalDiscountOnCart from "../../utils/getTotalDiscountOnCart.js";
 import formatMoney from "../../utils/formatMoney.js";
 import getEligibleShipping from "../../utils/getEligibleIShipping.js";
 import calculateDiscountAmount from "../../utils/calculateDiscountAmount.js";
@@ -45,14 +44,14 @@ export function createDiscountRecord(params, discountedItem) {
 /**
  * @summary Get the discount amount for a discount item
  * @param {Object} context - The application context
- * @param {Number} totalShippingPrice - The total shipping price
+ * @param {Number} totalShippingRate - The total shipping price
  * @param {Object} actionParameters - The action parameters
  * @returns {Number} - The discount amount
  */
-export function getTotalShippingDiscount(context, totalShippingPrice, actionParameters) {
+export function getTotalShippingDiscount(context, totalShippingRate, actionParameters) {
   const { discountMaxValue } = actionParameters;
 
-  const total = calculateDiscountAmount(context, totalShippingPrice, actionParameters);
+  const total = calculateDiscountAmount(context, totalShippingRate, actionParameters);
   if (typeof discountMaxValue === "number" && discountMaxValue > 0) {
     return Math.min(total, discountMaxValue);
   }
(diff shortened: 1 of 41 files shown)


packages/api-plugin-promotions-discounts/src/discountTypes/shipping/applyShippingDiscountToCart.js
```diff
@@ -1,5 +1,117 @@
...
+  const shippingDiscount = {
+    promotionId: promotion._id,
+    discountType: actionParameters.discountType,
+    discountCalculationType: actionParameters.discountCalculationType,
+    discountValue: actionParameters.discountValue,
+    discountMaxValue: actionParameters.discountMaxValue,
+    dateApplied: new Date(),
+    discountedItemType: "shipping",
+    discountedAmount: discountedItem.amount,
+    stackability: promotion.stackability,
```
brent-hoover, line 39: Are we applying the shipping discount to items? Not sure that's how we want to do that

packages/api-plugin-promotions-discounts/src/discountTypes/shipping/applyShippingDiscountToCart.js
```diff
@@ -1,5 +1,117 @@
...
+ * @param {Object} context - The application context
+ * @param {Number} totalShippingPrice - The total shipping price
+ * @param {Object} actionParameters - The action parameters
+ * @returns {Number} - The discount amount
+ */
+export function getTotalShippingDiscount(context, totalShippingPrice, actionParameters) {
+  const { discountCalculationType, discountValue, discountMaxValue } = actionParameters;
+  const calculationMethod = context.discountCalculationMethods[discountCalculationType];
+
+  const total = formatMoney(calculationMethod(discountValue, totalShippingPrice));
```
brent-hoover: This must be available in a util lib somewhere rather than copy-pasting it here?

brent-hoover (commented):
Seems ok just have some questions:

1. We are taking the discount on the shipping charge, not items right?
2. We are taking the discount only on shipping and not on handling, correct?
3. Have we taken into account when we may have a split shipment (i.e. two items in the shipping array)?

packages/api-plugin-promotions-discounts/src/discountTypes/shipping/applyShippingDiscountToCart.js
vanpho93: No, we are applying it to shipping only, as it shows on the Loom demo.

vanpho93:
> Seems ok just have some questions:
> 
> 1. We are taking the discount on the shipping charge, not items right?
> 2. We are taking the discount only on shipping and not on handling, correct?
> 3. Have we taken into account when we may have a split shipment (i.e. two items in the shipping array)?

1. => Yes, on the shipping charge
2. => Do we want to exclude handling? Currently handling get discounted too but I can exclude it
3. => We have it but we aren't able to demo it with the API because we didn't support split shipment in API level

brent-hoover:
@vanpho93 Ok, double-checking whether we should discount handling or not, so hold on this.

brent-hoover:
@vanpho93 Clarified with @brendan-shea-pm , discounts should only be taken against shipping, not handling

brent-hoover (changes requested):
Discounts don't display in quotes which is critical

brent-hoover:
@vanpho93 Can we add a shipping discount promotion to the sample data as well?

brent-hoover (changes requested):
Shipping quotes are not updated when a promotion becomes active after the cart already has quotes on it. Brian has identified the problem already and is working on a fix

packages/api-plugin-promotions/src/startup.js
```diff
@@ -22,7 +22,7 @@ const logCtx = {
 export default async function startupPromotions(context) {
   const { bullQueue } = context;
   await bullQueue.createQueue(context, "setPromotionState", { jobName: "checkForChangedStates" }, setPromotionState(context));
-  await bullQueue.scheduleJob(context, "setPromotionState", "checkForChangedStates", {}, "*/5 * * * *");
```
brent-hoover: It should not be one but since we're going to make it an env-var it can stay.

brent-hoover (approved):
Great work

brent-hoover:
@vanpho93 In the future please wait for all reviewers I've assigned to approve before merging

PR #6805, opened 2023-02-17
fix: sample image data not showing

### LINKED ISSUE #6797: Images for sample data not showing correctly on storefront

vanpho93:
Resolves #6797 
Impact: **minor**
Type: **bugfix**

## Issue

Sample data for image not showing

## Solution

* Change import path
* Fix error on get image

## Testing

* Run `pnpm install` again to rebuild file-collection
* Update env file and plugin.json
image

### DIFF
--- a/packages/file-collections-sa-gridfs/src/GridFSStore.js
+++ b/packages/file-collections-sa-gridfs/src/GridFSStore.js
@@ -46,7 +46,7 @@ export default class GridFSStore extends StorageAdapter {
     // Add range if this should be a partial read
     if (typeof startPos === "number" && typeof endPos === "number") {
       opts.start = startPos;
-      opts.end = endPos;
+      opts.end = endPos + 1;
     }
 
     debug("GridFSStore _getReadStream opts:", opts);
--- a/packages/file-collections/src/node/getFileDownloadHandler/requestRange.js
+++ b/packages/file-collections/src/node/getFileDownloadHandler/requestRange.js
@@ -22,7 +22,7 @@ export default function requestRange(headers, fileSize) {
   }
 
   const defaultRange = {
-    end: fileSize - 1,
+    end: fileSize,
     len: fileSize,
     partial: false,
     size: fileSize,
--- a/packages/file-collections/src/node/getFileDownloadHandler/requestRange.js
+++ b/packages/file-collections/src/node/getFileDownloadHandler/requestRange.js
@@ -22,7 +22,7 @@ export default function requestRange(headers, fileSize) {
   }
 
   const defaultRange = {
-    end: fileSize,
+    end: fileSize - 1,
     len: fileSize,
     partial: false,
     size: fileSize,
@@ -57,7 +57,7 @@ export default function requestRange(headers, fileSize) {
 
   // Fix invalid (non-numeric) ranges
   if (String(startByte) !== start) startByte = 0;
-  if ((String(endByte) !== end) || endByte === 0) endByte = fileSize;
+  if ((String(endByte) !== end) || endByte === 0) endByte = fileSize - 1;
 
   if (start >= end) {
     return {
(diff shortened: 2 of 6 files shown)


packages/file-collections/src/node/getFileDownloadHandler/requestRange.js
```diff
@@ -22,7 +22,7 @@ export default function requestRange(headers, fileSize) {
   }
 
   const defaultRange = {
-    end: fileSize - 1,
+    end: fileSize,
```
vanpho93: Tagging @aldeed to verify this change.

sujithvn (dismissed):
Went through all pages, all the images are loading correctly

brent-hoover (dismissed):

packages/file-collections/src/node/getFileDownloadHandler/requestRange.js
aldeed: I'm honestly not sure, but this doesn't seem to match what is done below when there's a range. It looks like we parse into `endByte` and `startByte`, but then we use the original string values in the returned object. Shouldn't it be `end: endByte` and `start: startByte` in the returned object? And if so, `endByte = fileSize - 1` will be different from the value of `end` in this default range.

vanpho93: @aldeed , I just updated the code.

there are cases to compare between old version and new version

If we have an image with length 300

old code:

missing range: failed
bytes=0-300: failed
bytes=0-150: failed
bytes=0-299: passed
bytes=0-MALFORMED_END: passed

new code:
missing range: passed
bytes=0-300: passed
bytes=0-150: failed
bytes=0-299: failed
bytes=0-MALFORMED_END: passed

I think bytes=0-300 for an image is standard for length 300. This is response header from an image request on our store front
Screen Shot 2023-02-28 at 16 15 31

brent-hoover (dismissed):

packages/file-collections/src/node/getFileDownloadHandler/requestRange.js
vanpho93: image

After checking [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests#single_part_ranges), I think we should support `bytes=0-299` instead of `bytes=0-3001.

aldeed: @vanpho93 Would that mean it was correct as it was? I think it would be useful to add several unit tests. Then it should be easy to confirm that the return object correctly represents the input header.

Also I think the main issue here is that the final return isn't returning the correct numbers for start/end. `start` and `end` are strings, so if this was TypeScript it would be complaining. It should be:

```js
return {
    end: endByte,
    len: partSize,
    partial: (partSize < fileSize),
    size: fileSize,
    start: startByte,
    unit
  };
```

packages/file-collections-sa-gridfs/src/GridFSStore.js
```diff
@@ -46,7 +46,7 @@ export default class GridFSStore extends StorageAdapter {
     // Add range if this should be a partial read
     if (typeof startPos === "number" && typeof endPos === "number") {
       opts.start = startPos;
-      opts.end = endPos;
+      opts.end = endPos + 1;
```
vanpho93, line 49: @aldeed to explain, before this fix, most of request has endPos as a string so they do not fall in to this `if`.

vanpho93:
@aldeed @zenweasel 

All cases work as expected after this change.

| Cases        | Old           | New  |
| ------------- |:-------------:| -----:|
| Missing range     | Failed | Pass|
| bytes=0-invalid      | Pass      |   Pass |
| bytes=0-length | Fail but take time     |    Fail immediately |
| bytes=0-(length-1) | Pass      |    Pass |
| bytes=0-(length / 2) | Fail      |    Pass |

aldeed (approved):

brent-hoover (approved):

PR #6739, opened 2023-01-03
feat: upgrade mongodb to 5.0

vanpho93:
Signed-off-by: vanpho93 <vanpho02@gmail.com>

Resolves #issueNumber
Impact: **major**
Type: **feature**

* Upgrade mongo image to v5
* Fix insertOne response

PR #6806, opened 2023-02-21
fix: duplicatePromotion endpoint issue

vannguyenn:
- Use `insertedId` instead of `insertedCount` for success check
- The response from `insertOne` doesn't include the `insertedCount` property

### DIFF
--- a/packages/api-plugin-promotions/src/mutations/duplicatePromotion.js
+++ b/packages/api-plugin-promotions/src/mutations/duplicatePromotion.js
@@ -24,8 +24,8 @@ export default async function duplicatePromotion(context, { shopId, promotionId 
   PromotionSchema.validate(newPromotion);
   validateTriggerParams(context, newPromotion);
   const results = await Promotions.insertOne(newPromotion);
-  const { insertedCount } = results;
-  if (!insertedCount) {
+  const { insertedId } = results;
+  if (!insertedId) {
     return {
       success: false,
       errors: [{


vanpho93 (commented):
We are using many insertedCount in other places, this issue could relate to mongodb upgrade so I think I will work on it.

zenweasel:
Does sort of make sense that `insertOne` doesn't return a `insertedCount` though as opposed to `insertMany`?

brent-hoover:
@vannguyenn @vanpho93 What was the outcome of this? If we think we have uncovered another issue can we create a ticket to track the work? I think Chloe's fix is still legit for this particular problem?

vanpho93:
@zenweasel the issue fixed here, just wait to merge to trunk
https://github.com/reactioncommerce/reaction/pull/6739/files#diff-76fff9e095f44a85888aae647be81d11e4001f0cc01ffdba2c0e20d839484bc5

brent-hoover:
@vanpho93 Providing backwards compatibility is good, but shouldn't any new code use the updated API i.e. use insertedId?

vanpho93:
> @vanpho93 Providing backwards compatibility is good, but shouldn't any new code use the updated API i.e. use insertedId?

Yes, you're right, I think this PR is good to go.

brent-hoover (approved):
Thanks for the fix @vannguyenn

PR #6816, opened 2023-02-27
feat: redeemed coupon information

### LINKED ISSUE #6793: As an operator I want to know the date and time a coupon was redeemed on a given order
As an operator I want to know the date and time a coupon was redeemed on a given order. In V.1 we won't have a place to view this in the UI but wanted to note that we should track this information

- [ ] Date and Time of Redemption
- [ ] Coupon ID
- [ ] Promo ID
- [ ] Order ID
- [ ] Customer ID (if available)

vanpho93:
Resolves #6793 
Impact: **major**
Type: **feature**

## Solution

Store more information and add API to query coupon logs.

### DIFF
--- a/packages/api-plugin-promotions-coupons/src/schemas/schema.graphql
+++ b/packages/api-plugin-promotions-coupons/src/schemas/schema.graphql
@@ -65,9 +65,6 @@ type CouponLog {
 
   "The time the coupon was used"
   createdAt: Date
-
-  "The log details for each time the coupon was used"
-  usedLogs: [JSONObject]
 }
 
 extend type Promotion {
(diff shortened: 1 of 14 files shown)


packages/api-plugin-promotions-coupons/src/schemas/schema.graphql
```diff
@@ -42,11 +42,44 @@ type Coupon {
...
+  accountId: ID
+
+  "The coupon code"
+  usedCount: Int
+
+  "The time the coupon was used"
+  createdAt: Date
+
+  "The log details for each time the coupon was used"
+  usedLogs: [JSONObject]
```
brent-hoover: Isn't this just going to be an array of a known entity? Seems like we could have a real shape of it defined?

brent-hoover (changes requested):
Just one comment about the GraphQL schema

packages/api-plugin-promotions-coupons/src/schemas/schema.graphql
vanpho93: Removed as we don't need it for now.

brent-hoover (approved):

PR #6823, opened 2023-03-10
feat: stackability resolution strategy

### LINKED ISSUE #6820: Stackability Resolution Strategy
Most promotions with be set to not stack with promotions of the same type. For Order promos, they don't stack with other Order promos. Same for Item and Shipping, but the stacking is checked per Item and Shipping Method.

When a promotion is added to a cart that already has a conflicting promotion, the outcome should be that the promotion that gives the customer the highest value discount should remain on the cart, and the lesser one removed.

When a promotion is triggered, check whether there are any conflicting promotion types in the order, item(s), or shipping method 

If there is a conflict, calculate the value of each conflicting promotion

Remove any promotion that does not have the highest value, so that only the highest value promotion remains

vanpho93:
Resolves #6820 
Impact: **major**
Type: **feature**

## Issue

Currently we pick promotions randomly.

## Solution

Pick the best combination of promotions and let other plugin override this implementation.

### DIFF
--- a/packages/api-plugin-promotions-discounts/src/handlers/getApplicablePromotions.test.js
+++ b/packages/api-plugin-promotions-discounts/src/handlers/getApplicablePromotions.test.js
@@ -1,0 +1,35 @@
+import mockContext from "@reactioncommerce/api-utils/tests/mockContext.js";
+import getApplicablePromotions from "./getApplicablePromotions.js";
+import getPromotionCombinations from "./getPromotionCombinations.js";
+import getHighestCombination from "./getHighestCombination.js";
+
+jest.mock("./getPromotionCombinations.js");
+jest.mock("./getHighestCombination.js");
+
+const promo1 = { _id: "1", promotionType: "order-discount", actions: [{ actionParameters: { discountCalculationMethod: "flat" } }] };
+const promo2 = { _id: "2", promotionType: "item-discount", actions: [{ actionParameters: { discountCalculationMethod: "percentage" } }] };
+const promo3 = { _id: "3", promotionType: "item-discount", actions: [{ actionParameters: { discountCalculationMethod: "fixed" } }] };
+const promo4 = { _id: "4", promotionType: "order-discount", actions: [{ actionParameters: { discountCalculationMethod: "fixed" } }] };
+const promo5 = { _id: "5", promotionType: "order-discount", actions: [{ actionParameters: { discountCalculationMethod: "flat" } }] };
+const promo6 = { _id: "6", promotionType: "shipping-discount", actions: [{ actionParameters: { discountCalculationMethod: "flat" } }] };
+
+test("getApplicablePromotions returns correct promotions", async () => {
+  const promotions = [promo1, promo2, promo3, promo4, promo5, promo6];
+
+  const cart = {
+    _id: "cartId"
+  };
+
+  const highestPromotions = [promo1, promo3, promo4];
+  const combinations = [[promo1, promo2], [promo1, promo3, promo4], [promo5], [promo6]];
+  getPromotionCombinations.mockReturnValueOnce(combinations);
+  getHighestCombination.mockReturnValueOnce(highestPromotions);
+
+  const applicablePromotions = await getApplicablePromotions(mockContext, cart, promotions);
+
+  const sortedPromotionsWithoutShippingDiscount = [promo1, promo5, promo2, promo3, promo4];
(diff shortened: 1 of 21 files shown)


packages/api-plugin-promotions-discounts/src/handlers/getApplicablePromotions.test.js
sujithvn, line 1: We could add 2 more promos with discountCalculationMethod = none and null

packages/api-plugin-promotions-discounts/src/handlers/getHighestCombination.js
```diff
@@ -0,0 +1,37 @@
...
+ * @returns {Promise<Array<Object>>} - The highest combination
+ */
+export default async function getHighestCombination(context, cart, combinations) {
+  const { promotions: { enhancers, utils } } = context;
+
+  const tasks = combinations.map(async (combinationPromotions) => {
+    let copiedCart = _.cloneDeep(cart);
+    for (const promo of combinationPromotions) {
+      // eslint-disable-next-line no-await-in-loop
+      const { affected, temporaryAffected } = await utils.actionHandler(context, copiedCart, promo);
```
sujithvn, line 26: I assume the return values `affected` & `temporaryAffected` will be defined by the `actionHandler` funcs of child plugins (ex `packages/api-plugin-promotions-discounts/src/discountTypes/order/applyOrderDiscountToCart.js`). I have not reviewed them now and will review while merging feat/promotions.
Meanwhile, I could not find the field `temporaryAffected` being returned by any handler. only affected & reason were returned

packages/api-plugin-promotions/src/utils/index.js
sujithvn, line 1: Should we add remaining entries from util folder here?

packages/api-plugin-promotions/src/utils/actionHandler.js
```diff
@@ -0,0 +1,27 @@
...
+  let affected = false;
+  let temporaryAffected = false;
+  let rejectedReason;
+  for (const action of promotion.actions) {
+    const actionFn = actionHandleByKey[action.actionKey];
+    if (!actionFn) continue;
+
+    // eslint-disable-next-line no-await-in-loop
+    const result = await actionFn.handler(context, enhancedCart, { promotion, ...action });
+    ({ affected, temporaryAffected, reason: rejectedReason } = result);
```
sujithvn, line 23: As mentioned earlier, I could not find the field `temporaryAffected` being returned by any handler. May be I missed.

packages/api-plugin-promotions-discounts/src/handlers/getPromotionCombinations.js
sujithvn, line 1: @vanpho93 

I am not sure if my understanding is correct, thought of verifying with you.

I added few additional promotions (assuming that it is valid to have more than 1+ explicit promos also)
New input:
```
const promotion1 = {
    _id: "promotionId1",
    triggerType: "implicit",
    discount: 2,
    type: "order",
    stackability: {
      key: "all"
    }
  };

  const promotion1ex = {
    _id: "promotionId1ex",
    triggerType: "explicit",
    discount: 2,
    type: "order",
    stackability: {
      key: "all"
    }
  };

  const promotion2 = {
    _id: "promotionId2",
    triggerType: "implicit",
    discount: 3,
    type: "item",
    stackability: {
      key: "all"
    }
  };

  const promotion2ex = {
    _id: "promotionId2ex",
    triggerType: "explicit",
    discount: 3,
    type: "item",
    stackability: {
      key: "all"
    }
  };

  const promotion3 = {
    _id: "promotionId3",
    triggerType: "implicit",
    discount: 4,
    type: "item",
    stackability: {
      key: "all"
    }
  };

  const promotion3ex = {
    _id: "promotionId3ex",
    triggerType: "implicit",
    discount: 4,
    type: "item",
    stackability: {
      key: "none"
    }
  };

  const promotion4 = {
    _id: "promotionId4",
    triggerType: "implicit",
    discount: 4,
    type: "item",
    stackability: {
      key: "per-type"
    }
  };
```

My expected output was :
```
[
    [promotion1, promotion4],
    [promotion1, promotion1ex, promotion2, promotion2ex, promotion3],
    [promotion3ex]
  ]
```

Actual output was :
```
[
    [promotion4],
    [promotion1, promotion1ex, promotion2, promotion2ex, promotion3],
    [promotion3ex]
  ]
```
I was expecting prom1(type=order) could combine with promo4(type=item)

sujithvn (approved):

PR #6839, opened 2023-05-15
Support Dynamic Password Reset URL

### LINKED ISSUE #6838: Allow configurable password reset URL
## Problem:

Currently, when we initiate a password reset through the `sendResetAccountPasswordEmail`,  the authorization plugin prepares an email with the following reset URL:

https://github.com/reactioncommerce/reaction/blob/ecb8ef62d8312ed62cb05da64bec1a86fe0b188d/packages/api-plugin-authentication/src/util/accountServer.js#L41

Where `STORE_URL` is a configurable environmental variable and `token` is the password reset token provided by `Accounts JS` that should be used later in the `resetPassword` mutation.

When the customer navigates to the generated link in the mail, they should land on the storefront page responsible for picking the new password.

The format of the generated URL is hardcoded in the authorization plugin and enforces the storefront to implement the password reset UI in one specific route.

An example password reset URL would look like this:

```
http://localhost:4000/?resetToken=awjdztqpjdasiawejaspo
```

This forces the storefront developers to implement a password reset login in the index route of the app which may not be desirable.

## Proposed solution:

Instead of hardcoding the URL, we can parametrize it by introducing a password reset path fragment environmental variable like `PASSWORD_RESET_PATH_FRAGMENT` and turn the password reset URL into:

```
${STORE_URL}/${PASSWORD_RESET_PATH_FRAGMENT}${token}
```

If we provide the default value of this env var to be `?resetToken=`, it will support backward compatibility.

Technically we can set the `STORE_URL` to a more specific route like `http://localhost:4000/password-reset` that will evaluate to an URL, but the name of the variable doesn't imply that it will be only used in the password reset scenario. That's why I think it's a better idea to add additional configurable fragment to provide flexibility.

tedraykov:
Resolves #6838
Impact: **minor**
Type: **feature**

## Issue

The authentication plugin generates a password reset email with a hardcoded URL.

## Solution

Instead of hardcoding the URL, we can parametrize it by introducing a password reset path fragment environmental variable like PASSWORD_RESET_PATH_FRAGMENT and turn the password reset URL into:

```
${STORE_URL}/${PASSWORD_RESET_PATH_FRAGMENT}${token}
```

If we provide the default value of this env var to be ?resetToken=, it will support backward compatibility.

Technically we can set the STORE_URL to a more specific route like http://localhost:4000/password-reset that will evaluate to an URL, but the name of the variable doesn't imply that it will be only used in the password reset scenario. That's why I think it's a better idea to add additional configurable fragment to provide flexibility.

## Breaking changes

None. The default value of the newly proposed environmental variable provides backward compatibility.

## Testing

1. Set up Opencommerce with a working email client.
2. Set up the `PASSWORD_RESET_PATH_FRAGMENT` to a desired path
3. Call the `sendResetAccountPasswordEmail` for an existing account.
4. Validate that the password reset URL in the received email matches the desired format.

### DIFF
--- a/.changeset/hip-owls-hug.md
+++ b/.changeset/hip-owls-hug.md
@@ -1,0 +1,5 @@
+---
+"@reactioncommerce/api-plugin-authentication": minor
+---
+
+Allow dynamic password reset URL in the email
--- a/packages/api-plugin-authentication/src/config.js
+++ b/packages/api-plugin-authentication/src/config.js
@@ -5,6 +5,7 @@ const { str } = envalid;
 export default envalid.cleanEnv(
   process.env,
   {
+    PASSWORD_RESET_PATH_FRAGMENT: str({ default: "?resetToken=" }),
     STORE_URL: str({ devDefault: "http://localhost:4000" }),
     TOKEN_SECRET: str({ default: "UPDATE_THIS_SECRET" })
   },
--- a/packages/api-plugin-authentication/src/util/accountServer.js
+++ b/packages/api-plugin-authentication/src/util/accountServer.js
@@ -14,7 +14,7 @@ export default async (app) => {
   if (accountsServer && accountsGraphQL) {
     return { accountsServer, accountsGraphQL };
   }
-  const { MONGO_URL, STORE_URL, TOKEN_SECRET } = config;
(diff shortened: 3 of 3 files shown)


sujithvn (dismissed):
I did not test this, but the changes looks good

brent-hoover (dismissed):

vanpho93 (approved):

PR #6841, opened 2023-05-19
Add AccountsJS configurations as env vars

tedraykov:
Impact: **minor**
Type: **feature**

## Issue

Developers can't customize the JWT access and refresh tokens' expiration date. Also, there is no way to configure whether the reset password mutation returns new auth tokens or not.

## Solution

This PR adds environmental variables that allow to configure the aforementioned accounts js server properties, i.e. the jwt access and refresh token expiration time and the password reset mutation return type.

## Breaking changes

None. The default values of the env vars are identical to the accounts js library default values.

### Testing the password reset returned tokens:
1. Set the `ACCOUNTS_JS_RETURN_TOKENS_AFTER_RESET_PASSWORD` env var to `true`.
2. Run the password reset workflow and run the `resetPassword` mutation with the `tokens` in the return body like so:
```graphql
mutation resetPassword($token: String!, $newPassword: String!) {
  resetPassword(token: $token, newPassword: $newPassword) {
    tokens {
      accessToken
      refreshToken
    }
  }
}
```
If the env var is set to true, you should see the newly generated access and refresh tokens in the response
If the env var is set to false, you should see `null` in the response

### Testing the jwt token expiration time:
1. Set the `ACCOUNTS_JS_ACCESS_TOKEN_EXPIRES_IN` to `7d` for example
2. Authenitcate with the `authenticate` mutation or with the Kinetic Admin
3. Overserve the generated jwt token expiration date after you decoded it:
Screenshot 2023-05-19 at 11 51 16
Screenshot 2023-05-19 at 11 51 26

### DIFF
--- a/packages/api-plugin-authentication/src/config.js
+++ b/packages/api-plugin-authentication/src/config.js
@@ -1,10 +1,13 @@
 import envalid from "envalid";
 
-const { str } = envalid;
+const { str, bool } = envalid;
 
 export default envalid.cleanEnv(
   process.env,
   {
+    ACCOUNTS_JS_RETURN_TOKENS_AFTER_RESET_PASSWORD: bool({ default: false }),
+    ACCOUNTS_JS_ACCESS_TOKEN_EXPIRES_IN: str({ default: "90m" }),
+    ACCOUNTS_JS_REFRESH_TOKEN_EXPIRES_IN: str({ default: "30d" }),
     PASSWORD_RESET_PATH_FRAGMENT: str({ default: "?resetToken=" }),
     STORE_URL: str({ devDefault: "http://localhost:4000" }),
     TOKEN_SECRET: str({ default: "UPDATE_THIS_SECRET" })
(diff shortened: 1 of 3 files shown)


vanpho93 (approved):

packages/api-plugin-authentication/src/config.js
```diff
@@ -1,10 +1,13 @@
...
 
-const { str } = envalid;
+const { str, bool } = envalid;
 
 export default envalid.cleanEnv(
   process.env,
   {
+    ACCOUNTS_JS_RETURN_TOKENS_AFTER_RESET_PASSWORD: bool({ default: false }),
+    ACCOUNTS_JS_ACCESS_TOKEN_EXPIRES_IN: str({ default: "90m" }),
+    ACCOUNTS_JS_REFRESH_TOKEN_EXPIRES_IN: str({ default: "30d" }),
```
sujithvn, line 10: We are providing 30d as default for refresh-token, but the default in accountsjs is 7d [link](https://github.com/accounts-js/accounts/blob/ff983b8f9603f6295f2b7c9830a7926bc36e81c4/packages/server/src/accounts-server.ts#L51)

sujithvn (approved):
All good, just my comment on the default for refresh-token expiry

